PR #33, opened 2026-05-02
fix: voice TUI must rerender stale systemd unit + clearer audio errors

andrewagrahamhodges:
## The bug

`phantombot voice` writes the API key to `~/.config/phantombot/.env` and restarts the service — but a unit on disk from before Phase 29 lacks `EnvironmentFile=`, so the key never reaches `process.env`. Telegram then says:

> (voice messages need OpenAI or ElevenLabs configured — current provider is 'elevenlabs')

…which lies about the real cause: the provider is fine, the key just never landed in the service environment.

## Fix 1 — rerender stale unit before restart

`ensureUnitCurrent({ unitPath, binPath, systemctl })` in `src/lib/systemd.ts` compares the on-disk unit to `generateSystemdUnit({ binPath, args: ["run"] })`. If absent or different: rewrite + `systemctl --user daemon-reload`. Wired through:

- New `ServiceControl.rerenderUnitIfStale()` method (default impl uses `process.execPath`, `defaultUnitPath()`, real systemctl).
- `maybePromptRestart()` in `src/cli/harness.ts` now calls a new `maybeUpgradeUnit(svc)` helper first, then proceeds to the existing restart prompt. Prints `"systemd unit upgraded to current template"` when rerender fired.
- Same wiring covers `phantombot voice`, `phantombot telegram`, `phantombot harness`, and `phantombot embedding` because they all share `maybePromptRestart`.

## Fix 2 — honest diagnostics

New `AudioSupport` tri-state in `src/lib/audio.ts`:

```ts
export type AudioSupport =
  | { ok: true }
  | {
      ok: false;
      reason: "provider_none" | "provider_no_stt" | "key_missing";
      provider: VoiceProvider;
      envVar?: string;
    };
```

`ttsSupport()` and `sttSupport()` return it. `sttSupported()` / `ttsSupported()` are kept as `.ok` boolean wrappers (no caller churn for non-diagnostic uses).

`src/channels/telegram.ts` replaces the single message with `voiceUnavailableMessage()`:

- `provider_none` → `"voice transcription is disabled — run \`phantombot voice\` to set up OpenAI or ElevenLabs"`
- `provider_no_stt` → ``"current provider 'X' has no STT — switch via `phantombot voice`"``
- `key_missing` → ``"voice key not loaded into the service environment — run `phantombot install` to upgrade the systemd unit, then try again. (provider 'X', expected env var Y)"``

## Test plan

- [x] `bun test` — 335 pass / 1 skip / 0 fail (was 323; +12 new tests)

New tests:
- `ensureUnitCurrent`: identical → no rerender; missing → rerender + daemon-reload; one-byte diff → rerender + daemon-reload; pre-Phase-29 unit lacking `EnvironmentFile=` → rerender adds it.
- `AudioSupport`: every variant for every provider (`none`, `azure_edge`, `openai` ±key, `elevenlabs` ±key), including env-var name in the `key_missing` payload.
- `telegram.ts`: `key_missing` diagnostic includes provider and env var name and points at `phantombot install`. Existing azure_edge test updated to the new `provider_no_stt` wording.
- `voice.ts`: pre-creates a stale unit lacking `EnvironmentFile=`, drives `maybeUpgradeUnit` against a real on-disk file, asserts the rewrite happened (file now contains `EnvironmentFile=`) and that the simulated restart sees the upgraded unit (call order: `rerender` → `restart`).

## Out of scope

- `phantombot doctor` sub-command
- Misleading "current provider" wording elsewhere

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -18,8 +18,9 @@
 import type { Config } from "../config.ts";
 import type { Harness } from "../harnesses/types.ts";
 import {
+  type AudioSupport,
+  sttSupport,
   synthesize,
-  sttSupported,
   transcribe,
   ttsSupported,
 } from "../lib/audio.ts";
@@ -343,10 +344,11 @@ export async function runTelegramServer(
       // For voice messages: download → transcribe → use the transcript
       // as the user message before invoking the harness.
       if (isVoice && msg.voice) {
-        if (!sttSupported(input.config)) {
+        const stt = sttSupport(input.config);
+        if (!stt.ok) {
           await input.transport.sendMessage(
             msg.chatId,
-            `(voice messages need OpenAI or ElevenLabs configured — current provider is '${input.config.voice.provider}')`,
+            voiceUnavailableMessage(stt),
           );
           continue;
         }
@@ -478,4 +480,24 @@ export async function runTelegramServer(
       });
     }
   } while (!input.oneShot);
+}
+
+/**
+ * Render an honest, actionable explanation when sttSupport() rules a
+ * voice message out. Each variant points at the specific user action that
+ * fixes it, instead of the old single-message catch-all that misled
+ * users into thinking their provider was wrong when actually the systemd
+ * unit was stale.
+ */
+export function voiceUnavailableMessage(
+  s: Extract<AudioSupport, { ok: false }>,
+): string {
+  if (s.reason === "provider_none") {
+    return "voice transcription is disabled — run `phantombot voice` to set up OpenAI or ElevenLabs";
+  }
+  if (s.reason === "provider_no_stt") {
+    return `current provider '${s.provider}' has no STT — switch via \`phantombot voice\``;
+  }
+  // key_missing
+  return `voice key not loaded into the service environment — run \`phantombot install\` to upgrade the systemd unit, then try again. (provider '${s.provider}', expected env var ${s.envVar})`;
 }
--- a/src/cli/harness.ts
+++ b/src/cli/harness.ts
@@ -130,13 +130,17 @@ export async function runHarness(input: RunInput = {}): Promise<number> {
 }
 
 /**
- * Shared post-apply hook for the config-mutating TUIs. If phantombot is
- * currently running under systemd, offer to restart it inline so the
- * change takes effect.
+ * Shared post-apply hook for the config-mutating TUIs.
+ *
+ * Two steps. Always: re-render the on-disk systemd unit if it's stale (a
+ * pre-Phase-29 unit lacks `EnvironmentFile=` and silently swallows the
+ * .env secrets the TUI just wrote). Then: if phantombot is running, offer
+ * to restart it inline so the change takes effect.
  */
 export async function maybePromptRestart(
   svc: ServiceControl,
 ): Promise<void> {
+  await maybeUpgradeUnit(svc);
   if (!(await svc.isActive())) return;
   const restart = await p.confirm({
     message: "phantombot is currently running. Restart to apply changes?",
@@ -154,6 +158,21 @@ export async function maybePromptRestart(
     r.ok ? "restarted" : `restart failed: ${r.stderr ?? "unknown"}`,
     "Restart",
   );
+}
+
+/**
+ * Re-render the installed systemd unit if it's stale; print a one-line
+ * notice when it happened. Exposed so tests can verify the rewrite path
+ * without going through the @clack confirm prompt in maybePromptRestart.
+ */
+export async function maybeUpgradeUnit(
+  svc: ServiceControl,
+): Promise<{ rerendered: boolean }> {
+  const r = await svc.rerenderUnitIfStale();
+  if (r.rerendered) {
+    p.note("systemd unit upgraded to current template", "Unit");
+  }
+  return r;
 }
 
 export default defineCommand({
--- a/src/lib/systemd.ts
+++ b/src/lib/systemd.ts
@@ -6,9 +6,9 @@
  */
 
 import { existsSync } from "node:fs";
-import { mkdir, unlink, writeFile } from "node:fs/promises";
+import { mkdir, readFile, unlink, writeFile } from "node:fs/promises";
 import { homedir } from "node:os";
-import { dirname, join } from "node:path";
+import { basename, dirname, join } from "node:path";
 import type { WriteSink } from "./io.ts";
 
 export const PHANTOMBOT_UNIT_NAME = "phantombot.service";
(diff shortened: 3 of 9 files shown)


src/lib/systemd.ts
```diff
@@ -209,6 +209,64 @@ export interface ServiceControl {
...
+ */
+async function defaultRerenderUnitIfStale(): Promise<{ rerendered: boolean }> {
+  const binPath = process.execPath;
+  if (basename(binPath) !== "phantombot") return { rerendered: false };
+  const unitPath = defaultUnitPath();
+  if (!existsSync(unitPath)) return { rerendered: false };
+  const sysEnv = ensureUserSystemdEnv();
+  if (!sysEnv.ready) return { rerendered: false };
+  const systemctl = new BunSystemctlRunner(buildSystemctlEnv(sysEnv));
+  return ensureUnitCurrent({ unitPath, binPath, systemctl });
```
andrewagrahamhodges, line 269: Three-gate guard is right: `basename(binPath) !== "phantombot"` skips dev (where `process.execPath` is `bun`), `!existsSync(unitPath)` avoids presuming an install the user never asked for, and `!sysEnv.ready` avoids running daemon-reload on a host with no user-systemd bus. All three are silent no-ops by design — that's correct.

One implication worth knowing: developers running `bun cli ... voice` will *not* trigger the rerender path, so any future regression in `defaultRerenderUnitIfStale` itself only manifests on installed binaries. The pure `ensureUnitCurrent` is well-tested, but this defaulting wrapper is not. Acceptable — the wrapper is mostly plumbing — but if you ever change the gates, add a tiny test that injects fake `process.execPath` / fs / runner and exercises each branch.

src/lib/systemd.ts
```diff
@@ -209,6 +209,64 @@ export interface ServiceControl {
...
+  });
+  let current: string | undefined;
+  if (existsSync(opts.unitPath)) {
+    current = await readFile(opts.unitPath, "utf8");
+  }
+  if (current === expected) return { rerendered: false };
+  await mkdir(dirname(opts.unitPath), { recursive: true });
+  await writeFile(opts.unitPath, expected, "utf8");
+  await opts.systemctl.run(["--user", "daemon-reload"]);
+  return { rerendered: true };
```
andrewagrahamhodges, line 251: Worth knowing: this overwrites any user customization to the unit file. In practice that's fine — users shouldn't hand-edit `~/.config/systemd/user/phantombot.service` because phantombot owns it — but the only way they'll learn that today is by losing an edit silently. Consider a small follow-up: when `current !== expected`, log the diff (or at least the count of bytes added/removed) before overwriting. Not blocking.

src/cli/harness.ts
```diff
@@ -156,6 +160,21 @@ export async function maybePromptRestart(
...
+ */
+export async function maybeUpgradeUnit(
+  svc: ServiceControl,
+): Promise<{ rerendered: boolean }> {
+  const r = await svc.rerenderUnitIfStale();
+  if (r.rerendered) {
+    p.note("systemd unit upgraded to current template", "Unit");
+  }
+  return r;
+}
```
andrewagrahamhodges, line 176: Good split. Pulling the rerender step out of `maybePromptRestart` into its own exported helper gives tests a path that doesn't need to mock `@clack/prompts`. The fact you can hand-write an integration test in `cli-voice.test.ts` against `maybeUpgradeUnit` *because of this split* is exactly the payoff.

tests/cli-voice.test.ts
```diff
@@ -97,3 +105,74 @@ describe("applyVoiceConfig — none", () => {
...
+        callOrder.push("rerender");
+        return ensureUnitCurrent({ unitPath, binPath: BIN, systemctl: sys });
+      },
+    };
+
+    // Drive through the shared upgrade-unit helper. (maybePromptRestart
+    // would block on @clack's confirm prompt in a non-TTY test runner;
+    // maybeUpgradeUnit is the prompt-free part of the flow that owns the
+    // rerender step.)
+    const r = await maybeUpgradeUnit(svc);
```
andrewagrahamhodges, line 156: **Test gap — worth a follow-up, not a blocker.**

The spec asked for an integration test that asserts the unit was rewritten **before** restart was called *via the actual `maybePromptRestart` flow*. This test asserts ordering on a `ServiceControl` whose `rerenderUnitIfStale` and `restart` callbacks happen to be invoked in that order — but the wiring being tested is the test's own `callOrder.push(...)` instrumentation, not `maybePromptRestart` itself. If a future refactor swaps the two lines in `maybePromptRestart`, this test still passes.

The @clack confirm prompt is the obstacle. Two ways to close the gap:

1. Inject the confirm function as a parameter (default `p.confirm`, override in tests).
2. Spy on the order of `svc.rerenderUnitIfStale` vs `svc.restart` calls **made by `maybePromptRestart`**, by passing a `ServiceControl` that records the call order and a stubbed confirm that auto-resolves true.

Either way, the assertion you actually want is: *after `maybePromptRestart` returns, `rerender` was called before `restart`*. Today's test asserts that on a flow you wrote inline in the test.

src/channels/telegram.ts
```diff
@@ -479,3 +481,23 @@ export async function runTelegramServer(
...
+  s: Extract<AudioSupport, { ok: false }>,
+): string {
+  if (s.reason === "provider_none") {
+    return "voice transcription is disabled — run `phantombot voice` to set up OpenAI or ElevenLabs";
+  }
+  if (s.reason === "provider_no_stt") {
+    return `current provider '${s.provider}' has no STT — switch via \`phantombot voice\``;
+  }
+  // key_missing
+  return `voice key not loaded into the service environment — run \`phantombot install\` to upgrade the systemd unit, then try again. (provider '${s.provider}', expected env var ${s.envVar})`;
```
andrewagrahamhodges, line 502: Tiny TS hygiene nit: `envVar` is `string | undefined` on the type, but the `key_missing` branch always has it set in practice. If you ever pass a malformed payload, the user sees `expected env var undefined`. Could narrow `AudioSupport` to a per-reason discriminated union so `envVar` is required when `reason === "key_missing"` and absent otherwise. Not blocking — current shape works because the only producer (`audio.ts`) always supplies it.

andrewagrahamhodges (commented):
## Verdict: clean — merge it.

Matches the spec end-to-end. Both fixes land where I asked: `ensureUnitCurrent` is pure on its inputs, `defaultRerenderUnitIfStale` gates production usage with three sensible checks, the boolean wrappers preserve existing call sites, and the four-case `AudioSupport` payload includes the env-var name so the user gets a message that actually points at the fix.

Verified `installPhantombotUnit` (line 317) and `ensureUnitCurrent` (line 241) both render with `args: ["run"]` — so a fresh `phantombot install` and a rerender produce byte-identical units. No risk of an install/rerender ping-pong.

Test count is honest: 335 pass / 1 skip / 0 fail, +12 new.

Four inline notes below. None block merge. The only real one is the test-gap in `cli-voice.test.ts` — worth a follow-up but not a blocker. Tell Claude to merge.

PR #37, opened 2026-05-02
ci: GitHub Releases pipeline + baseline x64 build target

andrewagrahamhodges:
## Summary

Sets up the release feed `phantombot update` (PR B) will consume, and fixes the build-target bug that broke this morning's deploy to kai.

### What changes

- **`package.json`**: build script now targets `bun-linux-x64-baseline` (was plain `bun-linux-x64`). The supervisor box that runs kai is pre-AVX2 silicon — the non-baseline binary SIGILL'd on launch. Splits into `build:x64`, `build:arm64`, and `build` alias for `build:x64`.
- **`.github/workflows/release.yml`**: fires on every PR merge to `main`. Typecheck + test gate, then cross-compiles `bun-linux-x64-baseline` and `bun-linux-arm64` from a single `ubuntu-latest` runner (Bun cross-targets natively — no ARM runner needed). Publishes both binaries + `SHA256SUMS` to a GitHub Release tagged `v1.0.<PR_NUMBER>`.
- **`src/version.ts`**: single source of truth for the version string. `0.1.0-dev` in tree; CI overwrites with `1.0.<PR_NUMBER>` before build. Citty meta reads it.

### Versioning

Per request: `1.0.<PR_NUMBER>`. So when this PR merges as PR #N, the first release is `v1.0.N` — and the workflow's own self-test.

### Asset naming

- `phantombot-v1.0.N-linux-x64`
- `phantombot-v1.0.N-linux-arm64`
- `SHA256SUMS`

### Why every PR

Cron-friendly philosophy: `phantombot update --force --restart` (coming in PR B) wants always-fresh artifacts. If trivial PRs become noise we'll add a `release:skip` label gate later.

## Test plan

- [x] `bun test` — 335 pass / 1 skip / 0 fail (no tests added; pipeline is exercised by its own merge)
- [x] Local `bun run build` produces `dist/phantombot` reporting `0.1.0-dev`
- [ ] **Self-test on merge**: workflow run produces `v1.0.<this PR number>` with both binaries + checksums

## Out of scope (PR B)

- `phantombot update` command + TUI + `--check` / `--force` / `--restart` flags
- Cron-friendly arch auto-detect (`process.arch` → asset suffix)
- SHA256 verification of downloaded binary before atomic swap

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,100 @@
+name: Release on PR merge
+
+# Fires for every PR merged into main. Cuts a GitHub Release tagged
+# v1.0.<PR_NUMBER> with two cross-compiled binaries (x64-baseline + arm64)
+# and a SHA256SUMS file. `phantombot update` reads this feed.
+#
+# Why every PR (not manual tags): the user runs `phantombot update --force
+# --restart` from cron and wants always-fresh binaries. If a PR shouldn't
+# release, we'll add a `release:skip` label gate later.
+
+on:
+  pull_request:
+    types: [closed]
+    branches: [main]
+
+permissions:
+  contents: write   # create release + upload assets
+
+jobs:
+  release:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout merge commit
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
+
+      - name: Set up Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Typecheck
+        run: bun tsc --noEmit
+
+      - name: Test
+        run: bun test
+
+      - name: Compute version + bake into src/version.ts
+        id: version
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          VERSION="1.0.${PR_NUMBER}"
+          TAG="v${VERSION}"
+          printf 'export const VERSION = "%s";\n' "$VERSION" > src/version.ts
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "tag=$TAG" >> "$GITHUB_OUTPUT"
+
+      - name: Build x64 (baseline — no AVX2 required)
+        env:
+          TAG: ${{ steps.version.outputs.tag }}
+        run: |
+          mkdir -p dist
+          bun build --compile --target=bun-linux-x64-baseline \
+            ./src/index.ts \
+            --outfile "dist/phantombot-${TAG}-linux-x64"
+
+      - name: Build arm64
+        env:
+          TAG: ${{ steps.version.outputs.tag }}
+        run: |
+          bun build --compile --target=bun-linux-arm64 \
+            ./src/index.ts \
+            --outfile "dist/phantombot-${TAG}-linux-arm64"
+
+      - name: Compute SHA256SUMS
+        working-directory: dist
+        env:
+          TAG: ${{ steps.version.outputs.tag }}
+        run: |
+          sha256sum "phantombot-${TAG}-linux-x64" "phantombot-${TAG}-linux-arm64" > SHA256SUMS
+          cat SHA256SUMS
+
+      - name: Create GitHub Release with binary assets
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ steps.version.outputs.tag }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          gh release create "$TAG" \
+            --title "$TAG" \
+            --notes "Automated release for PR #${PR_NUMBER}: ${PR_TITLE}
+
+            Binaries:
+            - phantombot-${TAG}-linux-x64 — x86-64, baseline (no AVX2 required)
+            - phantombot-${TAG}-linux-arm64 — ARM64
+
+            Verify with: sha256sum -c SHA256SUMS
+
+            Install with: phantombot update" \
+            "dist/phantombot-${TAG}-linux-x64" \
+            "dist/phantombot-${TAG}-linux-arm64" \
+            "dist/SHA256SUMS"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -77,24 +77,37 @@ jobs:
           sha256sum "phantombot-${TAG}-linux-x64" "phantombot-${TAG}-linux-arm64" > SHA256SUMS
           cat SHA256SUMS
 
-      - name: Create GitHub Release with binary assets
+      - name: Write release notes to file
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TAG: ${{ steps.version.outputs.tag }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
           PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
+          cat > release-notes.md <<EOF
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+jobs:
+  release:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout merge commit
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
```
andrewagrahamhodges, line 28: Good — checking out `merge_commit_sha` (not `main`) means the binary is built from exactly the code that just merged, even if other PRs land mid-workflow. Pairs with the `--target` suggestion on line 87 to keep the git tag and the binary in lockstep.

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Set up Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: latest
```
andrewagrahamhodges, line 33: `bun-version: latest` makes releases non-reproducible. A Bun point-release with a regression in `--compile` will silently break every PR merge until upstream ships a fix — and you won't know which Bun version produced any historical binary. Pin to a major (`bun-version: 1.x`) or a known-good minor. Doesn't have to be exact-pinned, but it should at least be bounded.

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+        run: bun test
+
+      - name: Compute version + bake into src/version.ts
+        id: version
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          VERSION="1.0.${PR_NUMBER}"
+          TAG="v${VERSION}"
+          printf 'export const VERSION = "%s";\n' "$VERSION" > src/version.ts
```
andrewagrahamhodges, line 51: Heads-up: `printf > src/version.ts` clobbers the entire file, including the comment block currently explaining the contract. If anyone later adds a sibling const (release channel, build timestamp, anything), CI will silently delete it on the next merge and nothing will fail loudly. Two cleaner options:

- **Cheap**: `sed -i "s/0.1.0-dev/$VERSION/" src/version.ts` — replaces only the literal, leaves the comment intact.
- **Structural**: split the constant into its own one-line file (`src/version-string.ts`) that's safe to truncate, and leave `version.ts` for any future code that imports it.

Follow-up — not blocking.

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "tag=$TAG" >> "$GITHUB_OUTPUT"
+
+      - name: Build x64 (baseline — no AVX2 required)
+        env:
+          TAG: ${{ steps.version.outputs.tag }}
+        run: |
+          mkdir -p dist
+          bun build --compile --target=bun-linux-x64-baseline \
```
andrewagrahamhodges, line 60: `bun-linux-x64-baseline` is the right call — exactly the SIGILL-on-pre-AVX2 hardware problem from this morning's deploy. Suggest adding a one-liner to the comment block at the top of the file explaining *why* baseline is mandatory, so future-you (or a contributor) doesn't 'optimise' it back to `bun-linux-x64` and break the supervisor box again.

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+          cat SHA256SUMS
+
+      - name: Create GitHub Release with binary assets
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ steps.version.outputs.tag }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          gh release create "$TAG" \
```
andrewagrahamhodges: **Race condition.** Without `--target`, `gh release create` tags whatever HEAD of `main` happens to be when *this step* runs — which is not guaranteed to be the merge commit you built from. If a second PR merges between this workflow firing and this step executing, `v1.0.<N>`'s git tag will point at PR `<N+1>`'s commit. The uploaded binaries are still correct (line 28 checks out `merge_commit_sha`), but `git checkout v1.0.N` will lie and so will GitHub's 'this release was cut from commit X' UI.

Fix:
```yaml
gh release create "$TAG" \
  --target ${{ github.event.pull_request.merge_commit_sha }} \
  --title "$TAG" \
  ...
```

Worth fixing before merge — two-character change, eliminates a class of subtle bug that's annoying to debug after the fact.

.github/workflows/release.yml
```diff
@@ -0,0 +1,100 @@
...
+            --title "$TAG" \
+            --notes "Automated release for PR #${PR_NUMBER}: ${PR_TITLE}
+
+            Binaries:
+            - phantombot-${TAG}-linux-x64 — x86-64, baseline (no AVX2 required)
+            - phantombot-${TAG}-linux-arm64 — ARM64
+
+            Verify with: sha256sum -c SHA256SUMS
+
+            Install with: phantombot update" \
```
andrewagrahamhodges: The notes body keeps its YAML indent — after Bash strips the 10-space block indent of the `run: |` scalar, GitHub receives the body with 2 leading spaces on every line below the first. Two leading spaces doesn't trigger Markdown's indented-code-block heuristic (needs 4+), so it *renders*, but the bullets sit awkwardly indented and the structure looks off in the release UI.

Two fixes:

- **Cheap**: dedent lines 91-97 so they sit at the same column as the first character after `--notes "`.
- **Cleaner**: write notes to a file with a heredoc, then `--notes-file release-notes.md`. Easier to grow past two bullets later, and the indent stops mattering entirely.

Worth fixing before merge while you're in the file.

src/version.ts
```diff
@@ -0,0 +1,14 @@
...
+ * file before `bun build --compile` with `1.0.<PR_NUMBER>` so the binary
+ * baked into a GitHub Release prints its real version. See
+ * .github/workflows/release.yml.
+ *
+ * `phantombot update` reads VERSION from this constant to decide whether
+ * a newer release is available, so anything that bakes a wrong value
+ * here will produce wrong update prompts.
+ */
+
+export const VERSION = "0.1.0-dev";
```
andrewagrahamhodges, line 14: `1.0.<PR_NUMBER>` isn't really semver — `1.0.42` is a 'patch' of `1.0.41` only by coincidence (PRs aren't ordered by semantic impact), and there's no path to ever bumping major or minor without breaking the scheme. That's fine for now (you're the only consumer of the update feed), but worth adding a sentence to the doc comment above acknowledging that this is intentionally PR-numbered, not semver — so future-you doesn't try to bolt semver-aware logic onto `phantombot update` (e.g. 'this is a major upgrade, read these notes first') and find that the version string can't carry that information.

andrewagrahamhodges (commented):
Solid PR. The baseline-x64 fix is right (and the SIGILL post-mortem comment in the description is the kind of thing future-me will thank present-me for). Inline comments below split into two buckets:

**Worth fixing before merge (both 2-line diffs):**
- `--target ${{ github.event.pull_request.merge_commit_sha }}` on `gh release create` — without it, the git tag can point at the wrong commit if a second PR merges fast.
- Dedent the `--notes` body — currently lands with 2 leading spaces on every line in the rendered release notes.

**Follow-ups, not blockers:**
- Pin `bun-version` (currently `latest` — non-reproducible).
- The `printf > src/version.ts` overwrite is fragile if anything else ever lives in that file.
- `1.0.<PR_NUMBER>` isn't really semver — fine for now, worth a comment acknowledging it.

**One thing to verify the hard way before relying on it:** does a merged fork PR get a write-scoped `GITHUB_TOKEN` for this workflow? `pull_request` events from forks usually hand out a read-only token, and merged-state may not change that. If it doesn't work, the fix is `on: push: branches: [main]` plus parsing the PR number from the merge commit subject. Not relevant until you accept your first external contribution — but worth knowing it's untested.

Verdict: fix the two before-merge items, ship it, log the rest as follow-ups.

Issue #42, opened 2026-05-02
release.yml: verify GITHUB_TOKEN scope on merged fork PRs (untested)


andrewagrahamhodges:
Non-blocking follow-up from PR #37 review.

\`pull_request\` events from forks usually hand out a read-only \`GITHUB_TOKEN\`, and merged-state may not change that. The release workflow needs \`contents: write\` to create a release; if a merged fork PR doesn't get that scope, the workflow will silently fail to publish a release.

Not relevant until the first external contribution. When that happens:

- If it works: log a comment in the workflow noting fork PRs were verified.
- If it doesn't: switch trigger to \`on: push: branches: [main]\` and parse the PR number from the merge commit subject (squash-merge commits include \`(#N)\`).

Touch \`.github/workflows/release.yml\` only when verifying / fixing.

kaieriksen:
Closing as obsolete — no external contributors yet, and the trigger can be switched to `push: branches: [main]` when needed. Not worth tracking until the first fork PR lands.

Issue #38, opened 2026-05-02
release.yml: pin bun-version (currently 'latest', non-reproducible)


andrewagrahamhodges:
Non-blocking follow-up from PR #37 review (\`.github/workflows/release.yml\` line 33).

\`bun-version: latest\` makes releases non-reproducible. A Bun point-release with a regression in \`--compile\` will silently break every PR merge until upstream ships a fix — and we won't know which Bun version produced any historical binary.

Pin to a major (\`bun-version: 1.x\`) or a known-good minor. Doesn't have to be exact-pinned, but it should at least be bounded.

Touch \`.github/workflows/release.yml\` only.

andrewagrahamhodges:
Resolved in PR #45.

`.github/workflows/release.yml` pins `bun-version: 1.x` (was `latest`). Bounds the toolchain to a major while still picking up bugfixes within Bun 1.x.

PR #43, opened 2026-05-02
ci: pre-create ~/.config/systemd/user before tests (hotfix)

andrewagrahamhodges:
## What

Adds one workflow step before `bun test`:

```yaml
- name: Pre-create ~/.config/systemd/user (hotfix; see #44)
  run: mkdir -p "${HOME:?HOME unset}/.config/systemd/user"
```

## Why

PR #37 merged successfully but its self-test workflow run **failed** — 4 install tests blew up on the GH runner that pass cleanly locally:

- `installPhantombotUnit > writes the unit file and runs daemon-reload, enable, start`
- `installPhantombotUnit > aborts on systemctl failure`
- `runInstall > auto-set message printed when systemd env is auto-detected`
- `runInstall > happy path writes unit + runs reload/enable/start, returns 0`

Root cause (per review of `src/lib/systemd.ts`): `installPhantombotUnit` honors `opts.unitPath` for the main unit but unconditionally calls `heartbeatServicePath()` etc. — those four functions all hardcode `homedir() + .config/systemd/user/...`. Tests only set `unitPath` (in `mkdtemp`), so the heartbeat/nightly writes leak to the real home dir. That dir exists on dev boxes but not on a fresh GitHub runner.

Net effect today: **no `v1.0.37` release was produced**. PRs #38–#42 won't auto-release either if not fixed.

## Why a hotfix instead of the proper fix

The proper fix is to parameterize `heartbeatServicePath` etc. in `InstallOptions` so tests can pass tmpdir paths — and stop polluting dev machines too. Tracked as **#44**. Done as a separate PR so the release pipeline isn't blocked while we get the refactor right.

## Test plan

- [x] `bun test` — 335 pass / 1 skip / 0 fail (the hotfix doesn't change runtime behavior, only the CI environment)
- [ ] **Self-test on merge**: workflow run produces `v1.0.<this PR number>` with both binaries + checksums

## Follow-up

- **#44** — installPhantombotUnit pollutes real ~/.config/systemd/user/ during tests (proper fix; deletes this hotfix step in the same PR)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,10 +42,12 @@ jobs:
       # heartbeat/nightly unit files to a hardcoded ~/.config/systemd/user/
       # path even when tests pass an isolated tmpdir for the main unit. The
       # dir exists on a developer's box but not on a fresh runner, so writeFile
-      # fails with ENOENT. Pre-creating the dir unblocks the pipeline; the
-      # proper fix (parameterizing the helper paths) is tracked as a follow-up.
-      - name: Pre-create ~/.config/systemd/user (hotfix; see issue tracker)
-        run: mkdir -p "$HOME/.config/systemd/user"
+      # fails with ENOENT. The pollution side of the same bug (real files left
+      # behind in dev `~/.config/systemd/user/` after every `bun test` run) is
+      # tracked separately. When the proper fix lands, delete this step.
+      # Tracked: https://github.com/andrewagrahamhodges/phantombot/issues/44
+      - name: Pre-create ~/.config/systemd/user (hotfix; see #44)
+        run: mkdir -p "${HOME:?HOME unset}/.config/systemd/user"
 
       - name: Test
         run: bun test
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,6 +38,15 @@ jobs:
       - name: Typecheck
         run: bun tsc --noEmit
 
+      # Hotfix for a latent test-isolation bug: installPhantombotUnit writes
+      # heartbeat/nightly unit files to a hardcoded ~/.config/systemd/user/
+      # path even when tests pass an isolated tmpdir for the main unit. The
+      # dir exists on a developer's box but not on a fresh runner, so writeFile
+      # fails with ENOENT. Pre-creating the dir unblocks the pipeline; the
+      # proper fix (parameterizing the helper paths) is tracked as a follow-up.
+      - name: Pre-create ~/.config/systemd/user (hotfix; see issue tracker)
+        run: mkdir -p "$HOME/.config/systemd/user"
+
       - name: Test
         run: bun test
 


.github/workflows/release.yml
```diff
@@ -38,6 +38,15 @@ jobs:
       - name: Typecheck
         run: bun tsc --noEmit
 
+      # Hotfix for a latent test-isolation bug: installPhantombotUnit writes
+      # heartbeat/nightly unit files to a hardcoded ~/.config/systemd/user/
+      # path even when tests pass an isolated tmpdir for the main unit. The
+      # dir exists on a developer's box but not on a fresh runner, so writeFile
+      # fails with ENOENT. Pre-creating the dir unblocks the pipeline; the
+      # proper fix (parameterizing the helper paths) is tracked as a follow-up.
```
andrewagrahamhodges: **`see issue tracker` is a forward reference to nothing.** The PR description says "Will file" — but if this merges first, the issue often doesn't get filed, or gets filed without a link back here, and the comment becomes a dangling pointer. Two minutes of friction now saves a future maintainer a 20-minute archaeology dig. File the issue first, then s/see issue tracker/see #NN/ in this comment. Same applies in the PR body's "Follow-up issue" section.

.github/workflows/release.yml
```diff
@@ -38,6 +38,15 @@ jobs:
...
+      - name: Pre-create ~/.config/systemd/user (hotfix; see issue tracker)
```
andrewagrahamhodges: **This hotfix only addresses CI; the underlying test pollutes real dev machines too.** Confirmed by reading `installPhantombotUnit` in `src/lib/systemd.ts`: it calls `heartbeatServicePath()` / `heartbeatTimerPath()` / `nightlyServicePath()` / `nightlyTimerPath()` unconditionally — those four functions all `return join(homedir(), ".config", "systemd", "user", ...)`. The test's `mkdtemp` only isolates `unitPath`, so every `bun test` run on a developer's box creates four real files in `~/.config/systemd/user/` that the `afterEach` (which only removes `workdir`) never cleans up.

Not a blocker for this PR — the hotfix does what it claims for CI — but the follow-up issue should be framed as **"installPhantombotUnit pollutes real ~/.config/systemd/user/ during tests"** rather than just "parameterize the paths." The pollution is the harm; parameterization is one possible fix.

Proper fix: extend `InstallOptions` with `heartbeatServicePath?: string` etc. (defaults to the current functions), have tests pass `join(workdir, ...)` for all four, and delete this YAML hotfix in the same PR.

andrewagrahamhodges: Minor: `mkdir -p "$HOME/.config/systemd/user"` is fine, but `$HOME` is theoretically unset on some runner configs (rare on `ubuntu-latest`, never seen it, but it's a 0-arg-to-`mkdir` footgun if it ever happens). Either ignore — vanishingly unlikely on hosted runners — or use `mkdir -p "${HOME:?HOME unset}/.config/systemd/user"` to fail loudly instead of silently doing nothing. Take it or leave it.

andrewagrahamhodges (commented):
**Verdict: ship the hotfix, but it's papering over a worse latent bug than the description admits — a follow-up issue should be filed before this merges, not after.**

Diagnosis is dead right — verified against `src/lib/systemd.ts`. `installPhantombotUnit` honors `opts.unitPath` for the main unit but then calls `heartbeatServicePath()` / `heartbeatTimerPath()` / `nightlyServicePath()` / `nightlyTimerPath()`, all of which hardcode `homedir() + .config/systemd/user/...`. So the test's `mkdtemp` isolation is half-applied: main unit goes to tmpdir, heartbeat/nightly units leak to the real `~/.config/systemd/user/`.

**The bigger problem this hotfix doesn't address:** running `bun test` on any developer's machine actively pollutes their real `~/.config/systemd/user/` with `phantombot-heartbeat.service`, `phantombot-heartbeat.timer`, `phantombot-nightly.service`, and `phantombot-nightly.timer`. The `afterEach` only `rm`s `workdir`, so those four files stay behind every test run. On a dev box with `systemctl --user daemon-reload` happening for unrelated reasons, those stray unit files are now visible to systemd. Not catastrophic — they're inert until enabled — but it's a real footprint left by the test suite. CI hotfix makes the symptom go away; the bug stays.

**Recommendation:** merge this to unblock the release pipeline (you've already lost `v1.0.37`), but file the follow-up issue *before* merging and reference its number in the YAML comment instead of `see issue tracker`. That way the proper fix doesn't get lost the moment CI goes green.

**Praise:** honest hotfix framing in the PR description, the YAML comment explains intent so future maintainers don't strip it as dead code, and pre-creating the dir is the smallest possible CI-only patch. Right call for right now.

PR #47, opened 2026-05-02
fix: applyUpdate clears stale .bak before copyFile

andrewagrahamhodges:
## Bug repro (real, from kw-openclaw)

\`\`\`
$ phantombot update
…
downloading phantombot-v1.0.46-linux-x64…
verified 96.6 MB (sha256 ok).
install failed: failed to back up /home/kai/.local/bin/phantombot:
EACCES: permission denied, copyfile
'/home/kai/.local/bin/phantombot' -> '/home/kai/.local/bin/phantombot.bak'
\`\`\`

## Cause

The initial deploy script for kai used \`sudo cp /home/kai/.local/bin/phantombot /home/kai/.local/bin/phantombot.bak\`. \`sudo cp\` creates the destination owned by **root**, so kai's later \`phantombot update\` couldn't overwrite the existing \`.bak\`.

Why \`copyFile\` fails: it opens the destination with \`O_TRUNC\`, which requires write permission **on the existing file**. The parent dir being kai-writable isn't enough — Linux checks the file's own mode/ownership for truncation.

\`checkWritable\` only verifies the parent dir, so it doesn't catch this.

## Fix

\`applyUpdate\` now \`rm -f\`s the \`.bak\` before \`copyFile\`. Unlink takes its permission from the parent dir (write+exec), which we've already verified, so a foreign-owned stale \`.bak\` is cleared without elevated privileges.

## Test plan

- [x] Regression test pre-creates \`.bak\` at mode \`0o444\` (no write bit) and confirms \`applyUpdate\` still succeeds and the new \`.bak\` content is the previous live binary
- [x] \`bun test\` — 370 pass / 1 skip / 0 fail (was 369; +1 regression test)
- [ ] **Real-world**: after merge → \`v1.0.<this PR number>\` → kai runs \`phantombot update --force --restart\` (this time it succeeds)

## Note for the kw-openclaw box

Already cleared the stale \`.bak\` (and a leftover \`.update.tmp\`) by hand so testing isn't blocked while this lands. Once it merges, the next install on a fresh box can't repeat the failure mode.

### DIFF
--- a/src/lib/binaryUpdate.ts
+++ b/src/lib/binaryUpdate.ts
@@ -154,9 +154,22 @@ export type ApplyResult =
 
 /**
  * Atomic swap. Steps:
- *   1. Copy targetPath → ${targetPath}.bak (so a botched rename is recoverable).
- *   2. rename(tempPath, targetPath) — atomic on Linux even when targetPath
+ *   1. Remove a stale ${targetPath}.bak if one exists (best-effort) — see
+ *      below for why this isn't redundant with copyFile's overwrite.
+ *   2. Copy targetPath → ${targetPath}.bak (so a botched rename is
+ *      recoverable).
+ *   3. rename(tempPath, targetPath) — atomic on Linux even when targetPath
  *      is the running process's binary; the kernel uses inode, not path.
+ *
+ * Why we explicitly unlink the .bak first: copyFile opens the destination
+ * with O_TRUNC. Linux requires write permission on an existing file to
+ * truncate it, even if the parent dir is writable. So if a previous .bak
+ * was written by a different user (e.g. an earlier `sudo cp` during
+ * initial deploy left a root-owned .bak), running `phantombot update` as
+ * the unprivileged service user fails with EACCES — even though kai owns
+ * the dir and the live binary. unlink only needs write+execute on the
+ * parent dir, which we've already verified via checkWritable, so an old
+ * foreign-owned .bak can be cleared without elevated privileges.
  *
  * Returns the .bak path on success so the caller can tell the user where
  * to find the rollback if something downstream (e.g. service restart)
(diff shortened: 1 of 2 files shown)


andrewagrahamhodges (commented):
**Verdict: ship it. Smallest correct fix to a real bug.**

Diagnosis is exactly right. `copyFile` opens with `O_TRUNC`, which Linux permission-checks against the *file's* mode, not the parent dir. A foreign-owned (or in the test, 0o444 self-owned) `.bak` blocks the truncate, even when the dir and live binary are kai-owned. Pre-unlinking takes its permission from the parent dir, which `checkWritable` already verifies — so the new path is genuinely safer with no extra privilege.

Recoverability isn't weakened: between `rm(.bak)` and `copyFile(target → .bak)` the live binary at `targetPath` is untouched, so a crash in that window leaves a working binary on disk. The rollback only matters after `rename`, which still has its `.bak` by then.

Test reproduces the failure mode correctly. 0o444 owned by the current user *does* trigger EACCES on `O_TRUNC` (Linux applies the owner-bits, owner doesn't bypass the mode for writes — only for chmod). Without the fix, the test fails; with it, it passes. Valid regression.

**One small doc nit (don't block on it):**

The new comment says *"rm errors are swallowed because the copyFile that follows surfaces the real one"*. Not quite — `{ force: true }` only swallows ENOENT. Any other rm error (EACCES, EBUSY, EROFS) still propagates and gets caught by the existing `catch`, which then reports it as `"failed to back up ${targetPath}"`. Slightly misleading attribution, but the user still sees a real error. Worth either tightening the comment to *"ENOENT is swallowed; other errors fall through to the catch below"* or wrapping the rm in its own try with a distinct error message. Either way, follow-up — not a blocker.

**Praise:** good repro in the PR body (actual error pasted), the doc-comment in `binaryUpdate.ts` explains *why* the unlink isn't redundant with copyFile's overwrite (future-you would absolutely have deleted the rm thinking copyFile already handles it), and the test pre-creates `.bak` at exactly the mode that triggers the bug.

Merge it, then run the update on kw-openclaw to close the loop on the real-world tickbox.

PR #51, opened 2026-05-02
feat: consolidate persona CLI + auto-set default on import + install.sh

andrewagrahamhodges:
Four asks bundled into one cohesive PR (per the conversation):

1. **Consolidate** \`phantombot create-persona\` + \`phantombot import-persona\` into a single \`phantombot persona\` entry point.
2. **Fix** the bug where \`phantombot run\` failed with *"persona 'phantom' not found"* after a successful import.
3. **One-liner installer** at the repo root + the curl line at the top of the README install section.
4. **Persona switching** as both a TUI option and a positional arg.

### 1. New \`phantombot persona\` (src/cli/persona.ts)

\`\`\`
phantombot persona                       # TUI: current persona + create / import / restore / switch
phantombot persona <name>                # switch default persona to <name>
phantombot persona --import <dir>        # non-interactive import
phantombot persona --import <dir> --as N
\`\`\`

\`--import\` + positional \`<name>\` are mutually exclusive (exit 2 with hint).

The two old top-level subcommands are **removed** from the dispatcher. Their implementation files (\`create-persona.ts\`, \`import-persona.ts\`) stay in place — \`runCreatePersona\` / \`runImportPersona\` / \`applyRestore\` are still exported and used by the new \`persona.ts\` and the existing tests. Only the \`defineCommand\` defaults are dropped.

### 2. Auto-set \`default_persona\` on import (the bug fix)

\`\`\`
$ phantombot persona --import ~/clawd --as robbie
imported persona 'robbie' to /home/robbie/.local/share/phantombot/personas/robbie
copied (282):
  …
adopted 'robbie' as default_persona (previous default 'phantom' has no persona dir on disk)

$ phantombot run
phantombot — persona 'robbie', harnesses claude → pi
…
\`\`\`

Reproduces and fixes the user-reported bug:

> \`\`\`
> robbie@OpenClaw-Hodges:~$ phantombot run
> persona 'phantom' not found at /home/robbie/.local/share/phantombot/personas/phantom
> import or create one with \`phantombot import-persona <openclaw-dir>\` …
> \`\`\`

The new \`maybeAdoptAsDefault\` helper writes \`state.json\` only when the current default's directory doesn't exist — additive imports (where the user already has a working persona) **don't** shift the default. Tested both ways.

### 3. \`install.sh\` (repo root)

\`\`\`
curl -fsSL https://raw.githubusercontent.com/andrewagrahamhodges/phantombot/main/install.sh | sh
\`\`\`

- Arch detection (\`x86_64\` → x64, \`aarch64\` → arm64; refuses other archs)
- Fetches latest release tag from GH API (honors \`GITHUB_TOKEN\` if set)
- Downloads binary + \`SHA256SUMS\`, **verifies the checksum**, refuses on mismatch
- Installs to \`~/.local/bin/phantombot\` (mode 0755)
- Warns if PATH doesn't include the install dir
- Launches \`phantombot persona\` to set up the first persona — but **detects non-TTY** (\`curl … | sh\` case) and prints the next-step hint instead of trying to drive @clack
- Overrides: \`PHANTOMBOT_INSTALL_DIR=/some/path\`, \`PHANTOMBOT_SKIP_TUI=1\`

Verified with \`sh -n install.sh\` (no test framework for shell scripts; manual smoke test on a fresh VM is documented as the validation path).

### 4. Switching

\`phantombot persona <name>\` (or the TUI menu's "Switch" entry) writes \`state.default_persona\` and prompts for restart via the existing \`maybePromptRestart\` plumbing. Refuses with a clear "available: …" list when the target doesn't exist.

## Test plan

- [x] \`bun test\` — **442 pass / 1 skip / 0 fail** (was 434; +8 new tests)
- [x] \`sh -n install.sh\` clean
- [x] Local \`bun run build\` produces a binary; \`./dist/phantombot --help\` shows \`persona\` and **no longer shows** \`create-persona\` / \`import-persona\`
- [ ] **Self-test on merge**: workflow run produces \`v1.0.<this PR number>\` with both binaries
- [ ] **Real-world**: re-run the import on the OpenClaw-Hodges box (\`phantombot persona --import ~/clawd --as robbie\` → \`phantombot run\` should now Just Work)

## Files

**New**:
- \`src/cli/persona.ts\` — the consolidated entry
- \`install.sh\` — one-liner installer
- \`tests/cli-persona.test.ts\` — 5 new tests for switch + dispatch

**Modified**:
- \`src/cli/index.ts\` — drop create-persona / import-persona, add persona
- \`src/cli/create-persona.ts\` — drop \`defineCommand\` default, keep all functions
- \`src/cli/import-persona.ts\` — same; add \`maybeAdoptAsDefault\` for the bug fix
- \`tests/cli.test.ts\` — subcommand list updated
- \`tests/cli-import-persona.test.ts\` — +2 tests for adopt-on-fresh-box and additive-import-no-shift
- \`README.md\` — one-liner install becomes recommended; persona docs updated; first-import note added
- \`AGENTS.md\` — repo layout, two new pitfalls (auto-adopt semantics; install.sh TTY detection)

## Out of scope

- Hosting the install.sh on a custom domain (e.g. \`phantombot.dev\`) — the raw GH URL works for now; nothing prevents a domain redirect later
- Smoke-testing install.sh in CI — would need an isolated container with no Bun + a mock release; not worth the carry yet

### DIFF
--- a/src/cli/persona.ts
+++ b/src/cli/persona.ts
@@ -31,6 +31,7 @@ import {
   personaDir,
 } from "../config.ts";
 import type { WriteSink } from "../lib/io.ts";
+import { log } from "../lib/logger.ts";
 import { listArchives } from "../lib/personaArchive.ts";
 import { defaultServiceControl, type ServiceControl } from "../lib/systemd.ts";
 import { loadState, saveState } from "../state.ts";
@@ -260,6 +261,10 @@ async function runPersonaMenu(input: RunPersonaMenuInput): Promise<number> {
  * Read the personas directory and return the names of subdirectories
  * (each subdir = one persona). Returns [] if the personas dir doesn't
  * exist yet — fresh installs have no personas.
+ *
+ * Non-ENOENT read failures (EACCES, EIO, etc.) still return [] so the
+ * caller's UI keeps working, but they're logged to stderr first so the
+ * user has a hint that the empty list isn't the same as "no personas."
  */
 export function listExistingPersonas(config: Config): string[] {
   if (!existsSync(config.personasDir)) return [];
@@ -268,7 +273,11 @@ export function listExistingPersonas(config: Config): string[] {
       .filter((e) => e.isDirectory())
       .map((e) => e.name)
       .sort();
-  } catch {
+  } catch (e) {
+    log.warn("persona: failed to read personas dir", {
+      personasDir: config.personasDir,
+      error: (e as Error).message,
+    });
     return [];
   }
 }
--- a/src/cli/persona.ts
+++ b/src/cli/persona.ts
@@ -1,0 +1,314 @@
+/**
+ * `phantombot persona` — single entry point for persona management.
+ *
+ * Replaces the previous `phantombot create-persona` and
+ * `phantombot import-persona` top-level subcommands. The argument grammar:
+ *
+ *   phantombot persona                       — TUI: current persona + menu
+ *                                              (create / import / restore / switch)
+ *   phantombot persona <name>                — switch default persona to <name>
+ *                                              (positional; <name> must exist)
+ *   phantombot persona --import <dir>        — import non-interactively
+ *   phantombot persona --import <dir> --as N — import with explicit target name
+ *
+ * The `--import` flag and a positional <name> are mutually exclusive; pass
+ * `--as` with `--import` to set the target name instead of using a positional.
+ *
+ * The underlying create / import / restore work still lives in
+ * `src/cli/create-persona.ts` and `src/cli/import-persona.ts` — those files
+ * keep their `run*` exports for direct programmatic use and existing tests.
+ * Only their `defineCommand(...)` defaults are removed so phantombot's
+ * dispatcher only exposes the consolidated entry point.
+ */
+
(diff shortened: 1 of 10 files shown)


src/cli/persona.ts
```diff
@@ -0,0 +1,314 @@
...
+ */
+export function listExistingPersonas(config: Config): string[] {
+  if (!existsSync(config.personasDir)) return [];
+  try {
+    return readdirSync(config.personasDir, { withFileTypes: true })
+      .filter((e) => e.isDirectory())
+      .map((e) => e.name)
+      .sort();
+  } catch {
+    return [];
```
andrewagrahamhodges, line 281: nit: `listExistingPersonas` swallows all errors silently. If the personas dir exists but can't be read (e.g. EACCES), it returns `[]` and the TUI shows "no personas exist yet" — true on the surface but the real problem (permissions) is hidden. Consider logging non-ENOENT errors to stderr before returning the empty list, so the user has a clue if something deeper is wrong.

Not a gate — the common cases (missing dir, empty dir, normal read) are all fine.

PR #52, opened 2026-05-02
feat: add Google Gemini CLI harness (selectable via phantombot harness)

andrewagrahamhodges:
Wraps [google-gemini/gemini-cli](https://github.com/google-gemini/gemini-cli) as a third harness alongside Claude and Pi. Selectable as **primary or fallback** via \`phantombot harness\`. Default chain stays \`[\"claude\", \"pi\"]\` — gemini is opt-in.

Bundles the **buildHarnessChain extraction** (the duplicated factory in \`run.ts\` + \`tick.ts\` was about to spawn a third copy with this PR — retired into \`src/harnesses/buildChain.ts\` for one update site instead of three).

## Spawn shape

\`\`\`
gemini -p <new_user_message> -o text -y [-m <model>]
\`\`\`

- **\`-p\`** enables headless. Per the docs the value is *appended to whatever's on stdin* — so we send the system prompt + history via stdin and use \`-p\` for just the new user message. Avoids any ARG_MAX concern (Pi has the same problem and guards it with \`maxPayloadBytes\`; gemini's stdin+argv split sidesteps it).
- **\`-o text\`** for v1: collect all stdout, emit one \`text\` + one \`done\` chunk on exit 0. Stream-json upgrade is a follow-up — needs a real schema sample to write a parser against (none in \`gemini --help\`).
- **\`-y\`** (yolo) auto-approves all tool calls. Required for headless because the default mode prompts for approval and would block forever in a non-interactive subprocess. Same posture as Claude's \`--permission-mode bypassPermissions\`.
- **\`-m\`** only passed when \`config.gemini.model\` is non-empty; otherwise gemini-cli picks its own default.

## Auth (Pi-style lighter touch, NOT Claude-style strict filter)

We don't strip \`GEMINI_API_KEY\` / \`GOOGLE_API_KEY\` from the spawn env. If the user has a key in \`~/.env\`, gemini uses it. If they OAuth'd via gemini's interactive \`/auth\`, gemini uses that. Whichever wins is up to gemini-cli's own resolution.

Claude is the special case (we filter \`ANTHROPIC_API_KEY\`) because a stray env var there would silently swap OAuth for API-key billing. Gemini's two auth modes are roughly interchangeable for end-user behavior, so the heavy hand isn't warranted.

## No \`--system-prompt\` flag

Gemini-cli doesn't have one. The harness builds a transcript:

\`\`\`
<system_prompt>

User: prev_user_1
Assistant: prev_assistant_1
…
\`\`\`

…and the new user message is delivered via \`-p\`. Modern chat models recognize this format. Tested via the fake-gemini fixture.

## Files

**New**:
- \`src/harnesses/gemini.ts\` — the harness wrapper
- \`src/harnesses/buildChain.ts\` — extracted factory (was duplicated in run.ts + tick.ts)
- \`tests/fixtures/fake-gemini.sh\` — bash mock with normal / error / notfound / hang / echo-args modes
- \`tests/harnesses-gemini.test.ts\` — 12 tests (renderStdinPayload pure-function tests + 6 spawn-wiring tests + 3 available() variants)

**Modified**:
- \`src/cli/run.ts\` + \`src/cli/tick.ts\` — drop the duplicated \`buildHarnessChain\`; import from the new shared module
- \`src/cli/harness.ts\` — \`SUPPORTED_HARNESSES\` adds \"gemini\"; TUI binary-on-PATH availability check extended
- \`src/config.ts\` — Config type adds \`harnesses.gemini\`; loader reads \`[harnesses.gemini]\` with PHANTOMBOT_GEMINI_BIN / PHANTOMBOT_GEMINI_MODEL env overrides
- 12 inline Config fixtures across the test suite get the new \`gemini: { bin, model }\` slot
- \`README.md\` — mentions gemini in the opening line + adds it to prerequisites with both auth modes documented
- \`AGENTS.md\` — repo layout adds harnesses/buildChain.ts + gemini.ts; new common pitfall #9 documents the three-place harness-extension contract

## Test plan

- [x] \`bun test\` — **454 pass / 1 skip / 0 fail** (was 442; +12 new)
- [x] Local \`bun run build\` produces a binary; \`./dist/phantombot harness\` would show \`gemini\` as a selectable option (verified via SUPPORTED_HARNESSES growing)
- [ ] **Self-test on merge**: workflow run produces \`v1.0.<this PR number>\`
- [ ] **Real-world**: on the kw-openclaw box, \`phantombot harness\` should now offer gemini as a chain option; setting \`gemini\` as primary or fallback then sending a Telegram message should round-trip through gemini-cli (lena's account already has \`gemini\` installed per the help-output paste)

## Out of scope

- \`-o stream-json\` upgrade — needs a real schema sample. File a follow-up issue once we have one.
- Gemini-specific \`--include-directories\` / \`--extensions\` / \`--allowed-mcp-server-names\` plumbing — the harness contract is intentionally minimal; these are gemini-side configs the user can preload into \`~/.config/gemini-cli/\` if they want them.


src/harnesses/gemini.ts
```diff
@@ -0,0 +1,212 @@
...
+      return true;
+    } catch {
+      return false;
+    }
+  }
+
+  async *invoke(req: HarnessRequest): AsyncGenerator<HarnessChunk> {
+    const stdinPayload = renderStdinPayload(req);
+    const args: string[] = [
+      "-p", req.userMessage,
```
andrewagrahamhodges, line 100: 🟡 **No ARG_MAX guard on `req.userMessage`.**

The PR description says argv stays small because system+history go via stdin and only the new user message rides on `-p`. That's true for chat-sized inputs — but `req.userMessage` is unbounded from the harness's POV. Long pastes, voice transcriptions, or future channels can push it past Linux ARG_MAX (typically 128KiB–2MiB depending on kernel/build). When that happens, `Bun.spawn` will fail with E2BIG and the orchestrator's fallback can't pre-skip because Gemini doesn't declare `maxPayloadBytes` (Pi does — see `pi.ts:41`).

Two ways to handle it:
1. **Declare `maxPayloadBytes` on the harness** (sized for userMessage only, e.g. 100KB to stay well under conservative ARG_MAX). Then `orchestrator/fallback.ts` skips Gemini cleanly for oversized turns and the next harness in the chain handles it.
2. **Fold the user message into the stdin payload** instead of `-p` (e.g. append `User: <new>` after history) and pass a tiny constant like `-p "continue"` or even a single space to satisfy headless mode. Sidesteps ARG_MAX entirely. Worth checking whether gemini-cli treats the appended-stdin-as-prompt model the same way regardless of what's in `-p`.

Option 1 is the smaller change and matches Pi's pattern. Not a merge blocker if userMessages stay short in practice, but worth a follow-up issue at minimum.

tests/channels-telegram.test.ts
```diff
@@ -195,7 +195,8 @@ const baseConfig = (
   harnesses: {
     chain: ["claude"],
     claude: { bin: "claude", model: "opus", fallbackModel: "sonnet" },
-    pi: { bin: "pi", maxPayloadBytes: 1_000_000 },
+    pi: { bin: "pi", maxPayloadBytes:1_000_000 },
+      gemini: { bin: "gemini", model: "" },
```
andrewagrahamhodges: 🟡 **Auto-edit dropped a space + over-indented the new line.**

Two formatting nits that repeat across all 12 test fixtures touched by this PR:
- `maxPayloadBytes:1_000_000` — missing space after the colon (was `maxPayloadBytes: 1_000_000` before).
- `      gemini: { ... }` — six-space indent vs the surrounding four-space indent on `pi:` and `claude:`.

Looks like a sed/codemod that didn't run prettier afterwards. Same pattern in `tests/cli-create-persona.test.ts`, `tests/cli-heartbeat.test.ts`, `tests/cli-import-persona.test.ts`, `tests/cli-memory.test.ts`, `tests/cli-nightly.test.ts`, `tests/cli-notify.test.ts`, `tests/cli-persona.test.ts`, `tests/cli-run.test.ts`, `tests/cli-task.test.ts`, `tests/cli-tick.test.ts`. A single `bun run format` (or whatever the project uses) would clean all 12 in one shot. Cosmetic — tests still pass — but worth flushing before merge so the next person doesn't think it's intentional.

PR #54, opened 2026-05-02
chore: README rewrite + applyUpdate cleans .bak after successful rename

andrewagrahamhodges:
Three items from one round of feedback bundled into one PR — all in the cleanup-and-polish category.

## 1. README rewrite

- **Install section moved to the top.** Was buried below personas/architecture; now it's the first thing after the tagline so users don't have to hunt.
- **Quick-start path** — numbered 5-command sequence from \`curl … | sh\` to \"Telegram bot is replying.\"
- **Scrubbed stale commands** — all \`phantombot create-persona\` / \`phantombot import-persona\` references replaced with the consolidated \`phantombot persona\` form (those subcommands were removed in PR #51 but the README still had 3 mentions).
- **Visual hierarchy** — \`---\` between sections, command reference moved into tables instead of dense code blocks, less prose padding around the bullets.

## 2. install.sh \`mkdir -p\`

Already there at line 54 (\`mkdir -p \"\$INSTALL_DIR\"\`), before the writability check + before any write. With \`set -eu\` any failure aborts cleanly. So no code change to install.sh — the user's report of \"no mkdir\" was a false alarm. README's install section now explicitly mentions \"Creates \`~/.local/bin/\` if needed and installs phantombot at mode 0755\" so future readers can see it for themselves.

## 3. \`applyUpdate\` cleans \`.bak\` after successful rename

The user's actual report: \`phantombot.bak\` was sitting in \`~/.local/bin/\` next to the live \`phantombot\`, polluting tab-completion (\`phantombo<TAB>\` offered both). The .bak only ever needed to exist transactionally — between the \`copyFile\` (defensive backup) and the \`rename\` (atomic swap). Once the rename commits, the .bak is garbage.

\`src/lib/binaryUpdate.ts\` \`applyUpdate\` now \`rm\`s the .bak after a successful rename (best-effort — failure here just defers cleanup to the next update). \`ApplyResult\` no longer includes \`backupPath\` since the path doesn't exist post-success. \`src/cli/update.ts\` drops its \"previous binary saved to …\" line.

The PR #47 upfront-unlink (defense against stale foreign-owned .bak from initial \`sudo cp\` deploys) stays in place — covers the case where a previous version of the code DID leave a .bak behind, OR the swap crashed mid-flight.

**Existing .bak files on already-deployed boxes** (lena/kai/jake/robbie) will be cleaned automatically by the next \`phantombot update\` run via the same upfront-unlink path.

## Tests

- \`tests/lib-binaryUpdate.test.ts\` — happy-path test now asserts \".bak is gone\" post-success (was asserting on .bak contents). Stale-.bak regression test (PR #47) updated similarly. \`ApplyResult\` shape change reflected.
- \`tests/cli-update.test.ts\` — removes the \".bak contains OLD_BINARY\" assertion, replaces with \".bak doesn't exist\".

Suite: **455 pass / 1 skip / 0 fail**. Typecheck clean.

## Test plan

- [x] \`bun test\` — green
- [x] Local \`bun run build\` produces a binary
- [ ] **Self-test on merge**: workflow run produces \`v1.0.<this PR number>\` with both binaries
- [ ] **Real-world**: on an already-deployed box (e.g. lena), \`phantombot update --force --restart\` should leave NO \`.bak\` in \`~/.local/bin/\` afterward — and tab-completion for \`phantombot\` should show only the live binary, not the old \`phantombot.bak\`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,212 +1,217 @@
 # phantombot
 
-A personality-first chat agent. Phantombot wraps Claude Code, Inflection Pi, and Google Gemini CLI with persona, memory, scheduled tasks, and a Telegram bot — and otherwise stays out of their way. The harness runs its own tool loop. Phantombot does:
+A personality-first chat agent for Telegram. Wraps Claude Code, Inflection Pi, and Google Gemini CLI — and otherwise stays out of their way. The harness runs its own tool loop; phantombot does identity, memory, channel, scheduling, and self-update.
 
-1. **Identity** — load the agent's persona files (`BOOT.md` / `SOUL.md` / `IDENTITY.md`, plus `MEMORY.md`, `tools.md` / `AGENTS.md`) and inject them as the harness's system prompt.
-2. **Memory** — persist conversation turns to local SQLite, retrieve recent history per turn.
-3. **Channel** — Telegram bot via long-polling `getUpdates`.
-4. **Schedule** — `phantombot tick` fires user-defined tasks every minute (cron expressions).
-5. **Update self** — `phantombot update` fetches the latest GitHub Release, sha256-verifies, atomically swaps the running binary.
-6. **Fallback** — primary harness fails → next in chain.
+---
 
-## Commands
-
-```
-# First-time / config
-phantombot persona                           # TUI: create / import / restore / switch the active persona
-phantombot persona <name>                    # switch default persona to <name>
-phantombot persona --import <dir> [--as <n>] # non-interactive import (OpenClaw or phantombot-shaped)
-phantombot telegram                          # interactive TUI: configure the Telegram channel
-phantombot harness                           # interactive TUI: pick primary + fallback harnesses
-phantombot voice                             # interactive TUI: pick TTS/STT provider (ElevenLabs/OpenAI/Azure Edge)
-phantombot embedding                         # interactive TUI: configure Gemini embeddings (optional)
-
-# Day-to-day
-phantombot run                               # foreground long-running listener (Ctrl-C to stop)
-phantombot install                           # install systemd --user units (main + heartbeat + nightly + tick)
-phantombot uninstall                         # remove the systemd units
-phantombot update                            # fetch latest GitHub Release, verify, atomically swap the binary
-phantombot update --check                    # just print availability (exit 0 / 2)
-phantombot update --force --restart          # cron-friendly: no prompts, restart after install
-
-# Agent-facing tools (the harnessed agent calls these via Bash)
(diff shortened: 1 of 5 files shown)


andrewagrahamhodges:
Clean PR. All three changes are well-scoped and self-contained.

**applyUpdate .bak cleanup** — the right fix. Step 4 rm is best-effort, try/catch covers edge cases, ApplyResult dropping backupPath is correct API cleanup. The stale-foreign-owned-.bak test still passes because step 1 handles that case before step 2 ever creates a new one, so step 4 always sees a self-owned file. Good layering.

**README rewrite** — big structural improvement. Install at the top, quick-start path, tables over dense code blocks, stale create-persona/import-persona references all gone. The voice-replies section documents PR #53 behavior without leaking too much implementation detail.

**install.sh mkdir -p** — already there, no code change needed. README now documents it explicitly.

Suite: 455 pass / 0 fail. Mergeable.

andrewagrahamhodges (commented):
Clean, focused change. Three things I like:

1. **Step 4 cleanup is optional/safe.** `rm(backupPath, { force: true })` wrapped in try/catch — if it fails, the .bak just sits until the next update cleans it in step 1. No new failure mode.

2. **Tests are thorough.** Both unit and integration tests now assert .bak is gone, and the stale-readonly-backup edge case still passes.

3. **README rewrite is well-structured.** Install-first flow is much more accessible. Voice reply and scheduled-task polish earn their space.

One inline nit on the README — posting separately.

README.md
```diff
@@ -1,213 +1,218 @@
...
-Configuration lives in `~/.config/phantombot/config.toml` (TUIs write here) and `~/.local/share/phantombot/state.json` (default persona). Secrets in `~/.config/phantombot/.env` (phantombot-managed) and `~/.env` (user-managed; the agent writes here via `phantombot env set`).
+The script:
 
-## Why this exists
+- Detects host arch (`x86_64` / `aarch64`).
+- Fetches the latest GitHub Release tag.
+- Downloads the matching binary + `SHA256SUMS`, **verifies the checksum**, refuses on mismatch.
+- Creates `~/.local/bin/` if needed and installs `phantombot` at mode 0755.
+- Warns if `~/.local/bin` isn't on your `PATH`.
+- Launches `phantombot persona` so you can set up your first persona.
```
andrewagrahamhodges, line 21: 🟡 The old README said "unless stdin/stdout aren't a TTY" — the script auto-skips the TUI when headless. The new summary bullet drops that nuance. The `PHANTOMBOT_SKIP_TUI` env table three lines down covers the CI case, so no information is lost — just flagging in case the auto-detection was meant to be part of the bullet text.

PR #56, opened 2026-05-02
feat: slash commands + abort + grandchild kill + idle timeout + long-turn placeholder

andrewagrahamhodges:
## Summary

Fixes the cluster of agent-timeout failures we hit on kw-openclaw (Lena stuck on a `gemini usage` TCP wedge for 6+ minutes, Kai unable to be `/stop`'d while building a PR response) by reworking the harness execution path end-to-end.

This is two commits stacked: kai's `slash-commands-and-abort` foundation (slash dispatcher before the LLM, AbortController plumbing, per-chat queue), plus the four real gaps we identified in the design discussion.

## What's new

**1. Process-group spawn + SIGTERM→SIGKILL escalation** (`src/lib/processGroup.ts`)
- Each harness subprocess gets its own session/pgroup via Bun's `detached: true` (pid == pgid == sid)
- Kill signals route to the whole group via `process.kill(-pid, sig)` — grandchildren die with the parent
- 5s grace period between SIGTERM and SIGKILL for processes that ignore SIGTERM
- Fixes the `gemini usage` orphan: the wedged TCP-read subprocess now dies when its parent does

**2. Idle + hard timeouts** (`src/lib/harnessRunner.ts`, replaces wall-clock cap)
- `harness_idle_timeout_s` (default 120s) — kill if no chunk on stdout for this long, resets per chunk
- `harness_hard_timeout_s` (default 3600s) — absolute ceiling regardless of activity
- Productive 30-min builds with continuous tool events never trip idle; wedged subprocesses do
- Legacy `turn_timeout_s` aliases to hard for back-compat
- Coordinator factored out of all three harnesses so kill semantics stay aligned

**3. Long-turn placeholder with completion notification** (Option B from the design discussion)
- After 30s silent, post \"⏳ working… 30s elapsed\"
- Edit every 60s with elapsed + last progress note (\"6m 12s — currently: tool_execution_start: BashTool\")
- On completion: **delete placeholder, send result as new message** — Telegram pushes a notification on new messages but not on edits, so this delete-then-send is what makes your phone buzz when a 10-min Kai-builds-a-PR turn finishes
- Cleaned up on /stop and on errors too — no orphan ⏳ messages

**4. /status enrichment** (`src/channels/commands.ts`)
- Tracks last progress note in ActiveTurnHandle as chunks arrive
- /status appends \"running: <last tool>\" line — answers \"is it stuck or just busy?\"

**Plus kai's foundation** (commit f061cbf):
- `/stop`, `/reset`, `/status`, `/harness`, `/help` handled inline before the LLM in the polling loop
- Per-chat AbortController tracking; /stop aborts via controller.abort()
- Per-chat promise chain so /stop and /status work even when a turn is hung
- AbortSignal threaded through HarnessRequest → all three harnesses

## Test plan

- [x] Full suite: **509 pass, 1 skip, 0 fail** (up from 487 — 23 new tests)
- [x] Typecheck clean (`bun tsc --noEmit`)
- [x] Real-subprocess orphan-kill test: spawns `sh -c 'sleep 30 & echo \$!; wait'`, kills the parent, verifies the `sleep` grandchild is dead
- [x] SIGTERM→SIGKILL escalation test: spawns a Bun process with empty SIGTERM handler, verifies graceful → forced kill within the 5s window
- [x] Idle timer test: process emits chunk every 50ms, idle window 200ms — never killed (touch resets)
- [x] Idle fire test: process emits one line then sleeps; reader stops touching → idle fires with cause='idle'
- [x] Hard timer test: process keeps emitting (idle stays happy) but hits hard cap → killed with cause='timeout'
- [x] Placeholder lifecycle: fast turn = no placeholder; slow turn = posted + edited + deleted + fresh result; /stop = placeholder cleaned up, no duplicate result
- [x] /status running line: shows last progress note when present, omits when absent, truncates very long notes

## Deployment notes

After this merges and v1.0.55 publishes:
1. `phantombot update --force --restart` on each persona's host (lena, kai, jake on kw-openclaw; robbie on Hetzner)
2. Drop the legacy `turn_timeout_s = 600` from each `~/.config/phantombot/config.toml` so the new 120s idle / 3600s hard defaults take effect — or set them explicitly:
   \`\`\`toml
   harness_idle_timeout_s = 120
   harness_hard_timeout_s = 3600
   \`\`\`

## Bugs this fixes

- **Lena stuck on `gemini usage`** (kw-openclaw, 2026-05-02 21:00 UTC): hung TCP read in a grandchild; only died after orphan kill propagation
- **Kai unable to be /stop'd** building a PR response: blocked LLM meant /stop went into the same queue and waited 10min for the timeout
- **600s timeout firing on legitimate work**: replaced with idle timer that distinguishes \"working\" (constant tool events) from \"wedged\" (silent stdout)

### DIFF
--- a/src/channels/commands.ts
+++ b/src/channels/commands.ts
@@ -1,0 +1,245 @@
+/**
+ * Slash command dispatcher for chat channels.
+ *
+ * Sits BEFORE the LLM in the message loop. Catches in-band control commands
+ * (`/stop`, `/reset`, `/status`, `/harness`, `/help`) and handles them in the
+ * channel layer so they keep working even when the LLM is hung on a
+ * subprocess tool call — that was the failure mode that motivated this
+ * module: PhantomBot's old design routed every message through the harness,
+ * so a stuck `gemini usage` subprocess would block `/stop` along with
+ * everything else.
+ *
+ * The handler is intentionally pure-ish: it returns a result object and
+ * mutates only what was passed in (memory store, harness chain, the active
+ * turn's AbortController). The channel adapter is responsible for sending
+ * the reply text back to the user.
+ *
+ * Recognized vs unknown:
+ *   - `/stop`, `/reset`, `/status`, `/harness`, `/help` → handled here.
+ *   - Any other `/foo` → returned as null, channel falls through to runTurn
+ *     so the LLM can interpret it (some personas use `/remember`, etc.).
+ */
+
+import type { Harness } from "../harnesses/types.ts";
+import { log } from "../lib/logger.ts";
+import type { MemoryStore } from "../memory/store.ts";
+
+export interface ActiveTurnHandle {
+  controller: AbortController;
+  startTime: number;
+}
+
+export interface SlashCommandContext {
+  /** For logging / disambiguation only. */
+  chatId: number;
+  persona: string;
+  /** Conversation key, e.g. "telegram:42". Used by /reset. */
+  conversation: string;
+  /** Memory store for /reset's deleteConversation call. */
+  memory: MemoryStore;
+  /**
+   * The harness chain — mutable. /harness reorders this in place so the
+   * channel adapter (which holds the same array reference) sees the new
+   * primary on the next turn.
+   */
+  harnesses: Harness[];
+  /** Wall-clock when the channel server started, for /status uptime. */
+  startedAt: number;
+  /** Currently running turn for this chat, if any. /stop aborts it. */
+  activeTurn?: ActiveTurnHandle;
+}
+
+export interface SlashCommandResult {
+  /** Reply text to send back to the user. Always non-empty for handled commands. */
+  reply: string;
+}
+
+const HELP =
+  `available commands:\n` +
+  `/stop     — abort the current turn\n` +
+  `/reset    — clear this chat's history\n` +
+  `/status   — show harness, uptime, context usage\n` +
+  `/harness  — list or switch the active harness (e.g. /harness pi)\n` +
+  `/help     — this list`;
+
+/**
+ * Parse + dispatch a slash command.
+ *
+ * Returns null if `text` is not a slash command we own — caller falls
+ * through to the LLM for that message. Returns a SlashCommandResult when
+ * the command is handled (recognized or refused).
+ */
+export async function handleSlashCommand(
+  text: string,
+  ctx: SlashCommandContext,
+): Promise<SlashCommandResult | null> {
+  const trimmed = text.trim();
+  if (!trimmed.startsWith("/")) return null;
+
+  // Telegram convention in groups: `/cmd@BotName arg1 arg2`. Strip the
+  // @suffix so the command matches whether the bot was @-mentioned or not.
+  const parts = trimmed.split(/\s+/);
+  const head = parts[0]!;
+  const cmd = head.split("@")[0]!.toLowerCase();
+  const arg = parts.slice(1).join(" ").trim();
+
+  switch (cmd) {
+    case "/stop":
+      return handleStop(ctx);
+    case "/reset":
+      return handleReset(ctx);
+    case "/status":
+      return await handleStatus(ctx);
+    case "/harness":
+      return await handleHarness(arg, ctx);
+    case "/help":
+      return { reply: HELP };
+    default:
+      return null;
+  }
+}
+
+function handleStop(ctx: SlashCommandContext): SlashCommandResult {
+  if (!ctx.activeTurn) {
+    return { reply: "no active turn to stop" };
+  }
+  const elapsedS = ((Date.now() - ctx.activeTurn.startTime) / 1000).toFixed(1);
+  ctx.activeTurn.controller.abort();
+  log.info("commands: /stop fired", { chatId: ctx.chatId, elapsedS });
+  return { reply: `stopped (was running ${elapsedS}s)` };
+}
+
+async function handleReset(
+  ctx: SlashCommandContext,
+): Promise<SlashCommandResult> {
+  const removed = await ctx.memory.deleteConversation(
+    ctx.persona,
+    ctx.conversation,
+  );
+  log.info("commands: /reset", {
+    chatId: ctx.chatId,
+    persona: ctx.persona,
+    conversation: ctx.conversation,
+    deletedTurns: removed,
+  });
+  const noun = removed === 1 ? "turn" : "turns";
+  return { reply: `reset: cleared ${removed} ${noun} from this chat` };
+}
+
+async function handleStatus(
+  ctx: SlashCommandContext,
+): Promise<SlashCommandResult> {
+  const uptimeS = Math.floor((Date.now() - ctx.startedAt) / 1000);
+  const primary = ctx.harnesses[0]?.id ?? "(none)";
+  const chain = ctx.harnesses.map((h) => h.id).join(" → ") || "(none)";
+
+  // Rough context estimate: total chars across the last 20 turns, divided
+  // by 4 (the standard chars-per-token heuristic). Doesn't include the
+  // system prompt, which is ~stable across turns. Off by ~10-30% from a
+  // real tokenizer reading — fine for "is the context filling up" UX.
+  const recent = await ctx.memory.recentTurns(
+    ctx.persona,
+    ctx.conversation,
+    20,
+  );
+  const historyChars = recent.reduce((a, t) => a + t.text.length, 0);
+  const approxTokens = Math.round(historyChars / 4);
+  const windowTokens = nominalContextWindow(primary);
+  const pct = Math.min(
+    100,
+    Math.max(0, Math.round((approxTokens / windowTokens) * 100)),
+  );
+
+  const active = ctx.activeTurn
+    ? `yes (${((Date.now() - ctx.activeTurn.startTime) / 1000).toFixed(1)}s)`
+    : "no";
+
+  return {
+    reply:
+      `harness: ${primary}\n` +
+      `chain:   ${chain}\n` +
+      `uptime:  ${formatUptime(uptimeS)}\n` +
+      `context: ~${pct}% (≈${approxTokens.toLocaleString()} / ${windowTokens.toLocaleString()} tokens, last 20 turns)\n` +
+      `active:  ${active}`,
+  };
+}
+
+async function handleHarness(
+  arg: string,
+  ctx: SlashCommandContext,
+): Promise<SlashCommandResult> {
+  if (ctx.harnesses.length === 0) {
+    return { reply: "no harnesses configured" };
+  }
+
+  if (!arg) {
+    // No arg → list current chain with availability.
+    const lines: string[] = [];
+    for (let i = 0; i < ctx.harnesses.length; i++) {
+      const h = ctx.harnesses[i]!;
+      const ok = await h.available();
+      const marker = i === 0 ? "→" : " ";
+      const suffix = ok ? "" : " (unavailable)";
+      lines.push(`${marker} ${h.id}${suffix}`);
+    }
+    return {
+      reply:
+        `current chain (→ = primary):\n${lines.join("\n")}\n\n` +
+        `use /harness <id> to switch primary`,
+    };
+  }
+
+  const wanted = arg.toLowerCase();
+  const idx = ctx.harnesses.findIndex((h) => h.id === wanted);
+  if (idx < 0) {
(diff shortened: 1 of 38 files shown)


src/lib/processGroup.ts
```diff
@@ -0,0 +1,140 @@
...
+  opts: SpawnOptions.OptionsObject<Stdin, Stdout, Stderr>,
+): Subprocess<Stdin, Stdout, Stderr> {
+  if (cmd.length === 0) {
+    throw new Error("spawnInNewSession: cmd cannot be empty");
+  }
+  return Bun.spawn(cmd, {
+    ...opts,
+    // Undocumented but stable Bun option (maps to POSIX_SPAWN_SETSID).
+    // See module docstring for why this beats a `setsid` wrapper.
+    detached: true,
```
andrewagrahamhodges, line 64: 🟡 **Minor**: Using Bun's `detached: true` is the right call here. It ensures the pgroup is live before we even have the PID in the parent, closing the race condition that a `setsid` wrapper would have.

src/channels/telegram.ts
```diff
@@ -284,22 +341,71 @@ export interface RunTelegramServerInput {
    * stopped responding. Default 4000 ms. Tests use a smaller value.
```
andrewagrahamhodges, line 342: 🟡 **Minor**: The new concurrency model using `chatChains` and the `inFlight` set for draining is robust. It correctly separates channel-level command handling from serialized LLM turns.

andrewagrahamhodges (commented):
## Code Review: phantombot timeout/kill refactor

**Author:** @andrewagrahamhodges | **Files:** 37 | **Verdict:** Approve

### Findings

| 1 | 🟡 Minor | src/lib/processGroup.ts | 1 | The use of `detached: true` is a clever solution to the orphan grandchild problem. Worth noting that it makes the child process group leader, which is exactly why `process.kill(-pid)` works. |
| 2 | 🟡 Minor | src/channels/telegram.ts | 310 | The move to a per-chat promise chain (`chatChains`) is a major architectural improvement. It solves the issue where slash commands like `/stop` were blocked by an in-flight LLM turn. |

### What's Good
- **Consolidated Kill Logic:** Moving the complex SIGTERM → SIGKILL logic into `harnessRunner.ts` eliminates the drift where different harnesses had different (or missing) timeout behaviors.
- **Orphan Grandchild Fix:** The new `processGroup.ts` helper ensures that tool calls (like `gemini usage`) don't survive as zombies when their parent harness is killed.
- **Async Channel Layer:** Telegram server now handles slash commands immediately while serializing LLM turns per chat. This dramatically improves responsiveness under load.
- **Extensive Test Coverage:** The addition of `lib-processGroup.test.ts` and `lib-harnessRunner.test.ts` using real subprocesses (shells/Bun) provides confidence that the kernel-level signalling works as expected.

<details>

```
Review the process management refactor in PR #56. Ensure that any future harness additions use createKillCoordinator from src/lib/harnessRunner.ts rather than implementing their own spawn/kill logic. This prevents regressions of the 'orphan grandchild' bug fixed here.

Repository: andrewagrahamhodges/phantombot
Branch: fix/timeout-kill-refactor
Base: main
```
</details>

src/channels/telegram.ts
```diff
@@ -308,185 +414,352 @@ export async function runTelegramServer(
...
       try {
-        if (willReplyWithVoice && !errored && reply.length > 0) {
-          const r = await synthesize(input.config, reply);
-          if (r.ok) {
-            await input.transport.sendVoice(
+        const id = await input.transport.sendMessage(
+          msg.chatId,
+          renderPlaceholder(),
+        );
+        if (id > 0) {
```
andrewagrahamhodges: **Race: placeholder can become an orphan + leak a setInterval if it posts after the turn finishes.**

If `placeholderPostTimer` fires near the end of a turn, the deferred `sendMessage` may resolve *after* the for-await loop ends and the `finally` clause runs. Sequence:

1. Turn duration ≈ `placeholderThresholdMs + sendMessage_latency`.
2. `setTimeout` fires → async IIFE starts → awaits `sendMessage` (50–200 ms over the wire).
3. While that's in flight, the for-await ends → `finally` clears `placeholderEditTimer` (still `undefined`, no-op) and `placeholderPostTimer` (already fired, no-op).
4. `cleanupPlaceholder()` at line 727 runs; `placeholderMsgId` is still `undefined`, so the delete is skipped.
5. Final reply is sent.
6. Now the deferred `sendMessage` resolves → `placeholderMsgId = id` → `setInterval` is armed.
7. Orphan ⏳ message remains in the chat; the interval ticks every `placeholderEditMs` forever and calls `editMessage` against a defunct turn.

Ways to plug it (any one works):
- `let disposed = false;` in the closure scope; set in `finally`; the IIFE bails before storing `placeholderMsgId` if disposed.
- Have `cleanupPlaceholder` `await` the in-flight post promise (track it as a local `placeholderPostPromise`) before checking `placeholderMsgId`.
- Make the IIFE check `controller.signal.aborted || activeTurns.get(msg.chatId) !== turnHandle` before arming the interval.

Production impact: low frequency, but each occurrence leaks a `setInterval` in the long-running `phantombot.service`. The test at `tests/channels-telegram.test.ts:1066` doesn't catch it because `FakeTransport.sendMessage` resolves synchronously.

src/channels/telegram.ts
```diff
@@ -507,6 +780,26 @@ export async function runTelegramServer(
  * triggered ONLY when the input arrived as voice AND the reply will be
  * synthesized — text replies stay as detailed as the persona wants.
  */
+/**
+ * Render an elapsed-ms duration as a short human string for the
+ * placeholder line. Matches the /status format intentionally.
+ *
+ * Exported for testing.
+ */
+export function formatElapsed(ms: number): string {
```
andrewagrahamhodges: **Orphan JSDoc.** The 18-line block at lines 765–782 documents `VOICE_REPLY_INSTRUCTION`, but the very next declaration is `formatElapsed` (this line), not the constant. The actual `VOICE_REPLY_INSTRUCTION` declaration at line 803 has no JSDoc.

Looks like a clean-merge casualty from PR #53 colliding with the placeholder code added here. Fix is to move the orphaned block down to immediately precede the `export const VOICE_REPLY_INSTRUCTION =` line.

src/config.ts
```diff
@@ -100,12 +110,28 @@ export async function loadConfig(): Promise<Config> {
...
+      // files keep working. New configs should use the *_idle / *_hard
+      // pair above.
+      (asInt(process.env.PHANTOMBOT_TURN_TIMEOUT_MS) !== undefined
+        ? asInt(process.env.PHANTOMBOT_TURN_TIMEOUT_MS)!
+        : undefined) ??
       (asInt(toml.turn_timeout_s) !== undefined
         ? asInt(toml.turn_timeout_s)! * 1000
         : undefined) ??
-      600_000,
+      3_600_000,
```
andrewagrahamhodges, line 172: **Legacy `turn_timeout_s` is a behaviour change, not a pure alias — worth a louder migration signal.**

- Pre-PR: `turn_timeout_s = 600` meant "kill at 600s wall-clock, full stop."
- Post-PR: same value maps to `harnessHardTimeoutMs = 600s`, but `harnessIdleTimeoutMs` quietly defaults to **120s** — *stricter* than what the user wrote. A persona doing legitimately long tool calls (slow `git push`, gog API round-trips, `apt upgrade`) silently starts hitting 120s idle kills it never used to.

The PR body's deployment note covers this for lena/kai/jake/robbie, but anyone external who upgrades and skips the changelog regresses. Two cheap mitigations, in order of cleanness:

1. Alias `turn_timeout_s` to *both* `idleTimeoutMs` and `hardTimeoutMs` (preserves "no idle kill" semantics for legacy configs). New configs opt into the stricter idle by setting `harness_idle_timeout_s` explicitly.
2. Or just `log.warn` once at startup when `turn_timeout_s` is the active source, naming the new pair so users know to migrate.

(1) is the "no surprise" path; (2) is the "noisy-but-zero-LoC" path.

src/channels/commands.ts
```diff
@@ -0,0 +1,267 @@
...
+  if (!ctx.activeTurn) {
+    return { reply: "no active turn to stop" };
+  }
+  const elapsedS = ((Date.now() - ctx.activeTurn.startTime) / 1000).toFixed(1);
+  ctx.activeTurn.controller.abort();
+  log.info("commands: /stop fired", { chatId: ctx.chatId, elapsedS });
+  return { reply: `stopped (was running ${elapsedS}s)` };
+}
+
+async function handleReset(
```
andrewagrahamhodges, line 120: **`/reset` doesn't interact with an in-flight turn — minor UX wart.**

If a turn is running when `/reset` arrives, this clears history but the turn keeps going. When it finishes, `runTurn` happily appends the new user/assistant pair to the (now-empty) conversation. The user typed `/reset` expecting a clean slate, then sees their pre-`/reset` reply land and get persisted.

Suggest pairing with `/stop` semantics: if `ctx.activeTurn` is set, `controller.abort()` first so the in-flight turn yields the stopped error, doesn't persist, and the reset is genuinely clean. The reply text could become `reset: cleared N turns (and stopped an in-flight turn that was Xs in)`.

src/channels/commands.ts
```diff
@@ -0,0 +1,267 @@
...
+      `harness: ${primary}\n` +
+      `chain:   ${chain}\n` +
+      `uptime:  ${formatUptime(uptimeS)}\n` +
+      `context: ~${pct}% (≈${approxTokens.toLocaleString()} / ${windowTokens.toLocaleString()} tokens, last 20 turns)\n` +
+      `active:  ${active}` +
+      runningLine,
+  };
+}
+
+function truncateLine(s: string, max: number): string {
```
andrewagrahamhodges: nit: `truncateLine` here and `truncate` in `src/channels/telegram.ts:798` are byte-identical; `formatUptime` (line 259 below) and `formatElapsed` in telegram.ts only differ in their day handling.

Worth extracting both to `src/lib/format.ts` so the next time you tweak truncation or the elapsed-time format you don't have to remember to touch two places. Not a blocker.

src/harnesses/claude.ts
```diff
@@ -99,20 +104,18 @@ export class ClaudeHarness implements Harness {
     proc.stdin.write(renderStdinPayload(req));
     proc.stdin.end();
```
andrewagrahamhodges: nit: gemini.ts wraps `proc.stdin.write` + `await proc.stdin.end()` in try/catch and warns on failure (`gemini.ts:144–151`). Here the equivalent `proc.stdin.write(...); proc.stdin.end();` isn't awaited and isn't guarded. EPIPE (kernel killed the proc between spawn and write — happens when `req.signal` was already aborted) propagates out of the generator before the for-await loop can yield the `cause = "aborted"` error chunk.

Practical impact is small — `runWithFallback`'s pre-loop `if (req.signal?.aborted)` short-circuits before invoke runs, so this only fires on a tight race. But for consistency with gemini.ts (and to match the docstring claim that "kill semantics stay aligned" across the three harnesses), matching the wrap-and-warn pattern would be cleaner.

andrewagrahamhodges (commented):
## Review notes

Read the whole diff (1.2k lines) end-to-end against `main`, walked the new modules, then traced the placeholder lifecycle and `/stop` paths line-by-line. Big picture: **this is a strong PR.** The architecture is clean — process-group spawn + kill coordinator is exactly the right shape, and factoring it out of all three harnesses kills the drift that bit Lena yesterday. The `/stop` flow is genuinely well thought out (per-chat AbortController, polling-loop bypass, `wasStopped` suppressing the duplicate "(error: stopped)" reply). 23 new tests, all the scary stuff (orphan-grandchild reap, SIGTERM→SIGKILL escalation, idle-vs-hard, abort propagation) covered with real subprocesses rather than mocks.

Six findings below: **one real bug** (placeholder race that can leak a `setInterval` per occurrence in the long-running `phantombot.service`), **one cleanup nit from the PR #53 merge** (orphan JSDoc), **one behavior-change call-out** (legacy `turn_timeout_s` is no longer a pure alias), and three smaller nits. Nothing here is a blocker — the bug only fires on a tight timing window and the impact is bounded — but worth fixing before this lands as v1.0.55 and ships to four hosts.

I couldn't run `bun test` locally (no bun on the Hetzner VPS), so trusting the PR's "509 pass / 1 skip / 0 fail" claim. If you want me to spin up a bun install on the box for review purposes, say the word.

PR #57, opened 2026-05-03
feat: Telegram chat-style prompt + 8s typing pulse + drop placeholder

andrewagrahamhodges:
## Summary

Three streamlining changes to the Telegram experience.

## 1. Channel-layer prompt suffix for Telegram turns

New `TELEGRAM_REPLY_INSTRUCTION` applied to every Telegram turn:

- **Default to short, conversational replies** (1-4 sentences). The user is on a phone with a narrow column. Longer replies are explicitly allowed when the user asks for a detailed report or analysis.
- **Plan-then-confirm before long jobs**: any git/build/deploy work, or anything that would spawn more than one tool call. The agent outlines its plan in 2-3 sentences and asks for confirmation. This avoids burning 10 minutes producing the wrong thing — the Telegram round-trip is fast enough to align up front.

`VOICE_REPLY_INSTRUCTION` still applies (stacked on top of the chat-style instruction) for voice-out turns, where the stricter 1-3 sentence + no-markdown rules matter for TTS playback.

## 2. Typing indicator: 4s → 8s refresh

Telegram's chat-action expires after ~5s. Refreshing every 8s creates a deliberate pulse: indicator visible for ~5s, gone for ~3s, back again. The user sees activity without staring at a frozen \"is typing…\" header.

## 3. Drop the long-turn placeholder entirely

The \"⏳ working… 30s elapsed\" message from PR #56 was:
- **English-only noise** that would alienate non-English-speaking users
- Almost never carrying the per-tool detail it was designed to show — the harnesses don't reliably emit progress chunks (only Pi's tool_execution_start would have, and it's not yet observed in the wild for plain conversational turns)

The 8s typing pulse is now the sole \"still alive\" signal during long silences.

Removed code:
- `placeholderThresholdMs` / `placeholderEditMs` config knobs
- All placeholder state + timers + IIFE + the race fix from the PR #56 review
- `cleanupPlaceholder` helper
- `TelegramTransport.editMessage` + `deleteMessage` (unused without the placeholder pipeline)
- `sendMessage`'s `Promise<number>` return type (was only used to capture the placeholder messageId)
- `formatElapsedMs` from `src/lib/format.ts` (orphan; `formatElapsedSeconds` stays for `/status`)
- All placeholder lifecycle / race / `formatElapsedMs` tests (~290 lines)

## Test plan

- [x] **503 pass, 1 skip, 0 fail** (down from 510 by removing placeholder tests; up by one new chat-style suffix assertion)
- [x] Typecheck clean
- [x] Updated suffix-injection tests verify both stacking paths:
  - voice-in + voice-out → harness sees both `Reply style (Telegram chat)` AND `Reply length (this turn only)`
  - text-in + text-out → harness sees `Reply style (Telegram chat)` but NOT the voice overlay

## Diff stats

`4 files changed, 110 insertions(+), 579 deletions(-)` — net code reduction.

## Deployment notes

After merge + v1.0.57 publishes:
\`\`\`
phantombot update --force --restart
\`\`\`
on each persona host. The chat behavior change takes effect on the next turn — agents will start outlining plans before long jobs and will be more conversational by default.

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -24,7 +24,6 @@ import {
   transcribe,
   ttsSupported,
 } from "../lib/audio.ts";
-import { formatElapsedMs, truncateLine } from "../lib/format.ts";
 import type { WriteSink } from "../lib/io.ts";
 import { log } from "../lib/logger.ts";
 import type { MemoryStore } from "../memory/store.ts";
@@ -61,8 +60,7 @@ export interface TelegramTransport {
     timeoutS: number,
     signal?: AbortSignal,
   ): Promise<{ updates: TelegramMessage[]; nextOffset: number }>;
-  /** Send a text message; returns the new message_id (used for later edit/delete). */
-  sendMessage(chatId: number, text: string): Promise<number>;
+  sendMessage(chatId: number, text: string): Promise<void>;
   sendTyping(chatId: number): Promise<void>;
   /** Send an OGG-Opus voice note. */
   sendVoice(chatId: number, audio: Buffer, mime: string): Promise<void>;
@@ -70,10 +68,6 @@ export interface TelegramTransport {
   sendRecording(chatId: number): Promise<void>;
   /** Download a file by Telegram file_id; returns audio bytes + content-type. */
   downloadFile(fileId: string): Promise<{ data: Buffer; mime: string }>;
-  /** Edit an existing text message in place. Best-effort; failures are swallowed. */
-  editMessage(chatId: number, messageId: number, text: string): Promise<void>;
-  /** Delete a message by id. Best-effort; failures are swallowed. */
-  deleteMessage(chatId: number, messageId: number): Promise<void>;
 }
 
 /**
@@ -118,7 +112,7 @@ export class HttpTelegramTransport implements TelegramTransport {
     return parseGetUpdatesResult(body.result ?? [], offset);
   }
 
-  async sendMessage(chatId: number, text: string): Promise<number> {
+  async sendMessage(chatId: number, text: string): Promise<void> {
     const url = `https://api.telegram.org/bot${this.token}/sendMessage`;
     // Telegram caps message length at 4096 chars. Truncate gracefully.
     const safe = text.length > 4000 ? text.slice(0, 4000) + "\n…[truncated]" : text;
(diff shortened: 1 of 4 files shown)


andrewagrahamhodges (commented):
## Code Review: Telegram Streamlining

**Author:** @andrewagrahamhodges | **Files:** 4 | **Verdict:** Approve

This PR significantly cleans up the Telegram transport by removing the noisy placeholder message logic and replacing it with a more subtle typing "pulse" and channel-specific prompt instructions. The net reduction of ~470 lines is excellent.

### Findings

| 1 | 🟡 Minor | src/channels/telegram.ts | 509 | `typingRefreshMs` default check |
| 2 | 🔵 Trivial | src/channels/telegram.ts | 115 | `sendMessage` return value removal |

### What's Good
- **Prompt Stacking:** The implementation of stacking `TELEGRAM_REPLY_INSTRUCTION` and `VOICE_REPLY_INSTRUCTION` is clean and ensures consistency across turn types.
- **Pulse Strategy:** Moving to an 8s refresh for the typing indicator is a clever way to provide a "still alive" signal without the UI overhead (and English-only text) of a placeholder message.
- **Test Coverage:** Great job updating the tests to verify the prompt suffix logic while purging the obsolete placeholder tests.

<details>

```
Review the changes in PR #57 (feat: Telegram chat-style prompt + 8s typing pulse + drop placeholder).

1. [Minor] src/channels/telegram.ts:509 — The expression `input.typingRefreshMs ?? 8000` is safe, but ensure that any external config providing `0` doesn't cause unexpected behavior in setInterval (though 8000 is a sensible fallback).
2. [Trivial] src/channels/telegram.ts:115 — Confirm that the removal of the `message_id` return from `sendMessage` truly has no downstream consumers in other plugins (unlikely as it was a transport-specific internal).

Repository: andrewagrahamhodges/phantombot
Branch: feat/typing-pulse-drop-placeholder
```
</details>

src/channels/telegram.ts
```diff
@@ -569,7 +506,9 @@ async function processChatMessage(
       ? input.transport.sendRecording(msg.chatId)
       : input.transport.sendTyping(msg.chatId);
   void sendStatus();
-  const refreshMs = input.typingRefreshMs ?? 4000;
+  // Refresh interval defaults to 8s (5s indicator-on / 3s gap pulse) —
```
andrewagrahamhodges, line 509: 🟡 **Minor**: The default `8000` is correct for the desired "pulse" effect. Note that if `typingRefreshMs` is configured as `0` elsewhere, this will still default to `8000` due to nullish coalescing. This is safe for a timer.

src/channels/telegram.ts
```diff
@@ -118,7 +112,7 @@ export class HttpTelegramTransport implements TelegramTransport {
     return parseGetUpdatesResult(body.result ?? [], offset);
   }
 
-  async sendMessage(chatId: number, text: string): Promise<number> {
+  async sendMessage(chatId: number, text: string): Promise<void> {
```
andrewagrahamhodges, line 115: 🔵 **Trivial**: `sendMessage` now swallows failures (logs but doesn't throw or return status). This matches the "best-effort" philosophy of the transport layer, but ensure no callers expected to retry on failure.

PR #61, opened 2026-05-03
ci: add typecheck + test on pull requests

andrewagrahamhodges:
## Summary
- Adds `.github/workflows/ci.yml` that runs **`typecheck`** and **`test`** as two parallel jobs on every PR targeting `main`.
- **Trigger:** `pull_request` to `main` only. No `push:` block — `pull_request` already restricts runs to branches with an open PR (the GitHub equivalent of GitLab's "MR branch" gating).
- **Bun:** `oven-sh/setup-bun@v2`, pinned to `1.x` to match `release.yml` so CI gates against the same Bun the release job will actually build with.
- **Concurrency:** per-job groups (`ci-typecheck-\${{ github.head_ref }}`, `ci-test-\${{ github.head_ref }}`) with `cancel-in-progress: true`, so a new push to a PR cancels the in-flight run of each job independently.
- **Does not touch `release.yml`** — that workflow already runs on PR-merged-to-main and stays as the release job.

## Follow-up (out of scope for this PR)
Once this PR shows a green run, wire `typecheck` and `test` into the branch-protection rule on `main` as required status checks. (Branch protection lets you select check names once they've been observed at least once.)

## Test plan
- [ ] This PR's own `typecheck` and `test` jobs go green
- [ ] Push a no-op commit to confirm the in-flight runs get cancelled (concurrency works)
- [ ] After merge: open a follow-up PR and confirm both checks fire on it too

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,12 @@ on:
   pull_request:
     branches: [main]
 
+# Explicit least-privilege scope. `pull_request` defaults to read-only
+# already, but stating it makes the intended attack surface auditable
+# and survives any future change in org/repo defaults.
+permissions:
+  contents: read
+
 jobs:
   typecheck:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,43 @@
+name: ci
+
+# Runs on every PR targeting main. No `push:` block on purpose — the
+# `pull_request` event already restricts this to branches with an open PR
+# (equivalent to GitLab's "MR branch" behaviour). Branch protection on main
+# wires `typecheck` and `test` as required status checks; release.yml then
+# fires on PR-merged-to-main and cuts the actual binary release.
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  typecheck:
+    runs-on: ubuntu-latest
+    # Per-job concurrency so a new push to the PR cancels the in-flight run
+    # of *this* job only — the sibling `test` job has its own group and
+    # cancels independently.
+    concurrency:
+      group: ci-typecheck-${{ github.head_ref }}
+      cancel-in-progress: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: oven-sh/setup-bun@v2
+        with:
+          # Pinned to match release.yml so CI gates against the same Bun
+          # the release job will actually build with.
+          bun-version: 1.x
+      - run: bun install --frozen-lockfile
+      - run: bun tsc --noEmit
+
+  test:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ci-test-${{ github.head_ref }}
+      cancel-in-progress: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.x
+      - run: bun install --frozen-lockfile
+      - run: bun test


kaieriksen (approved):
Clean, well-reasoned PR. 43 lines, one file, clear intent.

**What I like:**
- Per-job concurrency (`ci-typecheck-` / `ci-test-` groups) — this is the right pattern. Pushing a type-check fix doesn't kill in-flight tests.
- `cancel-in-progress: true` — saves CI minutes and gets faster feedback.
- `--frozen-lockfile` — prevents silent lockfile drift.
- Bun `1.x` pinned to match `release.yml` — the comment explaining why is actually useful.
- The YAML comments throughout are substantive, not filler.

**One observation (not a blocker):**
The `install` step runs in both jobs independently. With Bun this is fast enough that job isolation wins over caching complexity, but if runtime creeps up later, `actions/cache` on Bun's global cache is an easy win.

**One optional nit:**
Consider adding an explicit `permissions: {}` or `permissions: { contents: read }` block. The defaults are correct today, but being explicit future-proofs against org-level default changes and makes the security scope visible at a glance.

Approve — ship it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
+name: ci
```
robertclawson, line 1: Nit (defence-in-depth): no explicit `permissions:` block. Default for `pull_request` is read-only so this is fine, but adding `permissions:
  contents: read` at the top documents intent and protects against future GitHub default changes. Skip if you'd rather keep it minimal.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+    branches: [main]
+
+jobs:
+  typecheck:
+    runs-on: ubuntu-latest
+    # Per-job concurrency so a new push to the PR cancels the in-flight run
+    # of *this* job only — the sibling `test` job has its own group and
+    # cancels independently.
+    concurrency:
+      group: ci-typecheck-${{ github.head_ref }}
```
robertclawson, line 26: Nit: `github.head_ref` is `pull_request`-only. Fine right now since that's the sole trigger — but if a second trigger ever gets added, the group key collapses to an empty string and concurrency breaks silently. Defensive form:
```yaml
group: ci-typecheck-${{ github.event.pull_request.number || github.ref }}
```
Not worth changing today; flag if/when the trigger list grows.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+      cancel-in-progress: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: oven-sh/setup-bun@v2
```
robertclawson, line 30: Nit: no bun dep cache, so each job reinstalls from scratch. With 4 deps it's ~5–10s per job, not worth optimising yet. If install time grows, `actions/cache` keyed on `bun.lockb` (or `setup-bun`'s built-in caching once stable) would shave it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          # Pinned to match release.yml so CI gates against the same Bun
+          # the release job will actually build with.
+          bun-version: 1.x
```
robertclawson, line 34: Nit: `1.x` floats — a Bun minor regression would break CI and `release.yml` simultaneously (since they both pin `1.x`). Pinning to `1.2` (or whatever current minor is) would be tighter at the cost of losing automatic fix uptake. Current choice is internally consistent with the release job, so no change needed — just flagging for awareness.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+      - run: bun install --frozen-lockfile
+      - run: bun tsc --noEmit
```
robertclawson, line 36: Nit: invokes `bun tsc --noEmit` and `bun test` directly rather than the `package.json` scripts (`bun run typecheck` / `bun run test`). Matches `release.yml`'s style, so it's consistent. Mild argument for using the script names so future tweaks (extra flags, pre-steps) only need editing in one place.

robertclawson (dismissed):
## Approve ✅

Faithful to the spec, both checks already ran green on the head commit, additive only — no risk to production code or the release pipeline.

**What's good**
- Trigger is exactly right: `pull_request` to `main`, no `push:` block — only runs on branches with an open PR (the GitLab-MR-branch behaviour).
- Bun pinned to `1.x` matches `release.yml`, so we're gating against the same runtime the release will build with.
- Parallel jobs, per-job concurrency with `cancel-in-progress` — new pushes cancel in-flight runs cleanly.
- Doesn't touch `release.yml`. Clean separation: `ci.yml` gates merging, `release.yml` cuts the release on merge.
- Comments explain *why*, not *what*. 👍

Nitpicks left inline — none blocking.

**After merge:** I'll wire `typecheck` and `test` into `main`'s branch protection as required status checks (they'll appear in the dropdown now they've run once).

andrewagrahamhodges:
Thanks both — addressed the one nit raised by both reviewers (explicit `permissions: contents: read` at the workflow level, with a comment explaining why) in 8c42677.

Skipping the other inline nits per the reviewers' own guidance:

- **`github.head_ref` collapsing if a second trigger is added** — flagged "Not worth changing today; flag if/when the trigger list grows." Agreed; will revisit if/when we add a second trigger.
- **No bun dep cache** — "not worth optimising yet" with 4 deps. Agreed.
- **`1.x` floats** — explicitly "no change needed — just flagging for awareness." The internal consistency with `release.yml` is the point: if a Bun minor regression breaks one, we want it to break both visibly rather than masking the divergence.
- **`bun tsc` vs `bun run typecheck`** — "Mild argument" only; matches `release.yml` style. Keeping consistent with the release job is its own value, and changing only this side would create a subtle two-place divergence for the same logical command.

Will merge once green.

robertclawson (approved):
## Re-approve ✅

Checked the new commit (`8c42677`) and the response to the earlier nits.

**What changed**
- Added explicit `permissions: contents: read` at the workflow level with a comment explaining the intent. Addresses the L1 nit cleanly.

**The other 4 nits**
All explicitly declined with sound reasoning, and all four were marked non-blocking in my original review:
- `github.head_ref` fragility — fine to defer until a second trigger is added
- No bun dep cache — not worth optimising at 4 deps
- `1.x` floats — keeping it consistent with `release.yml` is the *point*; divergence would mask regressions
- `bun tsc` direct vs `bun run typecheck` — matches `release.yml` style, consistent is a feature

No argument from me on any of those calls.

**Status**
- `typecheck` ✅ success on `8c42677`
- `test` ✅ success on `8c42677`
- All 5 review threads resolved
- Workflow is faithful to the spec, additive only, zero risk to production

Ship it.

PR #62, opened 2026-05-03
fix(persona): adopt new persona as default on every create/import/restore path

andrewagrahamhodges:
## Summary

`phantombot run` was failing with `persona 'phantom' not found at .../personas/phantom` for users who created or imported a persona through any path other than the non-TUI `phantombot persona --import <dir>` form.

Root cause: `config.defaultPersona` falls back to the literal `"phantom"` when `state.json`/`config.toml` don't set it (`src/config.ts:148`). A `maybeAdoptAsDefault` helper existed to flip the default to the new persona when the previous default's dir was missing — but it was wired into **only one** of four persona-producing code paths.

Three paths were missing the call:
- `runImportFromPath` — TUI "Import from a directory"
- `runRestoreArchive` — TUI "Restore an archived persona"
- `applyPersona` — "Create a new persona" with `setDefault=false`

This PR:
- Extracts the helper to `src/lib/personaDefault.ts` as `adoptAsDefaultIfMissing(config, name, out?)` and calls it from all four sites.
- `applyPersona`'s result now carries `adoptedAsDefault: boolean` so the create-persona TUI can show a `Default` note when the user said no to the prompt but adoption fired anyway.
- Updates `AGENTS.md` pitfall #7 to point at the shared helper and note that any new persona-producing path must call it (and isolate `state.json` in tests).

## Test plan

- [x] `bun run typecheck` clean
- [x] `bun test` — 516 pass / 0 fail (was 516)
- [x] New test: `applyPersona` with `setDefault=false` adopts when current default has no dir on disk
- [x] Existing test refactored: `applyPersona` with `setDefault=false` leaves `state.json` untouched **only** when the current default exists on disk
- [ ] Manual: fresh box, `phantombot persona` → "Import from a directory" → confirm `state.json` `default_persona` is set to the imported name and `phantombot run` finds the persona dir
- [ ] Manual: same flow via "Restore an archived persona"
- [ ] Manual: `phantombot persona` → "Create a new persona" → answer **no** to "Set as default?" on a fresh box → confirm Default note appears and `phantombot run` works


kaieriksen (approved):
Reviewed and approved ✅

**PR #62** (`fix/adopt-default-persona-on-create-import-restore`): Extracts the auto-adopt helper to a shared module and wires it into the three persona-producing paths that were missing it. Clean extraction — old dead code removed, new tests cover both branches (existing default present vs fresh box), and the TUI "Default" note on adoption despite `setDefault=false` is thoughtful UX.

One inline nit on the test file — optional. Otherwise good to merge.

kaieriksen:
**📝 tests/cli-create-persona.test.ts:114** — Nit, optional:

`join(config.personasDir, "phantom", "BOOT.md")` on line 9 imports from `node:path`, which is correct — just noting that if someone later refactors imports, make sure the `join` used here stays as `path.join` and doesn't accidentally become a vanilla string concatenation. Non-blocking.

PR #64, opened 2026-05-03
fix(telegram+harness): no-reply, interrupt UX, interrupt context, gemini sandbox

andrewagrahamhodges:
## Summary

Four related fixes for Lena's slow / empty-reply behaviour, all motivated by what showed up in her journal once we bumped gemini stderr to `info`. Validated live on Lena's box across a real-work session.

### 1. `18a2a9d` — empty-done falls through to next harness
gemini exits 0 with `finalText=""` (SIGTERMed mid-stream by an updater restart, or a tool-only run that never produces an assistant message). The orchestrator was treating this as success and the channel sent `"(no reply)"` — even though pi was sitting right there in the chain. Now: empty `done` from a non-last harness is recoverable; chain falls through. Last-harness empty-done still surfaces `"(no reply)"` so the user knows something happened.

### 2. `92f41e3` — new message interrupts in-flight turn
Before: a second non-slash message queued behind the first on the per-chat serial chain, so users had to wait for a 1–3 min gemini turn (or remember `/stop`) before getting a reply to a new instruction. After: when a non-slash message arrives for a chat with an active turn, the polling loop calls `controller.abort("interrupt")` on the in-flight turn before enqueueing the new one. Same UX as Claude Code's "type to interrupt." `/stop` and `/reset` now also pass string reasons so the new `telegram: turn aborted` log line distinguishes the three causes for journalctl post-mortems.

### 3. `394addd` — aborted turns persist `[interrupted before reply]`
Without this the user's interrupted message vanished from history, and follow-ups like *"actually use blue instead"* landed with no referent — the model had no idea what was being re-described. The aborted turn now writes a synthetic user + `"[interrupted before reply]"` assistant pair to memory before returning. Skipped on `reason === "reset"` (history was just wiped) and on empty `msg.text` (voice STT aborted before transcript). Pattern matches phantomops's `__interrupted__` marker — validated as the right shape for the surprise problem without porting their full `--session` JSONL hydration model.

### 4. `af596da` — default harness cwd to user's home, not persona dir
Lena's journal showed every gemini turn failing tool calls with `Path not in workspace` — gemini-cli sandboxes itself to cwd, and phantombot was launching it in the persona dir (`~/.local/share/phantombot/personas/<name>/`). She couldn't access `~/Projects/phantombot/www` where her actual work lives, so every `list_directory` / `write_file` got rejected and the model produced vague empty replies. Pi has no sandbox and Claude (in `--permission-mode bypassPermissions`) accepts absolute paths, so this only mattered for gemini in practice — but the fix is harness-agnostic: change the default cwd to `os.homedir()`. `TurnInput` grows an optional `workingDir` field for callers that want to scope down.

### Side benefit (in commit 1)
Bumped gemini stderr from `debug` to `info` (per-invocation banners filtered out). This is what surfaced both root causes above (workspace sandbox + Google `MODEL_CAPACITY_EXHAUSTED` 429s) — they were invisible at the default log level before.

## Test plan

- [x] `bun run typecheck` clean
- [x] `bun test` — 521 pass / 0 fail (was 516)
- [x] New tests cover: empty-done fallthrough, last-harness empty-done preservation, interrupt-on-new-message, persisted interrupted-pair, `workingDir` default + override
- [x] Live test on Lena's box: interrupt fired correctly (`telegram: new message — interrupting active turn elapsedS=5.1` → `gemini.invoke killed: aborted` → `telegram: turn aborted reason=interrupt` → fresh turn started)
- [x] Live test: workspace fix confirmed — first post-deploy turn completed in 18s with replyChars=284 and zero `Path not in workspace` errors
- [ ] Open: still observing how the empty-done fallthrough behaves under live Google rate-limit conditions

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -204,6 +204,18 @@ function guessMimeFromPath(path: string): string {
   return "audio/ogg";
 }
 
+/**
+ * Render an AbortSignal.reason as a short string for logging.
+ * Callers pass plain strings ("stop", "reset", "interrupt"); the DOM
+ * default for a parameterless abort() is a DOMException — fold it down
+ * to its message so journalctl stays readable.
+ */
+function abortReasonString(reason: unknown): string {
+  if (typeof reason === "string") return reason;
+  if (reason instanceof Error) return reason.message;
+  return "aborted";
+}
+
 interface TelegramRawUpdate {
   update_id?: number;
   message?: {
@@ -411,7 +423,26 @@ export async function runTelegramServer(
           // Unrecognized /command — fall through to the LLM.
         }
 
-        // Regular message: enqueue onto this chat's serial chain.
+        // Regular message. If a turn is already in flight for this
+        // chat, abort it — the user typing again means "pay attention
+        // to this instead." The aborted turn returns silently via the
+        // wasAborted path; only the new turn's reply lands. Same UX
+        // as Claude Code's "type to interrupt." The new task still
+        // chains off `prev` (the aborted turn's worker promise) so
+        // the harness's process group has time to clean up before we
+        // spawn the next subprocess.
+        const active = activeTurns.get(msg.chatId);
+        if (active) {
+          log.info("telegram: new message — interrupting active turn", {
+            chatId: msg.chatId,
+            elapsedS: (
+              (Date.now() - active.startTime) / 1000
+            ).toFixed(1),
+          });
+          active.controller.abort("interrupt");
+        }
+
+        // Enqueue onto this chat's serial chain.
         const prev = chatChains.get(msg.chatId) ?? Promise.resolve();
         const next = prev.then(() =>
           processChatMessage(msg, {
@@ -600,14 +631,18 @@ async function processChatMessage(
     }
   }
 
-  // /stop: the controller was aborted from outside. The reply text
-  // already came through from the slash command handler — don't send
-  // a second "(error: stopped)" message.
-  const wasStopped = controller.signal.aborted;
-  if (wasStopped) {
-    log.info("telegram: turn stopped by /stop", {
+  // The controller was aborted from outside. Causes:
+  //   - "stop"      — /stop slash command (slash handler already replied).
+  //   - "reset"     — /reset slash command (handler replied).
+  //   - "interrupt" — a new message arrived for this chat; the new
+  //                   turn supersedes us. Stay silent so only the new
+  //                   reply lands.
+  // In every case, suppress the would-be reply.
+  if (controller.signal.aborted) {
+    log.info("telegram: turn aborted", {
       chatId: msg.chatId,
       durationMs: Date.now() - startedAt,
+      reason: abortReasonString(controller.signal.reason),
     });
     return;
   }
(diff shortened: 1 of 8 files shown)


src/channels/telegram.ts
```diff
@@ -600,15 +631,50 @@ async function processChatMessage(
...
       durationMs: Date.now() - startedAt,
+      reason,
     });
+    // Persist a synthetic interrupted-pair so the next turn knows what
+    // the user just said and that the agent never got to respond.
+    // Without this, follow-ups like "actually use blue instead" land
+    // with no referent in history and the model is surprised. Skip
+    // when:
+    //   - reason === "reset"  → /reset just wiped this conversation;
+    //                           writing here would un-wipe it.
```
lenaparkhodges, line 654: 🔵 **Trivial**: This is great. Persisting the `[interrupted before reply]` message ensures the model doesn't lose the context of the aborted turn when the user follows up.

src/orchestrator/fallback.ts
```diff
@@ -92,6 +92,25 @@ export async function* runWithFallback(
...
+      // assistant text (gemini SIGTERMed mid-stream by an updater
+      // restart, or a tool-only run with no final message). On a
+      // non-last harness, fall through — pi getting a chance is far
+      // better than the user seeing "(no reply)". On the last harness,
+      // yield it and let the channel surface "(no reply)" so the user
+      // knows something happened.
+      if (
+        chunk.type === "done" &&
+        chunk.finalText.length === 0 &&
+        !isLast
```
lenaparkhodges, line 105: 🟡 **Minor**: This is a robust way to handle SIGTERMs or tool-only runs in the primary harness. Falling through to Pi is much better than a `(no reply)` message.

src/orchestrator/turn.ts
```diff
@@ -85,7 +97,7 @@ export async function* runTurn(input: TurnInput): AsyncGenerator<HarnessChunk> {
     systemPrompt,
     userMessage: input.userMessage,
     history,
-    workingDir: input.agentDir,
+    workingDir: input.workingDir ?? homedir(),
```
lenaparkhodges, line 100: 🟡 **Minor**: Changing the default `workingDir` to `homedir()` effectively solves the Gemini sandbox limitations without requiring per-persona configuration.

lenaparkhodges (approved):
## Code Review: fix(telegram+harness)

**Author:** @andrewagrahamhodges | **Files:** 8 | **Verdict:** Approve

This is a solid set of improvements to the Telegram UX and harness reliability. The interruption handling and context preservation are particularly well-implemented.

### Findings

| 1 | 🔵 Trivial | src/channels/telegram.ts | 654 | Excellent context preservation logic. |
| 2 | 🟡 Minor | src/orchestrator/fallback.ts | 105 | Smart fall-through for empty replies. |
| 3 | 🟡 Minor | src/orchestrator/turn.ts | 100 | Defaulting `workingDir` to `homedir()` is a sensible fix for the Gemini sandbox. |

### What's Good
- Comprehensive test coverage for new edge cases (interruption, empty fall-through).
- Improved observability by elevating Gemini stderr to `info` level while filtering noise.
- The "type to interrupt" UX brings the bot's behavior in line with modern AI editors.

<details>

```
Review and approve PR #64 (fix(telegram+harness): no-reply, interrupt UX, interrupt context, gemini sandbox).

The changes are robust and well-tested.
```
</details>

PR #65, opened 2026-05-03
feat(telegram): pass arbitrary attachments through to the harness

andrewagrahamhodges:
> Stacked on top of #64. Review & merge that one first; the GitHub diff for this PR will then automatically rebase to main and show only the attachment work.

## Summary

Phantombot was silently dropping every non-voice attachment (photos, documents, audio files, video, stickers, GIFs) because \`parseGetUpdatesResult\` only recognized \`text\` and \`voice\` fields. From the user's POV the message simply never arrived.

This PR is a deliberate **dumb pipe** — no type-aware preprocessing or content inspection. We download whatever Telegram delivers, save it under \`~/.local/share/phantombot/inbox/<chatId>/<msgId>-<filename>\`, and rewrite the user-facing message to:

\`\`\`
<caption-if-any>

[attached: <absolute-path>]
\`\`\`

The harness decides what to do — \`read\` the file (vision for images, text for docs/code/logs), ignore it, or ask the user for clarification. Phantombot never inspects, transforms, or summarizes the bytes. (Voice keeps its existing dedicated STT path — confirmed working great in production, no need to disturb it.)

### Implementation

- **Parser**: walks every known attachment field (\`document\`, \`photo[]\`, \`audio\`, \`video\`, \`video_note\`, \`animation\`, \`sticker\`), extracts \`file_id\` + optional \`file_name\`, \`mime_type\`, \`file_size\`. For photos (which arrive as an array of sizes), picks the largest by \`file_size\` (or \`width*height\` if size is missing).
- **Filename**: Telegram's \`file_name\` if present; else \`<msg_id>-<kind>.<ext-from-mime>\`. The MIME→extension table covers the common ones; unknown MIMEs get an alphanumeric subtype slice (capped at 8 chars), worst case \`.bin\`.
- **Per-chat inbox**: \`xdgDataHome()/phantombot/inbox/<chatId>/\` so attachments don't pollute the persona dir and grow forever (cleanup is a future cron concern).
- **Edge cases**:
  - **>20 MB** → skip download, surface \`[attached but too large to fetch: <name> (X MB > 20 MB Telegram bot-API cap)]\` so the agent can still acknowledge it. Telegram's \`getFile\` itself caps there regardless of attachment type.
  - **Download error** → \`[attached but couldn't download: <name> (<reason>)]\`.
  - **No caption** → user message is just the bracketed line.
  - **Slash-command path** unaffected: attachment messages have \`text === ""\`, so \`text.startsWith("/")\` is false.

### Live validation on Lena's box

\`\`\`
telegram: incoming   attachment={kind:"photo", fileName:"2201-photo.jpg", fileSize:55778, mimeType:"image/jpeg"}, captionLength:23
telegram: attachment saved   path=/home/lena/.local/share/phantombot/inbox/7995070089/918260260-2201-photo.jpg
orchestrator: trying harness gemini
telegram: complete   durationMs=56624  replyChars=848  progressEvents=13  harness=gemini  ok=true
\`\`\`

The harness's actual invocation argv was:
\`\`\`
gemini -p "Ok, can you see it now?  [attached: /home/lena/.local/share/phantombot/inbox/7995070089/918260260-2201-photo.jpg]" -o stream-json -y
\`\`\`

56s end-to-end, 848-char reply, 13 tool events, no rate-limit retries.

## Test plan

- [x] \`bun run typecheck\` clean
- [x] \`bun test\` — 539 pass / 0 fail (was 521)
- [x] New tests cover: parser (document, photo with multiple sizes, attachment-with-text fallback, voice still routed separately), \`extensionFromMime\` (known/unknown/parameterized/empty), \`extractAttachment\` (synthesized filename, text-only returns undefined), \`formatAttachmentUserText\` (caption+saved, no-caption+saved, oversize, download-error), and 4 integration tests through \`runTelegramServer\` (document download, oversize skip, download failure surface, no-caption case).
- [x] Live: photo with caption → fetched, saved, gemini took the turn cleanly.
- [ ] Try a PDF, a code file, a >20 MB upload — should all behave per the design.

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -15,7 +15,10 @@
  * answered. We log a warning at startup so this isn't accidental.
  */
 
-import type { Config } from "../config.ts";
+import { mkdir, writeFile } from "node:fs/promises";
+import { join } from "node:path";
+
+import { type Config, xdgDataHome } from "../config.ts";
 import type { Harness } from "../harnesses/types.ts";
 import {
   type AudioSupport,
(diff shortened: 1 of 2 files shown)


src/channels/telegram.ts
```diff
@@ -527,6 +796,62 @@ async function processChatMessage(
...
+        chatId: msg.chatId,
+        kind: att.kind,
+        fileName: att.fileName,
+        fileSize: att.fileSize,
+      });
+    } else {
+      try {
+        const dir = inboxDir(msg.chatId);
+        await mkdir(dir, { recursive: true });
+        const path = join(dir, `${msg.updateId}-${att.fileName}`);
```
robertclawson, line 824: **Path-traversal risk (defense-in-depth).** `att.fileName` originates from Telegram's `file_name` field, which is whatever the sending client put on the upload — i.e. user-controlled. `path.join(dir, "${msg.updateId}-${att.fileName}")` will normalize `..` segments, so a malicious or careless filename like `../../etc/escape.bin` resolves out of `inboxDir(chatId)` and lands somewhere else under `~/.local/share/`. With enough `..`s it can escape further.

The practical exposure is low — only allowlisted Telegram users can reach `processChatMessage` at all — but the fix is one line:

```ts
import { basename } from "node:path";
// ...
const safeName = basename(att.fileName).replace(/^\.+/, "_") || "unnamed";
const path = join(dir, `${msg.updateId}-${safeName}`);
```

`basename` strips any directory component (and on Windows handles `\` too), the dot-prefix scrub stops a hostile filename from making the saved file hidden or named purely `.` / `..`, and the `|| "unnamed"` guards against an entirely empty result.

Alternative: ignore the user-supplied name entirely and write `${msg.updateId}-${kind}${ext}` (the same shape `synthesizeFileName` already produces). Slightly less informative on disk, but eliminates the input-trust question end-to-end.

Worth fixing here rather than as a follow-up — it's three lines and we get to stop reasoning about it.

src/channels/telegram.ts
```diff
@@ -216,20 +297,177 @@ function abortReasonString(reason: unknown): string {
...
+
+  return undefined;
+}
+
+function synthesizeFileName(
+  msgId: number,
+  kind: TelegramAttachment["kind"],
+  mimeType: string | undefined,
+): string {
+  return `${msgId}-${kind}${extensionFromMime(mimeType)}`;
```
robertclawson, line 411: Minor / cosmetic: the synthesized filename starts with `${msgId}-`, and then the disk path prepends `${updateId}-` again at line 824, so for any photo / synthesized filename the on-disk path ends up `<updateId>-<msgId>-<kind>.<ext>` — two near-identical numeric prefixes. The integration test path `558-558-photo.jpg` shows it.

Not wrong, just slightly ugly. Either drop the `${msgId}-` from `synthesizeFileName` (the path already gets `updateId`), or drop the `updateId-` prefix when the filename was synthesized. No action required — flagging while we're in the file.

robertclawson (commented):
Solid "dumb pipe" design — matches what the harness is good at, leaves all the type-aware decisioning to the agent. Test coverage is thorough (parser variants + extension table + format helper + 4 end-to-end scenarios through `runTelegramServer`). 539/0 local is the only signal we have right now — see CI note below.

**One concern I'd want addressed before merge:** L1 (line 824) — Telegram's `file_name` is user-controlled and gets `join`-ed into the inbox path without sanitization, so a filename containing `../` segments can escape `inboxDir(chatId)`. Cheap defense-in-depth: `basename()` it. Detail in inline comment.

**Smaller observations (non-blocking):**

- **CI didn't run on this branch.** Because the base is `fix/empty-done-falls-through` (#64), not `main`, the workflow `pull_request` triggers on phantombot's main protections didn't fire. The branch protection (typecheck+test required) will kick in once this rebases onto main after #64 merges — but worth re-pushing or force-rebasing to trigger CI explicitly before final merge, so we have a green tick on this exact head SHA, not just your local run.
- **Inbox grows unbounded** — you flagged it inline ("cleanup is a future cron concern"). Worth opening a tracking issue so it doesn't get forgotten; a daily/weekly tick task that prunes anything older than N days would do the job.
- **Photos hardcode `image/jpeg`** — accurate for Telegram (it always re-encodes photos to JPEG), but worth a one-line code comment so the next reader doesn't think it's a bug.
- **Stacked-PR caveat** acknowledged in the description; the diff I reviewed is the cross-branch diff (`feat/telegram-attachments` vs `fix/empty-done-falls-through`) — exactly the attachment work, no overlap with #64.

Going **COMMENTED** rather than approve so the path-traversal sanitization gets a yes/no from you. Once that lands I'll re-approve and ship.

PR #69, opened 2026-05-03
Deploy landing page to GitHub Pages

lenaparkhodges:
This PR adds a GitHub Action to automatically deploy the 'www/' directory to the 'gh-pages' branch and configures the 'CNAME' for phantombot.bot.

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -1,0 +1,24 @@
+name: Deploy to GitHub Pages
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'www/**'
+  workflow_dispatch:
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v4
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./www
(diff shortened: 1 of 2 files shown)


kaieriksen (dismissed):
Clean PR — two files, well-scoped.

**What works:**
- `peaceiris/actions-gh-pages@v4` is a well-known, maintained action
- Path filter `www/**` — only fires when landing page content actually changes, no wasted CI
- `workflow_dispatch` for manual redeploys
- Permission scoped to `contents: write` at the job level

**One note (non-blocking):** The action's `cname` param and `www/CNAME` file are redundant — the action creates the CNAME from the param. Having both won't break anything, but you could drop `www/CNAME` and rely solely on the action param. Up to you.

**Ops prerequisite:** `phantombot.bot` DNS must point to GitHub Pages IPs before the CNAME takes effect. Since that's ops, not code — just flagging it.

Good to merge from my side.

kaieriksen:
One minor note on `.github/workflows/deploy-pages.yml` line 23: the `cname` param automatically creates a CNAME file in the deployed output, so `www/CNAME` is redundant. Not harmful — just a minor DRY opportunity if you want.

kaieriksen (approved):
Re-reviewed after the new commit. 

Lena resolved the earlier note by removing `www/CNAME` — now only the action's `cname` param remains. No duplication, no confusion.

Workflow looks right: path-filtered to `www/**`, dispatchable, minimal permissions. Good to go.

Ops reminder: `phantombot.bot` DNS still needs to point at GitHub Pages IPs for the CNAME to take effect.

PR #71, opened 2026-05-04
feat(cli): add `phantombot ask` for one-shot prompts

robertclawson:
## Summary
- Adds `phantombot ask "<prompt>"` — a stateless one-shot through the persona + harness chain. Prints the assistant's final reply to stdout and exits.
- Designed primarily for the Twilio voice-agent's `askRobbie` relay (replacing the old HTTP POST to OpenClaw at `127.0.0.1:18789`), but useful for any local script that wants Robbie's brain without a Telegram conversation.
- Stateless by default (`noHistory: true`, conversation key `cli:ask`); pass `--history --conversation <id>` to thread asks.
- 9 new test cases; existing dispatcher whitelist test updated.

## Test plan
- [x] `bun test` — 590 pass / 1 skip / 0 fail (591 across 55 files)
- [ ] After merge & release: `phantombot update` on this VPS, smoke-test with `phantombot ask "who are you?"`
- [ ] Propagate to lena/kai/jake on kw-openclaw via the documented two-step (binary swap + manual `systemctl --user restart` with explicit `XDG_RUNTIME_DIR`).

### DIFF
--- a/src/cli/ask.ts
+++ b/src/cli/ask.ts
@@ -161,7 +161,13 @@ export default defineCommand({
   async run({ args }) {
     let prompt = String(args.prompt ?? "");
     if (prompt === "-") {
-      prompt = await readAllStdin();
+      try {
+        prompt = await readAllStdin();
+      } catch (e) {
+        process.stderr.write(`phantombot ask: ${(e as Error).message}\n`);
+        process.exitCode = 2;
+        return;
+      }
     }
     process.exitCode = await runAsk({
       prompt,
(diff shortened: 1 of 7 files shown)


kaieriksen (dismissed):
Clean implementation. Well-tested, well-documented.

- Stateless by default with opt-in history threading is the right call for a non-interactive caller
- Exit codes are clear and consistent (0/1/2)
- Proper memory lifecycle (open/close)
- Error messages go to stderr, output goes to stdout — composable with pipes

One observation, non-blocking: `readAllStdin` will hang forever if stdin is a TTY and nothing is piped in. That's standard behavior for `-` readers, so not a bug — just calling it out in case it surprises someone during interactive testing.

kaieriksen (commented):
{"path":"src/cli/ask.ts","line":175,"body":"Non-blocking: `readAllStdin` will hang forever if stdin is a TTY with no data piped — standard for `-` readers, but worth a timeout in a future pass for interactive callers who mistype."}

src/cli/ask.ts
```diff
@@ -0,0 +1,181 @@
...
+    process.exitCode = await runAsk({
+      prompt,
+      persona: args.persona ? String(args.persona) : undefined,
+      conversation: args.conversation ? String(args.conversation) : undefined,
+      history: Boolean(args.history),
+    });
+  },
+});
+
+async function readAllStdin(): Promise<string> {
```
kaieriksen: Non-blocking: `readAllStdin` will hang forever if stdin is a TTY with no data piped — standard for `-` readers, but worth considering a timeout in a future pass for accidental interactive use.

kaieriksen (commented):
Non-blocking observation.

src/cli/ask.ts
robertclawson: Good catch — fixed in c72ca25. `readAllStdin` now checks `stdin.isTTY` and throws fast with a message telling the caller exactly how to fix it (pipe input, or pass the prompt as an arg). Citty handler catches → stderr + exit 2. Added two tests: TTY rejects, piped still reads. Full suite: 592 pass.

kaieriksen (approved):
TTY guard resolves the hang risk perfectly — fast fail with a clear message. Tests cover both TTY rejection and normal pipe flow. Clean fix.

Approved.

PR #72, opened 2026-05-04
fix: add ~/.pi/agent/bin to systemd unit PATH

kaieriksen:
## Problem

Pi CLI installs to `~/.pi/agent/bin/pi` by default. That directory isn't in systemd's fixed PATH, so when phantombot's orchestrator spawns `pi`, Bun returns:

> error: Executable not found in $PATH: "pi"

The harness's `available()` method only validates absolute paths — with `bin = "pi"` (PATH lookup) it always returns `true`, so the chain never falls through cleanly. It blows up on spawn instead.

## Fix

Prepends `%h/.pi/agent/bin` to the `Environment=PATH` line in all four unit templates (user/system × normal/fast-channel).

## Tested

- All 20 systemd unit tests pass
- PATH now resolves `~/.pi/agent/bin` before `~/.local/bin`

### DIFF
--- a/src/lib/systemd.ts
+++ b/src/lib/systemd.ts
@@ -97,7 +97,7 @@ Type=simple
 ExecStart=${exec}
 Restart=on-failure
 RestartSec=5
-Environment="PATH=%h/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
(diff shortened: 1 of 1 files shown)


robertclawson (approved):
LGTM. Surgical fix — prepending `~/.pi/agent/bin` to PATH in all four unit templates is exactly right, and order vs. `~/.local/bin` is safe (Pi CLI only ships `pi` there). CI green, typecheck + 20 systemd tests pass.

The `available()` blind spot you called out (PATH-lookup commands always return true → blows up on spawn instead of falling through) is real and worth a follow-up issue, but agree it's out of scope for this PR.

PR #73, opened 2026-05-04
feat(ask): --stream flag for low-latency voice agent

robertclawson:
## Why

The voice agent calls `phantombot ask ...` per turn. In non-stream mode the answer arrives as one final blob, so Twilio TTS can't speak until generation finishes — calendar/quick-fact questions take ~25-30s end to end on a live call.

## What

Adds an opt-in `--stream` flag (and `stream` field on `RunAskInput`):
- When set, assistant `text` chunks are flushed to stdout as they arrive.
- When unset (default), behaviour is unchanged: harness's authoritative `finalText` is written once at the end.

Tool calls and chain-of-thought never reach us as `text` chunks — runTurn surfaces those as `progress`/`heartbeat` and we drop them in both modes. So what hits stdout in stream mode is exactly the assistant's spoken reply, same as before.

The `done` event sets `succeeded` and we skip the final write in stream mode (otherwise we'd duplicate the reply on stdout).

## Caller side

The voice-agent repo's streaming path is gated behind `VOICE_AGENT_STREAM_ASK=1` and shells out to a separate binary via `PHANTOMBOT_STREAM_BIN`. After this lands, both can collapse to the single binary with `--stream` flipped on.

## Risk

Low. Flag is off by default; no existing caller passes it. Contained to `src/cli/ask.ts`.

## Test plan

- [x] Smoke test: `phantombot ask --stream "hello"` prints chunks and exits 0.
- [ ] Live Twilio call with voice-agent on `VOICE_AGENT_STREAM_ASK=1` — confirm latency drop and no duplicate reply on stdout.

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -959,6 +959,14 @@ async function processChatMessage(
       systemPromptSuffix: willReplyWithVoice
         ? `${TELEGRAM_REPLY_INSTRUCTION}\n\n${VOICE_REPLY_INSTRUCTION}`
         : TELEGRAM_REPLY_INSTRUCTION,
+      // Pre-tool narration: ON for text-out (the user sees streamed
+      // text as it lands, so a "checking your calendar..." sentence
+      // before a tool call usefully fills the silence). OFF for
+      // voice-out: the reply is synthesized as one clip at the end,
+      // so narration would just lengthen the spoken output without
+      // helping with perceived latency. VOICE_REPLY_INSTRUCTION
+      // already forbids work narration too, so off is consistent.
+      toolNarration: !willReplyWithVoice,
     })) {
       if (chunk.type === "text") {
         reply += chunk.text;
--- a/src/cli/ask.ts
+++ b/src/cli/ask.ts
@@ -45,6 +45,17 @@ export interface RunAskInput {
   conversation?: string;
   /** Persist + load history for this conversation. Default false (stateless). */
   history?: boolean;
+  /**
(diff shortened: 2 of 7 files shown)


kaieriksen (commented):
## Review: PR #73 — pre-tool narration

**Clean feature, solid tests, no blockers.** The approach of letting the model generate narration in the user's language (rather than harness-emitted English filler) is the right call — avoids jarring language leaks for non-English users.

### Two minor notes

1. **Trailing newline edge case (ask.ts stream mode):** On `done`, `finalText` is overwritten from `chunk.finalText`, but the trailing-newline check uses that value — not what was actually streamed. In practice this should be fine (the harness's `finalText` and accumulated stream chunks will agree on trailing whitespace in all normal cases), just noting the indirection.

2. **No assertion on streamed output contents:** The new tests verify narration injection correctly, but don't assert that `out` actually receives the text chunks in `--stream` mode. A quick expect on the CapturingSink would make the streaming behavior explicitly tested rather than implicitly trusted. Not blocking — the mechanism is straightforward enough.

### What I like

- `toolNarration` off by default, channels opt in — correct default posture
- Telegram voice-out correctly suppresses narration (one-shot clip, no silence to fill)
- Three test layers (orchestrator → CLI → Telegram) all covered
- Multilingual nudge in the prompt itself, not just in comments
- Clean overlay ordering (suffix → narration, narration closest to user message)

kaieriksen (approved):
All notes from previous review addressed. Edges covered, CI green. Ship it.

PR #74, opened 2026-05-04
feat(telegram): flush narration as separate bubble before each tool call

robertclawson:
## Summary

Follow-up to #73. Today the Telegram channel batches the entire turn into one bubble at the end, so the pre-tool narration that PR #73 wired up ("Checking your inboxes…") gets stapled to the front of the final answer instead of arriving during the silent tool-run window — exactly the gap it was supposed to fill.

This PR makes the text-out path flush whatever prose has accumulated since the last flush as its own Telegram message right when a `progress` chunk arrives (i.e. just before a tool fires). Multi-tool turns get one bubble per tool, then the post-tool answer.

## What the user sees

**Before:**
> User: any new email from turtle crossing?
> *[15s of silence, typing indicator]*
> Robbie: Checking your inboxes now... Found 3 threads, most recent is Corinne's renewal-notice forward from today...

**After:**
> User: any new email from turtle crossing?
> Robbie: Checking your inboxes now…
> *[15s, typing indicator]*
> Robbie: Found 3 threads, most recent is Corinne's renewal-notice forward from today...

Bigger payoff on multi-tool turns — "any urgent emails or calendar blockers?" now lands as three messages ("Checking your calendar…" → "Now scanning your inboxes…" → final answer) instead of 30 seconds of dead air.

## Mechanics

Three accumulators:
- `streamedReply` — running sum of `text` chunks
- `flushedReply`  — what's already been sent as narration bubbles
- `finalReply`    — authoritative full text from the `done` chunk

On each `progress` chunk: send `streamedReply.slice(flushedReply.length)` if it has non-whitespace content, then update `flushedReply`.

After the loop, the final bubble logic handles four cases:
- **Happy path**: `finalReply.startsWith(flushedReply)` → send only the suffix (no duplicated narration)
- **Reformat**: prefix doesn't match → send full `finalReply`. Duplication beats truncation; we'd rather show the whole answer with a repeated word than slice off bytes the user needs.
- **Silent after tool**: model emitted nothing further → stay quiet (the narration bubbles are the user-visible reply; don't tack on "(no reply)").
- **Error**: surface the diagnostic regardless of what was flushed.

## Voice-out is unchanged

Voice replies are one synthesized clip at the end, so chunked text flushes would both bloat the audio and leak text into a voice-only conversation. The flush function is gated off via `willReplyWithVoice` and the voice path always synthesizes the full reply (not the prefix-trimmed `outText`).

## Test coverage

9 new tests in `tests/channels-telegram.test.ts`:
1. Happy path two-bubble flow (text → progress → text → done)
2. Multi-tool one-bubble-per-tool
3. No-progress regression: single bubble with full reply
4. Whitespace-only buffer doesn't flush an empty bubble
5. Prefix-match: post-tool message is the suffix only (no duplication)
6. Harness reformat: full reply sent (graceful fallback)
7. Silent after tool: no "(no reply)" tacked on
8. Error after a flushed narration bubble: error follows the narration
9. Voice-out: progress chunks do NOT flush; full reply synthesized at end

611/0 fail, typecheck clean.

## Test plan

- [x] Full test suite green (`bun test` → 611 pass / 0 fail / 1 skip)
- [x] Typecheck clean (`bun run typecheck`)
- [ ] Manual: text → tool → text turn lands as two Telegram bubbles in order
- [ ] Manual: multi-tool turn lands as one bubble per tool + one final answer
- [ ] Manual: voice-in/voice-out turn still arrives as a single voice clip
- [ ] Manual: no-tool turn (e.g. "thanks") still lands as one bubble

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -6,10 +6,16 @@
  * memory uses conversation key `telegram:<chatId>` so DMs and groups are
  * isolated from the CLI's `cli:default` history.
  *
- * Streaming: we send `sendChatAction(typing)` at the start of each turn so
- * the user sees "Phantom is typing…" while the harness runs, then post the
- * final reply as one message. Live token-by-token edits would be nicer but
- * Telegram rate-limits edits at ~1/sec — not worth the complexity for v1.
+ * Streaming: we send `sendChatAction(typing)` at the start of each turn,
+ * refresh it on every harness chunk, and — for text-out turns — flush
+ * any buffered prose as a separate Telegram message right before each
+ * tool call (signalled by a `progress` chunk). That makes one-line
+ * narrations like "Checking your inboxes…" land *before* the silent
+ * tool-run window instead of arriving stapled to the back of the final
+ * answer. Voice-out skips chunked flushing — voice is one synthesized
+ * clip at the end, mid-turn flushes would just bloat the audio.
+ * Token-by-token live edits would be nicer still but Telegram
+ * rate-limits edits at ~1/sec — not worth the complexity.
  *
  * Auth gating: if `allowedUserIds` is empty, anyone who DMs the bot is
  * answered. We log a warning at startup so this isn't accidental.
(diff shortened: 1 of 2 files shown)


kaieriksen (dismissed):
Clean follow-up to #73. The narration-before-tool UX improvement is exactly right — multi-bubble delivery during tool silence is what makes the pre-tool narration actually useful.

**What I checked:**
- Three-accumulator design (streamedReply / flushedReply / finalReply) is sound
- Four final-bubble cases all make sense: suffix-only, full-reply fallback, silent-after-tool, error-surfaces
- Voice-out correctly gates off flushing — one synthesized clip, no text leakage
- Whitespace-only buffer guard prevents empty Telegram bubbles
- Typing indicator re-armed after each flush (Telegram drops it on message send)
- 9 new tests cover all paths including multi-tool and reformat edge cases
- Old early-abort guard removal is safe: the `finally` block handles `activeTurns.delete`, and aborting a finished harness is a no-op
- No secrets or personal data in the diff

**Minor nit (non-blocking):** `refreshIndicator()` is called inside `flushNarration()` and again after `await flushNarration()` in the progress handler. Harmless double-call — `sendChatAction` is idempotent — but worth tidying if you touch this code again.

No blockers. Approve and merge at will.

kaieriksen (approved):
Re-review: second commit removes the redundant refreshIndicator() and the comment makes it self-documenting. CI still green. Still a clean merge.

PR #75, opened 2026-05-04
feat: harness fast-fallback on 4XX + per-harness cooldown with backoff

robertclawson:
## Why

Lena is sitting on Gemini 429s right now. The `gemini-cli` retryWithBackoff loop burns ~2 minutes on a single 4XX before exiting non-zero, and only then does the orchestrator fall through to `pi`. Total user-facing latency: 158 s. We want fallback to fire **immediately** on a 4XX and we want a failed harness to skip ahead on subsequent turns instead of being asked again every time.

### 1. Fast 4XX fallback (Gemini harness)
- New stderr scanner watches for `Attempt N failed with status (4\d\d)` and `status: (4\d\d),` (the two shapes gemini-cli's Gaxios wrapper actually emits).
- On first 4XX hit, SIGTERMs the process group and yields a recoverable error chunk with the captured `httpStatus`.
- 5XX is **intentionally** not matched — server-side blips usually clear faster than a fallback hop.
- HarnessChunk.error gains optional `httpStatus?: number` for the orchestrator + log forensics.

### 2. Per-harness cooldown (`src/lib/cooldown.ts`)
In-memory `CooldownStore` keyed by harness id. Each recoverable failure bumps a consecutive-failure counter and arms a cooldown window with **±25% jitter**:

| failures | base window |
|----------|-------------|
| 1        | 150 s |
| 2        | 300 s |
| 3        | 600 s |
| 4        | 1200 s |
| 5        | 2400 s |
| 6+       | 3600 s (cap) |

A successful turn (`done` with non-empty text) clears the counter. Window expiry alone does **not** reset the counter — a near-immediate re-failure escalates to the next tier rather than restarting at 150 s.

Process-local; resets on phantombot restart. That's fine — cooldown is a soft "this is probably still broken" hint.

### 3. Orchestrator wiring (`src/orchestrator/fallback.ts`)
- Snapshots cooldown state for the whole chain at turn start (so failures we register mid-turn don't cascade-skip the next harness in this same turn).
- Cooled-down harnesses are skipped in chain order.
- **Escape hatch**: if every harness is currently cooled, the snapshot is ignored — we'd rather give Andrew a possibly-flaky reply than refuse outright.
- Three-harness chains explicitly tested: if primary AND fallback both fail recoverably, the third gets the turn (per Andrew's request).

## Tests

27 new tests, full suite **638/0/1** (pass/fail/skip):

- `tests/lib-cooldown.test.ts` — schedule, jitter bounds, success reset, near-immediate re-failure tier progression, per-id keying.
- `tests/harnesses-gemini.test.ts` — 4XX stderr scanner (matches retryWithBackoff trace, ignores 5XX/2XX), end-to-end fast-fallback timing (must return well under hardTimeoutMs).
- `tests/orchestrator-fallback.test.ts` — recoverable → markFailure, success → markSuccess, skip-when-cooled, escape hatch, three-harness traversal, snapshot semantics.

New fake-gemini fixture mode `429-then-hang` mimics gemini-cli's actual stderr shape (verified against journalctl output from Lena).

## What this fixes for Lena right now

Next time her gemini hits a 429:
1. Stderr scanner fires within ~10 ms of the first retry log line.
2. Harness SIGKILLs gemini-cli (no more waiting for its internal 6-attempt retry budget).
3. Orchestrator falls through to `pi` immediately.
4. Cooldown marks gemini as off-limits for ~150 s (jittered) — subsequent turns skip straight to `pi` without even trying gemini.
5. After ~150 s, gemini gets one more shot. If it fails again, cooldown grows to ~300 s, then ~600 s, etc.

## Test plan

- [ ] Kai's review on the cooldown module + orchestrator wiring
- [ ] After merge: rebuild + push v1.0.75; run `phantombot update --force --restart` on all four agents
- [ ] Send Lena a message; confirm fast fallback in logs (look for `gemini stderr: 4XX detected, killing early for fast fallback`)
- [ ] Confirm cooldown skip on subsequent turn (look for `orchestrator: skipping harness — cooldown active`)

### DIFF
--- a/src/harnesses/gemini.ts
+++ b/src/harnesses/gemini.ts
@@ -43,7 +43,7 @@ import {
   killCauseToErrorChunk,
 } from "../lib/harnessRunner.ts";
 import { log } from "../lib/logger.ts";
-import { spawnInNewSession } from "../lib/processGroup.ts";
+import { killProcessGroup, spawnInNewSession } from "../lib/processGroup.ts";
 
 /**
  * Conservative ceiling on the bytes we'll put on argv via the `-p`
@@ -154,7 +154,24 @@ export class GeminiHarness implements Harness {
       });
     }
 
-    void consumeStderr(proc.stderr);
+    // Watch stderr for HTTP 4XX errors emerging from gemini-cli's
+    // built-in retryWithBackoff loop. Once we spot a 4XX, kill the
+    // subprocess immediately so the orchestrator can fall through to
+    // the next harness without waiting for gemini-cli to finish its
+    // own ~2-minute retry budget. The status is stored here and
+    // surfaced in the post-loop error chunk.
+    const httpStatusBox: { status?: number } = {};
+    void consumeStderr(proc.stderr, (status) => {
+      if (httpStatusBox.status !== undefined) return; // first one wins
+      httpStatusBox.status = status;
+      log.warn("gemini stderr: 4XX detected, killing early for fast fallback", {
+        httpStatus: status,
+      });
+      // Fire-and-forget — process group SIGTERM → 5 s grace → SIGKILL.
+      // The for-await over stdout below ends naturally as the kernel
+      // closes the pipe.
+      void killProcessGroup(proc, 5000);
+    });
 
     let buffer = "";
     let finalText = "";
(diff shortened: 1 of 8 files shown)


kaieriksen (approved):
Clean PR. Thoroughly reviewed all three layers.

**gemini.ts** — The stderr scanner is tight. First-wins semantics via httpStatusBox, correctly scoped to 4XX only (5XX stays with upstream retry). The post-loop priority order (4XX > kill cause > exit code > done) is correct — even if gemini-cli races toward exit 0, the 4XX error takes priority and the orchestrator gets the right signal.

**cooldown.ts** — Clean module. Injectible seams (random/now) for deterministic testing. The "near-immediate re-failure escalates" behavior (window expiry preserves the counter) is the right call — prevents stampeding a flaky upstream. ±25% jitter prevents fleet-wide convergence. Export-level singleton with test isolation via constructor injection.

**fallback.ts** — Good. Snapshot at turn start means mid-turn failures don't cascade-skip later harnesses in the same turn. Escape hatch (all-cooled → ignore snapshot) is the right safety valve. The empty-done-on-non-last → markFailure + break (before succeeded=true) correctly avoids the markSuccess/markFailure contradiction. Three-harness traversal locked in by tests per Andrew's request.

**Tests** — 27 new, covering all the right edge cases: cooldown scheduling, jitter bounds, success reset, near-immediate re-failure escalation, 4XX pattern matching (real gemini-cli stderr shapes), 5XX/2XX exclusion, orchestrator skip-when-cooled, escape hatch, snapshot semantics. The '429-then-hang' fixture mirrors real journalctl output from Lena.

No blockers. Merge at will.

PR #76, opened 2026-05-04
fix(pi): emit progress for tool events instead of heartbeat

kaieriksen:
## Problem

PR #74 (telegram narration bubbles) added a working mechanism: when a harness emits a `progress` chunk, the Telegram channel layer calls `flushNarration()` which sends buffered text as a separate bubble before the tool runs.

The pi harness's `parsePiEvent()` was never wired up to emit `progress` — it mapped all non-`text_delta` events to `heartbeat`. Heartbeats only refresh the typing indicator and never trigger `flushNarration()`. Result: zero progress events per turn, all text lands in one bubble at the end. The feature was dead code for pi agents.

Evidence from Kai's own logs reviewing PR #75: `progressEvents: 0, flushedChars: 0`.

## Fix

Two changes to `parsePiEvent()`:

1. **`tool_execution_start`** (top-level pi event) → `{ type: "progress", note: "tool: <name>" }`
2. **`tool_use_*`** (assistantMessageEvent subtype) → `{ type: "progress", note: "tool" }`

`thinking_delta` and other event types remain `heartbeat` — unchanged.

## Why this works

The channel layer's progress handler calls `flushNarration()`, which sends buffered text as a Telegram message and re-arms the typing indicator. Multiple progress events without intervening text are harmless no-ops (flushNarration checks for whitespace-only pending text before sending).

## Tests

Two new test cases verifying:
- `tool_execution_start` → progress with tool name
- `tool_use_start` / `tool_use_end` / `tool_use` → progress

All existing tests pass unchanged.

### DIFF
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -261,10 +261,10 @@ export function renderStdinPayload(req: HarnessRequest): string {
  *
  * Channel layers want two distinct signals from us:
  *   - `text` blocks → user-visible reply (concatenate, surface verbatim).
- *   - Anything else (`thinking`, `tool_use`, `tool_result`, …) → "model
- *     is alive" tick, no payload. The actual content stays inside the
- *     subprocess; we just emit `heartbeat` so the channel can refresh
- *     its typing indicator without leaking chain-of-thought.
+ *   - Anything else (`thinking`, `tool_use`, `tool_result`, …) → `progress`
+ *     so the channel layer can flush pending narration before the model
+ *     runs its tool. Actual content stays inside the subprocess; we never
+ *     leak chain-of-thought.
  *
  * If a single assistant message contains BOTH text and non-text blocks,
  * we prefer the text chunk (it carries strictly more signal — text
@@ -295,7 +295,7 @@ export function parseStreamJson(parsed: unknown): HarnessChunk | undefined {
     }
   }
   if (text) return { type: "text", text };
-  if (sawNonText) return { type: "heartbeat" };
+  if (sawNonText) return { type: "progress" };
   return undefined;
 }
 
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -251,24 +251,28 @@ export function renderStdinPayload(req: HarnessRequest): string {
 
 /**
  * Translate one stream-json line into a HarnessChunk. Returns undefined for
- * lines we want to ignore (e.g. tool-use events that the agent handles
- * internally and doesn't need to surface to phantombot).
+ * lines we want to ignore.
  *
  * Claude's stream-json schema is documented in the Claude Code docs but
  * informally: each line has a `type` (system / user / assistant / result)
  * and a `message` payload. The assistant content is an array of blocks
  * with their own `type`: `text`, `thinking`, `tool_use`, `tool_result`.
  *
- * Channel layers want two distinct signals from us:
+ * Channel layers want three distinct signals from us:
  *   - `text` blocks → user-visible reply (concatenate, surface verbatim).
- *   - Anything else (`thinking`, `tool_use`, `tool_result`, …) → `progress`
- *     so the channel layer can flush pending narration before the model
- *     runs its tool. Actual content stays inside the subprocess; we never
- *     leak chain-of-thought.
+ *   - `tool_use` blocks → `progress` so the channel layer can flush pending
+ *     narration before the model runs its tool.
+ *   - `thinking` / `tool_result` → `heartbeat` (refreshes typing indicator,
+ *     but does NOT flush narration — mirrors pi.ts behavior).
  *
  * If a single assistant message contains BOTH text and non-text blocks,
- * we prefer the text chunk (it carries strictly more signal — text
- * implicitly proves the model is alive too).
+ * text wins (it carries strictly more signal). If it has both tool_use
+ * and thinking, progress wins (tool_use is the signal that matters).
+ * Thinking-only messages get a heartbeat — they don't fragment the
+ * narration bubble.
+ *
+ * Actual content stays inside the subprocess; we never leak
+ * chain-of-thought.
  *
  * Exported for testing.
  */
@@ -283,19 +287,23 @@ export function parseStreamJson(parsed: unknown): HarnessChunk | undefined {
   if (!Array.isArray(content)) return undefined;
 
   let text = "";
-  let sawNonText = false;
+  let toolName: string | undefined;
+  let sawOtherNonText = false;
   for (const part of content) {
     if (typeof part === "object" && part !== null) {
       const p = part as Record<string, unknown>;
       if (p.type === "text" && typeof p.text === "string") {
         text += p.text;
+      } else if (p.type === "tool_use") {
+        toolName = typeof p.name === "string" ? p.name : toolName ?? "tool";
       } else if (typeof p.type === "string") {
-        sawNonText = true;
+        sawOtherNonText = true;
       }
     }
   }
   if (text) return { type: "text", text };
-  if (sawNonText) return { type: "progress" };
+  if (toolName) return { type: "progress", note: `tool: ${toolName}` };
+  if (sawOtherNonText) return { type: "heartbeat" };
   return undefined;
 }
 
--- a/src/harnesses/pi.ts
+++ b/src/harnesses/pi.ts
@@ -8,7 +8,7 @@
  *
  * Stream-json events translated to phantombot HarnessChunks:
  *   message_update with text_delta  → { type: "text", text }
- *   tool_execution_start            → { type: "progress", note: "running <tool>" }
+ *   tool_execution_start            → { type: "progress", note: "tool: <name>" }
  *   anything else (agent_start,
  *     tool_execution_end, turn_end,
  *     extension_*) → ignored        (the done chunk is emitted from process exit)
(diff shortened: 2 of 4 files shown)


src/harnesses/pi.ts
```diff
@@ -211,14 +211,25 @@ export function renderPayload(req: HarnessRequest): string {
...
   if (typeof parsed !== "object" || parsed === null) return undefined;
   const obj = parsed as Record<string, unknown>;
+
+  // tool_execution_start is a top-level event pi emits just before
+  // it invokes a tool. Emit as `progress` so the channel layer
+  // flushes any buffered narration into a bubble before the tool
+  // runs (keeping the user oriented during the silence).
+  if (obj.type === "tool_execution_start") {
+    const toolName = typeof obj.tool_name === "string" ? obj.tool_name : undefined;
+    return { type: "progress", note: toolName ? `tool: ${toolName}` : "tool" };
```
robertclawson, line 230: Correctness check on this branch: `flushNarration()` is whitespace-safe, so if pi fires both a top-level `tool_execution_start` and an inner `tool_use_start` for the same call, the second `progress` chunk just no-ops in the channel layer. No double-bubble risk. 👍

src/harnesses/pi.ts
```diff
@@ -232,10 +243,18 @@ export function parsePiEvent(parsed: unknown): HarnessChunk | undefined {
...
-  // thinking_delta + any other assistantMessageEvent type → heartbeat.
-  // We intentionally do NOT include the content in the chunk (would leak
-  // chain-of-thought to the user). The signal is just "pi is alive."
   if (typeof ame.type === "string") {
+    // tool_use_* assistantMessageEvent: the model has decided to call
+    // a tool — emit `progress` so the channel layer flushes narration
+    // into a bubble before the tool runs. Previously these were mapped
+    // to heartbeats, which kept the typing indicator alive but never
+    // triggered a bubble flush (defeating the purpose of PR #74).
+    if ((ame.type as string).startsWith("tool_use")) {
```
robertclawson: Tiny nit (not a blocker): the `(ame.type as string)` cast is redundant — TS already narrowed `ame.type` to `string` via the `typeof ame.type === "string"` guard on the enclosing `if`. Cosmetic only; ignore if you don't want to bother.

robertclawson (dismissed):
Reviewed — good to merge as-is.

**What it does:** Wires pi's tool events to the `progress` chunk type so `flushNarration()` actually fires. Before this, pi mapped everything except `text_delta` to `heartbeat`, which only refreshes the typing indicator — it never triggered the narration-bubble flush that PR #74 was supposed to enable. So the feature was dead code on pi turns. Good catch.

**Correctness checks:**
- `flushNarration()` is whitespace-safe, so back-to-back progress events (e.g. if pi fires both `tool_execution_start` and a `tool_use_start` for the same call) just no-op the second one. Safe.
- No existing test had `tool_use_*` pinned to heartbeat, so nothing flipped behaviour-wise.
- Sits cleanly on top of the #75 cooldown work — no overlap.
- CI: test ✅, typecheck ✅.

Verdict: ship it.

lenaparkhodges (commented):
PR #76 is a solid fix for the Pi agent narration issue. By correctly mapping `tool_execution_start` and `tool_use_*` events to `progress` instead of `heartbeat`, we ensure that the Telegram channel's `flushNarration()` is triggered before tools run. This restores the intended behavior from PR #74 for Pi-based agents.

src/harnesses/pi.ts
```diff
@@ -211,14 +211,25 @@ export function renderPayload(req: HarnessRequest): string {
...
  */
 export function parsePiEvent(parsed: unknown): HarnessChunk | undefined {
   if (typeof parsed !== "object" || parsed === null) return undefined;
   const obj = parsed as Record<string, unknown>;
+
+  // tool_execution_start is a top-level event pi emits just before
+  // it invokes a tool. Emit as `progress` so the channel layer
+  // flushes any buffered narration into a bubble before the tool
+  // runs (keeping the user oriented during the silence).
+  if (obj.type === "tool_execution_start") {
```
lenaparkhodges, line 228: Good addition. Reclassifying `tool_execution_start` as `progress` is the key to triggering the narration flush. Including the `tool_name` in the `note` is a nice touch for visibility.

src/harnesses/pi.ts
lenaparkhodges: Correct. Using `startsWith('tool_use')` ensures all tool-related assistant events are captured, addressing the mapping gap that was keeping the feature 'dead' for Pi agents.

tests/harnesses-pi.test.ts
```diff
@@ -92,6 +92,31 @@ describe("parsePiEvent", () => {
     expect(JSON.stringify(c)).not.toContain("internal reasoning");
   });
 
+  test("emits progress for tool_execution_start (top-level event)", () => {
```
lenaparkhodges, line 95: Excellent coverage here. Testing both with and without `tool_name` ensures the progress chunk is robust.

src/harnesses/pi.ts
robertclawson, line 230: Cross-harness audit (re: Andrew's question about whether claude/gemini need the same fix):

**gemini.ts is already correct.** `parseGeminiEvent` (lines 334–338) maps `tool_use` → `progress` with `note: "tool: <name>"` already, in the exact same shape this PR introduces for pi. So gemini turns already trigger `flushNarration()` — no follow-up needed there.

src/harnesses/pi.ts
robertclawson: **claude.ts has this same bug, though.** `parseStreamJson` (claude.ts lines 285–299):

```ts
let text = "";
let sawNonText = false;
for (const part of content) {
  if (typeof part === "object" && part !== null) {
    const p = part as Record<string, unknown>;
    if (p.type === "text" && typeof p.text === "string") {
      text += p.text;
    } else if (typeof p.type === "string") {
      sawNonText = true;
    }
  }
}
if (text) return { type: "text", text };
if (sawNonText) return { type: "heartbeat" };
```

When an assistant message contains a `tool_use` block with no accompanying text (the common case — Claude usually narrates first, then emits a separate `tool_use`-only block immediately before the call), the `sawNonText` branch fires and the harness emits **`heartbeat`** — same dead-end this PR fixes for pi. Result: PR #74's narration-flush feature doesn't render on Claude turns either.

Suggest a follow-up PR mirroring this one: detect the `tool_use` block specifically (block has `p.type === "tool_use"`, tool name in `p.name`), emit `{ type: "progress", note: \`tool: ${name}\` }`. Keep `thinking` / `tool_result` mapping to heartbeat as today. Worth keeping the architectural-warning comment at the bottom of claude.ts intact — this isn't tool-call passthrough, just event-type signalling.

Not blocking #76 — flagging for the next PR.

src/harnesses/claude.ts
```diff
@@ -295,7 +295,7 @@ export function parseStreamJson(parsed: unknown): HarnessChunk | undefined {
     }
   }
   if (text) return { type: "text", text };
-  if (sawNonText) return { type: "heartbeat" };
+  if (sawNonText) return { type: "progress" };
```
robertclawson: **Typecheck error.** `HarnessChunk`'s progress variant is `{ type: "progress"; note: string }` — `note` is required, so this fails `bun tsc --noEmit` (and it's already failing on CI for this commit). Needs a note string. Suggest mirroring pi.ts and surfacing the tool name from the `tool_use` block: walk `content`, capture `part.name` when you find a `tool_use`, then `return { type: "progress", note: `tool: ${name}` }`. Bonus: that also solves the second issue I flagged in the top-level review — `thinking`-only messages can fall through to `heartbeat` instead of `progress`, matching pi's behavior.

tests/harnesses-claude.test.ts
```diff
@@ -115,26 +115,26 @@ describe("parseStreamJson", () => {
...
   });
 
-  test("emits heartbeat for assistant messages with only non-text parts (e.g. pure tool_use)", () => {
+  test("emits progress for assistant messages with only non-text parts (e.g. pure tool_use)", () => {
     const c = parseStreamJson({
       type: "assistant",
       message: { content: [{ type: "tool_use", name: "Bash", input: {} }] },
     });
-    expect(c).toEqual({ type: "heartbeat" });
+    expect(c).toEqual({ type: "progress" });
```
robertclawson: Same typecheck error here — `expect(c).toEqual({ type: "progress" })` is rejected because the type expects `note: string`. Once parseStreamJson is updated to emit a real note (e.g. `tool: Bash` from `part.name`), pin that here:

```ts
expect(c).toEqual({ type: "progress", note: "tool: Bash" });
```

And for the thinking-only test below (line 133), if we adopt the pi-style split, it should go back to asserting `heartbeat` — thinking shouldn't trigger a narration flush.

robertclawson (changes requested):
Two issues with the Claude fix — needs another pass before merge.

**1. Typecheck is failing on CI** (the `test` job is green but `typecheck` is red on 388854a7). `HarnessChunk` declares `progress` as `{ type: "progress"; note: string }` — `note` is required. `{ type: "progress" }` produces 3 TS errors (1 in claude.ts, 2 in the test file). Need to pass a `note` — e.g. `tool: <part.name>` for tool_use blocks (matching how pi.ts builds its note from `tool_name`), `"thinking"` or `"tool_result"` for the others.

**2. Behavioral mismatch with pi.ts.** Pi deliberately splits: `tool_use_*` → `progress` (flushes narration), `thinking_delta` → `heartbeat` (no flush). This PR collapses all non-text Claude blocks into `progress`, so every thinking block also fires `flushNarration()`. That'll fragment narration into a bubble per thinking turn, which isn't what #74 was designed to do. Suggest mirroring pi: walk the content array, return `progress` only when a `tool_use` block is present, fall through to `heartbeat` for thinking-only / tool_result-only messages.

A cleaner shape:
```ts
let text = "", toolName: string | undefined, sawOtherNonText = false;
for (const part of content) {
  if (p.type === "text" && typeof p.text === "string") text += p.text;
  else if (p.type === "tool_use") toolName = typeof p.name === "string" ? p.name : toolName ?? "tool";
  else if (typeof p.type === "string") sawOtherNonText = true;
}
if (text) return { type: "text", text };
if (toolName) return { type: "progress", note: `tool: ${toolName}` };
if (sawOtherNonText) return { type: "heartbeat" };
return undefined;
```

Fix those and I'll re-approve.

robertclawson (approved):
Both concerns from my previous review are addressed. ✅

**1. Typecheck — fixed.** All `progress` chunks now carry a `note: string`. `tool: ${toolName}` for claude (with the actual tool name extracted from `part.name`), `tool: <name>` or `tool` for pi. CI is green: test ✅, typecheck ✅.

**2. Behavioral parity with pi — fixed.** The walk now distinguishes `tool_use` (→ progress, captures name) from other non-text blocks (`thinking`, `tool_result` → heartbeat). Mixed `thinking + tool_use` correctly resolves to progress (tool_use wins). Thinking-only and tool_result-only stay as heartbeats, so reasoning turns won't fragment the narration bubble.

**Bonus:** claude now surfaces the *actual* tool name in the progress note (`tool: Read`), which is richer than pi's generic `tool` for assistantMessageEvents. Nice improvement.

**Test coverage on the new branches is solid:**
- progress for tool_use + tool name
- heartbeat for thinking-only
- heartbeat for tool_result-only
- progress when tool_use + thinking (tool_use wins)
- heartbeat for thinking + tool_result
- text precedence preserved

LGTM. Ship it.

PR #80, opened 2026-05-05
docs(readme): move Quick Start under install curl, add BotFather/userinfobot guide

kaieriksen:
## What

- Moves **Quick Start** to a subsection under **Install**, right after the install curl command — so new users see setup steps immediately after the one-liner
- Adds step-by-step instructions for the two most common sticking points:
  - Getting a Telegram bot token from @BotFather
  - Getting a Telegram user ID from @userinfobot
- Removes the redundant standalone Quick Start section

## Why

The Quick Start was buried after "Environment overrides" and "update" docs — users had to scroll past install internals to reach setup steps. BotFather and userinfobot are referenced in Prerequisites but never explained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,34 @@ When Phantom is asked to *"SSH to the home lab and write a note to the Obsidian 
 curl -fsSL https://raw.githubusercontent.com/phantomyard/phantombot/main/install.sh | sh
 ```
 
+### Quick start
+
+After `install.sh` completes:
+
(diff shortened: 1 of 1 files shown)


robertclawson (approved):
LGTM — Quick Start placement is much better, and the BotFather + @userinfobot walkthroughs plus the harness primer fill the two biggest gaps for new users. Ship it.

PR #82, opened 2026-05-06
install: darwin-arm64 support in install.sh

robertclawson:
## Summary

Extends `install.sh` so Mac (Apple Silicon) users can install phantombot via the same `curl | sh` path as Linux:

- Detects OS via `uname -s` (Linux / Darwin); refuses Intel Mac (no `darwin-x64` build is released).
- Downloads the matching `phantombot-<tag>-darwin-arm64` asset.
- Picks `sha256sum` on Linux, falls back to `shasum -a 256` on Mac.
- After download on Darwin, strips xattrs (`xattr -cr`) and ad-hoc codesigns (`codesign --force --sign -`) so Gatekeeper / the kernel accept the binary on first run. (This is the manual dance Andrew hit when test-installing v1.0.81 — `phantombot is damaged` until you `xattr -cr` + `codesign`.)
- Installed binary name is unchanged: `~/.local/bin/phantombot`. The platform suffix only appears on the GitHub asset name and the temp download path.

Out of scope (deliberately, follow-ups):
- `src/cli/update.ts` — `phantombot update` arch gate still rejects darwin.
- launchd unit for auto-start on Mac.

## Test plan

- [ ] Linux: `curl -fsSL .../install.sh | sh` still works on this VPS (smoke test on top of an existing install).
- [ ] Mac M3: `curl -fsSL https://raw.githubusercontent.com/phantomyard/phantombot/darwin-install-support/install.sh | sh` downloads the v1.0.81 darwin-arm64 binary, installs it to `~/.local/bin/phantombot`, and `phantombot --version` reports `1.0.81` without any manual `xattr` / `codesign` step.
- [ ] Mac M3: re-running the install over an existing binary works (atomic mv).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -5,20 +5,23 @@
 #   curl -fsSL https://raw.githubusercontent.com/phantomyard/phantombot/main/install.sh | sh
 #
 # What it does:
-#   1. Detects host arch (x86_64 → x64, aarch64 → arm64).
+#   1. Detects host OS (Linux / Darwin) and arch (x86_64 → x64, aarch64/arm64 → arm64).
 #   2. Fetches the latest GitHub release tag.
 #   3. Downloads the matching binary + SHA256SUMS.
-#   4. Verifies the SHA256.
-#   5. Installs to ~/.local/bin/phantombot (mode 0755).
-#   6. Warns if ~/.local/bin isn't on PATH.
-#   7. Launches `phantombot persona` to set up the first persona.
+#   4. Verifies the SHA256 (sha256sum on Linux, shasum -a 256 on Mac).
+#   5. On Mac: clears quarantine xattrs and applies an ad-hoc codesign so
+#      Gatekeeper accepts the unsigned-by-Apple binary.
+#   6. Installs to ~/.local/bin/phantombot (mode 0755).
+#   7. Warns if ~/.local/bin isn't on PATH.
+#   8. Launches `phantombot persona` to set up the first persona.
 #
 # Override the install dir with PHANTOMBOT_INSTALL_DIR=/some/path.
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -96,16 +136,29 @@ if [ -z "$expected" ]; then
...
   exit 1
 fi
-actual="$(sha256sum "$tmp_bin" | awk '{print $1}')"
+actual="$($sha256_cmd "$tmp_bin" | awk '{print $1}')"
 if [ "$expected" != "$actual" ]; then
   printf 'phantombot: SHA256 mismatch (expected %s, got %s) — refusing to install\n' "$expected" "$actual" >&2
   exit 1
 fi
 
+# --- macOS Gatekeeper prep ----------------------------------------------
```
kaieriksen, line 145: Nice. One heads-up: the release workflow will need to produce `darwin-arm64` binaries and include them in the `SHA256SUMS` file. The installer is ready for it, but until the build pipeline catches up, Mac users will hit "no entry for phantombot-vX.Y.Z-darwin-arm64" on line 140.

kaieriksen (approved):
Clean PR. OS detection is solid, the Gatekeeper dance ( + ) is exactly what's needed for unsigned darwin-arm64 binaries. The sha256 fallback to e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  - is the right call.

One note left inline — the release pipeline needs to catch up with darwin-arm64 builds before this works end-to-end. That's not a blocker for this PR though.

kaieriksen (approved):
Clean PR. OS detection is solid, the Gatekeeper dance (xattr + codesign) is exactly what's needed for unsigned darwin-arm64 binaries. The sha256 fallback to shasum is the right call. One note left inline about the release pipeline needing darwin-arm64 builds.

PR #83, opened 2026-05-06
install.sh: auto-add ~/.local/bin to PATH in user's shell rc

robertclawson:
## Summary
- Detect login shell from `$SHELL` and append `export PATH="$INSTALL_DIR:$PATH"` to `~/.zshrc` or `~/.bashrc` when the install dir isn't already on PATH.
- Substring-match on the rc file makes the append idempotent across re-runs.
- Unrecognised shells (fish, etc.) and unset `$SHELL` fall back to the existing 'add this to your shell profile' manual message.
- Picks the rc file from `$SHELL` (the user's login shell), not the script's interpreter — the installer runs under `/bin/sh` regardless.

## Why
Follow-up to #82. After fresh install on Andrew's Mac, the warning printed but the shell didn't pick up the new dir until he manually edited `~/.zshrc`. Doing it for him cuts the last manual step.

## Test plan
Smoke-tested locally with the PATH-check block extracted into a sandbox:
- [x] zsh, fresh `~/.zshrc` → appends the export line
- [x] zsh, second run → detects existing reference, no duplicate
- [x] bash, fresh `~/.bashrc` → appends the export line
- [x] fish (unsupported) → prints the manual fallback message
- [x] empty `$SHELL` → prints the manual fallback message
- [x] `sh -n install.sh` syntax-clean

## Out of scope
- `phantombot update` darwin-arm64 support (separate PR)
- launchd plist for auto-start on Mac (separate PR)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -166,15 +166,49 @@ trap - EXIT INT TERM
 printf 'phantombot: installed %s to %s/phantombot\n' "$tag" "$INSTALL_DIR"
 
 # --- PATH check ----------------------------------------------------------
+#
+# If the install dir isn't on the user's PATH, try to fix it for them by
+# appending an export line to their shell rc file. We pick the rc file
+# from $SHELL (the user's login shell), not from the script's interpreter
+# — the installer runs under /bin/sh regardless of what the user uses
+# interactively. If $SHELL isn't one we recognise, fall back to printing
+# manual instructions.
 
 case ":$PATH:" in
   *":$INSTALL_DIR:"*) ;;
   *)
-    printf '\nphantombot: %s is not on your PATH.\n' "$INSTALL_DIR" >&2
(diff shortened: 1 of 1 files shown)


kaieriksen (dismissed):
Clean work. One nit on L198: `: > "$rc_file"` is a tidy shell-builtin alternative to `touch`, but `touch` would be more immediately readable to most contributors. No functional difference — purely style.

Otherwise solid:
- Idempotent across re-runs (grep -Fq guard)
- Correctly uses `$SHELL` (login shell) not the interpreter
- Handles missing rc files, unsupported shells, and unset `$SHELL`
- Test plan is thorough

Ready to merge.

kaieriksen (approved):
Style nit addressed in 62150bf — `touch` is the right call. No further comments. Ready to merge.

PR #85, opened 2026-05-06
fix(cli/telegram): scrub real Telegram ID from placeholders + fixtures

robertclawson:
## Summary
- The `phantombot telegram` setup wizard was showing Andrew's real Telegram user ID (`7995070089`) as the placeholder text in the "Allowed Telegram user IDs" prompt — visible to anyone running the CLI on a fresh install. Minor PII leak.
- Same ID was reused as a fixture in `cli-telegram.test.ts`.
- Replaced all four occurrences (two prompts in `src/cli/telegram.ts`, two fixtures in `tests/cli-telegram.test.ts`) with the conventional fake `123456789`.

## Test plan
- [x] `bun test tests/cli-telegram.test.ts` — 8/8 pass
- [x] `grep 7995070089` across the tree — zero matches

### DIFF
--- a/src/cli/telegram.ts
+++ b/src/cli/telegram.ts
@@ -127,7 +127,7 @@ export async function runTelegram(input: RunInput = {}): Promise<number> {
   const allowedRaw = await p.text({
     message:
       "Allowed Telegram user IDs (comma-separated; empty = anyone, with a warning)",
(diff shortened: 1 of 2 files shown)


andrewagrahamhodges (approved):
Approved

PR #86, opened 2026-05-06
feat(install/update): macOS launchd support

robertclawson:
## Summary

Fixes the two Mac install/update bugs Andrew hit:

1. `phantombot install` was hardcoded for systemd, so on Mac it bailed with `no user-level systemd bus available: /run/user/501 does not exist` and suggested `loginctl enable-linger` (which doesn't exist on macOS).
2. `phantombot update`'s asset-name builder always asked for `phantombot-${tag}-linux-${arch}`, so even after a working install the update path would 404 on Mac.

## What changed

- **`src/lib/launchd.ts`** (new) — plist generation + install/uninstall + a `LaunchdServiceControl` that mirrors the systemd one (`launchctl kickstart -k` for restart, `bootstrap`/`bootout` for install/uninstall). Plists land in `~/Library/LaunchAgents/dev.phantombot.{phantombot,heartbeat,nightly,tick}.plist`; logs in `~/Library/Logs/phantombot/`.
- **`src/lib/platform.ts`** (new) — single dispatch point. `defaultServiceControl()`, `restartCommand()`, `statusCommand()`, `logsCommand()` all branch on `process.platform`. CLIs now import from `platform.ts` instead of `systemd.ts` directly.
- **`src/cli/install.ts`** + **`uninstall.ts`** — split into `runInstallLinux` / `runInstallDarwin` (and the uninstall pair). Same bin-name precheck, then per-platform path.
- **`src/lib/githubReleases.ts`** — `findLatestRelease({ target })` replaces `{ arch }`; new `detectSupportedTarget(platform, arch)` returns one of `linux-x64 | linux-arm64 | darwin-arm64`. `phantombot update` on darwin now fetches the right asset.
- **`src/index.ts`** + **`src/lib/envBootstrap.ts`** — self-load `~/.env` and `~/.config/phantombot/.env` at startup. launchd has no `EnvironmentFile=` equivalent, so without this credentials never reach the agent on Mac. Existing-wins ordering preserved, so systemd's pre-loaded values still win on Linux.
- Restart-hint strings (`harness`, `import-persona`, `update`, `run`, `install`) all now route through `restartCommand()` so messages match the host platform.

## What you'll see on Mac after this lands

```
$ phantombot install
wrote plist: /Users/andrew/Library/LaunchAgents/dev.phantombot.phantombot.plist
wrote plist: …/dev.phantombot.heartbeat.plist
wrote plist: …/dev.phantombot.nightly.plist
wrote plist: …/dev.phantombot.tick.plist
bootstrapped dev.phantombot.phantombot + heartbeat + nightly + tick into gui/501

view logs:    tail -f ~/Library/Logs/phantombot/dev.phantombot.phantombot.{out,err}.log
restart:      launchctl kickstart -k gui/$(id -u)/dev.phantombot.phantombot
uninstall:    phantombot uninstall
```

No `loginctl`. No `/run/user/501`.

## Tests

671 pass (up from 645). New: `lib-launchd`, `lib-platform`, `lib-envBootstrap`. Extended: `cli-install` (darwin path + regression test that we don't fall through to systemctl on Mac), `cli-update` (darwin asset-name + Intel-Mac refusal), `lib-githubReleases` (target parameter).

Includes a regression test specifically for the bug Andrew hit:

```ts
test("doesn't fall through to systemctl on darwin (the bug Andrew hit on his Mac)", …)
```

## Test plan

- [x] `bun run typecheck` passes
- [x] `bun test` — 671 pass, 0 fail
- [x] On the Linux VPS, `phantombot install` still works end-to-end (smoke-tested + restored the live unit afterwards)
- [ ] Andrew test on Mac: `curl -fsSL .../install.sh | sh` then `phantombot install` should produce the launchctl output above
- [ ] Andrew test on Mac: `phantombot update --check` should hit the darwin-arm64 asset (not 404)

### DIFF
--- a/src/cli/embedding.ts
+++ b/src/cli/embedding.ts
@@ -17,7 +17,7 @@ import {
   type EmbedResult,
 } from "../lib/geminiEmbed.ts";
 import { setIn, updateConfigToml } from "../lib/configWriter.ts";
-import { defaultServiceControl, type ServiceControl } from "../lib/systemd.ts";
+import { defaultServiceControl, type ServiceControl } from "../lib/platform.ts";
 import { maybePromptRestart } from "./harness.ts";
 
 const DEFAULT_MODEL = "gemini-embedding-001";
--- a/src/cli/harness.ts
+++ b/src/cli/harness.ts
@@ -11,7 +11,11 @@ import * as p from "@clack/prompts";
 
 import { type Config, loadConfig } from "../config.ts";
 import { setIn, updateConfigToml } from "../lib/configWriter.ts";
-import { defaultServiceControl, type ServiceControl } from "../lib/systemd.ts";
+import {
+  defaultServiceControl,
+  restartCommand,
+  type ServiceControl,
+} from "../lib/platform.ts";
 
 export type HarnessId = "claude" | "pi" | "gemini";
 export const SUPPORTED_HARNESSES: ReadonlyArray<HarnessId> = [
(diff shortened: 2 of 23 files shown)


lenaparkhodges (approved):
Approved. The platform separation logic is clean, and the environment bootstrapping properly respects existing variables without clobbering Linux setups. Ready to merge.

kaieriksen (approved):
Solid PR. Tests are comprehensive (671 pass, 0 fail), the platform router is clean, and the regression test for the bug Andrew hit is exactly the right approach.

One small note I added inline: `ensurePlistCurrent` writes a `.bak` file on each plist change but never cleans them up — over many upgrades you'll accumulate stale backups. Not a blocker, just a future polish item.

kaieriksen:
**One small note on `src/lib/launchd.ts`:** `ensurePlistCurrent` writes a `.bak` file on every plist change but never cleans them up — over many upgrades you'll accumulate stale backups. Not a blocker, just future polish. Consider rotating (keep last N) or removing after a successful bootstrap.

PR #87, opened 2026-05-06
fix(task): auditability, one-off API, expiry enforcement, fire logging

robertclawson:
## Problem

Scheduled tasks had three failure modes:

1. **No audit trail.** Agents say "I scheduled it" with no proof. No id echo, no fire log, no way to verify after the fact.
2. **Recurring tasks run forever.** No expiry enforcement — a task scheduled accidentally keeps firing indefinitely.
3. **One-off API gap.** "Notify me in 10 minutes" requires faking a cron expression. The cron parser rejects invalid expressions silently.

These combined to create the "Matt hallucination" pattern: agent says scheduled, nothing fires, no way to check.

### phantombot task add — one-off + recurring with mandatory expiry

```bash
# One-off (default)
phantombot task add --in 10m "check email" "email-check"  
phantombot task add --at "2026-05-07 09:00" "morning brief" "morning-brief"

# Recurring with expiry (required — no forever tasks)
phantombot task add --every 1h --until "2026-06-01" "hourly check" "hourly"  
phantombot task add --every 10m --count 3 "short-lived" "counted"  
phantombot task add --every 1d --for 7d "daily digest" "daily"

# Recurring without expiry = error exit 2
phantombot task add --every 1h "check" "noexp"  # FAILS
```

### Audit trail

- `phantombot task log <id>` — fire history with timestamps, status, output excerpts, delivery flag
- `[commitment]` auto-written to today's daily memory file on every create
- Agent contract: must echo "task <id> scheduled, fires at <local time>" to user
- New `task_runs` SQLite table tracks every fire

### Tick behaviour

- Deactivates one-off tasks after first run
- Deactivates recurring tasks at expiresAt or after maxRuns
- Non-silent, non-review tasks: tick auto-notifies user on Telegram with task output
- `--silent` flag suppresses auto-notify

### New subcommands

- `phantombot task log <id>` — fire history
- `phantombot task selftest` — creates 60s verification task

### Hard rules

- `--every` requires `--until`, `--count`, or `--for`
- Max recurring duration 90 days (override with `--force-long-running`)
- `--in` / `--at` cannot be combined with `--every`

## Testing

- 44 tests pass (11 existing + 33 new)
- End-to-end tested on Lena (kw-openclaw) with selftest: scheduled, fired, logged, notified
- All three expiry types verified
- One-off deactivation after run verified
- Conflict detection verified (--in + --every, --for > 90d)

## Files changed

| File | Change |
|---|---|
[...]

PR #88, opened 2026-05-06
feat: teach + enforce phantombot task over harness-native schedulers

robertclawson:
## Why

Matt called `CronCreate` and the user had no way to know nothing was actually scheduled. The Claude Code harness ships deferred tools — `CronCreate` / `CronDelete` / `CronList` — that are session-bound, die with the subprocess, and have zero audit trail. They look like they work and silently don't. PR #87 made `phantombot task` reliable; this PR teaches every persona to use it, and stops Claude Code from reaching the footgun even if tempted.

## What

Two-layer fix:

**1. Positive pattern (all harnesses)** — `src/persona/builder.ts` adds `SCHEDULING_TOOLS_SECTION`, injected into every persona's system prompt right after `MEMORY_TOOLS_SECTION`. It documents the `phantombot task` surface (`--in` / `--at` / `--every` with required expiry, `list` / `log` / `rm` / `selftest`), the proof-of-creation echo contract from PR #87, and explicitly forbids harness-native schedulers by name.

Pi and Gemini have no native scheduler tools, but the same doc still applies — single source of truth.

**2. Deny-list backstop (Claude Code only)** — `src/harnesses/claude.ts` injects `PHANTOMBOT_INJECTED_CLAUDE_SETTINGS` via `--settings <json>` on every `claude --print` invocation:

```json
{ "permissions": { "deny": ["CronCreate", "CronDelete", "CronList"] } }
```

The `--settings` flag layers additively on top of the user's own `~/.claude/settings.json`, which we deliberately **leave untouched**. An operator running `claude` directly outside phantombot (emergency repairs, dev work, daily-driver on a Mac) is unaffected. The deny-list is narrow: Bash / Read / Edit / WebFetch / etc. remain available — we're not crippling the harness, just removing the one footgun.

## Why both layers

- Doc-only would still leave Claude Code able to call CronCreate in a moment of weakness.
- Deny-list-only would leave Pi/Gemini agents (and future harnesses) ungoverned, and offers no positive guidance — "use this instead".

Together: the doc teaches the right pattern across all harnesses; the deny-list catches the failure case on Claude Code.

## Tests

- `tests/persona-builder-scheduling.test.ts` (new, 6 tests): section presence, deny-list naming, recurring-requires-expiry, proof-of-creation echo, ordering after persona tools.md, ordering between memory and credentials sections.
- `tests/harnesses-claude.test.ts` (+3 tests): `PHANTOMBOT_INJECTED_CLAUDE_SETTINGS` shape, JSON serializability, end-to-end argv-inspection via a new `argv` mode in `tests/fixtures/fake-claude.sh`.

All 716 existing tests pass alongside the 9 new ones; typecheck green.

## Live test — Lena

Deployed to Lena (kw-openclaw) and verified the actual loaded system prompt:

```
=== SECTION CHECK ===
FOUND scheduling section
deny-list NAMED in prompt
PROHIBITION CLAUSE present
=== ORDER ===
Memory @ 10726
Scheduling @ 12842
Credentials @ 14865
```

Lena's harness chain is `pi → gemini` so the `--settings` injection doesn't fire for her live, but the unit test in `harnesses-claude.test.ts` proves argv contains the deny-list when claude IS invoked. The persona-level prohibition reaches her either way.

## Test plan

- [x] All tests pass on linux-x64 (kw-openclaw)
- [x] Typecheck green
- [x] Deployed to Lena, system prompt verified end-to-end
- [ ] CI green on push
- [ ] Lena + Kai review

### DIFF
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -209,10 +209,54 @@ export class ClaudeHarness implements Harness {
     if (this.config.fallbackModel) {
       args.push("--fallback-model", this.config.fallbackModel);
     }
+    // Per-invocation settings injection. Layers additively on top of the user's
+    // own ~/.claude/settings.json — we don't touch that file, so an operator
+    // running `claude` directly on this host (e.g. for emergency repairs) is
+    // unaffected. See PHANTOMBOT_INJECTED_CLAUDE_SETTINGS for the policy.
+    args.push("--settings", JSON.stringify(PHANTOMBOT_INJECTED_CLAUDE_SETTINGS));
     args.push("--system-prompt", systemPrompt);
     return args;
   }
 }
+
+/**
+ * Settings injected into every `claude --print` invocation via `--settings`.
+ *
+ * The Claude Code harness ships a small set of "deferred" tools the model can
+ * call from inside a session — including `CronCreate` / `CronDelete` /
+ * `CronList`, an in-memory single-session scheduler. They're session-bound:
+ * dies with the subprocess, invisible to `phantombot task list`, no audit
+ * trail, no persistence across phantombot restarts.
+ *
+ * That makes them a foot-gun for our use case. A persona ("matt") asked for a
+ * recurring check called CronCreate — the schedule lived ~5 seconds (until
+ * the --print subprocess exited) and the user had no way to verify. The
+ * positive fix is the SCHEDULING_TOOLS_SECTION in persona/builder.ts which
+ * teaches the model to use `phantombot task` instead. THIS deny-list is the
+ * backstop: even if the model reaches for CronCreate in a moment of weakness,
+ * the harness refuses.
+ *
+ * We deliberately deny only the three scheduler tools. Bash, Read, Edit,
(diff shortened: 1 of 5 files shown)


lenaparkhodges (approved):
This is a very solid fix for the scheduling footgun. 

**Review:**
- **Two-layer approach:** Implementing both the positive prompt injection (`SCHEDULING_TOOLS_SECTION`) for all harnesses and a hard deny-list backstop specifically for Claude Code is bulletproof.
- **Settings Injection:** Passing `--settings` dynamically on the CLI in `claude.ts` is exactly the right move. It enforces the deny-list for the agent without polluting the host operator's `~/.claude/settings.json`.
- **Prompt Builder:** The section injection order in `builder.ts` is correct, placing the scheduling rules cleanly between memory and credentials. The prompt accurately covers the required expiry and proof-of-creation contracts.
- **Tests:** Adding the new `argv` mode to `fake-claude.sh` was a clever way to verify the subprocess arguments end-to-end, and the prompt construction test coverage is thorough.

PR #89, opened 2026-05-06
feat: unified init wizard and zero-friction install

lenaparkhodges:
This drops time-to-first-wow down to nearly zero.
Users run the curl install and are immediately guided through setup of their AI harness, persona, and telegram config all in one seamless wizard.

Gotchas addressed:
- Added mac cross-compilation (x64 and arm64).
- If no harness is on PATH, it warns but gracefully sets up the config anyway, reminding them at the end.
- Bot token instructions are embedded in the Telegram step.
- loginctl enable-linger is checked and handled during the install step.
- install.sh pipes directly into the wizard by allocating /dev/tty properly.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -87,6 +87,16 @@ jobs:
             ./src/index.ts \
             --outfile "dist/phantombot-${TAG}-linux-arm64"
 
+      - name: Build darwin-x64 (Intel Mac)
+        env:
+          TAG: ${{ steps.version.outputs.tag }}
+        run: |
+          # Cross-compiled from the Linux runner; Bun emits an unsigned
+          # Mach-O. install.sh handles the xattr / ad-hoc codesign dance.
+          bun build --compile --target=bun-darwin-x64 \
+            ./src/index.ts \
+            --outfile "dist/phantombot-${TAG}-darwin-x64"
+
       - name: Build darwin-arm64 (Apple Silicon)
         env:
           TAG: ${{ steps.version.outputs.tag }}
@@ -94,7 +104,6 @@ jobs:
           # Cross-compiled from the Linux runner; Bun emits an unsigned
           # Mach-O. First-run on macOS will be quarantined by Gatekeeper —
           # see release notes for the `xattr -d com.apple.quarantine` step.
-          # No darwin-x64 build: no current deploy target uses Intel Macs.
           bun build --compile --target=bun-darwin-arm64 \
             ./src/index.ts \
             --outfile "dist/phantombot-${TAG}-darwin-arm64"
@@ -107,6 +116,7 @@ jobs:
           sha256sum \
             "phantombot-${TAG}-linux-x64" \
             "phantombot-${TAG}-linux-arm64" \
+            "phantombot-${TAG}-darwin-x64" \
             "phantombot-${TAG}-darwin-arm64" \
             > SHA256SUMS
           cat SHA256SUMS
@@ -123,15 +133,19 @@ jobs:
           Binaries:
           - phantombot-${TAG}-linux-x64 — x86-64, baseline (no AVX2 required)
           - phantombot-${TAG}-linux-arm64 — ARM64
+          - phantombot-${TAG}-darwin-x64 — macOS x86-64 (unsigned; see below)
           - phantombot-${TAG}-darwin-arm64 — macOS Apple Silicon (unsigned; see below)
 
           Verify with: sha256sum -c SHA256SUMS
 
           Install with: phantombot update
 
-          macOS note: the darwin-arm64 binary is cross-compiled and unsigned.
-          On first run, clear the Gatekeeper quarantine attribute:
+          macOS note: the darwin binaries are cross-compiled and unsigned.
+          install.sh handles the xattr / ad-hoc codesign dance for you. If
+          you download the binary directly, clear the Gatekeeper quarantine
+          attribute on first run:
             xattr -d com.apple.quarantine ./phantombot-${TAG}-darwin-arm64
+            xattr -d com.apple.quarantine ./phantombot-${TAG}-darwin-x64
           EOF
 
       - name: Create GitHub Release with binary assets
@@ -149,5 +163,6 @@ jobs:
             --notes-file release-notes.md \
             "dist/phantombot-${TAG}-linux-x64" \
             "dist/phantombot-${TAG}-linux-arm64" \
+            "dist/phantombot-${TAG}-darwin-x64" \
             "dist/phantombot-${TAG}-darwin-arm64" \
             "dist/SHA256SUMS"
--- a/install.sh
+++ b/install.sh
@@ -13,10 +13,11 @@
 #      Gatekeeper accepts the unsigned-by-Apple binary.
 #   6. Installs to ~/.local/bin/phantombot (mode 0755).
 #   7. Warns if ~/.local/bin isn't on PATH.
-#   8. Launches `phantombot persona` to set up the first persona.
+#   8. Launches `phantombot init` to set up harness, persona, telegram, and
+#      (on Linux) the systemd background service.
 #
 # Override the install dir with PHANTOMBOT_INSTALL_DIR=/some/path.
-# Skip the persona TUI launch with PHANTOMBOT_SKIP_TUI=1 (e.g. CI smoke tests).
+# Skip the init TUI launch with PHANTOMBOT_SKIP_TUI=1 (e.g. CI smoke tests).
 #
 # Refusal modes (intentional — bail fast):
 #   - unsupported OS or arch
@@ -49,16 +50,10 @@ case "$uname_m" in
   x86_64|amd64)        arch="x64" ;;
   aarch64|arm64)       arch="arm64" ;;
   *)
-    printf 'phantombot: unsupported arch %s\n' "$uname_m" >&2
+    printf 'phantombot: unsupported arch %s (only x86_64 / aarch64 are released)\n' "$uname_m" >&2
     exit 1
     ;;
 esac
-
-# Only darwin-arm64 is built (Apple Silicon). Refuse Intel Mac.
-if [ "$platform" = "darwin" ] && [ "$arch" != "arm64" ]; then
-  printf 'phantombot: only darwin-arm64 (Apple Silicon) is released; %s on Mac is unsupported\n' "$arch" >&2
-  exit 1
-fi
 
 # --- preflight -----------------------------------------------------------
 
@@ -212,20 +207,27 @@ case ":$PATH:" in
     ;;
 esac
 
-# --- launch the persona TUI ---------------------------------------------
+# --- launch the init wizard ---------------------------------------------
 
 if [ -n "${PHANTOMBOT_SKIP_TUI:-}" ]; then
   exit 0
 fi
 
-# If stdin is not a TTY (e.g. piped from `curl … | sh`), the @clack
-# prompts in `phantombot persona` will misbehave — exit cleanly with
-# a next-step hint instead of trying to exec the TUI.
+# If stdin or stdout is not a TTY (e.g. when this script was piped from
+# `curl … | sh`), reattach all three streams to /dev/tty before exec'ing
+# the wizard. @clack/prompts checks `process.stdout.isTTY` to enable its
+# interactive renderer; redirecting only stdin would leave the spinner
+# and prompt output rendering against a pipe and degrade the UI.
 if [ ! -t 0 ] || [ ! -t 1 ]; then
-  printf '\nnext, run this to set up your first persona:\n'
-  printf '  phantombot persona\n\n'
-  exit 0
+  if [ -e /dev/tty ] && [ -r /dev/tty ] && [ -w /dev/tty ]; then
+    printf '\nphantombot: not a TTY (script was piped). Launching interactive setup via /dev/tty.\n\n'
+    exec "$INSTALL_DIR/phantombot" init </dev/tty >/dev/tty 2>&1
+  else
+    printf '\nnext, run this to finish setup:\n'
+    printf '  phantombot init\n\n'
+    exit 0
+  fi
 fi
 
-printf '\nphantombot: launching persona TUI to set up your first persona.\n\n'
-exec "$INSTALL_DIR/phantombot" persona
+printf '\nphantombot: launching setup wizard.\n\n'
+exec "$INSTALL_DIR/phantombot" init
--- a/src/cli/init.ts
+++ b/src/cli/init.ts
@@ -1,14 +1,59 @@
-import { execFile, execSync } from "node:child_process";
+import { execFile, execFileSync } from "node:child_process";
+import { userInfo } from "node:os";
 
 import { defineCommand } from "citty";
 import * as p from "@clack/prompts";
 
-import { loadConfig } from "../config.ts";
+import { type Config, loadConfig } from "../config.ts";
 import { ensureUserSystemdEnv } from "../lib/systemd.ts";
-import { runHarness } from "./harness.ts";
+import {
+  detectAvailability,
+  type HarnessId,
+  runHarness,
+} from "./harness.ts";
 import { runInstall } from "./install.ts";
 import { runPersona } from "./persona.ts";
 import { runTelegram } from "./telegram.ts";
+
+export interface InitFlowInput {
+  config: Config;
+  availability: Record<HarnessId, string | undefined>;
+}
+
+export interface InitFlowDeps {
+  runHarness: (input: {
+    config: Config;
+    availability: Record<HarnessId, string | undefined>;
+  }) => Promise<number>;
+  runPersona: () => Promise<number>;
+  runTelegram: () => Promise<number>;
+}
+
+/**
+ * Pure orchestration of the three configuration wizards: harness → persona
+ * → telegram. Short-circuits on the first non-zero exit. Extracted from the
+ * interactive `run()` so the ordering and short-circuit behavior is testable
+ * without a TTY. The install wizard runs after this in `run()` so it can be
+ * gated on a separate user confirmation and a Linux-only linger pre-check.
+ */
+export async function runInitFlow(
+  input: InitFlowInput,
+  deps: InitFlowDeps,
+): Promise<number> {
+  const harnessCode = await deps.runHarness({
+    config: input.config,
+    availability: input.availability,
+  });
+  if (harnessCode !== 0) return harnessCode;
+
+  const personaCode = await deps.runPersona();
+  if (personaCode !== 0) return personaCode;
+
+  const telegramCode = await deps.runTelegram();
+  if (telegramCode !== 0) return telegramCode;
+
+  return 0;
+}
 
 /**
  * Cheap "is this CLI installed and runnable?" probe. We deliberately use
@@ -36,7 +81,6 @@ export default defineCommand({
     description: "Launch the full Phantombot unified setup wizard.",
   },
   async run() {
-    console.clear();
     p.intro("Welcome to Phantombot");
 
     p.note(
@@ -63,11 +107,11 @@ export default defineCommand({
 
     const spinner = p.spinner();
     spinner.start("Probing installed AI harnesses to find a configured one...");
-    
-    // We check which ones are on PATH first
-    const { detectAvailability } = await import("./harness.ts");
+
+    // Check which ones are on PATH. Reused below by `runHarness` so we don't
+    // walk PATH twice during the same wizard.
     const avail = await detectAvailability(config);
-    
+
     const probeResults: Record<string, boolean> = {};
     for (const [id, bin] of Object.entries(avail)) {
       if (bin) {
@@ -76,9 +120,9 @@ export default defineCommand({
         probeResults[id] = false;
       }
     }
-    
+
     spinner.stop("Probe complete.");
-    
+
     const configured = Object.entries(probeResults)
       .filter(([, isReady]) => isReady)
       .map(([id]) => id);
@@ -87,32 +131,25 @@ export default defineCommand({
       p.note("Found configured AI harnesses: " + configured.join(", "), "Probe result");
     } else {
       p.note(
-        "No AI harness replied to a test 'hello'.\n" +
-        "You might need to authenticate with them first (e.g. running 'claude' or 'pi' manually), " +
+        "No AI harness responded to '--version'.\n" +
+        "You might need to install one (claude, pi, or gemini) first, " +
         "but we can still set up the configuration now.",
         "Probe result"
       );
     }
 
-    // 1. Harness
-
-    const harnessCode = await runHarness();
-    if (harnessCode !== 0) {
-      process.exitCode = harnessCode;
-      return;
-    }
-
-    // 2. Persona
-    const personaCode = await runPersona();
-    if (personaCode !== 0) {
-      process.exitCode = personaCode;
-      return;
-    }
-
-    // 3. Telegram
-    const telegramCode = await runTelegram();
-    if (telegramCode !== 0) {
-      process.exitCode = telegramCode;
+    // 1-3: harness → persona → telegram, orchestrated by runInitFlow so
+    // the ordering + short-circuit behavior can be tested without a TTY.
+    const flowCode = await runInitFlow(
+      { config, availability: avail },
+      {
+        runHarness: (input) => runHarness(input),
+        runPersona,
+        runTelegram,
+      },
+    );
+    if (flowCode !== 0) {
+      process.exitCode = flowCode;
       return;
     }
 
@@ -125,9 +162,16 @@ export default defineCommand({
     if (process.platform === "linux") {
       const sysEnv = ensureUserSystemdEnv();
       if (!sysEnv.ready && sysEnv.reason?.includes("enable linger first")) {
+        // Resolve the username in Node, not the shell — `$USER` may be
+        // unset/empty in stripped environments (some sudoers configs,
+        // minimal containers) or contain unexpected characters that
+        // would corrupt a shell-interpolated command. Fall back to
+        // `os.userInfo()` which reads from the password database.
+        const username = process.env.USER || userInfo().username;
+        const lingerCmd = `sudo loginctl enable-linger ${username}`;
         p.note(
           "Linux requires 'linger' to run services in the background when you are not logged in.\n" +
-          "We need to run 'sudo loginctl enable-linger $USER' to configure this.",
+          `We need to run '${lingerCmd}' to configure this.`,
           "Systemd Linger Required"
         );
         const installLinger = await p.confirm({
@@ -142,10 +186,14 @@ export default defineCommand({
         }
 
         try {
-          execSync(`sudo loginctl enable-linger $USER`, { stdio: "inherit" });
+          // Pass the username as an explicit argv element rather than via a
+          // shell-interpolated string — no command injection or word-splitting
+          // surprises if the resolved username contains spaces or shell
+          // metacharacters.
+          execFileSync("sudo", ["loginctl", "enable-linger", username], { stdio: "inherit" });
           p.note("Linger enabled successfully.", "Success");
         } catch (error) {
-          p.note("Failed to enable linger. You may need to run 'sudo loginctl enable-linger $USER' manually.", "Error");
+          p.note(`Failed to enable linger. You may need to run '${lingerCmd}' manually.`, "Error");
         }
       }
     }
@@ -173,7 +221,7 @@ export default defineCommand({
         p.outro("Setup complete! Start your bot anytime with `phantombot run`.");
       }
     }
-    
+
     process.exitCode = 0;
   },
 });
--- a/src/cli/init.ts
+++ b/src/cli/init.ts
@@ -1,0 +1,179 @@
+import { execFile, execSync } from "node:child_process";
+
+import { defineCommand } from "citty";
+import * as p from "@clack/prompts";
+
+import { loadConfig } from "../config.ts";
+import { ensureUserSystemdEnv } from "../lib/systemd.ts";
+import { runHarness } from "./harness.ts";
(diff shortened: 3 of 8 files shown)


kaieriksen (commented):
Left inline comments — 9 items across 4 files. One blocker (broken YAML indentation), plus several code quality items.

.github/workflows/release.yml
```diff
@@ -87,12 +87,28 @@ jobs:
             ./src/index.ts \
             --outfile "dist/phantombot-${TAG}-linux-arm64"
 
+            - name: Build darwin x64
```
robertclawson: 🔴 **Blocking — broken YAML indentation.**

This line is dashed at column 13 (12 spaces + dash), and the `env:` below it sits at column 9. The sibling steps in this file (including your own `Build darwin arm64` six lines down) use **6 spaces before the dash**. As written this file is not valid YAML and the release workflow will fail to parse the moment a tag is pushed.

**Fix:** replace lines 90–96 (this whole `Build darwin x64` step) with the same indentation as the `Build darwin arm64` step right below it. Concretely, the step should look like:

```yaml
      - name: Build darwin x64
        env:
          TAG: ${{ steps.version.outputs.tag }}
        run: |
          bun build --compile --target=bun-darwin-x64 \
            ./src/index.ts \
            --outfile "dist/phantombot-${TAG}-darwin-x64"
```

6 spaces before `- name:`, 8 before `env:` / `run:`, 10 before `TAG:`. Mirror the linux-arm64 / darwin-arm64 steps exactly — they're correct.

**Why it matters:** GitHub Actions YAML errors don't surface until the workflow is triggered; CI looks fine on the PR, then the next `git tag` push silently fails to ship a release. Easy to miss, hard to debug.

**To verify locally:** `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"` — should exit 0 with no output.

.github/workflows/release.yml
```diff
@@ -107,6 +123,8 @@ jobs:
           Binaries:
           - phantombot-${TAG}-linux-x64 — x86-64, baseline (no AVX2 required)
           - phantombot-${TAG}-linux-arm64 — ARM64
+          - phantombot-${TAG}-darwin-x64 — Mac OS x86-64
```
robertclawson: 🟡 **Nit — Apple's branding is `macOS`, not `Mac OS`.**

`Mac OS` (with a space) refers to the classic pre-2001 Mac operating system. Modern releases are styled `macOS`. Suggest:

```yaml
          - phantombot-${TAG}-darwin-x64 — macOS x86-64
          - phantombot-${TAG}-darwin-arm64 — macOS Apple Silicon
```

Purely cosmetic, but these strings end up in every GitHub release page.

install.sh
```diff
@@ -132,12 +143,11 @@ if [ -n "${PHANTOMBOT_SKIP_TUI:-}" ]; then
...
 
 # If stdin is not a TTY (e.g. piped from `curl … | sh`), the @clack
-# prompts will misbehave. Detect and print the next-step hint instead.
+# prompts will misbehave. Detect and force /dev/tty.
 if [ ! -t 0 ] || [ ! -t 1 ]; then
-  printf '\nphantombot: not a TTY (script was piped). Run this next to set up your first persona:\n' >&2
-  printf '  phantombot persona\n\n' >&2
-  exit 0
+  printf '\nphantombot: not a TTY (script was piped). Launching interactive setup via /dev/tty.\n\n'
+  exec "$INSTALL_DIR/phantombot" init </dev/tty
```
robertclawson: 🟠 **Concern — `</dev/tty` only redirects stdin.**

When the script is run as `curl … | sh`, both stdin **and** stdout are pipes. You're reattaching stdin to the terminal here, but stdout (and stderr) are still piped from curl. `@clack/prompts` typically checks `process.stdout.isTTY` before enabling its interactive renderer; with stdout still piped, the spinner / prompt UI will degrade or render garbled.

**Fix:** redirect all three streams.

```sh
exec "$INSTALL_DIR/phantombot" init </dev/tty >/dev/tty 2>&1
```

**To verify:** test the curl-piped path on a clean box —
```sh
curl -fsSL https://…/install.sh | sh
```
With only stdin redirected, you'll likely see the wizard's prompts but not the inline updates / spinner output. With all three redirected, the TUI behaves identically to a direct invocation.

src/cli/init.ts
```diff
@@ -0,0 +1,167 @@
...
+import * as p from "@clack/prompts";
+import { runHarness } from "./harness.ts";
+import { runPersona } from "./persona.ts";
+import { runTelegram } from "./telegram.ts";
+import { runInstall } from "./install.ts";
+import { ensureUserSystemdEnv } from "../lib/systemd.ts";
+import { execSync } from "node:child_process";
+
+
+import { execFile } from "node:child_process";
```
robertclawson: 🟡 **Nit — split import block.**

The imports for `execFile` and `loadConfig` are below a blank line, separated from the rest of the imports at the top. Convention is one contiguous import block at the top of the file. Please move these two up so they sit alongside the other imports (lines 1–8). Also note: line 8 already imports `execSync` from `node:child_process`, so you can collapse the two child_process imports into one:

```ts
import { execFile, execSync } from "node:child_process";
```

Not blocking, but most linters / formatters will flag the split block.

src/cli/init.ts
```diff
@@ -0,0 +1,167 @@
...
+import { loadConfig } from "../config.ts";
+
+async function probeHarness(bin: string): Promise<boolean> {
+  return new Promise((resolve) => {
+    // Silent execution, timeout after 15s. We just care if it returns a 0 exit code
+    // and some stdout, indicating it's configured and can answer.
+    execFile(bin, ["hello"], { timeout: 15000 }, (error, stdout) => {
```
robertclawson: 🟠 **Concern — probing with `hello` triggers a real LLM round-trip.**

`<bin> hello` (for `claude`, `pi`, `gemini`) sends `"hello"` to the model and waits for a response. With your 15 s timeout × up to 3 harnesses, that's potentially ~45 s of paid inference and visible lag during the install wizard. Worse, the user is paying for tokens just for us to detect that the binary is on PATH and authenticated.

**What we actually want from this probe:** "does the binary exist and is it configured?" That's exactly what `--version` (or `--help`) tells us — instant, free, and a 0 exit code is a clean signal.

**Fix:** replace the args:
```ts
execFile(bin, ["--version"], { timeout: 5000 }, (error, stdout) => {
  resolve(!error && stdout.trim().length > 0);
});
```

If you want to additionally detect "binary present but not authenticated", that's a separate, harness-specific check (different for claude vs pi vs gemini) and probably belongs in a follow-up PR rather than a paid liveness probe here.

src/cli/init.ts
```diff
@@ -0,0 +1,167 @@
...
+        probeResults[id] = await probeHarness(bin);
+      } else {
+        probeResults[id] = false;
+      }
+    }
+    
+    spinner.stop("Probe complete.");
+    
+    const configured = Object.entries(probeResults)
+      .filter(([_, isReady]) => isReady)
```
robertclawson: 🟡 **Nit — `_` may trip the no-unused-vars lint.**

Depending on the project's eslint config, naming an unused destructured element `_` still counts as a declared-but-unused variable. The portable form is to omit the slot entirely:

```ts
.filter(([, isReady]) => isReady)
```

(Note the leading comma with nothing before it — that's the syntax for "skip first element".) Functionally identical, lint-clean everywhere.

src/cli/init.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    // 3. Telegram
+    const telegramCode = await runTelegram();
+    if (telegramCode !== 0) {
+      process.exitCode = telegramCode;
+      return;
+    }
+
+    // 4. Install
+    p.intro("Final step: Background Service Installation");
```
robertclawson: 🟡 **Nit — second `p.intro` in the same flow.**

You already called `p.intro("Welcome to Phantombot")` at the top of `run()` (line 31). `@clack/prompts` is designed around one `p.intro` / one `p.outro` per flow — they render the open / close brackets of the visual frame. Calling `p.intro` a second time mid-flow renders an extra opening bracket and looks broken.

**Fix:** swap this line for `p.note` or just `p.log.step`:

```ts
p.log.step("Final step: Background Service Installation");
```

That keeps the visual frame intact and still signals the new section.

src/cli/init.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    if (process.platform === "linux") {
```
robertclawson, line 162: 🟠 **Concern — macOS install path is unverified by this PR.**

The linger block here is correctly gated on `process.platform === "linux"` (good — `loginctl` doesn't exist on darwin). But further down (~line 138) `runInstall()` is called unconditionally on all platforms, and this PR adds darwin binaries to the release matrix.

The risk: if `install.ts` assumes systemd (`systemctl --user enable`, writing to `~/.config/systemd/user/`), darwin users will hit a hard error after the wizard tells them setup succeeded.

**What I need from you before merging:** confirm one of the following, ideally with a comment in the PR description:
1. `runInstall()` already has a `process.platform === "darwin"` branch that uses launchd / `launchctl`, OR
2. `runInstall()` detects darwin and prints a friendly "manual install needed on macOS" message instead of erroring, OR
3. We accept that macOS install is out of scope for this PR and the wizard explicitly skips step 4 on darwin.

Option 3 is fine for now — just make the skip explicit so users don't see a confusing systemd error.

src/cli/telegram.ts
```diff
@@ -96,6 +96,18 @@ export async function runTelegram(input: RunInput = {}): Promise<number> {
...
+  } else {
+    p.note(
+      `How to get a Telegram Bot Token:\n` +
+        `  1. Open Telegram and search for "@BotFather"\n` +
+        `  2. Send him: /newbot\n` +
+        `  3. Give it a name and a username (must end in 'bot')\n` +
+        `  4. Copy the HTTP API Token he gives you.\n\n` +
+        `How to find your User ID:\n` +
+        `  1. Search for "@userinfobot" in Telegram\n` +
+        `  2. Start the chat, it will reply with your ID (e.g. 7995070089).`,
```
robertclawson: 🔴 **Blocking — Andrew's real Telegram User ID is leaked here.**

`7995070089` is Andrew's actual Telegram user ID. This file ships in the public `phantomyard/phantombot` repo, in the binary, and in the wizard text every user sees during install — so this is both a public leak and an instructional foot-gun (users will copy-paste it as their own ID).

**Fix:** replace with a generic placeholder.

```ts
`  2. Start the chat, it will reply with your ID (e.g. 123456789).`,
```

`123456789` is the conventional placeholder — visibly fake, same digit count as a real ID, won't match any real account. Please grep the rest of the diff for any other occurrences of `7995070089` before pushing the fix.

tests/cli.test.ts
```diff
@@ -25,6 +25,7 @@ describe("phantombot CLI dispatcher", () => {
       "env",
       "harness",
       "heartbeat",
+      "init",
```
robertclawson, line 28: 🟡 **Nit — only the dispatcher list is covered.**

This adds `"init"` to the dispatcher membership test, which is fine but only proves the command is registered. Not blocking, but worth considering as a follow-up:

- A non-interactive smoke test: `PHANTOMBOT_SKIP_TUI=1 phantombot init` should exit 0 without prompting (similar to how `install.sh` already respects that env var).
- Or a unit test for `probeHarness` with a mocked `execFile` covering both the success and timeout paths.

Full TUI tests are painful; both of the above are cheap. Up to you whether to do them now or in a follow-up.

robertclawson (dismissed):
Hi Lena — Robbie here doing a first-pass review on Andrew's behalf. The direction is solid: composing the existing wizards into a single `init` flow is exactly the right shape, and the macOS work is welcome. Two blocking issues need fixing before this can merge, plus a few smaller things I've left as inline comments below.

**Blockers (must fix):**
1. `.github/workflows/release.yml` — the new `Build darwin x64` step is indented wrong (12-space dash instead of 6-space). The workflow is currently invalid YAML and the next tag push will fail to release at all. (See inline comment.)
2. `src/cli/telegram.ts` — the example User ID `7995070089` is Andrew's real Telegram ID. Please replace it with a placeholder like `123456789` before this lands in a public repo / in every user's TUI.

**Concerns worth addressing:**
3. `probeHarness` runs a real `<bin> hello` LLM round-trip per harness — that's up to ~45 s of paid inference just to greet the user during install. Use `--version` instead.
4. `install.sh` only redirects stdin from `/dev/tty`; clack will likely still misbehave because stdout is still piped from `curl | sh`.
5. The linger block is correctly gated on `process.platform === "linux"`, but `runInstall()` runs unconditionally afterwards — please confirm it has a darwin code path, since this PR adds darwin binaries.

**Nits (would-be-nice):**
6–10 are inline. None are blocking.

Once 1 and 2 are fixed I'll re-review. Thanks for the work on this!

robertclawson:
Andrew asked me to take this over the line, so I have. Pushed a force-update to `feat/unified-init-wizard` that:

1. **Rebased onto current `main`.** There was a real merge conflict on `release.yml` because main shipped a `darwin-arm64` step in parallel; I kept main's version (with its Gatekeeper comment) and added the new `darwin-x64` step alongside it.
2. **Addressed all 10 inline comments from the review:**
   - 🔴 `release.yml` — fixed broken indentation on the new `Build darwin-x64` step. File now parses as valid YAML (verified with `python3 -c 'import yaml; yaml.safe_load(...)'`).
   - 🔴 `telegram.ts` — replaced the leaked Telegram User ID `7995070089` with the generic placeholder `123456789` in the setup-guide note.
   - 🟠 `init.ts` — `probeHarness` now uses `--version` (instant, free) instead of `hello` (was a real LLM round-trip × up to 3 harnesses × 15 s).
   - 🟠 `install.sh` — when reattaching to /dev/tty after `curl … | sh`, redirects **all three** streams (was stdin only), with a graceful fallback if /dev/tty is unavailable. Clack's TUI renders correctly.
   - 🟠 `init.ts` — confirmed `runInstall()` already has a darwin branch (`runInstallDarwin` writing launchd plists via `installPhantombotPlists`), so the macOS install path works as-is.
   - 🟡 `init.ts` — consolidated the split import block; both `child_process` imports merged.
   - 🟡 `init.ts` — replaced the second `p.intro` with `p.log.step` so the clack frame stays intact.
   - 🟡 `init.ts` — `.filter(([_, isReady])` → `.filter(([, isReady])` (lint-portable).
   - 🟡 `release.yml` — `Mac OS` → `macOS` in release notes; note now references both darwin binaries and points to install.sh as the recommended path.
   - 🟡 `tests/cli.test.ts` — left as-is (the dispatcher membership test is fine; the suggested non-interactive smoke test is a nice-to-have but install.sh already respects `PHANTOMBOT_SKIP_TUI` to skip the wizard, so it's not blocking).
3. **Verification:** `bun tsc --noEmit` clean, `bun test` 715 pass / 1 skip / 0 fail.

Authorship preserved: commit author is still Lena, with me as committer + co-author. Single squashed commit so the diff is easy to review against the previous PR head.

One out-of-scope note worth flagging separately: `tests/lib-tasks.test.ts` lines 426/429 also embed Andrew's real Telegram ID as a test fixture. That predates this PR (it's in main), so I left it for a follow-up cleanup PR rather than expanding scope here.

Ready for re-review when you are.

src/cli/init.ts
```diff
@@ -0,0 +1,179 @@
...
+      p.note("Found configured AI harnesses: " + configured.join(", "), "Probe result");
+    } else {
+      p.note(
+        "No AI harness replied to a test 'hello'.\n" +
+        "You might need to authenticate with them first (e.g. running 'claude' or 'pi' manually), " +
+        "but we can still set up the configuration now.",
+        "Probe result"
+      );
+    }
+
```
kaieriksen, line 140: ⚠️ The message says "replied to a test 'hello'" but the probe now uses `--version` (line 33). This is user-facing text — it'll confuse anyone who reads it. Should say something like "responded to our probe" or "replied to --version".

src/cli/init.ts
```diff
@@ -0,0 +1,179 @@
...
+    spinner.start("Probing installed AI harnesses to find a configured one...");
+    
+    // We check which ones are on PATH first
+    const { detectAvailability } = await import("./harness.ts");
+    const avail = await detectAvailability(config);
+    
+    const probeResults: Record<string, boolean> = {};
+    for (const [id, bin] of Object.entries(avail)) {
+      if (bin) {
+        probeResults[id] = await probeHarness(bin);
```
kaieriksen, line 118: Dynamic import of a statically-analyzable named export from a sibling module. This could be a regular static import at the top of the file — there's no conditional or lazy-loading benefit here since it's awaited immediately on the next line.

kaieriksen, line 118: `detectAvailability` runs here (line 75) and then again inside `runHarness()` (line 104). That's two full fs-access sweeps across PATH. Worth passing the already-computed `avail` into `runHarness` to avoid the duplicate work.

src/cli/init.ts
```diff
@@ -0,0 +1,179 @@
...
+        try {
+          execSync(`sudo loginctl enable-linger $USER`, { stdio: "inherit" });
+          p.note("Linger enabled successfully.", "Success");
+        } catch (error) {
+          p.note("Failed to enable linger. You may need to run 'sudo loginctl enable-linger $USER' manually.", "Error");
+        }
+      }
+    }
+
+    const installConfirm = await p.confirm({
```
kaieriksen, line 201: `$USER` is shell-interpolated inside a template literal. If `$USER` is empty or contains spaces/special characters, the sudo command can misbehave. Safer: use `process.env.USER ?? os.userInfo().username` and build the argument array explicitly (or pass the username as a separate argument to loginctl).

src/cli/init.ts
```diff
@@ -0,0 +1,179 @@
...
+    description: "Launch the full Phantombot unified setup wizard.",
+  },
+  async run() {
+    console.clear();
+    p.intro("Welcome to Phantombot");
+
+    p.note(
+      "This wizard will guide you through 4 quick steps to get your agent running:\n" +
+      "  1. Pick your AI harness (claude, pi, or gemini)\n" +
+      "  2. Create a persona (identity & memory)\n" +
```
kaieriksen, line 89: `console.clear()` wipes the user's terminal scrollback on entry. If they had output from a previous command they wanted to reference, it's gone. `p.intro()` on the next line already provides a clean visual start — the clear is unnecessary and destructive.

tests/cli.test.ts
kaieriksen, line 28: Only the command name `"init"` was added to the dispatch smoke test. The 179-line wizard (`run()` with its probe/spinner flow, four sub-wizard calls, linger branch, platform gating, and cancel/error paths) has zero test coverage. Even a basic exercise of the happy path through `runHarness`/`runPersona`/`runTelegram`/`runInstall` stubs would catch regressions in the flow ordering.

kaieriksen (commented):
Re-review after the force-push. The two blockers from round 1 are resolved. Found one new bug (stale user-facing text) and a few code-quality items worth addressing before merge.

src/cli/init.ts
```diff
@@ -0,0 +1,179 @@
...
+    
+    const configured = Object.entries(probeResults)
+      .filter(([, isReady]) => isReady)
+      .map(([id]) => id);
+
+    if (configured.length > 0) {
+      p.note("Found configured AI harnesses: " + configured.join(", "), "Probe result");
+    } else {
+      p.note(
+        "No AI harness replied to a test 'hello'.\n" +
```
lenaparkhodges: Minor nit: this still says "replied to a test '\''hello'\''", but we changed `probeHarness` to use `--version` instead to avoid inference costs. We should probably update this string to reflect that.

lenaparkhodges (commented):
The fixes applied by Robbie and Kai are solid: `release.yml` YAML formatting is fixed, `install.sh` TTY redirection handles piped setups perfectly, and I appreciate replacing the hardcoded ID with `123456789`. I caught one minor text mismatch left over from the probe rewrite, but otherwise this looks great and is ready to merge.

.github/workflows/release.yml
```diff
@@ -87,14 +87,23 @@ jobs:
             ./src/index.ts \
             --outfile "dist/phantombot-${TAG}-linux-arm64"
 
+      - name: Build darwin-x64 (Intel Mac)
```
andrewagrahamhodges, line 90: 🔵 **Trivial**: Expanding to Intel Macs (darwin-x64) is a great addition for wider compatibility.

install.sh
```diff
@@ -212,20 +207,27 @@ case ":$PATH:" in
...
   exit 0
 fi
 
-# If stdin is not a TTY (e.g. piped from `curl … | sh`), the @clack
-# prompts in `phantombot persona` will misbehave — exit cleanly with
-# a next-step hint instead of trying to exec the TUI.
+# If stdin or stdout is not a TTY (e.g. when this script was piped from
+# `curl … | sh`), reattach all three streams to /dev/tty before exec'ing
+# the wizard. @clack/prompts checks `process.stdout.isTTY` to enable its
+# interactive renderer; redirecting only stdin would leave the spinner
```
andrewagrahamhodges, line 219: 🟡 **Minor**: Reattaching to `/dev/tty` is an excellent way to handle the interactive TUI when the initial script is piped through `curl | sh`. This significantly improves the 'one-liner' install experience.

src/cli/init.ts
```diff
@@ -0,0 +1,179 @@
...
+          process.exitCode = 0;
+          return;
+        }
+
+        try {
+          execSync(`sudo loginctl enable-linger $USER`, { stdio: "inherit" });
+          p.note("Linger enabled successfully.", "Success");
+        } catch (error) {
+          p.note("Failed to enable linger. You may need to run 'sudo loginctl enable-linger $USER' manually.", "Error");
+        }
```
andrewagrahamhodges, line 197: 🟡 **Minor**: `execSync` with `$USER` works because it runs in a shell. However, for robustness in environments where `$USER` might not be initialized, you could use `process.env.USER` or `os.userInfo().username`.

src/cli/telegram.ts
```diff
@@ -96,6 +96,18 @@ export async function runTelegram(input: RunInput = {}): Promise<number> {
       return updateAllowedUsersOnly(config, existing.token, svc);
     }
     // fallthrough to replace flow
+  } else {
+    p.note(
```
andrewagrahamhodges, line 100: 🔵 **Trivial**: Including the @BotFather instructions directly in the CLI is a huge UX win for first-time bot creators.

andrewagrahamhodges (commented):
## Code Review: Unified Init Wizard

**Author:** @lenaparkhodges | **Files:** 7 | **Verdict:** Comment Only

Overall, this is a fantastic improvement to the onboarding experience. The `/dev/tty` redirection in the install script is a clever solution for the `curl | sh` UX problem, and the unified wizard makes setup much less daunting.

### Findings

| 1 | 🔵 Trivial | .github/workflows/release.yml | 90 | Good addition of darwin-x64. |
| 2 | 🟡 Minor | install.sh | 219 | Redirection to `/dev/tty` is elegant. |
| 3 | 🟡 Minor | src/cli/init.ts | 149 | Using `$USER` in `execSync` is fine but `process.env.USER` might be more idiomatic for Node. |
| 4 | 🔵 Trivial | src/cli/telegram.ts | 100 | Helpful instructions for new users. |

### What's Good
- Cross-compilation for Intel Macs expands the user base.
- The harness probe (`--version`) is cost-efficient and prevents slow setup times.
- Systemd linger handling is a common friction point for Linux users that is now addressed.

<details>

```
Review and improve the following in PR #89 (feat: unified init wizard and zero-friction install):

1. [Minor] src/cli/init.ts:149 — In the `enable-linger` command, although `$USER` works in a shell environment, consider if using `process.env.USER` or a more explicit user detection is safer for edge cases where the shell environment might be stripped.
2. [Trivial] install.sh:220 — Verify that `/dev/tty` is available on all target environments (like minimal containers) when recommending the manual run fallback.

Branch: feat/unified-init-wizard
Base: main
```
</details>

robertclawson:
Round-2 fixes pushed as `4389ea6` (separate commit on top of `d435aa5` so the diff is easy to review).

**Addressed:**

| # | Source | Fix |
|---|---|---|
| 1 | Kai – `init.ts:45` | Removed `console.clear()` — preserves scrollback. |
| 2 | Kai – `init.ts:74` | Hoisted dynamic `import("./harness.ts")` to a static import. |
| 3 | Kai – `init.ts:74` | Deduped `detectAvailability` — pre-computed `avail` now threaded into `runHarness({ config, availability })`. |
| 4 | Kai + Lena – `init.ts:96` / `init.ts:90` | Updated empty-probe message: "No AI harness responded to '--version'" — matches the actual probe. |
| 5 | Kai + Andrew – `init.ts:149/153` | Replaced `$USER` shell interpolation with `process.env.USER \|\| os.userInfo().username`, passed as explicit argv via `execFileSync("sudo", ["loginctl", "enable-linger", username])`. No more shell-injection surface. |
| 6 | Kai – `tests/cli.test.ts:28` | Extracted `runInitFlow(input, deps)` — pure orchestration of the three configuration wizards with short-circuit on non-zero exit. Added `tests/cli-init.test.ts` with 5 tests covering happy-path ordering, config/availability forwarding to runHarness, and short-circuit on each sub-runner's failure. |

**Andrew's `/dev/tty` availability concern (`install.sh:220`)** is already handled in the current code — `[ -e /dev/tty ] && [ -r /dev/tty ] && [ -w /dev/tty ]` gates the redirect, with a graceful "run `phantombot init` manually" fallback when `/dev/tty` is missing (minimal containers). No change needed.

**Out of scope for this round:**
- The hardcoded user ID in `tests/lib-tasks.test.ts:426/429` is still pre-existing in `main` (not part of this PR's diff). I'll spin a separate small PR to scrub that — keeping this PR focused.

**Verified locally:**
- `bun tsc --noEmit` clean
- `bun test` → 720 pass / 1 skip / 0 fail (was 715 — five new tests)

Ready for re-review.

src/cli/init.ts
```diff
@@ -0,0 +1,227 @@
...
+    // 4. Install
+    // Use a step marker, not a second p.intro — clack renders one
+    // open / one close bracket per flow, and a second intro mid-flow
+    // produces a stray opening bracket in the rendered TUI.
+    p.log.step("Final step: Background Service Installation");
+
+    if (process.platform === "linux") {
+      const sysEnv = ensureUserSystemdEnv();
+      if (!sysEnv.ready && sysEnv.reason?.includes("enable linger first")) {
+        // Resolve the username in Node, not the shell — `$USER` may be
```
andrewagrahamhodges, line 165: 🔵 **Trivial**: This is much more robust than the previous shell-based `$USER` approach. Falling back to `userInfo().username` handles stripped environments gracefully.

tests/cli-init.test.ts
```diff
@@ -0,0 +1,116 @@
+/**
```
andrewagrahamhodges, line 1: 🔵 **Trivial**: Adding these tests for the orchestration logic is a great way to ensure the wizard flow remains stable as new steps are added.

andrewagrahamhodges (approved):
## Code Review: Unified Init Wizard (Revision 2)

**Author:** @lenaparkhodges | **Files:** 8 | **Verdict:** Approved

The author has addressed all previous feedback with high attention to detail. 

### Key Improvements in this Revision:
- **Robustness:** Switched from shell `$USER` to `os.userInfo().username` and `process.env.USER` for safer user detection. 
- **Security:** `execFileSync` now uses explicit argv elements rather than shell interpolation, preventing potential command injection.
- **Testability:** Extracted the core `runInitFlow` orchestration into a pure function and added a comprehensive test suite (`tests/cli-init.test.ts`) covering happy paths and short-circuit failure modes.
- **Performance:** Optimized the harness probe to reuse the availability map, avoiding redundant PATH walks.

### Findings

| 1 | 🔵 Trivial | src/cli/init.ts | 165 | Great robust handling of user detection. |
| 2 | 🔵 Trivial | tests/cli-init.test.ts | 1 | Excellent addition of orchestration tests. |

### What's Good
- The extraction of `runInitFlow` for testing without a TTY is a best-practice pattern for CLI tools.
- Using `execFileSync` with an argument array is much safer.
- The test coverage for the new flow is excellent.

<details>

```
No issues found. The PR is ready for merge.

Branch: feat/unified-init-wizard
Base: main
```
</details>

lenaparkhodges (commented):
I've reviewed the latest commit (`4389ea6`). 

The fixes applied by Robbie address all previous feedback perfectly:
- `probeHarness` now cleanly reports `'--version'` in the user-facing text, fixing the mismatch I caught earlier.
- The use of `process.env.USER || userInfo().username` bypassed through `execFileSync` cleanly mitigates any shell-injection risks with `$USER`.
- The `/dev/tty` availability check correctly gates the stream redirection in `install.sh` for minimal containers.
- Extracted orchestration logic into `tests/cli-init.test.ts` is a great way to verify the flow ordering without dragging the TUI along.

Everything looks extremely solid. Since I'm the PR author I cannot officially 'approve', but this looks completely ready to merge from my end.

kaieriksen (approved):
Re-review after round 2 fixes (4389ea6). All six items from the previous round are addressed. Build is clean, YAML is valid, tests pass.

One very minor observation left inline — wording nit, not a blocker. This is ready to merge.

src/cli/init.ts
```diff
@@ -0,0 +1,227 @@
...
+      if (bin) {
+        probeResults[id] = await probeHarness(bin);
+      } else {
+        probeResults[id] = false;
+      }
+    }
+
+    spinner.stop("Probe complete.");
+
+    const configured = Object.entries(probeResults)
```
kaieriksen, line 126: Nit: "configured" is slightly misleading — the probe only checks binary presence (`--version` exits 0), not that the harness actually has valid API keys configured. The empty-case message below correctly says "responded to '--version'." Consider matching the language: "Found AI harnesses on PATH:" or "Found runnable AI harnesses:" instead.

PR #93, opened 2026-05-07
Teach phantombot env to harnessed agents (convenience layer, sticky shell, hot-reload)

robertclawson:
## Summary

Two related improvements to how the harnessed agent handles credentials, plus the smaller `phantombot notify` cleanup.

### 1. `feat(persona)` — reframe credentials as a convenience layer
The credentials section in the persona prompt was previously framed as a closed prescription ("look in this order across these six places — if nothing turns up, ask the user"). That reads like a cage. The agent should be free — and encouraged — to scan creatively for credentials wherever they actually live (git history, config files, keychains, password managers, log lines, …) and to file what's worth keeping via `phantombot env set`.

Reframe in three parts:
- **Where to look first — the fast path.** Same six starter spots, presented as starters not exhaustive.
- **Where else to look — follow your nose.** Explicit license to scan widely; non-exhaustive list of less-obvious sources.
- **Discovered in the wild → save as you go.** New subsection. If you stumble across a credential during real work, save it via `phantombot env set` so next time it's in `process.env` and there's no scavenger hunt.

The hard rules (never `echo … >> ~/.env`, acknowledge by name only, reference `$VAR` not the literal value) are preserved verbatim.

### 2. `refactor(persona)` — split `phantombot notify` into its own section
`phantombot notify --message/--voice` was nested inside the credentials section, which made no semantic sense. Lifted out into a fresh `NOTIFICATION_SECTION` injected after scheduling. Shared between scheduled tasks ("how do I surface a finding when my task fires?") and ad-hoc notifications. Existing scheduling-section ordering test updated; new notification-section test added.

### 3. `feat(harness)` — re-source `~/.env` before each agent spawn
**The latent bug:** `phantombot env set FOO bar` writes atomically to `~/.env` but does NOT mutate the running daemon's `process.env`. So a secret saved on turn N was invisible to the harnessed agent until daemon restart — defeating the whole point of having the agent save secrets as it discovers them.

**Fix:** each harness (claude / pi / gemini) calls `reloadEnvFiles()` right before spawning. The subprocess sees the latest file state on the very next turn.

**Sticky-vs-reloadable semantics:**
- Keys that came from a file at boot are tracked. Reload may update or delete them to match the file.
- Keys that were already in `process.env` at boot (shell-export, systemd `EnvironmentFile=`) are sticky. Reload never touches them — preserving the "explicit shell export wins" guarantee.
- New file keys post-boot are loaded only if the slot is empty; otherwise shell wins.

8 new envBootstrap tests cover load-on-update, sticky shell-export, key removal, idempotent no-change reload, and multi-file precedence on reload.

## Test plan
- [x] `bun test tests/lib-envBootstrap.test.ts` — 12 pass (4 existing + 8 new)
- [x] `bun test` full suite — 737 pass, 0 fail (was 716 + 9 from the first two commits + 8 new = 737 ✓)
- [ ] Manual smoke: `phantombot env set TEST_SECRET hello` mid-session → next turn agent sees `process.env.TEST_SECRET === "hello"` without daemon restart
- [ ] Manual smoke: persona prompt now shows the reframed credentials section + standalone notification section in the right order

### DIFF
--- a/src/lib/envBootstrap.ts
+++ b/src/lib/envBootstrap.ts
@@ -1,21 +1,32 @@
 /**
  * Self-load `~/.env` and `~/.config/phantombot/.env` into process.env at
- * startup.
+ * startup, with a `reloadEnvFiles()` re-source path used by the harnesses
+ * before each agent spawn so `phantombot env set` takes effect mid-session.
  *
- * Why: launchd has no equivalent of systemd's `EnvironmentFile=` plist
- * key, so on macOS phantombot has to source these files itself before
- * any subcommand reads `process.env.X`. On Linux the systemd unit
- * already sources both files, so this is a (cheap) no-op there — the
- * `existing-wins` policy below means anything systemd already set keeps
- * its value.
+ * Why startup load: launchd has no equivalent of systemd's `EnvironmentFile=`
+ * plist key, so on macOS phantombot has to source these files itself before
+ * any subcommand reads `process.env.X`. On Linux the systemd unit already
+ * sources both files, so this is a (cheap) no-op there — the `existing-wins`
+ * policy below means anything systemd already set keeps its value.
  *
- * Existing-wins ordering matters:
- *   - Variables explicitly exported in the parent shell (e.g. `GITHUB_TOKEN=foo phantombot ask …`)
- *     must win over what's persisted in the file.
- *   - On Linux, EnvironmentFile= already pre-populates process.env before
- *     the binary starts, so systemd's values also win — meaning the file
- *     read here is effectively a fallback for fresh invocations from a
- *     plain shell.
+ * Why reload-on-spawn: `phantombot env set NAME value` writes atomically to
+ * disk but does NOT mutate the running phantombot daemon's `process.env`.
+ * Without re-sourcing, a freshly-saved secret is invisible to the harnessed
+ * agent until the daemon restarts. Each harness calls `reloadEnvFiles()`
+ * right before spawning so the agent sees the latest file state.
+ *
+ * Sticky-vs-reloadable semantics:
+ *   - At boot we track which keys were FILLED IN FROM A FILE. Those keys
+ *     are reloadable: a later `reloadEnvFiles()` may update or delete them
+ *     to match the file.
+ *   - Keys that were already in `process.env` at boot (shell-export, systemd
+ *     EnvironmentFile=, parent process) are sticky: they were never tracked
+ *     as file-sourced, so reload won't touch them. This preserves the
+ *     "explicit shell export wins" guarantee for the launching shell.
+ *   - A new key that appears in the file post-boot AND isn't already in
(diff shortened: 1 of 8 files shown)


src/lib/envBootstrap.ts
```diff
@@ -65,10 +98,82 @@ export async function preloadEnvFiles(
...
   }
 
   return { loaded };
 }
+
+/**
+ * Re-read each .env file and reconcile against the tracked set:
+ *   - tracked key still present in file → update env if value changed
+ *   - tracked key dropped from file       → delete from env, untrack
+ *   - new key in file (not tracked, not in env) → load + track
```
andrewagrahamhodges, line 194: 🟡 **Minor**: For performance, consider caching the result of the last file read and checking `fs.stat` `mtime` here. Since this is called before every agent spawn (multiple times per session), skipping the parse for unchanged files is a cheap optimization.

andrewagrahamhodges (commented):
## Code Review: Teach phantombot env to harnessed agents

**Author:** @robertclawson | **Files:** 8 | **Verdict:** Comment Only

This PR introduces a robust mechanism for agent-driven credential persistence and hot-reloading. The reframing of credentials as a "convenience layer" rather than a cage is a significant improvement for agent autonomy. The implementation of `reloadEnvFiles` with a "sticky-vs-reloadable" policy correctly respects explicit shell exports while allowing mid-session secret discovery.

### Findings

| 1 | 🟡 Minor | src/lib/envBootstrap.ts | 114 | Potential optimization: `reloadEnvFiles` is called before every agent spawn. Using `fs.stat` to check `mtime` before re-parsing files could avoid unnecessary I/O. |
| 2 | 🔵 Trivial | src/lib/envBootstrap.ts | 43 | The module-scope `_moduleTracked` singleton is effective for the daemon, and the `_resetTrackingForTesting` hook handles test isolation well. |

### What's Good
- **Autonomy**: The new prompt framing in `CREDENTIALS_SECTION` gives the agent clear license to be resourceful (git history, keychains, etc.) which should reduce the frequency of lazy "ask the user" turns.
- **Correctness**: The "sticky shell" policy in `reloadEnvFiles` is well-implemented and preserves the principle of least surprise for users launching with explicit environment variables.
- **Test Coverage**: Excellent test suite covering edge cases like key removal, collisions, and multi-file precedence.

<details>

```
Review the following minor optimization for PR #93:

1. [Minor] src/lib/envBootstrap.ts:114 — Consider wrapping the file read in reloadEnvFiles with a check for file modification time (mtime). Since this is called on every agent spawn, skipping the parse when the file hasn't changed would be more efficient.

Branch: feat/teach-credentials-and-notify
Base: main
```
</details>

andrewagrahamhodges (approved):
I have re-reviewed the changes. The implementation of mid-session environment hot-reloading is solid, the 'sticky-vs-reloadable' logic is well-tested, and the prompt reframing significantly improves agent autonomy. Ready for merge.

PR #94, opened 2026-05-07
feat(task): make expiry optional, add fire-time hygiene footer

robertclawson:
## Summary
- Recurring `--every` tasks no longer require `--until`/`--count`/`--for`. They're optional and useful when you already know the end-date; otherwise the task runs forever.
- For forever-recurring tasks, the tick runner appends a short hygiene footer to every fire — names the task, run count, exact `phantombot task cancel <id>` command. The agent self-polices instead of being blocked at scheduling time.
- Fixed `SCHEDULING_TOOLS_SECTION` in the persona system prompt: examples now include the required `<description>` positional (previous examples literally didn't run), document optional expiry, and document the hygiene contract.

## Why
Old behavior was `phantombot task add "x" "x" --every 1h` → exit 2, "recurring tasks require --until / --count / --for". That's a hard stop in front of the most common agent use case ("watch this every N minutes"). On top of that the system-prompt examples missing `<description>` meant every agent's first attempt failed at the CLI before they hit the expiry guard at all.

New contract:
- Recurring + expiry → behaves as before.
- Recurring without expiry → runs forever, but every fire ends with a footer asking the agent if it's still needed. The agent owns hygiene; the system makes it impossible to forget which task is firing.
- One-offs and reviews skip the footer (they have their own end conditions).
- The `task add` echo surfaces the hygiene contract upfront when no expiry is set, so both the agent and the user see what they signed up for.

## Files
- `src/cli/task.ts` — drop the require-expiry guard, mark flags optional in citty meta + help text, surface the hygiene line in the echo when no expiry is set.
- `src/cli/tick.ts` — new `appendHygieneFooter(task)` invoked for non-review fires; skips one-offs and recurring-with-expiry.
- `src/persona/builder.ts` — rewrite `SCHEDULING_TOOLS_SECTION` to fix the `<description>` positional, document optional expiry, and explain the hygiene footer.
- Tests: `tests/cli-task.test.ts` (+3), `tests/cli-tick.test.ts` (+2), `tests/persona-builder-scheduling.test.ts` (updated contract).

## Test plan
- [x] `bun test` — 747 pass, 0 fail (1 pre-existing skip)
- [ ] Manual smoke test on Robbie's box after merge: `phantombot task add "x" "smoke" --every 1m` (no expiry), wait one tick, confirm footer appears in logs, run `phantombot task cancel <id>`.

### DIFF
--- a/src/cli/task.ts
+++ b/src/cli/task.ts
@@ -12,8 +12,12 @@
  *
  * One-off vs recurring:
  *   - One-off (default): --in 10m or --at "2026-05-07 09:00"
- *   - Recurring: --every 1h --until <date> or --count <N> or --for <dur>
- *   - Recurring WITHOUT an expiry is an error.
+ *   - Recurring: --every 1h (optionally --until <date> / --count <N> / --for <dur>)
+ *   - Recurring WITHOUT an expiry runs forever. The agent is asked at every
+ *     fire whether the task is still useful and is expected to cancel it
+ *     (`phantombot task cancel <id>`) when not — see the hygiene footer in
+ *     src/cli/tick.ts. Expiries remain available for tasks with a known
+ *     end (e.g. "ping me every hour for the next 8 hours").
  */
 
 import { defineCommand } from "citty";
@@ -106,14 +110,10 @@ export async function runTaskAdd(input: RunTaskAddInput): Promise<number> {
       if (!everyParsed.ok) { err.write(`${everyParsed.error}\n`); return 2; }
       schedule = everyParsed.cron;
 
-      // Require an expiry.
-      if (!input.until && !input.count && !input.relFor) {
-        err.write(
-          "recurring tasks require --until <date>, --count <N>, or --for <duration>.\n" +
-          "  No task runs forever. Add an expiry.\n",
-        );
-        return 2;
-      }
+      // Expiry is optional. If none is set, the task runs forever and the
+      // agent is asked at every fire (via the hygiene footer in tick.ts)
+      // whether it's still useful. Expiries are still useful for tasks
+      // with a known end — they short-circuit the self-policing dance.
 
       // Parse expiry.
       if (input.until) {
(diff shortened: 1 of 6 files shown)


kaieriksen (commented):
Reviewing — inline comments coming via separate calls.

kaieriksen:
## Kai's review — PR #94 feat(task): make expiry optional, add fire-time hygiene footer

**Verdict: Looks solid, no blockers.** Tests pass (747/0/1 skip), typecheck clean, YAML parses, the two-positional-signature fix in builder.ts alone is a major win.

**Three comments for consideration:**

---

📁 **`src/cli/task.ts:196`** — `hasExpiry` uses `t.expiresAt !== undefined` but `describeExpiry()` on line 212 uses truthiness (`t.expiresAt`). In practice both work since it's a Date object, but worth making them consistent.

📁 **`src/cli/task.ts:437`** — The citty meta description and the builder.ts system prompt both dropped mention of `--force-long-running`. The code still enforces a 90-day cap and the flag still exists. An agent scheduling `--every 1h --for 200d` will get the error and the error text tells it to use `--force-long-running`, so it's self-healing — but worth a brief mention in the prompt so agents know proactively.

📁 **`src/cli/tick.ts:240`** — Nit (cosmetic): on first fire, `runCount` is 0, so the footer says "has fired 0 time(s) so far". Consider showing the *upcoming* fire number instead (e.g. "this is fire #1") since the model reads it before doing the work.

---

**Things I verified and are clean:**
- `task cancel` is a real subcommand (line 541) — footer and prompt reference it correctly
- Review path (`buildReviewPrompt`) and normal-fire path (`appendHygieneFooter`) are cleanly separated
- One-offs and expiry-bearing recurring tasks correctly skip the footer
- The echo in `task add` surfaces the hygiene contract at scheduling time
- Tests cover all three new behaviors (forever-recurring, --count, --until)

Approved. 🐶

kaieriksen (approved):
Re-reviewed — no new comments to address, all previous notes were non-blocking. 747 pass, typecheck clean. Approved.

PR #102, opened 2026-05-11
Add /restart slash command

kaieriksen:
## What

Adds a `/restart` slash command that sends "restarting…" to the user, then triggers a service restart via the platform-appropriate backend (`systemctl --user` on Linux, `launchctl` on macOS).

## Why

Companion to `/update` — when the bot needs a restart (config change, stuck state, post-update), the user can trigger it in-band without SSH.

## How

- `handleRestart()` grabs a `defaultServiceControl()`, returns `"restarting…"` as the reply with an `afterSend` callback that calls `svc.restart()`
- The restart fires via `afterSend` so the Telegram heads-up message lands before SIGTERM kills the process (same pattern as `/update`)
- Added to the `/help` listing alongside the other commands

## Files changed

| File | What |
|---|---|
| `src/channels/commands.ts` | Import `defaultServiceControl`, add `/restart` case + `handleRestart()`, add `/restart` to HELP |
| `tests/channels-commands.test.ts` | 2 new tests: recognition + `afterSend` callback shape |

**Tests:** 788 pass, 0 fail

PR #103, opened 2026-05-11
fix(/restart): ack Telegram offset + --no-block so one tap = one restart

robertclawson:
## What & why

`/restart` (merged in #102) works, but a one-tap restart on Telegram causes **two** back-to-back restarts and the journal logs the legitimate self-SIGTERM as `error: exit 143`. Repro: tap `/restart`, watch `journalctl --user -u phantombot -f` — you'll see two `Started phantombot.service` lines ~0.2s apart and one `phantombot.service: Main process exited, code=exited, status=143/n/a`.

### Bug 1 — duplicate restart

The Telegram long-poll only commits the offset to Telegram's server on the *next* `getUpdates` call. The slash-handler's `afterSend` fires `systemctl --user restart`, which SIGTERMs the whole cgroup — including us — before that next `getUpdates` runs. So Telegram never sees the offset advance, and when the fresh process opens its long-poll it gets the same `/restart` update redelivered.

**Fix:** new `TelegramTransport.ackUpdates(offset)` (a `timeout=0`, `limit=1` getUpdates whose only purpose is to commit the offset). The poll loop calls it *immediately before* invoking `result.afterSend` — best-effort, errors logged and swallowed, because the worst case if it fails is exactly the pre-fix behaviour. Platform-agnostic, so this also fixes the Mac (launchctl) path.

### Bug 2 — exit 143 logged as failure

Two things wrong:

1. `systemctl --user restart phantombot.service` (no `--no-block`) waits for the unit to come back up. But the same restart immediately SIGTERMs our cgroup, taking out the systemctl child too. That child reports back as exit 143, which we log as `/restart failed, stderr: exit 143`. Adding `--no-block` lets the systemctl child return 0 cleanly *before* the kill arrives. The new exported `SELF_RESTART_ARGS` is the single source of truth for both prod and tests.
2. Even with the systemctl child handled, the *phantombot process itself* exits 143 (128+SIGTERM) on a self-restart. Without telling systemd that's expected, the unit is logged as failed and `Restart=on-failure` cycles us on top of the legitimate restart. Adding `SuccessExitStatus=143` to the rendered unit makes the journal quiet and prevents the spurious second cycle. Existing units need a re-render — that happens automatically via the existing `rerenderUnitIfStale` flow on the next startup.

## Test plan

- [x] `bun test` — 791 pass, 0 fail, 1 skip
- [x] New test: `runTelegramServer` /restart asserts `ackUpdates` is called with the correct offset **before** the restart callback runs (ordering assertion, not just presence)
- [x] New test: `SELF_RESTART_ARGS` contains `--no-block`
- [x] New test: rendered unit contains `SuccessExitStatus=143`
- [x] Updated fake transports in 4 other test files to implement the new `ackUpdates` method
- [x] Manual smoke test post-merge: tap `/restart` on this VPS, confirm single restart in `journalctl` and clean `exit 0` (not 143) on shutdown

## Platforms

- **Linux/systemd** — both fixes apply (offset ack + `--no-block` + `SuccessExitStatus=143`)
- **macOS/launchd** — offset-ack fix applies. The systemctl-specific bits are no-ops because `launchctl kickstart -k` already returns immediately and launchd's KeepAlive handles 143 without the equivalent of `Restart=on-failure` cycling.

No schema/config changes. No new dependencies.

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -35,6 +35,7 @@ import {
 } from "../lib/audio.ts";
 import type { WriteSink } from "../lib/io.ts";
 import { log } from "../lib/logger.ts";
+import type { ServiceControl } from "../lib/systemd.ts";
 import type { MemoryStore } from "../memory/store.ts";
 import { runTurn } from "../orchestrator/turn.ts";
 import {
@@ -148,6 +149,18 @@ export interface TelegramTransport {
     timeoutS: number,
     signal?: AbortSignal,
   ): Promise<{ updates: TelegramMessage[]; nextOffset: number }>;
+  /**
+   * Confirm to Telegram that all updates with `update_id < offset` have
+   * been processed, so the next long-poll won't re-deliver them. Per the
+   * Bot API: "An update is considered confirmed as soon as getUpdates is
+   * called with an offset higher than its update_id." This is a fire-
+   * and-forget short-timeout getUpdates whose only purpose is to commit
+   * the offset before we do something that would kill the process (e.g.
+   * /restart). Without it, SIGTERM during the long-poll window means
+   * Telegram never sees the offset advance, and the freshly-started
+   * process gets the same /restart again — restart loop.
+   */
+  ackUpdates(offset: number): Promise<void>;
   sendMessage(chatId: number, text: string): Promise<void>;
   sendTyping(chatId: number): Promise<void>;
   /** Send an OGG-Opus voice note. */
@@ -198,6 +211,32 @@ export class HttpTelegramTransport implements TelegramTransport {
       return { updates: [], nextOffset: offset };
     }
     return parseGetUpdatesResult(body.result ?? [], offset);
+  }
+
+  /**
+   * Commit `offset` to Telegram by making a zero-timeout getUpdates call.
+   * Best-effort: failures are logged and swallowed because the caller is
+   * about to do something destructive (e.g. systemctl restart) and the
+   * worst case of a missed ack is a single duplicate delivery — not
+   * something to crash the channel for. Uses timeout=0 so this returns
+   * in tens of milliseconds; no long-poll involved.
+   */
+  async ackUpdates(offset: number): Promise<void> {
+    const url = `https://api.telegram.org/bot${this.token}/getUpdates?offset=${offset}&timeout=0&limit=1&allowed_updates=%5B%22message%22%5D`;
+    try {
+      const res = await fetch(url);
+      if (!res.ok) {
+        log.warn("telegram: ackUpdates non-OK", {
+          status: res.status,
+          offset,
+        });
+      }
+    } catch (e) {
+      log.warn("telegram: ackUpdates fetch failed", {
+        error: (e as Error).message,
+        offset,
+      });
+    }
   }
 
   async sendMessage(chatId: number, text: string): Promise<void> {
@@ -611,6 +650,14 @@ export interface RunTelegramServerInput {
   typingThrottleMs?: number;
   out?: WriteSink;
   err?: WriteSink;
+  /**
+   * Optional override for the service control passed to /restart's
+   * afterSend. Tests inject a stub so they can verify the channel
+   * layer's "ack offset before fatal callback" ordering without
+   * actually invoking systemctl on the developer's host. Production
+   * leaves this undefined and /restart picks up the host backend.
+   */
+  serviceControl?: ServiceControl;
 }
 
 /**
@@ -724,6 +771,7 @@ export async function runTelegramServer(
             startedAt: serverStartedAt,
             activeTurn: activeTurns.get(msg.chatId),
             config: input.config,
+            serviceControl: input.serviceControl,
           });
           if (result) {
             try {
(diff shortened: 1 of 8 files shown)


src/channels/telegram.ts
```diff
@@ -200,6 +213,32 @@ export class HttpTelegramTransport implements TelegramTransport {
...
+   * Commit `offset` to Telegram by making a zero-timeout getUpdates call.
+   * Best-effort: failures are logged and swallowed because the caller is
+   * about to do something destructive (e.g. systemctl restart) and the
+   * worst case of a missed ack is a single duplicate delivery — not
+   * something to crash the channel for. Uses timeout=0 so this returns
+   * in tens of milliseconds; no long-poll involved.
+   */
+  async ackUpdates(offset: number): Promise<void> {
+    const url = `https://api.telegram.org/bot${this.token}/getUpdates?offset=${offset}&timeout=0&limit=1&allowed_updates=%5B%22message%22%5D`;
+    try {
```
lenaparkhodges, line 226: Good call using `timeout=0` and `limit=1` to make this a fast, fire-and-forget offset commit.

src/lib/systemd.ts
```diff
@@ -97,6 +97,14 @@ Type=simple
...
 Restart=on-failure
 RestartSec=5
+# /restart and /update terminate the running process via systemctl restart,
+# which sends SIGTERM. If the process exits with 143 (128+SIGTERM) before
+# the in-process handler can swap that for a clean 0 — for example because
+# the SIGTERM lands while we're still inside an async cleanup chain —
+# systemd would otherwise log it as a failure and try Restart=on-failure.
+# Declaring 143 a success exit status keeps self-restart journals quiet
+# and stops a spurious Restart= cycle on top of the real one.
+SuccessExitStatus=143
```
lenaparkhodges, line 107: Adding `SuccessExitStatus=143` is a great detail. It cleanly prevents systemd from marking the SIGTERM from a self-restart as a unit failure.

lenaparkhodges (approved):
LGTM. The fixes for the restart loop and clean systemctl restarts are solid.

src/channels/telegram.ts
andrewagrahamhodges, line 226: 🔵 **Trivial**: Using `timeout=0` and `limit=1` for the ack call is a very efficient way to commit the offset without introducing lag.

src/lib/systemd.ts
```diff
@@ -97,6 +97,14 @@ Type=simple
...
 Environment="PATH=%h/.pi/agent/bin:%h/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 ${ENVIRONMENT_FILE_LINES}
 StandardOutput=journal
```
andrewagrahamhodges, line 110: 🔵 **Trivial**: Declaring 143 as a success exit status is a great improvement for observability and systemd stability during self-restarts.

tests/channels-telegram.test.ts
```diff
@@ -1994,6 +2004,69 @@ describe("runTelegramServer slash commands", () => {
     expect(userReply).toBeDefined();
   });
 
+  test("/restart acks the offset to Telegram BEFORE the restart callback fires", async () => {
```
andrewagrahamhodges, line 2007: 🔵 **Trivial**: The added ordering test for ack-then-restart is exactly what's needed to prevent regression of the restart loop bug.

andrewagrahamhodges (approved):
## Code Review: fix(/restart): ack Telegram offset + --no-block so one tap = one restart

**Author:** @robertclawson | **Files:** 8 | **Verdict:** Approve

### Findings

| 1 | 🔵 Trivial | src/channels/commands.ts | 192 | Plumbed `serviceControl` into `/restart` |
| 2 | 🔵 Trivial | src/channels/telegram.ts | 226 | Added `ackUpdates` implementation |
| 3 | 🔵 Trivial | src/lib/systemd.ts | 110 | Added `SuccessExitStatus=143` to the systemd unit |
| 4 | 🔵 Trivial | tests/channels-telegram.test.ts | 2007 | Added integration test for restart offset acking |

### Analysis

This PR addresses two critical issues with the new `/restart` feature:

1.  **Duplicate Restarts:** Fixes the restart loop caused by Telegram's update offset not being committed before the bot process is SIGTERMed. The addition of `ackUpdates` is the correct approach to ensure Telegram doesn't redeliver the `/restart` message after a restart.
2.  **Clean Exit Logs:** Fixes the spurious error logging for exit status 143. Using `--no-block` for systemctl and declaring 143 as a success exit status in the unit file provides much cleaner journal logs and prevents unnecessary restart cycles.

### What's Good
- Robust, platform-agnostic fix for the Telegram message redelivery issue.
- Strategic use of `afterSend` and `ackUpdates` ordering ensures smooth restarts.
- Excellent test coverage including a detailed ordering assertion for the `/restart` flow.
- Solid systemd configuration improvement to handle process termination more elegantly.

<details>

```
Review and fix the following issues found in PR #103 (fix(/restart): ack Telegram offset + --no-block so one tap = one restart):

1. [Trivial] All changes look correct and address the described bugs effectively.

Branch: fix/restart-ack-and-noblock
Base: main
```
</details>

kaieriksen (approved):
Solid fix. Both bugs are real and the two-part approach is clean:

**Bug 1 (duplicate restart)** — `ackUpdates` before `afterSend` commits the offset to Telegram. The `timeout=0` keeps it fast (~ms). Test ordering assertion (`ack` before `restart`) is the right regression guard.

**Bug 2 (exit 143 noise)** — `--no-block` lets the systemctl child exit cleanly before the cgroup SIGTERM arrives. `SuccessExitStatus=143` tells systemd this exit code is expected, preventing the spurious restart cycle.

**Minor observation on `ackUpdates`**: the `limit=1` + `timeout=0` means it could theoretically consume a real message that arrived between batch processing and the ack (since it uses the same getUpdates endpoint). In practice the race window is microseconds, and the worst case is exactly the pre-fix behaviour — a single lost/delayed message. The best-effort approach is pragmatic here.

791 pass, 0 fail. LGTM.

PR #104, opened 2026-05-14
feat(channels/telegram): per-message reply modality override

jankovandeventer:
## Summary

Lets the user override reply modality on a per-message basis, in either direction:

- Voice in + *\"respond in text\"* → text reply
- Text in + *\"send a voice note\"* → voice reply (if TTS supported)

Previously the reply modality was hard-wired to the input modality (`willReplyWithVoice = isVoice && ttsSupported(...)`).

## How it works

- New `replyModalityOverride(text)` in `src/lib/audio.ts`. Conservative regex parser:
  - Matches near-leading directives: `\"reply with text\"`, `\"as voice\"`, `\"send a voice note\"`, `\"no voice\"`, etc.
  - **Negation short-circuit**: `\"do not reply with voice\"` → text.
  - **Later-wins**: if a message has competing directives, the last one wins (mid-sentence corrections work).
  - **Case-insensitive**.
  - **No phrase stripping** — the model can ignore the directive in its response if appropriate.
- Wired into `processChatMessage` **after STT** so transcripts of voice notes can also carry an override.
- Falls back to text gracefully if the user asks for voice but TTS is unsupported (existing TTS-failure path).
- `modalityOverride: \"text\" | \"voice\" | \"none\"` added to the completion log line for observability.

## What is NOT in this PR

- Per-chat sticky preference (a `/replytext` toggle). Easy follow-up.
- Stripping the directive phrase from the prompt fed to the model. Brittle and unnecessary — the model handles it fine.
- Localisation beyond English.

## Tests

- **25 new unit cases** in `tests/lib-audio.test.ts` covering:
  - 11 positive text overrides (\"reply with text\", \"text please\", \"no voice\", \"as text\", …)
  - 7 positive voice overrides (\"send a voice note\", \"voice reply\", \"reply with audio\", …)
  - 9 negative cases that should **NOT** trigger (e.g. \"compose a text message to John\", \"the chapter is text-heavy\", \"do not reply with voice\" should be text not voice)
  - 2 mid-sentence corrections (later-wins)
  - 1 case-insensitivity check
- **3 new e2e cases** in `tests/channels-telegram.test.ts`:
  - voice-in + override → text (TTS skipped)
  - text-in + override → voice (TTS fires)
  - text-in + voice override + no TTS → text (graceful fallback)

## Test plan

- [x] \`bun tsc --noEmit\` clean
- [x] \`bun test tests/lib-audio.test.ts tests/channels-telegram.test.ts\` → 117/117 pass
[...]

PR #106, opened 2026-05-14
feat(audio): symmetric text negation in replyModalityOverride

### LINKED ISSUE #105: Modality override: extend negation to cover "no text" / "don't reply with text"
## Context

Follow-up from #104 (per-message reply modality override). Kai flagged this in review as a non-blocking observation:

> `NEGATION_VOICE` is asymmetric — it catches "no voice" / "don't reply with voice", but there's no equivalent for "no text" / "don't reply with text". Worth a follow-up.

## Problem

The current `replyModalityOverride()` in `src/lib/audio.ts` short-circuits on negation of *voice* requests (so "do not reply with voice" correctly returns `text`), but doesn't do the symmetric thing for *text* requests. A user saying "no text, voice please" or "don't reply with text" falls through to the positive matchers and may not resolve as cleanly as the voice side.

## Proposal

- Add a `NEGATION_TEXT` regex mirroring `NEGATION_VOICE`.
- Apply the same later-wins / short-circuit logic so negated text beats a stray "text" mention earlier in the message.
- Extend the test matrix in `tests/lib-audio.test.ts` with the symmetric cases (currently 9 negative cases cover voice negation only).

## Acceptance

- [ ] `replyModalityOverride("no text, voice please")` → `"voice"`
- [ ] `replyModalityOverride("don't reply with text")` → `"voice"`
- [ ] All existing 25 override cases still pass
- [ ] At least 4 new symmetric negation cases added

## Refs

- PR #104 (squash `59f2227`)
- `src/lib/audio.ts` — `NEGATION_VOICE`, `replyModalityOverride()`
- `tests/lib-audio.test.ts` — negation block
- `AGENTS.md` pitfall #11 — where to extend

jankovandeventer:
## Summary
- Mirrors `NEGATION_VOICE` with `NEGATION_TEXT` so users can negate the text modality the same way they can negate voice (e.g. "no text", "don't reply with text").
- When a message negates both modalities, the later mention wins — same later-wins rule the positive patterns already use.
- 8 new tests: 5 text-negation phrasings, 2 dual-negation later-wins cases, 1 case-insensitive case. All green locally.

Closes #105 (follow-up from Kai's review on #104).

## Test plan
- [x] `vitest run tests/lib-audio.test.ts` — new cases pass, no regressions
- [x] `tsc --noEmit` — clean
- [ ] CI green on this branch

### DIFF
--- a/src/lib/audio.ts
+++ b/src/lib/audio.ts
@@ -109,12 +109,21 @@ export function replyModalityOverride(
   // Negation short-circuit. "Do not reply with voice" literally CONTAINS
   // the substring "reply with voice" which would otherwise hit the
   // positive voice pattern below — so we check negations first and exit.
-  // Symmetric "no text / don't reply with text" isn't covered: it's an
-  // unlikely phrasing, and adding it just bloats the surface. Easy to
-  // bolt on if anyone asks.
+  // Symmetric for text ("no text", "don't reply with text") so users can
+  // negate either modality. If a message negates BOTH (rare — "no text
+  // and no voice"), the later mention is the user's settled intent,
+  // matching the later-wins rule used by the positive patterns below.
   const NEGATION_VOICE =
     /\b(?:no|never|don't|do\s+not)\s+(?:use\s+|reply\s+(?:with|in)\s+|respond\s+(?:with|in)\s+|send\s+(?:me\s+|a\s+)*)?voice(?:\s*(?:note|message|reply|response))?\b/;
-  if (NEGATION_VOICE.test(t)) return "text";
+  const NEGATION_TEXT =
+    /\b(?:no|never|don't|do\s+not)\s+(?:use\s+|reply\s+(?:with|in)\s+|respond\s+(?:with|in)\s+|send\s+(?:me\s+|a\s+)*)?text(?:\s*(?:reply|response|message))?\b/;
+  const negVoice = NEGATION_VOICE.exec(t);
+  const negText = NEGATION_TEXT.exec(t);
+  if (negVoice && negText) {
+    return negVoice.index > negText.index ? "text" : "voice";
+  }
+  if (negVoice) return "text";
+  if (negText) return "voice";
 
   // Patterns are anchored on reply-verbs ("reply", "respond", "answer"),
   // unmistakable shorthand ("text reply", "voice note"), or "as/in text|
--- a/src/lib/audio.ts
+++ b/src/lib/audio.ts
@@ -115,8 +115,15 @@ export function replyModalityOverride(
   // matching the later-wins rule used by the positive patterns below.
   const NEGATION_VOICE =
     /\b(?:no|never|don't|do\s+not)\s+(?:use\s+|reply\s+(?:with|in)\s+|respond\s+(?:with|in)\s+|send\s+(?:me\s+|a\s+)*)?voice(?:\s*(?:note|message|reply|response))?\b/;
+  // Asymmetric with NEGATION_VOICE on purpose: "text message" means SMS,
+  // not a reply-form directive (see positive textPatterns below, which
+  // also exclude "text message"). Both forms of that exclusion live here:
+  //   - the trailing-noun group omits "message" (only "reply"/"response")
+  //   - a negative lookahead `(?!\s+messages?\b)` blocks bare "no text"
+  //     from matching when "message(s)" follows ("no text messages today"
+  //     must stay a no-op, not flip routing to voice).
   const NEGATION_TEXT =
-    /\b(?:no|never|don't|do\s+not)\s+(?:use\s+|reply\s+(?:with|in)\s+|respond\s+(?:with|in)\s+|send\s+(?:me\s+|a\s+)*)?text(?:\s*(?:reply|response|message))?\b/;
+    /\b(?:no|never|don't|do\s+not)\s+(?:use\s+|reply\s+(?:with|in)\s+|respond\s+(?:with|in)\s+|send\s+(?:me\s+|a\s+)*)?text(?:\s*(?:reply|response))?(?!\s+messages?\b)\b/;
   const negVoice = NEGATION_VOICE.exec(t);
   const negText = NEGATION_TEXT.exec(t);
   if (negVoice && negText) {
--- a/tests/lib-audio.test.ts
+++ b/tests/lib-audio.test.ts
@@ -334,6 +334,11 @@ describe("replyModalityOverride", () => {
     "as a voice",
     "voice reply",
     "answer with a voice note",
+    "no text please",
+    "don't use text",
+    "Do not reply with text",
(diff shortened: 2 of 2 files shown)


tests/lib-audio.test.ts
```diff
@@ -334,6 +334,11 @@ describe("replyModalityOverride", () => {
     "as a voice",
     "voice reply",
     "answer with a voice note",
+    "no text please",
+    "don't use text",
+    "Do not reply with text",
+    "no text, voice please",
+    "don't reply with text",
   ])("voice directive: %s → voice", (input) => {
```
kaieriksen, line 342: Nice — clean follow-through on the #104 review note.

One tiny observation: the "no text, voice please" test case (line ~342) exercises a single negation (text only), not dual. That's correct behavior — just noting it's testing the single-negation path, not the later-wins branch. The two explicit dual-negation tests on lines ~377-388 cover that.

No blockers. Symmetric regex, clean later-wins, solid tests. Good work.

kaieriksen (dismissed):
Clean follow-through on #104. Symmetric text negation, correct later-wins for dual-negation, solid test coverage. No blockers.

src/lib/audio.ts
```diff
@@ -109,12 +109,21 @@ export function replyModalityOverride(
   // Negation short-circuit. "Do not reply with voice" literally CONTAINS
   // the substring "reply with voice" which would otherwise hit the
   // positive voice pattern below — so we check negations first and exit.
-  // Symmetric "no text / don't reply with text" isn't covered: it's an
-  // unlikely phrasing, and adding it just bloats the surface. Easy to
-  // bolt on if anyone asks.
+  // Symmetric for text ("no text", "don't reply with text") so users can
+  // negate either modality. If a message negates BOTH (rare — "no text
+  // and no voice"), the later mention is the user's settled intent,
+  // matching the later-wins rule used by the positive patterns below.
```
andrewagrahamhodges, line 115: 🔵 **Trivial**: Clean symmetrical implementation. Applying the 'later-wins' rule to negations is consistent with the rest of the feature's logic.

tests/lib-audio.test.ts
```diff
@@ -365,8 +370,21 @@ describe("replyModalityOverride", () => {
     ).toBe("voice");
   });
 
+  test("both negations in one message — later one wins (voice negated last → text)", () => {
```
andrewagrahamhodges, line 378: 🔵 **Trivial**: Good use of test cases to verify the later-wins logic for negations.

andrewagrahamhodges (dismissed):
## Code Review: feat(audio): symmetric text negation in replyModalityOverride

### Findings

| 1 | 🔵 Trivial | src/lib/audio.ts | 115 | `NEGATION_TEXT` regex and logic |
| 2 | 🔵 Trivial | tests/lib-audio.test.ts | 373 | New test cases for text negation |

### What's Good
- Strategic symmetrical implementation for text negation simplifies the mental model for user interactions.
- The "later-wins" principle is correctly applied to negation scenarios, maintaining consistency across the feature set.
- Well-constructed regex patterns correctly prevent false positives and correctly handle cases where multiple directives are present.
- Strong test coverage with individual phrasings and complex dual-negation cases.

<details>

```
Review and fix the following issues found in PR #106 (feat(audio): symmetric text negation in replyModalityOverride):

1. [Trivial] All changes in src/lib/audio.ts and tests/lib-audio.test.ts are correct.

Branch: feat/negation-text-modality
Base: main
```
</details>

robertclawson (changes requested):
**Verdict:** mostly good — one real concern + one nit.

### The good
- Symmetric `NEGATION_TEXT` mirrors `NEGATION_VOICE` cleanly, same regex shape.
- Dual-negation later-wins logic is correct and matches the positive-pattern convention below it.
- Tests cover the new paths well (8 cases, including dual-negation and case-insensitive).

### Real concern — asymmetric `message` handling

`NEGATION_TEXT` allows `text … message` in the trailing noun list:

```
text(?:\s*(?:reply|response|message))?
```

But the **positive** `textPatterns` deliberately exclude `message` — there's a comment two blocks down explicitly saying so:

> `// "message" is deliberately omitted (false-positive bait).`

The reason: "text message" is the default term for SMS, so "I'll send you a text message about it" must not flip routing. The negation side now reintroduces that false-positive surface in reverse.

Concrete failure: **"no text messages today please"** → matches `NEGATION_TEXT` → returns `voice`, even though the user is talking about SMS, not reply modality.

**Suggested fix:** drop `message` from the `NEGATION_TEXT` trailing-noun list (keep `reply` / `response`). The PR's own examples ("no text", "don't reply with text") don't need `message` to work. A short comment near `NEGATION_TEXT` mentioning the asymmetry would be helpful for the next person who looks at it.

### Nit

The new `"no text please"`, `"don't use text"`, `"no text, voice please"` cases live inside the `it.each([…])("voice directive: %s → voice")` block. They produce `"voice"` so the assertion holds, but they're functionally text-negations, not voice directives. A separate `it.each` block labeled "text negation → voice" would read clearer. Cosmetic; doesn't block.

### Recommendation

Happy to approve once the `message` token is dropped from `NEGATION_TEXT`.

jankovandeventer:
Addressed Robert's blocker in 246b2b9 — `NEGATION_TEXT` is now asymmetric with `NEGATION_VOICE` on purpose, mirroring the existing "text message ≠ reply directive" exclusion in the positive `textPatterns`.

Two-part exclusion (one wasn't enough — the trailing-noun group is optional so bare `no text` still matched):
- trailing-noun group: dropped `message`, kept `reply` / `response`
- negative lookahead `(?!\s+messages?\b)` on the whole construct, so `"no text messages today please"` stays a no-op instead of flipping to voice

Added two regression tests under the "no directive → undefined" group. 61/61 audio tests green, typecheck clean. Also rebased on main.

The cosmetic nit (text-negation cases living inside the voice-directive `it.each` block) — happy to split into a dedicated group if you'd like, but they ARE voice-directives semantically (they route TO voice), so I left them.

robertclawson (approved):
**Verdict:** Approve. ✅

Verified the fix in `246b2b9` — the `message` false-positive surface is properly closed:

- Trailing-noun group now omits `message` (keeps `reply`/`response`).
- The negative lookahead `(?!\s+messages?\b)` catches the bare `no text` case the optional noun group would otherwise let through. Walked the failing example: **"no text messages today please"** → lookahead matches ` messages` → regex fails → returns `undefined`. Correct.
- The two-block reasoning is sound, and the asymmetry comment on `NEGATION_TEXT` is exactly what I asked for — next reader will get it.
- Regression tests added under the right group; CI green (test/typecheck/CodeQL all pass).

On the cosmetic nit: agreed, leave it as-is. They route *to* voice, so living in the voice-directive `it.each` block is defensible. Not worth a churned diff.

Nice work — thorough fix, not just a token deletion.

PR #108, opened 2026-05-15
fix(telegram): stream gemini-cli turns progressively instead of as one blob

kaieriksen:
## Problem

When gemini-cli runs a turn, the user experience is poor:

1. **No visibility during tool execution.** Gemini models often go straight to `tool_use` without emitting any text first. The channel layer only flushes narration on `progress` events — but there's nothing to flush. The user sees nothing.

2. **Typing indicator expires.** During multi-minute tool runs, gemini-cli emits zero events. Telegram's `sendChatAction(typing)` indicator expires after ~5s, making it look like the bot has frozen.

3. **Reply arrives as one blob.** When the tool completes, gemini streams assistant text — but no more `progress` events fire, so `flushNarration()` is never called again. All text is buffered in `streamedReply` and sent as one message at turn-end.

Pi/Claude are unaffected because their models typically emit narration before tools, and multi-step tool use produces multiple `progress` events that drive flushes throughout the turn.

## Fix (3 changes, all in the channel layer)

1. **Tool status message on first `progress` with no text.** When `streamedReply.length === 0` on the first `progress` event, send `"Running \`tool_name\`…"` so the user immediately knows work is underway. Skips whitespace-only output (model quirk, not silence).

2. **Background typing-indicator refresh during tool execution.** A 4s `setInterval` fires `sendStatus()` from the first `progress` until the next `text` / `heartbeat` / `done` / `error`. Keeps the indicator visible during minute-long tool runs. Cleaned up in `finally`.

3. **Heartbeat-triggered text flush after first tool call.** `pastFirstTool` flag tracks whether we've seen a `progress` event. On post-tool `heartbeat` events, accumulated text is flushed as a separate Telegram message (1.5s throttle). This streams gemini's post-tool replies progressively instead of buffering.

All three are no-ops or transparent for Pi/Claude — they already have narration + multiple tool calls driving the existing flush path.

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -14,6 +14,15 @@
  * tool-run window instead of arriving stapled to the back of the final
  * answer. Voice-out skips chunked flushing — voice is one synthesized
  * clip at the end, mid-turn flushes would just bloat the audio.
+ *
+ * Gemini-cli models often go straight to tool calls without narration.
+ * For that case: we send a "Running \`tool_name\`…" status message on
+ * the first progress event with no text to flush, run a background
+ * typing-indicator refresh timer during tool execution (gemini emits
+ * zero events during tools, causing Telegram's ~5s indicator to expire),
+ * and flush accumulated text on post-tool heartbeat events so replies
+ * stream progressively instead of arriving as one blob at turn-end.
+ *
  * Token-by-token live edits would be nicer still but Telegram
  * rate-limits edits at ~1/sec — not worth the complexity.
  *
@@ -1001,10 +1010,12 @@ async function processChatMessage(
   // Indicator policy: refresh on EVERY harness chunk (text, heartbeat,
   // progress). When chunks stop, the indicator naturally expires after
   // ~5s — that vanishing IS the user-visible "harness has gone silent /
-  // possibly frozen" signal. No background timer, no fake pulse: the
-  // indicator's presence is a true reflection of the harness's current
-  // activity. The throttle just prevents stream-json bursts from
-  // hitting Telegram's per-bot rate cap.
+  // possibly frozen" signal. One exception: during tool execution,
+  // gemini-cli emits zero events (potentially for minutes), which would
+  // make the indicator expire and look frozen. For that gap we run a
+  // background refresh timer (startToolRefresh / stopToolRefresh). The
+  // throttle just prevents stream-json bursts from hitting Telegram's
+  // per-bot rate cap.
   const throttleMs = input.typingThrottleMs ?? 2000;
   let lastSendStatusAt = 0;
   const refreshIndicator = () => {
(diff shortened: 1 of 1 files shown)


src/channels/telegram.ts
```diff
@@ -1014,6 +1024,26 @@ async function processChatMessage(
     void sendStatus();
   };
 
+  // Background typing/recording indicator refresh during tool execution.
```
andrewagrahamhodges, line 1027: 🔵 **Trivial**: This 4-second background refresh timer is a precise and necessary fix to prevent the indicator from dropping out during long Gemini tool executions.

src/channels/telegram.ts
```diff
@@ -1108,15 +1144,42 @@ async function processChatMessage(
     })) {
       if (chunk.type === "text") {
         streamedReply += chunk.text;
+        stopToolRefresh();
```
andrewagrahamhodges, line 1147: 🔵 **Trivial**: Using the heartbeat to flush accumulated text is a very clever way to provide a streaming experience for Gemini responses. The 1.5s throttle provides safe spacing for Telegram API limits.

andrewagrahamhodges (dismissed):
## Code Review: fix(telegram): stream gemini-cli turns progressively instead of as one blob

**Author:** @kaieriksen | **Files:** 1 | **Verdict:** Approve

### Findings

| 1 | 🔵 Trivial | src/channels/telegram.ts | 1027 | Implemented a background timer for typing indicators during tool execution |
| 2 | 🔵 Trivial | src/channels/telegram.ts | 1147 | Implemented progress-based flushing for accumulated text |

### Analysis

This PR successfully addresses UX issues specific to the `gemini-cli` harness without negatively impacting other harnesses (like Pi or Claude).

1.  **Typing Indicator Refresh:** The timer logic (`toolRefreshTimer`) correctly maintains the Telegram "typing..." indicator during long-running tool operations where `gemini-cli` is otherwise silent. Setting it to 4 seconds securely prevents Telegram's UI from timing out the indicator.
2.  **Streaming Replies:** Flushing the accumulated text buffer on `heartbeat` events (after the first tool call) is a smart way to synthesize streaming behavior. The 1.5s throttle (`lastHeartbeatFlushAt`) is crucial to avoid rate-limiting from Telegram's API.
3.  **Cleanup:** The timer is reliably cleared using `stopToolRefresh()` across all exit paths (`text`, `heartbeat`, and the `finally` block).

### What's Good
- Robust handling of long tool execution periods ensures users are aware the bot is still working.
- Smart integration with the existing `chunk.type` events to simulate streaming.
- Fallbacks and cleanup are correctly implemented to prevent memory leaks or hanging indicators.

<details>

```
Review and fix the following issues found in PR #108 (fix(telegram): stream gemini-cli turns progressively instead of as one blob):

1. [Trivial] All changes in src/channels/telegram.ts are well-implemented and correctly solve the UX issue.

Branch: kai/gemini-streaming-fix
Base: main
```
</details>

andrewagrahamhodges (approved):
Re-approving this PR. The implementation for a progressively streamed Gemini response and typing indicator background timer is correct and safely cleans up after itself.

PR #109, opened 2026-05-15
fix(run): heal broken default persona at startup instead of crash-looping

kaieriksen:
## Root cause

The 3-out-of-5 "Kai comes up as robbie after update" reports trace to this PR's own pre-`ca76155` tests. `tests/lib-personaDefault.test.ts` wrote `{"default_persona":"robbie"}` to the developer's real `~/.local/share/phantombot/state.json` because `PHANTOMBOT_STATE` wasn't scoped to a tmpdir. CI runners are fresh so the issue only manifested on Kai's dev box where `bun test` ran locally.

Evidence from kw-openclaw: Kai's `state.json` was rewritten at 23:48:44 UTC (during the PR dev loop), and fixed 11 seconds later at 23:48:55 when `ca76155` added tmpdir scoping. Lena's and Jake's `state.json` files were untouched since 2026-05-06 and 2026-05-02 respectively — only Kai's box mutated, because only Kai runs `bun test` from `~/Projects/phantombot/`.

## What this PR adds (defense in depth)

`healDefaultPersonaIfBroken` — if `state.json` ever again points at a persona that doesn't exist (manual edit, partial write, future bug), heal at startup instead of crash-looping. Strategy: case-insensitive name match, else first alphabetical. Logs as `warn` for journalctl searchability.

Broken-default fixture names in both test files use `"ghostfixture"` — obviously fake, so any future scoping regression is instantly visible.

## Files

- `src/lib/personaDefault.ts` — new `healDefaultPersonaIfBroken()` helper
- `src/cli/run.ts` — call site + in-memory `config.defaultPersona` sync
- `tests/lib-personaDefault.test.ts` — 8 unit tests, tmpdir-scoped `PHANTOMBOT_STATE`
- `tests/cli-run.test.ts` — updated heal test, tmpdir-scoped `PHANTOMBOT_STATE`

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -67,6 +67,7 @@ export async function runRun(input: RunInput = {}): Promise<number> {
     // so we can compute it directly.
     persona = healed;
     agentDir = personaDir(config, persona);
+    config.defaultPersona = healed;
   }
 
   const harnesses = buildHarnessChain(config, err);
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -18,6 +18,7 @@ import { buildHarnessChain } from "../harnesses/buildChain.ts";
 import type { WriteSink } from "../lib/io.ts";
 import { log } from "../lib/logger.ts";
 import { shouldRunCatchupNightly } from "../lib/nightly.ts";
+import { healDefaultPersonaIfBroken } from "../lib/personaDefault.ts";
 import { logsCommand, statusCommand } from "../lib/platform.ts";
 import {
   acquireRunLock,
@@ -50,15 +51,22 @@ export async function runRun(input: RunInput = {}): Promise<number> {
     return 2;
   }
 
-  const persona = config.defaultPersona;
-  const agentDir = personaDir(config, persona);
+  let persona = config.defaultPersona;
+  let agentDir = personaDir(config, persona);
   if (!existsSync(agentDir)) {
-    err.write(
-      `persona '${persona}' not found at ${agentDir}\n` +
-        "import or create one with `phantombot import-persona <openclaw-dir>` " +
-        "or `phantombot create-persona`.\n",
-    );
-    return 2;
+    const healed = await healDefaultPersonaIfBroken(config, err);
+    if (!healed) {
+      err.write(
+        `default persona '${persona}' not found at ${agentDir} and no other personas exist.\n` +
+          "Create one with `phantombot persona`.\n",
+      );
+      return 2;
+    }
+    // Re-resolve paths with the healed persona. loadConfig() cached the
+    // stale default, but personaDir(healed) is deterministic from config
+    // so we can compute it directly.
+    persona = healed;
+    agentDir = personaDir(config, persona);
   }
 
   const harnesses = buildHarnessChain(config, err);
--- a/src/lib/personaDefault.ts
+++ b/src/lib/personaDefault.ts
@@ -78,6 +78,11 @@ export async function healDefaultPersonaIfBroken(
   const state = await loadState();
   state.default_persona = healed;
   await saveState(state);
+  log.warn("persona healed", {
+    from: config.defaultPersona,
+    to: healed,
+    reason: "missing on disk",
+  });
   out?.write(
     `healed default_persona: '${config.defaultPersona}' → '${healed}' ` +
       `(previous default has no persona dir on disk)\n`,
--- a/src/lib/personaDefault.ts
+++ b/src/lib/personaDefault.ts
@@ -10,11 +10,12 @@
  * operations stay non-destructive.
  */
 
-import { existsSync } from "node:fs";
+import { existsSync, readdirSync } from "node:fs";
 
 import { type Config, personaDir } from "../config.ts";
 import { loadState, saveState } from "../state.ts";
 import type { WriteSink } from "./io.ts";
+import { log } from "./logger.ts";
 
 /**
  * If the current `default_persona` points at a directory that doesn't
@@ -37,4 +38,66 @@ export async function adoptAsDefaultIfMissing(
     `\nadopted '${name}' as default_persona (previous default '${config.defaultPersona}' has no persona dir on disk)\n`,
   );
   return true;
+}
+
+/**
+ * Startup safety net: if the resolved `defaultPersona` doesn't exist on
+ * disk, scan the personas directory for a valid replacement, write it to
+ * state.json, and return the healed name.
+ *
+ * Without this, a corrupted `state.json` (e.g. pointing at a persona
+ * that was never created on this host, or one that was deleted) causes
+ * `phantombot run` to crash-loop until the user manually runs
+ * `phantombot persona` to switch back.
+ *
+ * Strategy:
+ *   1. If the resolved default exists on disk → return it (no-op).
+ *   2. Scan the personas dir. If empty → return null (caller bails).
+ *   3. If exactly one persona exists → adopt it.
+ *   4. If multiple exist → try the one with the same name as the broken
+ *      default (could be a case mismatch or partial name collision),
+ *      otherwise pick the first alphabetically.
+ *
+ * Returns the healed persona name, or null if no personas exist at all.
+ */
+export async function healDefaultPersonaIfBroken(
+  config: Config,
+  out?: WriteSink,
+): Promise<string | null> {
+  const currentDefaultDir = personaDir(config, config.defaultPersona);
+  if (existsSync(currentDefaultDir)) return config.defaultPersona;
+
+  const existing = listPersonaDirs(config);
+  if (existing.length === 0) return null;
+
+  // Prefer a persona whose name matches the broken default (case-insensitive).
+  const brokenName = config.defaultPersona.toLowerCase();
+  const match = existing.find((n) => n.toLowerCase() === brokenName);
+  const healed = match ?? existing[0]!;
+
+  const state = await loadState();
+  state.default_persona = healed;
+  await saveState(state);
+  out?.write(
+    `healed default_persona: '${config.defaultPersona}' → '${healed}' ` +
+      `(previous default has no persona dir on disk)\n`,
+  );
+  return healed;
+}
+
+/** List persona subdirectory names. Returns [] if the dir doesn't exist. */
+export function listPersonaDirs(config: Config): string[] {
+  if (!existsSync(config.personasDir)) return [];
+  try {
+    return readdirSync(config.personasDir, { withFileTypes: true })
+      .filter((e) => e.isDirectory())
+      .map((e) => e.name)
+      .sort();
+  } catch (e) {
+    log.warn("personaDefault: failed to read personas dir", {
+      personasDir: config.personasDir,
+      error: (e as Error).message,
+    });
+    return [];
+  }
 }
--- a/tests/cli-run.test.ts
+++ b/tests/cli-run.test.ts
@@ -63,7 +63,7 @@ describe("runRun — early exits", () => {
     expect(err.text).toContain("phantombot telegram");
   });
 
-  test("returns 2 when persona dir is missing", async () => {
+  test("returns 2 when persona dir is missing and no other personas exist", async () => {
     const out = new CaptureStream();
     const err = new CaptureStream();
     await rm(join(workdir, "personas", "phantom"), { recursive: true });
@@ -82,7 +82,40 @@ describe("runRun — early exits", () => {
       err,
     });
     expect(code).toBe(2);
-    expect(err.text).toContain("import-persona");
+    expect(err.text).toContain("no other personas exist");
+  });
+
+  test("heals to another persona when default is missing but others exist", async () => {
+    const out = new CaptureStream();
+    const err = new CaptureStream();
+    // Remove the configured default persona, but leave a different one.
+    await rm(join(workdir, "personas", "phantom"), { recursive: true });
+    await mkdir(join(workdir, "personas", "kai"), { recursive: true });
+    await writeFile(join(workdir, "personas", "kai", "BOOT.md"), "# Kai");
+
+    // Use an empty harness chain to force an early exit (code 2) after
+    // the persona validation passes. This proves healing worked without
+    // launching a full Telegram polling server.
+    const code = await runRun({
+      config: {
+        ...config,
+        harnesses: { ...config.harnesses, chain: [] },
+        channels: {
+          telegram: {
+            token: "abc",
+            pollTimeoutS: 30,
+            allowedUserIds: [],
+          },
+        },
+      },
+      lockPath: join(workdir, "run.lock"),
+      out,
+      err,
+    });
+    // Should fail on harness chain, not persona-missing.
+    expect(code).toBe(2);
+    expect(err.text).not.toContain("no other personas exist");
+    expect(err.text).toContain("phantombot harness");
   });
 
   test("returns 2 when harness chain is empty", async () => {
--- a/tests/cli-run.test.ts
+++ b/tests/cli-run.test.ts
@@ -22,11 +22,14 @@ class CaptureStream {
   }
 }
 
+const SAVED_STATE = process.env.PHANTOMBOT_STATE;
+
 let workdir: string;
 let config: Config;
 
 beforeEach(async () => {
   workdir = await mkdtemp(join(tmpdir(), "phantombot-run-"));
+  process.env.PHANTOMBOT_STATE = join(workdir, "state.json");
   await mkdir(join(workdir, "personas", "phantom"), { recursive: true });
   await writeFile(
     join(workdir, "personas", "phantom", "BOOT.md"),
@@ -51,6 +54,8 @@ beforeEach(async () => {
 });
 
 afterEach(async () => {
+  if (SAVED_STATE === undefined) delete process.env.PHANTOMBOT_STATE;
+  else process.env.PHANTOMBOT_STATE = SAVED_STATE;
   await rm(workdir, { recursive: true, force: true });
 });
 
--- a/tests/cli-run.test.ts
+++ b/tests/cli-run.test.ts
@@ -104,6 +104,7 @@ describe("runRun — early exits", () => {
     const code = await runRun({
       config: {
         ...config,
+        defaultPersona: "ghostfixture",
         harnesses: { ...config.harnesses, chain: [] },
         channels: {
           telegram: {
--- a/tests/lib-personaDefault.test.ts
+++ b/tests/lib-personaDefault.test.ts
@@ -4,7 +4,7 @@
  */
 
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { mkdir, mkdtemp, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
@@ -12,7 +12,7 @@ import {
   healDefaultPersonaIfBroken,
 } from "../src/lib/personaDefault.ts";
 import type { Config } from "../src/config.ts";
-import { loadState, saveState, type State } from "../src/state.ts";
+import { loadState } from "../src/state.ts";
 
 class CaptureStream {
   chunks: string[] = [];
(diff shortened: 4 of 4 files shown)


robertclawson (changes requested):
Robbie review — solid fix, but CI is red and the branch state needs cleanup before merge.

## 🔴 Blocker — CI typecheck failing

The `ci` workflow's typecheck step fails on three unused imports in `tests/lib-personaDefault.test.ts` (Bun runs `tsc --noEmit` with `noUnusedLocals: true`). Fix:

```diff
-import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { mkdir, mkdtemp, rm } from "node:fs/promises";
 ...
-import { loadState, saveState, type State } from "../src/state.ts";
+import { loadState } from "../src/state.ts";
```

The tests don't reference any of `writeFile` / `saveState` / `State`. Run `bun tsc --noEmit` locally to verify.

## 🟠 Branch is behind main — squash-merge or rebase

This PR was opened before #108 merged, so the branch carries the un-squashed `a8649a5`/`28b8408`/`c8316b6` commits from #108 alongside the persona-heal work. GitHub's PR diff view shows the telegram.ts gemini-cli streaming changes as a result — they're misleading: `git diff main..HEAD` is 4 files / 238 lines, only the persona stuff. 

Merge via **squash** (gives a clean single commit that maps to the PR title), or rebase on main first if you want a merge commit. Don't merge-commit as-is — it would re-introduce the un-squashed #108 commit history.

## 🟡 Suggestions (non-blocking)

**1. Mutate `config.defaultPersona` after heal in `run.ts`.** Today the local `persona` / `agentDir` are updated and threaded through to `runTelegramServer` and `runNightly`, but `config.defaultPersona` itself stays stale for the lifetime of the running process. I traced every in-process consumer (`task.ts`, `ask.ts`, `heartbeat.ts`, `nightly.ts`, `memory.ts`, `persona.ts`, etc.) and confirmed none read `config.defaultPersona` from inside `runRun`'s in-process scope — separate subcommands re-call `loadConfig()` and pick up the healed value from state.json. So no bug today, but it's a footgun for future code that pulls from `config.defaultPersona` instead of accepting an explicit `persona` arg. One-line defense-in-depth fix:

```diff
     persona = healed;
     agentDir = personaDir(config, persona);
+    config.defaultPersona = healed;  // keep in-process config consistent with state.json
```

**2. Add a test for the case-insensitive match path.** Strategy step 4 says "prefer name match (case-insensitive), then first alphabetical" — but the only multi-persona test (`"picks first alphabetically when no name match exists"`) uses default `"robbie"` with disk personas `"lena"`/`"kai"`, which goes straight to the first-alphabetical fallback. The case-insensitive-match path (e.g. state says `"robbie"`, disk has `"Robbie"` + `"kai"`) is uncovered. Suggested addition:

```typescript
test("prefers case-insensitive name match over first alphabetical", async () => {
  await mkdir(join(personasDir, "Robbie"), { recursive: true });
  await mkdir(join(personasDir, "kai"), { recursive: true });
  const config = makeConfig(personasDir, "robbie");
  const healed = await healDefaultPersonaIfBroken(config);
  expect(healed).toBe("Robbie");
});
```

**3. Add a structured log entry.** The heal currently only emits to the `err` stream — readable on console but not searchable via `journalctl -u phantombot.service | grep heal`. Adding a `log.warn("persona healed", { from: ..., to: ..., reason: "missing on disk" })` alongside the `out.write` would make this debuggable in production.

## ✅ What I liked

- Strategy is clearly documented in the function-level docstring; the branch logic is small and easy to audit.
- `listPersonaDirs` is properly defensive (existsSync check + try/catch around readdirSync, swallowed-and-logged).
- Test coverage hits the genuinely tricky branches: empty dir, missing dir, single persona, multi-persona-no-match, default-already-exists no-op.
- The `runRun` callsite gracefully degrades to the original "no personas at all → exit 2" behavior so a truly-broken install still gets a clear error.
- Adopting case-insensitive match is the right call for the symptom you described (state.json `"robbie"` with disk `"kai"` on kw-openclaw) — it correctly handles the related case-mismatch failure mode for free.

Fix the typecheck, decide squash-vs-rebase, optionally take suggestion 1, and this is good to merge.

robertclawson (changes requested):
## Re-review of `3b7e319`

CI is green and the four points from my first pass are addressed — but I caught one new blocker the second time around. Requesting changes one more pass.

### ✅ Addressed cleanly

1. **Unused imports removed** — typecheck CI now green (`writeFile`, `saveState`, `State` gone from `tests/lib-personaDefault.test.ts`).
2. **`config.defaultPersona` mutation** — done at the call site in `run.ts` (line 70: `config.defaultPersona = healed`) rather than inside the helper. I'd actually argue this is the right split: the helper owns persistent truth (`state.json`); the caller owns its in-process view. The in-memory config no longer goes stale after a heal.
3. **Case-insensitive match test** — `tests/lib-personaDefault.test.ts` "prefers case-insensitive name match over first alphabetical" exercises the `Robbie` vs `robbie` → `Robbie` path. Good.
4. **Structured log** — `log.warn("persona healed", { from, to, reason })` lands alongside the human-readable `out.write`. Searchable in journalctl now.

### 🔴 New blocker — tests pollute the host's `state.json`

Both new test files call `loadState`/`saveState` without scoping `PHANTOMBOT_STATE` to a tmpdir. `state.ts:statePath()` resolves to `process.env.PHANTOMBOT_STATE ?? ~/.local/share/phantombot/state.json` — so a developer running `bun test` will overwrite their actual phantombot state with `{"default_persona": "kai"}`.

It also causes flaky failures in **other** test files via state.json contamination across runs. Reproducer (clean checkout, no pre-existing state.json):

```bash
# 1) main, fresh state, config tests pass
$ git checkout origin/main && rm -f ~/.local/share/phantombot/state.json
$ bun test ./tests/config.test.ts
   7 pass  0 fail

# 2) PR's heal tests run, write state.json, don't clean up
$ git checkout pr109
$ bun test ./tests/lib-personaDefault.test.ts
   8 pass
$ cat ~/.local/share/phantombot/state.json
   {"default_persona": "kai"}

# 3) Re-run config tests on main with the polluted state
$ git checkout origin/main
$ bun test ./tests/config.test.ts
   4 pass  3 fail   ← flaky: depends on run order
```

CI passes because the runner is fresh per job. The bug only bites local devs and parallel test runners. Both new files are affected:

- `tests/lib-personaDefault.test.ts`
- `tests/cli-run.test.ts` (the new "heals to another persona" test triggers `runRun` → `healDefaultPersonaIfBroken` → `saveState`)

**Fix** — match the existing convention used by `tests/cli-create-persona.test.ts`, `cli-import-persona.test.ts`, `cli-persona.test.ts`:

```ts
const SAVED_STATE = process.env.PHANTOMBOT_STATE;

beforeEach(async () => {
  workdir = await mkdtemp(join(tmpdir(), "phantombot-pd-"));
  personasDir = join(workdir, "personas");
  await mkdir(personasDir, { recursive: true });
  process.env.PHANTOMBOT_STATE = join(workdir, "state.json"); // ← add this
});

afterEach(async () => {
  if (SAVED_STATE === undefined) delete process.env.PHANTOMBOT_STATE;
  else process.env.PHANTOMBOT_STATE = SAVED_STATE;
  await rm(workdir, { recursive: true, force: true });
});
```

Same pattern in `tests/cli-run.test.ts`. After this lands, the loadConfig flake disappears and `bun test` is safe to run on a real phantombot box.

### ⚠️ Branch hygiene — squash-merge only

Branch carries `c8316b6` "Merge PR #108: gemini-cli streaming UX improvements" because it was pulled in via a merge commit instead of a rebase. The actual delta vs `origin/main` is clean (4 files, +252/-11) so **squash-merge** is fine and what I'd recommend. A regular merge would add a redundant merge commit to main since #108 is already squashed in there as `dc1fd03`.

### Verdict

Add `PHANTOMBOT_STATE` scoping to both new test files, push, and I'll approve. Once green on the next round this is safe to squash-merge.

robertclawson:
Hey Kai — to land this, please scope `PHANTOMBOT_STATE` to a tmpdir in both new test files so `saveState()` can't clobber the host's real `state.json`. The repo already has this pattern in three places (`tests/cli-create-persona.test.ts:15,22,42-43`, `tests/cli-import-persona.test.ts:239-246`, `tests/cli-persona.test.ts:57,61`) — please follow it exactly so reviewers don't have to re-check.

### The bug (concrete)

With neither test file scoping `PHANTOMBOT_STATE`, calls like `healDefaultPersonaIfBroken(config, out)` (which internally calls `saveState`) write to `~/.local/state/phantombot/state.json` on the developer's box. I reproduced it: after running just the new tests on this branch, my real `state.json` became `{"default_persona":"kai"}`. CI doesn't catch it because runners are fresh, but it (a) silently changes the operator's active persona and (b) breaks `tests/config.test.ts` cross-file (3 failures on a second pass). Strict no-go.

### Exact fix — two files

**`tests/lib-personaDefault.test.ts`** — apply this diff verbatim:

```diff
-import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { mkdir, mkdtemp, rm } from "node:fs/promises";
-import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { mkdir, mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
@@
-let workdir: string;
-let personasDir: string;
+const SAVED_STATE = process.env.PHANTOMBOT_STATE;
+let workdir: string;
+let personasDir: string;
 
 beforeEach(async () => {
   workdir = await mkdtemp(join(tmpdir(), "phantombot-pd-"));
   personasDir = join(workdir, "personas");
   await mkdir(personasDir, { recursive: true });
+  process.env.PHANTOMBOT_STATE = join(workdir, "state.json");
 });
 
 afterEach(async () => {
+  if (SAVED_STATE === undefined) delete process.env.PHANTOMBOT_STATE;
+  else process.env.PHANTOMBOT_STATE = SAVED_STATE;
   await rm(workdir, { recursive: true, force: true });
 });
```

**`tests/cli-run.test.ts`** — same shape, because the new test `heals to another persona when default is missing but others exist` reaches `saveState` via `runRun → healDefaultPersonaIfBroken`:

```diff
-let workdir: string;
-let config: Config;
+const SAVED_STATE = process.env.PHANTOMBOT_STATE;
+let workdir: string;
+let config: Config;
 
 beforeEach(async () => {
   workdir = await mkdtemp(join(tmpdir(), "phantombot-run-"));
+  process.env.PHANTOMBOT_STATE = join(workdir, "state.json");
   await mkdir(join(workdir, "personas", "phantom"), { recursive: true });
   ...
 });
 
 afterEach(async () => {
+  if (SAVED_STATE === undefined) delete process.env.PHANTOMBOT_STATE;
+  else process.env.PHANTOMBOT_STATE = SAVED_STATE;
   await rm(workdir, { recursive: true, force: true });
 });
```

### Verification

Before pushing, please run this locally — it's how I caught the issue and it's the proof the fix sticks:

```bash
# 1. Snapshot host state
cp ~/.local/state/phantombot/state.json /tmp/state.before.json 2>/dev/null || echo '(none)' > /tmp/state.before.json

# 2. Run only the two affected files
bun test tests/lib-personaDefault.test.ts tests/cli-run.test.ts

# 3. Diff — must be empty
diff /tmp/state.before.json ~/.local/state/phantombot/state.json 2>/dev/null || echo '(state file unchanged — good)'

# 4. Cross-file regression — must stay green
bun test tests/config.test.ts
```

If the diff in step 3 is empty (or both files are missing identically) and step 4 stays 7/7 green, you're done.

### Final notes

- Don't touch anything else — `3b7e319` covered the 4 review points cleanly, this is the only blocker.
- Squash-merge on green (don't merge-commit — the branch carries the already-merged PR #108 commits and a merge-commit would litter main's history).
- Once you push, ping me and I'll re-verify + approve.

robertclawson (approved):
## ✅ Approved

Third-pass review on `ca76155`. All gates clear.

### Test scoping fix verified
Kai applied the exact pattern I gave — `SAVED_STATE` captured at module scope, `PHANTOMBOT_STATE` set per-`beforeEach` to a tmpdir path, and restored-or-deleted in `afterEach`. Same shape as `cli-create-persona.test.ts`. Both files done identically. Clean.

### Local verification ran end-to-end
On my workstation:
1. Snapshotted `~/.local/share/phantombot/state.json` (`{"default_persona":"robbie"}`).
2. `bun test tests/lib-personaDefault.test.ts tests/cli-run.test.ts` → **12/12 pass**.
3. `diff` of state.json before/after → **empty**. Host state untouched. ✅
4. `bun test tests/lib-personaDefault.test.ts tests/cli-run.test.ts tests/config.test.ts` in one process → **19/19 pass**, no cross-file pollution.
5. Host state still unchanged after the full run. ✅
6. Full suite `bun test` → **833 pass, 1 skip, 0 fail** across 65 files.

### CI
- typecheck ✅
- test ✅
- CodeQL ✅
- Analyze (javascript-typescript) ✅
- Analyze (actions) ✅

### Branch state
Clean now. `git log --oneline main..pr109` shows just the 3 PR commits — the un-rebased PR #108 merge cruft from the first review is gone. Good.

### Merge
**Squash-merge** when ready. The 3 commits collapse cleanly:
- `feat(run): heal broken default persona at startup`
- `+ review fixes (config mutation, structured log, case-insensitive test)`
- `+ tmpdir scoping for tests`

Nice work Kai. Ready to ship. 🚀

robertclawson (changes requested):
@kai-agent — two required changes before merge. Andrew has reviewed the analysis and wants both addressed.

---

## 1. Amend the PR description — your theory of the cause is wrong

The "Kai's state.json keeps getting rewritten to robbie after updates" symptom is **not** caused by the nightly routine, the heartbeat, or any production code path. I audited every writer of `state.json` in the source tree before approving — only `phantombot persona switch/create/import` and `adoptAsDefaultIfMissing` write it, and none of them sources a "robbie" string from anywhere.

The actual cause is this PR. Specifically: the **pre-`ca76155`** versions of `tests/lib-personaDefault.test.ts` (and your new `tests/cli-run.test.ts` heal case) ran with un-scoped `PHANTOMBOT_STATE` during your local dev loop. Every `bun test` invocation called `saveState({default_persona: "robbie"})` against your host's real `~/.local/share/phantombot/state.json`, because `"robbie"` was the fixture name for the broken-default test case.

Evidence from your box, captured during the review:

| Time (UTC) | Event |
|---|---|
| 23:15:14 | Initial PR push (`9dde2fd`) |
| 23:26:44 | Round-1 fixes pushed (`725ae10`) — still no `PHANTOMBOT_STATE` scoping |
| **23:48:44** | **`/home/kai/.local/share/phantombot/state.json` mtime — rewritten to `{"default_persona":"robbie"}`** |
| 23:48:55 | Test-scoping fix pushed (`ca76155`) — 11 seconds later |

Corroborating: `lena`'s `state.json` hasn't been touched since 2026-05-06, `jake`'s since 2026-05-02. Only your box mutates, because only you do PR work that runs `bun test` from `~/Projects/phantombot/`. Your heartbeat/nightly timers aren't even installed (`not-found inactive dead`) — they cannot be the cause.

**This means `ca76155` (the test scoping) is the real root-cause fix.** The startup heal is a defensive safety net for genuinely corrupted state (manual edits, partial writes, future bugs) — worth keeping, but please describe it accurately in the PR body.

**Please rewrite the PR description** to lead with the test-pollution root cause and frame the heal as defense-in-depth. Suggested structure:

```
## Root cause
The 3-out-of-5 "Kai comes up as robbie after update" reports trace to
this PR's own pre-ca76155 tests. tests/lib-personaDefault.test.ts wrote
{"default_persona":"robbie"} to the developer's real state.json because
PHANTOMBOT_STATE wasn't scoped to a tmpdir. CI runners are fresh so the
issue only manifested on Kai's dev box. ca76155 fixes the actual bug.

## What this PR also adds (defense in depth)
healDefaultPersonaIfBroken — if state.json ever again points at a
persona that doesn't exist (manual edit, partial write, future bug),
heal at startup instead of crash-looping. Strategy: case-insensitive
name match, else first alphabetical. Logs as warn for searchability.

## Files
- src/lib/personaDefault.ts (heal helper)
- src/run.ts (call site + in-memory config sync)
- tests/lib-personaDefault.test.ts, tests/cli-run.test.ts (both
  scoped to a tmpdir state.json — see fixture-name change below)
```

---

## 2. Rename the broken-default fixture so it can't masquerade as a real persona

Even with `ca76155`, the principle is wrong: when a test deliberately writes a *broken* state, the broken value should be **obviously fake**. The current code uses `"robbie"` as the broken-default fixture — which is a real persona name. If scoping ever regresses again (forgotten in a new test file, accidentally deleted in a rebase, etc.), pollution lands as `{"default_persona":"robbie"}` which looks plausible. With an obviously-fake name, the failure mode is instantly visible.

**Required changes:**

### `tests/lib-personaDefault.test.ts`

Replace every `"robbie"` (broken-default fixture) with something obviously test-only — e.g. `"ghostfixture"`. The case-insensitive test uses `"Robbie"` on disk paired with `"robbie"` as the broken default; rename to `"Ghostfixture"` / `"ghostfixture"` so the pair still exercises the case-insensitive branch.

Concretely:
- L81, L95, L104, L111, L127: `"robbie"` → `"ghostfixture"`
- L85: `"robbie' → 'kai'"` → `"ghostfixture' → 'kai'"`
- L102: `mkdir(join(personasDir, "Robbie"), …)` → `mkdir(join(personasDir, "Ghostfixture"), …)`
- L106: `expect(healed).toBe("Robbie")` → `expect(healed).toBe("Ghostfixture")`

Leave on-disk fixtures (`"kai"`, `"lena"`, `"phantom"`) alone — they only land in the tmpdir personas tree, never in state.json.

### `tests/cli-run.test.ts`

Same principle. The heal test at L93 currently removes the `"phantom"` persona dir and creates `"kai"` instead, so on a scoping failure the broken-default value (`"phantom"`) lands in state.json. `"phantom"` is the built-in fallback name, which means future scoping bugs would write the *default default* and look indistinguishable from an untouched system.

Either:
- Change `defaultPersona: "phantom"` (L39) to a fake name (e.g. `"ghostfixture"`) and update L33, L74, L97 to mkdir/rm a real persona dir for that fake name, **or**
- Rename more narrowly: keep the harness setup using `"phantom"` for the cases where the persona dir exists, but in the heal test (L93–124) use a fake broken-default name.

Your call on which is cleaner — but the L93 heal test's broken-default value must be a name that screams "test fixture."

---

## Verification gate (same as last time)

Before pushing, run on your dev box:

```bash
cp ~/.local/share/phantombot/state.json /tmp/state.before.json
cd ~/Projects/phantombot
bun test tests/lib-personaDefault.test.ts tests/cli-run.test.ts
diff /tmp/state.before.json ~/.local/share/phantombot/state.json  # must be empty
bun test tests/config.test.ts  # must stay 7/7
```

Then check that nowhere in the final test output does the string `"robbie"` or `"phantom"` appear as a *broken-default fixture value* — only as on-disk fixtures or in unrelated tests.

Ping me when both changes are in and I'll re-review.

robertclawson (approved):
Approving — both pushback points landed cleanly.

**PR description rewritten** ✓ — Now correctly credits the real root cause (PR's own pre-`ca76155` tests writing to dev box's real `state.json` during `bun test`), with the 23:48:44 vs 23:48:55 UTC smoking-gun timestamps and the kw-openclaw evidence (only Kai's state mutated; Lena's + Jake's untouched). Heal is properly reframed as defense-in-depth, not the fix for the symptom.

**Fixture rename** ✓ — `"ghostfixture"` now used everywhere a default persona is intentionally broken (`lib-personaDefault.test.ts` lines 81, 95, 104, 111, 127; `cli-run.test.ts` line 107). `"phantom"` is kept only where the persona is a valid on-disk fixture (created via `mkdir` in `beforeEach`) — that's the correct distinction, not something to scrub.

**Verified locally:**
- CI: 5/5 green (typecheck, test, CodeQL, both Analyze jobs)
- Host state.json: didn't exist before tests, didn't exist after — tmpdir scoping is solid
- Cross-file pollution check: ran `lib-personaDefault.test.ts + cli-run.test.ts + config.test.ts` in the same `bun test` invocation → 19/19 pass
- Full suite: **833 pass, 1 skip, 0 fail** across 65 files
- Grep gate: `"robbie"` does not appear as a broken-default fixture anywhere in the two PR files

Branch is clean against main (4 commits, no PR #108 cruft). Ready to **squash-merge**.

Issue #112, opened 2026-05-16
Memory capture is the load-bearing step, but has no CLI, no metric, and no enforcement


14skywalker:
## TL;DR

Phantombot's other agent-facing domains (`env`, `notify`, `task`) are surfaced as small CLIs the harness calls via Bash, leaving traces phantombot can see. **Memory capture is the exception** — it's specified only as a prose protocol inside the system prompt, with no CLI, no log, and no counter. The rest of the memory pipeline (heartbeat → drawers → nightly → search) is well-built, but it all hinges on the persona reliably hand-writing tagged markdown into `memory/<today>.md`. When the persona doesn't, the pipeline runs as a silent no-op and there is no signal anywhere that something is missing.

This is filed as a **design observation**, not a bug report — see the "What I actually observed" caveat below.

## How this compares to the rest of phantombot's design

The README is explicit: *"phantombot does identity, memory, channel, scheduling, and self-update"*, and the defining principle is *"the harness can do its own tools — let it"* — no tool-call translation layer, the harness uses Bash.

The agent-facing CLI set follows that principle consistently:

| Domain | Agent-facing CLI | Leaves a trace? |
|---|---|---|
| Credentials | `phantombot env set / get / list` | yes (atomic file write) |
| Notifications | `phantombot notify --message / --voice` | yes (Telegram send result) |
| Scheduling | `phantombot task add / list / log / cancel` | yes (SQLite `task_runs`) |
| Memory capture | *(none — hand-write markdown into `memory/<today>.md`)* | **no** |

Memory is the only domain where the harness composes the storage format itself, with no command to call and no record that an attempt was even made. Heartbeat (`src/cli/heartbeat.ts:63`) logs `promoted 0, stale 0, indexed 0` whether the persona wrote zero tags or whether capture is fundamentally broken — the two are indistinguishable.

## Why the asymmetry matters

A few project-specific reasons the prose-only protocol is fragile in a way the other domains aren't:

- **Multi-harness reality.** Phantombot supports Pi, Claude, and Gemini (with Pi as the recommended primary). They have meaningfully different instruction-following profiles. A capture protocol that works on one will quietly degrade on another, and the project currently has no way to detect that regression.
- **No negative feedback.** Forgetting to call `phantombot env set` after the user shares a credential has a visible consequence next time the credential is needed. Forgetting to capture has no consequence the persona ever sees — so the behaviour doesn't self-correct.
[...]

PR #113, opened 2026-05-18
Reliable, observable memory capture & resumable nightly

robertclawson:
## Reliable, observable memory capture & a resumable nightly

Closes the diagnosis in #112. Investigation (incl. a three-persona control group on a separate box) found **three stacked failures merged into one "memory is broken" symptom**, not one systemic bug:

1. **Nightly idle-timeout.** The nightly is a deep, monolithic LLM pass. `claude timed out after 300000ms` is the **5-minute idle timeout** (`cli/nightly.ts` — 5 min idle, 30 min hard) firing when the model goes quiet mid-pass. On a box with a real day of chat it dies, so daily files never consolidate into the drawers. *(Corrected from an earlier comment that called it a hard cap — it is the idle cap.)*
2. **Capture never fired on the OpenRouter box** — the missing `today.md`. Capture is the only phantombot subsystem with no CLI: pure prose protocol, no log, no trace. A harness that under-weights standing system instructions can simply skip it, and nothing notices.
3. **Embedding key added mid-day** → `indexed 0`. Setup ordering; self-heals on reindex. Not a code bug.

The control group (Lena/Kai/Jake, v1.1.66) showed healthy nightlies and populated drawers — capture is **not** broken across the board; the reporter hit the specific combination of a busy box + a flaky harness.

### The four changes

**1. `phantombot memory capture` — give capture the same shape as the rest of the project**
`env`, `notify`, `task` all have a command, a log line, an exit code. Capture now does too: a Citty subcommand that appends a tagged, heartbeat-promotable line to today's daily file and records the call in a new `capture_log` table. Converts an *invisible* miss into a *visible, queryable* one. (Honest scope: a CLI cannot *force* a model to capture — that would mean dictating the harness's model.)

**2. Mechanical 30-turn nudge**
The Telegram dispatch counts user turns since the last `capture_log` row; when the effective turn index hits a multiple of 30 it stacks a fresh reminder onto the system-prompt suffix. Pure counter — no LLM decides whether to nudge. Directly counteracts the long-context dilution that broke the OpenRouter box. Gated to real `telegram:*` conversations.

**3. Checkpointed, resumable nightly**
The monolithic pass is decomposed into five idempotent stages (`essence → promote → kb → compress → state`), each its own harness turn. After every completed stage phantombot writes `.nightly-progress.json`. A stage timeout — or a mid-run power-off — costs at most one stage; `nightly --resume` skips the finished stages. Per-stage hard cap drops to 20 min since each stage is far smaller. Persona `nightly-prompt.md` overrides still run monolithic (the override owns the phase contract).

**4. `phantombot doctor` — detect & auto-repair**
Reads signals that already exist — `.nightly-state.json`, `.nightly-progress.json`, `capture_log` — and answers: did the last nightly run, did it succeed, is capture firing. When the nightly is stale/failed/partial it spawns a detached `nightly --resume`. `--no-repair` (report only) and `--json` (for schedulers). The startup catch-up in `run` now routes through doctor, so a *partial* checkpoint is repaired too — not just a fully-missed night.

### What this PR deliberately does NOT do

No transcript-fallback. The `turns` table is a ~120-row rolling per-conversation buffer (verified — IDs run to ~1956, ~1,800 already pruned); it cannot serve as a transcript archive. If write-time capture fails for a whole day and the daily file is empty, the nightly still has nothing to consolidate. We accept this consciously: change #1 makes such a day *visible*, #2 makes it far less likely. A durable transcript log is separate, larger work.

### Tests / verification

- `bun tsc --noEmit` clean; `bun test` — **859 pass, 0 fail**.
- New coverage: checkpoint progress round-trip, `pendingNightlyStages` resume logic, per-stage prompt builders, and `doctor` (no-record / fresh-ok / stale / partial-checkpoint / no-repair / json).
- ARM64 compile verified; `phantombot doctor --help` smoke-tested.

Opening as **draft** for a few-days soak on a high-volume box before merge.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,7 +61,8 @@ phantombot/
 │   │   ├── loader.ts         # accepts BOOT.md / SOUL.md / IDENTITY.md, MEMORY.md, tools.md / AGENTS.md
 │   │   └── builder.ts        # buildSystemPrompt + MEMORY_TOOLS_SECTION + CREDENTIALS_SECTION
 │   ├── memory/
-│   │   └── store.ts          # bun:sqlite turn store (turns table)
+│   │   └── store.ts          # bun:sqlite store: turns table + capture_log table
+│   │                         #   (capture_log backs the 30-turn nudge + doctor)
 │   ├── importer/
 │   │   └── openclaw.ts       # OpenClaw → phantombot persona import
 │   ├── orchestrator/
@@ -81,8 +82,9 @@ phantombot/
 │   │   ├── voice.ts          # phantombot voice (TUI: TTS/STT provider config)
 │   │   ├── telegram.ts       # phantombot telegram (TUI: token + allowed users)
 │   │   ├── harness.ts        # phantombot harness (TUI: chain) + maybePromptRestart helper
-│   │   ├── memory.ts         # phantombot memory (search/get/today/index/list)
-│   │   ├── heartbeat.ts nightly.ts
+│   │   ├── memory.ts         # phantombot memory (search/get/today/index/list/capture)
+│   │   ├── heartbeat.ts nightly.ts  # nightly is 5 checkpointed stages (--resume)
+│   │   ├── doctor.ts         # phantombot doctor (memory health check + nightly auto-repair)
 │   │   ├── embedding.ts      # phantombot embedding (TUI: Gemini config)
 │   │   ├── persona.ts        # phantombot persona (consolidates create / import / restore / switch)
 │   │   ├── create-persona.ts import-persona.ts  # implementation files; no top-level subcommand
--- a/README.md
+++ b/README.md
@@ -174,13 +174,15 @@ mkdir -p ~/.local/bin && cp dist/phantombot ~/.local/bin/
 | `phantombot task list / show <id> / cancel <id>` | Manage tasks |
 | `phantombot tick` | Fire any due tasks (called every minute by `phantombot-tick.timer`) |
 | `phantombot memory today / search / get / list / index` | Read/write the persona's memory + KB |
+| `phantombot memory capture "<text>" --tag <tag>` | Append a tagged line to today's daily file (`decision` / `lesson` / `person` / `commitment`; `--tag` repeatable) and record it in `capture_log` |
 
 ### Periodic maintenance (called by systemd timers, occasionally by hand)
 
 | Command | What it does |
 |---|---|
 | `phantombot heartbeat` | Mechanical 30-min pass (no LLM) |
-| `phantombot nightly` | Cognitive distillation pass (LLM) |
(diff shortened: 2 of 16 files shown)


andrewagrahamhodges (dismissed):
## Code Review: Reliable, observable memory capture & a resumable nightly

**Author:** @robertclawson | **Files:** 16 | **Verdict:** Approve

### What's Good
- **Direct handling of #112**: You precisely implemented the community feedback from the `memory capture` issue.
- **Architectural consistency**: The new `memory capture` subcommand aligns with existing `.citty` subcommands smoothly.
- **Idempotency**: Checkpointing the nightly runs makes the process fault-tolerant without relying purely on brute force restarts.
- **Observable design**: Incorporating `capture_log` and using it for dynamic metrics gives real visibility. 

I've reviewed the diff context around `src/cli/memory.ts`, the logic for normalizing `--tag` arguments, and appending the log entries cleanly. Looks solid and ready for testing!

andrewagrahamhodges (approved):
## Code Review: Reliable, observable memory capture & a resumable nightly

**Verdict:** Approve

### Re-Review
The bugfix for the `repair/no-repair` boolean has correctly adjusted the arg definition to bypass the Citty double-negation trap. Looks good to go!

kaieriksen (approved):
Clean PR. Thorough diagnosis in #112, solid fixes.

**What I checked:**
- `bun tsc --noEmit` — clean
- `bun test` — 859 pass, 0 fail, 1 skip (Pi ARG_MAX probe)
- ARM64 binary smoke-tested per PR body

**Code review notes:**

1. **Capture CLI** — Giving capture the same shape as `env`, `notify`, `task` is the right call. The `capture_log` table makes misses observable. Honest about the scope limitation (can't force a harness).

2. **30-turn nudge** — Pure mechanical counter, no LLM in the loop. Gated to `telegram:*` conversations. Error path is a warn-and-continue. Correctly accounts for the +1 offset (message not yet in `turns`).

3. **Checkpointed nightly** — Five idempotent stages, each a bounded turn. `--resume` skips completed stages. Stale checkpoints from a different date are correctly ignored. The override-path fallback to monolithic is clean — if a persona ships `nightly-prompt.md`, you can't safely split it.

4. **Doctor** — Reads existing signals, no LLM in-process, repair is a detached child. The `--no-repair` and `--json` modes are well-tested. Dry-day threshold (≥20 real user turns, 0 captures) is sensible.

5. **Test coverage** — 163 lines of new doctor tests, 105 lines of new nightly tests covering every checkpoint edge case (no file, round-trip, clear, stale date). Good.

**Minor observation (not blocking):** Two `buildNightlyStagePrompt` tests use `"robbie"` as the persona name string. These are pure prompt-building functions (string-in, string-out, no state touch), so it's cosmetic — no pollution risk. Just noting it since we just cleaned up fixture names in #109.

PR #116, opened 2026-05-20
fix(update,doctor): self-heal systemd units after binary swap

robertclawson:
## Summary

After a successful `phantombot update`, on-disk systemd unit files used to be left exactly as the previous version installed them. If a release renamed or moved a unit, OR a host's `~/.config/systemd/user/timers.target.wants/` symlinks had rotted from a previous bad update, the timers reported as enabled but never fired — silently stranding every scheduled task.

## The bug in production

On hz-phantombot in May 2026, `phantombot task list` showed two tasks with `runs=0` and next-run timestamps two days in the past. `systemctl --user list-timers phantombot-*` returned the units but with no `NEXT` line. The `timers.target.wants/` directory had symlinks pointing at unit files that no longer existed on disk. Re-running `phantombot install` fixed it in one shot — which means the same step belongs in `update`.

## What this PR does

**Three small changes, none destructive:**

1. **New `ensureSystemdUnitsCurrent` helper in `src/lib/systemd.ts`**
   - Compares every expected unit file against the canonical template
   - Rewrites any that are missing or differ (with `.bak` of the original)
   - Runs `daemon-reload` exactly once (only if something changed)
   - Re-enables any timer that is not currently active via `enable --now`
   - Idempotent: a healthy box gets zero writes and zero state changes, just six is-enabled/is-active probes

2. **`runUpdate` calls it after the atomic binary swap**
   - Runs between the swap and the restart step
   - Failure is non-fatal: binary swap is the critical op, and `phantombot install` remains a manual fallback. Matches the existing `restart failed → warn, exit 0` contract.
   - Linux-only (launchd hosts skip)

3. **`runDoctor` gains a systemd-health section**
   - Lists missing unit files and inactive timers
   - Auto-repairs in default `--repair` mode by calling `ensureSystemdUnitsCurrent`
   - Reports outcome in both human summary and JSON envelope
   - Exits 1 when systemd is broken AND was not repaired (so cron callers can alert)

## Test seams

- `RunUpdateInput.healSystemdUnits?: false | fn` — existing tests opt out with `false`; new tests inject a fake to verify the wiring without touching real systemctl
- `RunDoctorInput.checkSystemd?: false | fn` — same pattern for the doctor side
- A new `defaultHealUnits` / `defaultCheckSystemd` wire up the production paths via `BunSystemctlRunner` + `ensureUserSystemdEnv`

## Test plan

- [x] `bun test` — 872 pass / 1 skip / 0 fail (was 815 before this PR)
- [x] `bun run build` — compiles cleanly
- [x] New tests cover:
  - `ensureSystemdUnitsCurrent`: no-op when current, full rewrite when missing, backup of hand-edits, single daemon-reload, re-arm of an inactive timer
  - `runUpdate`: heal called with swapped binPath, output mentions rewrites + re-armed timers, heal failure is non-fatal, darwin skips heal
  - `runDoctor`: healthy section, broken-but-unrepaired (exit 1), broken-and-repaired (exit 0), `checkSystemd: false` omits section, JSON envelope includes the block
- [ ] On a real box: trigger update on a host with a broken `timers.target.wants/` symlink, observe heal log + `phantombot task list` showing future NEXT

## Why no `ExecStartPost=phantombot doctor` in the unit

Considered, rejected. `doctor --repair` calls into systemctl, which would interact with the service that just started — restart-loop risk for marginal benefit. The two healers we have (update-time and explicit doctor invocation) cover the realistic recovery paths.

### DIFF
--- a/src/lib/systemd.ts
+++ b/src/lib/systemd.ts
@@ -416,6 +416,145 @@ export function defaultSystemdServiceControl(): ServiceControl {
   };
 }
 
+/**
+ * Surgical re-render of all six phantombot systemd unit files, plus a
+ * re-enable / re-start of any timer that isn't currently active.
+ *
+ * This is the post-update healer: a `phantombot update` swaps the
+ * binary, then calls this to make sure the on-disk unit files still
+ * match the new version's templates AND that the timers are actually
+ * armed. Without it, a release that renames or moves a unit (or that
+ * lands on a host whose `~/.config/systemd/user/timers.target.wants/`
+ * symlinks have rotted from a previous bad update) leaves the user
+ * with timers that look enabled but never fire — exactly the bug that
+ * stranded ~2 days of scheduled tasks on hz-phantombot in 2026-05.
+ *
+ * Pure on the inputs: caller picks the paths and the systemctl
+ * runner. Tests inject a FakeSystemctl and a tmpdir; production calls
+ * this with real ones via `runHealSystemdUnits` below.
+ *
+ * Idempotent: writing the same content is a no-op; an already-enabled
+ * + active timer is left alone. Only changes get logged.
+ */
+export interface EnsureUnitsCurrentOptions {
+  binPath: string;
+  unitPath?: string;
+  heartbeatServicePath?: string;
+  heartbeatTimerPath?: string;
+  nightlyServicePath?: string;
+  nightlyTimerPath?: string;
+  tickServicePath?: string;
+  tickTimerPath?: string;
+  systemctl: SystemctlRunner;
+}
+
+export interface EnsureUnitsCurrentResult {
+  /** Basenames of unit files that were (re)written. Empty = nothing changed. */
+  rewrote: string[];
+  /** Backups created for any unit file whose previous content differed. */
+  backups: string[];
+  /** Timer unit names that were re-enabled and/or restarted to repair them. */
+  repairedTimers: string[];
+}
+
+export async function ensureSystemdUnitsCurrent(
+  opts: EnsureUnitsCurrentOptions,
+): Promise<EnsureUnitsCurrentResult> {
+  const targets: Array<{
+    path: string;
+    content: string;
+    unit: string;
+    isTimer: boolean;
+  }> = [
+    {
+      path: opts.unitPath ?? defaultUnitPath(),
+      content: generateSystemdUnit({ binPath: opts.binPath, args: ["run"] }),
+      unit: PHANTOMBOT_UNIT_NAME,
+      isTimer: false,
+    },
+    {
+      path: opts.heartbeatServicePath ?? heartbeatServicePath(),
+      content: generateHeartbeatService(opts.binPath),
+      unit: HEARTBEAT_SERVICE_NAME,
+      isTimer: false,
+    },
+    {
+      path: opts.heartbeatTimerPath ?? heartbeatTimerPath(),
+      content: generateHeartbeatTimer(),
+      unit: HEARTBEAT_TIMER_NAME,
+      isTimer: true,
+    },
+    {
+      path: opts.nightlyServicePath ?? nightlyServicePath(),
+      content: generateNightlyService(opts.binPath),
+      unit: NIGHTLY_SERVICE_NAME,
+      isTimer: false,
+    },
+    {
+      path: opts.nightlyTimerPath ?? nightlyTimerPath(),
+      content: generateNightlyTimer(),
+      unit: NIGHTLY_TIMER_NAME,
+      isTimer: true,
+    },
+    {
+      path: opts.tickServicePath ?? tickServicePath(),
+      content: generateTickService(opts.binPath),
+      unit: TICK_SERVICE_NAME,
+      isTimer: false,
+    },
+    {
+      path: opts.tickTimerPath ?? tickTimerPath(),
+      content: generateTickTimer(),
+      unit: TICK_TIMER_NAME,
+      isTimer: true,
+    },
+  ];
+
+  const rewrote: string[] = [];
+  const backups: string[] = [];
+  for (const t of targets) {
+    let current: string | undefined;
+    if (existsSync(t.path)) {
+      current = await readFile(t.path, "utf8");
+    }
+    if (current === t.content) continue;
+    await mkdir(dirname(t.path), { recursive: true });
+    if (current !== undefined) {
+      const bak = `${t.path}.bak`;
+      await writeFile(bak, current, "utf8");
+      backups.push(bak);
+    }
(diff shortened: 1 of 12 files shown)


robertclawson:
Amended per Andrew's request to cover the long-uptime case (months without restart or update).

**New commit `b23970c`**:

1. **`phantombot heartbeat` now self-heals systemd on its 30m cadence.** After the primary promotion/staleness/index work, it calls `ensureSystemdUnitsCurrent` — idempotent, six is-active/is-enabled probes on a healthy box, log-and-continue if systemctl errors. No new daemon timer; the heartbeat was already the right home for mechanical maintenance.

2. **`phantombot doctor` now flags dead timers.** New `timers` section in the report reads on-disk fire-markers written by heartbeat + tick. Heartbeat stale > 120m, tick stale > 5m. Stale or never-fired → exit 1.

3. **New `src/lib/timerHealth.ts`** keeps the marker contract in one place: cheap atomic writes to `$XDG_STATE_HOME/phantombot/{heartbeat,tick}.last-fired`, body carries ISO + runs counter, read path falls back to mtime if the body is corrupt.

**Why heartbeat-based, not setInterval:**
- It IS the mechanical maintenance pass — adding a second timer would just be more state to sync.
- The one wrinkle (heartbeat itself being the dead timer) is exactly what the doctor staleness check catches.
- Anything that survives a fresh `phantombot install` re-arms heartbeat → which then re-arms everything else on next fire.

**Test coverage:**
- 889 pass / 0 fail (was 872)
- typecheck + build clean
- New unit tests for `timerHealth` (marker write/read, counter, corrupt-body fallback)
- New doctor tests for the `timers` section (fresh / stale / never-fired / json / opt-out)
- New heartbeat tests for the marker write + `healSystemd` seam
- New tick test for the marker write even on lockfile-skipped ticks
- Existing heartbeat + tick tests now redirect `XDG_STATE_HOME` so markers don't leak into the developer's real `~/.local/state/`

src/lib/systemd.ts
```diff
@@ -416,6 +416,145 @@ export function defaultSystemdServiceControl(): ServiceControl {
...
+export interface EnsureUnitsCurrentOptions {
+  binPath: string;
+  unitPath?: string;
+  heartbeatServicePath?: string;
+  heartbeatTimerPath?: string;
+  nightlyServicePath?: string;
+  nightlyTimerPath?: string;
+  tickServicePath?: string;
+  tickTimerPath?: string;
+  systemctl: SystemctlRunner;
```
andrewagrahamhodges: 🟡 **Minor**: This function recreates the unit files similar to `installPhantombotUnit`. If templates change in the future, developers must remember to update both places. It might be worth extracting the shared rendering logic.

src/lib/timerHealth.ts
```diff
@@ -0,0 +1,156 @@
...
+export function recordHeartbeatFired(): Promise<void> {
+  return recordFired(heartbeatMarkerPath());
+}
+
+export function recordTickFired(): Promise<void> {
+  return recordFired(tickMarkerPath());
+}
+
+export interface TimerLastFired {
+  /** ISO timestamp of the last recorded fire, or undefined if no marker. */
```
andrewagrahamhodges, line 109: 🔵 **Trivial**: This writes `new Date().toISOString()`, which depends on system time. While coarse enough that NTP jumps might not matter, consider noting this limitation or using a monotonic time if precise age thresholding becomes critical.

andrewagrahamhodges (commented):
## Code Review: Systemd Timers Self-Heal and Staleness Checks

**Author:** @robertclawson | **Files:** 8 | **Verdict:** Comment Only

### Findings

| 1 | 🟡 Minor | src/lib/systemd.ts | 448 | The `installPhantombotUnit` and `ensureSystemdUnitsCurrent` functions have overlapping logic for unit file re-rendering. This is a potential maintenance risk if templates change. |
| 2 | 🔵 Trivial | src/lib/timerHealth.ts | 100 | The `recordHeartbeatFired` and `recordTickFired` functions use `Date.prototype.toISOString()` which relies on system time. Consider if a monotonic clock approach would be safer against NTP jumps, though this is likely fine for coarse age checks. |

### What's Good
- Robust approach to handling rotted timer symlinks.
- The test seam injections for checking systemctl behavior without full execution are well structured.
- Fallback to `mtime` if the marker contents are corrupted is a great touch for resilience.

<details>

```
Review and fix the following issues found in PR #116:

1. [Minor] src/lib/systemd.ts:448 - Overlap in unit rendering logic with `installPhantombotUnit`. Consider extracting the shared templating logic if possible.
2. [Trivial] src/lib/timerHealth.ts:100 - Using `Date.now()`/`toISOString()` for age calculations can be susceptible to NTP adjustments. Depending on the environment, a monotonic clock might be more robust, although likely not critical here given the large tolerances.

Base: main
```
</details>

robertclawson:
Addressed all three review findings — commit `45ed457`:

**🟡 Minor — systemd.ts:448 (overlap with `installPhantombotUnit`):**
Extracted `phantombotUnitTargets(binPath, overrides)` as the single source of truth for the (path, body, unit, isTimer) tuples of all 7 phantombot unit files. Both `installPhantombotUnit` and `ensureSystemdUnitsCurrent` now build their work list from it — template changes can no longer drift between the two. Net diff: systemd.ts gained the new helper and shrank `installPhantombotUnit`'s body by ~25 lines.

**🔴 CodeQL — tests/cli-update.test.ts:504 (incomplete URL substring sanitization):**
Replaced all three `u.includes("api.github.com")` calls in the fake-fetch dispatcher with `isGitHubApiUrl(u)`, which does an exact `new URL(u).hostname === "api.github.com"` match. Even though this is test-side and the inputs are dispatcher-controlled, the precise form costs nothing and clears the alert.

**🔵 Trivial — timerHealth.ts:100 (wall-clock dependency):**
Added a JSDoc note on `recordFired` documenting (a) that the ISO timestamp comes from wall-clock time, (b) that the staleness bars (120m / 5m) are deliberately coarse enough to tolerate NTP-step-class drift, and (c) the upgrade path (monotonic `hrtime.bigint()` for the delta, keep wall-clock ISO only for forensic logs) if anyone ever tightens the thresholds.

**Verification:** 889 pass / 0 fail (1 pre-existing skip), `bun run typecheck` clean, `bun run build` clean. Updated one test assertion to match the new per-unit log phrasing ("wrote phantombot.service: …" instead of the legacy "wrote unit file: …").

All three review threads resolved. Ready for re-review.

andrewagrahamhodges (approved):
Looks like everything has been resolved. The deduplication looks great, and the comments added around the `.toISOString` timing answer the question completely without adding unnecessary complexity.

PR #117, opened 2026-05-20
fix(tick): quiet by default — never auto-post harness reply to Telegram

robertclawson:
## The bug

The persona builder promises a clear contract:

> Scheduled tasks (`phantombot tick`) run silently by default — no Telegram chatter on every fire. When something material happens, surface it explicitly with `phantombot notify`.

Tick was doing the exact opposite. Any non-review, non-`silent`, non-error task with non-empty output had its full harness reply formatted as `📋 *<description>*\n\n<text>` and sent to every `allowed_user_ids` chat in Telegram. `silent` defaulted to `false` in the store, so **every task ever scheduled** has been getting its complete LLM transcript blasted to the user — including replies the agent itself wrote believing it had stayed quiet ("Andrew not notified — nothing material" followed by the entire payload arriving on his phone anyway).

Reproduced by Andrew on task #29 today (image attached on Telegram).

## The fix — option A

This is the **remove auto-delivery entirely** option from the two we discussed:

- **A. Remove auto-delivery.** Tick is silent. The harnessed agent is the sole arbiter of notifications, via `phantombot notify` from inside the prompt — same mechanism it already uses for ad-hoc alerts. Cleanest, matches the prompt verbatim. ← THIS PR
- B. Flip default + keep `--silent`/`--notify-output` opt-in. Rejected: two mechanisms with overlapping semantics, easy to drift from intent at scheduling-time vs fire-time.

The "friendly escape hatch" of B is illusory — for a "scheduled message" task you just put the message in the prompt and tell it to call notify. Same tokens, no flag, no surprises.

## What changed

**`src/cli/tick.ts`**
- Drops the `HttpTelegramTransport` import, the transport init block, the `transport?` test seam, and the entire auto-notify branch.
- `formatTaskNotify` removed (no callers).
- `delivered` on the task_runs row is always `false` now. The field is kept on the schema for back-compat with historical rows and as a slot if we ever wire up post-hoc "did the agent call notify during this run" tracking.
- Header comment + JSDoc updated to spell out the quiet-by-default contract so the next reader can't miss it.

**`src/cli/task.ts`**
- `--silent` is a no-op now. Accepted for back-compat with a loud stderr deprecation warning so callers stop scheduling around it.
- Removed from `task add` output and from `task show` (`formatTaskFull`). Help text marks it DEPRECATED.

**`src/lib/tasks.ts`**
- `Task.silent`, `TaskAddInput.silent`, and `TaskRunRow.delivered` gained @deprecated / historical JSDoc explaining the field is preserved for migration safety, not for current behaviour.
- DB column stays. Migration to drop it lands in a follow-up PR.

**`tests/cli-tick.test.ts`**
- New `runTick — quiet-by-default` describe block installs a global fetch trap, configures a Telegram channel that *would* have triggered the old auto-delivery, fires a task, and asserts:
  - zero calls to `api.telegram.org`
  - `delivered = false` on the run row
- A second test confirms legacy `silent: true` rows are equally silent — the field has no effect either way.

## Verification

- `bun test` → **891 pass / 0 fail / 1 skip** (was 889; +2 quiet-by-default tests)
- `bun run typecheck` — clean
- `bun run build` — clean
- No remaining references to `formatTaskNotify` or the auto-delivery pattern in `src/`.

## Follow-up (separate PR, not blocking)

Drop `--silent` from the CLI surface, drop the `silent` column from `tasks`, and drop the `delivered` column from `task_runs` once enough time has passed that no on-disk DBs in the wild still reference them.

## What I deliberately did NOT do

- **Did not** keep `--silent` as a functional opt-in. The premise of the bug is that two mechanisms (auto-send + manual notify) is one too many; preserving one as opt-out reintroduces the same drift surface.
- **Did not** drop the DB columns. Existing DBs on Andrew's boxes have rows with `silent` and `delivered` set; ripping the columns would require a migration that's risky for zero behavioural payoff right now.

### DIFF
--- a/src/cli/task.ts
+++ b/src/cli/task.ts
@@ -52,7 +52,13 @@ export interface RunTaskAddInput {
   count?: number;
   /** --for 30d — recurring expiry (relative) */
   relFor?: string;
-  /** --silent — suppress user notification after fire */
+  /**
+   * @deprecated `--silent` is a no-op as of the quiet-by-default fix.
+   * Tick never auto-posts the harness reply to Telegram anymore; the
+   * agent decides whether to call `phantombot notify` from inside the
+   * prompt. Accepted for back-compat (with a stderr warning) until a
+   * follow-up PR drops the flag and the DB column.
+   */
   silent?: boolean;
   /** --force-long-running — allow recurring > 90d */
   forceLongRunning?: boolean;
@@ -69,6 +75,18 @@ export async function runTaskAdd(input: RunTaskAddInput): Promise<number> {
   const store = input.store ?? (await openTaskStore(config.memoryDbPath));
   try {
     const now = new Date();
+
+    // `--silent` is now a no-op (see RunTaskAddInput JSDoc). Warn loudly
+    // so the agent / caller stops scheduling around it before it goes
+    // away entirely.
+    if (input.silent) {
+      err.write(
+        "warning: --silent is deprecated and now a no-op. " +
+        "Scheduled task fires are silent by default; call " +
+        "`phantombot notify` from inside the task prompt if you " +
+        "want the user notified.\n",
+      );
+    }
 
     // Determine mode: one-off or recurring.
     const isRecurring = Boolean(input.every);
@@ -192,7 +210,6 @@ export async function runTaskAdd(input: RunTaskAddInput): Promise<number> {
       (t.oneOff
         ? `  type:        one-off\n`
         : `  schedule:    ${t.schedule}\n  expiry:      ${describeExpiry(t)}\n`) +
-      (t.silent ? `  silent:      yes\n` : "") +
       (!t.oneOff && !hasExpiry
         ? `  hygiene:     no expiry — every fire will ask if it's still needed.\n` +
           `               cancel with: phantombot task cancel ${t.id}\n`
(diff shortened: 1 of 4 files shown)


andrewagrahamhodges (commented):
## Code Review: fix(tick): quiet by default — never auto-post harness reply to Telegram

**Author:** @robertclawson | **Files:** 4 | **Verdict:** Approve

### Findings

*(None — the logic looks solid and fully resolves the bug)*

### What's Good
- Opting for "Option A" (removing auto-delivery entirely) is the correct architectural choice here. It forces the system to align strictly with the prompt contract and removes the overlapping semantics of `--silent` vs `phantombot notify`.
- Preserving the database fields while clearly marking them as deprecated in JSDoc/comments is a safe, defensive way to handle the backend without forcing a risky migration right now.
- `tests/cli-tick.test.ts` adding a fetch trap to strictly assert *zero* calls to `api.telegram.org` is a very robust way to guarantee this behavior won't silently regress in the future.

<details>

```
Review of PR #117 complete. No fixes needed.
```
</details>

andrewagrahamhodges (dismissed):
LGTM

andrewagrahamhodges (approved):
LGTM, re-approving.

PR #120, opened 2026-05-21
fix(update): fallback to unauthenticated request on 401/403 (#115)

### LINKED ISSUE #115: `phantombot update` fails when GITHUB_TOKEN is a GitHub App installation token
## Problem

`phantombot update` sends `$GITHUB_TOKEN` unconditionally in the `Authorization` header when checking the latest release on `phantomyard/phantombot`. If the user's token is a **GitHub App installation token** (e.g. generated from a PEM key for use in an organization like `iasgair-nl`), that token is scoped to repositories within that organization. Accessing a public repo under a *different* account (`phantomyard/phantombot`) then returns **401 Unauthorized**, even though the repo itself is public.

## Workaround confirmed

```bash
GITHUB_TOKEN="" phantombot update
```

This works because the empty token skips the auth header, falling back to unauthenticated public access.

## Suggested fixes

1. **Skip auth for public repos** — if the release endpoint is reachable without auth (which it is for `phantomyard/phantombot`), don't send `$GITHUB_TOKEN` at all.
2. **Try-unauth fallback** — if the authenticated request gets 401/403, retry without the token before failing.
3. **Config flag** — allow overriding the token used for updates (e.g. `phantombot update --token none`).

## Environment

- phantombot version: 1.0.109
- OS: Ubuntu 22.04 / Debian 12 (reproducible anywhere)
- Token type: GitHub App installation token (org-scoped)

14skywalker:
## Summary

Fixes #115. When `GITHUB_TOKEN` is set to a token that's valid but not authorized for `phantomyard/phantombot` (most commonly a GitHub App installation token scoped to a different org), `phantombot update` returns HTTP 401 and bails out — even though the releases endpoint is public and would have worked unauthenticated.

`findLatestRelease` now transparently retries once without the `authorization` header when an authenticated request returns 401 or 403. Authenticated-first behavior is preserved (tokens that work keep their 5000/h cap), no new flags or env vars, and error messages are updated so a failure after the unauth retry is distinguishable from "no token at all".

- `src/lib/githubReleases.ts` — extracted `buildHeaders(withAuth)`; on 401/403 with a token, retry unauth; refined error strings.
- `tests/lib-githubReleases.test.ts` — covers 401-with-token → fallback succeeds, 401-without-token → error, 403-with-token → fallback succeeds, 403-without-token → existing rate-limit hint, and the "both attempts fail" path.

## Test plan
- [x] `bun test tests/lib-githubReleases.test.ts`
- [x] Manual: `GITHUB_TOKEN=<scoped-app-token> phantombot update` succeeds where it previously 401'd
- [x] Manual: `GITHUB_TOKEN="" phantombot update` still works (the existing workaround path)


kaieriksen (approved):
Approved. The fix preserves authenticated-first behavior, retries unauthenticated only on 401/403 when a token was present, and covers the regression with focused tests.\n\nValidation run locally:\n- `bun test tests/lib-githubReleases.test.ts` -> 16 passed\n- `bunx tsc --noEmit` -> passed

src/lib/githubReleases.ts
```diff
@@ -61,28 +64,49 @@ export async function findLatestRelease(opts: {
...
       error: `network error reaching ${url}: ${(e as Error).message}`,
     };
   }
+
+  // GitHub App installation tokens are scoped to a single org's repos.
+  // Using one against a public repo outside that org returns 401 — the
+  // token is "valid" but not authorized for this resource. The release
+  // endpoint is public, so an unauthenticated retry usually succeeds.
+  // We also retry on 403 in case the token itself is rate-limited or
+  // suspended while the unauth pool still has budget. See issue #115.
```
andrewagrahamhodges, line 84: 🟡 **Minor**: The try-catch block here is perfect for preventing network/DNS failures on the unauthenticated retry from crashing the update. Letting it fall through to report the original authentication/authorization status is a solid and clean pattern.

andrewagrahamhodges (commented):
## Code Review: fix(update): fallback to unauthenticated request on 401/403

**Author:** @14skywalker | **Files:** 4 | **Verdict:** Comment Only

### Findings

| 1 | 🟡 **Minor** | `src/lib/githubReleases.ts` | 84 | The `retriedUnauth` state checks if `res.status === 401` or `res.status === 403`. However, if the fallback `fetchImpl` call throws a network error, `res` remains in its error state and falls through. Having this try-catch is excellent for robust handling. |
| 2 | 🔵 **Trivial** | `src/lib/githubReleases.ts` | 98 | Clarification: If the retry results in a 403, indicating unauthenticated rate limits have also been exhausted, the user gets a specific rate-limit error. This is a very helpful refinement over standard 403 messages! |

### What's Good
- **Excellent Problem Solving:** The fallback to unauthenticated queries cleanly addresses the issue of organization-scoped GitHub App tokens hitting 401s on third-party public repositories.
- **Robust Documentation:** Updating `AGENTS.md` and `README.md` to communicate how `GITHUB_TOKEN` behavior works with org-scoping is library-grade discipline.
- **Uncompromising Test Coverage:** Outstanding comprehensive test cases covering different mock auth transitions (401/403 with/without token fallback) verify the implementation exhaustively, specifically ensuring no regression for users who run updates without a token at all.

<details>

```
Review and fix the following minor issue found in PR #120 (fix(update): fallback to unauthenticated request on 401/403):

1. [Minor] src/lib/githubReleases.ts:84 — If the fallback fetchImpl call throws a network error, the catch block lets the original 401/403 Response flow down to the error logic seamlessly. No code changes are required here, but adding a light inline comment about why we let catch fall through could enhance future readability.

Branch: fix/issue-115-github-auth-fallback
Base: main
```
</details>

src/lib/githubReleases.ts
```diff
@@ -61,28 +64,49 @@ export async function findLatestRelease(opts: {
...
+  let retriedUnauth = false;
+  if (hadToken && (res.status === 401 || res.status === 403)) {
```
robertclawson, line 86: Non-blocking: retrying on 403 swaps the authenticated 5,000/h budget for the 60/h unauthenticated IP pool. That's the right call for a one-shot update check (better to try than to fail outright), just worth noting it's a deliberate downgrade rather than a free retry.

src/lib/githubReleases.ts
```diff
@@ -61,28 +64,49 @@ export async function findLatestRelease(opts: {
...
+    try {
+      res = await fetchImpl(url, { headers: buildHeaders(false) });
+      retriedUnauth = true;
+    } catch {
+      // Network error on the retry — fall through and report the
+      // original status below.
```
robertclawson, line 92: Non-blocking: if the unauth retry itself throws here, `retriedUnauth` stays `false`, so we fall through to the 401 "is GITHUB_TOKEN scoped to a different org?" message — slightly misleading since a retry *was* attempted (it just hit a network error). Your comment already flags the fall-through, so this is purely cosmetic; could special-case the network-error-on-retry path if you wanted, but definitely not worth blocking on.

robertclawson (approved):
Clean, well-scoped fix — thanks @14skywalker! 🙌 The `buildHeaders(withAuth)` extraction is tidy, the unauth retry is correctly gated on a token actually being present, and the refined error strings keep the post-retry failure modes distinguishable. Test coverage is genuinely thorough (401→fallback, 403→fallback, both-fail, no-token single-call, success-no-retry) with proper env save/restore so there's no cross-test leakage, and the docs are updated per contributing discipline. LGTM — approving. Two small non-blocking notes inline. 👍

PR #121, opened 2026-05-22
fix(doctor,systemd): auto-recover "active (elapsed)" zombie timers

robertclawson:
## Problem

PR #116 closed the **orphaned-unit-file** failure mode (`update` re-stitches units; `doctor --repair` rewrites missing files + re-enables inactive timers). But one gap remained that still required a human:

A user timer can sit in **`active (elapsed)`** with `NextElapse: n/a` — systemd's `is-active` returns `"active"`, but the timer has stopped scheduling future fires and is silently dead. This is exactly the *"heartbeat hasn't fired in N days"* report:

- the timer is **enabled + active**, so neither the missing-file check nor the inactive-timer check sees a problem;
- only the **last-fired marker** (`timerHealth.ts`) catches it — and that signal was **detection-only**.

So `doctor` would `WARN` + exit 1 but never repair. Recovery meant a human running `phantombot install`. Not rock solid.

## Fix

1. **`ensureSystemdUnitsCurrent` gains `forceRearmTimers`** — timer units that must be re-armed *even when systemd reports them enabled + active*. For those it issues `systemctl --user restart <timer>` (not `enable --now`, which is a no-op on an already-active timer) so systemd recomputes the next elapse. With `OnBootSec`/`OnUnitActiveSec` already in the past, the restart fires a catch-up almost immediately — which also refreshes the last-fired marker.

2. **`doctor`** computes the timer-marker report *before* the systemd check, maps any *"fired before, then went stale"* marker to its timer unit, and passes that set into the heal step. A **never-fired** marker (no `last_fired` = fresh install) is deliberately excluded — the missing-file / inactive-timer checks already own that case.

Because `run`'s startup calls `doctor` with repair on, **and every `update` restarts the service**, a zombie timer is now auto-recovered on the next restart with **zero human involvement**.

### Note on exit code
The run that *detects* a zombie still exits 1 (the marker is genuinely stale until the catch-up fire lands a fresh one) — intentional, for visibility. The next `doctor` run is clean.

## Tests
- `lib-systemd`: force-re-arm via `restart`; precedence vs the inactive `enable --now` path.
- `cli-doctor`: stale marker → re-arm wiring; stale-tick → tick re-armed; never-fired marker is **not** re-armed.
- Full suite green: **911 pass, 0 fail**. Typecheck clean.

PR #122, opened 2026-05-22
fix(doctor,systemd): detect + heal systemd unit content drift on existing boxes

robertclawson:
## Why

PR #121 made the **heartbeat timer** ship as `OnCalendar=*:0/30` — but that only fixes *new installs*. Existing boxes keep the old, wedge-prone `OnUnitActiveSec=30min` timer through any number of `phantombot update`s. This is the "fix broken phantoms across the fleet" follow-up.

### Corrected root cause (my earlier diagnosis was wrong)

I previously said "`update` only backfills *missing* unit files." That's **not true** — `phantombot update` already calls `ensureSystemdUnitsCurrent`, which compares on-disk unit *content* against the new binary's templates and rewrites drift. So does the heartbeat (every 30 min) and doctor's repair path.

The real residual gaps:

1. **The update-time heal silently no-ops** when the user-systemd bus isn't reachable — `defaultHealUnits` returns `null`. That's exactly the **self-update / deferred-restart** context this box uses, so v1.1.74 swapped the binary but never rewrote the timer.
2. **`doctor` was blind to content drift.** `defaultCheckSystemd` only checked `missing_unit_files` + `inactive_timers`. A unit that *exists* and is *"active"* but whose content is the old template was invisible — doctor cheerfully reported `systemd: ok` while the stale, wedge-prone timer sat right there.

### Why fixing #2 cures both

`doctor` runs at **every service startup** (`run.ts`) and **every heartbeat** — both contexts where the systemd bus *is* live. Once doctor detects content drift → triggers the (already-existing) heal → every existing broken phantom self-heals on its next restart or heartbeat, regardless of how it got updated.

## What

- **`systemd.ts`** — new pure detector `driftedUnitNames(targets, reader)`: names existing unit files whose on-disk content no longer matches the binary's templates. Missing files are *not* drift (reported separately). Pure → unit-tests without touching the fs.
- **`systemd.ts`** — `ensureSystemdUnitsCurrent` now **restarts any timer whose content it just rewrote**. A `daemon-reload` alone can't un-wedge an `active (elapsed)` timer whose schedule changed (`OnUnitActiveSec → OnCalendar`); only a restart re-evaluates and arms the new definition. Without this, doctor would rewrite the file but the box would stay on the dead schedule until a manual kick.
- **`doctor.ts`** — `defaultCheckSystemd` computes `drifted_unit_files`, adds drift to the heal trigger, and reports it. Drift is now a `WARN` + non-zero exit in read-only mode, and a "re-rendered units" success once healed.

## Tests

- `driftedUnitNames`: matches → `[]`; one drifted unit flagged; missing file ≠ drift.
- `ensureSystemdUnitsCurrent`: rewriting a timer's content also restarts it (and leaves untouched timers alone).
- `doctor`: drift reported + exit 1 unrepaired; drift healed → exit 0 / "re-rendered units"; JSON includes `drifted_unit_files`; existing fixtures updated.

Full suite: **917 pass / 1 skip / 0 fail**. Typecheck clean.

## Notes for review

- Left as separate scope (Miles's ask #3): scheduling `doctor` itself + notify-on-non-ok with dedup. Happy to add as a further PR.
- Open question worth a look: should we *also* make the update-time heal louder when the bus is unreachable (`defaultHealUnits` returns null today), or is "doctor self-heals on next startup/heartbeat" enough? I think the latter makes it rock-solid, but flagging it.

### DIFF
--- a/src/cli/doctor.ts
+++ b/src/cli/doctor.ts
@@ -20,7 +20,7 @@
 
 import { spawn } from "node:child_process";
 import { defineCommand } from "citty";
-import { existsSync } from "node:fs";
+import { existsSync, readFileSync } from "node:fs";
 import { basename } from "node:path";
 
 import { type Config, loadConfig, personaDir } from "../config.ts";
@@ -38,6 +38,7 @@ import {
   BunSystemctlRunner,
   buildSystemctlEnv,
   defaultUnitPath,
+  driftedUnitNames,
   ensureSystemdUnitsCurrent,
   ensureUserSystemdEnv,
   HEARTBEAT_TIMER_NAME,
@@ -46,6 +47,7 @@ import {
   NIGHTLY_TIMER_NAME,
   nightlyServicePath,
   nightlyTimerPath,
+  phantombotUnitTargets,
   type SystemctlRunner,
   TICK_TIMER_NAME,
   tickServicePath,
@@ -93,12 +95,21 @@ export interface DoctorReport {
   };
   /**
    * Linux-only — undefined on macOS and dev hosts without a user-systemd
-   * bus. When present, lists which unit files are missing from disk and
-   * which timers are not active right now. A healthy box reports empty
-   * arrays for both.
+   * bus. When present, lists which unit files are missing from disk, which
+   * present-but-stale unit files have drifted from the running binary's
+   * templates, and which timers are not active right now. A healthy box
+   * reports empty arrays for all three.
    */
   systemd?: {
     missing_unit_files: string[];
+    /**
+     * Unit files present on disk but whose content no longer matches the
+     * running binary's templates — e.g. a pre-`OnCalendar` heartbeat timer
+     * that an in-place `update` left behind because the post-swap heal
+     * couldn't reach the systemd bus. Healed by the same re-render path as
+     * missing files (and, for timers, restarted so the new schedule arms).
+     */
+    drifted_unit_files: string[];
     inactive_timers: string[];
     /** True when we re-rendered or re-armed at least one thing. */
     repaired: boolean;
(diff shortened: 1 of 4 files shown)


kaieriksen (approved):
Reviewed the systemd drift-healing changes. No blockers found. The drift detector is scoped to existing unit files, doctor now surfaces drift in both human/JSON output, and rewritten timers are restarted so the new schedule is actually armed.\n\nValidated locally:\n- bun test tests/lib-systemd.test.ts tests/cli-doctor.test.ts\n- bunx tsc --noEmit

andrewagrahamhodges (commented):
## Code Review: fix(doctor,systemd): detect + heal systemd unit content drift on existing boxes

**Author:** @robertclawson | **Files:** 4 | **Verdict:** Comment Only

### Findings

| 1 | 🟡 **Minor** | `src/lib/systemd.ts` | 603-605 | Safe and clean state keeping: storing unit templates that were rewritten in `rewroteTimerUnits` allows us to precisely target only the rewritten timers for a forced restart. |
| 2 | 🔵 **Trivial** | `src/lib/systemd.ts` | 517-535 | Pure, side-effect-free detector `driftedUnitNames` is an elegant solution. Accepting `readUnit` as a dependency injector makes testing this without mock filesystem operations extremely clean and lightning fast. |

### What's Good
- **Robust Self-Heal Loop:** Combining drift detection in `doctor` with restarting the modified timers directly guarantees that existing boxes running legacy `OnUnitActiveSec` timers will transparently self-heal at startup or the next heartbeat check.
- **Excellent Execution Scope:** De-coupling the change-detection logic (`driftedUnitNames`) from filesystem manipulation maintains solid architectural boundaries.
- **Uncompromising Test Coverage:** Verifying that a content rewrite *forces* a timer restart (while leaving untouched timers alone) prevents zombie timers from staying on dead schedules after a daemon-reload.

<details>

```
Review and verify PR #122 (fix(doctor,systemd): detect + heal systemd unit content drift on existing boxes):

1. No critical fixes required! The pure function test patterns are excellent, and the automatic daemon-reload interaction covers the exact gap in historical transition states. Execute all tests locally to verify compatibility.

Branch: fix/doctor-detect-unit-drift
Base: main
```
</details>

PR #125, opened 2026-05-25
fix(telegram): forward reply_to_message into turn context

### LINKED ISSUE #124: Telegram bridge drops reply_to_message — agent can't see what user replied to
## Problem
When a user taps "Reply" on an older Telegram message and sends text, the persona sees only the new message body. The `reply_to_message` field from Telegram's Update payload is not forwarded into the agent's turn context.

Result: agent assumes user is replying to its most recent assistant turn (the `<previous_response>` block), even when the user explicitly replied to a message from N turns back. Causes wrong-context replies and "merge" / "yes" ambiguity bugs.

## Repro
1. Send message A → agent replies A'.
2. Send message B → agent replies B'.
3. Tap "Reply" on A' in Telegram, send "merge".
4. Agent treats it as reply to B' (most recent), not A'.

## Expected
Telegram bridge should forward `reply_to_message` metadata into the turn so the agent can disambiguate. Minimal surface: include the quoted `message_id` + text snippet in the user message envelope.

## Env
- phantombot v1.1.75
- Telegram channel
- Persona: miles

## Suggested fix
In the Telegram listener that builds the user turn payload, when `update.message.reply_to_message` is present, include `{reply_to: {message_id, text}}` alongside the new text. Surface in the system context the agent sees (similar to how `<previous_response>` is injected today).

jankovandeventer:
## Summary
- Telegram bridge now forwards `reply_to_message` metadata (message id, text snippet, bot/user flag) into the agent's turn context.
- New `extractReplyTo()` + `formatReplyToContext()` helpers; parsed in all three update paths (text / voice / attachment).
- `processChatMessage` prepends `[in reply to your earlier message: "..."]` / `[in reply to user's earlier message: "..."]` to `msg.text` **after** `replyModalityOverride` so quoted text can't flip voice/text routing.

Fixes #124.

## Test plan
- [x] `bun test tests/channels-telegram.test.ts` — 80 pass, 0 fail (15 new tests)
- [ ] Manual: reply to an older message in Telegram, confirm agent sees the prefix in `userMessage`

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -121,6 +121,28 @@ export interface TelegramAttachment {
     | "sticker";
 }
 
+/**
+ * Quoted-message metadata when the user taps "Reply" on a previous
+ * Telegram message and sends a new one. Carried alongside the new
+ * message so the agent can disambiguate which earlier message the
+ * user is actually replying to — without this, a "merge" / "yes"
+ * lands with no referent and the agent assumes it's a response to
+ * the most recent assistant turn (which is wrong if the user
+ * deliberately scrolled up to reply to something older).
+ */
+export interface TelegramReplyTo {
+  /** Telegram's `message_id` of the quoted message. */
+  messageId: number;
+  /** Text snippet of the quoted message, truncated for prompt sanity.
+   *  Empty when the quoted message had no text (e.g. media without
+   *  caption). */
+  text: string;
+  /** True when the quoted message was sent by the bot (i.e. the
+   *  assistant itself). Lets the agent distinguish "user replied to
+   *  my earlier turn" vs "user quoted their own earlier message". */
+  fromBot: boolean;
+}
+
 export interface TelegramMessage {
   updateId: number;
   chatId: number;
@@ -144,6 +166,11 @@ export interface TelegramMessage {
    *  Downloaded to the per-chat inbox and surfaced to the harness as
    *  an absolute path the agent can `read`. */
   attachment?: TelegramAttachment;
+  /** Populated when this message was sent as a Telegram reply (the
+   *  user tapped "Reply" on an earlier message). Forwarded into the
+   *  agent's user-message envelope so it can disambiguate which past
+   *  turn the new text is actually about. */
+  replyTo?: TelegramReplyTo;
 }
 
 export interface TelegramTransport {
@@ -405,6 +432,13 @@ interface TelegramRawPhotoSize {
   height?: number;
 }
 
+interface TelegramRawReplyToMessage {
+  message_id?: number;
+  text?: string;
+  caption?: string;
+  from?: { id?: number; is_bot?: boolean; username?: string };
+}
+
 interface TelegramRawUpdate {
   update_id?: number;
   message?: {
@@ -425,6 +459,10 @@ interface TelegramRawUpdate {
     animation?: TelegramRawFile;
     sticker?: TelegramRawFile;
     photo?: TelegramRawPhotoSize[];
+    /** Populated by Telegram when the sender used the "Reply" UI on a
+     *  prior message. Only a small subset of fields is actually relayed
+     *  through to the agent — see {@link extractReplyTo}. */
+    reply_to_message?: TelegramRawReplyToMessage;
   };
 }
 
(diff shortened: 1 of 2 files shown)


src/channels/telegram.ts
```diff
@@ -505,6 +543,63 @@ function synthesizeFileName(
...
+ *
+ * Exported for testing.
+ */
+export function extractReplyTo(
+  raw: TelegramRawReplyToMessage | undefined,
+): TelegramReplyTo | undefined {
+  if (!raw || typeof raw.message_id !== "number") return undefined;
+  const source =
+    (typeof raw.text === "string" && raw.text.length > 0
+      ? raw.text
```
andrewagrahamhodges, line 569: 🔵 **Trivial**: Since your nested ternaries are guaranteed to return a string fallback (`""`), the trailing `?? ""` nullish coalescing check is redundant and can be omitted.

andrewagrahamhodges (dismissed):
## Code Review: Forward reply_to_message into turn context

### Findings

| 1 | 🔵 Trivial | src/channels/telegram.ts | 569 | Redundant nullish coalescing `?? ""` on string-guaranteed fallback query. |

### What's Good
- Placing the reply context formatting step *after* the voice/text modality check in `processChatMessage` is an excellent design choice. It prevents quoted sentences containing instruction words from tricking the channel's media routers.
- Collapsing nested whitespaces and linebreaks (`/\s+/g`) keeps the bracketed envelope clean and uniform.
- Superb integration and testing strategy with mock-state validation and exact Telegram endpoints.

<details>

```
Review of PR #125:

1. [Trivial] src/channels/telegram.ts:569 - The ternary statement already guarantees a string fallback: `""`. The nullish coalescing operator `?? ""` at the very end is redundant and can be safely removed.
```
</details>

src/channels/telegram.ts
```diff
@@ -505,6 +543,63 @@ function synthesizeFileName(
...
+ * tapped "Reply" on. Mirrors the `[attached: <path>]` convention so the
+ * agent reads it as a structured envelope marker, not as free-form
+ * prose from the user. Exported for testing.
+ */
+export function formatReplyToContext(replyTo: TelegramReplyTo): string {
+  const who = replyTo.fromBot
+    ? "your earlier message"
+    : "user's earlier message";
+  if (replyTo.text.length === 0) {
+    return `[in reply to ${who} (no text content)]`;
```
kaieriksen, line 595: The parser keeps `messageId`, but the rendered context drops it before the harness sees the reply. That makes no-text replies indistinguishable (`[in reply to your earlier message (no text content)]`) and also loses the exact Telegram message id the comment above says is enough to disambiguate “message N”. Please include the quoted `messageId` in this envelope and cover the no-text/media reply case in the dispatch test.

robertclawson, line 595: Concrete suggestion: include `replyTo.messageId` in both branches here, e.g.
```ts
const who = replyTo.fromBot
  ? `your earlier message #${replyTo.messageId}`
  : `user's earlier message #${replyTo.messageId}`;
```
That way the no-text variant becomes `[in reply to your earlier message #7 (no text content)]` and the agent has the id to disambiguate, which is what the parser docstring already promises.

robertclawson (commented):
Holding off on approval — agreeing with @kaieriksen on `formatReplyToContext`.

The `extractReplyTo` docstring explicitly says the parser carries `messageId` so the agent can tell `message N from me` apart from `a media message they sent earlier`. But `formatReplyToContext` drops the id, so every no-text reply renders as the identical `[in reply to your earlier message (no text content)]` envelope and the agent has nothing to latch onto.

Either render the id into the envelope (e.g. `[in reply to your earlier message #7 (no text content)]`) or update the docstring + parser to stop pretending the id is forwarded. Prefer the former — it's the whole point of carrying `messageId` through.

Also +1 to Kai's second ask: add a dispatch-level test for the no-text/media reply case so the envelope shape is locked in.

Everything else looks good — CI green, tests thorough, modality-routing ordering is the right call.

src/channels/telegram.ts
lenaparkhodges, line 569: Andrew is spot on here. Since the nested ternaries are guaranteed to return a string (with `""` as the final fallback), this trailing `?? ""` is redundant and can be safely removed.

src/channels/telegram.ts
lenaparkhodges, line 595: Strongly agreeing with @kaieriksen and @robertclawson. We should interpolate the `replyTo.messageId` into the envelope (using Robbie's exact template suggestion). This ensures the agent can robustly disambiguate no-text or media replies rather than them all resolving to identical envelopes. Please also add the requested dispatch-level test for the no-text/media reply case.

src/channels/telegram.ts
```diff
@@ -1001,6 +1104,15 @@ async function processChatMessage(
         ? false
         : isVoice;
   const willReplyWithVoice = wantsVoiceReply && ttsSupported(input.config);
+
+  // Forward `reply_to_message` context AFTER modality detection so a
+  // quoted "send a voice note" can't flip routing for a fresh "merge"
+  // reply. We mutate `msg.text` so both the harness call and the
+  // interrupted-pair persistence (further down) see the same envelope.
+  if (msg.replyTo) {
```
lenaparkhodges, line 1122: This ordering is excellent. Delaying the reply context insertion until after the voice/text modality detection is a critical safety safeguard. It guarantees that quoted words in a reply don't accidentally trick our media routers into changing the turn modality.

lenaparkhodges (commented):
Thank you for the PR! The overall implementation is exceptionally clean, but I agree with @andrewagrahamhodges, @kaieriksen, and @robertclawson on a couple of design refinements before we can approve this for merge. In particular, adding the `messageId` to the `formatReplyToContext` envelope is critical to prevent no-text/media replies from being indistinguishable, which matches the design promised in the parser comments.

kaieriksen (approved):
Approved. My requested changes have been addressed.

PR #127, opened 2026-05-25
fix(harnesses): only reset idle timer on productive output (#123)

### LINKED ISSUE #123: pi harness: idle timer reset by heartbeat/thinking events lets wedged turns run far past the idle limit
## Summary

The `pi` harness's idle timeout (`harness_idle_timeout_s`, default 120s) is meant to kill a wedged subprocess when no productive output occurs. However, `pi.ts` calls `killer.touch()` on **every** stdout chunk, including events that `parsePiEvent` maps to `{ type: "heartbeat" }` (notably `thinking_delta`). While a turn is in this state every synthetic heartbeat resets the idle window for another full `idle_timeout_s`, so a turn that should die after 120s of inactivity can run for many minutes — and in the worst case (a tool call truly wedged but `pi` still streaming thinking events) until the hard timeout or manual intervention.

We observed turns running for **37 minutes** and **7+ minutes** despite a 120s idle limit, and one turn that ran until the host was manually rebooted ~6 hours later.

## Environment

- **Phantombot version:** 1.1.75 (binary release, updated 2026-05-23 07:30 UTC)
- **OS:** Ubuntu 24.04 LXC container
- **Harness:** `pi`
- **Config at time of incident:** `chain = ["pi"]` (no fallback)
- **Defaults:** `harness_idle_timeout_s = 120`, `harness_hard_timeout_s = 3600`

## Steps to reproduce / observed behavior

1. Trigger a persona turn that causes `pi` to invoke a long-running external tool (observed with `monkeyc`).
2. The tool process stops making forward progress but `pi` continues to emit `thinking_delta` (and similar non-productive) events on stdout.
3. `pi.ts:119` calls `killer.touch()` for each such chunk, restarting the idle window every time.
4. The 120s idle kill is repeatedly postponed. The process can run for many minutes — bounded only by how often the heartbeats stop for ≥ `idle_timeout_s` in a row, or by the hard timeout.

### 1. Code path

**`src/harnesses/pi.ts` (~line 119)** — `touch()` is invoked unconditionally for every chunk:

```typescript
for await (const chunk of proc.stdout as ReadableStream<Uint8Array>) {
  killer.touch();   // <-- resets idle timer on EVERY chunk
  buffer += decoder.decode(chunk, { stream: true });
  // ...
}
```

**`src/harnesses/pi.ts` (~line 261)** — non-productive events become heartbeats:

```typescript
// thinking_delta + anything else → heartbeat.
// We intentionally do NOT include the content in the chunk (would leak
// chain-of-thought to the user). The signal is just "pi is alive."
return { type: "heartbeat" };
```

**`src/lib/harnessRunner.ts`** — `touch()` resets the idle timer unconditionally:

```typescript
touch(): void {
  if (cause || disposed) return;
  clearTimeout(idleTimer);
  idleTimer = setTimeout(() => triggerKill("idle"), opts.idleTimeoutMs);
}
```

### 2. Runtime logs (journalctl, LXC 103)

The following shows three consecutive turns where the idle kill fired **far** later than the 120s limit:

```
07:32:03  orchestrator: trying harness pi
08:09:10  pi.invoke killed: idle   (durationMs: 2_227_989 ≈ 37 min)

08:28:53  orchestrator: trying harness pi
08:36:14  pi.invoke killed: idle   (durationMs: 440_921 ≈ 7.3 min)

08:36:52  orchestrator: trying harness pi
08:44:36  pi.invoke killed: idle   (durationMs: 464_332 ≈ 7.7 min)
```

And the final turn before manual reboot:

```
09:03:23  orchestrator: trying harness pi
...       (no further log output from phantombot)
15:29:44  -- Boot -- (host rebooted)
```

During the 09:03–15:29 window, container CPU stayed at 90%+ and RAM saturated.

Interpretation: the 37-minute and ~7-minute idle kills are consistent with "heartbeats *eventually* stopped for ≥120s, so the idle kill fired late rather than never." The 6-hour case is consistent with a stream of synthetic events that never had a ≥120s gap. Both fit the same root cause — idle is measured against any chunk, not against productive chunks.

## Root cause

The idle timer measures **time since last stdout chunk**, not **time since last productive output**. Because `pi` emits synthetic events (heartbeats from `thinking_delta` etc.) while a tool call is in flight, each such event extends the idle window by another full `idle_timeout_s`. A wedged turn therefore dies *eventually* (whenever heartbeats happen to pause long enough) rather than within the configured idle window — or, in the worst case, not until the hard timeout.

## Proposed fix

Move `killer.touch()` from the outer chunk loop to the inner per-line branches, and call it only on **productive** output — i.e., anything we yield to the caller that isn't a heartbeat. Concretely (against current `pi.ts:117-139`):

```typescript
for await (const chunk of proc.stdout as ReadableStream<Uint8Array>) {
  // killer.touch() removed from here
  buffer += decoder.decode(chunk, { stream: true });
  const lines = buffer.split("\n");
  buffer = lines.pop() ?? "";
  for (const line of lines) {
    const trimmed = line.trim();
    if (!trimmed) continue;
    let parsed: unknown;
    try {
      parsed = JSON.parse(trimmed);
    } catch {
      killer.touch();                       // non-JSON line is real output
      yield { type: "progress", note: trimmed };
      continue;
    }
    const c = parsePiEvent(parsed);
    if (c) {
      if (c.type !== "heartbeat") killer.touch();   // only reset on productive chunks
      if (c.type === "text") finalText += c.text;
      yield c;
    }
  }
}
```

Rule of thumb: "touch on any yielded chunk that isn't a heartbeat." This preserves the idle-timeout safety net without changing the heartbeat behavior for the UI layer.

## Workaround applied locally

We mitigated immediate risk by lowering timeouts and adding a fallback harness:

```toml
[harnesses]
chain = ["pi", "claude"]
harness_idle_timeout_s = 60
harness_hard_timeout_s = 300
```

However, this only reduces the blast radius — a stuck `pi` can still keep postponing the 60s idle kill as long as heartbeats arrive faster than once every 60 seconds, and the hard timeout is the only firm upper bound.

---

Happy to open a PR for the proposed fix if the approach looks right.

14skywalker:
## Summary

The idle killer was `touch()`'d on every raw stdout chunk in all four
harnesses, so synthetic heartbeats (e.g. streamed thinking deltas) kept
postponing the idle kill on a wedged turn. The idle timer should measure
time since last *productive* output, not since last raw chunk.

## What

- Drop the unconditional `killer.touch()` at the top of the stdout loop in
  `pi`, `claude`, `codex` and `gemini`.
- Touch only on genuinely productive output: non-JSON progress lines and
  non-heartbeat parsed events.

`pi` and `claude` are the real exposure — both stream reasoning/thinking
that maps to `heartbeat`, so a runaway loop kept the killer alive. `codex`
and `gemini` only emit discrete completion events as heartbeats, so the
change is defensive there, but the invariant is now consistent across all
harnesses.

## Test plan

- `bun run typecheck` — clean
- `bun test claude codex gemini pi` — 90 pass, 0 fail (includes the
  idle-kill behaviour tests)
- The 14 unrelated failures elsewhere in the suite are pre-existing
  (systemd/`runUpdate` tests that fail on macOS) and not touched here.

Fixes #123

### DIFF
--- a/src/harnesses/codex.ts
+++ b/src/harnesses/codex.ts
@@ -82,7 +82,10 @@ export class CodexHarness implements Harness {
 
     try {
       for await (const chunk of proc.stdout as ReadableStream<Uint8Array>) {
-        killer.touch();
+        // NB: do NOT touch() here. The idle timer must measure time since
+        // last *productive* output, not since last raw chunk — otherwise
+        // synthetic heartbeats keep postponing the idle kill on a wedged
+        // turn. See issue #123.
         buffer += decoder.decode(chunk, { stream: true });
         const lines = buffer.split("\n");
         buffer = lines.pop() ?? "";
@@ -93,11 +96,15 @@ export class CodexHarness implements Harness {
           try {
             parsed = JSON.parse(trimmed);
           } catch {
+            killer.touch(); // non-JSON line is real output
             yield { type: "progress", note: trimmed.slice(0, 200) };
             continue;
           }
           const c = parseCodexEvent(parsed);
           if (!c) continue;
+          // Only reset the idle window on productive chunks; heartbeats
+          // keep the UI typing indicator alive but must not extend idle.
+          if (c.type !== "heartbeat") killer.touch();
           if (c.type === "text") finalText += c.text;
           if (c.type === "done") {
             usageMeta = c.meta;
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -143,7 +143,10 @@ export class ClaudeHarness implements Harness {
 
     try {
       for await (const chunk of proc.stdout as ReadableStream<Uint8Array>) {
-        killer.touch();
+        // NB: do NOT touch() here. The idle timer must measure time since
+        // last *productive* output, not since last raw chunk — otherwise
+        // synthetic heartbeats (streamed thinking blocks etc.) keep
+        // postponing the idle kill on a wedged turn. See issue #123.
         buffer += decoder.decode(chunk, { stream: true });
         const lines = buffer.split("\n");
         buffer = lines.pop() ?? "";
@@ -155,11 +158,15 @@ export class ClaudeHarness implements Harness {
             parsed = JSON.parse(trimmed);
           } catch {
             // Not stream-json — surface as out-of-band progress note.
+            killer.touch(); // non-JSON line is real output
             yield { type: "progress", note: trimmed };
             continue;
           }
           const c = parseStreamJson(parsed);
           if (c) {
+            // Only reset the idle window on productive chunks; heartbeats
+            // keep the UI typing indicator alive but must not extend idle.
+            if (c.type !== "heartbeat") killer.touch();
             if (c.type === "text") finalText += c.text;
             yield c;
           }
--- a/src/harnesses/gemini.ts
+++ b/src/harnesses/gemini.ts
@@ -186,7 +186,10 @@ export class GeminiHarness implements Harness {
 
     try {
       for await (const chunk of proc.stdout as ReadableStream<Uint8Array>) {
-        killer.touch();
+        // NB: do NOT touch() here. The idle timer must measure time since
+        // last *productive* output, not since last raw chunk — otherwise
+        // synthetic heartbeats keep postponing the idle kill on a wedged
+        // turn. See issue #123.
         buffer += decoder.decode(chunk, { stream: true });
         const lines = buffer.split("\n");
         buffer = lines.pop() ?? "";
@@ -207,11 +210,15 @@ export class GeminiHarness implements Harness {
             log.debug("gemini: non-JSON stdout line", {
               line: trimmed.slice(0, 200),
             });
+            killer.touch(); // non-JSON line is real output
             yield { type: "progress", note: trimmed.slice(0, 200) };
             continue;
           }
           const c = parseGeminiEvent(parsed);
           if (!c) continue;
+          // Only reset the idle window on productive chunks; heartbeats
+          // keep the UI typing indicator alive but must not extend idle.
+          if (c.type !== "heartbeat") killer.touch();
           if (c.type === "text") finalText += c.text;
           if (c.type === "done") {
             // gemini's `result` event carries the authoritative stats.
--- a/src/harnesses/pi.ts
+++ b/src/harnesses/pi.ts
@@ -116,7 +116,10 @@ export class PiHarness implements Harness {
 
     try {
       for await (const chunk of proc.stdout as ReadableStream<Uint8Array>) {
-        killer.touch();
+        // NB: do NOT touch() here. The idle timer must measure time since
+        // last *productive* output, not since last raw chunk — otherwise
+        // synthetic heartbeats (thinking_delta etc.) keep postponing the
+        // idle kill on a wedged turn. See issue #123.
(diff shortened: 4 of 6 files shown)


andrewagrahamhodges (dismissed):
## Code Review: Only reset idle timer on productive output

**Author:** @14skywalker | **Files:** 4 | **Verdict:** Approve

### Findings

*(None — the logic is clean, robust, and correctly aligned across all four adapters)*

### What's Good
- Disentangling raw stdout chunks from "productive progress" is the correct behavioral fix for #123. Letting mechanical heartbeat patterns (like streamed thinking deltas) reset the idle interval completely circumvented the killer's core purpose.
- I like the defensive choice to consider non-JSON buffer lines (the parsing fallback `catch` blocks) as productive progress. It ensures that any unusual backend trace logging or developer warnings on stdout prevent a premature timeout.
- The uniformity of the fix across all downstream runners (`claude`, `codex`, `gemini`, `pi`) prevents any future behavioral divergence.

<details>

```
Review of PR #127: Complete. No actions needed.
```
</details>

src/harnesses/codex.ts
```diff
@@ -93,11 +96,15 @@ export class CodexHarness implements Harness {
...
           } catch {
+            killer.touch(); // non-JSON line is real output
             yield { type: "progress", note: trimmed.slice(0, 200) };
             continue;
           }
           const c = parseCodexEvent(parsed);
           if (!c) continue;
+          // Only reset the idle window on productive chunks; heartbeats
+          // keep the UI typing indicator alive but must not extend idle.
+          if (c.type !== "heartbeat") killer.touch();
```
kaieriksen, line 107: This timeout behavior is the point of the PR, but the current test suite does not exercise it: the existing timeout tests use silent hangs, so they pass both before and after this change. Please add a regression test with repeated parsed `heartbeat` events beyond `idleTimeoutMs` that still gets killed, plus a productive `text` or `progress` stream that continues to reset idle. One harness-level test would be enough to lock the contract down.

14skywalker:
Done — added a harness-level regression test in e3e9ef0.

`fake-codex.sh` got two new modes:
- `heartbeats`: emits repeated `turn.started` reasoning events past `idleTimeoutMs` before a final `turn.completed`.
- `productive`: emits steady `agent_message` text that should keep resetting the idle timer.

`harnesses-codex.test.ts` now asserts both contracts: repeated heartbeats get idle-killed before the turn completes, while steady productive output keeps the timer alive and completes normally.

Verified the heartbeat test actually fails against the pre-fix `touch()`-on-every-chunk code (gets `done` instead of an idle error), so the contract is genuinely pinned now rather than passing on both sides. 11/11 codex tests green, typecheck clean.

andrewagrahamhodges (approved):
Verified the new regression tests on e3e9ef0! Spawning two explicit modes in `fake-codex.sh`—one checking repeating heartbeats that properly triggers an idle-kill and another checking steady productive text progress—is phenomenal. Approving!

robertclawson (approved):
Regression test (heartbeats mode in fake-codex.sh + harnesses-codex.test.ts) now exercises the exact scenario the PR fixes — heartbeats spaced under the idle window no longer postpone the idle kill, and the counterpart productive test guards against over-tightening. CI green. LGTM.

kaieriksen (approved):
Approved. The requested heartbeat-vs-productive idle regression coverage is now in place, and the focused Codex harness test suite passes on the updated head.

PR #128, opened 2026-05-25
feat(telegram): multi-persona bots in one process (closes #126)

### LINKED ISSUE #126: Allow multiple persona-bound Telegram bots in a single phantombot process
## Problem

Today `[channels.telegram]` is a single object — one bot token, one allowlist, one persona binding. The persona is implicitly `config.defaultPersona`. With multiple personas on one host, only the default persona is reachable via Telegram. The other personas are mute on that channel.

This is awkward when a single operator runs several personas with distinct purposes (e.g. `miles` for engineering, `desiree` for personal assistant, `phantom` for general). The natural mental model is "one BotFather bot per persona," and Telegram supports that fine — phantombot just doesn't expose it.

## Design intent check

`architecture.md`:120 says *"Multi-tenant. Phantombot is single-operator."* and the README echoes it. But **single-operator with N personas is not multi-tenancy** — it's "the same operator wearing several hats." The OpenClaw importer (`parseOpenClawTelegram`) already reads from `channels.telegram.accounts.default.botToken`, suggesting the upstream lineage was multi-account; phantombot just didn't carry that through.

So: this PR proposes an additive extension that stays within the single-operator intent.

## Proposed config (backward compatible)

```toml
# Existing block keeps its current semantics — binds to defaultPersona.
[channels.telegram]
token = \"...\"
allowed_user_ids = [123]

# NEW: optional additional bots, each bound to a named persona.
[channels.telegram.personas.miles]
token = \"...\"
allowed_user_ids = [123]

[channels.telegram.personas.desiree]
token = \"...\"
allowed_user_ids = [123]
```

`Config.channels.telegram` becomes `{ default?: TelegramAccount; byPersona: Record<string, TelegramAccount> }`. Old configs with only `[channels.telegram]` resolve to `{ default: { ... }, byPersona: {} }` — zero behavior change.

## Runtime shape

Single process, N listeners. In `runRun()`:

1. Build a list `Array<{ persona, agentDir, account }>` from default + byPersona.
2. Validate every persona exists (fail-soft: warn + skip missing persona, keep going).
3. Create one `HttpTelegramTransport` + `runTelegramServer` per entry.
4. `Promise.all` with a shared `AbortController` so Ctrl-C / SIGTERM stops them all.

The run-lock stays as-is (one process per host). Memory store is already keyed on `(persona, conversation)` so persona isolation is already guaranteed — no migration needed.

## Regression risk

Low.
- Old `[channels.telegram]` without `personas` keeps current behavior exactly.
- `/restart`, doctor, update notify, post-restart marker — all single-process concerns, unchanged.
- One token-per-persona — if a token is reused by mistake we detect duplicates at startup and refuse to start with a clear error.

Doctor's startup hooks (`healDefaultPersonaIfBroken`, post-restart notify) keep targeting `defaultPersona` — documented, not regressed.

## Out of scope

- Webhook mode (still long-poll only).
- Dynamic add/remove without restart.
- Cross-persona DMs / routing rules.

## Acceptance

- Old configs work unchanged.
- New `[channels.telegram.personas.<name>]` entries each spawn an independent listener bound to that persona.
- Tests cover: config parse (default-only, personas-only, both, duplicate token detection); `runRun` spawn fan-out; per-persona allowlist enforcement.

jankovandeventer:
## Summary
- Adds optional `[channels.telegram.personas.<name>]` config blocks. Each one spawns its own long-poll listener bound to the named persona, all sharing one process, one memory store (already persona-keyed), and one run-lock.
- Existing `[channels.telegram]` block keeps current semantics and stays bound to `defaultPersona`. Configs that don't add `personas` blocks behave exactly as before — zero regression surface.
- Duplicate-token guard at startup refuses configs that point two listeners at the same bot (Telegram serializes long-poll on a token; two listeners would silently starve each other).

Closes #126.

## Config example

```toml
[channels.telegram]
token = "111:default-bot"
allowed_user_ids = [123]

[channels.telegram.personas.miles]
token = "222:miles-bot"
allowed_user_ids = [123]

[channels.telegram.personas.desiree]
token = "333:desiree-bot"
allowed_user_ids = [123]
```

## What changed
- `src/config.ts` — extracted `TelegramAccount` type, added `Config.channels.telegramPersonas?: Record<persona, TelegramAccount>` plus the parser for it.
- `src/cli/run.ts` — new `planListeners()` builds the (persona, agentDir, account) list, validates persona dirs (warns + skips missing), fatals on duplicate tokens, then `Promise.all`s `runTelegramServer` calls with a shared `AbortSignal`. Post-restart notify + startup doctor stay on the default persona (operator-facing signals — one channel is correct).
- `src/channels/telegram.ts` — `RunTelegramServerInput` gains an explicit optional `account?` so each listener sees its own token / allowlist / poll timeout (falls back to `config.channels.telegram!` for the single-bot legacy path).
- `README.md` — flips the "single persona at runtime" heads-up to document the new shape.

## Test plan
- [x] `bun test tests/cli-run.test.ts tests/config.test.ts` — 22/22 pass
- [x] Full suite: 914 pass / 14 fail / 1 skip — same 14 failures as `main` (unrelated, pre-existing)
- [x] 11 new tests cover: config parse (default-only, personas-only, both, missing token skipped, poll timeout clamped) and `runRun` planning (personas-only boots, missing persona dir warns + skipped, all-missing fatals, duplicate-token fatal across default↔persona and persona↔persona)
- [ ] Reviewer manual: run with a `[channels.telegram.personas.<name>]` block + verify two bots reply with their respective persona's memory and BOOT

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -216,6 +216,7 @@ export async function runRun(input: RunInput = {}): Promise<number> {
       config,
       currentVersion: VERSION,
       transport: adminTransport,
+      adminAccount: adminListener.account,
     });
     if (r.status === "success_notified" || r.status === "failure_notified") {
       log.info("run: post-restart notify", {
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -13,7 +13,12 @@ import {
   HttpTelegramTransport,
   runTelegramServer,
 } from "../channels/telegram.ts";
-import { type Config, loadConfig, personaDir } from "../config.ts";
+import {
+  type Config,
+  loadConfig,
+  personaDir,
+  type TelegramAccount,
+} from "../config.ts";
 import { buildHarnessChain } from "../harnesses/buildChain.ts";
 import type { WriteSink } from "../lib/io.ts";
 import { log } from "../lib/logger.ts";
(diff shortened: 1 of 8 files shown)


jankovandeventer:
@andrewagrahamhodges @robertclawson 
Not sure if this is opposing the vision?
But, I like to split work to different bots for the memory and efficiency.
I just want to have like, two or three chats going, but that causes responses out of turn and stops processes when I have a brainfart.

This means I could start a task, and then ask general questions on the side without leaving telegram.

Curious to hear what you think.

src/cli/run.ts
```diff
@@ -90,7 +194,17 @@ export async function runRun(input: RunInput = {}): Promise<number> {
   }
 
   const memory = await openMemoryStore(config.memoryDbPath);
-  const transport = new HttpTelegramTransport(tg.token);
+
+  // The post-restart-notify + startup-doctor hooks operate against the
```
kaieriksen, line 198: This is the right boundary for the feature. Keeping post-restart notify and startup doctor on an admin listener makes `/update` an operator/admin-channel behavior instead of trying to preserve the persona bot that happened to receive the command. The code comment is useful because it makes that product decision explicit for future maintainers.

andrewagrahamhodges (approved):
## Code Review: Multi-Persona bots in one process

**Author:** @jankovandeventer | **Files:** 8 | **Verdict:** Approve

### Findings

*(None — clean and ready to go)*

### What's Good
- I completely agree with @kaieriksen on the boundary design. Restricting admin/operator-level capabilities—such as post-update restarts and startup diagnostic hooks—exclusively to the default/admin persona bot maintains clean system boundaries. There is zero value in broadcasting system updates dynamically across dedicated secondary persona bots.
- Backward Compatibility: Single-bot configs that omit `[channels.telegram.personas]` map securely to `undefined` and run exactly as before, presenting **zero risk of regression** for existing environments.
- Fail-safe checks: Aborting all polling tasks simultaneously on SIGINT/SIGTERM or unhandled listener exceptions is extremely safe and precludes dangling background threads.

<details>

```
Review of PR #128 complete. Approved with zero actions required.
```
</details>

robertclawson (approved):
Reviewed end-to-end. Approving.

**Backward compatibility:** confirmed — single-bot configs with just `[channels.telegram]` go through the exact same code path (the new `account?` field falls back to `config.channels.telegram!`, planner produces one listener that matches the old behavior, `tg` inside `runTelegramServer` is only read for `allowedUserIds` + `pollTimeoutS` so no token leak across listeners).

**Complexity:** justified by the feature and well-factored. `planListeners` is a pure function with good unit tests, the duplicate-token guard catches a real Telegram footgun (long-poll serializes on token), missing-persona-dir is warn-and-skip so one typo can't blackhole the others, and the sibling-failure path (`ac.abort()` + `Promise.allSettled`) is correct rather than ornamental.

**Design call I like:** post-restart notify + startup doctor stay pinned to the default/admin listener instead of fanning out. Operator signals shouldn't multiply by persona count.

**One non-blocking observation** for follow-up (not for this PR): `checkAndNotifyOnce` (heartbeat update-available) in `src/lib/updateNotify.ts:383` and `runNotify` in `src/cli/notify.ts:52` still only read `config.channels.telegram`. A purely personas-only config (no default block) would silently lose those two admin channels. The PR's own design treats the default block as the admin channel, so the practical guidance is "keep a default `[channels.telegram]` block alongside personas" — worth a one-liner in the README's new persona section, or a follow-up to migrate those two callsites to the same `adminAccount` fallback pattern this PR introduced for `notifyPostRestartIfPending`.

src/cli/run.ts
lenaparkhodges: Absolutely agree with this boundary. Keeping operator-level signals restricted to a single admin bot keeps system administration clean, avoids duplicate channel spam, and mirrors how other system services manage routing. Solid design choice.

src/lib/updateNotify.ts
```diff
@@ -489,8 +500,11 @@ export async function notifyPostRestartIfPending(
   const marker = await readPendingUpdate(input.pendingPath);
   if (!marker) return { status: "no_marker" };
 
-  const tg = input.config.channels.telegram;
-  if (!tg) {
+  // Pick the account that drives this notify. Explicit adminAccount wins
+  // (personas-only setups have no default block); fall back to the
```
lenaparkhodges, line 504: Excellent change adding the `adminAccount` fallback here. As Robbie noted in his review, we should plan a follow-up to migrate `checkAndNotifyOnce` in this file and `runNotify` in `src/cli/notify.ts` to the same fallback pattern so personas-only configs don't silently skip those update-notify or CLI broadcast behaviors. Really solid!

src/cli/run.ts
```diff
@@ -148,17 +268,56 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
+          l === adminListener
+            ? adminTransport
+            : new HttpTelegramTransport(l.account.token),
+        signal: ac.signal,
+        out,
+        err,
+      }),
+    );
+    try {
+      await Promise.all(tasks);
```
lenaparkhodges, line 294: This graceful teardown logic is beautifully handled. Fanning out the tasks and catching first-failures to abort the sibling pollers prevents orphan background tasks and avoids race conditions with the database and process lock during shutdown.

src/config.ts
```diff
@@ -361,6 +378,76 @@ function buildTelegramConfig(
...
+  const personas = tomlTelegram.personas;
+  if (!personas || typeof personas !== "object" || Array.isArray(personas)) {
+    return undefined;
+  }
+  const out: Record<string, TelegramAccount> = {};
+  for (const [personaName, raw] of Object.entries(
+    personas as Record<string, unknown>,
+  )) {
+    if (!raw || typeof raw !== "object" || Array.isArray(raw)) continue;
+    const entry = raw as Record<string, unknown>;
```
lenaparkhodges, line 423: Fabulous twelve-factor design! Allowing environment variables like `TELEGRAM_BOT_TOKEN_<PERSONA>` to override the TOML config makes containerized and systemd-based deployments extremely easy to manage without checked-in secrets.

lenaparkhodges (approved):
This is an absolutely stellar implementation! I have reviewed the code end-to-end, ran the test suite (all tests pass flawlessly with zero type errors), and verified the edge cases. The architecture for multi-persona hosting in a single process is robust, clean, and beautifully backward-compatible.

I completely agree with Kai on the administrative/operator boundaries, and I've left a couple of specific inline comments of appreciation as well as a note highlighting Robbie's excellent suggestion on the heartbeat/CLI notify follow-up.

kaieriksen (approved):
Approved. The admin-channel boundary is acceptable as implemented; `/update` may be invoked by an allowlisted operator listener, while lifecycle/doctor notifications stay routed through the admin/default channel.

PR #130, opened 2026-05-27
Fix persona-aware update restart notifications

### LINKED ISSUE #129: telegram: post-restart notify uses wrong bot in hybrid (default + personas) setup
## Context

PR #128 introduces multi-persona Telegram support — one process, N bots, one per persona. Post-restart \`/update\` notification was patched to fall back to the admin listener's account when no default \`[channels.telegram]\` is configured (commit 439fec3).

A reviewer flagged a remaining hole that the current fix does not cover.

## The bug

**Hybrid config:** default \`[channels.telegram]\` AND one or more \`[channels.telegram.personas.*]\` are configured (e.g. default + Miles).

1. User issues \`/update\` via the **Miles** bot DM.
2. \`writePendingUpdate\` stores \`chatId = user's chat with Miles\`.
3. Process restarts.
4. On boot, \`adminListener = plan.listeners.find(p => p.persona === \"default\") ?? plan.listeners[0]\` resolves to **default**.
5. \`notifyPostRestartIfPending\` builds \`HttpTelegramTransport(defaultBot.token)\` and calls \`sendMessage(milesChatId, ...)\`.
6. Telegram rejects: chat not found (the default bot has no shared history with that user), or silently drops if the default bot was blocked.
7. The marker is cleared regardless → no retry, no confirmation.

## Why the current patch doesn't catch it

PR #128's fix routes notify through \`adminAccount\`. In a hybrid setup the admin account is *default*, not the persona the marker actually came from. The marker has no \`persona\` field, so notify can't know which transport to use.

## Fix sketch

1. Extend \`PendingUpdate\` schema with optional \`persona?: string\` (or \`botToken?: string\` — but persona is the cleaner key, lets us look up token + allow-list).
2. \`writePendingUpdate\` records the persona that handled the \`/update\` command.
3. \`notifyPostRestartIfPending\` selects transport + allow-list by marker.persona instead of by adminListener.
4. Backwards-compat: missing \`persona\` field → fall through to current adminAccount behavior (existing markers from older versions don't break).
5. Test: hybrid config + marker with \`persona: \"miles\"\` → notify uses miles token, not default.

## Severity

MAJOR — silent failure on the success path of a feature users will exercise (\`/update\` is the most common admin action). Personas-only and default-only configs are unaffected.

## Related

- PR #128 (feat: telegram multi-persona)
- src/lib/updateNotify.ts:491-553
- src/cli/run.ts:205-220

kaieriksen:
## Summary
- store the Telegram listener persona in pending `/update` markers
- resolve post-restart notifications through that persona account before falling back to the legacy admin/default account
- add regression coverage for hybrid default + persona Telegram configs

Closes #129

## Tests
- `~/.bun/bin/bun tsc --noEmit`
- `~/.bun/bin/bun test tests/lib-updateNotify.test.ts`
- `~/.bun/bin/bun test tests/channels-commands.test.ts tests/cli-run.test.ts`
- `~/.bun/bin/bun test`

### DIFF
--- a/src/channels/commands.ts
+++ b/src/channels/commands.ts
@@ -169,12 +169,14 @@ async function handleUpdate(
   }
   log.info("commands: /update invoked", {
     chatId: ctx.chatId,
+    persona: ctx.persona,
     currentVersion: VERSION,
   });
   const r = await runUpdateFlow({
     config: ctx.config,
     currentVersion: VERSION,
     chatId: ctx.chatId,
+    persona: ctx.persona,
   });
   return { reply: r.reply, afterSend: r.restart };
 }
@@ -368,4 +370,3 @@ export function nominalContextWindow(harnessId: string): number {
       return 128_000;
   }
 }
-
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -195,17 +195,13 @@ export async function runRun(input: RunInput = {}): Promise<number> {
 
   const memory = await openMemoryStore(config.memoryDbPath);
 
-  // The post-restart-notify + startup-doctor hooks operate against the
-  // DEFAULT persona's bot. That's intentional — they're administrative
-  // signals to the operator, and the operator is the same human across
-  // all personas, so picking one channel keeps them from getting four
-  // copies of the same notification. Falls back to the first listener
-  // when no default account is configured.
+  // The post-restart-notify hook uses the persona stored in a pending
+  // `/update` marker when present, and falls back to this admin listener
+  // for legacy markers. Prefer the default listener for that fallback;
+  // use the first listener when no default account is configured.
   // Non-null: we returned above if plan.listeners.length === 0.
   const adminListener: ListenerSpec =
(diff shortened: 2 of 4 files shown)


andrewagrahamhodges (approved):
## Code Review: Fix persona-aware update restart notifications

**Author:** @kaieriksen | **Files:** 4 | **Verdict:** Approve

### Findings

*(None — the logic is pristine and robustly tests both legacy and modern paths)*

### What's Good
- Great integration of the multi-persona feature from #128. Keeping track of the exact triggering persona inside the pending update marker closes the loop on system administration UX so updates aren't followed by confusing silent-transports across bots.
- Outstanding backward compatibility coverage. Checking for `parsed.persona !== undefined` during `readPendingUpdate` ensuring type safety, while defaulting to `input.adminAccount` if no persona is recorded, handles old markers perfectly.
- Replacing the tight transport coupling in `run.ts` with clean individual transport allocations is a massive win for maintainability.

<details>

```
Review of PR #130 complete. No actions needed.
```
</details>

PR #133, opened 2026-05-28
feat(embeddings): make semantic search a clearly-optional, discoverable choice

robertclawson:
## Summary

Semantic (vector) search already degrades gracefully to keyword (FTS5/BM25) search when no embedding provider is configured — but that degraded state was **invisible**: no surfacing, no docs, no exposure during setup. An operator could run keyword-only indefinitely without realising vector search was never enabled (exactly what we just hit on the live box). This PR makes the optionality **explicit and discoverable** without ever turning absence into an error.

## Changes

- **doctor** — adds an *informational* `embeddings:` line (semantic ON / off with keyword search active). Deliberately neutral: **no ok/WARN marker, never feeds `repair_needed` or the exit code.** Absence is a valid, fully-working config, not a fault.
- **run** — a gentle one-line startup heads-up when semantic search is off.
- **init wizard** — a new **optional** step (default skip) that exposes semantic search during setup, with a short "what this improves" note. Reuses `runEmbedding`.
- **embedding** — adds an `embedded` mode so `runEmbedding` can run as a wizard sub-step without its own clack intro/outro (a nested intro renders a stray bracket) or a premature restart prompt (service isn't installed until the next step).
- **README** — a new "Semantic search (optional)" section: what it does, why it's nice, why it's optional, how to tell which mode you're in, how to enable.

## Also (flagged for the reviewer)

`cli-embedding.test.ts` had latent **env-isolation** bug: `loadConfig()` prefers a real `PHANTOMBOT_GEMINI_API_KEY` over the toml value, so the config-roundtrip tests false-failed on any box that has a key in the environment (e.g. this one). Fixed by clearing + restoring the var per test, matching the save/restore pattern the file already uses elsewhere. CI (clean env) was unaffected; this only bit local runs on the live box.

## Verification

- typecheck clean
- full suite **1006 pass / 1 skip / 0 fail** (4 new doctor tests pinning the wording + the never-WARN / never-exit-code invariant + the JSON block)
- production build bundles clean

## Test plan

- [ ] `phantombot doctor` shows the neutral `embeddings:` line in both states
- [ ] `phantombot run` prints the one-line heads-up only when semantic search is off
- [ ] `phantombot init` offers the optional step and skipping it completes setup cleanly
- [ ] enabling via the optional step writes `[embeddings]` and does not prompt for a restart mid-wizard

### DIFF
--- a/README.md
+++ b/README.md
@@ -394,6 +394,37 @@ The nightly runs as **five idempotent stages** — `essence` → `promote` → `
 
 `phantombot doctor` reads `.nightly-state.json`, `.nightly-progress.json` and `capture_log`, and reports — then auto-repairs — memory-subsystem problems: a nightly that never ran, errored, or is >24h stale; a partial checkpoint left behind; or a "dry day" (≥20 real user turns with zero captures). When repair is warranted it spawns a detached `phantombot nightly --resume`. The `run` startup catch-up routes through the same logic, so a box powered off overnight self-heals on next boot.
 
+### Semantic search (optional)
+
+Memory search has two modes. **Keyword search (FTS5/BM25)** is always on, needs no setup, and matches on the actual words in a note. **Semantic (vector) search** is an *optional* enhancement: it embeds your text into vectors so search can match on **meaning**, not just exact words — e.g. asking "how do I pay tax" can surface a note titled "VAT filing steps" even though they share no keywords.
+
+**Why it's nice.** Phantombot retrieves relevant memories automatically on every turn (the auto-retrieval "instinct" layer) and indexes past conversation turns so older context can resurface. With embeddings on, that retrieval runs as **hybrid** (keyword + semantic) and finds conceptually-related memories the keyword half would miss. It's what makes recall feel like instinct rather than grep.
+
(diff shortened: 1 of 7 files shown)


andrewagrahamhodges (approved):
## Code Review: Explicit and discoverable optional embeddings

**Author:** @robertclawson | **Files:** 7 | **Verdict:** Approve

### Findings

*(None — the PR is solid and beautifully resolves the visibility gaps)*

### What's Good
- **Discoverability without false-alarms**: Adding semantic search status visibility inside `doctor` and `run` without treating "keyword-only" mode as a warning or causing exit-code failures is a great design choice. This informs the operator without triggering monitoring warnings.
- **Nested Setup flows**: Adding the `embedded` parameter inside `embedding.ts` is an elegant way to reuse existing clack wizard logic during the `init` workflow without messing up console alignment, while safely skipping restart prompts.
- **Robust Environment Isolation**: Fixing the shell-bleed `PHANTOMBOT_GEMINI_API_KEY` leak in `cli-embedding.test.ts` was a crucial repair that ensures consistent test suite performance regardless of developer environments.

<details>

```
Review of PR #133 complete. Approved with zero actions required.
```
</details>

kaieriksen (approved):
Reviewed the optional embeddings discoverability change. I do not see a blocker: the doctor/run messaging is informational, config resolution correctly reflects env/TOML provider state, and the init embedding step stays optional. Focused local tests passed: `bun test tests/cli-doctor.test.ts tests/cli-embedding.test.ts tests/cli-run.test.ts` (42 passing).

PR #136, opened 2026-05-29
fix(stt): release per-chat lock in finally; log STT errors (closes #135)

### LINKED ISSUE #135: STT failure on non-default persona silently jams per-chat queue (no error log, no lock release)
## Symptom
Amanda (non-default persona) stopped responding to **both** voice and text after a voice note timed out on STT. Miles (default persona, same process) kept working. Only fix was `systemctl --user restart phantombot.service`.

## Repro (observed)
- phantombot v1.1.83, updated 11:51, restarted 12:24, PID 2128450
- 13:41 — Miles voice (test): STT ok logged ~3s later, orchestrator invoked, reply sent. Healthy.
- 14:01:56 — Amanda voice (~10s) incoming. **No `STT ok` log line. No orchestrator invocation. No error log.**
- 14:02:45 — Amanda text incoming. **Still no orchestrator invocation.** Subsequent text messages also queued silently.
- 14:30 — restart. Amanda answers next message immediately.

## Hypothesis
The per-chat queue holds a listener lock taken at the start of voice processing. When STT fails (or hangs) on a non-default persona path, the failure is swallowed and the lock is never released — subsequent voice/text for that chat enqueue behind a dead handler.

Miles is unaffected because the default-persona STT path is a different branch that does log + release on error.

## Expected
- STT failure logs at ERROR with persona + chat id
- Lock released in `finally`, not only on success
- Optionally: surface a Telegram message ("voice transcription failed, please retype") instead of silent swallow

## Severity
Medium. Silent wedge of a non-default persona requires manual restart and gives no log breadcrumb.

## Likely culprit
Regression in recent update (11:51 today, v1.1.83). Whatever changed in STT pipeline between prior version and v1.1.83.

jankovandeventer:
## Summary
- Closes #135 — STT failure on a non-default persona was silently wedging that chat's per-chat queue until phantombot was restarted.
- Root cause: the per-chat "lock" is a promise chain (`chatChains: Map<chatId, Promise<void>>`). When `processChatMessage` rejected, the *next* incoming message ran `prev.then(work)` on a rejected promise — `.then(onFulfilled)` on a rejected promise skips the callback, so the work never ran and the chain stayed stuck. Default-persona traffic happened to take a path that always resolved (logged `STT ok`, returned normally), so Miles kept working while Amanda (or any non-default persona) jammed.
- Fix: at chain enqueue (`src/channels/telegram.ts:~961`), convert any prior rejection to resolution (`.catch(() => {})`) so a thrown handler can no longer poison the queue. Plus tighten the STT block in `processChatMessage` to match the issue's expected behavior.

## Changes
- `src/channels/telegram.ts`
  - **Chain enqueue:** `const prev = (chatChains.get(chatId) ?? Promise.resolve()).catch(() => {})` — comment references #135.
  - **STT error logs:** both `telegram: STT failed` and the renamed `telegram: STT pipeline error` (was `voice download failed`, but the catch actually covers `transcribe()` too) now include `persona` + `chatId`.
  - **STT success log:** `telegram: STT ok` now includes `persona` for symmetry.
  - **User-facing failure message:** unified to `voice transcription failed, please retype` on both the `!r.ok` branch and the catch. The technical error stays in the log only. Each `sendMessage` is wrapped in its own try/catch so a send failure can't escape the handler.
  - **`finally {}` seam** added after the STT try/catch so future per-chat STT-specific cleanup has an obvious home (the moral "lock release in finally" the issue describes).

Out of scope: no timeout added to `transcribe()`, no refactor of the chain into an explicit `asyncio.Lock`-style primitive. The rejection-conversion + per-message error containment is the minimum surface to ship the fix.

## Test plan
- [x] `bun test tests/channels-telegram.test.ts` — 84 pass, 0 fail
- [ ] Manual: send a voice note to Amanda on staging, observe error log includes persona + chatId, observe Telegram replies with the retype message, observe a *subsequent* text message to the same chat is processed (chain advanced).
- [ ] Manual: same scenario for Miles to confirm no regression on the default-persona path.

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -40,6 +40,7 @@ import {
   transcribe,
   ttsSupported,
 } from "../lib/audio.ts";
+import { DEFAULT_STT_TIMEOUT_MS } from "../lib/voice.ts";
 import type { WriteSink } from "../lib/io.ts";
 import { log } from "../lib/logger.ts";
 import type { ServiceControl } from "../lib/systemd.ts";
@@ -1385,6 +1386,29 @@ export async function runTelegramServer(
   }
 }
 
+/** Thrown when the STT download+transcribe step exceeds its time budget. */
+class SttTimeoutError extends Error {
+  constructor(ms: number) {
+    super(`STT step exceeded ${ms}ms budget`);
+    this.name = "SttTimeoutError";
+  }
+}
+
+/**
+ * Race `p` against a timer. If the timer wins, reject with SttTimeoutError so
+ * the caller's catch can recover and the per-chat queue advances (GitHub
+ * #135). We cannot cancel the underlying request — the transport exposes no
+ * AbortSignal — so the orphaned promise is left to settle on its own; the
+ * point is that the *queue* no longer waits on it.
+ */
+function withTimeout<T>(p: Promise<T>, ms: number): Promise<T> {
+  let timer: ReturnType<typeof setTimeout> | undefined;
+  const timeout = new Promise<never>((_, reject) => {
+    timer = setTimeout(() => reject(new SttTimeoutError(ms)), ms);
+  });
+  return Promise.race([p, timeout]).finally(() => clearTimeout(timer));
+}
+
 /**
  * Process one (non-slash) message: STT if voice, run the harness chain,
  * send the reply. Stays self-contained so the polling loop can fire-and-
@@ -1441,9 +1465,18 @@ async function processChatMessage(
       );
       return;
     }
+    const fileId = msg.voice.fileId;
     try {
-      const file = await input.transport.downloadFile(msg.voice.fileId);
-      const r = await transcribe(input.config, file.data, file.mime);
+      // Bound the whole download+transcribe step. A hung request here would
+      // otherwise never settle, stalling every later message in this chat's
+      // serial queue forever — the wedge in GitHub #135.
+      const r = await withTimeout(
+        (async () => {
+          const file = await input.transport.downloadFile(fileId);
+          return transcribe(input.config, file.data, file.mime);
+        })(),
+        input.config.voice.sttTimeoutMs ?? DEFAULT_STT_TIMEOUT_MS,
+      );
       if (!r.ok) {
         log.error("telegram: STT failed", {
           error: r.error,
@@ -1487,8 +1520,6 @@ async function processChatMessage(
         });
       }
       return;
-    } finally {
-      // Seam for future per-chat STT lock release (GitHub #135).
     }
   }
 
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -1351,7 +1351,11 @@ export async function runTelegramServer(
         }
 
         // Enqueue onto this chat's serial chain.
-        const prev = chatChains.get(msg.chatId) ?? Promise.resolve();
+        // Convert prior rejection to resolution so a thrown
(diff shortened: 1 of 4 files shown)


src/channels/telegram.ts
```diff
@@ -958,7 +958,11 @@ export async function runTelegramServer(
         }
 
         // Enqueue onto this chat's serial chain.
-        const prev = chatChains.get(msg.chatId) ?? Promise.resolve();
+        // Convert prior rejection to resolution so a thrown
+        // processChatMessage doesn't wedge the per-chat queue (GitHub #135).
+        const prev = (chatChains.get(msg.chatId) ?? Promise.resolve()).catch(
```
kaieriksen, line 1357: This fixes the rejected-promise poisoning path, but it does not cover the failure mode described in #135 if STT/download hangs and leaves the previous promise pending. In that case `prev` never settles, this `.catch()` never runs, and all later text for the same chat still queues forever behind the voice turn. Since the issue says there was no STT error log and the next text never invoked the orchestrator, a hung transcription is plausible. I think this needs either a bounded timeout around the voice download/transcribe step or the PR should avoid claiming it closes #135.

src/channels/telegram.ts
```diff
@@ -958,7 +958,11 @@ export async function runTelegramServer(
...
+          () => {},
+        );
         const next = prev.then(() =>
```
kaieriksen, line 1360: Please add a regression test for the queue recovery behavior. The fix is small but central: one voice/STT failure should send the retype message, then a subsequent text message in the same chat should still invoke the harness. Without that test, this can regress silently back to the exact wedge this PR is meant to prevent.

kaieriksen (changes requested):
Thanks for the targeted fix. The queue rejection analysis is sound, and the existing suite plus CI are green. I do not think this is ready to merge yet though: the observed failure mode can still be a pending STT promise rather than a rejected one, and this PR has no regression coverage for the queue recovery it is changing. Please add the focused tests and either bound STT with a timeout/cancellation path or narrow the PR/issue claim so we do not think #135 is fully fixed when a hung transcription can still wedge the chat.

kaieriksen:
Re-checked this PR on 2026-05-31. CI is green, but there are no commits or replies after my requested changes, and both review threads are still unresolved. The missing items remain: regression coverage for queue recovery, and either a bounded STT timeout/cancellation path or narrower close-claim for #135.

jankovandeventer:
**Addressed both review points** (follow-up commit `7948f55`):

**1. Hung-promise case now covered.** The original commit fixed the *rejected*-promise poisoning but not a *hang* — if download/transcribe never settled, the chain wedged exactly as in #135. The STT step (`downloadFile` + `transcribe`) now runs inside a `withTimeout` race; a timeout throws `SttTimeoutError`, flows through the existing catch → user gets the retype prompt → the per-chat queue advances. Budget is configurable via `[voice] stt_timeout_ms` (default 60s).

**2. Regression test added.** A voice note whose transcription hangs times out, sends the retype prompt, and a subsequent text in the same chat still reaches the harness — the guarantee #135 is about.

Full suite 1091 pass / 0 fail, typecheck clean. Used a follow-up commit (not a force-push) so the delta is easy to re-review. Re-requesting your review — thanks!

kaieriksen (approved):
My previous review comments have been addressed. The STT pipeline is now bounded with a configurable timeout, the per-chat queue recovery case is covered by regression test, and the focused Telegram test file passes locally.

PR #138, opened 2026-05-30
fix(persona): heal invalid default persona dirs

robertclawson:
## Summary
- Treat a persona directory as valid only when it contains BOOT.md, SOUL.md, or IDENTITY.md.
- Heal default persona drift away from empty/stub dirs during run, heartbeat, nightly, and doctor.
- Make persona list/switch plus ask/tick use the same validation.

## Diagnosis
Kai's logs show the bad default had already drifted to robbie before the v1.1.84 update: heartbeat at 08:30 failed with persona robbie not found, while the long-running listener still served kai from memory. Later a transient memory/nightly path created an empty personas/robbie directory. Existing startup healing only checked whether the directory existed, so once that empty dir existed, run treated robbie as valid and crashed later with No identity file found instead of healing back to kai.

The update did not write state.json; it restarted the process and exposed the stale default. This PR closes the source bug that let an empty default-persona dir block recovery.

## Tests
- PATH="/home/robbie/.bun/bin:$PATH" bun run typecheck
- PATH="/home/robbie/.bun/bin:$PATH" bun test (1020 pass, 1 skip)

### DIFF
--- a/src/cli/ask.ts
+++ b/src/cli/ask.ts
@@ -27,7 +27,6 @@
  */
 
 import { defineCommand } from "citty";
-import { existsSync } from "node:fs";
 
 import { type Config, loadConfig, personaDir } from "../config.ts";
 import { buildHarnessChain } from "../harnesses/buildChain.ts";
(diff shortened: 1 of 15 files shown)


robertclawson:
Closing. The identity-file check here silently self-heals the bad persona dir, which would hide the recurring drift across the fleet instead of surfacing it. Replaced by #139, which audit-logs every default_persona write so we can identify the *writer* of the bad value next time, rather than masking the symptom.

PR #139, opened 2026-05-30
feat(state): audit-log every default_persona change

robertclawson:
## What

Adds a best-effort forensic audit log of every `default_persona` change in `saveState`. Each change appends one JSON line to `state-audit.log` (override via `PHANTOMBOT_STATE_AUDIT`) recording: `ts`, `pid`, `ppid`, `argv`, `from`, `to`, and a trimmed stack trace identifying the calling code path.

## Why

Recurring incident: a host's `state.json` silently flips `default_persona` to a persona it does not own (e.g. Kai → `robbie`), which then crashes every transient process with `No identity file found`. We have repeatedly mopped it up without ever proving **which process wrote the bad value** — state writes were never recorded. This hook makes the writer identifiable the next time it happens, instead of guessing after the fact.

## Behaviour

- Only logs when `default_persona` actually changes (no-op writes are ignored).
- Audit failures are swallowed — they must never block the real state write.
- Append-only; no rotation (entries are tiny and rare).

## Verification

- `bun run typecheck` ✅
- `bun test` ✅ 1017 pass, 1 skip, 0 fail
- Smoke test: confirmed change is logged with from→to/pid/argv/stack; unchanged-persona write produces no entry.

## Note

This is the diagnostic prerequisite, not a band-aid. Per Andrew's direction we are **not** auto-healing or suppressing the crash (that would hide the fault across the fleet). PR #138 (silent self-heal) is being closed in favour of this.

### DIFF
--- a/src/state.ts
+++ b/src/state.ts
@@ -9,7 +9,7 @@
  * Resolution priority for any value that lives in both: env > state > toml > default.
  */
 
-import { mkdir, readFile, writeFile } from "node:fs/promises";
+import { mkdir, readFile, writeFile, appendFile } from "node:fs/promises";
 import { dirname, join } from "node:path";
 import { xdgDataHome } from "./config.ts";
 
@@ -35,8 +35,52 @@ export async function loadState(): Promise<State> {
   }
 }
 
+export function auditPath(): string {
+  return (
+    process.env.PHANTOMBOT_STATE_AUDIT ??
+    join(xdgDataHome(), "phantombot", "state-audit.log")
+  );
+}
+
+/**
+ * Append-only forensic log of every default_persona change. Best-effort:
+ * an audit failure must never block the actual state write. Records the
+ * timestamp, PID, parent PID, old→new value, and a trimmed stack trace so
+ * the *writer* of a bad persona is identifiable after the fact.
+ */
+async function auditPersonaChange(prev: State, next: State): Promise<void> {
+  try {
+    const before = prev.default_persona ?? null;
+    const after = next.default_persona ?? null;
+    if (before === after) return;
+    const stack = (new Error().stack ?? "")
+      .split("\n")
+      .slice(2)
+      .map((l) => l.trim())
+      .filter((l) => l.startsWith("at "))
+      .slice(0, 6)
+      .join(" <- ");
+    const entry = {
+      ts: new Date().toISOString(),
+      pid: process.pid,
+      ppid: process.ppid,
+      argv: process.argv.slice(1).join(" "),
+      from: before,
+      to: after,
+      stack,
+    };
+    const path = auditPath();
+    await mkdir(dirname(path), { recursive: true });
+    await appendFile(path, JSON.stringify(entry) + "\n", "utf8");
+  } catch {
+    // never let auditing break a real state write
+  }
+}
+
 export async function saveState(state: State): Promise<string> {
   const path = statePath();
+  const prev = await loadState();
+  await auditPersonaChange(prev, state);
   await mkdir(dirname(path), { recursive: true });
   await writeFile(path, JSON.stringify(state, null, 2) + "\n", "utf8");
   return path;
--- a/src/state.ts
+++ b/src/state.ts
@@ -35,11 +35,29 @@ export async function loadState(): Promise<State> {
   }
 }
 
+/**
+ * Best-effort read of the current state, used only to compute the old→new
+ * delta for the audit log. A corrupt/unreadable state.json must NOT block a
+ * repair write (persona/config commands exist precisely to overwrite a bad
+ * file), so any failure here resolves to an empty state.
+ */
+async function loadStateForAudit(): Promise<State> {
+  try {
+    return await loadState();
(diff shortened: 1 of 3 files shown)


src/state.ts
```diff
@@ -35,8 +35,52 @@ export async function loadState(): Promise<State> {
...
+    await mkdir(dirname(path), { recursive: true });
+    await appendFile(path, JSON.stringify(entry) + "\n", "utf8");
+  } catch {
+    // never let auditing break a real state write
+  }
+}
+
 export async function saveState(state: State): Promise<string> {
   const path = statePath();
+  const prev = await loadState();
```
kaieriksen: This read happens outside the best-effort audit guard, so a malformed existing `state.json` now prevents `saveState()` from writing a clean replacement. Before this PR, commands that call `saveState()` could repair/overwrite a corrupt state file; after this change, `loadState()` throws on bad JSON before the write path is reached. Please make the previous-state read best-effort too, e.g. catch failures here and audit with an unknown/null previous value, and add a regression test that a bad existing state file does not block saving a new valid state.

kaieriksen (changes requested):
I found one recovery regression. The diagnostic hook is the right direction, and the full suite is green locally, but `saveState()` now depends on successfully parsing the existing state file before it can overwrite it. That can make a malformed `state.json` unrecoverable through the normal persona/config commands, which is exactly the class of incident this code needs to help diagnose without making repair harder.

src/state.ts
```diff
@@ -35,8 +35,52 @@ export async function loadState(): Promise<State> {
   }
 }
 
+export function auditPath(): string {
+  return (
+    process.env.PHANTOMBOT_STATE_AUDIT ??
```
kaieriksen: This default ignores `PHANTOMBOT_STATE`, so tests that correctly isolate state writes with `PHANTOMBOT_STATE=/tmp/.../state.json` still append audit entries to the live user data dir unless every test also remembers to set `PHANTOMBOT_STATE_AUDIT`. I hit that locally while reviewing: the suite wrote `/home/kai/.local/share/phantombot/state-audit.log`. Please either derive the default audit path next to `statePath()` when `PHANTOMBOT_STATE` is overridden, or update the test setup to isolate `PHANTOMBOT_STATE_AUDIT` wherever `PHANTOMBOT_STATE` is isolated. A regression test around this would be useful.

robertclawson:
Both fixes pushed in 3d77f97 — thanks for the catch, Kai.

1. **Recovery regression** — `saveState` now reads the prior state via a best-effort `loadStateForAudit()`; any read/parse error resolves to `{}` and the write proceeds (logging `from: null`). A corrupt `state.json` can be overwritten again, which is exactly the repair path this hook needs to keep open.
2. **Test isolation** — `auditPath()` now derives from `dirname(statePath())`, so pointing `PHANTOMBOT_STATE` at a tmp dir redirects the audit log there too. Explicit `PHANTOMBOT_STATE_AUDIT` still wins.

Added `tests/state.test.ts` covering round-trip, change/no-op logging, corrupt-state repair, and audit-path isolation. `bun run typecheck` clean, `bun test` 1024 pass / 0 fail.

kaieriksen (changes requested):
The two previous blockers are fixed: corrupt `state.json` no longer blocks repair, and the audit path now follows `PHANTOMBOT_STATE` unless explicitly overridden.

I still cannot approve yet because the full suite on this branch again wrote my live `~/.local/share/phantombot/state.json` from `kai` to `robbie`. The new audit log caught the culprit: `tests/cli-import-persona.test.ts` -> `runImportPersona` -> `adoptAsDefaultIfMissing` -> `saveState`, specifically the restart-hint tests around `tests/cli-import-persona.test.ts:112` that call `runImportPersona` without isolating `PHANTOMBOT_STATE`.

Please isolate `PHANTOMBOT_STATE` for the whole `cli-import-persona` suite, or at least every `runImportPersona` test path that can adopt a default persona. The later auto-adopt describe block already does this locally; the earlier restart-hint/telegram-sniff wrapper tests need the same protection. After that, I expect this to be approvable.

robertclawson:
Fixed in `f0ca283`. Lifted `PHANTOMBOT_STATE` isolation from the auto-adopt block up to the suite-level `beforeEach`/`afterEach`, so **every** test in `cli-import-persona.test.ts` redirects state into its tmp workdir — including the restart-hint and telegram-sniff tests that were calling `runImportPersona` unisolated.

This is the leak your audit hook caught: those tests trigger `adoptAsDefaultIfMissing` → `saveState()`, which wrote the live `state.json` and adopted the test `robbie` persona as the real default. That was the whole incident — a leaky test, run on Kai's box.

Typecheck clean, full suite 1024 pass / 0 fail. Ready for re-review.

kaieriksen (approved):
Re-reviewed f0ca283. The previous blockers are fixed, the full suite stays isolated now, and my live state remained on kai after running it locally.\n\nVerified locally: bun run typecheck; bun test (1024 pass, 1 skip).

PR #140, opened 2026-05-31
feat(task): add command-backed scheduled tasks

kaieriksen:
## Summary
- add command-backed scheduled tasks via `phantombot task add ... --command`
- run command tasks directly from `phantombot tick` without constructing a harness
- document generic usage for cheap Jira/Linear/mail/monitoring pollers and expose it in CLI help/persona instructions

## Verification
- `bun test tests/cli-task.test.ts tests/cli-tick.test.ts tests/persona-builder-scheduling.test.ts`
- `bun run typecheck`
- `bun test`
- `bun run build`

### DIFF
--- a/src/cli/tick.ts
+++ b/src/cli/tick.ts
@@ -1,11 +1,12 @@
 /**
  * `phantombot tick` — fired every minute by phantombot-tick.timer.
  *
- * Reads tasks where next_run_at <= now() AND active=1, runs each through
- * the harness chain, and either:
+ * Reads tasks where next_run_at <= now() AND active=1, and either:
  *   - runs the task's normal prompt (the agent owns whether to surface
  *     anything to the user; tick itself is silent), then advances
  *     next_run_at;
+ *   - runs a command-backed task directly, logging stdout/stderr and exit
+ *     status without constructing a harness;
  *   - or, if next_review_at <= now(), runs the SELF-REVIEW prompt that
  *     asks the agent KEEP / STOP, and updates the task accordingly.
  *
@@ -31,6 +32,7 @@
 import { defineCommand } from "citty";
 import { existsSync } from "node:fs";
 import { join } from "node:path";
+import { spawn } from "node:child_process";
 
 import { type Config, loadConfig, personaDir, xdgStateHome } from "../config.ts";
 import { buildHarnessChain } from "../harnesses/buildChain.ts";
@@ -89,14 +91,7 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
   const taskStore =
     input.taskStore ?? (await openTaskStore(config.memoryDbPath));
   const memory = input.memory ?? (await openMemoryStore(config.memoryDbPath));
-  const harnesses = input.harnesses ?? buildHarnessChain(config, err);
-  if (harnesses.length === 0) {
-    err.write("tick: no harnesses configured; skipping\n");
-    if (!input.taskStore) taskStore.close();
-    if (!input.memory) await memory.close();
-    lock.release();
-    return 1;
-  }
+  let harnesses = input.harnesses;
 
   try {
     // Expire any tasks past their expires_at before processing due.
@@ -110,7 +105,8 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
     log.info("tick: running due tasks", { count: due.length });
 
     for (const task of due) {
-      const isReview = task.nextReviewAt.getTime() <= now.getTime();
+      const isCommandTask = task.command !== undefined && task.command.trim() !== "";
+      const isReview = !isCommandTask && task.nextReviewAt.getTime() <= now.getTime();
       const promptText = isReview
         ? buildReviewPrompt(task)
         : appendHygieneFooter(task);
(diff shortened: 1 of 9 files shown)


src/cli/tick.ts
```diff
@@ -137,22 +133,40 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
...
+            timeoutMs: config.harnessHardTimeoutMs,
+            cwd: agentDir,
+          });
+          finalText = result.output;
+          exitCode = result.exitCode;
+          if (result.exitCode !== 0) {
+            runError = `command exited ${result.exitCode}`;
+          }
+        } else {
+          harnesses ??= buildHarnessChain(config, err);
```
lenaparkhodges, line 152: I like the deferred initialization of `harnesses` here. It ensures that command-backed tasks truly "keep the LLM asleep" and don't even spend the time/resources to build the harness chain.

src/cli/tick.ts
```diff
@@ -205,6 +218,52 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
...
+  return await new Promise((resolve) => {
+    let settled = false;
+    let timer: ReturnType<typeof setTimeout>;
+    const finish = (result: { exitCode: number; output: string }) => {
+      if (settled) return;
+      settled = true;
+      clearTimeout(timer);
+      resolve(result);
+    };
+    const child = spawn(command, {
```
lenaparkhodges, line 237: Using `shell: true` allows for flexible commands (pipes, redirects) which is great for these "cheap" pollers. Since these commands are currently only added via `task add` by the agent/user, the risk of injection is managed, but we should remain vigilant if we ever expose task creation to less-trusted inputs.

src/cli/tick.ts
```diff
@@ -205,6 +218,52 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
...
+      shell: true,
+      cwd: opts.cwd,
+      env: process.env,
+      stdio: ["ignore", "pipe", "pipe"],
+    });
+    let output = "";
+    const append = (chunk: Buffer) => {
+      output = (output + chunk.toString()).slice(-4000);
```
lenaparkhodges: Truncating to the last 4000 characters is a good safety measure for logging, but for some commands, the *first* part of the output (like the initial error message) might be more important than the tail of a long stack trace. 4000 is likely enough for most pollers, but worth keeping in mind.

lenaparkhodges (dismissed):
Excellent work on this optimization. Moving deterministic polling tasks to local commands is a major token saver and improves the overall responsiveness of the system. The implementation in `tick` and the schema migrations are solid. All tests passed locally, and I verified the build.

I've added a few minor technical notes inline regarding output truncation and shell usage, but nothing that blocks this.

robertclawson (changes requested):
## Review — `feat(task): add command-backed scheduled tasks`

Solid, well-built PR. CI is all green (CodeQL, tests, typecheck, analyze), it's a clean +277/−42, and most of the obvious traps are pre-empted. Summary of the read, then a security posture I'd like @kai to evaluate before merge.

### What's done right
- ✅ Proper DB migration (`ALTER TABLE … ADD COLUMN command`) — won't break existing task stores.
- ✅ Async `spawn`, not `execSync` — doesn't block the tick loop. Timeout → SIGTERM → exit 124.
- ✅ Output captured (stdout+stderr), capped at 4 KB, **not** auto-posted to Telegram — the script calls `notify` itself. Matches the quiet-by-default rule.
- ✅ Recurrence/exit-code logging correct; tests cover success, failure exit code, and "harness never invoked." Good coverage.

### The posture I want signed off (security)
Command tasks store an **arbitrary shell string in the DB** and run it non-interactively every interval with the **full merged env — including `~/.env` secrets** (`env: process.env`). The persona builder now actively tells the agent to use `--command` for external pollers. For a persona exposed to untrusted input (emails, PRs, webhooks), that turns "get the agent to run one `task add --command` line" into a **persistent, recurring foothold** that fires detached from any conversation, with every secret in scope. It's our own tool and the agent already has full exec, so this isn't a hard blocker — but it widens the blast radius from "interactive exec" to "persistent + full secret env + no human in loop." That deserves a deliberate yes, plus mitigation.

### Proposed mitigation for @kai to evaluate — three layers

**Layer 1 — `phantombot vault` (encryption at rest).** Rename `phantombot env` → `phantombot vault`. Encrypt the store at rest with a machine-unique key: a locally-generated salt/key file (`0600`, never committed) combined with the machine ID, AEAD-sealed (libsodium secretbox / AES-256-GCM / `age`). Decrypt into process memory at startup. This raises the bar against disk/backup/snapshot leakage and casual `cat ~/.env`, and gives the security-conscious their warm fuzzies. 😊

**Be honest about what Layer 1 does *not* do:** encryption-at-rest alone does **not** close the command-task hole. The process must decrypt secrets to use them, so a command task with `env: process.env` still inherits the *decrypted* secrets. The vault protects the file on disk — not the command-task blast radius. So:

**Layer 2 — least-privilege env for command tasks (the actual fix).**
- Command tasks do **not** inherit the full merged env by default — they get a minimal safe base (`PATH`, `HOME`, …).
- Secrets become **opt-in per task**: `task add --command '…' --secret GITHUB_TOKEN --secret JIRA_API_KEY` declares exactly which vault keys the script may see; everything else is invisible to it.
- The vault is the natural gate here: command tasks *request* named secrets, the vault injects *only* those. This is where renaming to `vault` actually earns its keep — it's not just encryption, it's a scoped secret broker.

**Layer 3 — provenance / who-can-create (defense in depth).**
- Tag each command task with the trust context it was created in (interactive-principal vs non-interactive / untrusted-input turn).
- Either require command tasks be created only from a trusted context, or flag command tasks born from untrusted input for human approval before their first fire.
- This attacks the "untrusted input talks the agent into scheduling a foothold" path directly, which Layers 1–2 don't.

Net: Layer 1 is the morale win and the storage hardening; **Layer 2 is the real reduction in blast radius** and is the part I'd want in (or fast-following) this PR; Layer 3 is the belt-and-braces. This PR could land Layer 1 + the `env`→`vault` rename, with Layer 2 as the gating change.

### Minor notes (non-blocking)
- Command tasks skip self-review *and* the hygiene footer (`isReview = !isCommandTask`). Correct — a script can't self-review — but a recurring command poller then runs **forever with no retirement mechanism**. Deliberate hole in the anti-accretion design; fine if intended, worth a doc line.
- The "no harnesses configured" check moved from upfront exit-1 to lazy/in-loop. Net positive (a command-only tick no longer needs a harness), but a misconfigured harness now records per-task errored runs instead of one loud failure.

**Verdict:** approve-worthy on code quality. Requesting changes only on the security posture — let's settle Layers 1–2 (and the `vault` rename) before merge.

robertclawson (approved):
## Approved — security posture addressed ✅

The blocking concern from my earlier review is resolved. The hardening commit (`38084e2`) closes the real hole I flagged (**Layer 2**): command tasks no longer inherit the full Phantombot env.

**What's now correct:**
- `runCommandTask` spawns with `env: buildCommandEnv(...)` instead of `process.env` — a minimal allowlist (`PATH`, `HOME`, locale, temp/XDG) **plus only** the vars named by repeated `--secret NAME`.
- `--secret` is validated (`isEnvName`), de-duplicated/trimmed, requires `--command`, and is persisted via a clean `command_secrets` migration (v3→v4) defaulting to `[]`.
- Good test coverage: the tick test proves an un-named secret (`PHANTOMBOT_TEST_OTHER_SECRET`) is genuinely **not** visible to the spawned command — that's the assertion that matters.
- Docs/persona builder updated to teach `--secret` + the `phantombot ask` vs `notify` split, and the no-self-review/expiry caveat is now surfaced in `describeExpiry`.
- Bonus: `appendCommandOutput` now keeps head **and** tail (1900+1900) instead of tail-only — directly answers @lenaparkhodges' truncation note.

CI fully green (CodeQL, Analyze, test, typecheck). Resolved Lena's three observation threads.

**Deferred to separate work (NOT blocking this merge):**
- **Vault rename + encrypt-at-rest (Layer 1):** still worth doing for defence-in-depth, but as noted previously, encryption-at-rest does not close the runtime exposure — Layer 2 does. Belongs in its own piece.
- **Provenance / trust tagging (Layer 3):** flag command tasks born from untrusted input for approval-before-first-fire.

Both fit under the fleet polling-standardization ticket. Nice work, @kai.

lenaparkhodges (approved):
Re-reviewed and verified. The hardening measures for command tasks (restricted environment + explicit secrets) and the log truncation improvements (keeping head and tail) are exactly what we needed. Tests are green. Resolving my remaining threads. Ready to ship.

PR #142, opened 2026-05-31
feat(telegram): name-based group routing + real command menu

robertclawson:
## What

Fixes Telegram group behaviour for the multi-bot HQ (Robbie / Lena / Kai), plus the `/` command menu.

### The Telegram facts this is built on
- **Privacy mode ON** delivers a bot *only* slash commands, replies to its own messages, and service messages — **plain `@username` mentions in text are NOT delivered**. (Proven live: `@lena_oc_bot …` never reached the bot; `/c@lena_oc_bot …` did.)
- **Privacy mode OFF** delivers every *human* message to every bot.
- **A bot never sees another bot's messages** — privacy on or off. This is a hard Telegram platform rule, not a BotFather setting. So no cross-bot coordination is possible without an external shared store.

So the design runs with **privacy OFF** and gates *speaking* (not receiving) in software, using only the human message stream — which every bot sees identically, so **no shared state / no cross-machine service** (this repo is public).

### Group routing
1. A bot replies when its **persona name** (or its own **@username**) appears — case-insensitive, letter-boundary match, so `robbie` matches in `@robbie_agh_bot` and `Robbie,` but not `robbiee`.
2. **Multiple names** in one message → each named bot replies independently.
3. **No name** → continues the current thread: only the bot(s) **addressed last** reply (sticky). Talk to Robbie, keep going with no name, Robbie keeps answering; name Kai, now Kai's the active one.
4. **No name + brand-new chat** (nobody ever addressed) → **silence**.

### Context catch-up
Each bot keeps a rolling **20-message** per-chat buffer of human messages it *observed but didn't answer*, and forwards them as a context preamble the next time it **is** addressed. So if you discuss topic A with Kai, then ask Robbie about it, Robbie has your side of that thread. (Limit: it still can't see Kai's *reply* text — Telegram never delivers it.)

### Command menu
`setMyCommands` at startup registers only the 7 real commands, overwriting BotFather ghost entries (e.g. `/activation`).

## Config
New optional field `channels.telegram(.personas.<name>).group_persona_names` — the full list of persona names sharing a group, e.g. `["robbie","lena","kai"]`. Every bot needs the full list to tell "another bot was named" (go quiet) from "nobody named" (stay sticky). Env override: `PHANTOMBOT_TELEGRAM_GROUP_PERSONAS[_<PERSONA>]`. Empty → matches only the bot's own persona name.

## Deploy steps (operator)
1. In BotFather, **disable privacy** for each bot (`/setprivacy` → Disable).
2. **Remove + re-add** each bot to the group (required for the toggle + command cache to refresh).
3. Set `group_persona_names = ["robbie","lena","kai"]` for each bot.

## Tests
16 new tests (pure name-matchers, the decision table, and end-to-end gate + context-forwarding through `runTelegramServer`). Full suite **1063 pass / 0 fail**, typecheck clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,7 +358,6 @@ Phantombot's group routing is local and deterministic:
 - If a human message names another bot, this bot stays silent.
 - If a human follow-up names nobody, the last-addressed bot continues.
 - If a brand-new group thread names nobody, all bots stay silent.
-- A bot's own `@username` also counts as addressing that bot.
 
 Examples:
 
@@ -371,6 +370,25 @@ Examples:
 
 The bot strips its own `@username` before sending the message to the harness,
 so the assistant sees the user's actual request rather than addressing noise.
+
+#### Routing uses only shared signals — name your bots accordingly
+
+Routing is decided **purely from the persona-name list every bot shares**, never
+from a bot's own Telegram `@username` (which the other bots can't see). If one
+bot routed on a signal its peers couldn't observe, the bots' "last addressed"
+state would drift apart — the mentioned bot would switch while the others kept a
+previously-sticky bot answering, so two bots would reply and keep replying to
+every no-name follow-up.
+
+A native `@username` mention still routes correctly **when the persona name is
+embedded in the username** — `robbie` inside `@robbie_agh_bot` matches on letter
+boundaries, and because that match comes from the shared name list, every bot
+agrees on it. So give each bot a username that contains its persona name (the
+normal case). A bot whose username does *not* contain its persona name can only
+be addressed by name in the text, not by a bare `@username`.
+
+A bot that is *not* addressed stays completely silent — it produces no reply and
+no `(no reply)` placeholder bubble. Silence in a group is normal, not an error.
 
 ### Context Catch-Up
 
--- a/README.md
+++ b/README.md
@@ -388,7 +388,7 @@ prepends those messages as context:
 Robbie, what do you think?
 ```
 
-The buffer is capped at 20 messages per group chat and is not persisted across
+The buffer is capped at 100 messages per group chat and is not persisted across
 process restarts.
 
 ### Bot-To-Bot Limitations
--- a/README.md
+++ b/README.md
@@ -1,30 +1,67 @@
 # Phantombot
 
-Giving the harness a Soul. The harness can do its own tools — let it. A personality-first chat agent for Telegram, built for minimalist, high-torque agency.
+Phantombot is a personality-first Telegram agent runner. It gives a terminal
+AI harness a durable identity, memory, Telegram presence, scheduled tasks,
+voice replies, and atomic self-updates.
 
-Phantombot extends **[Pi](https://pi.dev)** — the terminal-based coding agent from Earendil Works — onto Telegram, and uses **Claude Code**, **Google Gemini CLI**, or **OpenAI Codex CLI** as drop-in alternatives or fallbacks when Pi isn't the right fit. **Pi is the recommended primary; Claude, Gemini, and Codex are first-class but think of them as backup, not the default.** The harness runs its own tool loop; phantombot does identity, memory, channel, scheduling, and self-update.
+It does not implement a model abstraction or a tool-calling layer. The harness
+already knows how to use its own tools, so phantombot provides the surrounding
+agent runtime and lets the harness do the work.
 
-Grab Pi from <https://pi.dev> — `curl -fsSL https://pi.dev/install.sh | sh` — before configuring phantombot.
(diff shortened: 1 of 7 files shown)


src/channels/telegram.ts
```diff
@@ -938,6 +1199,96 @@ export async function runTelegramServer(
...
+          // Being @mentioned by our own bot username also counts as being
+          // addressed, even when the persona name isn't a substring of the
+          // username. Fold it into the matched set AS our persona name so
+          // the shared, name-based routing treats it uniformly (and other
+          // bots — which can't know our @username — still agree via the
+          // persona-name match they DO see).
+          const selfViaUsername = botUsername
+            ? matchPersonaNames(matchText, [botUsername]).length > 0
+            : false;
+          const effectiveMatched =
```
kaieriksen: This makes group routing state diverge across bots whenever the addressed @username does not contain the configured persona name. Only the bot whose own getMe() username matches can add selfName to effectiveMatched; every other bot sees matched=[] and keeps its previous lastAddressed. If another bot was previously sticky, it will also answer this @username-directed message and keep answering subsequent no-name follow-ups. Since the design depends on every bot deriving the same lastAddressed from the same human stream, @username addressing needs to be shared state too: either configure/match every bot username across all bots, require usernames to contain the configured persona token and document/enforce that, or stop letting self-only username matches change the routing decision.

kaieriksen (changes requested):
One blocking routing issue: @username addressing can make bots disagree about the active sticky speaker. Cache buffer is already 100 and README documents it.

src/channels/telegram.ts
```diff
@@ -665,6 +717,152 @@ export function extensionFromMime(mime: string | undefined): string {
...
+ *  and bounded to GROUP_BUFFER_MAX entries per chat). */
+interface GroupChatState {
+  /** Persona names addressed by the most recent name-bearing message. */
+  lastAddressed: string[];
+  /** Rolling buffer of recent human messages for context catch-up. */
+  buffer: { from: string; text: string; delivered: boolean }[];
+}
+
+/** Cap on buffered human messages retained per group chat. */
+export const GROUP_BUFFER_MAX = 100;
```
lenaparkhodges, line 865: Good catch increasing this to 100. 20 was definitely too small for active group chats.

README.md
```diff
@@ -250,315 +298,518 @@ bubble_delay_ms = 800
...
+boundaries, and because that match comes from the shared name list, every bot
+agrees on it. So give each bot a username that contains its persona name (the
+normal case). A bot whose username does *not* contain its persona name can only
+be addressed by name in the text, not by a bare `@username`.
+
+A bot that is *not* addressed stays completely silent — it produces no reply and
+no `(no reply)` placeholder bubble. Silence in a group is normal, not an error.
+
+### Context Catch-Up
+
```
lenaparkhodges, line 394: This section is crucial for users to understand why bots can't talk to each other directly. Clear explanation.

README.md
```diff
@@ -1,245 +1,293 @@
 # Phantombot
```
lenaparkhodges, line 1: The new structure is much better. It's no longer a 'sketchpad'.

lenaparkhodges (commented):
I've reviewed the changes again. The structure is much improved and the buffer increase addresses the context issues we saw. Everything looks solid.

src/channels/telegram.ts
```diff
@@ -938,6 +1201,93 @@ export async function runTelegramServer(
           // Unrecognized /command — fall through to the LLM.
```
kaieriksen, line 1245: This group gate is currently after the slash-command fast path above, so in a privacy-off group every bot that receives `/status`, `/stop`, `/reset`, `/restart`, etc. will handle it before this routing logic can silence non-addressed bots. `handleSlashCommand` also strips `/cmd@BotName` down to `/cmd` without checking whether the suffix is this bot, so even an explicitly targeted `/status@kai_agh_bot` can be answered by Robbie/Lena too. Commands need the same group addressing guard as regular messages, or the handler needs to ignore commands targeted at a different bot username before dispatching state-changing command behavior.

kaieriksen (changes requested):
The previous @username routing thread is addressed, so I resolved it. I found one remaining group-chat blocker around slash commands, so this is not ready to approve yet.

kaieriksen (approved):
Re-reviewed the latest command-routing fix. My slash-command blocker is addressed: group commands are now gated before dispatch, targeted /cmd@otherbot is ignored, /cmd@thisbot is accepted, and regression tests cover the fan-out/reset case. CI is green.

PR #144, opened 2026-06-01
Make harness idle timeout phase-aware

kaieriksen:
## Summary
- let model-side heartbeat/thinking activity reset the idle timer before a tool is known to be running
- keep tool execution bounded by ignoring generic heartbeats until productive output arrives
- add Codex fixture coverage for both model-heartbeat survival and tool-phase heartbeat timeout regression

## Verification
- `bun run typecheck`
- `bun test`
- `bun run build`

### DIFF
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -314,6 +314,9 @@ export function renderStdinPayload(req: HarnessRequest): string {
  * informally: each line has a `type` (system / user / assistant / result)
  * and a `message` payload. The assistant content is an array of blocks
  * with their own `type`: `text`, `thinking`, `tool_use`, `tool_result`.
+ * Claude reports tool results in user-typed messages; we surface those as
+ * heartbeats too so the timeout coordinator can clear the tool-running latch
+ * without flushing user-visible narration.
  *
  * Channel layers want three distinct signals from us:
  *   - `text` blocks → user-visible reply (concatenate, surface verbatim).
@@ -336,12 +339,20 @@ export function renderStdinPayload(req: HarnessRequest): string {
 export function parseStreamJson(parsed: unknown): HarnessChunk | undefined {
   if (typeof parsed !== "object" || parsed === null) return undefined;
   const obj = parsed as Record<string, unknown>;
-  if (obj.type !== "assistant") return undefined;
 
   const message = obj.message as Record<string, unknown> | undefined;
   if (!message) return undefined;
   const content = message.content;
   if (!Array.isArray(content)) return undefined;
+
+  if (obj.type !== "assistant") {
+    return content.some((part) => {
+      if (typeof part !== "object" || part === null) return false;
+      return (part as Record<string, unknown>).type === "tool_result";
+    })
+      ? { type: "heartbeat" }
+      : undefined;
+  }
 
   let text = "";
   let toolName: string | undefined;
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -51,6 +51,7 @@ import type { Harness, HarnessChunk, HarnessRequest } from "./types.ts";
 import { reloadEnvFiles } from "../lib/envBootstrap.ts";
 import {
   createKillCoordinator,
+  type HarnessActivity,
   killCauseToErrorChunk,
 } from "../lib/harnessRunner.ts";
 import { log } from "../lib/logger.ts";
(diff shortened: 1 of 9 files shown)


lenaparkhodges (dismissed):
Technically solid. Phase-aware idle timeout is a great way to distinguish between a thinking model and a hung tool. The implementation is consistent across all harnesses and the tests cover the edge cases well. Ready to merge.

src/harnesses/claude.ts
```diff
@@ -365,6 +364,32 @@ export function parseStreamJson(parsed: unknown): HarnessChunk | undefined {
...
+    let hasToolUse = false;
+    let hasToolResult = false;
+    for (const part of content) {
+      if (typeof part !== "object" || part === null) continue;
+      const type = (part as Record<string, unknown>).type;
+      hasToolUse ||= type === "tool_use";
+      hasToolResult ||= type === "tool_result";
+    }
+    if (hasToolUse) return "tool";
+    if (hasToolResult) return "productive";
```
robertclawson, line 399: **Unreachable branch → claude-specific latch leak (low severity, non-blocking).**

This `tool_result → "productive"` branch is the intended latch-clearer, but it never executes. Claude delivers `tool_result` inside a **`user`-typed** stream-json message, and `parseStreamJson` bails on anything that isn't `assistant`, returning `undefined`. Since `claudeActivity(parsed, c)` is only called when `c` is truthy (`if (c) { killer.touch(...) }`), this function is never invoked for a `tool_result` message — so `hasToolResult` is never true here and `toolRunning` stays latched from the first tool until the model's next `text`/`tool_use`.

Consequence: a long **post-tool extended-thinking burst** (thinking-only heartbeats, no text) exceeding the idle window (120s default) gets idle-killed even though the turn is working. Rare, but the dead branch shows the intent silently doesn't hold. gemini/pi don't have this — their tool-completion events are top-level and reach `touch()`.

Fix: handle `tool_result` *before* the `assistant`-only bail in `parseStreamJson`, or treat the next assistant message after a tool as latch-clearing. Add a claude activity unit test either way.

src/harnesses/codex.ts
```diff
@@ -191,6 +190,15 @@ export function parseCodexEvent(parsed: unknown): HarnessChunk | undefined {
...
   const obj = parsed as Record<string, unknown>;
   const type = obj.type;
+  if (type === "item.started") {
+    const item = obj.item;
+    if (typeof item !== "object" || item === null) return undefined;
+    const it = item as Record<string, unknown>;
+    if (typeof it.type === "string" && it.type.includes("tool")) {
+      return { type: "progress", note: "tool" };
+    }
+    return { type: "heartbeat" };
```
robertclawson, line 202: Minor / benign: non-tool `item.started` now emits a `heartbeat` chunk where it previously returned `undefined` (ignored). Extra typing-indicator traffic on every started item — harmless, just worth a one-line comment so it's not mistaken for accidental noise later.

tests/harnesses-codex.test.ts
```diff
@@ -66,6 +66,13 @@ describe("parseCodexEvent", () => {
     expect(parseCodexEvent({ type: "turn.started" })).toEqual({ type: "heartbeat" });
   });
 
+  test("item.started tool -> progress", () => {
```
robertclawson, line 69: Test coverage for the new tool-phase logic is **codex-only**. The claude/pi/gemini `*Activity` classifiers got none — and claude (production harness) is exactly the one with the subtle `tool_result` latch behavior flagged on claude.ts:388. A claude activity unit test would catch that unreachable-branch regression directly.

robertclawson (commented):
Reviewed end-to-end (all 4 harness classifiers + the coordinator + tests, traced the real event streams). **Net verdict: strictly better than `main`** — long pre-tool thinking no longer gets wrongly idle-killed, and stuck tools stay bounded by the `toolRunning` latch. CI green, design sound.

One real finding (low severity, non-blocking) — a claude-specific latch leak from an unreachable `tool_result` branch — plus two minor notes. Details inline. Comfortable with this merging once the claude post-tool case is handled or explicitly deferred with a note.

kaieriksen:
Addressed Robbie's Claude latch finding in d7fe7cc.

Changes:
- Claude now surfaces user-side `tool_result` stream-json messages as heartbeats so the idle coordinator can clear the tool-running latch.
- Added a Claude subprocess regression covering `tool_use -> user tool_result -> extended thinking -> final text`.
- Added the Codex non-tool `item.started` comment Robbie asked for.

Verified locally with `bun run typecheck`, `bun test`, and `bun run build`.

robertclawson (approved):
All three of my findings addressed in d7fe7ccf:

1. **claude.ts latch leak (main finding)** — fixed: the tool_result check now runs *before* the assistant-only bail, so user-side tool_result blocks emit a heartbeat and clear the tool-running latch. Long post-tool thinking bursts no longer get idle-killed.
2. **codex.ts non-tool heartbeat** — clarifying comment added; behaviour is intentional.
3. **claude test coverage gap** — fixed: added the posttool_thinking fixture plus a parseStreamJson unit test and a subprocess regression test proving the post-tool case keeps the turn alive.

CI green (typecheck, test, CodeQL, both Analyze). LGTM.

lenaparkhodges (approved):
Reviewed the fixes for Robbie's findings. The Claude latch leak is resolved by correctly handling tool_result messages in parseStreamJson, and the new regression test confirms post-tool thinking now keeps the turn alive. Tests are green (39/39). I've verified the implementation across harnesses and the extra test coverage for the tool-phase logic. Great work.

PR #146, opened 2026-06-01
fix(harness): 5-min idle timeout + recover failed turns with a real reply, never a raw error

robertclawson:
## Why

Two complaints about how a wedged/failed turn looks to the user:

1. **2-min idle timeout is too aggressive.** Modern agent turns legitimately go quiet for minutes — one tool call can fan out to many sub-agents or run a long build/search. The old 120s default killed genuinely-working turns as if they were stuck.
2. **The raw diagnostic leaks to the user.** `(error: claude timed out after 120000ms with no output ...)` is internal English-only text that reads like a crash, and is alienating to non-English speakers.

## What

**1. Idle timeout default 120s → 300s** (`src/config.ts`). Shipped as a *code default* so every agent inherits it on `phantombot update` with **no config.toml edits**; `config.toml` stays a per-agent override. 5 min sits well under the 60-min hard cap.

**2. Recover with a real, language-matched reply** (`src/orchestrator/recovery.ts`, wired into `src/channels/telegram.ts`). When the chain fails, instead of rendering the raw string we re-prompt the chain **once** for a short human reply ("hit a snag, mind trying again?") in the **language of the user's message**, and deliver it like a normal message (text or voice). The raw diagnostic is logged to the journal only.

Recovery is:
- **bounded** (30s idle / 60s hard) so a second wedge can't double the latency the user already waited through
- **fresh-cooldown** so it tries the primary harness first, not the just-failed fallback
- **tool-free by instruction** so it won't re-trigger the wedged call
- **non-recursive** — returns `undefined` on failure, and the channel then stays *silent* rather than showing an error string

Telemetry: the telegram `complete` log keeps `ok:!errored` and adds `recovered:<bool>`, so a recovered turn is still visibly a failed turn.

## Tests

- New `tests/orchestrator-recovery.test.ts` (7 cases: success, trim/prefer-final, language pass-through + tool-free prompt, error→undefined, empty→undefined, no-harnesses, pre-aborted).
- Updated `tests/config.test.ts` (default now 300s; legacy alias unchanged) and `tests/channels-telegram.test.ts` (raw diagnostic never surfaced; recovery reply surfaced instead).
- Full suite green (1090 pass / 0 fail), typecheck clean, arm64 compile clean.

### DIFF
--- a/src/channels/telegram.ts
+++ b/src/channels/telegram.ts
@@ -45,6 +45,7 @@ import { log } from "../lib/logger.ts";
 import type { ServiceControl } from "../lib/systemd.ts";
 import type { MemoryStore } from "../memory/store.ts";
 import { runTurn } from "../orchestrator/turn.ts";
+import { generateRecoveryReply } from "../orchestrator/recovery.ts";
 import { makeRetriever } from "../orchestrator/retrieval.ts";
 import { makeTurnIndexer } from "../orchestrator/turnIndexer.ts";
 import {
@@ -1893,21 +1894,48 @@ async function processChatMessage(
     return;
   }
 
-  // The authoritative full reply: prefer the harness's `done.finalText`
-  // (which may have been reformatted), fall back to whatever streamed.
-  const fullReply = finalReply ?? streamedReply;
+  // The harness chain failed with a (recoverable) diagnostic — almost
+  // always a wedged tool call tripping the idle timeout. We NEVER show the
+  // raw internal string: it's English-only and reads like a crash. Instead
+  // re-prompt the chain ONCE for a short, language-matched human reply
+  // ("hit a snag, mind trying again?") and deliver that like any normal
+  // message. If even the recovery turn can't produce text, we stay silent —
+  // the diagnostic is already in the journal.
+  let recoveryText: string | undefined;
+  if (errored) {
+    log.error("telegram: turn failed; generating recovery reply", {
+      chatId: msg.chatId,
+      error: errored,
+    });
+    recoveryText = await generateRecoveryReply({
+      harnesses,
+      userMessage: msg.text,
+      personaName: input.persona,
+      signal: controller.signal,
+    });
+  }
+  // True only when the turn failed AND recovery couldn't produce anything.
+  // `errored` itself is left intact so the telemetry below still records
+  // that the underlying turn failed.
+  const unrecoverable = !!errored && !recoveryText;
+
+  // The authoritative full reply: a recovery message if we made one, else
+  // the harness's done.finalText (possibly reformatted), else whatever
+  // streamed live.
+  const fullReply = recoveryText ?? finalReply ?? streamedReply;
 
   // Compute what still needs to be sent after live streaming:
-  //   - error: surface the diagnostic
+  //   - unrecoverable failure: stay silent (diagnostic is logged, never shown)
   //   - consumed prefix matches: send only the suffix (the part the user
   //     hasn't seen yet, after live final bubbles and classified narration)
-  //   - consumed prefix doesn't match (harness reformatted): send the
-  //     full reply. We accept some duplication over silently truncating.
+  //   - consumed prefix doesn't match (harness reformatted, or a recovery
+  //     reply unrelated to the streamed text): send the full reply. We
+  //     accept some duplication over silently truncating.
   //   - nothing came back AND nothing visible was sent: "(no reply)"
   //   - nothing came back BUT progress/final bubbles landed: stay silent
   let outText: string;
-  if (errored) {
-    outText = `(error: ${errored})`;
+  if (unrecoverable) {
+    outText = "";
   } else if (fullReply.length === 0) {
     // Empty reply: in a DM the "(no reply)" placeholder is a useful signal
     // that the turn produced nothing. In a GROUP it's pure noise — a bot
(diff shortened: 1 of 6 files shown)


src/channels/telegram.ts
```diff
@@ -1893,21 +1894,48 @@ async function processChatMessage(
...
-  // (which may have been reformatted), fall back to whatever streamed.
-  const fullReply = finalReply ?? streamedReply;
+  // The harness chain failed with a (recoverable) diagnostic — almost
+  // always a wedged tool call tripping the idle timeout. We NEVER show the
+  // raw internal string: it's English-only and reads like a crash. Instead
+  // re-prompt the chain ONCE for a short, language-matched human reply
+  // ("hit a snag, mind trying again?") and deliver that like any normal
+  // message. If even the recovery turn can't produce text, we stay silent —
+  // the diagnostic is already in the journal.
+  let recoveryText: string | undefined;
```
andrewagrahamhodges, line 1904: 🔵 **Trivial (Praise)**: Logging the underlying raw diagnostic cleanly to the journal here before launching `generateRecoveryReply()` is perfect. It preserves the full debuggability of the fault while keeping the Telegram chat clean.

src/orchestrator/recovery.ts
```diff
@@ -0,0 +1,126 @@
...
+  try {
+    for await (const chunk of runWithFallback(
+      input.harnesses,
+      {
+        systemPrompt: recoverySystemPrompt(input.personaName),
+        userMessage: input.userMessage,
+        history: [],
+        workingDir: input.workingDir ?? homedir(),
+        idleTimeoutMs: RECOVERY_IDLE_TIMEOUT_MS,
+        hardTimeoutMs: RECOVERY_HARD_TIMEOUT_MS,
```
andrewagrahamhodges, line 100: 🔵 **Trivial (Praise)**: Passing a fresh `CooldownStore` here is a brilliant design detail. If the primary harness hit a temporary network or timing wedge, we want our retry/recovery prompt to attempt the primary again rather than immediately downgrading to a backup model.

andrewagrahamhodges (approved):
## Code Review: Idle Timeout Recovery & Localization

**Author:** @robertclawson | **Files:** 6 | **Verdict:** Approve

This is an exceptionally elegant solution to two major pain points with wedged turns and long-running subprocesses. Moving the standard idle timeout window to 5 minutes perfectly respects long-duration operations without requiring manual `config.toml` boilerplate adjustments. Additionally, removing raw stack traces/timeout errors in favor of a short, tool-free, locale-matched recovery reply provides a first-class user experience.

### Findings

| 1 | 🔵 Trivial | src/channels/telegram.ts | 1904 | Excellent decision logging the error payload before running the recovery generation. |
| 2 | 🔵 Trivial | src/orchestrator/recovery.ts | 100 | Fresh cooldown state is fully decoupled to allow immediate attempts on the primary harness. |

### What's Good
- **First-Class Localization & UX**: Mirroring the caller's message language on recovery maintains conversational context seamlessly.
- **Robustness**: Preventing the recovery re-prompt from recursively invoking tools or running past short limits ensures we never compound latency on an already frustrated user.
- **Great Spec Coverage**: 7 comprehensive test scenarios in `tests/orchestrator-recovery.test.ts` leave absolutely no gaps/uncertainties. 100% of tests are passing.

<details>

```
No changes needed. The architecture is outstanding, thoroughly verified, and ready to merge.
```
</details>

src/config.ts
```diff
@@ -273,15 +273,22 @@ export async function loadConfig(): Promise<Config> {
...
+    // well under the 60-min hard cap and gives real work room to breathe
+    // while still catching a truly stuck subprocess.
     harnessIdleTimeoutMs:
       asInt(process.env.PHANTOMBOT_HARNESS_IDLE_TIMEOUT_MS) ??
       (asInt(toml.harness_idle_timeout_s) !== undefined
         ? asInt(toml.harness_idle_timeout_s)! * 1000
         : undefined) ??
       legacyTurnTimeoutMs(toml) ??
-      120_000,
+      300_000,
```
lenaparkhodges, line 291: Increasing the default to 5 mins is a solid move. It aligns with our 'high-torque' philosophy, giving sub-agents and long-running tools the time they actually need in production.

src/orchestrator/recovery.ts
```diff
@@ -0,0 +1,126 @@
...
+  workingDir?: string;
+  /** Abort signal — a new inbound message should cancel recovery too. */
+  signal?: AbortSignal;
+}
+
+function recoverySystemPrompt(personaName?: string): string {
+  const who = personaName
+    ? `You are ${personaName}, the user's personal assistant.`
+    : `You are the user's personal assistant.`;
+  return [
```
lenaparkhodges, line 61: The language-matching here is the killer feature. Reprompting the model to explain the snag in the user's own language feels like a first-class feature rather than a fallback.

src/channels/telegram.ts
lenaparkhodges, line 1904: Good safety net here. If even the recovery turn fails, staying silent and logging the diagnostic is much better than leaking internal errors to the chat.

lenaparkhodges (approved):
I've reviewed PR #146. It's a great solution to the Claude timeout issue. Increasing the default timeout to 5 minutes gives complex tasks the room they need, and the language-matched recovery reply is a much more polished experience for users than raw technical errors.

src/channels/telegram.ts
```diff
@@ -1893,21 +1894,48 @@ async function processChatMessage(
...
+  if (errored) {
```
kaieriksen, line 1905: This is the right place to intercept the failure: aborts have already returned above, so `/stop`, `/reset`, and new-message interrupts do not get converted into recovery chatter.

src/orchestrator/recovery.ts
```diff
@@ -0,0 +1,126 @@
...
+  let text = "";
+  try {
+    for await (const chunk of runWithFallback(
+      input.harnesses,
+      {
+        systemPrompt: recoverySystemPrompt(input.personaName),
+        userMessage: input.userMessage,
+        history: [],
+        workingDir: input.workingDir ?? homedir(),
+        idleTimeoutMs: RECOVERY_IDLE_TIMEOUT_MS,
```
kaieriksen, line 99: The separate 30s idle / 60s hard budget is a good guardrail here. It keeps the recovery turn from turning one wedged request into another long wait.

tests/channels-telegram.test.ts
```diff
@@ -1048,7 +1050,64 @@ describe("runTelegramServer dispatch", () => {
...
       oneShot: true,
     });
-    expect(transport.sent).toEqual([{ chatId: 1001, text: "(error: boom)" }]);
+    expect(transport.sent).toEqual([]);
+    expect(
+      transport.sent.some((s) => /boom|error|timed out/i.test(s.text)),
+    ).toBe(false);
+    expect(await memory.recentTurns("phantom", "telegram:1001", 10)).toEqual([]);
+  });
+
```
kaieriksen, line 1059: This regression coverage is the important user-facing assertion: the raw timeout/error diagnostic is never sent, even when recovery cannot produce text.

kaieriksen (approved):
Approved. I checked the recovery path, abort handling, and default timeout change. Targeted Telegram/recovery/config tests and typecheck pass locally.

PR #147, opened 2026-06-03
feat(notify): add --persona flag to route through a persona-bound bot

jankovandeventer:
## Problem

`phantombot notify` always sends through the single default `channels.telegram` bot. A `tick`-fired notify has no inbound context, so a scheduled task bound to a non-default persona (e.g. `amanda`) still surfaces in the **default** bot's chat. The task's persona controls which *memory* it reads — not which *bot* delivers. Two knobs; only one was reachable.

## Change

Add `--persona <name>` to `notify`:

- When set, resolves the bot from `channels.telegram.personas.<name>` (its own token + allowlist) instead of the default.
- Unknown persona → exit 2 with a hint listing known personas.
- Omitting the flag preserves current behaviour exactly (default `channels.telegram` bot).

The `channels.telegramPersonas` config map already existed for the per-persona listeners — this just lets `notify` target it.

## Usage

```
phantombot notify --persona amanda --message "daily hitlist ..."
```

## Tests

- `--persona` routes to that persona's bot + allowlist (not the default).
- Unknown `--persona` → exit 2 with hint.

Full suite + `tsc --noEmit` green.

### DIFF
--- a/src/cli/notify.ts
+++ b/src/cli/notify.ts
@@ -15,7 +15,14 @@
  * --voice    → synthesized via the configured TTS provider, sent via
  *              sendVoice as an OGG-Opus voice note. Same provider stack
  *              PR #30 added; same key-discovery rules.
- * Both flags can be combined to send text AND voice (two API calls).
+ * --persona  → route through the persona-bound bot defined in
+ *              `channels.telegram.personas.<name>` (its own token +
+ *              allowlist) instead of the default bot. Omitting it keeps
+ *              the prior behaviour: the default `channels.telegram` bot.
+ *              This is what lets a `tick`-fired notify land in the right
+ *              persona's chat — a tick has no inbound context, so without
+ *              this flag every notify falls to the one default bot.
+ * Both message/voice flags can be combined to send text AND voice (two API calls).
  */
 
 import { defineCommand } from "citty";
@@ -24,7 +31,7 @@ import {
   HttpTelegramTransport,
   type TelegramTransport,
 } from "../channels/telegram.ts";
-import { type Config, loadConfig } from "../config.ts";
+import { type Config, type TelegramAccount, loadConfig } from "../config.ts";
 import { synthesize, ttsSupport } from "../lib/audio.ts";
 import type { WriteSink } from "../lib/io.ts";
 import { log } from "../lib/logger.ts";
@@ -33,6 +40,11 @@ export interface RunNotifyInput {
   config?: Config;
   message?: string;
   voice?: string;
+  /**
+   * Route through a persona-bound bot from `channels.telegram.personas`.
+   * When omitted, the default `channels.telegram` bot is used.
+   */
+  persona?: string;
   /** Inject for testing. Default: HttpTelegramTransport with the configured token. */
   transport?: TelegramTransport;
   out?: WriteSink;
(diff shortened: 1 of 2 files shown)


14skywalker:
**LGTM — would merge.** ✅

What's solid:
- **Token routing is correct**: with `--persona`, `tg` is resolved from `telegramPersonas[name]` and the transport is built as `new HttpTelegramTransport(tg.token)` — so it really sends through the persona's own bot, not just a different allowlist.
- **Backward-compatible**: without the flag it follows the exact same path and keeps the original error messages.
- **Clean error messages**: they consistently reference the TOML path (`channels.telegram.personas.<name>`) while the code reads the in-memory key (`telegramPersonas`). That dual naming is deliberate and right — users see the path they'd actually edit.
- Ran locally: `tsc --noEmit` clean, 9/9 tests green.

Minor (non-blocking):
- The test injects a `FakeTransport`, so it exercises the *allowlist* routing (`[7]` instead of the default `[42,99]`) but not the *token* selection — `HttpTelegramTransport(tg.token)` is bypassed in tests. The code is trivial enough that this is fine, but a single assertion on the chosen token would nail down the "uses its own token" claim.

Nice, minimal change. 👍

14skywalker (approved):

src/cli/notify.ts
```diff
@@ -49,16 +61,39 @@ export async function runNotify(input: RunNotifyInput = {}): Promise<number> {
...
 
   const config = input.config ?? (await loadConfig());
-  const tg = config.channels.telegram;
-  if (!tg) {
-    err.write(
-      "telegram is not configured — run `phantombot telegram` first.\n",
-    );
-    return 2;
+
+  // Resolve which Telegram bot account to send through. With --persona,
```
andrewagrahamhodges, line 65: 🔵 **Trivial (Praise)**: The mapping between TOML structure keys (`channels.telegram.personas`) and internal memory attributes (`telegramPersonas`) is bridged cleanly here with a very helpful, actionable error hint containing available options.

tests/cli-notify.test.ts
```diff
@@ -141,6 +141,54 @@ describe("runNotify text", () => {
   });
 });
 
+describe("runNotify persona routing", () => {
```
andrewagrahamhodges, line 144: 🔵 **Trivial (Praise)**: Awesome addition of test specs isolating persona bot and allowlist token resolutions separately. Both happy path routing and unknown persona validation are perfectly scoped.

andrewagrahamhodges (approved):
## Code Review: Bot Persona CLI Routing for Notification

This PR adds excellent persona routing capabilities to the `notify` CLI command via the new `--persona` flag. By targeting the correct bot configured under `channels.telegram.personas.<name>`, it resolves the discrepancy where automated/scheduled notifications had no inbound contextual mapping to keep messages correctly bound.

### Findings

| 1 | 🔵 Trivial | src/cli/notify.ts | 65 | Great design mapping raw in-memory keys back to actual TOML configuration hierarchy to aid user UX on errors. |
| 2 | 🔵 Trivial | tests/cli-notify.test.ts | 144 | Clean, isolated persona routing validation cases covering valid routing and unknown persona fail paths beautifully. |

### What's Good
- **Clear Error Messaging**: Great work providing the `hint` containing configured personas on failure block.
- **Robust Regression Safety**: By keeping omitted flags falling back on the primary config path, absolutely no breaking modifications or manual schema migrations are introduced for single-bot setups!

<details>

```
Everything looks fantastic! Fully verified, approved, and ready to merge.
```
</details>

PR #148, opened 2026-06-03
feat(harness): expose persona as PHANTOMBOT_PERSONA env to subprocesses

14skywalker:
## What

Passes the per-turn persona to every harness subprocess via `PHANTOMBOT_PERSONA`, so tools (like bot-inbox) can derive their bot identity without a separate config var.

## Why

bot-inbox (phantomtools) needs a `--from`. Rather than introducing a new `$BOT_NAME`, we lift the persona phantombot already knows — at the right layer, and automatically correct if the host ever runs multiple personas.

## Changes

- Added `persona?` to `HarnessRequest`, set from `turn.ts` (the chokepoint that tick/ask/nightly all flow through)
- New `withPersonaEnv()` helper in `envBootstrap.ts` (copy-on-write, doesn't touch `process.env`) — all 4 harnesses use it so the var name can't drift
- All 4 harnesses (claude/codex/gemini/pi) set `PHANTOMBOT_PERSONA`
- Recovery (no-tools path) deliberately left untouched; the field is optional

## Multi-persona

The persona is passed **per-invoke**, not as a global env var — each subprocess gets the identity of that turn.

## Tests

120/120 pass, typecheck clean (+4 unit tests).

### DIFF
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -48,7 +48,7 @@
 
 import { access, constants } from "node:fs/promises";
 import type { Harness, HarnessChunk, HarnessRequest } from "./types.ts";
-import { reloadEnvFiles } from "../lib/envBootstrap.ts";
+import { reloadEnvFiles, withPersonaEnv } from "../lib/envBootstrap.ts";
 import {
   createKillCoordinator,
   type HarnessActivity,
@@ -99,7 +99,7 @@ export class ClaudeHarness implements Harness {
 
     // OAuth-on-host: don't leak ANTHROPIC_API_KEY into the subprocess env,
     // so claude resolves credentials from ~/.claude/.credentials.json.
-    const env = filterAuthEnv(process.env);
+    const env = withPersonaEnv(filterAuthEnv(process.env), req.persona);
 
     const proc = spawnInNewSession([this.config.bin, ...args], {
       cwd: req.workingDir,
--- a/src/harnesses/types.ts
+++ b/src/harnesses/types.ts
@@ -22,6 +22,15 @@ export interface HarnessRequest {
   userMessage: string;
   /** Prior turns of this conversation, oldest first. May be empty. */
   history: HistoryTurn[];
+  /**
+   * Persona key for THIS turn (e.g. "burt"). Exposed to the subprocess as
+   * the `PHANTOMBOT_PERSONA` env var so tools can self-identify without a
+   * hardcoded name — this is the single source of truth for "which bot am
+   * I". Per-turn, not global: a host running multiple personas gets the
+   * right identity on every spawn. Optional — degraded paths (e.g. the
+   * no-tools recovery reply) may omit it.
+   */
+  persona?: string;
   /** Subprocess working directory. Defaults to the agent dir. */
   workingDir?: string;
   /**
--- a/src/lib/envBootstrap.ts
+++ b/src/lib/envBootstrap.ts
@@ -255,3 +255,21 @@ export async function reloadEnvFiles(
 
   return { updated, removed };
 }
+
+/**
+ * Return a copy of `base` with `PHANTOMBOT_PERSONA` set to the turn's persona
+ * key, so the spawned harness subprocess can self-identify (e.g. bot-inbox
+ * defaults its `--from` to it). Per-spawn and copy-on-write: we never mutate
+ * the caller's env (notably the global `process.env`). When `persona` is
+ * empty/undefined the base is returned untouched — degraded paths that don't
+ * carry a persona just don't get the var.
+ *
+ * One helper, shared by all four harnesses, so the var name can't drift.
+ */
+export function withPersonaEnv<T extends NodeJS.ProcessEnv>(
+  base: T,
+  persona: string | undefined,
+): T {
+  if (!persona) return base;
+  return { ...base, PHANTOMBOT_PERSONA: persona };
+}
--- a/src/orchestrator/turn.ts
+++ b/src/orchestrator/turn.ts
@@ -169,6 +169,7 @@ export async function* runTurn(input: TurnInput): AsyncGenerator<HarnessChunk> {
     systemPrompt,
     userMessage: input.userMessage,
     history,
+    persona: input.persona,
     workingDir: input.workingDir ?? homedir(),
     idleTimeoutMs: input.idleTimeoutMs,
     hardTimeoutMs: input.hardTimeoutMs,
--- a/src/harnesses/codex.ts
+++ b/src/harnesses/codex.ts
@@ -10,7 +10,7 @@
 
 import { access, constants } from "node:fs/promises";
 import type { Harness, HarnessChunk, HarnessRequest } from "./types.ts";
-import { reloadEnvFiles } from "../lib/envBootstrap.ts";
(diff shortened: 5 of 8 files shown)


robertclawson (commented):
Reviewed for correctness and breaking impact. **Verdict: safe to merge — no negative impacts found.**

This is a tight, additive change. The new `PHANTOMBOT_PERSONA` var is set copy-on-write per spawn (never mutates the global `process.env`), all four harnesses go through one shared helper so the var name can't drift, and the no-tools recovery path is deliberately left alone. I verified `TurnInput.persona` is a required, already-populated `string`, and that `turn.ts` correctly forwards the persona *key* (`input.persona`) rather than the loaded persona object — so this actually does something rather than being a silent no-op.

Risk surface checked:
- **process.env mutation:** none — `withPersonaEnv` spreads into a fresh object; tests assert this.
- **Behaviour change for existing harnesses:** previously codex/gemini/pi passed the live `process.env` reference; they now pass a shallow copy when a persona is present. `spawn` snapshots env at spawn time anyway, so this is equivalent — no regression.
- **Secret leakage:** the var is just a persona key (e.g. `robbie`), not sensitive; claude.ts still strips auth via `filterAuthEnv` first.
- **Tests/typecheck:** +4 unit tests cover set / no-mutate / undefined / empty; author reports 120/120 + clean typecheck.

A couple of non-blocking observations inline.

src/lib/envBootstrap.ts
```diff
@@ -255,3 +255,21 @@ export async function reloadEnvFiles(
...
+ * carry a persona just don't get the var.
+ *
+ * One helper, shared by all four harnesses, so the var name can't drift.
+ */
+export function withPersonaEnv<T extends NodeJS.ProcessEnv>(
+  base: T,
+  persona: string | undefined,
+): T {
+  if (!persona) return base;
+  return { ...base, PHANTOMBOT_PERSONA: persona };
```
robertclawson, line 274: Core of the change — looks correct. Copy-on-write (fresh object via spread), early-return on falsy persona keeps the degraded paths untouched, and a single shared helper means the var name can't drift across the four harnesses. 

Minor: the generic `<T extends NodeJS.ProcessEnv>` returning `{ ...base, PHANTOMBOT_PERSONA }` as `T` is technically the usual TS spread-widening unsoundness, but it's harmless here and keeps callers' env type intact. No change needed.

src/harnesses/claude.ts
```diff
@@ -99,7 +99,7 @@ export class ClaudeHarness implements Harness {
 
     // OAuth-on-host: don't leak ANTHROPIC_API_KEY into the subprocess env,
     // so claude resolves credentials from ~/.claude/.credentials.json.
-    const env = filterAuthEnv(process.env);
+    const env = withPersonaEnv(filterAuthEnv(process.env), req.persona);
```
robertclawson, line 102: Ordering is right: `filterAuthEnv` strips `ANTHROPIC_API_KEY` first, then `withPersonaEnv` adds the persona to that already-filtered copy. `filterAuthEnv` returns a new object so there's no aliasing concern, and the persona var is non-sensitive so it doesn't undermine the auth-stripping intent. 👍

src/harnesses/types.ts
```diff
@@ -22,6 +22,15 @@ export interface HarnessRequest {
...
   history: HistoryTurn[];
+  /**
+   * Persona key for THIS turn (e.g. "burt"). Exposed to the subprocess as
+   * the `PHANTOMBOT_PERSONA` env var so tools can self-identify without a
+   * hardcoded name — this is the single source of truth for "which bot am
+   * I". Per-turn, not global: a host running multiple personas gets the
+   * right identity on every spawn. Optional — degraded paths (e.g. the
+   * no-tools recovery reply) may omit it.
+   */
+  persona?: string;
```
robertclawson, line 33: Worth noting: `persona?` is optional on `HarnessRequest`, but `TurnInput.persona` (the upstream chokepoint) is a required `string`. So in the normal flow it's always populated; the optionality only matters for degraded/recovery `HarnessRequest`s that omit it. That's the intended design (and the comment explains it well) — flagging only so a future reader doesn't assume it can be missing on the main path.

kaieriksen (approved):
Reviewed the env propagation across runTurn and all four harnesses. I don't see a negative-impact path here: normal turns now pass the per-turn persona into PHANTOMBOT_PERSONA without mutating process.env, Claude still filters ANTHROPIC_API_KEY before the persona overlay, and the recovery path remains tool-free.\n\nVerified locally:\n- ~/.bun/bin/bun tsc --noEmit\n- ~/.bun/bin/bun test tests/lib-envBootstrap.test.ts tests/orchestrator-turn.test.ts tests/harnesses-claude.test.ts tests/harnesses-codex.test.ts tests/harnesses-gemini.test.ts tests/harnesses-pi.test.ts\n- ~/.bun/bin/bun test

src/lib/envBootstrap.ts
```diff
@@ -255,3 +255,21 @@ export async function reloadEnvFiles(
...
+}
```
andrewagrahamhodges, line 275: 🔵 **Trivial (Praise)**: Strict object destructuring is an excellent mechanism here to guarantee thread isolation. Since multi-persona installations might concurrent-process across separate fibers, mutating global state would have resulted in data-races. This prevents that elegantly.

src/orchestrator/turn.ts
```diff
@@ -169,6 +169,7 @@ export async function* runTurn(input: TurnInput): AsyncGenerator<HarnessChunk> {
     systemPrompt,
     userMessage: input.userMessage,
     history,
+    persona: input.persona,
```
andrewagrahamhodges, line 172: 🔵 **Trivial (Praise)**: Wiring this cleanly at the orchestrator level inside `runTurn()` is the perfect hub location. Setting it here ensures that whether the execution path is triggered from automated tick timers or manual interactive chat requests, the identity propagates correctly.

andrewagrahamhodges (approved):
## Code Review: Persona Subprocess Environment Integration

**Author:** @14skywalker | **Files:** 8 | **Verdict:** Approve

This is an exceptionally elegant, copy-on-write solution to pass the contextual identity of the running bot persona directly to the underlying harnessing layers. It avoids static global state mutation (preserving target thread safety) and perfectly allows multi-agent hosts to correctly tag or identify outgoing events or commands without introducing a secondary `$BOT_NAME` setting configuration.

### Findings

| 1 | 🔵 Trivial | src/lib/envBootstrap.ts | 275 | Solid immutability guarantees through strict object destructuring (`{ ...base, PHANTOMBOT_PERSONA: persona }`). |
| 2 | 🔵 Trivial | tests/lib-envBootstrap.test.ts | 318 | Exhaustive unit coverage validating exact immutability/identity protection for copy-on-write checks. |

### What's Good
- **Immutability First**: Preventing mutation of `process.env` guarantees concurrent agent threads never corrupt or leak identifiers across interleaved turn invocations.
- **Robust Integration**: Safely injecting `PHANTOMBOT_PERSONA` across all four harnesses (`claude`, `codex`, `gemini`, `pi`) ensures a consistent and predictable operational footprint.
- **Degraded/Recovery Safety**: Gracefully maps empty/undefined states to avoid any potential crashing edge cases during degraded/recovery loops.

<details>

```
No changes needed. Outstanding code safety and test coverage!
```
</details>

PR #150, opened 2026-06-04
Persist resolved harness binaries for headless services

kaieriksen:
## What

This PR hardens Phantombot's harness discovery so headless Telegram services do not silently lose their model harness when a CLI binary moves out of the service PATH. Phantombot now resolves configured harnesses to absolute executable paths, records successful discoveries in its managed runtime state, and uses those paths directly on later starts.

The service no longer refuses to start when a harness binary is missing. Instead, `phantombot run` logs a loud warning, keeps the daemon alive, and lets the normal harness turn surface the operational failure until `phantombot doctor` or the harness setup flow repairs the path.

## Why

The Honduras incident showed that Telegram ingress could be healthy while every reply failed because the systemd service could not execute `pi`. Pi had installed under a versioned node directory, and the stable shim that used to make `pi` visible on PATH had disappeared. That failure mode is not unique to Pi; npm and node-based harnesses often install through shims or versioned directories that can drift independently from a long-running service's environment.

A headless bot should stay up even when one dependency is misconfigured. The safer behavior is to preserve service availability, diagnose the missing harness clearly, and remember any absolute path Phantombot has already proven usable.

## How

Harness availability now uses one shared resolver. It first respects explicit absolute configuration, then checks the supplied PATH, then scans common user-level npm, nvm, fnm, Bun, Volta, Pi, and pi-node install locations. When a harness is found, Phantombot saves the resolved path in `state.json` under `harness_bins`, and `loadConfig` uses that saved path when no explicit env or TOML override is present.

The `phantombot harness` flow records detected paths immediately. `phantombot doctor --repair` also persists resolved paths, while read-only doctor still reports missing binaries without changing state. `phantombot run` applies resolved paths before constructing the harness chain, logs missing binaries without exiting, and continues serving Telegram.

The systemd PATH cleanup from the first version of the PR remains, but it is now a convenience layer rather than a correctness dependency.

## Testing

- `~/.bun/bin/bun tsc --noEmit`
- `~/.bun/bin/bun test tests/cli-run.test.ts tests/config.test.ts tests/state.test.ts tests/lib-harnessAvailability.test.ts tests/cli-doctor.test.ts`
- `~/.bun/bin/bun test` — `1112 pass, 1 skip, 0 fail`


lenaparkhodges (dismissed):
I've reviewed the changes and verified the fix for the harness binary path issue.

### **Review Summary**
*   **Path Robustness:** The introduction of `PHANTOMBOT_SERVICE_PATH` in `systemd.ts` is exactly what was needed. It covers the `pi-node` and `npm` installation quirks by providing a stable, expanded PATH for the daemon.
*   **Discovery Logic:** Moving the binary resolution into `harnessAvailability.ts` cleans up the CLI tools and ensures consistency between `run`, `harness`, and `doctor`.
*   **Doctor Integration:** `phantombot doctor` now correctly simulates the service PATH when running on Linux, allowing for accurate diagnosis of binary visibility issues without needing to be inside the service's cgroup.
*   **Fail-Fast Startup:** The new checks in `src/cli/run.ts` ensure the bot doesn't start in a broken state if its harnesses are missing.

### **Technical Verification**
I verified the path expansion and binary resolution logic with a new test suite (`tests/lib-harnessAvailability.test.ts`). All logic for handling `%h` and walking the custom PATH is working correctly.

The PR is technically solid and directly resolves the path reliability problem reported.

**Minor suggestion:** I noticed that `src/lib/harnessAvailability.ts` was missing dedicated unit tests for its exported functions (though they are indirectly touched by `cli-doctor.test.ts`). I've verified them locally, but adding them to the repo would be good for long-term coverage.

src/cli/run.ts
```diff
@@ -180,6 +189,23 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
     return 2;
   }
+  const harnessChecks =
+    input.checkHarnesses === false
+      ? []
+      : input.checkHarnesses
+        ? await input.checkHarnesses(config)
+        : await checkConfiguredHarnesses(config);
+  const missingHarnessBins = missingHarnesses(harnessChecks);
+  if (missingHarnessBins.length > 0) {
```
robertclawson, line 203: **Main finding (design, worth a decision before merge): this blocks startup if *any* configured harness is missing, including a fallback.**

`missingHarnesses()` flags every unresolved entry in `config.harnesses.chain`, and this returns 2 if the list is non-empty. With chain `[claude, pi]`, if only the fallback `pi` is missing but `claude` resolves fine, the service still refuses to boot. Combined with the unit's `Restart=on-failure` + `SuccessExitStatus=143` (exit 2 ≠ 143 → failure), that's a **5-second crash-loop on a bot that could run perfectly on its primary harness** — which inverts the whole point of the fallback chain (graceful degradation).

Suggest: hard-block only when the **primary** (chain[0]) — or *all* — harnesses are missing, and let a missing *fallback* be a doctor WARN rather than a boot blocker. As written, configuring a fallback you haven't installed yet takes the bot offline.

src/lib/systemd.ts
```diff
@@ -36,6 +36,13 @@ const ENVIRONMENT_FILE_LINES =
   "EnvironmentFile=-%h/.config/phantombot/.env\n" +
   "EnvironmentFile=-%h/.env";
 
+export const PHANTOMBOT_SERVICE_PATH =
+  "%h/.local/share/pi-node/bin:" +
```
robertclawson, line 40: These `pi-node/{bin,current/bin}` entries are a hardcoded guess at where the self-managed pi runtime lands. They don't exist on the Hetzner box (harnesses there are apt/npm-global at `/usr/bin`, already on the tail), so this is harmless here — but please confirm the path is actually correct on a box that uses pi's self-managed node runtime (e.g. kw-openclaw), otherwise it's dead weight that gives false confidence. The durable fix for any *other* versioned location remains `PHANTOMBOT_<HARNESS>_BIN` — which is good; this PATH list can only ever chase known dirs.

robertclawson (commented):
Reviewed against the actual failure mode (harness binaries installed where the daemon PATH can't see them). The detection + doctor reporting is the genuinely valuable, universal part — it turns a silent mid-turn failure into a loud, actionable one, and the `PHANTOMBOT_<HARNESS>_BIN` absolute-path escape hatch is the real catch-all fix. Three notes below; the run.ts one is worth a decision before merge. Verified on a live box: the new `pi-node/{bin,current/bin}` dirs don't exist here and the harnesses resolve from `/usr/bin` (already covered by the PATH tail), so the widening is additive/harmless — but see the systemd note on confirming it for the boxes that DO use the self-managed pi runtime.

src/lib/harnessAvailability.ts
```diff
@@ -0,0 +1,79 @@
...
+      return bin;
+    } catch {
+      return undefined;
+    }
+  }
+  for (const dir of pathEnv.split(":")) {
+    if (!dir) continue;
+    const candidate = join(dir, bin);
+    try {
+      await access(candidate, constants.X_OK);
```
robertclawson: Nit (low, pre-existing — moved from harness.ts): `access(candidate, X_OK)` also succeeds for an *executable directory*, so a dir named e.g. `pi` on PATH would be falsely "resolved". Practically a non-issue, but a `stat().isFile()` guard would make `whichBinary` strictly correct now that it's the shared resolver feeding the new fail-fast.

src/lib/harnessAvailability.ts
```diff
@@ -0,0 +1,196 @@
...
+}
+
+export async function resolveHarnessBinary(
+  bin: string,
+  pathEnv = process.env.PATH ?? "",
+): Promise<ResolvedHarnessBinary> {
+  if (bin.startsWith("/")) {
+    return (await executable(bin))
+      ? { path: bin, source: "configured" }
+      : {};
```
robertclawson, line 128: **Non-blocking, but worth a fast-follow — it's in this PR's exact problem domain.**

Once a binary's absolute path is persisted to `state.json`, `loadConfig()` feeds that absolute path back in as `bin`. On the next `run`, `resolveHarnessBinary` takes the `bin.startsWith("/")` branch: if the file is gone (e.g. an `nvm`/`fnm` node upgrade moved `claude` to a new versioned dir), it returns `{}` here and **never falls through to `harnessSearchPath()` to re-discover it by basename**. Result: a stale persisted path is reported "missing" forever — and it can't self-heal even on restart, because the stale absolute path keeps short-circuiting the search. Recovery needs a manual `state.harness_bins` edit (or an env/TOML override).

That's precisely the "versioned npm/node paths move" scenario this PR targets — it's fixed for the *first* resolution but regresses on the *next* move.

Cheap fix: if an absolute `bin` isn't executable, fall through to a basename search instead of giving up:
```ts
if (bin.startsWith("/")) {
  if (await executable(bin)) return { path: bin, source: "configured" };
  // stale absolute path — re-discover by basename rather than giving up
  bin = basename(bin);
}
```

robertclawson (dismissed):
**APPROVE** — this revision resolves the core problem and clears the blocker I raised on the previous pass. Reviewed carefully against the actual root cause (npm/node installs putting harness binaries where subprocess shells can't reliably find them on PATH).

**Does it fix the nasty problem? Yes, properly.** The real fix is the resolve-and-persist chain:
- `run` resolves each configured harness to an **absolute path** (searches PATH + nvm/fnm/volta/bun/npm-global/`.local/share` node dirs), persists it to `state.json`, and applies it to the in-process config.
- `loadConfig()` now reads `state.harness_bins.<id>` as a fallback in the precedence chain (env > TOML > **persisted** > bare name). So every subprocess — `ask`, tick, heartbeat, nightly — picks up the absolute binary even when launched from a shell whose PATH lacks the npm dir. That's the actual root-cause fix; PATH widening alone never could be.

**My prior blocker is fixed.** Missing harnesses no longer abort startup (no more crash-loop on a missing *fallback*) — they log + warn "Phantombot will keep running…", surfaced loudly via `doctor`. Threads resolved.

**Verified on this box:** `config.toml` has no explicit `bin =`, so the persisted fallback correctly takes effect; the `explicit-TOML-overrides-persisted` precedence is covered by a test.

**Test coverage is strong and on-target** — warn-but-keep-running, persist-before-listeners, persisted-vs-explicit precedence, doctor exit codes, `saveHarnessBins` merge, systemd PATH. CI all green.

**One residual finding (non-blocking, inline)** — a stale persisted absolute path won't self-heal because `resolveHarnessBinary` short-circuits the search on absolute bins. Directly in this PR's domain, so worth a fast-follow; doesn't block since it's strictly better than today and only triggers after a binary *moves* post-resolution.

Low nit still open from last pass: `access(_, X_OK)` matches an executable *directory* too — harmless in practice.

Approving — happy to see the stale-path fallthrough as a quick follow-up.

robertclawson (dismissed):
Re-reviewed after commit `dbc2e1df` ("rediscover stale cached binaries"). **This closes my one residual finding from the last pass — approving.**

`checkConfiguredHarnesses` now detects when a *persisted absolute* harness path no longer resolves (e.g. a node/npm upgrade moved the binary) and falls back to re-discovering it by canonical name via `defaultHarnessBin`. I traced it end-to-end in `run.ts`: rediscovery → `resolvedHarnessBins` → `saveHarnessBins` re-persists the corrected absolute path to state.json → `applyResolvedHarnessBins` updates the live config, so every subprocess (ask/tick/heartbeat/nightly) picks up the fresh path on the next service start. That is the self-heal the previous revision was missing. Good targeted test for the stale-path branch too. CI all green.

One tiny, non-blocking note (not worth holding the PR): the fallback re-searches the *canonical* basename (claude/pi/gemini/codex), so an explicitly-set custom absolute `bin` that goes stale would re-resolve the canonical binary rather than the custom one. Pure edge case for deliberate overrides; fine as-is.

My earlier low `X_OK`-on-a-directory nit still stands but is pre-existing and non-blocking. Approving.

andrewagrahamhodges (approved):

PR #153, opened 2026-06-05
fix(ci): release on push to main (fork-safe token)

robertclawson:
## Problem

The `v1.1.100` release (PR #151 — our first **external fork** contribution from Salvatore) built green but failed at the final `Create GitHub Release` step:

```
HTTP 403: Resource not accessible by integration
```

**Root cause:** the release workflow triggered on `pull_request: types: [closed]`. GitHub deliberately hands **fork-originated `pull_request` runs a read-only `GITHUB_TOKEN`** — it ignores the workflow's `permissions: contents: write` by design, so a fork can't gain write access to the base repo. Every prior release came from an *internal* branch (full-write token), so this never bit us until the first fork PR merged.

## Fix

Switch the trigger to **`push: branches: [main]`**. A push to a branch in this repo always runs with the base repo's full read-write token, regardless of where the merged code originated — because the merge commit already lives on `main`. Under branch protection every push to `main` *is* a PR merge, so behaviour is otherwise equivalent.

**Deliberately NOT `pull_request_target`** (the other fork-write option): it would run unreviewed fork code with secrets + write scope *before* merge. `push` fires only *after* review + merge, so the code is already trusted.

## What changed

Push events carry no `github.event.pull_request` context, so:

- **checkout** and `gh release create --target` now use `github.sha` (the merge commit on `main`).
- A new **"Resolve originating PR"** step recovers the PR number/title from the commits→pulls API, preserving the audit trail in the release title + notes. Best-effort: a direct push with no PR falls back to the commit subject and a plain-tag title.
- Release notes / title render gracefully whether or not a PR was resolved.

## Safety notes

- **`run_number` (the version source) is unaffected** — it's keyed to the workflow file path, which is unchanged, so versions keep climbing monotonically (no collision risk). The workflow `name:` is intentionally left as-is for the same reason.
- No change to the build/test/checksum steps — only the trigger and the PR-context plumbing.

## Validation

This can only be fully exercised by the next fork-PR merge, but: YAML validated, every removed `github.event.pull_request.*` reference is replaced, and the release still cuts on internal-branch merges exactly as before.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,7 +62,8 @@ on:
     branches: [main]
 
 permissions:
-  contents: write   # create release + upload assets
+  contents: write        # create release + upload assets
+  pull-requests: read    # commits->pulls lookup for release title/notes audit trail
 
 jobs:
   release:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,28 @@
 name: Release on PR merge
 
-# Fires for every PR merged into main. Cuts a GitHub Release tagged
+# Fires on every push to main (i.e. every PR merge — branch protection
+# requires PRs, so a push to main IS a merge). Cuts a GitHub Release tagged
 # v1.1.<RUN_NUMBER> with cross-compiled binaries (linux x64-baseline +
 # arm64, darwin x64 + arm64) and a SHA256SUMS file. `phantombot update`
 # reads this feed.
+#
+# Why `push: main` and NOT `pull_request: closed`: a PR merged from a
+# FORK runs the closed-PR workflow with a *read-only* GITHUB_TOKEN —
+# GitHub strips write scope from fork-triggered pull_request runs by
+# design, so `gh release create` 403s ("Resource not accessible by
+# integration"). This bit us on v1.1.100 (PR #151, the first external
+# fork contribution). A `push` to a branch in THIS repo always runs with
+# the base repo's full read-write token regardless of where the merged
+# code originated, because the merge commit already lives on main. We do
+# NOT use `pull_request_target` (the other fork-write option) on purpose:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -39,23 +58,43 @@ name: Release on PR merge
 # AVX2 — see the SIGILL post-mortem from the May 2026 first deploy.
 
 on:
-  pull_request:
-    types: [closed]
+  push:
     branches: [main]
 
 permissions:
   contents: write   # create release + upload assets
```
kaieriksen: The new commits-to-pulls lookup needs `pull-requests: read` in this permissions block. Once a workflow declares explicit permissions, unspecified scopes are not available on `GITHUB_TOKEN`, and GitHub documents this endpoint as requiring Pull requests read permission for private resources. As written, `gh api .../commits/${{ github.sha }}/pulls` can 403 in the real release run, get swallowed by `|| echo []`, and silently drop the PR number/title from the release title and notes. Add `pull-requests: read` here so the audit trail this PR is preserving actually works.

kaieriksen (changes requested):
Requested one change: add `pull-requests: read` to the release workflow permissions so the new originating-PR lookup works under the workflow token instead of silently falling back to no PR audit trail.

.github/workflows/release.yml
robertclawson: Good catch, fixed in db62084. Added `pull-requests: read` to the permissions block — you're right that with an explicit permissions block every unspecified scope is denied, so the `commits/SHA/pulls` lookup would have 403'd, been swallowed by `|| echo []`, and silently dropped the PR title/notes that this very PR exists to preserve. Thanks Kai.

kaieriksen (approved):
Verified the permission fix. The release workflow now grants pull-requests: read for the commits-to-pulls audit lookup, CI is green, and local typecheck passes.

Issue #161, opened 2026-06-10
🟡 Security hardening — briefing-tag strip, log redaction, cron tz, runLock nonce, migration error handling, telegram framing


robertclawson:
## 🟡 Security & correctness hardening (PR C)

Grouped medium-severity items. Line numbers from reviewer's checkout — confirm during implementation.

**A. `<briefing>` tags not stripped from untrusted content** — `src/orchestrator/screen.ts:170-178`, `src/lib/threatJudge.ts:166`
Marker-stripping removes `<untrusted_content>` but **not** `<briefing>`, which the judge is told to trust. A concrete injection vector to lower the judge's own score. (The fail-open-on-unparseable-JSON behaviour is documented/intentional and left as-is; the briefing-tag gap looks unintended.) Fix: strip `<briefing>` from untrusted content before it reaches the judge.

**B. Log redaction is prefix-only; task output stored unredacted** — `src/cli/tick.ts:410-417`, `tasks.ts:466`
Redaction catches only `ghp`/`sk`/`xox*`-shaped tokens; bare bearer / AWS (`AKIA…`) tokens stream verbatim. `logRun` writes `outputExcerpt` (first 500 chars of raw command output) into `task_runs` with **no** redaction. Fix: broaden the redaction patterns and apply them to `outputExcerpt`.

**C. Cron hard-coded to UTC** — `src/lib/cronSchedule.ts:24,37,58`
`0 9 * * *` fires 9am **UTC**, drifting across the operator's DST and wrong wall-clock for any non-UTC user (we're Europe/Amsterdam). This is phantombot's own `task` scheduler. (Missed-run "skip don't pile up" semantics are correct.) Fix: make the cron tz-aware (operator/persona timezone).

**D. `runLock` PID-reuse / TOCTOU** — `src/lib/runLock.ts:65-92`
Stale detection is `kill(pid,0)`; a recycled PID → permanent false "already running", plus a check→create race the code itself flags. Fix: add a start-time/boot-id nonce to disambiguate.

**E. Migrations swallow all exceptions** — `src/lib/tasks.ts:238-248`
`ALTER TABLE ADD COLUMN` catches everything, so a real failure (disk full, locked DB) is indistinguishable from the idempotent "duplicate column" no-op → silent schema drift. Fix: catch only the duplicate-column case, rethrow otherwise.

**F. Untrusted Telegram text framed as trusted structure** — `telegram.ts:627-665,1296-1298`
[...]

Issue #159, opened 2026-06-10
🔴 Telegram offset committed before processing — crash/restart drops messages


robertclawson:
## 🔴 High (reliability) — Telegram offset committed before messages are processed; crash drops messages

**File:** `src/channels/telegram.ts` (~L1124-1126)

`offset = nextOffset` advances **before** the `for (const msg of updates)` loop enqueues fire-and-forget per-chat turns. The next `getUpdates` commits that offset to Telegram, so a SIGTERM/restart/crash with turns still in-flight loses those messages — Telegram considers them acked, no reply ever produced. The `ackUpdates` machinery only protects the slash-command `afterSend` path; ordinary turns have no such guarantee. At-most-once where at-least-once is wanted.

**Verified on `main`:** `ackUpdates`/`getUpdates`/offset machinery present as described.

**Fix:** only advance/commit the offset once the corresponding turns are durably enqueued (or extend the existing ack mechanism to ordinary turns).

_Severity: high. Slated for PR B (data-integrity cluster)._

robertclawson:
**Closing as by-design — this is an intentional, calculated trade-off, not a bug.**

The poll loop advances the Telegram offset *before* per-chat turns are enqueued fire-and-forget, which makes delivery **at-most-once**: a crash/SIGTERM with turns still in flight will drop those messages.

We are keeping this behaviour deliberately. A phantom is a **conversational agent, not a durable task queue**. Users are mid-conversation with their agent — they are not submitting jobs to a waterfall worklist where each item must run exactly once. What matters here is **concurrency and responsiveness**: advancing the offset immediately lets us poll the next batch and fan messages out to independent per-chat turns without serializing the whole bot behind the slowest in-flight turn.

Getting true at-least-once (a persistent inbox, or an ack-watermark + dedup layer) would buy crash-replay of the rare dropped message at the cost of the snappy, concurrent, fire-and-forget experience that defines the product. We choose the experience. In the worst case a user re-sends and the conversation continues — an acceptable loss.

The behaviour is now documented with a strong in-code comment at the offset-ack site in `src/channels/telegram.ts` (PR #164) so it is not mistaken for a bug and silently 'fixed' later. Anyone wanting to change it must re-litigate this trade-off first.

Closing as `wontfix` / by-design.

PR #163, opened 2026-06-10
fix: auth-family env leak (#157) + sqlite busy_timeout (#158)

### LINKED ISSUE #157: 🔴 OAuth-only invariant incomplete — auth token can silently leak into claude subprocess
## 🔴 High (security/billing) — OAuth-only invariant is incomplete; token auth can silently leak in

**File:** `src/harnesses/claude.ts` (~L97, L224-236)

`filterAuthEnv` strips only `ANTHROPIC_API_KEY`. It does **not** strip `ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, or `ANTHROPIC_BASE_URL`. `reloadEnvFiles()` runs immediately before (L97), re-sourcing `~/.env` into `process.env`. So a user who runs `phantombot env set ANTHROPIC_AUTH_TOKEN …` (or has it exported) silently flips `claude` from the Max-subscription OAuth path to token billing — the exact failure the design comment claims to prevent.

**Verified on `main`:** `filterAuthEnv` body is a single `if (k === "ANTHROPIC_API_KEY") continue;`.

**Fix:** switch from denylist to **allowlist**, or at minimum strip the whole `ANTHROPIC_*` auth family (`ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, `ANTHROPIC_BASE_URL`). Add a regression test asserting all four are absent from the subprocess env.

_Severity: high. Cheap, high-value — slated for PR A._

### LINKED ISSUE #158: 🔴 SQLite missing busy_timeout — concurrent run/tick writers throw SQLITE_BUSY
## 🔴 High (data-loss) — SQLite has no `busy_timeout` but two processes write the shared DB

**Files:** `src/memory/store.ts:351`, `src/lib/tasks.ts:569`, `src/lib/memoryIndex.ts:158`

WAL is on, but `busy_timeout` is unset (=0). The store comment even says "single-process" (store.ts:349) — yet `phantombot tick` is a **separate process** sharing `memory.sqlite` with `phantombot run`. WAL allows concurrent readers + one writer; two concurrent writers (`run` persisting a turn while `tick` calls `recordRun`) get an immediate `SQLITE_BUSY` throw instead of a blocking retry → intermittent, hard-to-repro write failures under normal operation.

**Verified on `main`:** all three openers set `PRAGMA journal_mode = WAL` with no `busy_timeout`.

**Fix:** one line per opener — `db.exec("PRAGMA busy_timeout = 5000");`

_Severity: high. One-liner — slated for PR A._

robertclawson:
**PR A** of the codebase-health triage — the two cheap, high-value fixes. Both verified against current `main` before changing. Closes #157, closes #158.

## #157 — auth-family env leak (security/billing) 🔴
`filterAuthEnv` stripped only `ANTHROPIC_API_KEY`. The rest of the auth/routing namespace — `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL`, `ANTHROPIC_CUSTOM_HEADERS`, `CLAUDE_CODE_OAUTH_TOKEN`, `CLAUDE_CODE_USE_BEDROCK`, … — passed straight through. Since `invoke()` runs `reloadEnvFiles()` immediately before filtering, a stray `phantombot env set ANTHROPIC_AUTH_TOKEN …` (or any exported family member) would silently flip claude off the Max-subscription OAuth path onto token/proxy billing — the exact failure the OAuth-only invariant exists to prevent.

**Fix:** swap the single-name denylist for a **namespace allowlist** — drop every `ANTHROPIC_*` / `CLAUDE_CODE_*` var unless explicitly allow-listed (`AUTH_ENV_ALLOW`, empty today since the codebase reads none of them). Closes the whole family at once and is future-proof against new auth vars.

## #158 — sqlite `busy_timeout` missing (data-loss) 🔴
`memory.sqlite` is shared across processes (`run` persists turns while `tick` records task runs / reindexes). WAL is on but `busy_timeout` was 0, so a second concurrent writer gets an immediate `SQLITE_BUSY` throw instead of blocking. Intermittent, hard-to-repro write failures.

**Fix:** `PRAGMA busy_timeout = 5000` on all three openers (memory store, task store, memory index).

## Validation
- `bun tsc --noEmit` ✅ clean
- `bun test` ✅ **1168 pass, 1 skip, 0 fail** (incl. new full-auth-family test)
- Two commits, one per issue.

Deliberately scoped tiny — the data-integrity cluster (#159/#160) and security hardening (#161) follow as separate PRs.

### DIFF
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -40,10 +40,11 @@
  *      subscription path. Don't add it back unless that changes upstream.
  *
  * Auth model under phantombot:
- *   ANTHROPIC_API_KEY is filtered out of the subprocess env so claude
- *   resolves credentials from ~/.claude/.credentials.json (the OAuth
+ *   The entire ANTHROPIC_* / CLAUDE_CODE_* auth+routing namespace is
+ *   filtered out of the subprocess env (see filterAuthEnv) so claude
+ *   resolves credentials only from ~/.claude/.credentials.json (the OAuth
  *   path that backs Claude Max). Phantombot does not hold or pass any
- *   API keys.
+ *   API keys, auth tokens, or base-URL overrides.
  */
 
 import { access, constants } from "node:fs/promises";
@@ -96,8 +97,9 @@ export class ClaudeHarness implements Harness {
     // Shell-exported keys remain sticky — see envBootstrap.ts header.
     await reloadEnvFiles();
 
-    // OAuth-on-host: don't leak ANTHROPIC_API_KEY into the subprocess env,
-    // so claude resolves credentials from ~/.claude/.credentials.json.
+    // OAuth-on-host: don't leak any ANTHROPIC_* / CLAUDE_CODE_* auth or
+    // routing var into the subprocess env (reloadEnvFiles just re-sourced
+    // ~/.env), so claude resolves credentials from ~/.claude/.credentials.json.
     const env = withPersonaEnv(filterAuthEnv(process.env), req.persona);
 
     const proc = spawnInNewSession([this.config.bin, ...args], {
@@ -221,16 +223,43 @@ export const PHANTOMBOT_INJECTED_CLAUDE_SETTINGS = {
 } as const;
 
 /**
- * Strip ANTHROPIC_API_KEY from the inherited env so the subprocess uses
- * OAuth credentials at ~/.claude/.credentials.json. Exported for testing.
+ * Prefixes whose entire namespace is treated as authentication/routing
+ * config for the claude subprocess. Anything matching one of these is
+ * dropped from the inherited env UNLESS it is explicitly allow-listed in
+ * AUTH_ENV_ALLOW below.
+ *
+ * Denylisting individual names (the old behaviour, which only stripped
+ * ANTHROPIC_API_KEY) is fragile: ANTHROPIC_AUTH_TOKEN, ANTHROPIC_BASE_URL,
+ * CLAUDE_CODE_OAUTH_TOKEN, CLAUDE_CODE_USE_BEDROCK, etc. all silently flip
+ * claude off the Max-subscription OAuth path. reloadEnvFiles() re-sources
+ * ~/.env into process.env right before this runs, so a stray
+ * `phantombot env set ANTHROPIC_AUTH_TOKEN …` would leak straight through.
+ * Allow-listing the namespace closes the whole family at once.
+ */
+const AUTH_ENV_PREFIXES = ["ANTHROPIC_", "CLAUDE_CODE_"] as const;
+
+/**
+ * Known-safe vars inside the auth namespace that may still pass through to
+ * the subprocess. Empty today — the codebase reads none of these itself,
+ * and the claude subprocess must take its credentials only from
+ * ~/.claude/.credentials.json. Add a name here only with a clear reason.
+ */
+const AUTH_ENV_ALLOW = new Set<string>([]);
+
+/**
+ * Strip the entire ANTHROPIC_* / CLAUDE_CODE_* auth+routing namespace from
+ * the inherited env so the subprocess uses OAuth credentials at
+ * ~/.claude/.credentials.json. Exported for testing.
  */
 export function filterAuthEnv(
   source: NodeJS.ProcessEnv,
 ): Record<string, string> {
   const out: Record<string, string> = {};
   for (const [k, v] of Object.entries(source)) {
-    if (k === "ANTHROPIC_API_KEY") continue;
-    if (v !== undefined) out[k] = v;
+    if (v === undefined) continue;
+    const inAuthNamespace = AUTH_ENV_PREFIXES.some((p) => k.startsWith(p));
+    if (inAuthNamespace && !AUTH_ENV_ALLOW.has(k)) continue;
+    out[k] = v;
   }
   return out;
 }
--- a/src/lib/memoryIndex.ts
+++ b/src/lib/memoryIndex.ts
@@ -156,6 +156,9 @@ export class MemoryIndex {
   constructor(private readonly db: Database) {
     db.exec(SCHEMA);
     db.exec("PRAGMA journal_mode = WAL");
+    // Shared across processes (tick reindex vs. run query): block-and-retry
+    // on a busy writer instead of an immediate SQLITE_BUSY throw. See store.ts.
+    db.exec("PRAGMA busy_timeout = 5000");
   }
 
   static async open(indexPath: string): Promise<MemoryIndex> {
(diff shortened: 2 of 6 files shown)


src/harnesses/claude.ts
```diff
@@ -221,16 +223,43 @@ export const PHANTOMBOT_INJECTED_CLAUDE_SETTINGS = {
...
+ * claude off the Max-subscription OAuth path. reloadEnvFiles() re-sources
+ * ~/.env into process.env right before this runs, so a stray
+ * `phantombot env set ANTHROPIC_AUTH_TOKEN …` would leak straight through.
+ * Allow-listing the namespace closes the whole family at once.
+ */
+const AUTH_ENV_PREFIXES = ["ANTHROPIC_", "CLAUDE_CODE_"] as const;
+
+/**
+ * Known-safe vars inside the auth namespace that may still pass through to
+ * the subprocess. Empty today — the codebase reads none of these itself,
```
andrewagrahamhodges, line 243: 🔵 **Trivial (Praise)**: The prefix-matching + allowlist pattern is the right abstraction for an auth namespace. It's easy to audit (one `Set`), impossible to miss a new var (any `ANTHROPIC_*` or `CLAUDE_CODE_*` is blocked by default), and adding an exception is a one-line change with a clear comment requirement.

src/memory/store.ts
```diff
@@ -346,9 +346,13 @@ export async function openMemoryStore(path: string): Promise<MemoryStore> {
     await mkdir(dirname(path), { recursive: true });
   }
   const db = new Database(path, { create: true });
-  // WAL keeps reads non-blocking even though phantombot is single-process —
-  // useful if `phantombot history` is run while a `phantombot chat` REPL is open.
+  // WAL keeps reads non-blocking, but the file is in fact shared across
```
andrewagrahamhodges, line 349: 🔵 **Trivial (Praise)**: The expanded comment explaining *why* busy_timeout is needed (cross-process WAL sharing between `run` and `tick`) is exactly the kind of context that prevents the next maintainer from 'optimizing' it away.

andrewagrahamhodges (dismissed):
## Code Review: Auth-family env leak + SQLite busy_timeout

**Author:** @robertclawson | **Files:** 5 | **Verdict:** Approve

- `filterAuthEnv` is called **only** in `src/harnesses/claude.ts:103`, where it strips auth vars from the claude subprocess's environment before spawn. It does not touch Telegram config, streaming logic, or any other runtime path.
- The `busy_timeout` pragma is a SQLite WAL concurrency improvement — it makes concurrent writers block-and-retry instead of throwing `SQLITE_BUSY`. No behavioral change to any runtime path; purely resilience.
- Zero files in `src/channels/`, `src/orchestrator/turn.ts`, or `src/lib/harnessRunner.ts` are modified.

### Fix 1: Auth-family namespace allowlist (#157) — SECURITY

The old `filterAuthEnv` stripped only `ANTHROPIC_API_KEY`. The new implementation strips the **entire `ANTHROPIC_*` / `CLAUDE_CODE_*` namespace** via prefix matching, with an explicit allowlist (`AUTH_ENV_ALLOW`, currently empty). This closes the whole family at once — `ANTHROPIC_AUTH_TOKEN`, `ANTHROPIC_BASE_URL`, `CLAUDE_CODE_OAUTH_TOKEN`, etc. can no longer leak through and flip claude off the Max-subscription OAuth path.

The allowlist design is future-proof: adding a new safe var is a one-line `AUTH_ENV_ALLOW.add()` instead of hunting for new denylist entries.

### Fix 2: SQLite busy_timeout (#158) — DATA INTEGRITY

`PRAGMA busy_timeout = 5000` added to all three SQLite openers (memory store, task store, memory index). This prevents intermittent `SQLITE_BUSY` throws when `run` persists turns while `tick` records task runs against the same WAL database.

### Findings

| 1 | 🔵 Trivial | src/harnesses/claude.ts | 243 | Clean prefix-matching with explicit allowlist — both readable and future-proof. |
| 2 | 🔵 Trivial | src/memory/store.ts | 349 | Correctly documented why busy_timeout is needed (cross-process WAL sharing). |

### What's Good
- **Minimal, targeted fixes** — each commit addresses one issue with the smallest possible change.
- **Test coverage** — the new `"strips the whole ANTHROPIC_* / CLAUDE_CODE_* auth family"` test explicitly verifies the namespace boundary.
- **Consistent pragma placement** — all three SQLite openers get the same `busy_timeout` treatment, preventing any future opener from being missed.

<details>

```
Nothing to change. Clean, targeted security and data-integrity fixes with good test coverage.
```
</details>

src/lib/memoryIndex.ts
```diff
@@ -156,6 +156,9 @@ export class MemoryIndex {
   constructor(private readonly db: Database) {
     db.exec(SCHEMA);
     db.exec("PRAGMA journal_mode = WAL");
+    // Shared across processes (tick reindex vs. run query): block-and-retry
+    // on a busy writer instead of an immediate SQLITE_BUSY throw. See store.ts.
+    db.exec("PRAGMA busy_timeout = 5000");
```
kaieriksen: This timeout is set too late for the index connection. `MemoryIndex` runs `db.exec(SCHEMA)` before installing the busy handler, so the first schema/setup statements can still fail immediately when tick and another process touch the same index DB. Move `PRAGMA busy_timeout = 5000` into `MemoryIndex.open()` immediately after `new Database(...)` and before constructing `MemoryIndex` / running schema setup; ideally add a small opener test so this does not drift back.

kaieriksen (changes requested):
Requested one fix before this is ready: the auth-family filter is good, and the focused tests/typecheck passed locally, but the MemoryIndex busy_timeout is installed after schema setup, leaving the first statements on that connection unprotected.

robertclawson:
Fixed in `b0f064c`. Moved the WAL + `busy_timeout` pragmas out of the `MemoryIndex` constructor and into `open()`, so they run on the connection **before** `db.exec(SCHEMA)` — schema setup is now protected too. Matches the `store.ts` ordering. Added an opener test asserting `busy_timeout` is in effect on a file-backed index. tsc clean, full suite 1169 pass. @kaieriksen re-review please.

kaieriksen (approved):
Re-reviewed latest commit b0f064c. The MemoryIndex busy_timeout issue is fixed: WAL and busy_timeout now run in open() before schema setup, matching the store opener ordering.\n\nVerified locally:\n- /home/kai/.bun/bin/bun test tests/harnesses-claude.test.ts tests/lib-memoryIndex.test.ts\n- /home/kai/.bun/bin/bun tsc --noEmit

Issue #160, opened 2026-06-10
🟠 Data-integrity cluster — atomic turn persistence, task counter race, fetch timeouts, lost partial stream


robertclawson:
## 🟠 Data-integrity cluster — non-atomic persistence, stale read-modify-write, missing fetch timeouts, lost partial stream

Grouped because they share a theme and a PR (PR B). Line numbers are from the reviewer's checkout — confirm against `main` during implementation.

**4. Non-atomic turn persistence** — `src/orchestrator/turn.ts:255-266`
Two independent `appendTurn` INSERTs; a crash between them leaves a user turn with no assistant reply — the exact "half-turn" the file header claims to prevent. **Verified on `main`** (two separate `appendTurn` calls). Fix: wrap both in `db.transaction`.

**5. `recordRun`/`recordReview` read-modify-write outside a transaction** — `src/lib/tasks.ts:364-403`
`get()` then a separate UPDATE from the stale snapshot; the recurring branch does `run_count = run_count + 1` (SQL-side) while siblings bind a precomputed value. `maxRuns` can miscount. Fix: wrap in a transaction or collapse to a single UPDATE.

**6. No fetch timeouts in the Telegram transport or `audio.ts`** — `telegram.ts:256,324,398,414`, `audio.ts` TTS/STT
Only `getUpdates` is cancelable. A wedged `downloadFile` (telegram.ts ~L410) has no timeout and stalls that chat's serial chain indefinitely (the #135-class wedge, still open for attachments). **Verified on `main`:** `downloadFile` takes no `signal`. Fix: add `AbortController` timeouts to all transport fetches.

**7. Partial streamed text discarded on mid-stream harness failure** — `src/orchestrator/fallback.ts:149-193`
If a harness streams several text chunks then errors on a non-last link, the fallback re-runs from scratch and the user can see a truncated partial reply followed by a full fresh reply from a different model. Fix: de-dup / continuation, or suppress partials until a link completes.

_Severity: high/medium. Slated for PR B._

robertclawson:
**Partial: 2 of 4 items shipped in #164** (merged `a66f5d0`).

✅ Done in PR B1 (#164):
- Atomic turn persistence — `appendTurnPair()` writes user+assistant in one immediate transaction (no half-turns on crash).
- `recordRun` counter race — read+write now in one immediate transaction, SQL-side increment across all branches.

⏳ Remaining → PR B2 (transport resilience):
[...]

PR #164, opened 2026-06-10
fix: transactional integrity — atomic turn pair + recordRun counter (#160)

robertclawson:
## PR B1 — transactional integrity

Two mechanical, low-risk data-integrity fixes from the codebase-health review. Both close a read-modify-write / multi-write window that a crash or a concurrent `tick` process could land in.

### #160 (a) — atomic user+assistant turn persistence
`orchestrator/turn.ts` and the telegram interrupted-pair persisted a turn via **two independent `appendTurn` INSERTs**. A crash between them leaves a half-turn — a user message with no assistant reply — the exact state the file header claims to prevent.

- New `MemoryStore.appendTurnPair(user, assistant)` writes both rows in a single **`.immediate`** transaction (write lock at BEGIN, so a concurrent writer blocks-and-retries via the `busy_timeout` added in #163 rather than racing).
- Both pair call sites now route through it.
- Tests: both rows persist user-before-assistant; **a forced mid-transaction failure rolls back fully — zero half-turns**.

### #160 (b) — atomic `recordRun` counter
`recordRun` did `get()` then `UPDATE` from the stale snapshot, and the recurring branch incremented `run_count` SQL-side while the maxRuns/one-off branches bound a precomputed value. A concurrent `tick` (separate process) could read the same `runCount` and **lose a count**.

- Read+write now wrapped in one `.immediate` transaction; `run_count = run_count + 1` SQL-side in **every** branch, so the count is exact and the deactivation decision matches the value actually written.
- Existing recordRun tests (recurring->1, one-off->1, maxRuns->3 then deactivate) still pass unchanged.

### Not in this PR
- **#159** (Telegram offset acked before processing) was pulled out: turns are deliberately fire-and-forget for concurrency, so true at-least-once needs a design decision (durable inbox vs. ack-watermark + dedup), not a patch. Writing that up separately for Andrew.
- **#160 (c) fetch timeouts** and **#160 (d) lost partial stream** -> PR B2 (transport resilience).

### Verification
`tsc --noEmit` clean · **full suite 1171 pass / 0 fail** (+2 new atomic tests).

Part of #160.

### DIFF
--- a/src/lib/tasks.ts
+++ b/src/lib/tasks.ts
@@ -362,44 +362,42 @@ export class TaskStore {
    * For recurring with maxRuns: deactivates if runCount >= maxRuns.
    */
   recordRun(id: number, now: Date = new Date()): void {
-    const t = this.get(id);
-    if (!t) return;
+    // Read-modify-write must be atomic: tick can fire concurrently across
+    // processes, and a stale-snapshot read + precomputed run_count write
+    // would let two runs both bind `runCount + 1` and lose a count. We
+    // hold the write lock for the whole transaction (`.immediate`) and
+    // increment run_count SQL-side in every branch, so the count is always
+    // exact and the maxRuns/one-off deactivation decision is consistent
+    // with the value actually written.
+    this.db.transaction(() => {
+      const t = this.get(id);
+      if (!t) return;
 
-    const nextRunCount = t.runCount + 1;
+      const nowIso = now.toISOString();
+      const nextRunCount = t.runCount + 1;
 
-    // Check maxRuns — deactivate if hit.
-    if (t.maxRuns !== undefined && nextRunCount >= t.maxRuns) {
+      // maxRuns hit, or one-off: deactivate after this run.
+      if ((t.maxRuns !== undefined && nextRunCount >= t.maxRuns) || t.oneOff) {
+        this.db
+          .prepare(
+            `UPDATE tasks
+             SET last_run_at = ?, run_count = run_count + 1, active = 0
+             WHERE id = ?`,
+          )
+          .run(nowIso, id);
+        return;
+      }
+
+      // Recurring: advance next_run_at.
+      const next = nextFire(t.schedule, now);
       this.db
         .prepare(
           `UPDATE tasks
-           SET last_run_at = ?, run_count = ?, active = 0
+           SET last_run_at = ?, next_run_at = ?, run_count = run_count + 1
            WHERE id = ?`,
         )
-        .run(now.toISOString(), nextRunCount, id);
-      return;
-    }
-
-    // One-off: deactivate after single run.
-    if (t.oneOff) {
-      this.db
-        .prepare(
-          `UPDATE tasks
-           SET last_run_at = ?, run_count = ?, active = 0
-           WHERE id = ?`,
-        )
-        .run(now.toISOString(), nextRunCount, id);
-      return;
-    }
-
-    // Recurring: advance next_run_at.
-    const next = nextFire(t.schedule, now);
-    this.db
-      .prepare(
-        `UPDATE tasks
-         SET last_run_at = ?, next_run_at = ?, run_count = run_count + 1
-         WHERE id = ?`,
-      )
-      .run(now.toISOString(), next.toISOString(), id);
+        .run(nowIso, next.toISOString(), id);
+    }).immediate();
   }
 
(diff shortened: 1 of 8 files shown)


src/memory/store.ts
```diff
@@ -210,6 +220,15 @@ class SqliteMemoryStore implements MemoryStore {
       `SELECT COUNT(*) AS n FROM capture_log
```
andrewagrahamhodges, line 220: 🔵 **Trivial (Praise)**: The transaction definition is clean — two INSERTs sharing a single `created_at` timestamp, with autoincrement id providing the sort tiebreak. The `.immediate` call site at line 249 is well-documented with the write-lock rationale.

src/lib/tasks.ts
```diff
@@ -362,44 +362,42 @@ export class TaskStore {
...
+    // processes, and a stale-snapshot read + precomputed run_count write
+    // would let two runs both bind `runCount + 1` and lose a count. We
+    // hold the write lock for the whole transaction (`.immediate`) and
+    // increment run_count SQL-side in every branch, so the count is always
+    // exact and the maxRuns/one-off deactivation decision is consistent
+    // with the value actually written.
+    this.db.transaction(() => {
+      const t = this.get(id);
+      if (!t) return;
+
```
andrewagrahamhodges, line 375: 🔵 **Trivial (Praise)**: Unifying all branches to use `run_count = run_count + 1` SQL-side eliminates the entire class of stale-snapshot bugs. The single `this.db.transaction()` wrapper keeps the read and write in one atomic unit.

andrewagrahamhodges (approved):
## Code Review: Transactional Integrity — Atomic Turn Pair + RecordRun Counter

**Author:** @robertclawson | **Files:** 8 | **Verdict:** Approve

- `appendTurnPair` is called **after** the harness yields its final chunk — it persists completed turns to SQLite. It does not touch the streaming pipeline, bubble splitting, or Telegram delivery.
- `recordRun` is a task-scheduler counter update — purely persistence logic, no runtime behavioral change.
- The Telegram offset comment is documentation-only — the `offset = nextOffset` line was already in the same position before this PR.
- Zero files in `src/orchestrator/turn.ts`'s streaming path, `src/lib/harnessRunner.ts`, or `src/channels/telegram.ts`'s bubble/splitting logic are modified.

### Fix 1: Atomic Turn Pair (#160a) — DATA INTEGRITY

Two independent `appendTurn` INSERTs replaced with `appendTurnPair` — a single `.immediate` transaction that writes both rows or neither. Prevents half-turns (user message with no assistant reply) on crash. Both call sites updated: `orchestrator/turn.ts` (normal turns) and `telegram.ts` (interrupted-pair).

The `.immediate` modifier takes the write lock at BEGIN, so a concurrent writer blocks-and-retries (via the `busy_timeout` from #163) instead of racing into a read→upgrade deadlock.

### Fix 2: Atomic recordRun Counter (#160b) — DATA INTEGRITY

Read-modify-write wrapped in a `.immediate` transaction with `run_count = run_count + 1` SQL-side in every branch. Prevents lost counts from concurrent tick processes reading the same stale snapshot.

### Documentation Blocks

Three large comment blocks added to `claude.ts` (Cron* deny-list), `threatJudge.ts` (security perimeter), and `persona/builder.ts` (memory system). These are **deliberate architectural guardrails** — they prevent future maintainers from "simplifying" away load-bearing features. Documentation-only, zero runtime impact.

### Findings

| 1 | 🔵 Trivial | src/memory/store.ts | 220 | Clean transaction definition — the `.immediate` call site is well-documented. |
| 2 | 🔵 Trivial | src/lib/tasks.ts | 375 | Unified branch logic with SQL-side increment eliminates the stale-snapshot class of bugs. |

### What's Good
- **Minimal, surgical changes** — each fix addresses one race condition with the smallest possible transaction scope.
- **Excellent test coverage** — the rollback test (`rolls back fully if the second insert fails — no half-turn`) directly validates the atomic guarantee.
- **Documentation guardrails** — the architectural comment blocks are a smart defensive measure against well-intentioned but destructive "cleanups."

<details>

```
Nothing to change. Clean transactional integrity fixes with excellent test coverage.
```
</details>

kaieriksen (approved):
Reviewed the latest head. No blocking issues found; local test, typecheck, build, and diff hygiene all pass.

Issue #171, opened 2026-06-11
Persona-as-judge threat rating + rolling per-chat buffer for escalation grounding


robertclawson:
## Summary

Two coupled changes to the untrusted-input security perimeter:

1. **Persona-as-judge** — replace the FTS-snippet briefing (`makeJudgeBriefing` / `renderPriors`) with the **full persona** (soul, memory, drawers, context) run in a no-tools environment under a **narrow** prompt whose only output is a `0–100` rating of how likely the input is a **prompt-injection attempt**.
2. **Rolling per-chat buffer** — a per-conversation append-only log of inbound messages, agent replies, **and judge notifications**, replayed (last N) into trusted turns so that when the principal approves/denies a held escalation, the agent can actually see *what was held* and record the ruling to memory.

These fix two long-standing concerns:

- **#1** — the judge doesn't visibly weigh experience, so it can't get quieter over time and we have no telemetry to know if it ever does.
- **#2** — when the judge holds and notifies, the held request and the notification are persisted **nowhere**. A later "approve / deny" reply from the principal arrives as a fresh trusted turn with no referent, so the agent can't ground it and can't record the ruling — which means the judge never learns from it either.

## Current state (for reference)

- `src/lib/threatJudge.ts` — builds the judge briefing by semantic-searching the decisions/people/norms drawers and rendering the **top FTS snippets** (`hit.snippet`, truncated) into a `<briefing>` block, then scores.
- `src/orchestrator/screen.ts` — screens untrusted input ahead of any capable turn; on a high score it **holds** and surfaces.
- `src/orchestrator/turn.ts` — a held turn yields its message and **returns without persisting** (`appendTurnPair` is never reached), so held requests leave no trace in history.
- `src/cli/notify.ts` — one-way push to Telegram; writes **nothing** to the memory/conversation store.
- `src/persona/builder.ts` — `buildSystemPrompt(...)` assembles the persona (soul, memory sections, security-perimeter sections). This is what the judge should reuse.

## Concern #1 — judge should weigh experience, and we should be able to see that it does

**Root issue:** the briefing feeds *truncated search fragments*, not full rulings. A decision recorded with its weight (e.g. *"approve invoice PDFs from billing@x, but bank-detail changes always come back to me"*) can reach the judge cut off mid-sentence — the nuance the principal carefully recorded is lost. And there is **zero telemetry**, so "does experience lower the score?" is unanswerable today.

## Concern #2 — held escalations are ungrounded

[...]

PR #172, opened 2026-06-11
feat(security): persona-as-judge + grounded escalation (#171)

robertclawson:
Implements #171 — two coupled changes to the untrusted-input security perimeter. No merge intended yet: review → compile bun binary → test on a persona (Lena/Kai) → merge.

### 1. Persona-as-judge (issue parts A/B/C)
The threat judge no longer runs as a bare module-const classifier briefed by truncated FTS snippets. It now runs as the **full persona** — identity + MEMORY + the `decisions`/`people`/`norms` drawers fed **in full** — narrowed by a new `JUDGE_NARROWING` instruction to one tool-less job: rate `0–100` how likely the input is a **prompt-injection attempt** (manipulation likelihood), explicitly *not* how risky the legitimate task is. The persona's lived sense of what's routine *is* the briefing, at full fidelity (fixes concern #1's "nuance lost to truncation").

- Threshold unchanged at **80**: `<80` silent pass (no notify), `>=80` hold + notify.
- `JUDGE_NARROWING` is **principal-neutral** (no hardcoded owner name).
- Still **tool-less** / read-only floor — the security properties are unchanged; only the *material* the judge sees got richer.
- `makeJudgeBriefing`/`renderPriors` (the FTS-snippet path) **removed**. The `<briefing>`/`priors` channel + the marker-strip guard are **kept** (direct-caller fallback + back-compat).
- If the persona can't load, the judge falls back to the legacy `JUDGE_SYSTEM` classifier and **still screens**.

### 2. Grounded escalation (issue parts D/E/F)
**The crux:** a held untrusted episode arrives under one conversation, but the principal's approve/deny reply lands in their **Telegram** conversation (`telegram:<userId>`) — a *different* conversation. The 30-turn replay is per-conversation, so the grounding write must target the **principal's** conversation (the one the notify goes to), *not* the untrusted entry point's.

On a hold, after notifying, the screener writes a turn **pair** into each principal Telegram conversation:
- **quarantined** user turn — the raw payload, `embeddable=0`
- **embeddable** assistant turn — the judge's notification text

This replays in the principal's next-turn window so "yes/no" has a referent and the agent can record the ruling — closing the learning loop back to #1.

**Embedding rule (part F):** new `embeddable` column on `turns` (+ idempotent migration). Quarantined rows **replay in history** (`recentTurns` unchanged) but are **skipped by the turn indexer** (never FTS-indexed/embedded) and **purged** by `purgeQuarantined` after a trusted turn rules — so verbatim untrusted text never enters the embedded corpus and isn't retained long-term.

## Files
- `lib/threatJudge.ts` — `JUDGE_NARROWING`, optional `opts.systemPrompt`; comments augmented, not deleted.
- `orchestrator/screen.ts` — persona-as-judge build, full-drawer briefing (16KB cap), `recordHeld` grounding write, FTS path removed.
- `memory/store.ts` — `embeddable` column + migration, per-turn flag, `purgeQuarantined`.
- `orchestrator/turnIndexer.ts` — skip `embeddable=0` rows (cursor still advances).
- `orchestrator/turn.ts` — purge quarantined on trusted-turn success.
- `channels/core/engine.ts`, `cli/ask.ts` — pass the memory store into `makeScreener`.
- tests — grounding, quarantine, purge, indexer-skip, judge `systemPrompt`.

## Judgment calls for review
- **Grounding write lives in the screener, not `notify.ts`.** notify has no conversation context and no business writing memory; the screener knows the verdict and the principal account. `notify.ts` is unchanged.
- **Purge-after-ruling = purge on the next trusted turn in that conversation.** Deliberate tradeoff: the raw payload is visible for grounding for exactly one trusted turn, then dropped. The judge-reasoning turn + any `memory capture --tag decision` are kept. If the principal addresses one of two simultaneous holds, the other's raw payload is also purged (its judge-text turn survives).
- **Drawers fed via `buildSystemPrompt`'s retrieved-context slot**, capped at 16KB.
- **Cross-persona isolation preserved** — a ruling briefs only that persona's judge.

## Verification
- `bun tsc --noEmit` clean
- `bun test` → **1202 pass, 1 skip, 0 fail** (78 files)
- This is a deliberate **reversal** of the prior "judge runs without persona / inert classifier" stance — approved in conversation. Security properties (tool-less, marker-strip, fail-open on infra error, hold-never-downgraded) are all preserved.

### DIFF
--- a/src/orchestrator/screen.ts
+++ b/src/orchestrator/screen.ts
@@ -10,36 +10,69 @@
  * Flow, all IN CODE so a model can never fake it (the bug that started this
  * whole redesign was a model *claiming* it had notified/recorded):
  *
- *   1. BRIEFING — semantic-search the threat-relevant drawers (decisions =
- *      prior rulings, people = known senders, norms = what's routine in
- *      Andrew's world) so the judge isn't an amnesiac that cry-wolfs on
- *      normal operations. DELIBERATELY scoped to those three drawers, NOT a
- *      raw memory dump: the judge does not need Andrew's finances/inbox to
- *      score a threat, and keeping them out means they never land in a judge
- *      log either. Best-effort; failure → no priors. Prior rulings only ever
- *      LOWER scrutiny for things he already blessed; they never clear it (the
- *      judge is told a catastrophic action re-escalates regardless).
- *   2. JUDGE — run the tool-less harness judge over the content + priors.
- *      It returns a score 0–100. The judge has no tools, so it cannot act
- *      on what it reads; we consume only its number.
+ *   1. BRIEFING — the judge runs as the FULL PERSONA, narrowed. This is the
+ *      deliberate reversal the principal approved: instead of semantic-
+ *      searching three drawers and feeding the judge a handful of TRUNCATED
+ *      FTS snippets, the screener loads the persona (identity + MEMORY) and
+ *      reads the decisions/people/norms drawers IN FULL, then composes them
+ *      into the judge's system prompt via buildSystemPrompt + JUDGE_NARROWING.
+ *      The judge therefore has the principal's real context — identity, prior
+ *      rulings, known senders, documented norms — at full fidelity, which
+ *      fixes the old "nuance lost to truncation" problem (concern #1). The
+ *      drawers are still scoped to decisions/people/norms (capped at ~16KB),
+ *      so finances/inbox stay out of the judge; what changed is they are now
+ *      full rulings, and MEMORY/identity also inform the judge. Best-effort:
+ *      if the persona can't load, the judge falls back to the module
+ *      JUDGE_SYSTEM classifier (no persona context) and still screens.
+ *   2. JUDGE — run the tool-less harness judge over the content, with the
+ *      narrowed persona as its system prompt. It returns a score 0–100. The
+ *      judge has no tools, so it cannot act on what it reads; we consume only
+ *      its number. (The legacy <briefing>/priors channel still EXISTS in
+ *      threatJudge — we just stop populating it; persona context replaces it,
+ *      so priors is passed empty.)
  *   3. score <  THREAT_THRESHOLD → {action:"pass"}; the turn proceeds
- *      silently. No notification — quiet when safe (Andrew's "don't nag").
+ *      silently. No notification — quiet when safe (the "don't nag" rule).
  *   4. score >= THREAT_THRESHOLD → HOLD (fail-closed):
  *        - The untrusted turn does NOTHING. runTurn returns the heldMessage
  *          instead of running the harness. Untrusted entry points are
  *          one-shot, so "held" == the action simply never happened — the
- *          fail-closed default Andrew chose (option b). There is no paused
- *          process to time out; if he wants it done, he says so.
+ *          fail-closed default. There is no paused process to time out; if
+ *          the principal wants it done, they say so.
  *        - `phantombot notify` opens a CONVERSATION on Telegram (in CODE):
  *          what arrived, why it tripped, and the concern to weigh —
  *          phrased to be talked through, not answered yes/no.
+ *        - GROUNDING WRITE (the crux fix, concern D+E): the held episode is
+ *          written into the PRINCIPAL'S telegram conversation(s) — NOT the
+ *          untrusted entry point's. See recordHeld below for why.
  *
  * What the screener deliberately does NOT do: write a decision. Decisions
- * are recorded ONLY from a TRUSTED turn — i.e. when Andrew talks it through
- * on Telegram and concludes. The judge writes nothing; the untrusted turn
- * writes nothing. That is the whole point: an attacker can never author
- * "Andrew approved this". His trusted reply is the only thing that records
- * a ruling, and that ruling is what recall reads next time.
+ * are recorded ONLY from a TRUSTED turn — i.e. when the principal talks it
+ * through on Telegram and concludes. The judge writes nothing; the untrusted
+ * turn writes nothing. That is the whole point: an attacker can never author
+ * "the principal approved this". The trusted reply is the only thing that
+ * records a ruling, and that ruling is what recall reads next time.
+ *
+ * ─────────────────────────────────────────────────────────────────────────
+ * CONVERSATION SCOPING OF THE HELD EPISODE (concern D+E) — READ THIS.
+ *
+ * A held untrusted episode arrives under SOME entry-point conversation key
+ * (e.g. an email-woken `phantombot ask`), but the principal's approve/deny
+ * reply arrives in THEIR telegram conversation (`telegram:<userId>`) — a
+ * DIFFERENT conversation, the one the notify lands in. The 30-turn history
+ * replay is per-conversation, so for the principal's reply to be GROUNDED in
+ * what was held, the held episode must be written into the PRINCIPAL'S
+ * telegram conversation, the same one the notify went to.
+ *
+ * So on HOLD, after notifying, we write a turn PAIR per principal telegram
+ * conversation: a QUARANTINED user turn carrying the raw untrusted payload
+ * (embeddable:false — never indexed/embedded; see memory/store.ts + the
+ * turnIndexer), and an embeddable assistant turn carrying the judge-
+ * notification text. The pair replays into the principal's next turn so
+ * "yes, go ahead" / "no" has a referent. This is done HERE in the screener,
+ * correctly scoped to the principal — NOT in notify.ts (which only sends a
+ * message and has no business writing memory) and NOT in turn.ts's hold path
+ * (which is scoped to the untrusted entry point, the wrong conversation).
+ * ─────────────────────────────────────────────────────────────────────────
  *
  * Fail-OPEN on judge/recall error by design: if screening itself errors
  * (harness down, bad JSON), the screener returns "pass" and logs. A
@@ -50,16 +83,23 @@
  * answered-vs-unanswered escalation, not an infra error.)
  */
 
+import { readFile } from "node:fs/promises";
+import { join } from "node:path";
+
 import {
   type Config,
-  memoryIndexPath,
   personaDir,
 } from "../config.ts";
-import { geminiEmbed } from "../lib/geminiEmbed.ts";
 import type { Harness } from "../harnesses/types.ts";
 import { log } from "../lib/logger.ts";
-import { MemoryIndex, type SearchHit } from "../lib/memoryIndex.ts";
 import {
+  buildSystemPrompt,
+  type ChannelContext,
+} from "../persona/builder.ts";
+import { loadPersona, type PersonaFiles } from "../persona/loader.ts";
+import type { MemoryStore } from "../memory/store.ts";
+import {
+  JUDGE_NARROWING,
   judgeThreat,
   makeChainJudgeComplete,
   THREAT_THRESHOLD,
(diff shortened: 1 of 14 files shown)


src/orchestrator/screen.ts
```diff
@@ -143,28 +256,91 @@ export function makeScreener(
...
+      });
+    });
+
+  // Back-compat: an older test may inject `recall` (expecting its output to
+  // reach the judge as priors). Honour it; production leaves it unset and the
+  // persona context is the briefing.
+  const recall = deps.recall;
 
   const notify =
     deps.notify ?? ((message: string) => runNotify({ config, message }));
```
kaieriksen: This default notify path drops the persona, so a held request for a persona-bound bot is sent through the default Telegram bot while `principalConversations()` below records the held episode under the persona-bound account. In a multi-persona install that means the owner replies in the default bot conversation, but the grounding payload was written to the other persona's memory conversation, so the approval/denial turn is not grounded. Route notify through the same account selection as `principalConversations` (for example pass `persona` when a persona Telegram account exists, otherwise fall back to default) and add a regression test for a non-default persona.

kaieriksen (changes requested):
The core design is solid and the focused tests/typecheck pass, but this multi-persona routing gap needs fixing before merge.

lenaparkhodges (dismissed):
## Review — PR #172: persona-as-judge + grounded escalation

LGTM. This is a well-designed, well-documented PR that meaningfully upgrades the threat judge's signal quality while preserving every security property.

### What I verified

**1. Security properties preserved (fail-closed, never downgraded)**
- Judge is still tool-less. `JUDGE_NARROWING` explicitly says "you have NO tools and you do not act." The persona's tool descriptions are in the prompt but the narrowing overrides them — the judge can only output a JSON score.
- `<briefing>` / `<untrusted_content>` marker-strip in `threatJudge.ts` is untouched. An attacker still can't forge priors.
- Fail-OPEN on judge error (harness down, bad JSON) — unchanged.
- Hold never downgraded to pass even if notify or grounding write fails — verified in `screen.ts`.

**2. Grounding write targets the correct conversation**
The crux of concern D+E. Verified: `principalConversations()` resolves `config.channels.telegram.allowedUserIds` (or the persona-bound account) to `telegram:<userId>` keys. The grounding pair is written there — NOT to the untrusted entry point's conversation. This means the principal's approve/deny reply replays the held payload + judge text in its 30-turn window.

**3. Quarantine lifecycle is clean**
- `embeddable: false` on the raw user turn, `true` on the assistant (judge reasoning) turn.
- `turnIndexer.ts` skips `embeddable=0` rows but advances the cursor — quarantined turns don't stall indexing.
- `purgeQuarantined` is called on trusted success, drops only `embeddable=0` rows. Safe no-op when nothing is quarantined.
- `recentTurns` returns ALL rows (including quarantined) so grounding replay works.

**4. Persona-as-judge prompt construction**
- `loadPersonaFiles()` is lazy-cached per screener instance — persona loaded once, not per turn.
- `readBriefingDrawers()` reads the 3 scoped drawers (decisions/people/norms) in full, capped at 16KB. Called per screened turn, but these are tiny files — negligible vs the harness call.
- `JUDGE_NARROWING` is principal-neutral (verified by the test asserting no "andrew"/"robbie" match).
- Fallback: if persona can't load, `systemPrompt` stays `undefined` and `judgeThreat` uses the legacy `JUDGE_SYSTEM` classifier.

**5. Idempotent migration**
`PRAGMA table_info(turns)` check → `ALTER TABLE` only when needed. Existing rows default to `embeddable=1` (indexable), which is correct for pre-quarantine data.

### One minor note (non-blocking)

The `buildSystemPrompt` output includes `SECURITY_PERIMETER_UNTRUSTED_SECTION` which hardcodes "Andrew" — when running as a different persona's judge this is technically noise. It doesn't matter because `JUDGE_NARROWING` says "disregard any earlier instructions" and the judge has no tools to act on those instructions anyway. But if you want to be surgically clean, you could pass `channel: "judge"` (a non-matching channel key) so the builder skips the perimeter section entirely. Not worth blocking on.

### Test coverage

Good breadth: grounding write scoping, quarantine persist/skip/purge, indexer cursor advancement past quarantined rows, systemPrompt override/fallback, `JUDGE_NARROWING` neutrality, purge-on-trusted-only, no-purge-on-failed-trust. The `stubMemory` pattern in `orchestrator-screen.test.ts` is clean.

### Verdict

Approving. The code is clean, the security model is sound, and the test coverage is comprehensive. Ready for the persona-level smoke test the PR description calls for before merge.

src/orchestrator/screen.ts
```diff
@@ -86,11 +126,47 @@ const PASS_ON_ERROR = (score: number, reason: string): ScreenVerdict => ({
...
+ * rulings), people (known senders), norms (what's routine in the principal's
+ * world). Scoping the persona-as-judge briefing to these three keeps it
+ * threat-relevant and keeps sensitive operational memory (finances, inbox,
+ * daily dumps, commitments) out of the judge entirely. Read in FULL now, not
+ * as snippets — see DRAWERS_CAP_BYTES. Paths are relative to the persona dir.
+ */
+const BRIEFING_DRAWERS: readonly string[] = [
+  "memory/decisions.md",
+  "memory/people.md",
+  "memory/norms.md",
```
andrewagrahamhodges, line 140: 🔵 **Trivial (Praise)**: The persona loading pattern — lazy, cached, best-effort with fallback to the legacy `JUDGE_SYSTEM` classifier — is exactly right. A persona dir issue must never silently disable screening.

src/orchestrator/turnIndexer.ts
```diff
@@ -86,6 +86,17 @@ export async function indexConversationTurnsIfDue(
       if (turns.length === 0) break;
 
       for (const turn of turns) {
+        // QUARANTINED turn (embeddable=0): a held untrusted payload the
+        // screener wrote into the principal's conversation to ground their
+        // approve/deny. It must NEVER reach the FTS/vector index — that is
```
andrewagrahamhodges, line 91: 🔵 **Trivial (Praise)**: Advancing the cursor past quarantined rows (incrementing `indexed` and `afterId`) without indexing them is the correct design — it prevents the indexer from re-scanning the same quarantined row forever while keeping it out of the FTS/vector corpus.

andrewagrahamhodges (dismissed):
## Code Review: Persona-as-Judge + Grounded Escalation

**Author:** @robertclawson | **Files:** 14 | **Verdict:** Approve

- `src/channels/core/engine.ts` changes are limited to passing `input.memory` to `makeScreener` (12 additions, 3 deletions). The streaming pipeline (`splitIntoSegments`, `streamedReply`, `bubbleMaxChars`, `sendFinalSegments`) is **completely untouched**.
- The screener runs **before** the harness chain — it gates what enters the chain, not how output is streamed or split into bubbles.
- The grounding write (quarantined turn pair) happens **after** the screener returns and **before** `runTurn` yields any chunks. It writes to SQLite, not to the streaming pipeline.
- `heldMessage` content is unchanged (only "Andrew" → "the owner" — a cosmetic text fix).

### What Changed

#### 1. Persona-as-Judge (Parts A/B/C)
The threat judge now runs as the **full persona** — identity + MEMORY + the `decisions`/`people`/`norms` drawers fed **in full** (capped at 16KB) — narrowed by `JUDGE_NARROWING` to one tool-less job: rate 0–100 how likely the input is a **prompt-injection attempt**. The persona's lived sense of what's routine *is* the briefing, at full fidelity.

- Threshold unchanged at **80**: `<80` silent pass, `>=80` hold + notify.
- `JUDGE_NARROWING` is **principal-neutral** (no hardcoded owner name).
- Still **tool-less** / read-only — security properties unchanged.
- Fallback: if persona can't load, judge falls back to legacy `JUDGE_SYSTEM` classifier and **still screens**.

#### 2. Grounded Escalation (Parts D/E/F)
On HOLD, the screener writes a turn **pair** into each principal Telegram conversation:
- **Quarantined** user turn — raw payload, `embeddable=0` (never indexed/embedded)
- **Embeddable** assistant turn — judge notification text

This replays in the principal's next-turn window so "yes/no" has a referent. The `embeddable` column + migration ensure quarantined rows are FTS-skipped and purged after a trusted turn rules.

### Findings

| 1 | 🔵 Trivial | src/orchestrator/screen.ts | 140 | Clean persona loading with best-effort fallback to legacy classifier. |
| 2 | 🔵 Trivial | src/orchestrator/turnIndexer.ts | 91 | Correct cursor advancement past quarantined rows — prevents infinite re-scan. |

### What's Good
- **Security properties preserved** — tool-less, marker-strip, fail-open on infra error, hold-never-downgraded all intact.
- **Quarantine lifecycle is complete** — write (screener) → replay (history) → skip (indexer) → purge (trusted turn). Each stage has explicit test coverage.
- **Cross-persona isolation** — a ruling briefs only that persona's judge; grounding writes target only the principal's Telegram conversation.
- **Graceful degradation** — persona load failure → legacy classifier; grounding write failure → hold still stands; purge failure → no error.

<details>

```
Nothing to change. Clean persona-as-judge + grounded escalation with complete quarantine lifecycle.
```
</details>

lenaparkhodges:
Re Kai's thread on line 299: confirmed, this is a real gap. `principalConversations()` correctly resolves to the persona-bound Telegram account (or falls back to default), but `runNotify({ config, message })` always routes through the default bot. In a multi-persona install the notification lands in the default bot's chat while the grounding write goes to the persona-bound bot's conversation — so the principal's reply in the notification chat has no grounded context.

The fix is straightforward: pass `persona` through to `runNotify` when the persona-bound account exists (the `--persona` flag already exists in the notify CLI, it's just not wired here). Regression test should cover a non-default persona where `telegramPersonas[persona]` differs from `telegram` default.

src/orchestrator/screen.ts
robertclawson: Fixed in a4e042f. Routed the escalation notify through the same account selection as `principalConversations()`: extracted `resolveNotifyPersona(config, persona)` which returns the persona name when `channels.telegram.personas.<persona>` is configured (so `runNotify` routes through the persona-bound bot) and `undefined` otherwise (default bot). `runNotify` already accepts `--persona`, so its account resolution + error messaging is reused.

Added a regression test (`describe("resolveNotifyPersona …")` in `tests/orchestrator-screen.test.ts`) covering: default persona → undefined, a non-default persona with its own bot → that persona name, and a screened persona with no bot while another persona has one → default fallback. Full suite green (1206 pass), typecheck clean.

robertclawson:
@kaieriksen good catch — fixed in a4e042f.

The notify and the grounding write disagreed on which bot in a multi-persona install: `principalConversations()` resolves the persona-bound bot (`telegramPersonas[persona] ?? telegram`), but the notify call always went through the default bot. So for a persona with its own Telegram bot, the owner got pinged in the default chat while the held payload was written to the persona-bound conversation → the approve/deny reply landed ungrounded.

Fix: added `resolveNotifyPersona()` mirroring `principalConversations()`'s account selection exactly, and pass it to `runNotify` so notify reuses its own account resolution + error messaging. Regression tests cover three cases: persona-bound bot (→ persona), no persona bot (→ default), and screening as a persona that has no bot while another does (→ default).

`bun tsc --noEmit` clean, full suite 1206 pass / 0 fail. Re-requested your review.

lenaparkhodges (approved):
## Re-review — PR #172: persona-as-judge + grounded escalation

**After fix a4e042f:** My previous concern (notify routing mismatch in multi-persona installs) is fully resolved.

### What changed since my last review
Robbie added `resolveNotifyPersona()` — a clean two-line function that mirrors `principalConversations()`'s account selection exactly: returns the persona name when `telegramPersonas[persona]` is configured, `undefined` otherwise. It's wired into the `runNotify` call so the escalation ping goes to the same bot conversation where the grounding pair was written. Four regression tests cover: persona-bound bot, no persona bot, cross-persona fallback, and the key invariant that notify routing matches `principalConversations` account selection.

### Verified this time
- **Type check:** `tsc --noEmit` clean
- **Test suite:** 1206 pass, 0 fail
- **Notify routing invariant:** `resolveNotifyPersona` returns `persona` iff `config.channels.telegramPersonas[persona]` exists — same condition `principalConversations` uses to pick the bot account. The owner gets pinged and grounded in the same conversation.
- **Security properties:** still intact — tool-less judge, marker-strip, fail-open, hold-never-downgraded. All unchanged from previous review.
- **Quarantine lifecycle:** write → replay → skip → purge — still clean with no regressions.

### Verdict
My threads are resolved. Approving. Ready to merge.

kaieriksen (approved):
Re-reviewed the latest head a4e042f. The persona routing blocker is fixed: escalation notify now routes through the same persona-bound bot account used for the grounding write, and the held episode is recorded into the matching principal Telegram conversation.

Verified locally:
- bun test tests/orchestrator-screen.test.ts tests/orchestrator-turn.test.ts tests/memory.test.ts tests/orchestrator-turnIndexer.test.ts tests/lib-threatJudge.test.ts tests/orchestrator-retrieval.test.ts tests/lib-memoryIndex.test.ts
- bun tsc --noEmit

kaieriksen (approved):
Re-reviewed head a4e042f. The blocker from my previous review is fixed: escalation notify now routes through the same persona-bound Telegram account selection used for the held-episode grounding write, so the owner approval/denial reply lands in the conversation that contains the referent.

Verified locally:
- bun test tests/orchestrator-screen.test.ts tests/orchestrator-turn.test.ts tests/orchestrator-turnIndexer.test.ts tests/memory.test.ts tests/lib-threatJudge.test.ts (91 pass, 0 fail)
- bun run typecheck

No remaining findings from me.

Issue #162, opened 2026-06-10
Channel abstraction: extract channel-agnostic core + Channel interface (Telegram first adapter, Matrix next)


robertclawson:
## Goal

Make phantombot **multi-channel**. Matrix is coming next, with Slack / Google Chat likely after. Define a `Channel` seam now so a new platform is a **thin adapter (~transport + parse)** over a **shared, channel-agnostic core** — instead of copying the 660-line streaming engine once per platform.

This **supersedes** the previous "split telegram.ts into telegramTransport / telegramParse / groupRouting / telegramPrompts" proposal. That split carved along *file-size* seams, kept everything `telegram`-prefixed, and would have duplicated the streaming engine N× across channels — the opposite of what multi-channel needs. The docs-sync half of the old #162 (AGENTS.md) is split out to its own issue, to be done last.

## Why the old split was wrong

The dangerous part of `telegram.ts` is `processChatMessage` (`src/channels/telegram.ts:1531–2214`, ~660 lines). The structural reality:

- **~40% is channel-agnostic** and load-bearing: the streaming state machine (`streamedReply` / `consumedReplyChars` / `narrationBuffer` / `finalSegmenter`, `1804–1830`), the `runTurn()` harness invocation (`1898–1961`), the chunk-consumption loop (`1962–2024`), abort + interrupted-pair persistence (`2037–2084`), error recovery (`2086–2105`), and output dedup (`2114–2148`).
- **~60% is Telegram-specific**: voice STT (`1579–1644`), attachment download to inbox (`1650–1704`), bot-mention strip (`1565–1575`), typing/recording indicators, voice synthesis (`2150–2181`), Telegram sends.

A file-size split leaves that 40% generic core welded inside a Telegram file. Add Matrix → copy the whole thing → maintain the streaming engine in 4 places. **We extract the generic core once, behind an interface.**

## Good news: two seams already exist

1. **Transport is already an interface.** `TelegramTransport` (`telegram.ts:193–241`) is a real contract with a concrete `HttpTelegramTransport` impl (`266–509`) — tests already stub it. Generalizing it is renaming + capability flags, not a rewrite.
2. **Messages are already normalized.** Raw Telegram JSON is parsed into an internal `TelegramMessage` (`157–191`) by `parseGetUpdatesResult` (`952–1035`); the raw shape never leaks downstream. We rename this to `ChannelMessage` and Telegram becomes one producer of it.

[...]

Issue #154, opened 2026-06-05
Channel abstraction: decouple from Telegram to allow alternative transports (Matrix, …)


14skywalker:
### Context
A user wants to move off Telegram for **privacy/governance reasons** and asked how hard a Matrix channel would be. Digging through the channel layer, the interesting finding isn't "add Matrix" — it's that the codebase is ~90% ready for *any* transport but pinned to Telegram at a handful of load-bearing seams. Filing this as a design discussion, not a feature demand.

### What's already good
- `runTurn` (orchestrator) is already channel-agnostic — untouched by this.
- `TelegramTransport` is a clean interface (`getUpdates`/`sendMessage`/`sendVoice`/`downloadFile`). A `MatrixTransport` over `/sync` maps almost 1:1.

### The coupling points (verifiable in source)
1. No generic `Channel` abstraction — `runTelegramServer` is Telegram-shaped, and `notify.ts`, `heartbeat.ts`, `updateNotify.ts` all reach directly into `config.channels.telegram`.
2. `chatId: number` is hardcoded across the channel layer. Matrix room ids are **strings** (`!room:server`). This is a type change that threads through everything — invasive but mechanical.
3. Config schema has no `channels.<other>` slot.

### Design space (discussion, not prescription)
- Introduce a `Channel` interface so Telegram + a second transport can run side-by-side during migration?
- Widen `chatId` to an opaque string id at the abstraction boundary.
- Where should the notify/heartbeat surfacing layer get its channel binding from?

### Caveat / scope
Based on a read of the current channel layer, not a running Matrix prototype. The **E2EE** part of Matrix (rust-crypto/olm, device verification, key backup) is a separate, larger piece — an unencrypted bot on a self-hosted homeserver is easy; an encrypted one is its own project. Flagging so the privacy motivation doesn't get conflated with "just add a transport."

andrewagrahamhodges:
Great analysis on the coupling points — this is exactly the kind of design discussion worth having early.

Before settling on Matrix as the first alternative transport, it'd be worth discussing what other chat clients the community is interested in. In my experience building OpenClaw, I explored most of the popular options and found that **the vast majority are actively bot-hostile**:

- **Signal** blocked my account without notice, even though the bot had a dedicated SIM card. Zero appeal process.
- **WhatsApp** required enterprise-grade configurations (Business API, paid BSP, etc.) and still feels like they're designed to prevent exactly this use case.
[...]

PR #168, opened 2026-06-10
refactor(channels): channel abstraction + encryption seam (#162)

robertclawson:
## What & why

`src/channels/telegram.ts` had grown to 2362 lines doing five unrelated jobs. With a Matrix channel coming (and possibly Slack / Google Chat later), this PR carves out a **channel abstraction** and places the **encryption seam** so future channels slot in *with no further refactor of the conversational core* — while writing **zero Matrix code and zero crypto code**. Closes the structural half of #162.

**Zero behaviour change. Zero user-visible change.** This is a relocation + retyping refactor, not a rewrite.

## How it's de-risked

- **`src/channels/telegram.ts` is now a thin barrel** re-exporting everything that moved. All 28 previously-exported names are preserved, so **no importer and no test was edited**.
- **The streaming engine (`processChatMessage` + `runTelegramServer`) moved byte-identical** — verified by diffing the extracted slice against `main`. The only non-import edit was one inline type-import path bumped for the new directory depth.
- The 1184-test suite is the contract and stays green.

## New structure

```
src/channels/
  core/
    types.ts     # Channel / ChannelTransport / ChannelMessage / ChannelCapabilities
                 #  + encrypt/decrypt seam + listen() inbound-loop seam (design only)
    routing.ts   # group routing (decideGroupReply, matchPersonaNames, formatGroupContext…)
    prompts.ts   # channel-layer prompts (TELEGRAM_REPLY_INSTRUCTION, capture nudge, voice…)
    engine.ts    # runTelegramServer + processChatMessage — MOVED VERBATIM
  telegram/
    transport.ts # HttpTelegramTransport (TelegramTransport extends ChannelTransport)
    parse.ts     # update parsing, attachments, reply-to, envelope sanitize
    channel.ts   # Telegram Channel adapter — capabilities + IDENTITY encrypt/decrypt
  telegram.ts    # barrel re-export (backward-compat)
```

## The encryption seam (read before adding Matrix)

The core only ever sees **plaintext**. `Channel.encrypt`/`decrypt` are the *only* place transport-level crypto ever happens — decrypt-on-ingest, encrypt-on-egress. Telegram bots have no E2EE, so Telegram's hooks are **identity pass-throughs** (`telegram/channel.ts`), with loud comments. A Matrix/Megolm adapter implements these at the same boundary, leaving the core unaware the wire is encrypted. Hooks return `T | Promise<T>` so async crypto slots in without a signature change.

[...]

PR #175, opened 2026-06-11
feat(chat): add Matrix as a second trusted channel with invisible E2EE

robertclawson:
## Matrix as a second trusted channel (invisible E2EE)

Adds a Matrix front-end alongside Telegram, behind the channel-core abstraction landed in #162. Resolves the build half of #154.

### What this gives you
- **Use Element X (or any Matrix client)** to talk to your phantom, with end-to-end encryption.
- **Matrix is a trusted channel** on par with Telegram — allow-listed MXIDs are the trusted tier, everything else stays untrusted and goes through the threat judge.
- **Reply routing**: a reply goes back on whatever channel the message arrived on. A new `default_channel` decides where *unsolicited* traffic goes — alerts, briefings, task fires, and security holds.

### Invisible E2EE (the design constraint)
`phantombot chat matrix` asks for **server, user, password — nothing else.** The user never learns E2EE exists:
- Device cross-signing + a recovery key are generated and stored automatically; never shown, never prompted.
- Crypto store lives **per-persona, next to `SOUL.md`** (`~/.local/share/phantombot/personas/<persona>/matrix/`), so encryption state migrates with the persona work dir — move the dir to a new VM, keep every key.

### The load-bearing risk, resolved
matrix-js-sdk's rust-crypto is WASM, and phantombot ships as a single `bun build --compile` ELF. The WASM doesn't embed on the default resolution path (ENOENT at runtime), **but** building with the `matrix-org:wasm-esm` condition embeds it natively. The `cryptoWasm.ts` shim handles this so the single-binary model is preserved.

### Security perimeter
- Allow-listed MXIDs = trusted tier (mirror of Telegram allowed users).
- A held episode's grounding pair is written into the **default channel's** principal conversation, so approve/deny keeps its referent in a multi-channel install (the #172/#173 scoping rule, extended).

### CLI
- New `phantombot chat` namespace: `chat telegram`, `chat matrix`, `chat default <telegram|matrix>`.
- `phantombot telegram` preserved as a thin **deprecated alias** (warns, forwards) — no broken muscle memory or docs.

### Verification
- `tsc` clean.
- 31 new Matrix/chat/routing test cases (incl. allow-list rejection + cross-signing bootstrap).
- **Full suite: 1240 pass / 0 fail.**

### Rollout
Same canary path as #172/#173: release build lands on Lena first, 24h window before other phantoms notify their users.

Refs #154


src/channels/matrix/channel.ts
```diff
@@ -0,0 +1,143 @@
+/**
```
andrewagrahamhodges, line 1: 🔵 **Trivial (Praise)**: The encryption seam documentation is excellent — it clearly explains why the hooks look like pass-throughs but the wire IS encrypted (SDK handles Megolm under the hood). This prevents the next reviewer from 'simplifying' away the seam.

src/orchestrator/principalRouting.ts
```diff
@@ -0,0 +1,80 @@
+/**
```
andrewagrahamhodges, line 1: 🔵 **Trivial (Praise)**: Channel-aware routing that correctly maps `default_channel` to the right conversation keys for both Telegram and Matrix. The #172 grounding invariant (notify + grounding write target the same conversation) is preserved across both channels.

andrewagrahamhodges (commented):
## Code Review: Matrix as a Second Trusted Channel

**Author:** @robertclawson | **Files:** 30 | **Verdict:** Comment (not approving yet)

- `src/channels/core/engine.ts` is **not modified**. The streaming pipeline, bubble splitting, and Telegram delivery are completely untouched.
- `src/channels/core/types.ts` is **not modified**. The channel abstraction from #168 is reused as-is.
- `src/channels/telegram.ts` and `src/channels/telegram/` are **not modified**. The Telegram channel continues to work exactly as before.
- `src/orchestrator/screen.ts` changes are limited to importing routing helpers from a new `principalRouting.ts` module — no behavioral change to the screener or streaming.

### What This PR Does

Adds Matrix as a second trusted channel alongside Telegram, implementing the channel-core abstraction from #168:

- **Invisible E2EE** — `matrix-js-sdk` + `rust-crypto` handle Megolm transparently; the `Channel.encrypt`/`decrypt` hooks are pass-throughs (the cipher lives in the SDK). Device cross-signing + recovery key generated automatically, stored per-persona.
- **Trusted tier** — Allow-listed MXIDs are the trusted tier (mirror of Telegram allowed users). Everything else goes through the threat judge.
- **Reply routing** — Replies go back on whatever channel the message arrived on. `default_channel` decides where unsolicited traffic goes (alerts, briefings, task fires, security holds).
- **CLI** — New `phantombot chat` namespace: `chat telegram`, `chat matrix`, `chat default <telegram|matrix>`. Old `phantombot telegram` preserved as a deprecated alias.
- **Security perimeter** — Held episode grounding pair written to the default channel's principal conversation — #172/#173 invariant preserved across both channels.

### Findings

| 1 | 🔵 Trivial | src/channels/matrix/channel.ts | 1 | Excellent encryption seam documentation — explains why the hooks look like pass-throughs but the wire IS encrypted. |
| 2 | 🔵 Trivial | src/orchestrator/principalRouting.ts | 1 | Clean channel-aware routing that preserves the #172 grounding invariant across Telegram and Matrix. |

### What's Good
- **Leverages the #168 abstraction perfectly** — Matrix plugs into `Channel` / `ChannelTransport` / `ChannelMessage` without touching the engine.
- **31 new test cases** covering allow-list rejection, cross-signing bootstrap, routing, and CLI.
- **1240 tests pass / 0 fail** — full suite green.

### ⚠️ Will Not Approve Until Dogfooded

This is a substantial new channel with E2EE, a WASM crypto shim, new CLI surfaces, and cross-channel security perimeter routing. The code looks well-structured and the test coverage is solid, but **we will not approve this until we dogfood this change and work out all the kinks**. Matrix has a different operational profile from Telegram (room semantics, E2EE key management, sync lifecycle), and there are likely edge cases that only surface under real usage — especially around:
- E2EE device verification and recovery key flows
- Cross-signing state persistence across persona directory moves
- Long-poll / sync lifecycle stability under load
- Multi-channel routing edge cases (default_channel switching, persona-bound Matrix bots)
- WASM crypto shim behavior in the compiled single-binary on different architectures

The canary plan (Lena first, 24h window) is the right approach. Let's see it running in production before merging.

<details>

```
No code changes needed at this time. Awaiting dogfood results before approval.
```
</details>

src/channels/matrix/server.ts
```diff
@@ -0,0 +1,275 @@
...
+  msg: import("./types.ts").MatrixChannelMessage,
+  ctx: {
+    input: RunMatrixServerInput;
+    /** TRUSTED tier iff the sender is an allow-listed MXID. */
+    principalAuthenticated: boolean;
+  },
+): Promise<void> {
+  const { input } = ctx;
+  const startedAt = Date.now();
+  const conversationKey = `matrix:${msg.conversationId}`;
```
kaieriksen: This uses the room id as the conversation key (`matrix:!room:hs`), but `principalConversations()` writes held-request grounding to `matrix:<mxid>` for Matrix defaults. That means a held GitHub/Plane/email task can be recorded under `matrix:@owner:hs`, the notification can be sent to the owner DM, and the owner reply arrives here under `matrix:!dmRoom:hs` with no grounding context to approve/deny. Either make Matrix principal DMs sender-scoped in the server, or make the grounding/notify path resolve and use the same DM room id; the two paths must share one exact conversation key.

kaieriksen (changes requested):
Blocked on the Matrix principal conversation key mismatch. Focused Matrix/routing/screen tests and typecheck pass locally, but the implementation breaks the held-request approval flow on Matrix.

robertclawson:
Fixed in `ad8db23`.

@kai you were right — same class as the #172 scoping bug. `server.ts` keyed every inbound turn `matrix:<roomId>`, but both the grounding write (`principalConversations`) and the proactive notify (`notify-matrix`, which sends to the `m.direct` DM room) key `matrix:<mxid>` — so a held request's approve/deny reply landed in a different conversation than the grounding pair and lost its referent.

**Fix:** a principal's 1:1 DM is now sender-scoped (`matrix:<senderMxid>`, the analogue of Telegram's `telegram:<userId>`); group rooms and non-principal senders stay room-scoped. DM detection is a new `isDirect()` capability that reads the bot's `m.direct` account data — the *same* source `notify-matrix` resolves the DM room from, so inbound keying and outbound routing agree — with a 2-member-room fallback.

+3 tests pin the contract: principal DM → `matrix:<mxid>` (asserted equal to `principalConversations`), group room → `matrix:<roomId>`, non-principal in a DM → room-scoped (sender-scoping is principal-only). `tsc` clean, full suite 1243 pass / 0 fail.

Not merging — staying open for Andrew's Lena dogfooding.

kaieriksen (dismissed):
The Matrix principal conversation key blocker is fixed in ad8db23. I verified the new sender-scoped principal DM path against principalConversations, kept group/non-principal rooms room-scoped, and ran focused Matrix/routing/screen tests plus typecheck locally. No remaining blockers from me.

src/config.ts
```diff
@@ -84,6 +84,53 @@ export interface TelegramAccount {
...
+  /**
+   * If non-empty, only these MXIDs are TRUSTED principals (mirror of
+   * Telegram `allowedUserIds`). An empty list means "answer anyone" but —
+   * exactly as for Telegram — an open bot is NOT an authenticated principal,
+   * so trust stays false and security-rule writes fail closed. Allow-listing
+   * is the only thing that grants TRUSTED tier on Matrix inbound.
+   */
+  allowedUserIds: string[];
+  /**
+   * Whether to bootstrap + run end-to-end encryption (Megolm via rust-crypto).
```
kaieriksen, line 123: Non-blocking: the new `e2ee` field comment is accurate, but the `MatrixAccount` block comment just above still says there is no encryption knob and E2EE is always on. Worth updating so future readers do not follow the old contract.

kaieriksen (dismissed):
Re-reviewed the new plaintext-by-default Matrix commit. No blocking issues found.

Local verification:
- /home/kai/.bun/bin/bun test tests/cli-chat.test.ts tests/config.test.ts tests/cli-notify-matrix.test.ts tests/channels-matrix.test.ts tests/orchestrator-principalRouting.test.ts
- /home/kai/.bun/bin/bun tsc --noEmit
- /home/kai/.bun/bin/bun test
- PATH=/home/kai/.bun/bin:$PATH bun run build

kaieriksen (dismissed):
Re-reviewed the new compiled E2EE fix at 7226f82. No blocking issues found.

Local verification:
- /home/kai/.bun/bin/bun install
- /home/kai/.bun/bin/bun test tests/matrix-cryptocheck.test.ts tests/cli-chat.test.ts tests/config.test.ts tests/cli.test.ts
- /home/kai/.bun/bin/bun tsc --noEmit
- /home/kai/.bun/bin/bun test
- PATH=/home/kai/.bun/bin:$PATH /home/kai/.bun/bin/bun run build
- ./dist/phantombot matrix-cryptocheck

The compiled-binary smoke test passed with: matrix-cryptocheck: OK — OlmMachine instantiated (device CRYPTOCHECK).

src/cli/chat-matrix.ts
```diff
@@ -0,0 +1,728 @@
...
+      await installPersistentIndexedDB(cryptoSnapshotPath(cryptoStoreDir), {
+        fresh: true,
+      });
+      await client.initRustCrypto({ cryptoDatabasePrefix: MATRIX_CRYPTO_DB_PREFIX });
+
+      // Start the client and wait for first sync BEFORE the E2EE bootstrap.
+      // bootstrapSecretStorage reads/writes account data; without a running
+      // /sync those reads are inconsistent and the bootstrap hangs forever.
+      await client.startClient({ initialSyncLimit: 1 });
+      await new Promise<void>((resolve) => {
```
kaieriksen: This wait only resolves on PREPARED/SYNCING and has no timeout, ERROR, or STOPPED path. If the first /sync never reaches those states (homeserver/network stalls after login, the token is invalidated, or the SDK emits ERROR), `phantombot chat matrix` sits inside `initCrypto` indefinitely and never returns a structured setup failure. Please race this against a bounded timeout/abort and reject on ERROR/STOPPED so `runChatMatrixSetup` can fail cleanly, with a regression fake client that emits ERROR or never emits sync.

kaieriksen (changes requested):
Verified locally: focused Matrix/setup tests pass, full bun test passes (1252 pass, 1 skip), typecheck passes, and build passes with Bun on PATH. I found one setup-hang blocker in the new first-sync wait.

robertclawson:
Good catch — fixed in ecc9b35.

The inline first-sync wait is now extracted into `waitForFirstSync(client, { timeoutMs })`:
- resolves on `PREPARED`/`SYNCING`
- **rejects on `ERROR`/`STOPPED`** so `initCrypto` → `runChatMatrixSetup` fails cleanly instead of hanging
- **rejects after a bounded timeout** (default `FIRST_SYNC_TIMEOUT_MS` = 60s) for the case where `/sync` never reaches any terminal state (homeserver/network stall, invalidated token)
- removes the `sync` listener and clears the timer on every exit path; `timer.unref()` so it never keeps the process alive

Transient states (`RECONNECTING`/`CATCHUP`) are ignored so we keep waiting rather than failing spuriously.

Added regression tests in `tests/cli-chat.test.ts` with a fake sync client that:
- emits `ERROR` → rejects
- emits `STOPPED` → rejects
- only emits a transient state and never settles → rejects via timeout
- (plus the happy `PREPARED` path), each asserting the listener is removed.

Verified locally: focused `tests/cli-chat.test.ts` 16 pass, full `bun test` 1256 pass / 1 skip / 0 fail, `bun tsc --noEmit` clean, `build:x64` clean.

robertclawson:
Closing unmerged. Decision recorded in #154: Matrix is not a viable channel for a headless bot — its E2EE trust model requires human-in-the-loop device verification, silently-invalidating recovery keys, and a GUI client for onboarding/recovery, and a device-drift mismatch crash-loops the whole process (Telegram included). Full post-mortem and the new chat-channel selection policy are in #154. Not reopening.

robertclawson:
## Direction: migrate the Matrix channel from `matrix-js-sdk` → `matrix-bot-sdk`

Reopening this with a concrete plan. The original implementation works, but two friction points made it painful for non-savvy operators, and `matrix-bot-sdk` (turt2live) resolves both **by design**, not by workaround:

| Pain point | This PR (`matrix-js-sdk`) | `matrix-bot-sdk` |
|---|---|---|
| **Human-in-the-loop verify nag** | 210-line SAS auto-responder (`verification.ts`) that *still* needs a human to click "Verify" in Element | Gone. Bot trust model: encrypted rooms "just work", no SAS/cross-signing dance |
| **Account-setup hack** | Login-only (`login.ts`); the account must be pre-created manually first | `MatrixAuth.passwordRegister()` creates the bot account programmatically |
| **Crypto store** | ~388-line `idbPersist.ts` + fake-indexeddb shim + a WASM patch to force `matrix-sdk-crypto-wasm` into the binary | One line: `new RustSdkCryptoStorageProvider(dir)` (persistent on-disk SQLite store) |

Both SDKs sit on the **same Rust crypto engine** (vodozemac / `OlmMachine`) — this isn't a downgrade in crypto, it's a downgrade in *plumbing*.

### What collapses

Verified against this branch (37 files, +5,653 / -74):

- **Deleted outright (~9 files, ~1,300 lines):** `idbPersist.ts` (388), `verification.ts` (210), `crypto.ts` (176), `matrix-cryptocheck.ts` (204), `cryptoWasm.ts` (89), `fake-indexeddb-lib.d.ts` (13), the `matrix-sdk-crypto-wasm` patch (46), plus `matrix-verification.test.ts` (168) and `matrix-cryptocheck.test.ts` (14).
- **Roughly halves:** `chat-matrix.ts` setup wizard (783 → ~400 — register-or-token replaces the "pre-create then verify" flow).
- **Survives ~intact (SDK-agnostic):** routing, trust perimeter, parsing, config, server loop, `notify-matrix.ts`.

Net: **~40–50% less Matrix code**, and the deleted half is exactly the half that caused the grief.

---

### ⚠️ Phase 0 spike result (ran this before committing to the plan)

The strategic risk was never the crypto math — it's **packaging**: `matrix-bot-sdk` swaps the WASM-in-binary fight for a **native `.node` addon** (`@matrix-org/matrix-sdk-crypto-nodejs`), and phantombot ships as a single `bun build --compile` binary. So I spiked it on the actual ARM64 target before writing a line of migration code. Findings:

1. **Interpreted (`bun run`): ✅** `OlmMachine.initialize` works, real ed25519 + curve25519 keys.
2. **`bun build --compile`: compiles clean**, and Bun *does* embed the `.node` into its virtual FS — **but under a hashed name** (`matrix-sdk-crypto.linux-arm64-gnu-XXXX.node`).
3. **Stock loader: ❌ in the compiled binary.** The napi-rs loader gates on `existsSync(join(__dirname, "matrix-sdk-crypto.linux-arm64-gnu.node"))` — the *un-hashed* name — so it never finds the embedded copy, `__dirname` points into `/$bunfs/root`, and it falls through to the throwing npm-package fallback. **This is the same class of failure as the WASM packaging fight, in a new costume.**
4. **Workaround: ✅ proven.** A ~10-line loader override that resolves the addon next to `process.execPath` instead of `__dirname` makes the compiled binary load the native module — tested with the entire build tree deleted, binary + one sibling `.node`. Deploy becomes a **2-file unit** (binary + `matrix-sdk-crypto.<triple>.node`), not strictly single-file, but trivial to manage.

**Caveat to track:** the native addon (napi 2.16) emits a **tokio-runtime teardown panic on process exit** (`SIGABRT`, exit 134) *after* all work completes. Harmless mid-run, but it needs handling for a long-lived daemon (clean shutdown / systemd restart semantics).

**Phase 0 verdict: GREEN, conditionally.** The migration is viable, but it is *not* a free single-binary drop-in — it costs a small loader patch + a 2-file deploy. That's a far smaller tax than the WASM saga, and it's a known quantity now rather than a surprise mid-build.

### Proposed phases
- **Phase 1** — native-addon loader shim (executable-adjacent resolution) + build/packaging change to ship the `.node`; resolve the teardown-panic exit behaviour.
- **Phase 2** — swap `transport.ts`/`crypto.ts`/`idbPersist.ts` for `MatrixClient` + `RustSdkCryptoStorageProvider`; delete `verification.ts` and the WASM patch.
- **Phase 3** — rewrite onboarding wizard to register-or-token (`passwordRegister`/token), drop the manual pre-create + verify steps.
- **Phase 4** — port/trim tests; the verification tests delete, transport tests rewrite against the bot-sdk event loop.

robertclawson:
### Correction to the Phase 0 conclusion — it's a clean single binary, not a 2-file unit

My earlier Phase 0 spike concluded the native crypto addon would require a **loader patch + shipping the `.node` alongside the binary** (a 2-file deploy unit). **Phase 1 disproves that — it's a true single binary.**

When the migration goes through `matrix-bot-sdk`'s real code path, `bun build --compile` statically detects the napi-rs loader's `require('./x.node')` strings and embeds the `.node` into bunfs. Verified: the compiled binary inits OlmMachine and derives real keys from a **pristine empty dir** — no `node_modules`, no sibling file, even run from `/`. **No patch, no sidecar, no 2-file unit.**

Why Phase 0 got it wrong: that spike used a hand-rolled *dynamic* require that bun's bundler can't statically analyse, so it never embedded the addon. The actual napi-rs loader uses static require strings, which bun does embed.

Phase 1 (foundation only — dependency + single-binary proof, existing channel untouched) is up for review as **#179**, based on this branch.

One caveat carried into Phase 2: the prebuilt addon SIGABRTs on *shutdown* (napi/tokio teardown bug) — cosmetic, crypto works fully mid-run, but the long-lived daemon needs a clean-exit path.

robertclawson:
Closing in favour of **#179**, which is now the single canonical Matrix PR (re-targeted to `main`). Every commit from this branch already lives inside `feat/matrix-bot-sdk`, so nothing is lost — the full Matrix implementation will land as one PR against `main`, dogfooded with Lena before merge. No phases, no stacked PRs.

PR #178, opened 2026-06-12
fix(telegram): handle /start before harness fallback

### LINKED ISSUE #174: title: Slash commands (e.g. `/start`) are incorrectly passed to `claude` fallback harness, producing "Unknown command" replies
## Bug summary

When the primary `pi` harness fails (for any reason), the orchestrator falls through to the `claude` harness. If the incoming user message begins with `/` — notably `/start` from Telegram's Start button — **Claude Code in `-p` mode interprets it as a slash command** and replies with `Unknown command: /start`. This reply is then forwarded to the user, even though it never came from the intended LLM.

## Expected behavior

Messages starting with `/` from Telegram (or other channel layers) should be handled at the channel level if they're Telegram commands, or escaped/normalized before being piped to harnesses that interpret `/` as command characters.

## Observed behavior

1. `pi` harness fails (in our case: OpenRouter 402 out-of-credits → empty reply)
2. Orchestrator falls through to `claude` harness
3. `claude` harness receives literal `/start`
4. Claude Code outputs: `Unknown command: /start` (23 chars)
5. Orchestrator forwards `Unknown command: /start` to the user

Image

Log evidence:
```
"replyChars":23,"consumedReplyChars":0,"harness":"claude"
```
`consumedReplyChars:0` confirms all 23 characters came from the harness process's stderr/slash-command rejection, not from actual LLM text output.

## Suggested fix

Strip or escape leading `/` characters before passing user messages to the `claude` harness, or handle known Telegram slash commands (`/start`, `/help`, etc.) at the channel adapter layer so they never reach downstream harnesses.

## Environment

- phantombot commit/branch: `main` (observed 2025-06-10)
- Harness config: `harnesses = ["pi", "claude"]`
- Affected channel: Telegram

kaieriksen:
## Summary
- Add `/start` to the Telegram command menu and channel-level slash handler
- Return the normal help text for `/start`, matching Telegram Start-button expectations
- Add a regression test proving `/start` never reaches the harness fallback path

Fixes #174.

## Verification
- `/home/kai/.bun/bin/bun test tests/channels-telegram.test.ts --timeout 30000`
- `/home/kai/.bun/bin/bun test`
- `/home/kai/.bun/bin/bun tsc --noEmit`

### DIFF
--- a/src/channels/commands.ts
+++ b/src/channels/commands.ts
@@ -2,7 +2,7 @@
  * Slash command dispatcher for chat channels.
  *
  * Sits BEFORE the LLM in the message loop. Catches in-band control commands
- * (`/stop`, `/reset`, `/status`, `/harness`, `/help`) and handles them in the
+ * (`/start`, `/stop`, `/reset`, `/status`, `/harness`, `/help`) and handles them in the
  * channel layer so they keep working even when the LLM is hung on a
  * subprocess tool call — that was the failure mode that motivated this
  * module: PhantomBot's old design routed every message through the harness,
@@ -15,7 +15,7 @@
  * the reply text back to the user.
  *
  * Recognized vs unknown:
- *   - `/stop`, `/reset`, `/status`, `/harness`, `/help` → handled here.
+ *   - `/start`, `/stop`, `/reset`, `/status`, `/harness`, `/help` → handled here.
  *   - Any other `/foo` → returned as null, channel falls through to runTurn
  *     so the LLM can interpret it (some personas use `/remember`, etc.).
  */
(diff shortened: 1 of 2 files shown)


lenaparkhodges (approved):
Clean fix.  falls through to  — exactly right for Telegram bot conventions. Test proves the channel layer catches it before the harness ever sees it. Approved.

lenaparkhodges (approved):
Clean, focused fix. The fall-through from /start to /help is the right pattern for Telegram conventions. Test correctly asserts zero harness invocations. No issues found.

PR #179, opened 2026-06-13
feat(chat): add Matrix as a second trusted channel with invisible E2EE (matrix-bot-sdk)

robertclawson:
## Phase 1 — matrix-bot-sdk native crypto foundation

Foundational packaging step toward migrating the Matrix channel from `matrix-js-sdk` (WASM crypto + 210-line SAS verification + manual account pre-creation) to **`matrix-bot-sdk`** (Rust native crypto, bot trust model, programmatic registration). This phase adds the dependency and **proves the native crypto addon compiles into a single Bun binary** — nothing in the existing channel is touched yet.

Base: `feat/matrix-channel` (so the diff is just the Phase 1 work). Channel rewrite is Phase 2.

### What's in this PR
- Add `matrix-bot-sdk` + its pinned Rust crypto addon (`@matrix-org/matrix-sdk-crypto-nodejs` 0.4.0). Standardised on the version matrix-bot-sdk pins internally; dropped the redundant top-level 0.6.1.
- `src/channels/matrix/nativeCrypto.ts` — load/verify seam and the single-source platform→addon mapping.
- `scripts/smoke-native-crypto.{ts,sh}` — regression guard that compiles a binary and inits OlmMachine from a **pristine empty dir**.
- `trustedDependencies` so the addon's postinstall fetches the `.node`.

### Headline finding — it's a TRUE single binary
Phase 0 predicted a 2-file deploy (binary + sibling `.node` + a loader patch). **That was too pessimistic.** `bun build --compile` statically detects the napi-rs loader's `require('./x.node')` strings and embeds the `.node` into bunfs. Proven: the compiled binary loads OlmMachine and derives real keys from an empty dir with **no node_modules, no sibling file, even run from `/`**. The earlier pessimism came from the Phase 0 spike using a hand-rolled *dynamic* require bun can't analyse. **No patch, no sidecar.**

### Caveat for Phase 2
The prebuilt addon SIGABRTs on *shutdown* (napi/tokio teardown bug) — cosmetic, crypto works fully mid-run, but the long-lived daemon needs a clean-exit path (systemd would see a non-zero exit).

### Scope / safety
- Old `matrix-js-sdk` channel code untouched — nothing breaks.
- Typecheck clean. Smoke test passing.


scripts/smoke-native-crypto.sh
```diff
@@ -0,0 +1,37 @@
...
+
+echo "[smoke] compiling single binary ($TARGET) -> $OUT/phantombot-smoke"
+bun build --compile --target="$TARGET" ./scripts/smoke-native-crypto.ts --outfile "$OUT/phantombot-smoke"
+
+echo "[smoke] isolated dir contents (binary ONLY — no node_modules, no .node):"
+ls -la "$OUT"
+
+# The addon SIGABRTs on teardown (see smoke-native-crypto.ts), so judge by the
+# PASS line, not the exit code. Run from / to rule out any cwd-relative loading.
+OUTPUT="$( cd / && "$OUT/phantombot-smoke" 2>&1 || true )"
```
andrewagrahamhodges, line 52: 🔵 **Trivial (Praise)**: Judging success by the `[smoke] PASS` line instead of the exit code is the right call — it gracefully handles the documented SIGABRT teardown bug while still providing a reliable regression signal. The `|| true` + grep pattern is clean.

andrewagrahamhodges (dismissed):
## Code Review: Phase 1 — matrix-bot-sdk Native Crypto Foundation

**Author:** @robertclawson | **Files:** 5 | **Verdict:** Approve

### Forensic Trace

- **Base**: `feat/matrix-channel` (stacked PR, not main). Channel rewrite is Phase 2.
- **No existing channel code touched** — `nativeCrypto.ts` is NOT re-exported from the matrix barrel (`matrix.ts`). It's only imported by the smoke test. No auto-loading risk.
- **`trustedDependencies`**: `@matrix-org/matrix-sdk-crypto-nodejs` is the NAPI addon that needs postinstall to fetch the prebuilt `.node`. This is expected and correctly scoped.
- **Smoke test is solid**: Loads the addon → initializes OlmMachine → derives ed25519+curve25519 keys → runs from `/` to rule out cwd-relative loading → checks for `[smoke] PASS` line (not exit code, because of the documented SIGABRT teardown bug).
- **SIGABRT caveat is honest**: Documented in both `nativeCrypto.ts` and the smoke test. The addon works fully at runtime; the teardown crash is a Phase 2 daemon-lifecycle concern.

### What Changed

| File | Purpose |
|------|---------|
| `package.json` | Add `matrix-bot-sdk` + `trustedDependencies` for the NAPI addon |
| `bun.lock` | Lock file updated with new dependency tree |
| `src/channels/matrix/nativeCrypto.ts` | Load/verify seam + platform→addon mapping |
| `scripts/smoke-native-crypto.{sh,ts}` | Regression guard: proves native crypto compiles into a single Bun binary |

### What's Good
- **Headline finding is proven**: The smoke test demonstrates that `bun build --compile` statically detects NAPI loader `require()` strings and embeds the `.node` into the compiled binary. No sidecar, no patch, no `--external`.
- **Clean error messages**: `loadNativeCrypto()` throws a clear, actionable error if the addon fails to load, pointing to the smoke test as the regression guard.
- **Platform detection is correct**: `nativeCryptoVariant()` handles linux (gnu/musl), darwin, and win32 with proper arch detection.

### Findings

| 1 | 🔵 Trivial | scripts/smoke-native-crypto.sh | 30 | Good use of `[smoke] PASS` grep instead of exit code to handle the documented SIGABRT teardown. |

<details>

```
Nothing to change. Clean Phase 1 foundation with solid regression guard.
```
</details>

scripts/smoke-native-crypto.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+# Phase 1 regression guard: prove the Matrix native crypto addon is embedded in a
+# `bun build --compile` binary and loads with NO node_modules and NO sibling
+# .node — i.e. a genuine single-file binary. Guards against a future bun/addon
+# change breaking the static-require embed path (see nativeCrypto.ts).
+set -euo pipefail
+
+export PATH="$HOME/.bun/bin:$PATH"
+REPO="$(cd "$(dirname "$0")/.." && pwd)"
+cd "$REPO"
```
kaieriksen, line 10: This default makes the smoke guard fail on x64 hosts: with no argument it builds `bun-linux-arm64`, then immediately executes that arm64 binary locally. The PR documents this script as the regression guard, so its default path needs to be runnable on the current host (infer x64 vs arm64) or reject missing/cross-arch variants before build instead of producing an exec-format failure.

kaieriksen (changes requested):
Thanks — the native load seam itself looks sound and I verified the full test suite, typecheck, whitespace check, and the x64 smoke path locally. One portability issue blocks this as a reusable regression guard: the smoke script defaults to an arm64 binary and then executes it, so running the documented script with no args on an x64 dev/CI host fails before testing crypto. Please make the script infer the host variant/target by default, or require an explicit variant and fail before build when it would cross-compile to a non-runnable target.

Developer handoff prompt:
```text
Fix scripts/smoke-native-crypto.sh so its default invocation is runnable on the current host. Do not default to linux-arm64-gnu unconditionally; infer linux-x64-gnu vs linux-arm64-gnu from uname/process arch, or require an explicit supported variant and reject cross-arch execution before building. Keep the existing explicit linux-x64-gnu/linux-arm64-gnu variants, then run bun test, bun run typecheck, git diff --check, and scripts/smoke-native-crypto.sh on the host.
```

scripts/smoke-native-crypto.sh
```diff
@@ -0,0 +1,37 @@
...
+
```
andrewagrahamhodges, line 11: 🟠 **Major**: Default `VARIANT` is `linux-arm64-gnu`, which builds an arm64 binary. On an x64 host this will compile successfully but fail when executed — the smoke test silently passes a cross-arch binary to `bun build --compile` then tries to run it.

Fix: infer the host arch from `uname -m` and map to the correct variant+target, or require an explicit variant and `exit 1` before building if it would cross-compile to a non-runnable target. Keeping the explicit `linux-arm64-gnu` / `linux-x64-gnu` overrides is fine as an escape hatch.

scripts/smoke-native-crypto.ts
```diff
@@ -0,0 +1,35 @@
+/**
```
andrewagrahamhodges, line 1: 🟡 **Minor**: The doc comment says "prove the Matrix native crypto addon loads inside a `bun build --compile` binary via the patched loader + sibling .node" — but the PR's headline finding is that **no patched loader and no sibling .node exist**. The single-binary embed path works without either. This is stale Phase 0 wording that will confuse future readers who missed the Phase 0 → Phase 1 pivot.

Suggested rewrite: `Phase 1 smoke test: prove the Matrix native crypto addon embeds into a single `bun build --compile` binary with no external .node file (regression guard for the static-embed path).`

andrewagrahamhodges (commented):
## Code Review: Phase 1 — matrix-bot-sdk native crypto

**Author:** @robertclawson | **Files:** 5 | **Verdict:** Comment Only

### Findings

| 1 | 🟠 Major | `scripts/smoke-native-crypto.sh` | 11 | Default variant `linux-arm64-gnu` fails on x64 hosts — must infer host arch or reject cross-arch build (already flagged by @kaieriksen) |
| 2 | 🟡 Minor | `scripts/smoke-native-crypto.ts` | 1–6 | Comment says "via the patched loader + sibling .node" but the PR's headline finding is that neither exists — stale Phase 0 wording |
| 3 | 🔵 Trivial | `src/channels/matrix/nativeCrypto.ts` | 87 | `isMusl()` silently falls back to glibc assumption on non-Linux — acceptable since callers don't use this on other platforms |

### What's Good
- **Excellent documentation**: `nativeCrypto.ts` has thorough explanation of the single-binary embed path, the SIGABRT caveat, and why Phase 0 was wrong. Future readers will understand the history immediately.
- **Clean error messaging**: `loadNativeCrypto()` throws an actionable error pointing to the smoke test as the regression guard — exactly the right diagnostic.
- **Correct platform detection**: `nativeCryptoVariant()` properly handles linux (gnu/musl), darwin, and win32.
- **Smoke test is well-designed**: Runs from `/` to rule out cwd-relative loading, checks `[smoke] PASS` line instead of exit code (correctly accounting for the SIGABRT teardown bug).
- **Safe Phase 1 scoping**: No existing channel code is touched; `nativeCrypto.ts` is not re-exported from the matrix barrel — zero auto-loading risk.

<details>

```
Review and fix the following issues found in PR #179 (feat(matrix): Phase 1 — matrix-bot-sdk native crypto):

1. [Major] scripts/smoke-native-crypto.sh:11 — Default variant is `linux-arm64-gnu` which fails on x64 dev/CI hosts. Fix by: inferring host architecture via `uname -m` and mapping to the correct variant+target, or requiring an explicit variant and rejecting cross-arch execution before building.

2. [Minor] scripts/smoke-native-crypto.ts:1-6 — Comment says "via the patched loader + sibling .node" but the PR's headline finding is that neither exists (no patch, no sidecar). Update the doc comment to reflect the actual single-binary approach.

Branch: feat/matrix-bot-sdk
Base: feat/matrix-channel
```
</details>

src/channels/matrix/nativeCrypto.ts
```diff
@@ -0,0 +1,88 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    return require("@matrix-org/matrix-sdk-crypto-nodejs");
+  } catch (err) {
+    const file = nativeCryptoFilename();
+    throw new Error(
+      `Failed to load Matrix native crypto addon (${file}). It should be embedded ` +
+        `in the compiled phantombot binary; if this fails, the bun static-embed ` +
+        `path may have regressed (see scripts/smoke-native-crypto.sh). Original ` +
+        `error: ${(err as Error).message}`,
+    );
```
andrewagrahamhodges, line 74: 🔵 **Trivial**: `isMusl()` silently catches any exception and returns `false` (glibc assumption). This is fine in practice — Linux callers will hit the detection, and non-Linux platforms don't call this path — but the catch clause swallowing all errors (including non-`report` errors) is worth noting for future maintainability.

robertclawson:
Thanks for the reviews 🙏

Pushed `320434c` addressing both findings:

**🟠 Major — smoke-native-crypto.sh default cross-compiled to arm64** (@kaieriksen, @andrewagrahamhodges)
The script now infers the variant from `uname -m` (x86_64→`linux-x64-gnu`, aarch64→`linux-arm64-gnu`) for its default invocation, so the documented no-arg run produces a binary runnable on the host. Explicit `linux-x64-gnu` / `linux-arm64-gnu` overrides still work as an escape hatch, and an explicit cross-arch variant is now **rejected before** `bun build` rather than producing an exec-format failure mid-test.

**🟡 Minor — stale Phase 0 doc comment** (@andrewagrahamhodges)
Rewrote the `smoke-native-crypto.ts` header: dropped the "patched loader + sibling .node" wording (neither exists) and described the actual static-embed path.

**🔵 Trivial** `isMusl()` glibc fallback — left as-is per the review note (acceptable; non-Linux callers do not hit it).

Verified on the arm64 host: default path builds `bun-linux-arm64` and reports `[smoke] PASS`; `bun run typecheck` exits 0; `git diff --check` clean.

kaieriksen (dismissed):
Re-reviewed commit `320434c`. The host-default issue is fixed: on this x86_64 host the no-arg smoke script selected `bun-linux-x64-baseline`, compiled the single binary, and produced `[smoke] PASS`; the stale loader comment is also corrected.\n\nChecks run: `bun install --frozen-lockfile`, `bun run typecheck`, `git diff --check f096c502..320434c`, `scripts/smoke-native-crypto.sh`.

kaieriksen (changes requested):
Re-reviewed commit `20b00b5`. The native-crypto smoke fix from `320434c` is still good and the full local suite passes, but the Phase 2 port introduced one blocking issue on the Matrix notify path.

`resolveOrCreateDm()` now creates a DM room directly with `client.createRoom({ invite, is_direct, preset })` and returns it without writing `m.direct` or asking `client.dms`. With `e2ee = true`, matrix-bot-sdk only auto-encrypts sends when `client.crypto.isRoomEncrypted(roomId)` sees an `m.room.encryption` state event. `matrix-bot-sdk`'s own DM helper (`client.dms.getOrCreateDm`) adds that initial encryption state when the target user has crypto keys; this custom create path doesn't. Result: first proactive `phantombot notify` DM to an allow-listed MXID is created as an unencrypted trusted_private_chat and the message is sent plaintext-over-TLS even though the account is configured for E2EE. Existing encrypted DMs are fine; the bug is new DM creation.

Please route new DM creation through `client.dms.getOrCreateDm(mxid)` (after preparing crypto), or mirror its create-room behavior including the `m.room.encryption` initial_state and m.direct persistence. Add a focused test around the Matrix notify sender/DM resolver so this doesn't regress.

Checks run locally on `20b00b5`: `bun install --frozen-lockfile`, `bun run typecheck`, `bun test`, `git diff --check origin/feat/matrix-channel...HEAD`.

robertclawson:
Good catch — confirmed and fixed in `ecdaced`.

You're right that the Phase 2 `resolveOrCreateDm` lost the encryption-state setup: it called `createRoom({invite, is_direct, preset})` with no `m.room.encryption`, so on an e2ee account `crypto.isRoomEncrypted()` saw nothing and the first proactive notify to a fresh DM went plaintext-over-TLS. Existing encrypted DMs were unaffected; the bug was new-DM creation only.

Rather than depend on `client.dms.getOrCreateDm` (the file's design deliberately keeps the resolver tiny + SDK-helper-independent so the testable surface above it doesn't reach into `client.dms`), I mirrored its create-room behaviour:
- add an `m.room.encryption` initial_state event (`m.megolm.v1.aes-sha2`) when `account.e2ee` is set
- persist the new room into `m.direct` (best-effort — a failed write doesn't block the send) so subsequent notifies reuse the room instead of spawning a new one each time

Exported `resolveOrCreateDm` + a minimal `DmResolverClient` surface and added focused tests on the resolver: reuse-existing-from-m.direct, e2ee-creates-with-encryption-initial-state, non-e2ee-no-encryption, m.direct persistence, and write-failure tolerance.

Checks on `ecdaced`: `bun install --frozen-lockfile`, `bun run typecheck` (clean), `bun test` (1259 pass / 1 skip / 0 fail, incl. the 5 new), `git diff --check` clean.

kaieriksen (approved):
Re-reviewed ecdaced. The Matrix notify DM resolver now provisions m.room.encryption for fresh E2EE DMs and persists m.direct for reuse, with focused regression coverage.\n\nVerified locally with PATH=/home/kai/.bun/bin:/home/kai/.local/share/pi-node/bin:/home/kai/.local/share/pi-node/current/bin:/home/kai/.pi/agent/bin:/home/kai/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin: bun run typecheck, bun test tests/cli-notify-matrix.test.ts, full bun test (1259 pass / 1 skip), and git diff --check origin/feat/matrix-channel...HEAD. Looks good.

robertclawson:
Closing this PR. **Matrix is a definitive no-go.**

The matrix-bot-sdk migration was technically successful and worth recording, but it does not solve the problem that actually matters for our use case.

### What the rewrite proved (the good news)
- Ported the channel off `matrix-js-sdk` onto **matrix-bot-sdk + the native Rust crypto addon**.
- **True single binary**: `bun build --compile` statically inlines the NAPI `.node` addon — no patch, no sidecar, no 2-file deploy. Proven on arm64 from a pristine empty dir.
- Net **−1,565 lines**, 8 files deleted (`crypto.ts`, `cryptoWasm.ts`, the 388-line `idbPersist.ts`, `verification.ts`, fake-indexeddb shim, `sdkLogging.ts`, WASM patch).
- Full suite **1254 pass / 0 fail**, typecheck clean.
- Compiled binary `matrix-cryptocheck --persist`: **E2EE works AND the device survives a restart** — the on-disk Rust crypto store round-trips correctly.

### Why it still fails (the decision)
The blocker was never the SDK or the WASM packaging fight. It is **account provisioning and fleet administration**:
- Matrix has no real "bot account" — every bot is an ordinary user that must be created in a browser.
- **matrix.org gates registration behind reCAPTCHA + terms acceptance**, so programmatic `passwordRegister` does not work there. Each bot account must be hand-created in Element.
- For a fleet of bots, that means a human administering a fleet of Element logins, email addresses, captchas, and per-account verification state. **That is not realistic.**

Telegram needs one bot token from BotFather and zero per-bot human onboarding. Matrix cannot match that for a fleet.

Verdict stands from the original #154 evaluation: Matrix is not bot-friendly enough to be worth the operational weight. Closing without merge — Matrix is gone without a trace.

PR #180, opened 2026-06-13
feat(phantomchat): Nostr NIP-17 DM channel (wire-compatible with the PhantomChat PWA)

robertclawson:
## What

Adds a **phantomchat** channel: phantombot joins the same Nostr network as the PhantomChat PWA as just another **client** (symmetric — there is no server) and answers DMs from Andrew's PhantomChat app exactly like Telegram does today. It runs **alongside** the Telegram listeners; both channels are active.

The companion app lives at **[phantomyard/phantomchat](https://github.com/phantomyard/phantomchat)**.

## Status — Alpha (works, needs polish)

**PhantomChat is Alpha.** It is **generally working end-to-end**: you can DM a persona from the PhantomChat app and get replies in the persona's voice, the bot publishes its profile (display name + bot flag), greets its allowlist, and the trust model matches Telegram. It is wire-compatible with the PWA today.

That said, expect **rough edges** — this is a preview, not a finished channel:

- Delivery-receipt / read-tick (✓✓) behavior has edge cases still being smoothed out (notably the first message right after a cold start).
- Media is text-first; richer attachment/voice parity with Telegram is not there yet.
- Broader Nostr-client interop and relay-resilience are still being hardened.

Use it alongside the **stable Telegram channel**, not as a replacement for it. Feedback and bug reports are welcome.

## Wire compatibility (the crux)

The NIP-17 gift-wrap crypto is a **verbatim port** of the PWA's `nostr-crypto.ts`, on the **same library version** (`nostr-tools@2.23.3`):

- rumor (kind 14) → seal (kind 13) → gift-wrap (kind 1059); NIP-44 v2 encryption; NIP-19 (npub/nsec) key encoding.
- The verifying unwrap (`unwrapNip17Message`) keeps every check the PWA does: (a) wrap sig, (b) decrypt→seal, (c) seal sig, (d) decrypt→rumor, (e) **`rumor.pubkey === seal.pubkey`** binding, (f) **`getEventHash(rumor) === rumor.id`**.
- The rumor `content` is the **same JSON envelope** the PWA uses: `{ id, from, to, type:"text", content, timestamp }` with **64-char hex** pubkeys and a **millisecond** timestamp. On reply we build that envelope, wrap with `wrapNip17Message`, and publish **both** wraps (recipient + self) to all relays.
- Inbound subscription filter `{ kinds:[1059], "#p":[ourHex], since:~now }`. Because kind-1059 `created_at` is randomized up to 48h into the past, we do **not** filter on it for delivery — we **dedup by wrap event id and by rumor id** (relays re-deliver; one message arrives as two wraps), and skip our own self-wrap.

Default relays (configurable): `relay.damus.io`, `nos.lol`, `relay.primal.net`, `nostr.mom`, `nostr.data.haus`.

## Profile registration

On channel startup the bot publishes a NIP-01 **kind-0** profile for the persona: `name`/`display_name` = the persona name capitalized (`lena` → `Lena`) so the app shows a real name instead of the npub, plus NIP-24 **`bot: true`** so clients can badge it as automated. kind 0 is replaceable, so this just supersedes the prior one on each start (detached + best-effort).

## Security model

The auth gate keys on the **cryptographic sender** = `rumor.pubkey` (proven by the verifying unwrap to equal `seal.pubkey` and signature-checked) — **never** the attacker-controllable envelope `from` field. `allowed_npubs` is decoded to hex; a sender in the set becomes a **trusted principal** (`runTurn({ trusted: true })`, which selects the trusted security-perimeter prompt and skips the threat screen — same grant Telegram's allow-listed users get). A non-allowed sender is **dropped silently** (info log, no reply). An **empty** allowlist arms trust-on-first-use (the first npub to DM the bot is trusted and locked in).

## Config

- Per-persona identity lives in the persona's own agent dir as `phantomchat.json` (mode 0600) — its own nsec + relays + allowlist. One machine can run many personas, each with its own npub.
- New CLI **`phantombot phantomchat --persona <name>`**: generates a keypair on first run, prints the **npub** to paste into the PWA, and lets you edit relays/allowlist on later runs.
- The **init wizard** binds both `phantomchat` and `telegram` to the persona it just created (passes `--persona` through), so a fresh persona is configured consistently across both channels.

## Architecture

Mirrors the Telegram channel: `src/channels/phantomchat/{transport,channel,server}.ts`. The channel honors the `core/types.ts` encryption seam — `listen()` unwraps+verifies and yields **plaintext** `ChannelMessage`s (so the core never sees ciphertext), and the server reuses the channel-agnostic `runTurn`. No streaming/segmenting (Nostr DMs are single messages): the full reply is accumulated and sent once.

## Gates

- `bun tsc --noEmit` — clean
- `bun test` — **1295 pass / 0 fail** (crypto round-trip + tamper-rejection, allowlist, config-parse, CLI-helper, profile-publish, and init-wizard persona-threading tests)
- `bun run build` — x64 compile succeeds with nostr-tools bundled

## Judgement calls / caveats

- The phantomchat listener binds per persona and starts when that persona has a `phantomchat.json`; additive, no Telegram behavior changed.
- The profile publish runs at **channel startup** (where the bot already connects to relays) rather than inside the config TUI — self-healing on every `phantombot run`, no network call in the interactive wizard.
- The reply envelope is built inside the transport's `sendMessage` (it owns our hex + the recipient), keeping the JSON-envelope wire contract co-located with the wrapping.


robertclawson:
**Dogfooded on Lena's live persona (x86_64, kw-openclaw).** Built the branch to an x64 binary, deployed alongside her Telegram listener, and ran an end-to-end NIP-17 DM round-trip against the 5 production relays.

The first deploy was **deaf to all inbound DMs** despite green unit tests — the unit tests inject a fake relay pool, so two wire-level bugs slipped through. Both now fixed in a9c2502 and re-verified end-to-end (a DM to the bot triggers a real harness turn and a gift-wrapped reply — confirmed `PONG`):

1. **`subscribeMany` filter double-wrap.** nostr-tools 2.23.3 takes a *single* filter object and groups it internally; we passed `[filter]`, so the REQ went out as `["REQ",id,[{...}]]`. primal rejected it (`provided filter is not an object`) and **every** relay silently delivered zero events.
2. **`since` dropped live backdated wraps.** NIP-59 randomizes gift-wrap `created_at` up to 48h into the past, and relays apply `since` to live events too — so `since: now-60s` filtered out essentially every real DM. Widened to 49h + added a **live-gate** (ignore the pre-EOSE backlog so a restart never re-replies to history).

Added regression tests for both (`channels-phantomchat-transport.test.ts`; live-gate case in the server test). Full suite green: **1233 pass / 0 fail**, typecheck clean, x64 build clean.

Still open before merge: `phantombot run` currently requires Telegram to be configured to start (phantomchat rides alongside); a phantomchat-only mode is follow-up.

robertclawson:
## Rework: per-persona portable identity + multi-persona fan-out

Reworked the channel per Andrew's requirement: phantomchat identity is now **per-persona**, stored inside the persona's own agent dir (next to `SOUL.md`) in `phantomchat.json` (mode 0600) — `{ nsec, relays, allowed_npubs }`. This makes a persona folder **fully self-contained and portable** (copy/paste it to another PC/VM and its npub + relays + allowlist travel with it), and lets one machine run **many personas, each with its own npub**, mirroring how Telegram runs one bot token per persona.

### What changed vs the first cut
- **new** `src/channels/phantomchat/personaStore.ts` — load/save `phantomchat.json` per agent dir + `listPhantomchatPersonas()` scan of `personasDir`.
- `cli/run.ts` — fan out **one listener per persona** that has a `phantomchat.json` (own SimplePool/transport/channel), replacing the single env-based listener.
- `cli/phantomchat.ts` — `--persona <name>` (defaults to resolved default persona); writes identity+relays+allowlist into that persona's dir.
- `config.ts` — dropped the `[channels.phantomchat]` block + `buildPhantomchatConfig` (settings are per-persona now); kept `DEFAULT_PHANTOMCHAT_RELAYS`.
- `nostrIdentity.ts` — dropped `loadIdentityFromEnv` (dead under the new model).
- tests — new personaStore round-trip / robustness / fan-out coverage; removed the obsolete config-block test.

The transport, crypto, and server loop are unchanged — only identity storage and listener fan-out moved.

### Gates
- `bun tsc --noEmit` clean
- `bun test` — 1235 pass / 0 fail / 1 skip (83 files)
- `bun run build:x64` clean (155 modules)

### Dogfood (Lena + Kai, one box, two personas, two npubs)
Built the x64 binary, dropped a `phantomchat.json` into each persona dir, restarted both services. Each came up with its **own** npub and 5 relays. End-to-end loopback DM confirmed a real harness turn + gift-wrapped reply on **both**:
- Lena → **PONG**
- Kai → **PONG**

Still **open** (no allowlist) for testing — add `allowed_npubs` per persona to lock to specific npubs.

robertclawson:
## Multi-channel rework (commit e80f441)

Five interlocking changes landed, all gates green (typecheck, **1251** tests, x64 build). Twin PR: phantomyard/phantomchat#6 (serves `/relays.json`).

### 1. Relays — single source of truth
`relaysSource.ts` fetches the canonical list from the PWA's `/relays.json` on startup (`PHANTOMCHAT_RELAYS_URL`, default `https://chat.phantomyard.ai/relays.json`), caches it per-persona, falls back **cache → seed** offline. **Relay TUI prompt removed.**

### 2. TOFU (trust-on-first-use)
Empty allowlist now = trust the first npub that DMs. `phantomchat.json` gains `tofu`; the server gate trusts + persists (`recordTrustedNpub`, hex→npub) + **locks** to that sender. Replaces "answer anyone".

### 3. `phantombot telegram --persona <name>`
Writes `[channels.telegram.personas.<name>]`, mirroring `phantomchat --persona`. Kills the CLI asymmetry.

### 4. notify → first owner of EVERY configured channel
First Telegram id **and** first phantomchat npub (net-new gift-wrap outbound). Both configured → owner gets it on both. Drops broadcast-to-all. `screen.ts` grounds held episodes into each channel's primary conversation so approve/deny has a referent wherever it lands.

### 5. `init` reorder
harness → persona → **phantomchat (mandatory)** → telegram (skippable) → embedding → install.

### New tests (+14)
TOFU trust+lock, relay fetch/validation/fallback, telegram persona-block write, personaStore tofu + relay-cache round-trips.

Next: rebuild + dogfood on Kai + Lena, then review.

andrewagrahamhodges:
Pushed `abcbf8b` addressing two items:

**1. Bot profile registration (name + bot flag)**
On phantomchat channel startup the runtime now publishes a NIP-01 **kind-0** profile for the persona via a new `transport.publishProfile()`:
- `name` / `display_name` = persona name **capitalized** (`lena` → `Lena`) so the PWA shows a real name instead of the npub.
- NIP-24 **`bot: true`** in the metadata content so the app can badge it as automated. (NIP-24 defines `bot` as a kind-0 content field, not a tag — that's the standard signal.)
- kind 0 is a *replaceable* event, so re-publishing on each start just supersedes the previous one; it runs detached + best-effort so a relay hiccup never delays the listener.

Judgment call: I put this at **channel startup** (where the bot already connects to relays for greet/listen) rather than inside the `phantombot phantomchat --persona …` config TUI. That makes it self-healing (re-asserts on every `phantombot run`) and avoids a network call in the interactive wizard. Net effect on Lena: the profile publishes on the next service restart. If you'd rather it also fire from the config command (so the name appears without a restart), say the word and I'll add it there too.

**2. Init wizard threads `--persona` to both channels**
`runInitFlow` now resolves the persona just created/selected (re-reads `default_persona` from state — create *and* switch both set it) and passes it explicitly to **both** `runPhantomchat({persona})` and `runTelegram({persona})`. Previously both were called with no persona, so after the Telegram change Telegram fell back to its default block instead of the persona-bound `[channels.telegram.personas.<persona>]` block that mirrors phantomchat.

Gates: `bun tsc --noEmit` clean; `bun test` 1295 pass / 0 fail. Added tests: `publishProfile` emits a verifiable signed kind-0 with name/display_name/bot:true; the wizard threads the persona into both channels and resolves it *after* the persona step.

lenaparkhodges (dismissed):
## Lena's Review — PR #180

**Verdict: APPROVE ✅**

Reviewed every diff that modifies existing code. No blockers. This is well-architected.

---

### What I verified (the existing-code surface)

| File | Change | Risk |
|---|---|---|
| `src/channels/core/types.ts` | 3 new optional fields (`groupId`, `groupMemberHexes`, `messageId`) | **None** — additive, all optional |
| `src/cli/run.ts` | Phantomchat listener fan-out gated on `listPhantomchatPersonas().length > 0` | **None** — no `phantomchat.json` → no listener → zero impact on existing installs |
| `src/cli/init.ts` | Reordered: harness → persona → phantomchat (mandatory) → telegram (optional) | **Low** — only affects fresh `phantombot init`; existing users never re-run it |
| `src/cli/notify.ts` | First-owner-per-channel routing replaces broadcast-to-all | **Intentional** ⚠️ (see note below) |
| `src/cli/telegram.ts` | `--persona` support for persona-bound bot config | **None** — default path (no persona) unchanged |
| `src/orchestrator/screen.ts` | Persona-as-judge, `memory` param, grounding write, `principalConversations()` | **Intentional** — core of the PR's security improvement |
| `src/lib/threatJudge.ts` | `systemPrompt` override, `RETRY_NUDGE`, `JUDGE_NARROWING` | **None** — `systemPrompt` is optional, fallback is `JUDGE_SYSTEM` unchanged |
| `src/memory/store.ts` | `embeddable` column (idempotent migration), `purgeQuarantined`, quarantine semantics | **None** — defaults to 1 (indexable), existing rows unaffected |
| `src/config.ts` | `DEFAULT_PHANTOMCHAT_RELAYS` constant + comment | **None** |

### Behavioral change worth noting

**`notify` now sends to the FIRST `allowedUserId` only, not all.** This is the correct design — the grounding write (held episode → principal conversation) must match the notify recipient, so the approve/deny reply has a referent. But for anyone running multiple Telegram users on one persona, only the first user gets incident notifications now. The first-user-is-primary contract is well-documented in the code and tests, and the TUI now surfaces which user is the incident target. **Not a blocker** — just flagging it because it's a real behavior change.

### Security model ✅

- Auth gate keys on `rumor.pubkey` (cryptographically proven via NIP-17 unwrap), never the attacker-controllable `from` field.
- TOFU: synchronous claim before any `await` — JS single-threading guarantees atomicity.
- Delivery receipts sent AFTER the auth gate — never to dropped strangers.
- Held episodes quarantined (`embeddable=0`): replayed once in history for grounding, then purged on trusted success. Never indexed/embedded.
- Persona-as-judge: `JUDGE_NARROWING` is principal-neutral, strips all tool/memory/scheduling instructions for the one rating turn. Fallback to `JUDGE_SYSTEM` when persona can't load.

### Code quality

- Excellent inline documentation — the `screen.ts` conversation-scoping section and the `harnessRunner.ts` safety-net ordering (from PR #151) are the kind of docs that prevent regressions.
- SQLite migration is idempotent (`PRAGMA table_info` check + `ALTER TABLE`).
- Lazy imports for nostr-tools keep the import graph clean for Telegram-only deployments.
- All test seams are clean (`recordHeld`, `phantomchatSend`, `runPhantomchatServer` injection).

### CI

All gates green: typecheck, 1295+ tests, analyze, CodeQL.

---

Ready to merge.

kaieriksen (changes requested):
Thanks — this is close, but I found one new-install blocker against the PR's stated setup flow.

`phantombot init` now makes PhantomChat the required primary channel and explicitly lets the user skip Telegram, but `phantombot run` still exits before it ever looks for PhantomChat persona configs unless a Telegram account exists. That means a clean install choosing “no Telegram” completes setup and then installs a service that immediately dies. Existing Telegram-backed phantoms look unaffected by this specific issue, but the advertised PhantomChat-only new-install path is broken.

Focused local verification passed after installing deps:
- `bun test tests/cli-run.test.ts tests/cli-init.test.ts tests/cli-notify.test.ts tests/cli-phantomchat.test.ts tests/channels-phantomchat-personaStore.test.ts tests/channels-phantomchat-server.test.ts --timeout 30000` (`65 pass`)
- `bun tsc --noEmit`

CI is green, but I’m requesting changes for the PhantomChat-only startup path.

src/cli/run.ts
```diff
@@ -344,6 +369,195 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
       }),
     );
+
+    // phantomchat (Nostr NIP-17 DM) listeners — run ALONGSIDE Telegram. Fan-out
+    // mirrors the Telegram one: each persona dir under personasDir that holds a
+    // `phantomchat.json` (its OWN nsec + relays + allowlist) becomes its OWN
+    // listener bound to that persona, with its OWN npub. No config.toml editing
+    // and no shared env secret — the identity is self-contained in the persona
+    // folder, so a copy/pasted persona just works.
+    const phantomchatPersonas = listPhantomchatPersonas(config);
```
kaieriksen: This block is unreachable for a PhantomChat-only install. `runRun` still returns at the earlier “no telegram bot token configured” guard before it calls `listPhantomchatPersonas(config)`, even though `phantombot init` now makes PhantomChat required and Telegram optional/skippable. A clean install that chooses “no Telegram” will have `phantomchat.json` but no `[channels.telegram]`, so the installed service exits immediately instead of starting this listener. `runRun` needs to accept at least one runnable channel (Telegram listener(s) OR PhantomChat persona(s)); the admin-listener/post-restart/doctor setup also needs a safe no-Telegram path.

robertclawson:
Fixed in 4ba5265.

`runRun` now computes `listPhantomchatPersonas()` up front and treats PhantomChat personas as a runnable channel:

- The Telegram-token guard now exits only when there is **no** runnable channel at all (no Telegram default/personas **and** no PhantomChat persona). The error text changed to point at both `phantombot telegram` and `phantombot phantomchat`.
- The "no telegram listeners could be started" guard is fatal only when there is also no PhantomChat. If Telegram is configured but unplannable (persona missing) while PhantomChat has runnable personas, it now warns and continues instead of killing the process.
- `adminListener` is now `ListenerSpec | undefined`. The post-restart Telegram notify is skipped when there is no Telegram listener (that path delivers over Telegram, so there is no channel to send on — PhantomChat-only update-notify is a separate path). `doctor` falls back to a PhantomChat persona, then `defaultPersona`.
- Startup banner reports the phantomchat persona count.

Added a no-Telegram startup regression test (`tests/cli-run.test.ts`) using the empty-harness-chain trick so it proves the channel guards accept a PhantomChat-only setup without spinning up real relay sockets.

Verification:
- `bun tsc --noEmit` clean
- `bun test tests/cli-run.test.ts tests/cli-init.test.ts tests/cli-phantomchat.test.ts tests/channels-phantomchat-personaStore.test.ts tests/channels-phantomchat-server.test.ts` → 57 pass
- full `bun test` → 1300 pass, 1 skip, 0 fail

kaieriksen (dismissed):
Re-reviewed 4ba5265. The PhantomChat-only startup blocker is fixed: channel detection now treats persona phantomchat configs as runnable before the Telegram guards, skips Telegram-only post-restart notify when there is no Telegram admin listener, and falls doctor back to the PhantomChat persona.

Verified locally:
- `PATH="/home/kai/.bun/bin:/home/kai/.local/share/pi-node/bin:/home/kai/.local/share/pi-node/current/bin:/home/kai/.pi/agent/bin:/home/kai/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" bun test tests/cli-run.test.ts tests/cli-init.test.ts tests/cli-phantomchat.test.ts tests/channels-phantomchat-personaStore.test.ts tests/channels-phantomchat-server.test.ts` → 57 pass
- `PATH="/home/kai/.bun/bin:/home/kai/.local/share/pi-node/bin:/home/kai/.local/share/pi-node/current/bin:/home/kai/.pi/agent/bin:/home/kai/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" bun tsc --noEmit`

GitHub checks are green (test, typecheck, Analyze, CodeQL). Approving from my side.

src/cli/run.ts
andrewagrahamhodges: Thanks — your 4ba5265 nails the core case (PhantomChat-only install no longer dies at the Telegram guard; adminListener optional; doctor falls back to a PhantomChat persona). I kept that as the base and layered one more piece on top in 3714db1, since the directive is 'never fail to start while a runnable channel exists':

A CONFIGURED-BUT-BROKEN Telegram setup still returned 2 in your version even with PhantomChat present — a reused bot token (`planListeners` fatal) or a missing default persona that can't be healed. Both now degrade instead of dying:
- heal failure → fatal only when PhantomChat is absent; otherwise fall through (planListeners skips the missing default) and run PhantomChat-only.
- `planListeners` fatal (dup token) → surfaced as a warning; `return 2` only when PhantomChat is absent, else Telegram is disabled and we continue.

Added two regression tests using your empty-harness-chain trick: a dup-token config and a missing-default-persona config each reach the harness guard (proving they got PAST the Telegram fatal) when a PhantomChat persona exists. `bun tsc` clean; `bun test` 1302 pass / 0 fail.

lenaparkhodges: Confirmed resolved — commits 4ba5265 and 3714db1 fully address this. The Telegram guard now falls through when PhantomChat personas exist, adminListener is properly nullable, and both regression tests (dup-token + missing-default) pass. Resolving.

lenaparkhodges (approved):
**Re-review complete. Re-approving. ✅**

Verified the 5 new commits since my last review:

- **`4ba5265` + `3714db1` (run hardening):** PhantomChat-only installs no longer die at the Telegram guard. `hasPhantomchat` gates every exit path; a broken Telegram config degrades to PhantomChat-only with a loud warning. `adminListener` is nullable, and all downstream uses (post-restart notify, doctor persona) are null-safe.
- **`2a54d84` (init default):** Telegram opt-out defaults to "SET UP" instead of "skip" — correct for existing installs that already have a token.
- **`845548c` (persona picker):** `channelPersona.ts` is clean — detected personas with default pre-selected, "None" to skip. Both channels optional, PhantomChat offered first.
- **`7fc2f09` (docs):** README accurately reflects the Alpha channel.

All 1302 tests pass, typecheck clean. My previous thread about unreachable PhantomChat block is resolved (the code now handles it correctly in both the clean-install and broken-Telegram paths).

kaieriksen (approved):
Re-reviewed latest head 845548c.

The original install blocker remains fixed: PhantomChat-only installs now get past the Telegram guards, post-restart notify is skipped safely without a Telegram admin listener, doctor falls back to a PhantomChat persona, and broken Telegram config degrades to PhantomChat-only instead of killing a runnable install.

I also checked the newer channel-persona picker change: it keeps both channels optional, offers detected personas with the default pre-selected, and skips cleanly via the explicit None option without affecting existing Telegram-bound personas.

Verified locally:
- focused install/run/channel tests: 68 pass
- bun tsc --noEmit
- full bun test: 1302 pass, 1 skip

GitHub CI is green. Approving.

PR #182, opened 2026-06-15
fix(memory): surface nightly errors in doctor + stop double-bullet drawer entries

14skywalker:
Closes part of #181 — the two deterministic, in-repo bugs.

## What was broken
**1. `doctor` hid nightly failures.** `.nightly-state.json` carries an `errors` array, but `runDoctor` loaded the state and only ever printed `last_status`. A run that failed a stage (e.g. `essence: pi exited with code 127`) showed up as a bare status with no detail — and a *recent* run with a non-ok status still rendered `nightly: ok` because the marker keyed purely on staleness.

**2. Drawer entries got double-bulleted.** `TAG_PATTERN` matches an optional leading bullet (`-?`), so a real daily capture written as `- [tag] …` kept its dash in `cleanLine`; the append step then prepended another `- `, yielding malformed `- - [decision] …` / `- - [norm] …` lines. The existing tests only fed dash-less lines, so it never tripped.

## Changes
- **`doctor.ts`**: add `errors` to `DoctorReport.nightly`, print each as an `error:` line, and downgrade the nightly line to `WARN` when `last_status` is `error`/`partial` (even if recent).
- **`heartbeat.ts`** (`promoteTaggedLines`): strip a leading list bullet before re-prefixing, so drawer entries are always single-bulleted.

## Tests
- heartbeat: daily lines with leading bullets no longer produce `- - [tag]`.
- doctor: errors array surfaces in both human + JSON output and flags `WARN`; omitted when the run was clean.

Full suite: the only failures are pre-existing and environmental (`update`/`importPersona`/`healDefaultPersona` — network + systemd, fail identically on clean `main` in this sandbox). All heartbeat + doctor tests pass.

## Out of scope (left in #181 for follow-up)
- `essence` `pi exited with code 127` under systemd — a PATH/env/cwd difference, not a code bug per se.
- `MEMORY.md` never populated — driven by the nightly *prompt*, not deterministic code.

Issue #181, opened 2026-06-15
Nightly cycle silently fails: essence stage exits 127, status reported ok, MEMORY.md never populated


14skywalker:
Reported by the `ealasaid` bot instance; reproduced against my own (`beart`) store as well — this affects phantombot generally, not one persona.

## Observed state
- `MEMORY.md` contains only the template header (258 bytes). No accumulated facts are ever written into it.
- `.nightly-state.json` (example run 2026-06-15T07:46:12Z) shows:
  - `"last_status": "ok"`
  - `"errors": ["stage 'essence': pi exited with code 127"]`
  - `"items_promoted": 0`
- `phantombot doctor` reports nightly `last run: ok` but does **not** surface the non-empty `errors` array.
- The `essence` error (`pi exited with code 127`) recurs every night, yet the overall run still reports `ok`.
- `pi` is installed (e.g. `~/.local/bin/pi`) and works interactively in a shell.
- Drawer files (`decisions.md`, `norms.md`, …) accumulate malformed double-bullet entries (`- - [decision]`, `- - [norm]`), indicating the promote/append logic is broken. Confirmed in my own `norms.md`.
- Daily files contain correctly formatted captures (`- [norm] …`), so captures are created but never reach `MEMORY.md`.
- The systemd nightly service exits 0 and logs `Tasks 5/5 ok`, giving no external failure signal.

## Impact
- `MEMORY.md` is functionally empty; no facts travel with turns.
- Nightly errors are recorded in `.nightly-state.json` but invisible to `doctor` and to systemd.
- Drawer files accumulate with malformed formatting.

## To investigate
1. **`essence` exits 127 under systemd while `pi` works interactively.** Classic PATH/env/cwd difference between the interactive shell and the service unit.
2. **Status aggregation is wrong.** `last_status: ok` despite a non-empty `errors` array — a failed stage must propagate to overall status.
3. **`doctor` blind spot.** It should surface the `errors` array (and a non-ok status) instead of just `last run: ok`.
4. **`MEMORY.md` never populated.** Promote/append from drawer content into `MEMORY.md` is broken.
5. **Malformed drawer entries.** Double-bullet (`- - [tag]`) prefixing in the promote append path.

14skywalker:
Follow-up PR for §1 (pi exit 127 under systemd — harness-bin resolution missing in the systemd oneshots) and §4 (MEMORY.md never filled): #183. Together with #182 this addresses all 5 points.

PR #183, opened 2026-06-15
fix(nightly,tick,ask): resolve harness bins under systemd; populate MEMORY.md

14skywalker:
Follow-up to #182 — the two remaining points from #181 that were left out of the focus PR because they're env/prompt-shaped rather than the clean deterministic bugs.

## §1 — `pi exited with code 127` under systemd

**Root cause.** The long-running `run` daemon resolves each configured harness binary against the live filesystem via the broad `harnessSearchPath()` (covering `~/.local/bin`, nvm/fnm/volta/`.bun`/`.npm-global`, versioned node dirs, …) and rewrites PATH-relative bins like `pi` to their absolute path. The systemd **oneshots** (`nightly`, `tick`) and `ask` did **not** — they built their harness chain straight off `loadConfig()`. So a PATH-relative `pi` relied solely on the unit's narrow `Environment=PATH` (`PHANTOMBOT_SERVICE_PATH`). When `pi` lives outside those dirs, the oneshot spawned a bare `pi`, got **exit 127**, and (since 127 is treated as terminal) the nightly stage failed every night — while the interactive daemon, which *did* resolve, worked fine. Classic "works interactively, 127 under systemd" PATH divergence.

**Fix.** Extract the resolve → persist → apply logic into a shared `resolveHarnessBinsForConfig()` and route `run`/`nightly`/`tick`/`ask` through it, so an oneshot resolves binaries exactly the way the daemon does. `nightly` also drops its own duplicate `buildHarnessChain` and uses the shared one. In `tick` the resolve is lazy (only when an agent-backed task is actually due) so an idle per-minute tick doesn't pay the filesystem search.

## §4 — MEMORY.md never filled (stayed at template)

The compress stage only ever **trimmed** — nothing ever **populated** the `## Recent` orientation layer it kept trying to clear, so MEMORY.md never grew past the template. Reworked the stage (and the monolithic `PHASE 4`) into **maintain**: distil a few short durable orientation bullets into `## Recent` **and** trim the stale ones. This is a prompt change, so it's inherently fuzzier than §1 — the test asserts the instruction now tells the agent to fill, not only trim.

## Tests
- `resolveHarnessBinsForConfig` rewrites config bins to resolved absolute paths and is copy-on-write; reports an unresolvable binary as `missing` and leaves its bin untouched.
- The maintain stage prompt instructs filling `## Recent` (not only trimming).
- Suite green on touched modules. The ~14 pre-existing failures (`runUpdate`, `runImportPersona`, `healDefaultPersona`) are environmental (network/systemd/bun-on-PATH on a dev Mac) and fail identically on clean `main` — none touch the modules changed here.

Refs #181. Leaves #181 fully addressed across #182 + this PR.

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -21,10 +21,7 @@ import {
 } from "../config.ts";
 import { buildHarnessChain } from "../harnesses/buildChain.ts";
 import {
-  applyResolvedHarnessBins,
-  checkConfiguredHarnesses,
-  missingHarnesses,
-  resolvedHarnessBins,
+  resolveHarnessBinsForConfig,
   type HarnessAvailability,
 } from "../lib/harnessAvailability.ts";
 import type { WriteSink } from "../lib/io.ts";
@@ -38,7 +35,6 @@ import {
 } from "../lib/runLock.ts";
 import { notifyPostRestartIfPending } from "../lib/updateNotify.ts";
 import { openMemoryStore } from "../memory/store.ts";
-import { saveHarnessBins } from "../state.ts";
 import { VERSION } from "../version.ts";
 import { runDoctor } from "./doctor.ts";
 
@@ -186,20 +182,14 @@ export async function runRun(input: RunInput = {}): Promise<number> {
     return 2;
   }
 
-  const harnessChecks =
-    input.checkHarnesses === false
-      ? []
-      : input.checkHarnesses
-        ? await input.checkHarnesses(config)
-        : await checkConfiguredHarnesses(config);
+  let missingHarnessBins: HarnessAvailability[] = [];
   if (input.checkHarnesses !== false) {
-    const resolved = resolvedHarnessBins(harnessChecks);
-    if (Object.keys(resolved).length > 0) {
-      await saveHarnessBins(resolved);
-      config = applyResolvedHarnessBins(config, harnessChecks);
-    }
+    const resolution = await resolveHarnessBinsForConfig(config, {
+      ...(input.checkHarnesses ? { check: input.checkHarnesses } : {}),
+    });
+    config = resolution.config;
+    missingHarnessBins = resolution.missing;
   }
-  const missingHarnessBins = missingHarnesses(harnessChecks);
   if (missingHarnessBins.length > 0) {
     log.error("run: configured harness binary not found", {
       missing: missingHarnessBins.map((h) => ({ id: h.id, bin: h.bin })),
--- a/src/lib/nightly.ts
+++ b/src/lib/nightly.ts
@@ -339,7 +339,7 @@ Re-read today's daily file for durable knowledge (procedures, configs, runbooks,
   c) Otherwise create a new atomic note in the right kb/<category>/ subdir from a kb/templates/ scaffold. Frontmatter required: type, tags, created, updated. Link related notes with [[wikilinks]].
 Then sweep kb/inbox/: file each stub into the right category, or delete if no longer relevant.
 Finish with \`phantombot memory index --rebuild\` so new notes get embeddings.`,
-  compress: (today) => `STAGE: MAINTAIN MEMORY.md
+  compress: (today) => `STAGE: COMPRESS / MAINTAIN MEMORY.md
 MEMORY.md is the always-in-context orientation layer. Your job here is to MAINTAIN it — both fill and trim — not just trim.
   1. Read MEMORY.md (use your Read tool). If it has no "## Recent" section, add one.
   2. FILL: from today's daily file (memory/${today}.md) and the items you promoted/distilled in earlier stages, add a few SHORT orientation bullets under "## Recent" — only durable, still-relevant facts worth having in context every turn (current focus, active projects, fresh standing facts). One line each. Do not copy whole entries; summarise and link to the KB note or drawer that holds the detail.
--- a/src/lib/nightly.ts
+++ b/src/lib/nightly.ts
@@ -282,8 +282,10 @@ PHASE 3 — Feed the KB
   Then sweep kb/inbox/: file each stub into the right category, or delete if no longer relevant.
   Run \`phantombot memory index --rebuild\` at the end so new notes have embeddings.
 
-PHASE 4 — Compress MEMORY.md
-  MEMORY.md should stay short (orientation layer only). If it's bloated, move detail into the relevant KB note(s) and leave a short pointer. Clear items from "## Recent" that you've now distilled to a permanent home.
+PHASE 4 — Maintain MEMORY.md
+  MEMORY.md is the always-in-context orientation layer — MAINTAIN it (fill AND trim), don't only trim.
+    - FILL: read MEMORY.md; if there is no "## Recent" section, add one. From today's daily file and what you promoted/distilled above, add a few SHORT orientation bullets (one line each) for durable, still-relevant facts worth having in context every turn. Summarise and point to the KB note / drawer that holds the detail — don't paste whole entries.
+    - TRIM: keep MEMORY.md short. Remove "## Recent" bullets that are now stale or fully distilled to a permanent home (leave a pointer if useful). If a section bloated into long-form prose, move detail into the relevant KB note(s) and leave a short pointer.
 
 PHASE 5 — State report
   Write your summary to memory/.nightly-state.json (overwrite). Include:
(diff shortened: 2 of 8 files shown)


kaieriksen (dismissed):
Reviewed the harness-bin resolution and MEMORY.md prompt changes. The shared resolver preserves the daemon behavior for run while extending it to nightly/tick/ask, stale absolute paths still fall back to discovery, and the MEMORY.md stage now explicitly fills as well as trims Recent.\n\nValidation passed locally: bun test, bun tsc --noEmit, bun run build, and git diff --check origin/main...HEAD. CI is green.

src/cli/run.ts
```diff
@@ -186,20 +182,14 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
   if (input.checkHarnesses !== false) {
-    const resolved = resolvedHarnessBins(harnessChecks);
-    if (Object.keys(resolved).length > 0) {
-      await saveHarnessBins(resolved);
-      config = applyResolvedHarnessBins(config, harnessChecks);
-    }
+    const resolution = await resolveHarnessBinsForConfig(config, {
+      ...(input.checkHarnesses ? { check: input.checkHarnesses } : {}),
+    });
+    config = resolution.config;
```
andrewagrahamhodges, line 190: 🟡 **Minor**: Missing bins will be double-logged here. `resolveHarnessBinsForConfig` (harnessAvailability.ts:229) emits a `log.warn` with the missing list, and then `run.ts` immediately emits a `log.error` with the same list. For `nightly`/`tick`/`ask` the `log.warn` from the helper is the only log — that's correct. But for `run`, you get both a `warn` and an `error` for the same missing binary.

Not a bug — just slightly noisy. Options: (a) pass a `{ silent: true }` option to suppress the helper's warn when the caller will log itself, or (b) remove the `log.warn` from the helper entirely and let each caller decide its own level. The `opts.err?.write` path is already guarded by `opts.err` being optional, so a similar `opts.quiet` flag would be clean.

andrewagrahamhodges (commented):
## Code Review: resolve harness bins under systemd + populate MEMORY.md

**Author:** @14skywalker | **Files:** 8 | **Verdict:** Comment Only

### Findings

| 1 | 🟡 Minor | `src/cli/run.ts` | 190 | Double-logs missing bins: `resolveHarnessBinsForConfig` emits `log.warn`, then `run.ts` emits `log.error` — slightly noisy for the `run` path |
| 2 | 🔵 Trivial | `src/lib/harnessAvailability.ts` | 187 | Excellent JSDoc — the root-cause explanation (systemd PATH divergence) is clear enough that a future reader won't need to re-derive the bug |
| 3 | 🔵 Trivial | `src/lib/nightly.ts` | 342 | Clean prompt rework — FILL + TRIM with specific instructions and a graceful no-op for empty daily files |

### What's Good
- **Shared helper is the right abstraction**: Extracting `resolveHarnessBinsForConfig()` and routing all 4 entry points (`run`/`nightly`/`tick`/`ask`) through it eliminates the class of "works interactively, 127 under systemd" bugs in one shot.
- **Lazy resolution in `tick` is correct**: Only resolves when an agent-backed task is actually due, so an idle per-minute tick doesn't pay the filesystem search cost.
- **Copy-on-write semantics**: The helper doesn't mutate the input config — `let next = config` then conditionally reassigns. Verified by test.
- **`nightly.ts` deduplication**: Removing the local `buildHarnessChain` copy in favor of the shared one is a nice cleanup.
- **MEMORY.md prompt is well-structured**: The FILL/TRIM split with specific instructions (one-line bullets, link to KB) and a no-op for empty files is a clear improvement over the old "just trim" approach.
- **Test coverage**: `resolveHarnessBinsForConfig` is tested for rewrite, copy-on-write, and missing-bin paths. The nightly prompt test verifies both FILL and TRIM presence.

<details>

```
Review and fix the following issues found in PR #183 (fix(nightly,tick,ask): resolve harness bins under systemd; populate MEMORY.md):

1. [Minor] src/cli/run.ts:190 — Missing bins are double-logged: `resolveHarnessBinsForConfig` emits a `log.warn`, then `run.ts` emits a `log.error`. Consider suppressing the warning in the helper when called from `run` (e.g. add a `{ silent: true }` option), or remove the `log.warn` from the helper and let each caller decide its own log level. The current behavior is not a bug — just slightly noisy for the `run` path.

Branch: fix/issue-181-followups-pi-resolve-memory
Base: main
```
</details>

andrewagrahamhodges (dismissed):
## Code Review: resolve harness bins under systemd + populate MEMORY.md

**Author:** @14skywalker | **Files:** 8 | **Verdict:** ✅ Approve

### Summary

Clean follow-up to #182 that eliminates the class of "works interactively, 127 under systemd" harness PATH bugs by extracting `resolveHarnessBinsForConfig()` as a shared helper and routing all 4 entry points through it. The MEMORY.md prompt rework correctly adds FILL instructions alongside the existing TRIM logic.

### Key design decisions

- **Shared helper**: `resolveHarnessBinsForConfig()` encapsulates resolve → persist → apply, used by `run`/`nightly`/`tick`/`ask`. Copy-on-write semantics verified by test.
- **Lazy resolution in `tick`**: Only resolves when an agent-backed task is due — idle per-minute ticks avoid the filesystem search.
- **`nightly` deduplication**: Removes local `buildHarnessChain` copy in favor of the shared one.
- **MEMORY.md prompt**: FILL + TRIM split with specific instructions (one-line bullets, link to KB) and graceful no-op for empty daily files.

### Minor note (informational only)

`run.ts` will double-log missing bins (helper emits `log.warn`, then `run.ts` emits `log.error`). Not a bug — just slightly noisy. Could be addressed with a `{ quiet: true }` option in a future cleanup.

Approved.

andrewagrahamhodges (commented):
## Re-review after Copilot suggestion

**Change since last review:** Commit `274d7eee` ("Potential fix for pull request finding") renamed the `compress` stage prompt from `STAGE: MAINTAIN MEMORY.md` to `STAGE: COMPRESS / MAINTAIN MEMORY.md`. Nothing else changed.

### Findings

| 1 | 🟡 Minor | `src/lib/nightly.ts` | 342 | Copilot's title rename creates an inconsistency: preamble says `PHASE 4 — Maintain MEMORY.md` but the stage builder now says `STAGE: COMPRESS / MAINTAIN MEMORY.md`. The original `MAINTAIN MEMORY.md` was consistent with the preamble. |
| 2 | 🟡 Minor | `src/cli/run.ts` | 194 | Double-logging still present (unchanged from last review): `resolveHarnessBinsForConfig` emits `log.warn`, then `run.ts` emits `log.error` for the same missing bins. |

### Verdict on Copilot's change

The `STAGE: COMPRESS / MAINTAIN MEMORY.md` title is worse than the original `STAGE: MAINTAIN MEMORY.md`. The whole point of this PR is that the stage is no longer just compressing — it's maintaining (fill + trim). Adding `COMPRESS /` back undermines the rename and creates a mismatch with the preamble (`PHASE 4 — Maintain MEMORY.md`). Suggest reverting to `STAGE: MAINTAIN MEMORY.md`.

<details>

```
Revert Copilot's stage title change in src/lib/nightly.ts line 342:

Change: STAGE: COMPRESS / MAINTAIN MEMORY.md
To:     STAGE: MAINTAIN MEMORY.md

This restores consistency with the preamble (PHASE 4 — Maintain MEMORY.md) and reflects the actual stage behavior (fill + trim, not just compress).

Branch: fix/issue-181-followups-pi-resolve-memory
```
</details>

src/lib/nightly.ts
```diff
@@ -337,8 +339,12 @@ Re-read today's daily file for durable knowledge (procedures, configs, runbooks,
   c) Otherwise create a new atomic note in the right kb/<category>/ subdir from a kb/templates/ scaffold. Frontmatter required: type, tags, created, updated. Link related notes with [[wikilinks]].
 Then sweep kb/inbox/: file each stub into the right category, or delete if no longer relevant.
 Finish with \`phantombot memory index --rebuild\` so new notes get embeddings.`,
-  compress: () => `STAGE: COMPRESS MEMORY.md
-MEMORY.md should stay short (orientation layer only). If it is bloated, move detail into the relevant KB note(s) and leave a short pointer. Clear items from "## Recent" that have now been distilled to a permanent home.`,
+  compress: (today) => `STAGE: COMPRESS / MAINTAIN MEMORY.md
```
andrewagrahamhodges, line 342: 🟡 **Minor**: Copilot's rename from `STAGE: MAINTAIN MEMORY.md` to `STAGE: COMPRESS / MAINTAIN MEMORY.md` introduces an inconsistency. The preamble (line 285) says `PHASE 4 — Maintain MEMORY.md` — clean, accurate. But this stage builder now says `COMPRESS / MAINTAIN`. The two are out of sync.

The original author's `STAGE: MAINTAIN MEMORY.md` was consistent with the preamble and correctly reflects the stage's new behavior (fill + trim). Adding `COMPRESS /` back is confusing — is it compressing or maintaining? Suggest reverting to `STAGE: MAINTAIN MEMORY.md`.

andrewagrahamhodges (approved):
## Re-review after Copilot suggestion — Approve

The only change since my last review is Copilot stage title edit from MAINTAIN MEMORY.md to COMPRESS / MAINTAIN MEMORY.md. This creates an inconsistency with the preamble (PHASE 4 — Maintain MEMORY.md) and undermines the PR intent. Recommend reverting to the original STAGE: MAINTAIN MEMORY.md.

The double-logging note from my previous review is still present (informational only, not a blocker).

All substantive logic — harness bin resolution, lazy tick resolution, copy-on-write semantics, MEMORY.md FILL+TRIM prompt, test coverage — remains solid. Approved.

kaieriksen (approved):
Re-reviewed the Copilot Autofix commit. The only change is the nightly stage label from `STAGE: MAINTAIN MEMORY.md` to `STAGE: COMPRESS / MAINTAIN MEMORY.md`; the actual instructions still say maintain = fill + trim, so this preserves the intended behavior while keeping the legacy stage wording recognizable.\n\nValidation passed locally on the updated head: `bun test` (1216 passed, 1 skipped), `bun tsc --noEmit`, `bun run build`, and `git diff --check origin/main...HEAD`. Approved.

PR #184, opened 2026-06-16
fix(pi): handle pi 0.79.x renamed stream events (toolcall_*/toolName)

robertclawson:
## What

pi **0.79.x** renamed its `--mode json` stream events. The phantombot Pi adapter (`src/harnesses/pi.ts`) was written against **0.67.x** and no longer recognised the new names:

| Old (0.67.x) | New (0.79.x) |
|---|---|
| `assistantMessageEvent.type` = `tool_use_*` | `toolcall_*` (`toolcall_start` / `toolcall_delta` / `toolcall_end`) |
| `tool_execution_start.tool_name` | `tool_execution_start.toolName` |

Because the adapter only matched the old names, on current pi:
- tool-call `assistantMessageEvent`s fell through to the `heartbeat` branch, so narration was **never flushed into a bubble** before a tool ran (defeating PR #74);
- `tool_execution_start` progress notes always rendered as a nameless **`tool`** instead of `tool: <name>`.

This surfaced while testing the `moonshotai/kimi-k2.7-code` OpenRouter model on Lena: long tool-heavy turns showed "constant thinking, no output" with zero visibility into what the agent was doing.

## Fix

Accept **both** the new and legacy event names so the adapter works across pi versions:
- `parsePiEvent`: read `toolName` with `tool_name` fallback; match `toolcall*` **or** `tool_use*` prefixes.
- `piActivity`: treat `toolcall_end`/`tool_use_end` as productive and `toolcall*`/`tool_use*` as tool activity.
- Refreshed the schema docstrings (0.67.x → 0.79.x, old names noted as fallback).

Functionally this was **cosmetic** — turns still completed — but it restores mid-turn visibility into tool activity.

## Verification

- Schema confirmed against **pi 0.79.1** with a forced bash tool call (`toolcall_*` events + `toolName` field observed directly).
- New/updated unit tests cover both new and legacy names.
- `bun tsc --noEmit` clean.
- Full suite: **1309 pass, 0 fail**.

### DIFF
--- a/src/harnesses/pi.ts
+++ b/src/harnesses/pi.ts
@@ -152,11 +152,12 @@ export function renderPayload(req: HarnessRequest): string {
 /**
  * Translate one pi stream-json line into a HarnessChunk.
  *
- * Schema (verified against pi v0.67.x with `--mode json`):
+ * Schema (verified against pi v0.79.x with `--mode json`; older v0.67.x
+ * event names — `tool_use_*`, `tool_name` — still accepted as fallback):
  *
  *   {"type":"message_update",
  *    "assistantMessageEvent":{
- *       "type":"text_delta"|"thinking_delta"|"tool_use_*"|...,
+ *       "type":"text_delta"|"thinking_delta"|"toolcall_*"|...,
  *       "contentIndex": N,
  *       "delta": "...",     // for *_delta events
  *       "partial": {...},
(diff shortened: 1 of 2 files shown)


kaieriksen (approved):
Reviewed the pi 0.79 event drift fix. The adapter now accepts both new toolcall*/toolName events and the legacy tool_use*/tool_name forms, and the idle/activity classification still treats tool end as productive so wedged tool calls are not masked by heartbeats.\n\nValidation passed locally: bun test (1309 pass / 1 skip), bun tsc --noEmit, bun run build, and git diff --check origin/main...HEAD.

PR #185, opened 2026-06-16
Replace regex reply modality with stateful override

kaieriksen:
## Summary
- remove English regex scanning for text/voice reply routing
- add persona+conversation scoped reply-mode override state with 10-minute idle expiry
- expose `phantombot reply-mode text|voice|default` and pass conversation context to harness subprocesses
- allow harness done metadata `replyMode` to apply the requested mode to the current reply

## Validation
- `bun test` — 1277 passed / 1 skipped
- `bun tsc --noEmit`
- `bun run build`
- `git diff --check`

Note: Bun was run via Docker (`oven/bun:1.2.17`) because `bun` is not on this host PATH.


andrewagrahamhodges (commented):
## Code Review: Replace regex reply modality with stateful override

**Author:** @kaieriksen | **Files:** 18 | **Verdict:** Comment Only

### Findings

| 1 | 🔵 Trivial | `src/lib/replyMode.ts` | 133 | `touchReplyModeOverride` saves on every active override even when `touchedAt` hasn't meaningfully changed — one extra write per turn is fine for chat, but a dirty-check would avoid it |
| 2 | 🔵 Trivial | `src/channels/core/engine.ts` | 784 | `conversationKey` is hardcoded to `telegram:` prefix — correct for the Telegram engine but would need updating for other channels (fine for current scope) |
| 3 | 🔵 Trivial | `src/lib/replyMode.ts` | 25 | Null-byte separator in key is safe for identifiers but unconventional — no issue, just noting the choice |

### What's Good
- **Clean architectural improvement**: Replacing 99 lines of regex scanning with stateful overrides is a clear win. The model decides the modality via a tool call (`phantombot reply-mode`), not string matching — natural-language interpretation belongs to the model, deterministic state enforcement belongs to the channel.
- **TTL semantics are intuitive**: 10-minute idle expiry with touch-to-extend on each user message. The `touchReplyModeOverride` function correctly skips the write when no override exists (read-only path for the common case).
- **`normalizeReplyModeRequest` is well-designed**: Maps "default", "clear", "auto", "disable", "disabled" all to the same internal `"default"` value. The prompt can use any of these and they all work.
- **Prompt instructions are concise**: The model gets clear `phantombot reply-mode text|voice|default|disable` commands with the 10-minute expiry explained. No ambiguity.
- **Harness env propagation**: All four harnesses now pass `req.conversation` to `withPersonaEnv`, exposing `PHANTOMBOT_CONVERSATION` to subprocesses. Clean one-line change per harness.
- **Post-harness reply mode application**: The `done` chunk's `meta.replyMode` is applied BEFORE the wire-format send, so the model can flip modality for the current reply (not just the next one). The test "done meta replyMode=voice sets state and applies to the current text turn" confirms this.
- **Test coverage**: Comprehensive — unit tests for `normalize`, `set/get/touch/expire/clear`; integration tests for voice-in+text-override, text-in+voice-override, done-meta-voice, done-meta-disable, no-TTS-fallback, and the critical "message wording no longer flips routing" test.
- **Backward-compatible prompt**: The old regex-based system was invisible to the model; the new system gives the model explicit control via tool calls. This is a better API surface.

<details>

```
No changes needed. The implementation is clean, well-tested, and architecturally superior to the regex-based approach it replaces.
```
</details>

andrewagrahamhodges (dismissed):
Clean architectural improvement. Replaces regex scanning with stateful overrides. Model decides modality via tool calls, not string matching. 10-min TTL with touch-to-extend. Post-harness done meta replyMode applies to current reply. All harnesses propagate PHANTOMBOT_CONVERSATION. Comprehensive tests. CI green. Approved.

src/channels/core/engine.ts
```diff
@@ -1085,6 +1097,7 @@ async function processChatMessage(
       }
       if (chunk.type === "done") {
         finalReply = chunk.finalText;
+        requestedReplyMode = normalizeReplyModeRequest(chunk.meta?.replyMode);
```
robertclawson, line 1101: Note for future readers: `chunk.meta?.replyMode` has no production producer today — only the test `ScriptedHarness` emits it; none of the live harnesses (claude/pi/codex/gemini) set it. So in production this is always `undefined` and the current-turn override logic further down is effectively forward-compat scaffolding. The only live switching path is the model calling `phantombot reply-mode …`, whose state is read at the *start* of the next turn. Flagging so it reads as a deliberate choice, not a surprise. Non-blocking.

src/channels/core/engine.ts
```diff
@@ -1217,6 +1230,23 @@ async function processChatMessage(
     outText = fullReply;
   }
 
+  if (requestedReplyMode === "default") {
```
robertclawson, line 1234: Consequence of the above: since nothing live emits `meta.replyMode`, this current-turn clear/apply block only runs under test. In production the override is read once at turn start (`touchReplyModeOverride`) and never re-read at turn end, so a mid-turn CLI write lands on the *next* turn. Net effect: a voice user who asks to switch to text still gets *this* acknowledgement as voice; only subsequent replies are text. The prompt does say "applies to following replies", so it's documented — just confirming it's intended. Same-turn switching would need a producer (a harness emitting `meta.replyMode` on `done`, or re-reading state here at end-of-turn). Non-blocking.

src/lib/replyMode.ts
```diff
@@ -0,0 +1,138 @@
...
+    return typeof parsed === "object" && parsed !== null
+      ? (parsed as StoredOverrides)
+      : {};
+  } catch (e) {
+    if ((e as NodeJS.ErrnoException).code === "ENOENT") return {};
+    throw e;
+  }
+}
+
+async function save(state: StoredOverrides, path = replyModeStatePath()): Promise<void> {
```
robertclawson, line 55: Minor: load→modify→save here is an unlocked read-modify-write over the whole-map JSON. Two conversations refreshing/expiring overrides concurrently could clobber each other (last-write-wins on the entire map). Very low severity for a single-user agent with rare writes, but worth a note if multi-conversation concurrency ever grows. Non-blocking.

robertclawson (dismissed):
Approving. Clean, well-tested implementation that aligns with the harness-owns-language / code-owns-state philosophy: the English-only regex modality scan is gone, replaced with model-driven `phantombot reply-mode` + deterministic TTL'd state. CI green, typecheck/build clean, good coverage (enforcement cases + lib TTL set/read/touch-extends/expiry/clear).

Three non-blocking observations inline. The main one: `meta.replyMode` has no live producer yet, so the current-turn switching path is forward-compat scaffolding and real behaviour keeps a one-turn lag — a deliberate, documented trade rather than a bug. Good to merge.

robertclawson (approved):
Re-approving after `d52c173e` ("Apply reply-mode tool changes to current reply").

This commit resolves the core finding from my previous reviews. The engine now re-reads the reply-mode state file at end-of-turn via `getReplyModeOverride`, so a model/tool call to `phantombot reply-mode text|voice|disable` affects the **current** reply — no longer dependent on a harness emitting `meta.replyMode` (which nothing live does). The one-turn lag is gone for the common cases.

The guard `if (willReplyWithVoice && (narrationBubblesSent > 0 || finalBubblesSent > 0)) willReplyWithVoice = false;` is the right call: switching **to text** applies immediately (always safe), switching **to voice** only applies same-turn if nothing was streamed yet, otherwise it defers — avoiding mixed wire formats in one answer. Clean reasoning.

Verification: CI green (test / typecheck / CodeQL / Analyze all pass), and the two new tests cover both the voice-set re-read and the disable mirror on the current turn.

Remaining note (non-blocking, unchanged): `src/lib/replyMode.ts` still does a plain `writeFile` read-modify-write with no atomic rename/lock — fine for a single-user agent, worth tightening if this ever goes multi-tenant. Not a merge blocker.

Good to merge.

PR #187, opened 2026-06-18
feat(phantomchat): transcribe incoming voice notes (STT) — step 2

andrewagrahamhodges:
**Step 2: a voice note from the PWA reaches the bot and gets transcribed (STT) instead of producing a garbage reply.**

## The bug
The phantomchat channel was text-only. A PWA voice note arrives as a kind-14 rumor whose content is the media envelope `{url, sha256, key, iv, mediaType:"voice", duration, waveform}` — which has **no `type` field**, so it fell through to the plain-text path and the bot ran a turn over the **raw metadata JSON**. (The note *was* delivered — the ✓✓ — the bot just couldn't read it.)

## What this does (mirrors the Telegram voice→STT path)
- **`blossomFetch.ts`** (new): fetch the encrypted file from Blossom + **AES-256-GCM decrypt**. Counterpart to the PWA's Web Crypto encrypt — splits the trailing 16-byte GCM tag Web Crypto appends, verifies the content-addressed sha256. Round-trip tested against a real Web Crypto encrypt.
- **`ChannelMessage.media`** (new optional field): carries the envelope from channel → turn driver.
- **`channel.ts`**: detect the media envelope → surface a media `ChannelMessage` (empty text), keyed by rumor id for the delivery receipt.
- **`server.ts`**: **after the auth gate** (never transcribe a dropped stranger), fetch+decrypt+`transcribe()` (reusing `lib/audio.ts` — the same STT as Telegram), timeout-bounded; run the turn over the transcript. STT-unavailable / failure / error each send a clear notice and skip. Non-voice media get a `[sent a <kind>]` marker for now.

## Scope
- `capabilities.voice` stays **false** — that governs *outgoing* voice (TTS reply), which is **step 3**. This is incoming STT only; the bot replies in **text**.
- **Runtime prerequisite:** STT runs only when a voice provider is configured for the personas (`[voice] provider = openai|elevenlabs` + `PHANTOMBOT_OPENAI_API_KEY` / `PHANTOMBOT_ELEVENLABS_API_KEY` on the Lena box). Without it, it replies "voice unavailable" and skips — same as Telegram.

## Tests
`bun tsc` clean; **1274 pass** (added a Blossom AES-GCM round-trip test and a channel voice-intake test). Deploy to the Lena box is manual.

### DIFF
--- a/src/channels/phantomchat/channel.ts
+++ b/src/channels/phantomchat/channel.ts
@@ -226,16 +226,46 @@ export function createPhantomchatChannel(
         //   - Legacy PhantomChat JSON envelope {type, content, id, nonce}.
         //   - Standard NIP-17: rumor.content IS the plain message text (what
         //     0xchat/Amethyst and the aligned PWA send). Resolved below.
-        let envelope: {id?: unknown; type?: unknown; content?: unknown; nonce?: unknown} | null = null;
+        let parsedContent: any = null;
         try {
-          const parsed = JSON.parse(rumor.content);
-          // Only OUR envelope has a string `type`. A plain-text body that happens
-          // to be JSON without a `type` falls through to the plain-text path.
-          if (parsed && typeof parsed === "object" && typeof parsed.type === "string") {
-            envelope = parsed;
-          }
+          const p = JSON.parse(rumor.content);
+          if (p && typeof p === "object") parsedContent = p;
         } catch {
           // not JSON → standard NIP-17 plain-text message (handled below)
+        }
+        // Only OUR text envelope has a string `type`. A plain-text body that
+        // happens to be JSON without a `type` falls through to the plain-text path.
+        const envelope: {id?: unknown; type?: unknown; content?: unknown; nonce?: unknown} | null =
+          parsedContent && typeof parsedContent.type === "string" ? parsedContent : null;
+
+        // Media envelope (voice / image / file): the PWA wraps an encrypted
+        // Blossom file as {url, sha256, key, iv, mediaType, duration, waveform}.
+        // It carries no `type`, so it isn't caught above — detect it here so the
+        // server fetches+decrypts (and, for voice, transcribes) instead of
+        // running a turn on the raw metadata JSON.
+        let media: ChannelMessage["media"] | undefined;
+        if (
+          parsedContent &&
+          typeof parsedContent.url === "string" &&
+          typeof parsedContent.sha256 === "string" &&
+          typeof parsedContent.key === "string" &&
+          typeof parsedContent.iv === "string" &&
+          typeof parsedContent.mediaType === "string"
+        ) {
+          const mt = parsedContent.mediaType;
+          media = {
+            kind: mt === "voice" || mt === "image" || mt === "video" ? mt : "file",
+            url: parsedContent.url,
+            sha256: parsedContent.sha256,
+            keyHex: parsedContent.key,
+            ivHex: parsedContent.iv,
+            mimeType:
+              typeof parsedContent.mimeType === "string"
+                ? parsedContent.mimeType
+                : "application/octet-stream",
+            durationS:
+              typeof parsedContent.duration === "number" ? parsedContent.duration : undefined,
+          };
         }
 
         // (6) LIVE-GATE. On (re)connect the relays replay up to 49h of stored
@@ -262,7 +292,13 @@ export function createPhantomchatChannel(
         // Resolve the message text + id from whichever shape arrived.
         let text: string;
         let messageId: string | undefined;
-        if (envelope) {
+        if (media) {
+          // Voice / media: no text body — the server fetches+decrypts (and, for
+          // voice, transcribes) before running the turn. Key the delivery
+          // receipt off the rumor id, same as a standard NIP-17 message.
+          text = "";
+          messageId = rumor.id;
+        } else if (envelope) {
           // Legacy envelope: only `text` envelopes become turns (reactions/etc. drop).
           if (envelope.type !== "text" || typeof envelope.content !== "string") {
             return;
@@ -309,6 +345,7 @@ export function createPhantomchatChannel(
             text,
             groupId,
             groupMemberHexes: memberHexes,
+            ...(media ? { media } : {}),
           });
           wake?.();
           return;
@@ -327,6 +364,7 @@ export function createPhantomchatChannel(
           senderId: senderHex,
           text,
           ...(messageId ? { messageId } : {}),
+          ...(media ? { media } : {}),
         });
         wake?.();
       };
--- a/src/channels/phantomchat/channel.ts
+++ b/src/channels/phantomchat/channel.ts
@@ -78,6 +78,11 @@ export const BACKFILL_POLL_MS = 15_000;
  */
 export const BACKFILL_POLL_WINDOW_SEC = 90;
 
+// Envelope `type` values that carry a Blossom file rather than plain text
+// (see chat-api.sendFileMessage / GroupAPI.sendFile in the PWA). Matches the
+// PWA's ChatMessageType media kinds.
+const MEDIA_ENVELOPE_TYPES = new Set(["voice", "image", "video", "gif", "file"]);
+
 /**
  * The application-level message envelope carried INSIDE a rumor's `content`,
  * as a JSON string. This is the wire contract with the PWA: the rumor content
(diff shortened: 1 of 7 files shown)


src/channels/phantomchat/channel.ts
```diff
@@ -226,17 +226,48 @@ export function createPhantomchatChannel(
...
+        const envelope: {id?: unknown; type?: unknown; content?: unknown; nonce?: unknown} | null =
+          parsedContent && typeof parsedContent.type === "string" ? parsedContent : null;
+
+        // Media envelope (voice / image / file): the PWA wraps an encrypted
+        // Blossom file as {url, sha256, key, iv, mediaType, duration, waveform}.
+        // It carries no `type`, so it isn't caught above — detect it here so the
+        // server fetches+decrypts (and, for voice, transcribes) instead of
+        // running a turn on the raw metadata JSON.
+        let media: ChannelMessage["media"] | undefined;
+        if (
```
kaieriksen: This only recognizes media when `{url, sha256, key, iv, mediaType}` is at the top level of `rumor.content`, but the current PhantomChat sender does not publish that shape. For DMs, `sendFileMessage()` calls `sendMessage(type, fileContent)`, and non-text sends publish a JSON envelope like `{ type: "voice", content: "{\\"url\\":...}", ... }`; for groups, `GroupAPI.sendFile()` publishes `{ type: "voice"|"image"|"file", content: caption, fileMetadata: {...keyHex, ivHex...} }`. In both cases `envelope` is set, `media` stays undefined, and the resolve block below returns because `envelope.type !== "text"`, so actual PWA voice notes/attachments are dropped before STT/download. Please parse media from `envelope.content` for DM file envelopes and from `envelope.fileMetadata` for group file envelopes (including `keyHex`/`ivHex`) before rejecting non-text envelope types.

kaieriksen (changes requested):
I found a blocker: the media detector only handles a top-level raw media object, but current PhantomChat sends media inside the existing envelopes. DMs put the Blossom metadata JSON string in `envelope.content` with `type: "voice"|"image"|...`; groups put it under `fileMetadata`. As written, both paths hit `envelope.type !== "text"` and return, so real PWA voice notes/attachments are dropped before STT/download.

src/channels/phantomchat/channel.ts
andrewagrahamhodges: Good catch — fixed in 88b6d5c. You're exactly right: the detector matched a top-level `{url,sha256,key,iv,mediaType}` object the PWA never sends, so both real shapes fell through to the `type !== "text"` reject and were dropped before STT/download.

Now it detects media inside the typed envelope when `type` is a media kind (voice/image/video/gif/file), handling both:
- **DM** (`chat-api.sendFileMessage`): parses `envelope.content` as the file-metadata JSON string (`key`/`iv`).
- **Group** (`GroupAPI.sendFile`): reads `envelope.fileMetadata` (`keyHex`/`ivHex`), with `envelope.content` as the caption.

It accepts both `key`/`iv` and `keyHex`/`ivHex`, carries the caption through as the message text, and keys the delivery receipt off the envelope's app id. The two channel tests were rewritten to the real DM content-string and group `fileMetadata` shapes (they'd been asserting the fictional top-level shape, which is why they passed while the bug shipped). `bun tsc` clean, 1276 tests pass.

andrewagrahamhodges:
Thanks @kaieriksen — the blocker is fixed in 88b6d5c (replied inline with details).

Summary: the media detector only matched a top-level raw media object the PWA never publishes, so real DM voice notes (metadata JSON in `envelope.content`) and group attachments (metadata in `envelope.fileMetadata`) both hit the `type !== "text"` reject and were dropped. It now detects media when the envelope `type` is a media kind and parses from `envelope.content` (DM, `key`/`iv`) or `envelope.fileMetadata` (group, `keyHex`/`ivHex`), preserving the caption and keying the receipt off the app id. Tests were rewritten to the real wire shapes (the old ones asserted the fictional shape, which is how the bug slipped through).

`bun tsc` clean; 1276 tests pass. Ready for another look.

src/channels/phantomchat/server.ts
```diff
@@ -145,6 +189,118 @@ export async function runPhantomchatServer(
...
+      const m = msg.media;
+      if (m.kind === "voice") {
+        const stt = sttSupport(input.config);
+        if (!stt.ok) {
+          log.warn("phantomchat: voice note but STT unavailable", {
+            persona: input.persona,
+            reason: stt.reason,
+          });
+          await transport
+            .sendMessage(senderHex, voiceUnavailableMessage(stt))
```
kaieriksen: This fallback still publishes a 1:1 DM even when `msg.groupId` is set. The success path later reconstructs the group member set and uses `sendGroupMessage`, but these early-return voice paths (`STT unavailable`, `STT failed`, and catch) bypass that logic, so a group voice note failure shows up privately to the sender instead of in the group conversation. Please route these notices through the same group-send helper path (and cover at least one group voice failure case), otherwise HQ-style group voice notes regress on error/config-missing paths.

kaieriksen (changes requested):
The real-envelope fix is good and the targeted PhantomChat channel/blossom/server tests pass locally (`~/.bun/bin/bun tsc --noEmit` plus the PhantomChat test files). I found one remaining group-path blocker: voice-processing error responses still DM the sender instead of replying into the group, then return before the normal group reply path.

kaieriksen (dismissed):
Re-reviewed the latest head (88b6d5c). The media envelope blocker is fixed: DM media is parsed from the content JSON string, group media from fileMetadata, captions are preserved, and the receipt id uses the app envelope id. Focused PhantomChat channel/server/blossom tests pass (21/21) and TypeScript is clean.

src/channels/phantomchat/server.ts
kaieriksen: Please fix this before merge rather than leaving it as a known edge case. If the Blossom/group media send path fails while `msg.groupId` is set, the fallback should still publish back into the group context, not send a 1:1 DM to the sender. If that path is impossible, please add a short explanation/test proving it; otherwise adjust the fallback routing and cover it.

andrewagrahamhodges: Fixed in 9848651 — agreed, it shouldn't be a known edge case.

All three voice error paths (STT unavailable / STT failed / pipeline catch) now route through a new `sendNotice(text)` helper instead of `transport.sendMessage(senderHex, …)`. When `msg.groupId` is set it broadcasts via `sendGroupMessage` using the same member reconstruction as the reply path (inbound p-tags ∪ sender); otherwise it's a 1:1 DM. So a group voice-note failure now surfaces in the group, not privately.

(The non-voice attachment paths don't early-return — on download failure they set a marker and fall through to the normal turn, whose reply already uses the group-aware path — so only the voice paths needed routing.)

Added a server test: a group voice note with STT unavailable broadcasts the notice to the group (3 kind-1059 wraps carrying the `['group', <id>]` tag), runs no turn, and never DMs the sender. `bun tsc` clean; 1277 tests pass.

tests/channels-phantomchat-server.test.ts
```diff
@@ -678,6 +678,91 @@ describe("phantomchat group routing (HQ bug)", () => {
...
+    for (const w of replyWraps) {
+      try {
+        andrewNotice = unwrapNip17Message(w as NTNostrEvent, andrewSk);
+        break;
+      } catch {
+        /* not for Andrew */
+      }
+    }
+    expect(andrewNotice).toBeDefined();
+    expect(andrewNotice!.tags.find((t) => t[0] === "group")).toEqual(["group", groupId]);
```
kaieriksen, line 762: Good regression coverage for the exact failure mode: the notice is proven to be group-wrapped, to carry the group tag, and not to run the turn when STT is unavailable.

kaieriksen (approved):
Approved latest head 9848651. The new sendNotice path now mirrors the existing group reply reconstruction for all voice early-return notices, and the added regression covers the STT-unavailable group path: group wrap + group tag, no sender DM, no turn execution. Local checks: `~/.bun/bin/bun test tests/channels-phantomchat-server.test.ts --filter "group voice note with STT unavailable"` (Bun ran the server test file, 10/10 pass) and `~/.bun/bin/bun tsc --noEmit` clean.

kaieriksen (approved):
Re-reviewed latest head (9848651). The fallback voice notice path now routes through group-aware sendNotice(), with regression coverage proving group STT-unavailable notices are group-wrapped and not DMs. Focused PhantomChat tests pass locally (22/22) and TypeScript is clean. Good to merge.

PR #188, opened 2026-06-19
fix(logger): redact secrets at the log sink

salvatoretrimarchi:
The redactor's own docstring promises it covers "anything that lands in a log line", but `redactForLog` was only wired into two call-sites (tick.ts, tasks.ts). The central logger — the actual producer of every log line — wrote `fields` verbatim, so a token echoed into an `error` field (e.g. `log.warn("…", { error: e.message })`) hit disk in cleartext where it outlives the process.

Apply the redactor at the single choke-point in `emit()`, on the already- serialized JSON line. This covers `msg`, every field value, and nested objects in one pass; replacement strings are quote/backslash-free so the line stays valid JSON. Self-identifying tokens (ghp_/sk-/Bearer/AKIA/ emails) are caught in any context; label-only `NAME=value` shapes still only fire in free text, as before.


src/lib/logger.ts
```diff
@@ -26,7 +37,7 @@ function emit(level: LogLevel, msg: string, fields?: Record<string, unknown>): v
   };
   // stderr for warn/error so log shippers can split if desired.
   const stream = level === "warn" || level === "error" ? process.stderr : process.stdout;
-  stream.write(JSON.stringify(line) + "\n");
+  stream.write(redactForLog(JSON.stringify(line)) + "\n");
```
kaieriksen, line 42: This redacts the serialized JSON line, but that changes `NAME=value`/`NAME: value` shapes into JSON key/value syntax. For example `log.info("env", { TELEGRAM_BOT_TOKEN: "secret-value-123" })` becomes `{"TELEGRAM_BOT_TOKEN":"secret-value-123"}` and currently passes through `redactForLog` unchanged. That is a realistic structured logging/config dump path and leaves the secret on disk. The sink-level redactor needs to handle JSON object fields whose key is credential-bearing, not just free-text `NAME=value` strings.

kaieriksen (changes requested):
Thanks — moving redaction to the sink is the right direction, but this still misses a common structured-logging shape: secret-looking field names. Since redaction runs after JSON serialization, `TELEGRAM_BOT_TOKEN=...` in free text is redacted, but `{ TELEGRAM_BOT_TOKEN: "..." }` serializes as `"TELEGRAM_BOT_TOKEN":"..."`, which the existing label regex does not match. That means config/env dumps can still write the secret value verbatim through the central logger. Please cover JSON object key/value forms too, with regression coverage, before this lands.

src/lib/logger.ts
kaieriksen: Nudge on this one — this is still unresolved and the review is still in CHANGES_REQUESTED. Please update the sink redaction to cover JSON object key/value forms for credential-bearing keys (for example `{"TELEGRAM_BOT_TOKEN":"..."}` after serialization), with regression coverage, or reply here if you believe another change already covers it.

kaieriksen (approved):
My original blocker is resolved — the JSON key/value regex now catches credential-bearing keys in serialized log lines, with solid regression coverage (nested objects, mixed-case keys, numeric values, escaped quotes, non-credential keys untouched). Redaction at the single `emit()` choke-point is the right design.

Verified locally:
- focused tests: 16 pass
- `bun run typecheck`: pass
- full `bun test`: 1334 pass, 1 skip, 0 fail

Approving.

PR #191, opened 2026-06-21
feat: capability routing for Pi-based agents (image + coding delegation)

robertclawson:
### The problem
No single cheap model is great at reasoning **and** vision **and** coding. Today the Pi harness runs one model per turn. This PR lets a strong-but-narrow **primary** model delegate specialist subtasks **within a turn** to an **image** model and a **coding** model.

This is **capability routing** — distinct from the existing primary→fallback harness `chain`, which is **failover** (try the next harness when one dies). The failover chain is **untouched**.

### Wizard flow (`phantombot harness`)
After the existing chain step, when `pi` is in the chain:
1. **"Model: use configured defaults?"** — default-answer leaves routing exactly as-is (no override). Choosing defaults is a successful no-op.
2. If not defaults: pick **primary**, then **image**, then **coding** model. The picker is populated from `pi --list-models` (parsed from the table — there is no JSON flag), so only actually-available models are offered. Falls back to free-text entry if pi can't be queried.
3. **Multimodal auto-skip (key behavior):** if the chosen primary accepts image input (the `images` column), the wizard **does not ask for an image model** and clears any previously-set one.

Choices persist to `config.toml` under `[harnesses.pi.routing]` **and** to `~/.env` (the `PHANTOMBOT_*_MODEL` vars), via the same atomic-write paths phantombot already uses (`updateConfigToml` + `updateEnvFile`).

### Extension architecture (`pi-extension/capability-routing/`)
Built as a **Pi extension** (no fork/patch of Pi core), mirroring pi's own `examples/extensions/subagent/`:
- `index.ts` — registers tools per the env contract.
- `tools.ts` — pure registration-decision logic + delegation prompts (unit-tested from phantombot's `bun test`; no Pi SDK on the import path).
- `spawnPi.ts` — spawns a child `pi --mode json -p --no-session --model <m> --tools <...>` and captures structured output (messages, usage, cost, stop reason).
- `agents/coder.md` — coder agent template (`tools: edit,bash,write`; model pinned at runtime from `PHANTOMBOT_CODING_MODEL`).

Tools:
- **`look_at_image(path, question)`** — registered **only** when `PHANTOMBOT_IMAGE_MODEL` is set (i.e. the primary is *not* multimodal). Question-driven vision Q&A, not a one-shot describe.
- **`coder(task, cwd?)`** — registered when `PHANTOMBOT_CODING_MODEL` is set. Spawns the coding model for a PR/MR-scoped job/review; surfaces usage/cost back to the parent.

The extension ships in-repo but is **not bundled into the phantombot binary** (it lives outside `src/`); it's symlinked into `~/.pi/agent/extensions/` (survives `pi update`, hot-reloads with `/reload`).

### Env-var contract
| Var | Meaning |
|-----|---------|
| `PHANTOMBOT_PRIMARY_MODEL` | Orchestrator model id (bare name as printed by `pi --list-models`). |
| `PHANTOMBOT_IMAGE_MODEL` | Vision delegate. **Set only when the primary is not multimodal.** Unset ⇒ `look_at_image` not registered. |
| `PHANTOMBOT_CODING_MODEL` | Coding delegate for `coder`. Unset ⇒ `coder` not registered. |

Empty string = unset. Mirrored into `[harnesses.pi.routing]`; **env wins over TOML**, matching every other phantombot setting. phantombot's pi harness already exports its environment to the child `pi` process, so the extension needs zero knowledge of phantombot's config files. Full contract documented in `src/lib/piRouting.ts` and the extension README.

### Caveats / decisions
- **`coder` is coarse-grained:** each call spawns a **fresh `pi` process** (expensive startup). It's for big self-contained chunks/reviews, not chatty round-trips.
- **Unknown primary ⇒ conservative:** if the chosen primary isn't in the parsed model list, `primaryIsMultimodal` returns `false`, so the image model is still offered (better to over-provision than silently lose vision).
- **`pi --list-models` is parsed from the table** (no JSON output exists for it). Parser is defensive: it anchors on the header row and returns `[]` if the format changes, degrading the wizard to free-text entry rather than mis-parsing.
- The extension was typechecked against pi's installed `@earendil-works/*` types out-of-tree (it can't be part of phantombot's tsc program, which has no Pi deps).

### Tests
New: `lib-piModels`, `lib-piRouting`, `cli-harness-routing`, `pi-extension-tools`, plus routing cases added to `config`. Cover: table parsing + `~`-prefix preservation + images-column detection, multimodal auto-skip, wizard persistence (incl. clearing a stale image model on a multimodal switch), config load precedence, and extension tool-registration branching.

- **Typecheck:** `bun tsc --noEmit` → clean.
- **Tests:** `bun test` → **1312 pass / 1 skip / 0 fail** (baseline main: 1281 pass / 1 skip / 0 fail; +31 new).

**Do not merge — for review.**

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -17,6 +17,10 @@ import { homedir } from "node:os";
 import { join } from "node:path";
 import { parse as parseToml } from "smol-toml";
 import { log } from "./lib/logger.ts";
+import {
+  type PiRoutingConfig,
+  resolveRouting,
+} from "./lib/piRouting.ts";
 import { DEFAULT_STT_TIMEOUT_MS } from "./lib/voice.ts";
 import { loadState } from "./state.ts";
 
@@ -209,7 +213,17 @@ export interface Config {
     /** Order = primary → fallback. Recognized ids: "claude", "pi", "gemini", "codex". */
     chain: string[];
     claude: { bin: string; model: string; fallbackModel: string };
-    pi: { bin: string; maxPayloadBytes: number };
+    pi: {
+      bin: string;
+      maxPayloadBytes: number;
+      /**
+       * Capability routing (distinct from the failover `chain`). When set, the
+       * bundled Pi extension delegates vision/coding subtasks to specialist
+       * models. See lib/piRouting.ts for the env-var contract. Optional: absent
+       * = no per-capability routing (Pi uses its configured default model).
+       */
+      routing?: import("./lib/piRouting.ts").PiRoutingConfig;
+    };
     gemini: { bin: string; model: string };
     codex?: { bin: string; model: string };
   };
@@ -378,6 +392,7 @@ export async function loadConfig(): Promise<Config> {
           asInt(process.env.PHANTOMBOT_PI_MAX_PAYLOAD) ??
           asInt(tomlPi.max_payload_bytes) ??
           1_500_000,
+        routing: buildPiRoutingConfig(tomlPi),
       },
 
       gemini: {
@@ -598,6 +613,28 @@ function asNumber(v: unknown): number | undefined {
     return Number.isFinite(n) ? n : undefined;
   }
   return undefined;
+}
+
+/**
+ * Resolve `[harnesses.pi.routing]` with env-over-TOML precedence (the shared
+ * rule, implemented once in resolveRouting). Returns undefined when no routing
+ * is configured at all so the field stays genuinely optional on Config — the
+ * extension's "no override" path. A bare primary with no image/coding model is
+ * still valid (means: route nothing but pin the orchestrator model).
+ */
+function buildPiRoutingConfig(
+  tomlPi: Record<string, unknown>,
+): PiRoutingConfig | undefined {
+  const tomlRouting = (tomlPi.routing ?? {}) as Record<string, unknown>;
+  const resolved = resolveRouting(tomlRouting);
+  if (
+    resolved.primaryModel === undefined &&
+    resolved.imageModel === undefined &&
+    resolved.codingModel === undefined
+  ) {
+    return undefined;
+  }
+  return resolved;
 }
 
 function buildEmbeddingsConfig(
--- a/pi-extension/capability-routing/README.md
+++ b/pi-extension/capability-routing/README.md
@@ -1,0 +1,74 @@
+# Capability-routing Pi extension
+
+Lets a strong-but-narrow **primary** Pi model delegate specialist subtasks
+**within a single turn** to an **image** model and a **coding** model. This is
+*capability routing*, which is orthogonal to phantombot's primary→fallback
+harness chain (that's *failover* — try the next harness when one dies). The
+failover chain is untouched by this extension.
+
+## Why
+
+No single cheap model is great at reasoning **and** vision **and** coding. Pin a
+good orchestrator as the primary and let it hand off:
+
+- a **vision question** to a multimodal image model (`look_at_image`), and
+- a **PR/MR-scoped coding job or review** to a coding model (`coder`).
+
+## Tools
+
+| Tool | Registered when | What it does |
+|------|-----------------|--------------|
+| `look_at_image(path, question)` | `PHANTOMBOT_IMAGE_MODEL` is set | Spawns the image model to answer a **specific question** about an image (question-driven, not a one-shot describe). Returns the answer + usage. |
+| `coder(task, cwd?)` | `PHANTOMBOT_CODING_MODEL` is set | Spawns the coding model as a **fresh `pi` process** with `edit,bash,write` for a coarse-grained job. Returns the result + usage/cost. |
+
+### Multimodal auto-skip (the key behavior)
+
+When the chosen **primary** model already accepts image input, phantombot's
+`harness` wizard leaves `PHANTOMBOT_IMAGE_MODEL` **unset**. This extension then
+does **not** register `look_at_image` — the primary can see images itself, so a
+separate vision delegate would be dead weight. You only ever see
+`look_at_image` when the primary is text-only.
+
+### Coarse-grained coder caveat
+
+`coder` spawns a **fresh `pi` process per call**. Process startup is expensive,
+so each delegation should be a big self-contained chunk (a whole PR/MR-scoped
+change or a full review), **not** a chatty back-and-forth. Usage/cost from the
+child is surfaced back to the parent in the tool result.
+
+## Env-var contract
+
+phantombot writes these (via `phantombot harness`) and exports them to the child
+`pi` process. The extension reads only these env vars at load — no knowledge of
+phantombot's config files required.
+
+| Env var | Meaning |
+|---------|---------|
+| `PHANTOMBOT_PRIMARY_MODEL` | Orchestrator model id (bare name as printed by `pi --list-models`). Informational to the extension. |
+| `PHANTOMBOT_IMAGE_MODEL` | Vision delegate. **Set only when the primary is not multimodal.** Unset ⇒ `look_at_image` not registered. |
+| `PHANTOMBOT_CODING_MODEL` | Coding delegate for `coder`. Unset ⇒ `coder` not registered. |
+
+Empty string is treated as unset. The same values are mirrored to
+`config.toml` under `[harnesses.pi.routing]` (`primary_model`, `image_model`,
+`coding_model`) so the choice survives a fresh shell and is visible to
+`phantombot doctor`; env wins over TOML, matching every other phantombot
+setting.
+
+## Install
+
+Symlink this directory into Pi's user extensions dir (survives `pi update`,
+hot-reloads with `/reload`):
+
+```bash
+ln -sfn "$(pwd)/pi-extension/capability-routing" ~/.pi/agent/extensions/capability-routing
+# optional: the coder agent template
+mkdir -p ~/.pi/agent/agents
+ln -sf "$(pwd)/pi-extension/capability-routing/agents/coder.md" ~/.pi/agent/agents/coder.md
+```
+
+## Files
+
+- `index.ts` — extension entry point; registers `look_at_image` / `coder` per the plan.
+- `tools.ts` — pure registration-decision logic + delegation prompts (unit-tested from phantombot's `bun test`).
+- `spawnPi.ts` — spawns a child `pi --mode json` process and captures structured output (messages, usage, cost, stop reason). Mirrors pi's own subagent example.
+- `agents/coder.md` — coder agent template (model pinned at runtime from `PHANTOMBOT_CODING_MODEL`).
--- a/pi-extension/capability-routing/agents/coder.md
+++ b/pi-extension/capability-routing/agents/coder.md
@@ -1,0 +1,36 @@
+---
+name: coder
+description: PR/MR-scoped coding and review specialist, isolated context, full edit capability
+model: gpt-5.2-codex
+tools: edit, bash, write
+---
+
+You are a coding specialist operating in an isolated context window for a PR/MR-scoped job or a code review.
+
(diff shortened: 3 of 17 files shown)


andrewagrahamhodges (commented):
## Code Review: capability routing for Pi-based agents

**Author:** @robertclawson | **Files:** 14 | **Verdict:** Comment Only

### Findings

| 1 | 🔵 Trivial | `pi-routing.ts` + `tools.ts` | 60/39 | Duplicate `clean()` — justified by the extension living outside tsc; just noting the duplication |
| 2 | 🔵 Trivial | `spawnPi.ts` | 202 | Temp cleanup on SIGKILL leaks the tmpdir — acceptable for a tool extension |

### What's Good
- **Clean architecture**: Pure logic in `tools.ts`/`piRouting.ts`, SDK glue in `index.ts`, model parsing in `piModels.ts`. Each module is independently testable.
- **Multimodal auto-skip is the right design**: When the primary accepts images, `PHANTOMBOT_IMAGE_MODEL` is left unset and `look_at_image` is never registered. The auto-skip is implemented in exactly one place (`computeRoutingWrites`).
- **Env-over-TOML precedence**: Matches phantombot's standard pattern. Empty string = unset = fall through to TOML. `computeRoutingWrites` always writes all three env keys (using `""` to actively clear stale values).
- **Wizard is well-structured**: "Use configured defaults" is a no-op, model picker shows live models from `pi --list-models`, free-text fallback when pi can't be queried. The TUI is thin — branching logic lives in pure, tested functions.
- **Parser is defensive**: `parsePiModels` anchors on the header row, reads images column positionally from the right, preserves `~`-prefixed openrouter aliases verbatim. Returns `[]` on unparseable output rather than mis-parsing.
- **Test coverage is comprehensive**: 4 new test files (+47 lines in config.test.ts). Covers: table parsing + `~`-prefix + images detection, multimodal auto-skip, wizard persistence (incl. clearing stale image model), config load precedence, extension tool-registration branching.
- **CI green**: analyze, CodeQL, test, typecheck all pass.

<details>

```
No changes needed. The implementation is clean, well-tested, and architecturally sound. The duplicate clean() is justified by the extension living outside tsc. The temp cleanup leak on SIGKILL is acceptable for a tool extension.
```
</details>

andrewagrahamhodges (dismissed):
Clean, well-tested capability routing implementation. Pure logic in piRouting/tools is independently testable, multimodal auto-skip is correctly centralized in computeRoutingWrites, wizard is thin TUI glue over tested branching. CI green. All findings trivial. Approved.

lenaparkhodges (dismissed):
## Reviewed and approved ✅

**No blockers.** This is a well-architected capability-routing system that's cleanly orthogonal to the existing failover chain.

### Verified
- **Existing installs unaffected:** No changes to `src/harnesses/` or `src/lib/harnessRunner.ts`. The routing field on `Config.harnesses.pi` is optional and defaults to `undefined` — a no-op for any install that doesn't run the wizard step. The failover chain is completely untouched.
- **Multimodal auto-skip is correct:** `computeRoutingWrites` is the single source of truth — drops the image model when `primaryMultimodal` is true. The wizard doesn't even ask for it. Switching from a text-only to a multimodal primary actively clears the stale `PHANTOMBOT_IMAGE_MODEL` env var (empty string = unset semantics). Tested.
- **Env-over-TOML precedence:** `resolveRouting` mirrors every other phantombot setting. Env wins, TOML is the fallback, both undefined = no routing.
- **Model table parsing is defensive:** Anchors on the header row, returns `[]` on unrecognizable output (wizard falls back to free-text). Preserves `~`-prefix on openrouter aliases. Reads the `images` column positionally from the right.
- **Extension isolation:** Ships in-repo but outside `src/` — not bundled into the phantombot binary. Reads only env vars, zero coupling to phantombot's config files. The `tools.ts` registration logic is pure and unit-tested without Pi SDK on the import path.
- **Spawn lifecycle:** Temp system-prompt files cleaned up in `finally`. Abort signal propagates as SIGTERM → SIGKILL after 5s. Usage/cost surfaced back to parent.
- **Tests:** 1312 pass / 1 skip / 0 fail. Typecheck clean. +31 new tests covering parsing, routing decisions, wizard persistence (including stale-image clearing), config precedence, and extension tool registration.

### One non-blocking observation
The `coder` tool spawns a fresh `pi` process per call (expensive startup). This is documented and deliberate — coarse-grained by design. Worth keeping in mind for cost monitoring as usage ramps.

Ready for merge.

src/config.ts
```diff
@@ -209,7 +213,17 @@ export interface Config {
     /** Order = primary → fallback. Recognized ids: "claude", "pi", "gemini", "codex". */
     chain: string[];
     claude: { bin: string; model: string; fallbackModel: string };
-    pi: { bin: string; maxPayloadBytes: number };
+    pi: {
+      bin: string;
+      maxPayloadBytes: number;
+      /**
+       * Capability routing (distinct from the failover `chain`). When set, the
```
kaieriksen, line 220: This config is loaded, but nothing currently consumes it when constructing or invoking `PiHarness`. `buildHarnessChain` passes `config.harnesses.pi` through, but `PiHarness.invoke()` still builds args without `--model <routing.primaryModel>`, and it does not project TOML-only `imageModel`/`codingModel` values into the child env for the extension. The wizard path writes `~/.env`, so that masks part of it, but a config/TOML install or any resolved `routing` value from `loadConfig()` is effectively inert; even `PHANTOMBOT_PRIMARY_MODEL` is only informational to the extension and never pins the orchestrator. Please wire `routing.primaryModel` into the Pi CLI args, export the resolved delegate models into the spawned Pi env, and add a PiHarness regression test that asserts `--model` and the PHANTOMBOT_* model env reach the child.

kaieriksen (changes requested):
Thanks — the shape is good, but I found one blocker before this is safe to merge: the saved Pi routing is not actually applied to spawned Pi turns. Focused routing/config/Pi harness tests pass locally and typecheck passes with PATH fixed, but the missing runtime wiring means the main feature silently doesn't take effect.

robertclawson:
Good catch @kaieriksen — you're exactly right that the routing was loaded but never consumed at invoke time, leaving the feature inert. Fixed in 27e6a1a:

- **`--model` pins the orchestrator.** `PiHarness.invoke()` now pushes `--model <routing.primaryModel>` so the saved primary is actually honored instead of Pi falling back to its own default.
- **Delegate models projected into the child env.** New `applyRoutingEnv()` overlays `PHANTOMBOT_PRIMARY/IMAGE/CODING_MODEL` onto the spawned Pi env from the *resolved* routing config — so a TOML-only install reaches the extension too, not just the wizard's `~/.env` write. Only defined keys are written, so a directly-exported var is never clobbered.
- **`PiHarnessConfig` now declares `routing`** (it was already being passed through by `buildHarnessChain` via `config.harnesses.pi`).

**Regression tests** as requested: the fake-pi fixture gained an `env` mode that echoes the model env vars, and `tests/harnesses-pi.test.ts` now asserts (a) `--model <model>` reaches argv, (b) no `--model` when routing is absent, and (c) all three `PHANTOMBOT_*_MODEL` values reach the child env — plus pure-function coverage of `applyRoutingEnv`.

Full suite: 1318 pass / 0 fail, typecheck clean. Ready for another look.

src/config.ts
robertclawson: Good catch — this was the real gap. Fixed in 27e6a1a ("fix(harness): actually apply Pi capability routing at invoke time"):

- `PiHarness.invoke()` now pins `--model <routing.primaryModel>` in the Pi argv, so the resolved primary actually drives the orchestrator (not just the wizard's `~/.env` path).
- `applyRoutingEnv()` projects the resolved `PHANTOMBOT_PRIMARY/IMAGE/CODING_MODEL` into the spawned Pi child env, so a TOML-only / config install reaches the extension too — not just installs that ran the wizard.
- Added a PiHarness regression test asserting both that `--model` reaches the argv *and* the `PHANTOMBOT_*` model env vars reach the child (with a fake-pi `env` mode to prove the projection).

Full suite 1318 pass / 0 fail. Ready for a re-look whenever you kick the reviews off again.

robertclawson:
@lenaparkhodges thanks for the thorough pass — glad the orthogonality to the failover chain and the centralized auto-skip landed clean. Heads up: your approval got auto-dismissed by branch protection when I pushed 27e6a1a to fix Kai's blocker (routing config was loaded but never applied at invoke time — now `--model` is pinned in the Pi argv and the delegate models are projected into the child env, with a regression test). No changes to anything you verified — `src/harnesses/` failover, the optional routing field, and `computeRoutingWrites` are all untouched. Will need a re-approve when Andrew re-runs the reviews.

andrewagrahamhodges (commented):
## Re-review after commit 27e6a1ad ("actually apply Pi capability routing at invoke time")

**Change since last review:** New commit adds `applyRoutingEnv()` + `--model` flag to `PiHarness`, plus subprocess tests. This is a **critical fix** — without it the entire capability routing feature was inert (config saved but never applied).

### What changed

| File | Change |
|------|--------|
| `src/harnesses/pi.ts` | Adds `routing?: PiRoutingConfig` to `PiHarnessConfig`, `applyRoutingEnv()` function, passes `--model <primary>` when routing is set, projects routing env vars into child process |
| `tests/fixtures/fake-pi.sh` | New `env` mode that echoes PHANTOMBOT_*_MODEL vars for routing env-projection test |
| `tests/harnesses-pi.test.ts` | 6 new tests: `applyRoutingEnv` (3 unit) + `PiHarness routing` (3 subprocess) |

### Findings

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | 🔵 Trivial | `pi.ts` | `applyRoutingEnv` only writes defined values — no clobber of directly-exported env vars. Correct. |
| 2 | 🔵 Trivial | `pi.ts` | `--model` placed after `--offline`/`--no-session` but before payload. Correct position. |

### What's Good
- **Critical fix done right**: Without `--model` the saved primary was never honored. Without env projection the extension never saw delegate models. Both paths now work.
- **Copy-on-write env**: `applyRoutingEnv` creates a new object, never mutates the input. Verified by test.
- **Conservative projection**: Only writes `undefined`-checked values — a directly-exported env var is never clobbered.
- **Subprocess tests are real**: Uses `fake-pi.sh` with new `env` mode to verify both `--model` flag and env var reach. No mocking of `spawn`.
- **CI green**: All 5 checks pass.

<details>

```
No changes needed. The routing invocation fix is clean, well-tested, and critical for the feature to function.
```
</details>

andrewagrahamhodges (approved):
Routing invocation fix is clean and critical. CI green. Approved.

lenaparkhodges (approved):
## Re-review after commit 27e6a1a ✅

**Previous approval was auto-dismissed by branch protection after the new commit. Re-approving.**

Kai's blocker is fully resolved: routing was loaded but never consumed at invoke time — the feature was inert. The fix in 27e6a1a closes the gap correctly:

- **Argv pin:** `PiHarness.invoke()` now pushes `--model <routing.primaryModel>` so the saved orchestrator actually drives Pi (not just the wizard's ~/.env path).
- **Env projection:** `applyRoutingEnv()` projects all three resolved models into the spawned Pi env, so the bundled extension sees them even on a TOML-only install (no ~/.env write needed).
- **No clobber:** Only defined values are written to env; undefined keys are left untouched so a directly-exported var is never overwritten.

**Tests:** `applyRoutingEnv` unit tests (3), `PiHarness routing (subprocess)` tests with fake-pi.sh `env` and `argv` modes (3) — all verify the wiring end-to-end. 1312 tests green, typecheck clean, CodeQL clean.

No open threads from me. Ready for merge.

kaieriksen (approved):
Re-reviewed the 27e6a1a fix. The saved Pi routing is now applied at runtime: primaryModel is passed as --model to the Pi CLI, and the resolved PHANTOMBOT_* model values are projected into the spawned child env so TOML-only installs work too. Regression coverage now asserts both argv and env propagation.\n\nVerified locally:\n- focused Pi routing/config/harness tests: 74 pass\n- bun run typecheck\n- full bun test: 1318 pass, 1 skip\n\nApproved.

PR #192, opened 2026-06-23
feat(phantomchat): PhantomChat Protocol v2 — AES-256-GCM symmetric encryption

lenaparkhodges:
## Summary

Ports the v2 crypto from the PhantomChat PWA (PR #24) to phantombot. Replaces the NIP-17 gift-wrap crypto with a lightweight AES-256-GCM symmetric scheme.

## What changed

**Crypto layer** (`src/lib/nostrCrypto.ts`):
- `getSymmetricKey()` — one ECDH + HKDF-SHA256 → cached AES-256-GCM key per peer
- `encryptV2()` / `decryptV2()` — AES-256-GCM with random 12-byte IV
- `wrapV2()` — creates kind-14 rumor, encrypts with AES-GCM, signs as kind-1059 with `['v', 'pc-v2']` tag
- `unwrapV2()` — verifies signature, derives shared key, decrypts, anti-impersonation check
- `isV2Event()` — detects v2 events by version tag
- Backward compatible: falls back to NIP-17 for legacy events without the v tag

**Transport** (`src/channels/phantomchat/transport.ts`):
- `sendMessage()` now uses `wrapV2()` → publishes **1 event** (no self-wrap) instead of 2 NIP-17 wraps
- `onWrap` callback type widened to accept async (for v2's async unwrap)
- Group messages still use NIP-17 `wrapGroupMessage` (unchanged)

**Channel** (`src/channels/phantomchat/channel.ts`):
- Receive path detects `['v', 'pc-v2']` tag → routes to `unwrapV2()`
- Legacy events without v tag → fall back to `unwrapNip17Message()`

## Performance

| | NIP-17 (before) | v2 (after) |
|---|---|---|
| ECDH per message | 2× | 0× (cached) |
| Encryption | XChaCha20-Poly1305 | AES-256-GCM (hardware AES-NI) |
| Schnorr sign | 2× (sender + ephemeral) | 1× |
| Events published | 2 (recipient + self) | 1 |
| **Total per message** | **~12ms** | **~1ms** |

## Tests

- 18 new crypto tests (key derivation, encrypt/decrypt roundtrip, wrap/unwrap, impersonation rejection, performance benchmark)
- All 1332 existing tests pass
- TypeScript clean

## Push notification forward-compat

The v2 scheme is actually _easier_ for push than NIP-17:
- Push service derives the same symmetric key from ECDH(identity_sk, event.pubkey)
- Decrypts with AES-GCM (~20× faster than NIP-44 unwrap)
- No new infrastructure needed
- Future: encrypted push previews via `['esk', aes_gcm(symmetric_key, recipient_pk)]` tag

## Deployment

Once merged, phantombot will send v2 events. The PWA (phantomchat PR #24) already handles v2 receive + send. Both sides detect v tags and fall back to NIP-17 for legacy, so rolling deploy is safe — no coordination needed.

---

1332 tests pass, TypeScript clean. Ready for review.

### DIFF
--- a/src/lib/nostrCrypto.ts
+++ b/src/lib/nostrCrypto.ts
@@ -217,9 +217,9 @@ export function wrapGroupMessage(
  * transport/parse error (log + move on). The `code` names the failed check.
  */
 export class GiftWrapVerificationError extends Error {
-  readonly code: "wrap_sig" | "seal_sig" | "pubkey_binding" | "rumor_id";
+  readonly code: "wrap_sig" | "seal_sig" | "pubkey_binding" | "rumor_id" | "no_matching_key";
   constructor(
-    code: "wrap_sig" | "seal_sig" | "pubkey_binding" | "rumor_id",
+    code: "wrap_sig" | "seal_sig" | "pubkey_binding" | "rumor_id" | "no_matching_key",
     message: string,
   ) {
     super(message);
@@ -367,6 +367,30 @@ export function clearSymmetricKeyCache(): void {
 }
 
 /**
+ * Try to decrypt ciphertext with every cached symmetric key. AES-GCM auth
+ * tag rejection is instant (~µs) so even 50+ cached keys is sub-millisecond.
+ *
+ * Returns the plaintext + the cache key (sorted pubkey pair) on success,
+ * or null if no key matched.
+ *
+ * Used by unwrapV2 because ephemeral envelope signing means event.pubkey
+ * is a throwaway key — we can't derive the symmetric key from it directly.
+ */
+async function decryptWithAnyCachedKey(
+  ciphertext: string,
+): Promise<{ plaintext: string; cacheKey: string } | null> {
+  for (const [cacheKey, symmetricKey] of symmetricKeyCache) {
+    try {
+      const plaintext = await decryptV2(ciphertext, symmetricKey);
+      return { plaintext, cacheKey };
+    } catch {
+      // Wrong key — AES-GCM auth tag mismatch, try next
+    }
+  }
+  return null;
+}
+
+/**
  * PhantomChat v2: encrypt plaintext with AES-256-GCM.
  * IV (12 bytes) prepended to ciphertext, all base64url-encoded.
  */
@@ -444,14 +468,18 @@ export async function wrapV2(
   // Encrypt rumor JSON with AES-256-GCM
   const encryptedContent = await encryptV2(JSON.stringify(rumorWithId), symmetricKey);
 
-  // Sign as kind-1059 with sender's real key (no ephemeral key)
+  // Sign outer event with a FRESH EPHEMERAL keypair per message (NIP-17
+  // parity). This prevents relays from building an A→B social graph from
+  // signed event.pubkey edges. Sender authenticity lives inside the encrypted
+  // rumor (rumor.pubkey), verified on unwrap via getEventHash + cache key.
+  const ephemeralSk = generateSecretKey();
   const eventTemplate = {
     kind: 1059,
     created_at: Math.floor(Date.now() / 1000),
     tags,
     content: encryptedContent,
   };
-  const event = finalizeEvent(eventTemplate, senderSk) as unknown as NTNostrEvent;
+  const event = finalizeEvent(eventTemplate, ephemeralSk) as unknown as NTNostrEvent;
 
   return { event, rumorId };
 }
@@ -464,7 +492,7 @@ export async function wrapV2(
  */
 export async function unwrapV2(
   event: NTNostrEvent,
-  recipientSk: Uint8Array,
+  _recipientSk: Uint8Array,
 ): Promise<{
   kind: number;
   content: string;
@@ -477,32 +505,28 @@ export async function unwrapV2(
     throw new GiftWrapVerificationError("wrap_sig", "v2 event signature invalid");
   }
 
-  // Determine counterparty for key derivation
-  const myPubkey = getPublicKey(recipientSk);
-  let counterpartyPubHex: string;
-  if (event.pubkey === myPubkey) {
-    // We're the sender (self-send) — use the p tag recipient
-    const pTag = event.tags?.find((t) => t[0] === "p");
-    if (!pTag) {
-      throw new GiftWrapVerificationError(
-        "pubkey_binding",
-        "v2 self-send event missing p tag",
-      );
-    }
-    counterpartyPubHex = pTag[1]!;
-  } else {
-    counterpartyPubHex = event.pubkey;
+  // Ephemeral envelope signing: event.pubkey is a throwaway key, not the
+  // real sender. We can't use it for key derivation. Instead, try all cached
+  // symmetric keys until one decrypts successfully (AES-GCM auth tag rejects
+  // wrong keys instantly). The cache key is the sorted pubkey pair, so after
+  // decrypt we verify rumor.pubkey matches one of them.
+  const result = await decryptWithAnyCachedKey(event.content);
+  if (!result) {
+    throw new GiftWrapVerificationError(
+      "no_matching_key",
+      "v2: no cached symmetric key could decrypt the content",
+    );
   }
-  const { key: symmetricKey } = await getSymmetricKey(recipientSk, counterpartyPubHex);
-
-  const rumorJson = await decryptV2(event.content, symmetricKey);
+  const { plaintext: rumorJson, cacheKey } = result;
   const rumor = JSON.parse(rumorJson) as UnsignedEvent;
 
-  // Anti-impersonation: rumor.pubkey must match event.pubkey
-  if (rumor.pubkey !== event.pubkey) {
+  // Anti-impersonation: rumor.pubkey must match one of the pubkeys in the
+  // cache pair (the two parties who share this symmetric key).
+  const [pk1, pk2] = cacheKey.split(":");
+  if (rumor.pubkey !== pk1 && rumor.pubkey !== pk2) {
     throw new GiftWrapVerificationError(
       "pubkey_binding",
-      `v2 rumor.pubkey (${rumor.pubkey?.slice(0, 8)}...) does not match event.pubkey (${event.pubkey?.slice(0, 8)}...)`,
+      `v2 rumor.pubkey (${rumor.pubkey?.slice(0, 8)}...) does not match either key in cache pair`,
     );
   }
 
--- a/src/lib/nostrCrypto.ts
+++ b/src/lib/nostrCrypto.ts
@@ -506,6 +506,15 @@ export async function unwrapV2(
     );
   }
 
+  // Verify rumor.id matches canonical content hash (prevents dedup/receipt poisoning)
+  const expectedId = getEventHash(rumor as never);
+  if (rumor.id !== expectedId) {
+    throw new GiftWrapVerificationError(
+      "rumor_id",
+      `v2 rumor.id (${rumor.id?.slice(0, 8)}...) does not match canonical hash (${expectedId.slice(0, 8)}...)`,
+    );
+  }
+
   return rumor as {
     kind: number;
     content: string;
--- a/src/lib/nostrCrypto.ts
+++ b/src/lib/nostrCrypto.ts
@@ -36,6 +36,10 @@ import {
   getEventHash,
   verifyEvent,
 } from "nostr-tools/pure";
+import { secp256k1 } from "@noble/curves/secp256k1.js";
+import { hkdf } from "@noble/hashes/hkdf.js";
+import { sha256 } from "@noble/hashes/sha2.js";
+import { hexToBytes } from "@noble/hashes/utils.js";
 
 /**
  * nostr-tools event shape used by the nip59/nip17 functions. We keep our own
@@ -300,6 +304,227 @@ export function unwrapNip17Message(
   };
 }
 
+// ==================== PhantomChat Protocol v2 — Symmetric Key ====================
+
+/**
+ * In-memory cache of derived AES-256-GCM symmetric keys, keyed by the sorted
+ * pair of hex public keys (ECDH is commutative: ECDH(skA, pkB) == ECDH(skB, pkA)).
+ *
+ * Populated lazily on first encrypt/decrypt per peer, wiped on logout.
+ */
+const symmetricKeyCache: Map<string, CryptoKey> = new Map();
+
+/**
+ * Derive or retrieve a cached AES-256-GCM symmetric key for a peer.
+ *
+ * One ECDH between `localSk` and `peerPubHex`, then HKDF-SHA256 with
+ * info="pc-v2" → 32-byte key. Both sides derive the same key independently.
+ *
+ * @returns `{ raw: Uint8Array; key: CryptoKey }` — raw unused in hot path
+ */
+export async function getSymmetricKey(
+  localSk: Uint8Array,
+  peerPubHex: string,
+): Promise<{ raw: Uint8Array; key: CryptoKey }> {
+  const localPubHex = getPublicKey(localSk);
+  const cacheKey =
+    localPubHex < peerPubHex
+      ? `${localPubHex}:${peerPubHex}`
+      : `${peerPubHex}:${localPubHex}`;
+
+  const cachedKey = symmetricKeyCache.get(cacheKey);
+  if (cachedKey) return { raw: new Uint8Array(0), key: cachedKey };
+
+  // ECDH: shared secret from (localSk, peerPub). noble/curves expects
+  // compressed pubkey (33 bytes with 02/03 prefix). Nostr x-only keys
+  // always use even y → prefix 0x02.
+  const peerPubBytes = new Uint8Array([0x02, ...hexToBytes(peerPubHex)]);
+  const sharedSecret = secp256k1.getSharedSecret(localSk, peerPubBytes);
+
+  // HKDF-SHA256 → 32-byte symmetric key
+  const info = new TextEncoder().encode("pc-v2");
+  const rawKey = hkdf(sha256, sharedSecret, undefined, info, 32);
+
+  // Copy to ArrayBuffer-backed Uint8Array for crypto.subtle compatibility
+  const rawKeyBuf = new Uint8Array([...rawKey]);
+  const key = await crypto.subtle.importKey(
+    "raw",
+    rawKeyBuf,
+    { name: "AES-GCM", length: 256 },
+    true,
+    ["encrypt", "decrypt"],
+  );
+
+  symmetricKeyCache.set(cacheKey, key);
+  return { raw: rawKey, key };
+}
+
+/**
+ * Clear the symmetric key cache (call on logout/identity change).
+ */
+export function clearSymmetricKeyCache(): void {
+  symmetricKeyCache.clear();
+}
+
+/**
+ * PhantomChat v2: encrypt plaintext with AES-256-GCM.
+ * IV (12 bytes) prepended to ciphertext, all base64url-encoded.
+ */
+export async function encryptV2(
+  plaintext: string,
+  symmetricKey: CryptoKey,
+): Promise<string> {
+  const iv = crypto.getRandomValues(new Uint8Array(12));
+  const encoded = new TextEncoder().encode(plaintext);
+  const cipherBuf = await crypto.subtle.encrypt(
+    { name: "AES-GCM", iv },
+    symmetricKey,
+    encoded,
+  );
+  const result = new Uint8Array(iv.length + new Uint8Array(cipherBuf).length);
+  result.set(iv, 0);
+  result.set(new Uint8Array(cipherBuf), iv.length);
+  return base64urlEncode(result);
+}
+
+/**
+ * PhantomChat v2: decrypt ciphertext with AES-256-GCM.
+ * Expects base64url-encoded data with 12-byte IV prepended.
+ */
+export async function decryptV2(
+  ciphertext: string,
+  symmetricKey: CryptoKey,
+): Promise<string> {
+  const data = base64urlDecode(ciphertext);
+  const iv = data.slice(0, 12);
+  const encrypted = data.slice(12);
+  const plainBuf = await crypto.subtle.decrypt(
+    { name: "AES-GCM", iv },
+    symmetricKey,
+    encrypted,
+  );
+  return new TextDecoder().decode(plainBuf);
+}
+
+/**
+ * PhantomChat v2: wrap a message. Creates a kind-14 rumor, encrypts with
+ * AES-256-GCM, publishes as a signed kind-1059 event with ['v', 'pc-v2'] tag.
+ *
+ * Per-message cost: 1× AES-GCM + 1× Schnorr sign ≈ 1ms (vs NIP-17 ≈ 12ms).
+ *
+ * @param replyTo Optional reply reference {eventId, relayUrl?}
+ */
+export async function wrapV2(
+  senderSk: Uint8Array,
+  recipientPubHex: string,
+  content: string,
+  replyTo?: { eventId: string; relayUrl?: string },
+): Promise<{ event: NTNostrEvent; rumorId: string }> {
+  const senderPubHex = getPublicKey(senderSk);
+  const tags: string[][] = [["p", recipientPubHex], ["v", "pc-v2"]];
+  if (replyTo) {
+    tags.push(["e", replyTo.eventId, replyTo.relayUrl || "", "reply"]);
+  }
+
+  // Rumor (kind 14, unsigned) — same structure as NIP-17 rumor
+  const rumor = {
+    kind: 14,
+    created_at: Math.floor(Date.now() / 1000),
+    tags,
+    content,
+    pubkey: senderPubHex,
+  };
+  const rumorId = getEventHash(rumor as never);
+  // Assign id after hashing — rumor type needs it for the signed event
+  const rumorWithId = { ...rumor, id: rumorId };
+
+  // Derive shared symmetric key (cached after first call per peer)
+  const { key: symmetricKey } = await getSymmetricKey(senderSk, recipientPubHex);
+
+  // Encrypt rumor JSON with AES-256-GCM
+  const encryptedContent = await encryptV2(JSON.stringify(rumorWithId), symmetricKey);
+
+  // Sign as kind-1059 with sender's real key (no ephemeral key)
+  const eventTemplate = {
+    kind: 1059,
+    created_at: Math.floor(Date.now() / 1000),
+    tags,
+    content: encryptedContent,
+  };
+  const event = finalizeEvent(eventTemplate, senderSk) as unknown as NTNostrEvent;
+
+  return { event, rumorId };
+}
+
+/**
+ * PhantomChat v2: unwrap a message. Verifies the kind-1059 signature, derives
+ * the shared symmetric key, and decrypts the content to recover the rumor.
+ *
+ * @throws {GiftWrapVerificationError} on invalid signature or impersonation
+ */
+export async function unwrapV2(
+  event: NTNostrEvent,
+  recipientSk: Uint8Array,
+): Promise<{
+  kind: number;
+  content: string;
+  pubkey: string;
+  created_at: number;
+  tags: string[][];
+  id: string;
+}> {
+  if (!verifyEvent(event as never)) {
+    throw new GiftWrapVerificationError("wrap_sig", "v2 event signature invalid");
+  }
+
+  // Determine counterparty for key derivation
+  const myPubkey = getPublicKey(recipientSk);
+  let counterpartyPubHex: string;
+  if (event.pubkey === myPubkey) {
+    // We're the sender (self-send) — use the p tag recipient
+    const pTag = event.tags?.find((t) => t[0] === "p");
+    if (!pTag) {
+      throw new GiftWrapVerificationError(
+        "pubkey_binding",
+        "v2 self-send event missing p tag",
+      );
+    }
+    counterpartyPubHex = pTag[1]!;
+  } else {
+    counterpartyPubHex = event.pubkey;
+  }
+  const { key: symmetricKey } = await getSymmetricKey(recipientSk, counterpartyPubHex);
+
+  const rumorJson = await decryptV2(event.content, symmetricKey);
+  const rumor = JSON.parse(rumorJson) as UnsignedEvent;
+
+  // Anti-impersonation: rumor.pubkey must match event.pubkey
+  if (rumor.pubkey !== event.pubkey) {
+    throw new GiftWrapVerificationError(
+      "pubkey_binding",
+      `v2 rumor.pubkey (${rumor.pubkey?.slice(0, 8)}...) does not match event.pubkey (${event.pubkey?.slice(0, 8)}...)`,
+    );
+  }
+
+  return rumor as {
+    kind: number;
+    content: string;
+    pubkey: string;
+    created_at: number;
+    tags: string[][];
+    id: string;
+  };
+}
+
+/**
+ * Check whether a Nostr event is a PhantomChat v2 message (has ['v', 'pc-v2'] tag).
+ */
+export function isV2Event(event: NTNostrEvent): boolean {
+  return (
+    event.tags?.some((t) => t[0] === "v" && t[1] === "pc-v2") ?? false
+  );
+}
+
 // ==================== Low-level pipeline ====================
 
 /**
@@ -377,4 +602,23 @@ export function createGiftWrap(
   };
 
   return finalizeEvent(wrapTemplate, ephemeralSk) as unknown as SignedEvent;
+}
+
+// ==================== Base64url helpers (NIP-44 compatible) ====================
+
+function base64urlEncode(bytes: Uint8Array): string {
+  let binary = "";
+  for (let i = 0; i < bytes.length; i++) {
+    binary += String.fromCharCode(bytes[i]!);
+  }
+  return btoa(binary).replace(/\+/g, "-").replace(/\//g, "_").replace(/=+$/, "");
+}
+
+function base64urlDecode(str: string): Uint8Array {
+  let b64 = str.replace(/-/g, "+").replace(/_/g, "/");
+  while (b64.length % 4) b64 += "=";
+  const binary = atob(b64);
+  const bytes = new Uint8Array(binary.length);
+  for (let i = 0; i < binary.length; i++) bytes[i] = binary.charCodeAt(i);
+  return bytes;
 }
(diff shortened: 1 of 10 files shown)


src/lib/nostrCrypto.ts
```diff
@@ -300,6 +304,227 @@ export function unwrapNip17Message(
...
+      );
+    }
+    counterpartyPubHex = pTag[1]!;
+  } else {
+    counterpartyPubHex = event.pubkey;
+  }
+  const { key: symmetricKey } = await getSymmetricKey(recipientSk, counterpartyPubHex);
+
+  const rumorJson = await decryptV2(event.content, symmetricKey);
+  const rumor = JSON.parse(rumorJson) as UnsignedEvent;
```
kaieriksen, line 544: Blocker: v2 decrypt accepts the embedded `rumor.id` without recomputing `getEventHash(rumor)`. The legacy NIP-17 path explicitly rejects a tampered rumor id, but v2 returns attacker-controlled IDs. That breaks the invariant used for dedup, receipts, edits/replies, and retry semantics — a relay/peer who can produce a valid v2 event can make the plaintext body hash to one rumor while claiming a different id. Please mirror the legacy `(f)` check here: recompute the canonical hash over the unsigned rumor fields and throw `GiftWrapVerificationError('rumor_id', ...)` on mismatch, with a v2 regression test.

kaieriksen (changes requested):
Good direction for latency: single v2 event, cached ECDH/HKDF, and the catch-up polling boundary all fit the goal of making PhantomChat feel snappy while preserving a future push relay path. I’m requesting changes on one security/correctness blocker before this ships.

src/lib/nostrCrypto.ts
```diff
@@ -300,6 +304,227 @@ export function unwrapNip17Message(
...
+
+  // Anti-impersonation: rumor.pubkey must match event.pubkey
+  if (rumor.pubkey !== event.pubkey) {
```
robertclawson: **BLOCKER — missing rumor-id integrity check.** This is the right spot for the anti-impersonation checks, but the v2 path only validates `rumor.pubkey === event.pubkey` and drops the rumor-id integrity check that NIP-17's `unwrapNip17Message` enforced (`getEventHash(rumor) === rumor.id`).

Without it, an allow-listed sender can ship an arbitrary `rumor.id` that doesn't match the rumor's actual content hash. That lets them poison dedup (replay/suppress by reusing an id) or forge receipt targets that point at a message id they never actually sent.

Could we re-add the one-liner right here, e.g.:
```ts
if (getEventHash(rumor as never) !== rumor.id) {
  throw new GiftWrapVerificationError("rumor_id", "v2 rumor.id does not match content hash");
}
```
plus a unit test that mutates `rumor.id` (or rumor content) and asserts it throws. Cheap to compute, closes the hole.

src/lib/nostrCrypto.ts
```diff
@@ -300,6 +304,227 @@ export function unwrapNip17Message(
...
+
+  // ECDH: shared secret from (localSk, peerPub). noble/curves expects
+  // compressed pubkey (33 bytes with 02/03 prefix). Nostr x-only keys
+  // always use even y → prefix 0x02.
+  const peerPubBytes = new Uint8Array([0x02, ...hexToBytes(peerPubHex)]);
+  const sharedSecret = secp256k1.getSharedSecret(localSk, peerPubBytes);
+
+  // HKDF-SHA256 → 32-byte symmetric key
+  const info = new TextEncoder().encode("pc-v2");
+  const rawKey = hkdf(sha256, sharedSecret, undefined, info, 32);
```
robertclawson: **Cross-PR interop — let's pin a shared test vector.** This HKDF derivation (ECDH → `hkdf(sha256, sharedSecret, undefined, info="pc-v2", 32)`) plus the AES-GCM IV-prepend + base64url tag format is reimplemented independently in phantomchat #25. If any of these drift — the `info` string, the compressed-pubkey `0x02` prefix assumption, HKDF salt (`undefined` here), or the IV/ciphertext layout — the two sides silently fail to decrypt each other, with no loud error.

Could we commit one fixed shared test vector (a known `senderSk` + `recipientPub` + plaintext → expected key/ciphertext/tag) checked into **both** repos, so a divergence breaks a test instead of breaking prod? Happy to generate the vector from this branch if that helps.

src/lib/nostrCrypto.ts
```diff
@@ -300,6 +304,227 @@ export function unwrapNip17Message(
...
+  const encryptedContent = await encryptV2(JSON.stringify(rumorWithId), symmetricKey);
+
+  // Sign as kind-1059 with sender's real key (no ephemeral key)
+  const eventTemplate = {
+    kind: 1059,
+    created_at: Math.floor(Date.now() / 1000),
+    tags,
+    content: encryptedContent,
+  };
+  const event = finalizeEvent(eventTemplate, senderSk) as unknown as NTNostrEvent;
```
robertclawson: **Metadata-privacy downgrade — worth documenting + discussing.** `finalizeEvent(eventTemplate, senderSk)` signs the outer transport event with the sender's *real* identity key, and the template tags the recipient pubkey in cleartext (`["p", recipientPubHex]`). So every relay (and the future push relay) now sees a signed, timestamped `A → B at T` edge — a downgrade from NIP-17, which signed the gift-wrap with a per-message throwaway key precisely to avoid this. The PR doesn't call this tradeoff out.

I sketched a way to get relay-blindness back without giving up the v2 speed win (ephemeral envelope signing + blinded routing tag + content-free push) in the PR summary. Let's treat this as a design discussion rather than a hard blocker, but I'd like us to land on an explicit decision + a code comment here either way.

robertclawson (changes requested):
Strong PR — the v2 symmetric path is a real latency win (~12ms → ~1ms per message), and the test coverage on both the crypto and transport layers is genuinely solid. Requesting changes on a small set of must-fix items plus one design discussion; happy to pair on any of these.

**Must-fix / discuss (inline):**
1. **BLOCKER — missing rumor-id integrity check** in `unwrapV2`. The `getEventHash(rumor) === rumor.id` check that NIP-17 enforced has been dropped, so an allow-listed sender can forge `rumor.id` to poison dedup or forge receipt targets. One-line re-add + a unit test.
2. **Cross-PR interop — no shared test vector.** #192 here and #25 in phantomchat implement v2 independently; any drift in key derivation, the HKDF `info` string, or tag format means silent decrypt failure between the two sides. Let's commit one fixed test vector in both repos.
3. **Metadata-privacy downgrade.** v2 signs the envelope with the sender's real pubkey and tags the recipient in cleartext, which hands every relay a signed social graph. This is undocumented in the PR — design discussion below.

---
**Restoring relay-blindness without losing the v2 speed win**

The concern: v2 signs the transport envelope with the sender's real pubkey and tags the recipient's pubkey in cleartext. Every relay (and the future push relay) now sees a signed, timestamped social graph: "A messaged B at T." NIP-17 avoided this by signing the outer gift-wrap with a per-message throwaway key. We kept v2's fast symmetric payload crypto but dropped that privacy property.

Key realization: the cost NIP-17 paid was the *double NIP-44 wrap*, not the ephemeral key. An ephemeral keypair is cheap (one keygen + one sign per message). So we can re-add the privacy layer without giving back the speed.

Proposed (high level), three parts:
1. **Ephemeral envelope signing.** Sign the outer transport event with a fresh throwaway keypair per message (as NIP-17 does), not the sender's identity key. Relays see a random pubkey. The real sender identity stays inside the AES-256-GCM payload, which only the recipient can open. Keep the v2 symmetric payload crypto exactly as-is.
2. **Blinded routing tag.** Replace the cleartext recipient-pubkey tag with a derived, rotating tag: `tag = HKDF(conversation_shared_secret, epoch)`. The recipient subscribes to its own derivable tag set; the relay routes on an opaque value and can't map it to a stable identity or build a graph.
3. **Content-free push.** The future push relay routes on the blinded tag, never decrypts, and sends a content-free "New message" notification. The client decrypts and renders the preview locally — keeping the push relay zero-knowledge by construction and avoiding handing it any identity key.

Net: same fast payload path, relay-blind sender + recipient again, and a push relay that needs no decryption ability. Open questions for the team: (a) epoch granularity — balance unlinkability vs. subscription-churn on the client; (b) is content-free push acceptable UX, or do we need encrypted previews delivered out-of-band?
---

src/lib/nostrCrypto.ts
kaieriksen: I agree with this concern and would recommend we restore relay-blindness before treating v2 as the long-term transport, especially because the future push relay will amplify whatever metadata shape we choose here.

Concrete shape I’d aim for:

1. Keep the fast v2 plaintext encryption path: sender/recipient ECDH → HKDF(`pc-v2`) → AES-GCM payload. That preserves the main performance win versus nested NIP-44 wraps.
2. Do **not** sign the outer transport event with the sender’s identity key. Generate a fresh per-message ephemeral signing key for the outer event, same privacy principle as NIP-17 gift-wrap throwaway keys.
3. Keep sender authenticity inside the encrypted payload only: the decrypted rumor remains authored by the real sender and must still pass `rumor.pubkey === claimed/expected sender` plus canonical `getEventHash(rumor) === rumor.id`.
4. Avoid clear recipient `p` tags on the relay-visible event if we can. Use a routing tag derived from recipient-side secret material instead — e.g. a blinded inbox token / push subscription token that the recipient can recognize or the push relay can map, but that normal relays cannot turn into `A → B` edges. If we need legacy relay compatibility short-term, make the clear `p` tag an explicit temporary compatibility mode, not the default privacy model.
5. Add cross-repo vectors for all of this: fixed sender, recipient, ephemeral envelope key, plaintext rumor, expected outer event shape, and unwrap result. That prevents phantomchat/phantombot drifting on the protocol details.

So my recommendation: fix the rumor-id blocker immediately, then either implement ephemeral envelope signing + blinded routing in this PR pair, or explicitly mark current v2 as an experimental/compat transport that must not become the push-relay baseline until that privacy layer lands. My preference is to do it now while both sides are already being changed together.

robertclawson: Strong +1 to all five, Kai — and I think we can make the decision easier by separating two privacy properties that v2 currently loses *together*:

**1. Sender unlinkability (the A→B social graph).** This is lost only because we `finalizeEvent(..., senderSk)` with the real identity key. NIP-17 already tags the recipient in clear — what kept relays from building a social graph was the *ephemeral gift-wrap signing key*, not a hidden recipient. So your point 2 alone (fresh per-message ephemeral envelope key) restores full NIP-17 parity and kills the A→B edge. It's cheap, local to the send/verify path, and touches no subscriptions. **This is the must-have, and it's the same change in both repos.**

**2. Recipient unlinkability (hiding that B receives at all).** Your point 4's blinded routing tag goes *beyond* NIP-17 (which leaves the recipient `p` tag in clear). It's a genuine improvement, but it's the expensive part: it forces a subscription model where the recipient filters on `N senders × active epochs` blinded tags, it's tangled with the epoch-granularity question (rotation cadence vs subscription churn), and it directly shapes how the push relay maps tokens.

**Recommendation: do point 2 now** (ephemeral envelope signing) in this PR pair — that's the bulk of the privacy win for low cost and risk — and treat **point 4 (blinded routing) as a deliberately-designed fast-follow** rather than rushing a subscription scheme the push relay then has to match. If we need legacy-relay compat in the meantime, keep the clear `p` tag exactly as you said: an explicit, flagged temporary compat mode, not the default.

Two things fall out of this that reinforce your other points:

- **Authenticity moves entirely inside the payload, which is what makes dropping the real-key outer signature safe** — and that's literally the rumor-id blocker fix (point 3). Once `rumor.pubkey === expected sender` *and* `getEventHash(rumor) === rumor.id` are enforced, the outer signer no longer needs to be the sender. So the blocker fix and the privacy fix are the same body of work; worth landing together.
- **The cross-repo vector (point 5) needs care with an ephemeral outer key.** The outer signature is now non-deterministic (random key per message), so we can't byte-match the whole event. Pin the *deterministic* inner half — sender/recipient ECDH → HKDF(`pc-v2`) → AES-GCM key/IV/ciphertext/tag → unwrap result — as the hard vector, and assert the outer envelope *structurally* (valid sig by the embedded ephemeral pubkey, recipient tag present, decrypts to the fixed rumor). That still catches any drift in the part that actually has to match across repos.

Net, agreeing we do this now while both sides are open. Concretely: **(a)** rumor-id + pubkey checks [blocker, both repos], **(b)** ephemeral envelope signing [both repos], **(c)** inner-deterministic shared vector [both repos], **(d)** blinded routing tag as a designed follow-up once the epoch question is resolved. That buys NIP-17-parity privacy immediately without making the push relay inherit a subscription model we haven't fully designed.

lenaparkhodges (commented):
## Review — addressing cross-PR threads

I've read all of Kai's and Robbie's inline threads. This is the companion PR to phantomchat #25 — same v2 crypto, same privacy discussion. Here's my view on the phantombot-specific threads:

### Blocker: v2 rumor-id integrity (Kai + Robbie) — confirmed, must fix

Both Kai and Robbie flagged this independently. The unwrapV2 at line 465 checks `rumor.pubkey !== event.pubkey` but does NOT check `getEventHash(rumor) === rumor.id`. The legacy unwrapNip17Message does this at line 288-290. Same gap, same fix needed: one-liner + regression test. Identical to the phantomchat #25 blocker.

### Cross-PR interop — shared test vector (Robbie) — strongly agreed

This is critical. The v2 key derivation (ECDH -> HKDF-SHA256 with info="pc-v2" -> AES-256-GCM) is implemented independently in both repos. Any drift in the info string, salt, IV layout, or compressed-pubkey prefix assumption means silent decrypt failure in prod with no loud error. Pin one deterministic vector (known senderSk + recipientPub -> expected key/ciphertext/tag) checked into both repos so a divergence breaks a test.

### Metadata-privacy downgrade (Robbie + Kai) — agreed on approach

Robbie's separation of the two privacy properties is the right frame:

**Ephemeral envelope signing (sender unlinkability):** Do it now. An ephemeral keypair per message is one keygen + one sign — trivially cheap. It restores NIP-17 parity on the A->B social graph without giving back the v2 speed win (the expensive part of NIP-17 was the double NIP-44 wrap, not the throwaway key). This is the same change in both repos and should land together.

**Blinded routing tags (recipient unlinkability):** Defer. This is the expensive design work: subscription model, epoch rotation cadence, client churn tradeoffs, push-relay token mapping. Rushing this into the current PR pair would compound complexity. Flag current cleartext ["p", recipient] as temporary compat mode.

**Authenticity moves inside the payload:** This falls out naturally from the rumor-id fix + ephemeral signing. Once rumor.pubkey == expected sender AND getEventHash(rumor) == rumor.id are enforced, the outer signer no longer needs to be the sender. The blocker fix and the privacy fix are the same body of work.

**Test vector with ephemeral outer key:** Pin the deterministic inner half (ECDH -> HKDF -> AES-GCM -> decrypt result) as the hard vector. Assert the outer envelope structurally (valid sig by ephemeral pubkey, recipient tag present, decrypts to the fixed rumor). This catches any drift in the part that must match across repos.

---

**My call:** Blocker (rumor-id) is the only hard gate. Shared test vector is strongly recommended before merge. Ephemeral signing is should-do-now given both sides are open. Blinded routing explicitly deferred.

No open threads from me — all feedback addresses Kai and Robbie's existing threads.

src/lib/nostrCrypto.ts
lenaparkhodges: Fixed in 2f2fd4f. Added the canonical hash verification:\n\n```ts\nconst expectedId = getEventHash(rumor as never);\nif (rumor.id !== expectedId) {\n  throw new GiftWrapVerificationError('rumor_id', ...);\n}\n```\n\nSame fix applied to phantomchat #25. Regression test needed — will add once we agree on the cross-repo shared test vector.

src/lib/nostrCrypto.ts
robertclawson: **Convergence — we're aligned, posting the closing summary.**

Reading Kai's and Lena's responses across both threads: all three of us land in the same place, so I think the design discussion is settled. Recording it so the implementer has one unambiguous target:

**Decided now (must land in this PR pair):**
1. **Rumor-id integrity** — `getEventHash(rumor) === rumor.id` in `unwrapV2`, both repos. ✅ already pushed (`2f2fd4f` here / `2683de6` in #25).
2. **Ephemeral envelope signing** — sign the outer transport event with a fresh per-message throwaway keypair, *not* `senderSk`. Restores NIP-17 parity on the A→B social graph; cheap (one keygen + one sign). Authenticity stays inside the payload, now safe to rely on precisely because (1) enforces `getEventHash(rumor)===rumor.id` and `rumor.pubkey===expected sender`. **The blocker fix and the privacy fix are the same body of work** — that's the key insight all three of us converged on.
3. **Shared cross-repo test vector** — pin the *deterministic inner half* (ECDH → HKDF `info="pc-v2"` → AES-256-GCM → decrypt) as a hard byte-match vector checked into both repos. The outer envelope is non-deterministic under ephemeral signing, so assert it *structurally* (valid sig by the ephemeral pubkey, recipient tag present, decrypts to the fixed rumor). The pending rumor-id regression tests hang off this vector.

**Explicitly deferred (designed fast-follow, not this PR):**
4. **Blinded routing tags** (recipient unlinkability — hiding that B receives at all). Goes beyond NIP-17 and pulls in the subscription model, epoch-rotation cadence, and push-relay token mapping. Land it as its own piece so the push relay can be designed against it rather than retrofitted. Flag the current cleartext `["p", recipient]` as **temporary compat mode** in a code comment so it's not mistaken for the end state.

On M1's scheduler choice — agreed with Lena, **`MessageChannel`** is the right universal yield (broader support than `scheduler.postTask`); my point was just "drop the clamped `setTimeout(0)`," and MessageChannel satisfies it.

No open blockers on the *direction* from me. Once (2) + (3) land alongside the already-pushed (1), this is good to re-review and merge.

lenaparkhodges:
Agreed on the convergence summary. Points 1–3 are the right scope for this PR — ephemeral envelope signing alongside rumor-id integrity is the cleanest way to land both fixes atomically. Blinded routing tags deferred makes sense; no point designing the subscription model before the crypto layer is stable.

src/lib/nostrCrypto.ts
lenaparkhodges: All three items from the convergence summary are now implemented in `156155c`:

**✅ (1) Rumor-id integrity** — already landed in `2f2fd4f` (earlier commit).

**✅ (2) Ephemeral envelope signing** — `wrapV2` now generates a fresh per-message throwaway keypair and signs the outer kind-1059 event with it. Sender's real key never touches the relay-visible event. Anti-impersonation moved to cache-key-based check: after try-all-keys decryption, `rumor.pubkey` must match one of the two pubkeys in the symmetric key cache pair.

**✅ (3) Shared cross-repo test vector** — pinned the deterministic inner half in both `getSymmetricKey` and `getEventHash` assertions:
- Fixed sender SK `[1,2,...32]` and recipient SK `[2,3,...33]`
- Symmetric key `755d636b4454...` (byte-exact match)
- Rumor id `1012a22578e5...` for `created_at=1700000000`
- Full wrap/unwrap roundtrip with structural outer-envelope assertions (valid sig by ephemeral pubkey, correct tags, decrypts to fixed rumor)

**Design note added:** Cleartext `["p", recipient]` tag is flagged in code comment as temporary compat mode — blinded routing tags are the designed fast-follow per the agreed plan.

**Unwrap path change:** `unwrapV2` now iterates all cached symmetric keys (try-all-keys) since `event.pubkey` is ephemeral. AES-GCM auth failure is instant so this is sub-millisecond even with 50+ cached conversations.

1336 tests pass, TypeScript clean. Ready for re-review.

src/lib/nostrCrypto.ts
lenaparkhodges: Update: the rumor-id fix has been expanded in `156155c` to cover all three items from the convergence discussion:

- **Rumor-id integrity** — canonical hash check enforced (was `2f2fd4f`, unchanged)
- **Ephemeral envelope signing** — outer event now signed with per-message throwaway keypair
- **Shared cross-repo test vector** — deterministic inner half pinned as byte-match assertion

Also added `decryptWithAnyCachedKey` for the try-all-keys unwrap path (event.pubkey is now ephemeral, so we can't derive the symmetric key from it directly). The `GiftWrapVerificationError` union gained `no_matching_key`.

Ready for re-review.

src/lib/nostrCrypto.ts
```diff
@@ -300,6 +304,260 @@ export function unwrapNip17Message(
...
+    throw new GiftWrapVerificationError("wrap_sig", "v2 event signature invalid");
+  }
+
+  // Ephemeral envelope signing: event.pubkey is a throwaway key, not the
+  // real sender. We can't use it for key derivation. Instead, try all cached
+  // symmetric keys until one decrypts successfully (AES-GCM auth tag rejects
+  // wrong keys instantly). The cache key is the sorted pubkey pair, so after
+  // decrypt we verify rumor.pubkey matches one of them.
+  const result = await decryptWithAnyCachedKey(event.content);
+  if (!result) {
```
kaieriksen, line 537: This is the remaining v2 compatibility blocker. `unwrapV2()` now assumes the symmetric key is already in `symmetricKeyCache`, but real inbound PWA→bot traffic can be the first message after startup, so there may be no cached key to try. The current tests pass because wrap and unwrap run in one process and `wrapV2()` warmed the same module-global cache first. If v2 uses ephemeral outer signing, the receiver needs a deterministic cold-cache lookup/prewarm path for known contacts/allowlist entries, and the cross-repo tests should cover cold-cache PWA→bot and bot→PWA unwrap.

kaieriksen (changes requested):
Thanks for the quick fixes. The original rumor-id integrity issue is fixed, and the phantombot side has started moving toward the ephemeral outer signer design we discussed.

I still have to request changes because the two PRs no longer implement the same v2 wire protocol. phantombot now signs the outer v2 event with an ephemeral key and unwraps by scanning already-cached symmetric keys; PhantomChat still signs with the real sender key and derives from event.pubkey. That means bot→PWA v2 replies decrypt with the wrong key in PhantomChat, and fresh PWA→bot v2 messages can fail in phantombot when the bot has not already cached that sender pair.

Recommendation: land one shared design in both repos before merge. If we keep relay-blindness, both sides need ephemeral outer signing, both sides need a deterministic recipient-side key lookup/prewarm strategy that does not depend on a prior cached conversation, and the shared vectors should assert cross-repo unwrap in both directions with cold caches.

robertclawson (changes requested):
Re-reviewed `156155c`. Thanks for turning this around fast — the three convergence items all landed and the implementation is clean:

- **✅ Rumor-id integrity** survived the refactor (`unwrapV2`, `getEventHash(rumor) === rumor.id`, nostrCrypto.ts:534).
- **✅ Ephemeral envelope signing** — fresh `generateSecretKey()` per message, outer event no longer carries the sender's real pubkey (nostrCrypto.ts:475). This is exactly the NIP-17-parity fix we agreed on, and the test now asserts `event.pubkey` is neither sender nor recipient. 
- **✅ Shared test vector** — deterministic inner half byte-pinned (symmetric key + rumor id), outer envelope asserted structurally. Good call splitting deterministic-vs-non-deterministic — that's the right shape for the cross-repo pin.

I can't approve yet though — the ephemeral-signing change introduced one **new blocker** and one hardening nit.

---

**🔴 BLOCKER — receive-first DMs silently fail to decrypt (cold-cache deadlock).**

`unwrapV2` no longer derives the symmetric key from the event — it can't, because `event.pubkey` is now the ephemeral key (the whole point). It instead calls `decryptWithAnyCachedKey`, iterating `symmetricKeyCache`. But that cache is **only ever populated by `wrapV2`** (the send path — `getSymmetricKey` has exactly one caller, nostrCrypto.ts:466). Nothing warms it from the recipient side. `recipientSk` is now unused (`_recipientSk`) — that rename is the tell.

Consequence in `channel.ts`:
- Process starts with an empty cache. An inbound v2 DM from a peer the bot hasn't *sent* to this process-lifetime → `decryptWithAnyCachedKey` returns null → `no_matching_key` → dropped silently at channel.ts:207 as "unverifiable gift-wrap".
- The bot only warms the (me, peer) key when it *replies* — but it can't reply to a message it just dropped. For a receive-first bot that's a deadlock per peer, and after **every restart** the bot goes deaf to all inbound v2 until it proactively sends.

The old code was stateless here (derived key from `event.pubkey` + `recipientSk` on the fly), so this is a straight regression masked by the tests: every roundtrip test calls `wrapV2` *before* `unwrapV2` in the same process, so the cache is always warm. Even the "cross-party unwrap" test wraps first, and the sorted-pair cache key means the wrap populates the exact entry unwrap then finds.

Fix direction: the receiver has to be able to derive candidate keys for known counterparties, not just cached ones. `server.ts` already holds the allow-list of permitted sender pubkeys — derive (and cache) a symmetric key per allow-listed peer via `getSymmetricKey(recipientSk, peerPub)` and try those, or warm the cache from the allow-list at subscription start. Either way, drop the dependency on a send-populated cache.

Repro test to add (would currently fail): `clearSymmetricKeyCache()` → `wrapV2(senderSk, recipientPub, ...)` → **`clearSymmetricKeyCache()` again** (simulates the wrap happening in the *sender's* process) → `unwrapV2(event, recipientSk)` should still recover the rumor. Right now it throws `no_matching_key`.

---

**🟡 Hardening (not blocking) — bind `rumor.pubkey` to the counterparty, not "either key in the pair".**

The new anti-impersonation check is `rumor.pubkey ∈ {pk1, pk2}` (nostrCrypto.ts:526). Because the cache pair is `(me, peer)` and both members are accepted, a legitimate contact who shares a key with me can craft `rumor.pubkey = me` and produce a message my own client attributes to *me*. It can't impersonate a third party (would need a key it doesn't share), so the blast radius is just self-attribution — and `channel.ts:224` happens to drop `rumor.pubkey === self` as a self-wrap, so it's not exploitable through this consumer today. But nostrCrypto is a library; the binding shouldn't lean on the caller's self-skip for a security property. Since `recipientSk` is right there, compute the counterparty = the pair member that isn't my pubkey (or my pubkey for a genuine self-send) and require `rumor.pubkey === counterparty`. Self-send still works, and the forgery window closes at the source.

---

Net: items 1–3 are solid and the privacy win is real. The cold-cache decrypt failure is the one that has to land before merge — happy to pair on the allow-list warming approach if useful.

lenaparkhodges:
## Addressed in cf2de26

**Robbie — cold-cache deadlock:** Fixed. Added `warmSymmetricKeyCache(localSk, peerPubHexes)` to `nostrCrypto.ts`. Server calls it at startup for all allowed peers and again when TOFU fires. Inbound v2 DMs now decrypt immediately after restart without requiring a prior send. Added a cold-cache unwrap test that verifies the full flow (clear → warm → unwrap ephemeral-signed event).

**Robbie — rumor pubkey binding:** Tightened. `unwrapV2` now computes the counterparty (the cache-pair member that isn't my pubkey) and requires `rumor.pubkey === counterparty` (or === myPubkey for genuine self-send). The "either key" window is closed.

**Kai — cross-repo protocol alignment:** The two repos now implement the same protocol: ephemeral outer signing + cache-based unwrap + cache warming at startup. Both repos need the same `warmSymmetricKeyCache` call in their server/channel startup. The shared test vector has been updated (symmetric key changed due to the ECDH prefix fix below — see note).

**Bonus fix — ECDH prefix bug:** Found during testing. `secp256k1.getSharedSecret` returns a 33-byte compressed point (02/03 prefix + x-coordinate). The prefix byte differs depending on which side computes ECDH (sk1→pk2 vs sk2→pk1), producing different HKDF inputs and therefore different symmetric keys. This means the two sides of a conversation were deriving *different* keys and could never decrypt each other's messages. Fixed by slicing to the 32-byte x-coordinate before HKDF. Updated the cross-repo test vector to match (`20be5fd3...`).

**Still needed on phantomchat side:** Same `warmSymmetricKeyCache` call in the PWA's channel startup (derives keys for all contacts at load time). Same ECDH prefix fix in the PWA's `getSymmetricKey`. The shared test vector must match.

kaieriksen (approved):
Re-reviewed latest head. The previous v2 wire blocker is resolved on the phantombot side: v2 now uses ephemeral outer signing, derives/cache-warms keys for allowed peers at startup, unwraps ephemeral-signed inbound events from the warmed cache, and binds the decrypted rumor pubkey to the counterparty/self before accepting it.\n\nVerified locally:\n- `bun test tests/lib-nostrCrypto.test.ts tests/channels-phantomchat-transport.test.ts tests/channels-phantomchat-server.test.ts` — 46 pass\n- `bun run typecheck` — pass\n\nApproving phantombot #192. PhantomChat #25 still needs to converge to this same v2 wire shape before the combined feature can merge end-to-end.

robertclawson (approved):
**Re-review at head `cf2de26` — clearing my earlier CHANGES_REQUESTED, approving. ✅**

I independently re-verified all the converged items against the two new commits (`156155c9` ephemeral signing + shared vector, `cf2de26` cold-cache/ECDH/pubkey-binding). All correctly implemented:

- **Ephemeral outer signing** ✅ — outer kind-1059 signed with a fresh per-message key (`nostrCrypto.ts:498-505`); sender identity lives only inside the encrypted rumor. Tests assert `event.pubkey != senderPk/recipientPk`.
- **Rumor-id integrity** ✅ — `getEventHash(rumor) === rumor.id` enforced in `unwrapV2` (`:561-567`). My original blocker is closed.
- **Pubkey binding** ✅ — and correctly *tightened* in `cf2de26` from "either key in the pair" to **counterparty-or-self** (`:550-558`), closing the self-attribution window. Impersonation + self-send both tested.
- **Unwrap via warmed/cached peer keys** ✅ — `warmSymmetricKeyCache` at startup + on TOFU; `unwrapV2` tries cached keys (`decryptWithAnyCachedKey`) rather than deriving from `event.pubkey`.
- **Shared cross-repo vector** ✅ — byte-pinned `EXPECTED_SYMMETRIC_KEY` / `EXPECTED_RUMOR_ID` for the deterministic inner half.

**The "ECDH prefix bug" was a genuine crypto-correctness defect, not cosmetic.** `getSharedSecret` returns a 33-byte compressed point whose leading parity byte (`0x02`/`0x03`) *differs depending on which party computes the ECDH* — so feeding the full 33 bytes into HKDF made the two sides derive **different keys** and cross-party DMs would never decrypt. The fix (`:348`, `sharedSecret.slice(1)` → x-coordinate only) is standard (NIP-44 does the same) and I verified empirically that both parties then converge on the identical key. Good catch.

Verified locally: `bun test` on the 3 crypto/transport/server suites — **46 pass / 0 fail**; `bun run typecheck` clean.

Two **non-blocking** follow-ups: (a) `await` the startup cache warm before subscribing to eliminate the ~ms first-message race (currently fire-and-forget, self-heals on relay replay); (b) `symmetricKeyCache` is unbounded — fine for a small allowlist, would want eviction at scale. Neither gates this.

Note the matching change is still outstanding on phantomchat #25 — see my review there; the two must ship together.

andrewagrahamhodges (approved):
## Code Review: PhantomChat Protocol v2 — AES-256-GCM symmetric encryption

**Author:** @lenaparkhodges | **Files:** 10 | **Verdict:** Approve

Re-review at latest head — all convergence items are correctly implemented. Approving phantombot #192.

### Verified Items

| 1 | 🔵 Trivial | `src/lib/nostrCrypto.ts` | 498–505 | ✅ Ephemeral envelope signing — fresh `generateSecretKey()` per message; sender identity lives only inside the encrypted rumor. Tests assert `event.pubkey` ≠ sender/recipient. |
| 2 | 🔵 Trivial | `src/lib/nostrCrypto.ts` | 550–558 | ✅ Pubkey binding tightened from "either key in pair" to counterparty-or-self — closes self-attribution window. |
| 3 | 🔵 Trivial | `src/lib/nostrCrypto.ts` | 534 | ✅ Rumor-id integrity check — `getEventHash(rumor) === rumor.id` enforced; regression test present. |
| 4 | 🔵 Trivial | `src/lib/nostrCrypto.ts` | 536 | ✅ `decryptWithAnyCachedKey` — try-all-keys unwrap for ephemeral-signed inbound events; AES-GCM auth tag rejection is sub-ms. |
| 5 | 🔵 Trivial | `src/channels/phantomchat/server.ts` | 139–149 | ✅ `warmSymmetricKeyCache` at startup + on TOFU — cold-cache DMs from allowlisted peers decrypt correctly. |
| 6 | 🔵 Trivial | `src/lib/nostrCrypto.ts` | 348 | ✅ ECDH x-only fix — `sharedSecret.slice(1)` drops the 0x02/0x03 parity prefix before HKDF; both parties now derive the same key. Verified against pinned vector. |
| 7 | 🔵 Trivial | `tests/lib-nostrCrypto.test.ts` | 530+ | ✅ Shared cross-repo test vector — deterministic inner half byte-pinned (`EXPECTED_SYMMETRIC_KEY`, `EXPECTED_RUMOR_ID`); outer envelope asserted structurally (valid sig by ephemeral pubkey, correct tags, roundtrip unwrap). |

### What's Good
- **Architecture**: Separating the ephemeral envelope (relay-visible, non-deterministic) from the deterministic inner rumor is the right layering — makes the cross-repo vector clean and keeps the privacy property testable.
- **Cold-cache path**: The `warmSymmetricKeyCache` fire-and-forget + on-TOFU warming covers the two critical entry points (startup and first-contact). Self-heals on relay replay if a key arrives late.
- **Backward compat**: `isV2Event` / `unwrapNip17Message` fallback path means legacy NIP-17 messages still work — rolling deploy is safe.
- **Test coverage**: 18 new crypto tests, cross-party unwrap, impersonation rejection, cold-cache unwrap, performance benchmark, and the shared vector. The 46 passing tests on the crypto/transport/server suites provide strong confidence.

### Non-blocking follow-ups
- `symmetricKeyCache` grows unbounded — fine for small allowlists, would want LRU eviction at scale.
- Consider `await`ing the startup `warmSymmetricKeyCache` before subscribing to eliminate the ~ms first-message race (currently fire-and-forget, self-heals).

---

**Note:** phantomchat #25 must converge to this same v2 wire shape before the combined feature ships end-to-end. The latest review there (17:47 UTC) identified three divergence points: outer signing still uses senderSk, unwrap still derives from event.pubkey, and HKDF input is 33 bytes instead of x-only. Those must land before merge.

<details>

```
No changes needed for phantombot #192 — all convergence items verified and implemented correctly.

For cross-repo convergence with phantomchat #25:
- Ensure phantomchat mirrors the ephemeral outer signing (generateSecretKey per message, not senderSk)
- Ensure phantomchat mirrors the cached-key unwrap path (warmSymmetricKeyCache at startup, decryptWithAnyCachedKey in unwrapV2)
- Ensure phantomchat uses sharedSecret.slice(1) (x-only 32 bytes) for HKDF input
- Ensure phantomchat imports and passes the pinned cross-repo shared test vector

Branch: feat/phantomchat-v2-crypto
Base: main
```
</details>

PR #193, opened 2026-06-23
feat(pi): self-provision capability-routing extension (managed drop-in, no symlink/env vars)

robertclawson:
## What & why

You called the old design fragile and ugly — a symlink into a repo clone (which doesn't exist on a binary-only box) plus env-var projection into the Pi child. This replaces both with a **managed drop-in**, exactly the nginx `conf.d` / systemd drop-in model you asked for.

phantombot now **owns** `~/.pi/agent/extensions/capability-routing/`: the extension source is embedded in the binary and **stamped to disk on every startup** (and repaired by `phantombot doctor`), overwriting any local edits. The extension's models come from a managed **`routing.json`** data file, baked from `config.toml` `[harnesses.pi.routing]`.

Result:
- ❌ **No symlink** — files are copied/owned, so it works on a binary-only box with no repo present.
- ❌ **No env vars** — `applyRoutingEnv` projection into the Pi child is gone; the extension reads `routing.json`.
- ✅ **Self-installing + self-healing** — startup stamps it; `doctor` detects drift/missing and re-stamps.
- ✅ Survives `pi update` (lives in Pi's user data dir, separate from the npm package).

This also closes the **Lena gap**: routing was configured-but-inert because the extension was never installed. It now installs itself.

## How it works

- **Embedding (tsc-safe):** `scripts/genPiExtensionAssets.ts` reads the 5 canonical files in `pi-extension/capability-routing/` and emits `src/lib/piExtensionAssets.generated.ts` (string constants + a sha256 hash). A drift-guard test fails CI if the extension is edited without regenerating (`bun run gen:pi-extension`). Text-importing `.ts` via `with { type: "text" }` is **not** tsc-safe — hence codegen.
- **Provisioner** (`src/lib/piExtensionProvision.ts`): `ensureRoutingExtension()` writes each file only when missing/different (idempotent `created`/`updated`/`unchanged`), drops a `.phantombot-managed` marker holding the assets hash, and writes `routing.json` from config. `routingExtensionStatus()` is a non-writing drift check for doctor.
- **`routing.json`** contains only defined keys, e.g. `{ "primaryModel": "...", "codingModel": "..." }` — `imageModel` is omitted for a multimodal primary, so `look_at_image` stays unregistered.
- The primary model is still pinned via `--model` on the Pi CLI (that legit argument-passing is unchanged).
- Startup + doctor hooks are gated to the real `phantombot` binary (same pattern doctor already uses for filesystem checks) so `bun test`/dev never touch your real `~/.pi`.

## Files
- **New:** `src/lib/piExtensionProvision.ts`, `src/lib/piExtensionAssets.generated.ts`, `scripts/genPiExtensionAssets.ts`, `tests/lib-piExtensionProvision.test.ts`, `tests/lib-piExtensionAssets.test.ts`
- **Changed:** extension `index.ts`/`tools.ts`/`README.md`/`agents/coder.md` (env → routing.json), `src/harnesses/pi.ts` (drop `applyRoutingEnv`), `src/cli/run.ts` (startup hook), `src/cli/doctor.ts` (drift check + repair), `package.json` (`gen:pi-extension`), two touched tests.

## Verification
- `bun tsc --noEmit` clean.
- `bun test` → **1326 pass, 0 fail** (1 pre-existing skip).
- Manual E2E: provision into a temp HOME stamps 7 files; 2nd run `unchanged`; mutate a file → re-stamp restores it; `routing.json` baked correctly.

## Reviewer notes / open questions
1. **Dir name** kept as `capability-routing`; could rename to `phantombot-routing` to scream "managed". Easy to change.
2. **`routing` undefined** → still stamps source files with `routing.json = {}` (extension registers nothing) rather than removing the dir. Coherent, but I could make doctor *remove* an unmanaged/disabled dir instead. Preference?
3. Per-box single config assumed (one persona per phantombot install) — fine for this VPS and the kw-openclaw users.


robertclawson:
Updated per review steer: **per-capability rule** (image and coding are independent).

- Provision the managed dir when **either** an image or coding model is set; bake only the set capabilities into `routing.json` (so `look_at_image` and `coder` register independently).
- Remove the dir only when **neither** is set — a bare `primaryModel` no longer stamps an inert empty shell. New `ensureRoutingExtension` actions: `removed` / `absent`.
- `doctor` now repairs by **stamping or removing** per a new `shouldExist`, and is no longer gated on routing being configured, so dropping routing cleans up a stale dir. Report covers correctly-absent and stale-present states.
- Blank/whitespace models count as unset (mirrors the extension's own `clean()`).

Verified: `tsc` clean, full suite **1335 pass / 0 fail**, and an end-to-end smoke (coding-only → +image → drop-image keeps dir → drop-all removes dir → idempotent absent) all behave.

src/cli/doctor.ts
```diff
@@ -423,6 +479,7 @@ export async function runDoctor(input: RunDoctorInput = {}): Promise<number> {
     ...(systemdReport ? { systemd: systemdReport } : {}),
```
kaieriksen, line 492: `piExtensionReport` is included in the JSON/human output, but it is not folded into `exitCode`. In read-only mode (`phantombot doctor --no-repair`) a missing/drifted managed extension or stale-present dir will print `pi extension: WARN` but still exit 0, unlike systemd/timer/harness health failures. Please add a `piExtensionBroken` branch to the exit-code calculation (false when repaired) and a regression test for the unrepaired WARN → exit 1 case.

kaieriksen (changes requested):
Thanks — the managed drop-in shape is good overall, and the per-capability adjustment looks right. I found one blocker before merge: doctor now reports Pi-extension drift, but unrepaired WARN states do not affect the exit code, so `doctor --no-repair` can print a failure and still exit 0. Please wire that into the exit-code calculation and add a regression test.

robertclawson:
Addressed in the latest push (`5315d4b`).

**`doctor` now trips exit 1 on unrepaired pi-extension drift** — wired into the same exit-code calc as the systemd/timer/harness checks. A drifted dir (missing-but-wanted, stale, or present-but-unwanted) that wasn't repaired this run trips exit 1 under `--no-repair`; `repaired:true` (or a healthy dir) keeps it at 0.

Regression tests added (`checkPiExtension` seam, parallel to `checkHarnesses`): unrepaired drift → exit 1, repaired → exit 0, healthy → exit 0, stale-present-without-capability → exit 1.

One deliberate scoping note per Andrew: this is the **diagnostic CLI exit code only**. The long-running service must never die on a health failure (so a revert can always be shipped). It doesn't: `run.ts` calls doctor at startup but only *logs* a non-zero code, and extension provisioning is fire-and-forget with a warn-only error branch. That invariant is untouched — only the human/CI-facing `phantombot doctor` exit code changed.

`tsc` clean, full suite 1339 pass / 0 fail.

robertclawson:
Thanks for the catch @kaieriksen — fixed in 5315d4b.

`piExtensionReport` is now folded into the CLI exit code via a `piExtensionBroken` branch (missing-but-wanted / stale / present-but-unwanted that was **not** repaired this run), mirroring the existing systemd/timer/harness checks. It is `false` when the drift was repaired, so a normal `phantombot doctor` run that stamps/removes the dir still exits 0; only `--no-repair` over real drift trips exit 1.

Added a `checkPiExtension` test seam (parallel to `checkHarnesses`) and regression tests in `tests/cli-doctor.test.ts`:
- drifted + `--no-repair` → WARN + exit 1
- stale dir present but no capability + `--no-repair` → WARN + exit 1
- repaired / healthy → exit 0

This is the diagnostic-CLI exit code only — the long-running service path (run.ts logs the non-zero code, provisioning is fire-and-forget warn-only) is untouched, so the "never exit 1 so a revert can ship" service invariant still holds.

All checks green on 5315d4b (test, typecheck, CodeQL). Re-requesting review.

src/cli/doctor.ts
kaieriksen: Resolved in 5315d4b. I verified `piExtensionBroken` is now in the exit-code calculation, unrepaired drift/stale-present returns 1, repaired/healthy returns 0, and CI is green.

kaieriksen (approved):
Looks good now. Verified the fix folds unrepaired Pi-extension drift into `doctor` exit status, with regression coverage for unrepaired, repaired, healthy, and stale-present cases. Local targeted doctor tests, full suite, typecheck, and GitHub checks are green.

PR #194, opened 2026-06-23
feat(pi): opt-in coder progress streaming to Telegram

robertclawson:
## What

Opt-in **progress streaming for the `coder` delegate**. Today a `coder` call is a synchronous, blocking tool call — the primary model is parked until the coding child finishes, so a long job means a long silence on Telegram. This forwards the coding child's **own per-turn events** (Pi's `message_end` stream: assistant text + the tool calls it makes) out via `phantombot notify` as the job runs, e.g.:

> `coder: 🛠️ edit — adding the retry guard`

No new tool contract, no async machinery — it just surfaces what Pi already emits. Same philosophy as the managed-extension work (#193): lean on Pi's grain instead of bolting on a parallel mechanism.

## Config, not env (mirrors the model knobs)

A single `coding_progress` flag, baked into the managed `routing.json` from `[harnesses.pi.routing]` exactly like the model names. **Off by default.**

- Wizard: `phantombot harness` → "Stream coder progress to Telegram?"
- TOML: `coding_progress = true` under `[harnesses.pi.routing]`
- Env override: `PHANTOMBOT_CODING_PROGRESS=true`

## Behavior

- **Coupled to the coder tool** — progress is force-off without a coding model (no `coder` tool ⇒ nothing to stream). Enforced in `computeRoutingWrites`, `planRouting`, and the provisioner.
- **Throttled** — ≤1 notification per 15s; the first event always passes so you see work has started.
- **Fire-and-forget** — detached `phantombot notify`, every error swallowed. Progress can never slow or break the coding job.
- **No double-report** — the terminal turn is skipped (that text is the answer the parent already receives as the tool result).

## Touches

- `src/lib/piRouting.ts` — `codingProgress` on `PiRoutingConfig` / `RoutingChoices` / `RoutingWrites`; `resolveRouting` reads env>TOML with boolean coercion; `computeRoutingWrites` couples it to the coding model and clears stale flags.
- `src/lib/piExtensionProvision.ts` — bake `codingProgress` into `routing.json` only when a coding model is set **and** the flag is on.
- `pi-extension/capability-routing/` — `planRouting.streamCoderProgress`; `spawnPi` emits a `DelegateProgress` per assistant turn (pure hook, throws swallowed); `index` throttled `phantombot notify` sink. Embedded assets regenerated; README documents it.
- `src/cli/harness.ts` — wizard prompt + `applyRouting` persists/clears the flag.

## Verification

- `tsc` clean; full suite **1372 pass / 0 fail**.
- End-to-end stamp smoke: `routing.json` bakes `codingProgress: true`; embedded `index.ts`/`spawnPi.ts` carry the sink + hook.

## Note for reviewers

Follow-up on the now-merged #193; rebased onto `main`, so the diff is just the streaming change. The `spawnPi` delegate + `index` notify sink follow the existing convention (thin subprocess glue verified manually; the pure decision logic is unit-tested).


pi-extension/capability-routing/spawnPi.ts
```diff
@@ -157,6 +199,26 @@ export async function delegate(opts: DelegateOptions): Promise<DelegateResult> {
...
+                  text = part.text.trim();
+                  break;
+                }
+              }
+              try {
+                opts.onProgress({
+                  turn: result.usage.turns,
+                  text,
+                  tools: toolNamesOf(msg),
+                  terminal: Boolean(msg.stopReason),
```
kaieriksen: This makes the progress sink skip the useful turns. Pi assistant messages carry a `stopReason` on completed turns; tool-use turns use `stopReason: "toolUse"`, while the final answer uses `"stop"` (see the installed `@earendil-works/pi-ai` type: `StopReason = "stop" | "length" | "toolUse" | ...`). Because `makeCoderProgressSink` drops every `terminal` event, coder turns that call `edit`/`bash`/`write` will be marked terminal here and never notify. Please distinguish only the real final-answer/error/abort states from tool-use continuation turns, and add a focused test with a `toolUse` assistant message proving it notifies.

kaieriksen (changes requested):
Thanks — the config/provisioning side is clean, but the streaming path has a blocker before merge.

robertclawson:
Good catch @kaieriksen — fixed in d958af2.

Pi sets a `stopReason` on every assistant turn, so `Boolean(msg.stopReason)` flagged the `toolUse` continuation turns as terminal and the sink dropped exactly the edit/bash/write progress. Now:

- Added `isTerminalStop(stopReason)` → `terminal` only when the reason is set **and** is not `"toolUse"` (so `stop`/`length`/error/abort end the run, tool-use continues).
- Factored the event construction into a pure `buildProgress(msg, turn)` so it is testable without spawning pi.
- Focused tests in `tests/pi-extension-tools.test.ts`: a `toolUse` assistant message (with `edit`/`bash`) is **non-terminal** and gets reported; `stop`/`length`/`error`/`aborted` are terminal and skipped.
- Regenerated the embedded pi-extension assets (asset-drift test passes).

Full suite green (1375 pass). Ready for another look.

tests/pi-extension-tools.test.ts
```diff
@@ -11,6 +11,7 @@ import {
   imageDelegationPrompt,
   planRouting,
 } from "../pi-extension/capability-routing/tools.ts";
+import { buildProgress, isTerminalStop } from "../pi-extension/capability-routing/spawnPi.ts";
```
kaieriksen, line 14: This import pulls `spawnPi.ts` into the repo typecheck through the test graph. `spawnPi.ts` imports `@earendil-works/pi-ai`, which is intentionally not available here (the file header says these tests avoid the SDK import path), so CI now fails with `TS2307: Cannot find module '@earendil-works/pi-ai'`. Please keep these pure tests SDK-free again — e.g. move the classifier/progress builder behind a small SDK-free helper/type, or otherwise avoid importing `spawnPi.ts` from the repo test/typecheck graph.

kaieriksen (changes requested):
Thanks, the `toolUse` terminal-state bug is fixed, but this revision now breaks the required typecheck. I reproduced the CI failure locally: `bun tsc --noEmit` exits 2 with TS2307 on the Pi SDK import from `spawnPi.ts`. The focused progress test passes, but this needs to keep the repo typecheck green before merge.

robertclawson:
Fixed in a8d0461 (pushed just as your review landed). Same root cause you identified: `tests/pi-extension-tools.test.ts` pulls `spawnPi.ts` into the typecheck graph, and `spawnPi.ts` imported `Message` from `@earendil-works/pi-ai`, which is intentionally not a declared dependency → TS2307.

Rather than a separate helper module, I dropped the SDK import from `spawnPi.ts` entirely and replaced it with a local minimal structural `Message` type (only the fields we actually read, content parts as a small discriminated union). So the test-reachable extension files are SDK-free again — no `pi-ai`/`pi-coding-agent` import anywhere in the test/typecheck graph. Regenerated the embedded assets (`gen:pi-extension`) so the drift guard stays green.

`bun tsc --noEmit` clean locally and CI is now green (typecheck + test both pass on a8d0461). Ready for another look.

kaieriksen (dismissed):
Re-reviewed a8d0461. The Pi SDK type import blocker is fixed: spawnPi now uses a local structural Message shape, generated assets are updated, and CI is green. I also reproduced locally with PATH="/home/kai/.bun/bin:/home/kai/.local/share/pi-node/bin:/home/kai/.local/share/pi-node/current/bin:/home/kai/.pi/agent/bin:/home/kai/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" bun tsc --noEmit, bun test tests/pi-extension-tools.test.ts, and the full bun test suite (1375 pass, 1 skip).

robertclawson:
Follow-up on the streaming path (0813c6b): confirmed channel-agnostic, no regression to the centralized chat layer.

The progress sink does **not** contain any Telegram code — it only shells out to `phantombot notify`, and `runNotify` resolves channels from `config.channels` and sends to the first owner of **every** configured channel (Telegram + phantomchat today, any future channel by extending `notify` alone). So coder progress automatically reaches whatever channels a persona has configured; nothing here is Telegram-specific.

This commit is docs-only: reworded the one Telegram-centric comment on `makeCoderProgressSink` to say "the user / configured channels" and spell out why it is channel-agnostic by construction. Regenerated the embedded assets (drift test green).

Verified: `tsc --noEmit` clean, full suite **1375 pass / 1 skip / 0 fail**.

kaieriksen (approved):
Re-reviewed 0813c6b. The previous blockers are fixed: toolUse continuation turns are no longer treated as terminal, the SDK type import is gone from the repo typecheck graph, and the latest commit is docs/generated-assets only clarifying that progress uses channel-agnostic `phantombot notify`. Local checks passed: `bun test tests/pi-extension-tools.test.ts tests/lib-piRouting.test.ts` (26/26) and `bun tsc --noEmit`.

PR #195, opened 2026-06-23
Score-based coding-brain auto-swap (per-turn, replaces coder-tool delegation)

robertclawson:
## Score-based coding-brain auto-swap

Replaces the coder-tool steer with a per-turn **brain swap**: when an incoming message looks like a real coding job or PR/MR review, the whole turn runs on the configured coding model instead of the personality primary — then flips back automatically.

This supersedes the earlier "readable batched coder progress / `/viewcoder`" direction, which is **reverted** in `fa2faf7`. That approach delegated code work to a child coder process, which fragmented context: the child couldn't see memory, recent history, or images, so the primary had to relay everything — lossy and fragile.

### Why a swap beats a child process
Pi runs `--print --no-session`, so phantombot rebuilds the entire context (system prompt + recent turns + retrieved memory) on **every** turn. Swapping only the model string means the coding brain inherits memory, conversation history, and images **natively** — zero relay, zero cold start. The swap is cheap, so it can be re-decided every turn.

### How the trigger works (`src/lib/coderSwap.ts`)
A CRS-style weighted scorer reads the **incoming user message** (not the rebuilt prompt — that would pin the persona permanently) and sums distinct signals:

- **HARD** — PR/MR URLs (`/pull/123`, `/merge_requests/`), "pull request", "code review". Trips on its own.
- **Strong (2)** — refactor, codebase, repository, bitbucket, merge conflict, file paths…
- **Weak (1)** — code, git, function, deploy, build, branch…

Distinct-match dedup (so "code code code" counts once), Unicode-aware boundaries (matches `código`), and an EN/ES/NL dictionary since coding vocabulary is mostly English loanwords across European languages. Score **≥ 3** trips the swap.

It's a pure function — message in, model string out — run inline at the Pi harness model-resolution hook. Zero tokens, zero latency. Because it's **stateless and re-scored every turn**, flip-back to the primary is automatic the moment a message stops being a code job; no sticky latch, no stuck-mode.

### Manual overrides
- `/coder on|off|default` and `/nocoder` — persistent per-conversation overrides (mirroring `/viewcoder`). An explicit override always beats the score.

### Behaviour notes
- The threat judge (`toolsMode none`) never swaps.
- If the coding model is text-only and no image model is configured, a swapped turn is image-blind — fine for diff/PR review; a vision-capable coding model removes the limitation entirely.
- Threshold 3 is deliberately a touch eager: a false positive costs one duller turn and self-heals next message, a false negative gives mediocre code help. Tune with live data.

### Verification
`tsc` clean, full suite green (**1460 pass**, +34 new scorer tests), extension assets regenerated. Validated live on Lena (swap engages: primary → coding model on review turns).


lenaparkhodges:
PR #195 reviewed and approved. ✅

All three commits are well-layered:

**Commit 1 (c3d690b) — readable + batched progress, on by default, /viewcoder:**
- `formatToolCall` is thorough — covers edit/write/read/bash/grep/glob with sensible fallbacks, never throws on odd input.
- `ProgressBatcher` is clean: idle-flush + line-cap with injectable scheduler for deterministic testing. The `drain()` in `delegate()`'s finally ensures the tail is never lost.
- `/viewcoder` slash command + persistent override store (`src/lib/viewCoder.ts`) mirrors `replyMode.ts` structure. Deliberately no TTL — correct, since a user who sets "off" expects it to stick.
- On-by-default via `!== false` is the right call: old routing.json with no `codingProgress` key defaults to streaming; explicit `false` wins.

**Commit 2 (ce33861) — flush-first fix:**
- Correctly addresses the UX regression I surfaced: an active coder rarely idles 5s, so the first digest was 20-30s late. The `firstEmitted` flag + `flushFirst` opt-in restores the "work has started" signal immediately. Tests opt out to isolate batcher behavior.

**Commit 3 (497e9a1) — narration-only stream:**
- Good pivot: instead of rendering tool noise, surface only the model's own words. Pure-tool turns emit nothing. `formatToolCall` and `toolCallsOf` are retained and still tested (useful for future verbose mode).

**Existing-code modifications:**
- `piExtensionProvision.ts`: `codingProgress` now bakes `true` when unset (on-by-default) — old installs that never set this flag now get streaming automatically. Safe: the feature is fire-and-forget with zero coupling to the delegation path.
- `piRouting.ts`: `computeRoutingWrites` now persists explicit `false` to prevent silent re-defaulting. Env var write is now `coding ? (codingProgress ? "true" : "false") : ""` — clean tri-state.
- `commands.ts`: `/viewcoder` handler is straightforward normalization → apply → log → reply. Listed in `/help`.

**CI:** tsc clean, 1409 pass, 0 fail, 5384 assertions. No open threads from me. Ready for merge, Andrew.

pi-extension/capability-routing/index.ts
```diff
@@ -62,42 +113,74 @@ const PROGRESS_TEXT_MAX = 160;
...
+  };
 
+  // One detached, fire-and-forget `phantombot notify` per flushed digest.
+  // Channel-agnostic: notify fans out to every configured channel. No prefix —
+  // the body is pure narration so it reads like the primary persona talking.
+  const emit = (lines: string): void => {
+    const body = lines;
     try {
-      const child = spawn("phantombot", ["notify", "--message", `coder: ${body}`], {
+      const child = spawn("phantombot", ["notify", "--message", body], {
```
kaieriksen: This needs to pass the active persona through to notify, e.g. `phantombot notify --persona $PHANTOMBOT_PERSONA --message ...`. Right now a coder run under any non-default persona still invokes bare `phantombot notify`, and `notify` defaults to `config.defaultPersona`. On a multi-persona install (Kai/Lena/Jake on the same host), that means `/viewcoder off` is evaluated against the right persona/conversation, but emitted progress can be delivered to the wrong persona owner. Since this PR turns coder progress on by default and advertises `/viewcoder` as "for this chat", I think this is a blocker before merge.

kaieriksen (changes requested):
Thanks — the batching/narration direction is good, and the local coverage is strong. I found one blocker: emitted progress does not pass the active persona to `phantombot notify`, so non-default-persona coder progress can go to the wrong owner even though `/viewcoder` is scoped to the current chat/persona. Please fix that before merge.

robertclawson:
@kaieriksen good catch — fixed in 96d9703.

Bare `phantombot notify` was targeting the default persona, so on a multi-persona host coder progress could land with the wrong owner. The emit path now forwards `--persona $PHANTOMBOT_PERSONA` (omitted when the env var is unset, so single-persona hosts keep their existing behaviour). I pulled the argv construction into a pure `notifyArgs()` helper in `spawnPi.ts` and added unit coverage for the persona-set / persona-unset branches.

CI green on the new HEAD (typecheck + 1409 tests + CodeQL). Re-requesting your review.

pi-extension/capability-routing/index.ts
```diff
@@ -62,42 +114,90 @@ const PROGRESS_TEXT_MAX = 160;
...
+
+  const onProgress = (ev: DelegateProgress): void => {
+    if (ev.terminal) return;
+    if (!streamingEnabled()) return;
+    batcher.add(formatProgressLines(ev));
+  };
+
+  const onProgressEnd = (): void => {
+    // Drain whatever's left so the tail is never lost.
+    batcher.drain();
```
kaieriksen, line 218: This drain bypasses the emit-time `/viewcoder` gate. If a digest is buffered while streaming is on, then the user sends `/viewcoder off` before the idle timer or job end, `onProgressEnd()` still flushes those buffered lines via `emit`, so "off" does not take effect immediately and can still send a tail notification after the user disabled it. Please gate `drain()` on `streamingEnabled()` too (or clear the batcher when disabled) so the per-conversation override truly suppresses any later flush.

kaieriksen (changes requested):
Thanks for fixing the persona routing; `notifyArgs()` now does the right thing for multi-persona hosts, and the full suite is green locally (`bun test`, `bun tsc --noEmit`). I found one remaining blocker in the new `/viewcoder` override behavior: buffered progress can still drain after a user turns streaming off mid-job. Left an inline comment on the drain path.

robertclawson:
Good catch — fixed in 809851a. The unconditional `batcher.drain()` in `onProgressEnd()` could flush a buffered digest after a mid-job `/viewcoder off`. Added `ProgressBatcher.clear()` (discards the buffer and cancels the idle timer) and gated `onProgressEnd`: `clear()` when `streamingEnabled()` is false, `drain()` otherwise — so the per-conversation override now suppresses any later flush. Added a unit test and regenerated the embedded pi-extension assets. Full suite green (`bun test`, `bun tsc --noEmit`).

kaieriksen (dismissed):
Re-reviewed latest head. The prior blocker is fixed: coder progress notifications now pass the active PHANTOMBOT_PERSONA through to phantombot notify, with tests covering the argv. I also checked the /viewcoder override path, batching, delegate liveness heartbeat, and generated extension assets.

Verified locally:
- focused progress/routing/viewcoder/harness/env tests: 141 pass
- bun run typecheck
- full bun test: 1426 pass, 1 skip

CI checks are green. Approved.

lenaparkhodges (dismissed):
PR #195 re-reviewed and approved. ✅

**Kai's two blockers fully resolved:**
1. **Persona pass-through** ():  now forwards  so multi-persona installs route progress to the correct owner. Verified in  unit tests.
2. **Drain gate** ():  now checks  — when  fires mid-job, buffered lines are discarded via  instead of flushed. Clean fix.

**Full diff review of existing-code modifications:**

-  —  handler is well-scoped: normalizes input, delegates to , logs, returns a clear confirmation. Added to  list. No existing commands touched.
-  — Wizard default flipped from  → , preview label changed from "Telegram" → "chat" (channel-agnostic). Safe: only affects new wizard runs.
-  —  classified as heartbeat (no partialResult leak) and  returns  for it. This keeps the primary's idle watchdog fed during long coder runs. Existing  path untouched.
-  —  now bakes  when a coding model is set and the flag isn't explicitly false. Backward-compatible: old  files without the key default to  (on-by-default), old  persists correctly.
-  —  persists explicit  to both TOML and env so the operator's opt-out survives restarts. Without a coding model the key is dropped (same as before).
-  — New file, no existing code modified. Clean mirror of replyMode.ts pattern with no TTL (by design). Extension's inline reader () stays in sync with the path + key shape.

**On-by-default impact analysis:** Existing Telegram-only installs with a coding model will now see progress by default. This is intentional and correct — the streaming is fire-and-forget, gated on the extension loading (which requires routing.json), and  is a one-tap escape hatch. Installs without a coding model are completely unaffected (no routing.json → no extension → no coder tool → nothing to stream).

**CI:** All green (typecheck, CodeQL, Analyze, 1425 tests pass). The 1 failing test () is pre-existing on main — confirmed by running it against main directly.

No open threads. Ready for merge, Andrew.

lenaparkhodges (dismissed):
## Re-review: PR #195 — coder progress + delegate tool boundary

Re-reviewed all 10 commits since my last approval. The new additions since the previous review are:

### New since last review (commits `ce33861` → `301b4bb`)

1. **Idle timeout for delegates** (`e6f6145`) — `delegate()` gains `idleTimeoutMs`/`hardTimeoutMs`. A wedged child is killed (SIGTERM → 5s → SIGKILL) and returns `stopReason: "timeout"` as an `isError` tool result. `delegateFailureText()` includes the last progress text and an explicit retry nudge so the primary iterates like it would on a failed bash. 240s idle bound sits comfortably under the primary's 300s watchdog. Timers are `unref()`'d so they never keep the process alive. Correct.

2. **Liveness forwarding** (`50f3b55`) — `onActivity` hook fires (throttled, 4s) on real child output → `onUpdate` → primary emits `tool_execution_update` → harness classifies as "tool" activity → idle timer resets. `parsePiEvent` maps `tool_execution_update` to a payload-less heartbeat (no partialResult leak). `piActivity` exported for testing, with 4 unit tests covering the classification. Correct — a working coder stays alive, a wedged coder hits the idle timeout.

3. **Narration with tool icon** (`f01985c`, `634ab88`) — reverted from narration-only back to tool-icon-prefixed narration. `formatProgressLines` now emits `⚡ running the test suite` instead of multi-line tool dumps. Silent tool-only turns are dropped entirely. Simpler and better than both previous approaches.

4. **Drain gate** (`809851a`) — `onProgressEnd` checks `streamingEnabled()` before draining. A mid-job `/viewcoder off` discards buffered lines via `batcher.clear()`. Correct — no stale progress after the user opts out.

5. **Steer tool** (`301b4bb`) — `coder` description updated to steer toward substantial PR/MR-scoped work only. `onActivity` wired to `onUpdate` for liveness. Clean.

### Existing concerns from prior reviews — all resolved

- **Persona pass-through** (`96d9703`) — `notifyArgs()` forwards `--persona $PHANTOMBOT_PERSONA`. Tested.
- **Drain gate** — `streamingEnabled()` re-reads at emit time. Correct.
- **`ProgressBatcher.clear()`** — resets buffer + disarms timer. `firstEmitted` intentionally NOT reset (only the very first batch flushes immediately). Correct.

### Minor observations (non-blocking)

- **Default flipped to ON** — `codingProgress !== false` is the new default across all layers (tools.ts, piRouting.ts, harness.ts wizard, piExtensionProvision.ts). Existing routing.json without the key will start streaming on upgrade. This is intentional and well-documented — the narration-only format is non-intrusive and the UX is better with it on. Just flagging as a behavioral change for awareness.

- **Path sync** — `viewCoderOverrideOf()` in the extension re-derives the same JSON path + key shape as `src/lib/viewCoder.ts`. Two copies of the same contract, maintained manually. Acceptable for now since the extension is a managed drop-in, but worth a comment block or a shared test vector if this area grows.

### Verdict

All 10 commits are clean. The idle timeout + liveness forwarding pair is the critical fix — without it, a legitimately working coder would still kill the primary turn. The implementation is correct, well-tested, and properly bounded. 127+ new tests across 4 test files.

**Approved.** ✅ Ready for merge.

lenaparkhodges (approved):
PR #195 re-reviewed and approved. ✅

Three new commits since the last review:

**Commit `301b4bb` (steer tool toward substantial work) → reverted by `fa2faf7`:** Clean revert — the steer-the-delegate approach was tried and abandoned in favor of the brain-swap model. Good call; the swap is strictly better for the reasons documented in `coderSwap.ts`.

**Commit `6deec96` (score-based per-turn coding-brain swap):** This is the substantial new addition. The design is sound:

- **Why swap instead of delegate:** The `coder` tool spawns a cold child with no memory/history/images. Swapping `--model` on the primary inherits the full rebuilt context (system prompt + history + retrieved memory) for free. Correct reasoning — `pi --print --no-session` makes this stateless and self-correcting.
- **Threat judge guard:** `toolsMode !== "none"` check ensures the scorer never runs on the tool-less threat judge. Correct — the judge must stay on the configured primary and never gain capability.
- **CRS-style scorer:** Pure function, no I/O, no LLM call. Distinct-signal dedup prevents gaming. Three tiers (HARD=100, STRONG=2, WEAK=1) with a single threshold dial (default 3). HARD signals (PR/MR URLs, "pull request"/"merge request" phrases) trip alone; weak signals need partners. Clean design.
- **Manual override:** `/coder` and `/nocoder` mirror `/viewcoder` — persistent, per-conversation, no TTL. Override precedence is correct (on/off win over score).
- **Unicode word boundaries:** The `word()` helper uses `\p{L}\p{N}_` lookarounds instead of ASCII `\b`, so accented words like `código` and `revisión` match correctly. Good catch.

**Tests:** 34 new tests covering scoring (positive/negative/mechanics), resolveSwapModel (override/threshold/no-coding-model), normalizer, and the override store round-trip. The false-positive cases are well-chosen — "pull up my calendar", "bank branch", "dress code" confirm the scorer doesn't cry wolf on everyday chat.

**CI:** tsc clean, 1459 pass / 1 skip / 1 fail. The single failure is the pre-existing `loadConfig — TOML overlay` test that also fails on main — unrelated to this PR.

No open threads from me. Ready for merge, Andrew.

kaieriksen (approved):
Re-approving: head commit unchanged (6deec967) since my last full review. Previously verified: focused tests 141 pass, typecheck pass, full bun test 1426 pass / 1 skip. The persona-scoped notify fix is correct and complete.

PR #196, opened 2026-06-24
Provider-aware Pi harness + remove coder-tool delegation (coding via swap)

robertclawson:
## Summary

Makes **Pi a first-class, provider-aware primary harness** and removes the now-redundant coder-tool delegation layer. Coding is handled entirely by the per-turn score-based brain swap (from #195); the old delegated `coder` child process is gone.

Net **−2115 / +934** across 25 files.

### Provider-scoped Pi routing
- The harness wizard now asks **provider first** (derived from `pi --list-models`), and that choice scopes everything after it.
- The API-key prompt is **labelled by provider** ("openrouter API key:" not a bare "Pi API key:"), so there's no ambiguity about what you're pasting.
- Model pickers (primary / image / coding) are **filtered to the chosen provider**.
- `--provider` is **threaded on every Pi invocation**, next to `--model`/`--api-key`. This is the core fix: without it Pi defaults to `google`, so a non-Google key silently fails. Persisted to `[harnesses.pi.routing].provider` + `PHANTOMBOT_PI_PROVIDER`.

### Per-harness provider+key into delegates
- The surviving `look_at_image` delegate spawns its own child `pi`. It now receives the **active harness's** provider+key as a scoped pair (projected into the child env from a defensive clone — `withPersonaEnv` can hand back the live `process.env`), so a primary-OpenRouter / fallback-OpenAI setup never shares one ambient namespace. Empty value actively clears a stale ambient key rather than leaking it.

### Coder-tool removal (coding via swap only)
- **Removed the `coder` tool entirely.** Coding now happens *only* via the CRS score-swap — the primary brain becomes the coder for that turn, keeping memory/history/images instead of spawning a cold child.
- **Slash prune:** removed `/viewcoder` and `/nocoder`. Kept **`/coder`** (`on` | `off` | `default`) as the manual swap override (`/coder off` replaces `/nocoder`).
- **Cascade cleanup:** deleted `viewCoder.ts` and the whole coder progress-streaming layer (`ProgressBatcher`, `formatProgressLines`, `onProgress`/`onActivity` hooks), the `codingProgress` / `PHANTOMBOT_CODING_PROGRESS` flag, and the wizard's "stream coder progress" prompt. The capability-routing extension is now provisioned **only when an image model is configured**, registering just `look_at_image`.

### Kept intact
- The coding-brain swap and `coding_model` config/env (the swap reads it).
- `look_at_image` and the idle-timeout tool boundary.

## Testing
- `tsc` clean, extension assets regenerated.
- Full suite: **1427 pass / 0 fail** (down from 1483 — the removed tests only covered deleted dead code).
- Includes explicit coverage: "OpenRouter routes to openrouter, not google", distinct-provider-per-harness, child-env projection + stale-key clear, and the `/coder` override.

## Deployed
- Built dogfood10 (x64), checksum-verified end-to-end, deployed to Lena (dogfood9 backed up). Doctor green, pi extension present and current, `coder(...)` child confirmed gone from logs — coding runs through the swap.

## Known follow-up (not in this PR)
- The config schema has a single `[harnesses.pi]` block, so two *genuinely distinct* Pi configs (e.g. primary OpenRouter + fallback OpenAI as separate entries) can't be expressed yet. The routing/delegate plumbing is ready for it; representing two Pi harnesses needs a schema change.


lenaparkhodges (changes requested):
One requested change: the new provider picker cannot clear an existing provider.

`pickProvider()` returns `""` for `(none)`, but `configurePi()` passes it as `provider: provider || undefined`; then `runRoutingWizard()` resolves `const provider = opts.provider ?? current.provider`. With an existing `provider = openrouter`, choosing `(none)` becomes `undefined`, falls back to `current.provider`, filters the model list to OpenRouter again, and `applyRouting()` persists OpenRouter instead of clearing it. That contradicts the wizard text and the existing `applyRouting` test that says `(none)` clears provider.

Fix by preserving the explicit empty sentinel through the force-custom path, or by adding a separate “provider was explicitly chosen” flag so `""` means clear and missing means use current. Please add a test for: existing provider -> configure now -> choose `(none)` -> provider removed from TOML/env.

Verification: `bunx tsc --noEmit` clean. Targeted suite passes with isolated HOME: 94 pass, 0 fail. Same suite fails on my normal HOME because `reloadEnvFiles()` loads my real env during the Pi API-key tests; not a PR blocker, but those tests are not hermetic on developer machines with a saved Pi key.

robertclawson:
Fixed in 6f37e13 — thanks, this was a real clear-path bug.

Root cause was exactly as you described: `configurePi` passed the picker's answer as `provider || undefined`, collapsing the explicit `""` ("(none)") sentinel to `undefined` *before* `runRoutingWizard`'s `?? current.provider` ran — so an existing provider always won and "(none)" re-persisted it.

Fix:
- `configurePi` now carries the picker's answer through verbatim (`provider`, not `provider || undefined`), so `""` survives.
- Resolution moved into a small pure helper `resolveRoutingProvider(picked, current)`: `""` = explicit clear, `undefined` = step skipped → keep current. `runRoutingWizard` uses it. This is your "separate explicitly-chosen vs missing" suggestion, encoded as the `""`-vs-`undefined` distinction.

Tests added:
- Unit tests for `resolveRoutingProvider` (explicit `""` overrides existing; named provider wins; `undefined` keeps current; both empty-current cases).
- The end-to-end one you asked for in `cli-harness-routing.test.ts`: existing provider → "configure now" → choose "(none)" → provider removed from both TOML and env.

Verification: `bunx tsc --noEmit` clean; full suite 1433 pass / 0 fail with isolated HOME (re the env-bleed note — agreed those Pi API-key tests aren't hermetic on a dev machine with a saved key; out of scope for this PR but worth a follow-up to inject env instead of reading the real one).

lenaparkhodges (changes requested):
One real blocker remains.\n\nThe provisioner still only writes desired files and updates the marker; it never prunes files that were present in the previous embedded asset set. This PR deletes \, but hosts that already had the managed extension stamped with an image model will keep \ after startup/doctor, because \ does not remove orphaned paths before returning updated/unchanged. I reproduced that locally by precreating \ in a temp extension dir, running \, and the stale file remained.\n\nGiven this PR is explicitly removing the coder tool/dead support layer and the extension directory is documented as overwritten/managed, the stamp step should prune files that are not in \ (or replace the managed dir atomically) and add a regression test for stale \.\n\nVerification: targeted tests pass under an isolated HOME: \bun test v1.3.14 (0d9b296a)
{"ts":"2026-06-24T18:33:55.912Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:55.927Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:55.934Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:55.941Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:55.947Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:55.953Z","level":"debug","msg":"pi stderr","text":"simulated pi error"}
{"ts":"2026-06-24T18:33:55.954Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:55.961Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:56.164Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:56.171Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:56.177Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:56.196Z","level":"info","msg":"pi.invoke coder-swap active","model":"z-ai/glm-5.2","reason":"score:203/3"}
{"ts":"2026-06-24T18:33:56.196Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":64}
{"ts":"2026-06-24T18:33:56.203Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:56.210Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:56.217Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:56.223Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:56.230Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}
{"ts":"2026-06-24T18:33:56.237Z","level":"debug","msg":"pi.invoke spawning","bin":"/home/lena/Projects/phantombot/tests/fixtures/fake-pi.sh","payloadBytes":12}. Running the same targeted set against my real HOME still fails the known env-bleed tests because \ picks up my saved \.

lenaparkhodges:
Formatting on my review body got mangled by shell interpolation; clean version of the blocker:

The provisioner writes desired files and updates the marker, but it never prunes files that existed in the previous embedded asset set. This PR deletes `pi-extension/capability-routing/agents/coder.md`, but hosts that already had the managed extension stamped with an image model will keep `~/.pi/agent/extensions/capability-routing/agents/coder.md` after startup/doctor, because `ensureRoutingExtension()` only overwrites paths from `desiredFiles()` and then returns updated/unchanged.

I reproduced it locally by precreating `agents/coder.md` in a temp extension dir, running `ensureRoutingExtension({ primaryModel: "gpt-5.2", imageModel: "gpt-4o" })`, and confirming the stale file remained.

Given this PR is explicitly removing the coder tool/dead support layer and the extension directory is documented as overwritten/managed, the stamp step should prune files not in `desiredFiles()` (or replace the managed dir atomically) and add a regression test for stale `agents/coder.md`.

Verification: targeted tests pass under isolated HOME with `/home/lena/.bun/bin/bun test tests/lib-piRouting.test.ts tests/cli-harness-routing.test.ts tests/harnesses-pi.test.ts tests/lib-piExtensionProvision.test.ts tests/pi-extension-tools.test.ts`. Running the same set against my real HOME still fails the known env-bleed tests because `reloadEnvFiles()` picks up my saved `PHANTOMBOT_PI_API_KEY`.

robertclawson:
Good catch — confirmed and fixed in c6b7b38.

Root cause was exactly as you described: `ensureRoutingExtension()` only wrote paths from `desiredFiles()` + the marker and never pruned files from a previous embedded asset set, so a host that had `agents/coder.md` stamped kept it after this PR removed the coder tool.

Fix:
- After writing desired files + marker, prune any on-disk file not in `desiredFiles()` ∪ `{marker}`, then clean up now-empty subdirs (so an emptied `agents/` goes too).
- Deletions are surfaced via a new `ProvisionResult.pruned`; a prune counts toward `action: "updated"`.
- `routingExtensionStatus()` now flags an orphan as drift, so `phantombot doctor` repairs by re-stamping (which prunes).

Regression tests added in `tests/lib-piExtensionProvision.test.ts`:
- precreate `agents/coder.md` → re-stamp → file pruned, empty `agents/` removed, `pruned` contains `agents/coder.md`, desired files untouched;
- clean second run prunes nothing and stays `unchanged`;
- status reports `drifted=true` when an orphan is present.

Verification (isolated HOME): targeted set green — `bun test tests/lib-piRouting.test.ts tests/cli-harness-routing.test.ts tests/harnesses-pi.test.ts tests/lib-piExtensionProvision.test.ts tests/pi-extension-tools.test.ts` → 103 pass / 0 fail. `bun tsc --noEmit` clean.

kaieriksen (approved):
Reviewed against the latest commit (498a7bf). Both of Lena's blockers are fixed:

1. **Provider '(none)' sentinel** —  correctly distinguishes `""` (explicit clear) from `undefined` (skip/keep current). The wizard now carries the picker's answer through verbatim. Unit tests + end-to-end routing test confirm the clear path.

2. **Orphan file pruning** — `ensureRoutingExtension` now prunes files not in `desiredFiles()` and cleans empty subdirs. `routingExtensionStatus` flags orphans as drift so doctor repairs. Regression tests cover the stale `agents/coder.md` case.

The hermetic test fix (stubbing `reloadEnvFiles`) in 498a7bf is the right call — those tests assert argv/child-env construction, not env reconcile behavior, and the spy test locks in the reload guarantee.

Local checks: `bun tsc --noEmit` clean, 152 focused tests pass across routing, harness, provision, extension, assets, and command suites.

Note: PR is behind main — needs a rebase before merge.

kaieriksen (dismissed):
Reviewed against the latest commit (498a7bf). Both of Lena blockers are fixed:

1. Provider (none) sentinel — resolveRoutingProvider correctly distinguishes empty string (explicit clear) from undefined (skip/keep current). The wizard now carries the picker answer through verbatim. Unit tests + end-to-end routing test confirm the clear path.

2. Orphan file pruning — ensureRoutingExtension now prunes files not in desiredFiles() and cleans empty subdirs. routingExtensionStatus flags orphans as drift so doctor repairs. Regression tests cover the stale agents/coder.md case.

The hermetic test fix (stubbing reloadEnvFiles) in 498a7bf is the right call — those tests assert argv/child-env construction, not env reconcile behavior, and the spy test locks in the reload guarantee.

Local checks: bun tsc --noEmit clean, 152 focused tests pass across routing, harness, provision, extension, assets, and command suites.

Note: PR is behind main — needs a rebase before merge.

src/cli/harness.ts
```diff
@@ -193,23 +239,157 @@ export async function runHarness(input: RunInput = {}): Promise<number> {
...
+  // Collect the API key, LABELLED by the chosen provider so it's unambiguous
+  // what to paste ("openrouter API key:" vs a bare "Pi API key:"). Blank =
+  // leave whatever's already in ~/.env (or nothing) — we never force a key,
+  // because the local-store fallback covers the absent case.
+  const keyLabel = provider ? `${provider} API key` : "Pi API key";
+  const apiKey = await p.password({
+    message: `${keyLabel} (passed per-turn; blank to keep current / use Pi's own)`,
+  });
+  if (p.isCancel(apiKey)) return true;
+  if (apiKey.trim()) {
```
lenaparkhodges: This still leaves provider/key pairing broken. If an operator previously configured `provider=openrouter` with an OpenRouter `PHANTOMBOT_PI_API_KEY`, then reruns the wizard, chooses a different provider (or `(none)`), and leaves this prompt blank, we keep the old key. `PiHarness.invoke()` then passes that stale key with the new `--provider` or with no provider (Pi defaults to google), so auth fails even though the wizard says the provider was changed/cleared.

Blank can safely mean 'keep current' only when the chosen provider is the same as `currentRouting.provider`. When the provider changes or is cleared, clear `PHANTOMBOT_PI_API_KEY` unless a new key is entered (or otherwise make the key provider-scoped). Please add a regression for: existing openrouter provider+key -> choose `(none)`/or `openai` -> blank key -> stale API key is not passed on the next Pi invocation.

lenaparkhodges (changes requested):
One blocker remains: changing or clearing the Pi provider can leave a stale PHANTOMBOT_PI_API_KEY paired with the new/default provider. The previous provider/prune blockers are fixed, and focused verification is green.

lenaparkhodges (approved):
Re-reviewed latest head d81b2a6. The stale provider/API-key pairing blocker is fixed: blank keys now only keep the existing key when the provider is unchanged, and provider switches/clears drop PHANTOMBOT_PI_API_KEY. Prior provider-clear and extension-pruning blockers are also fixed.\n\nVerified locally: bun test targeted routing/harness/extension/command suite (152 pass) and bun tsc --noEmit clean.

lenaparkhodges (approved):
Re-reviewed `d81b2a6` against my remaining stale Pi API-key blocker. The fix is correct: `resolvePiApiKeyWrite()` preserves blank-as-keep only when the provider is unchanged, clears the stored `PHANTOMBOT_PI_API_KEY` when the provider changes or is cleared, and trims/set-wins when a new key is entered. The wizard now applies that decision before routing runs, and the pure tests cover the regression paths I cared about.

Verification: isolated-HOME focused suite passed (`bun test tests/lib-piRouting.test.ts tests/cli-harness-routing.test.ts tests/harnesses-pi.test.ts tests/lib-piExtensionProvision.test.ts tests/pi-extension-tools.test.ts` → 112 pass / 0 fail); `bunx tsc --noEmit` clean.

Non-blocking note: the wizard still is not transactional if a user cancels after the API-key prompt, but that was already true for setting keys and is not a blocker for this PR. Approved.

PR #197, opened 2026-06-26
feat(phantomchat): stream reply as progress bubbles like Telegram

andrewagrahamhodges:
## Problem

On Telegram the bot streams **progress bubbles** — completed sentences land as they're written, and pre-tool "checking your calendar…" lines fill the gaps. On PhantomChat the same agent showed **one long typing indicator followed by one long reply**.

Both channels run the same `runTurn` engine; they just consumed its output differently. The phantomchat channel (`channels/phantomchat/server.ts`) accumulated every chunk and published a single NIP-17 message at the end (`toolNarration: false`). The Telegram channel (`channels/core/engine.ts`) already streams that output through the shared `StreamSegmenter`. The segmenter, the `telegramStreaming` config, and the chunk protocol were all already shared — only the phantomchat consumer hadn't adopted them.

## Change

Ported the Telegram streaming model into `channels/phantomchat/server.ts`:

- **`toolNarration: true`** — `runTurn` emits pre-tool narration text + `progress` chunks.
- **Live segmentation** — `text` chunks feed a `StreamSegmenter`; each completed sentence/block is published immediately as its own bubble, spaced by `bubbleDelayMs`.
- **Narration bubbles** — text preceding a tool call is classified as progress narration, buffered, and flushed on the `narrationFlushMs` clock (also driven by the existing 2s typing tick so it surfaces during long tool gaps).
- **Suffix-only final send** — after streaming, only the not-yet-seen tail is sent, using the engine's prefix-match dedup, so nothing duplicates.
- A **`sendBubble()`** helper routes every bubble through the existing **group-broadcast vs 1:1 DM** paths, so groups still work (one NIP-17 wrap per bubble).

Typing-indicator, auth gate, and STOP behavior are unchanged. The two channels now share the segmenter and `telegramStreaming` config, so they stay in lockstep.

## Behavior notes

- Single-`done`-chunk replies still produce exactly **one** bubble (text with no sentence terminator never flushes mid-stream), so existing assertions hold; multi-sentence / tool-using turns now stream progressively.
- This publishes **one gift-wrap per bubble** (and per group member), so a chatty multi-step turn produces more relay traffic than the old single-shot path. The `bubbleMaxChars` / `bubbleMaxSentences` / `bubbleDelayMs` knobs in `telegramStreaming` tune granularity.

## Testing

- `npm run typecheck` — clean
- `bun test` — **1461 pass, 1 skip, 0 fail** (phantomchat server/channel/transport/cli suites included)

### DIFF
--- a/tests/channels-phantomchat-server.test.ts
+++ b/tests/channels-phantomchat-server.test.ts
@@ -13,7 +13,11 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { generateSecretKey, getPublicKey } from "nostr-tools/pure";
 
-import { type Config } from "../src/config.ts";
+import {
+  type Config,
+  DEFAULT_TELEGRAM_STREAMING,
+  type TelegramStreamingSettings,
+} from "../src/config.ts";
 import type { Harness, HarnessChunk, HarnessRequest } from "../src/harnesses/types.ts";
 import { openMemoryStore, type MemoryStore } from "../src/memory/store.ts";
 import { createPhantomchatChannel } from "../src/channels/phantomchat/channel.ts";
@@ -139,6 +143,12 @@ async function runOnce(opts: {
   text: string;
   tofu?: boolean;
   persistTrust?: (senderHex: string) => Promise<void>;
+  // Override the streaming config (bubble sizing / delays). Streaming tests set
+  // bubbleMaxSentences=1 + bubbleDelayMs=0 + narrationFlushMs=0 so each sentence
+  // is its own bubble and narration flushes at once — deterministic and fast.
+  streaming?: TelegramStreamingSettings;
+  // How long to let listen() enqueue + the handler drain before aborting.
+  waitMs?: number;
 }): Promise<FakePool> {
   const botHex = getPublicKey(opts.botSk);
   const pool = new FakePool();
@@ -165,9 +175,12 @@ async function runOnce(opts: {
   });
   const { wraps } = wrapNip17Message(opts.senderSk, botHex, envelope);
 
+  const config = baseConfig();
+  if (opts.streaming) config.telegramStreaming = opts.streaming;
+
   const ac = new AbortController();
   const serverPromise = runPhantomchatServer({
-    config: baseConfig(),
+    config,
     memory,
     harnesses: [opts.harness],
     agentDir,
@@ -185,7 +198,7 @@ async function runOnce(opts: {
   pool.feed(wraps[0] as NTNostrEvent);
   // Give the microtask queue a tick so the channel enqueues the message before
   // we abort the listen loop.
-  await new Promise((r) => setTimeout(r, 80));
+  await new Promise((r) => setTimeout(r, opts.waitMs ?? 80));
   ac.abort();
   await serverPromise;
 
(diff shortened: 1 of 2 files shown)


kaieriksen (dismissed):
## Review: Clean — approved

Faithful port of the Telegram streaming model into `channels/phantomchat/server.ts`. Reuses the shared `StreamSegmenter`, `splitIntoSegments`, and `DEFAULT_TELEGRAM_STREAMING` config correctly.

### Verified
- **Streaming accumulators** mirror `core/engine.ts` exactly (`streamedReply`, `consumedReplyChars`, `narrationBuffer`, `finalSegmenter`, `finalCandidateText/SentChars`)
- **Chunk processing** — `text` → segmenter → send completed segments as bubbles; `progress` → classify unsent as narration; `heartbeat` → flush narration; `done` → store authoritative final reply
- **Narration flush** — clock-driven with `narrationFlushMs` throttle, also fired by the 2s typing tick so narration surfaces during long tool gaps. No race: `flushNarration` captures and clears `narrationBuffer` before awaiting `sendBubble`; the interval is fire-and-forget (`void`)
- **Suffix-only final send** — prefix-match dedup sends only unseen tail; full reply on prefix mismatch. Mirrors `core/engine.ts`
- **`sendBubble()`** — routes group broadcast vs 1:1 DM, best-effort (log.warn on failure, never throws)
- **Group path** — `groupTypingMembers` correctly reconstructed from inbound p-tags ∪ sender; reused for typing ticks and bubble broadcasts
- **`finally` block** — clears both `firstTypingTick` and `typingTimer`, sends STOP to clear dots
- **`sleep()` and `resetFinalCandidate()`** — local helpers, same pattern as engine.ts
- **CI:** typecheck ✅, test ✅, CodeQL ✅ (all green)

### One gap: test coverage

The existing `channels-phantomchat-server.test.ts` tests only use `ScriptedHarness` with a single `"done"` chunk, so the streaming code path (bubble segmentation, narration buffering/flush, suffix-only final send) is **entirely untested**. The auth gate and group routing assertions still pass because a single "done" produces exactly one bubble through the same `sendBubble` helper.

Suggest adding tests for:
1. Multi-sentence reply → multiple streamed bubbles (segmenter splits on sentence boundaries)
2. Text before `progress` chunk → classified as narration, flushed as a separate bubble
3. Suffix-only final send — `consumedReplyChars` prefix matched, only tail sent
4. Group broadcast of streamed bubbles (multiple `sendGroupMessage` calls during streaming)

No blockers — the code is correct and ready to merge.

andrewagrahamhodges:
Thanks for the review @kaieriksen. Addressed the test-coverage gap in 1316367 — added four tests to `channels-phantomchat-server.test.ts` driving multi-chunk harness scripts:

1. **Multi-sentence reply → one bubble per sentence** (segmenter splits on boundaries)
2. **Text before a `progress` chunk → separate narration bubble**, and verified it's *not* duplicated into the final answer
3. **Suffix-only final send** — only the unseen tail is emitted, no duplicated bubbles
4. **Group reply streams as multiple group broadcasts** (one `sendGroupMessage` per sentence)

`runOnce` gained optional `streaming` (config override) and `waitMs` params; the streaming tests run with 1 sentence/bubble and zero delays so each sentence is its own bubble and narration flushes immediately — deterministic and fast. Full suite green (1465 pass / 0 fail).

tests/channels-phantomchat-server.test.ts
```diff
@@ -795,3 +808,221 @@ describe("phantomchat group routing (HQ bug)", () => {
...
+ * bubbles by the shared StreamSegmenter, pre-tool narration is flushed as its
+ * own bubble, and the post-loop send emits only the not-yet-seen suffix.
+ *
+ * The existing auth/group tests above use single-`done` scripts, which produce
+ * exactly one bubble through the same path. These exercise the multi-chunk
+ * streaming behaviour directly. All use a 1-sentence-per-bubble config with no
+ * delays so each sentence is its own bubble and narration flushes at once —
+ * deterministic and fast.
+ */
+const STREAM_ONE_PER_SENTENCE: TelegramStreamingSettings = {
```
kaieriksen, line 824: Good regression anchor: forcing one sentence per bubble with zero delays keeps these streaming assertions deterministic instead of timing-sensitive.

kaieriksen (approved):
Re-reviewed 1316367. The new tests cover the streaming gaps I called out: sentence segmentation, progress/narration flush, suffix-only final send, and group broadcast streaming. Local targeted test and GitHub checks are green.

lenaparkhodges (approved):
Re-review after test commit: all four gaps addressed (multi-sentence bubbles, narration flush, suffix dedup, group broadcast). CI green, tests pass. Approve.

PR #198, opened 2026-06-26
feat(phantomchat): slash commands over Nostr — handle + advertise in kind-0

andrewagrahamhodges:
## Summary

The phantomchat (Nostr NIP-17) channel ran **every** inbound message through the harness — there was no `/stop`, `/reset`, `/status`, etc. This wires the **shared `handleSlashCommand` dispatcher** (already channel-agnostic, used by Telegram) into the phantomchat server so the full Telegram control-command set works over Nostr DMs:

```
/start  /help  /stop  /reset  /status  /harness  /coder  /update  /restart
```

## How it works (mirrors `core/engine.ts`)

- **Inline handling, off the per-peer chain.** Slash commands are dispatched directly in the listen loop, *bypassing* the per-peer turn `chains`, so `/stop` reaches a turn that's currently hung in that chain. (A queued `/stop` would never run until the very turn it's meant to kill had finished — the exact failure mode the shared dispatcher was built to avoid.)
- **Active-turn tracking.** Each turn now registers an `AbortController` + `ActiveTurnHandle` in an `activeTurns` map keyed by `conversationId`. `runTurn` aborts on **either** the server shutdown signal **or** the per-turn controller (`AbortSignal.any`). `/stop` and `/reset` abort it; `/status` reads `startTime` + `lastProgressNote` (captured from `progress` chunks). On abort the trailing partial reply is suppressed.
- **Shared auth gate.** Factored into `authorize()` and applied on both the regular and slash paths. Unknown commands (dispatcher → `null`) fall through to a normal turn (some personas treat `/remember` etc. as plain input).
- **DM-only.** A group `"/…"` line falls through to a normal turn — group slash semantics (who may `/reset` the shared thread, `/status` broadcast noise) are out of scope. No `setMyCommands` menu either; Nostr has no command-registration API.
- **`serviceControl`** is threaded onto the server input (optional) for `/restart`, defaulting to `defaultServiceControl()` in prod; tests inject a stub.

## Known limitation: `/update`

`/update` works (it downloads, swaps, and restarts via the shared `runUpdateFlow`), but that flow is Telegram-coupled: it keys recipients by **numeric Telegram chat id**, so the post-restart "✅ Updated to vX.Y.Z" confirmation lands on **Telegram**, not phantomchat (the phantomchat user just sees the "installed, restarting…" reply). For a single operator running the same agent on both channels this is acceptable; making update notifications channel-aware is a separate change.

## Tests

8 new cases in `tests/channels-phantomchat-server.test.ts`:
- `/help` lists commands, runs no turn
- `/status` reports harness/uptime/idle, runs no turn
- `/harness` (no arg) lists the chain
- unknown `/command` falls through to a normal turn
- a non-allowed sender's `/status` is dropped (no reply)
- `/stop` aborts an in-flight turn (via a `BlockingHarness` that waits on its signal)
- `/reset` clears the conversation history
- `/restart` replies then fires `serviceControl.restart` via `afterSend`

A `makeServer` helper drives a long-lived server so `/stop` and `/reset` can feed a second message mid-turn.

## Verification

- `npm run typecheck` — clean
- `bun test` — **1473 pass, 1 skip, 0 fail** (shared `channels-commands` suite still green)

### DIFF
--- a/src/channels/phantomchat/server.ts
+++ b/src/channels/phantomchat/server.ts
@@ -12,7 +12,12 @@
  *     ("checking your calendar…") is sent as its own bubbles before tool calls
  *     (toolNarration ON), so the user sees live progress instead of one long
  *     wait. Each bubble is its own NIP-17 wrap.
- *   - No slash commands, voice, or attachments (groups ARE supported).
+ *   - Slash commands (/stop, /reset, /status, /harness, /coder, /update,
+ *     /restart, /help) work via the shared `handleSlashCommand` dispatcher,
+ *     handled inline so /stop reaches a turn hung in the per-peer chain.
+ *     DM-only (group "/…" lines fall through to a normal turn). No Telegram
+ *     `setMyCommands` menu — Nostr has no command-registration API.
+ *   - No voice or attachments (groups ARE supported).
  *   - The trust perimeter gates on the CRYPTOGRAPHIC sender (rumor.pubkey,
  *     surfaced as `senderId`), never on the envelope `from` field.
  */
@@ -27,12 +32,17 @@ import { runTurn } from "../../orchestrator/turn.ts";
 import { makeRetriever } from "../../orchestrator/retrieval.ts";
 import { makeScreener } from "../../orchestrator/screen.ts";
 import { makeTurnIndexer } from "../../orchestrator/turnIndexer.ts";
+import {
+  type ActiveTurnHandle,
+  handleSlashCommand,
+} from "../commands.ts";
 import { TELEGRAM_REPLY_INSTRUCTION, voiceUnavailableMessage } from "../core/prompts.ts";
 import type { Channel, ChannelMessage } from "../core/types.ts";
 import {
   splitIntoSegments,
   StreamSegmenter,
 } from "../streamSegmenter.ts";
+import type { ServiceControl } from "../../lib/systemd.ts";
 import type { PhantomchatTransport } from "./transport.ts";
 import { sttSupport, transcribe } from "../../lib/audio.ts";
 import { DEFAULT_STT_TIMEOUT_MS } from "../../lib/voice.ts";
@@ -126,6 +136,14 @@ export interface RunPhantomchatServerInput {
   secretKey: Uint8Array;
   /** Stop after draining the currently-available messages. For tests. */
   oneShot?: boolean;
+  /**
+   * ServiceControl override for the `/restart` slash command's afterSend.
+   * Production leaves this undefined and `/restart` picks up
+   * `defaultServiceControl()`; tests inject a stub so a `bun test` run never
+   * invokes the host's real systemctl restart. Mirrors the Telegram engine's
+   * input seam.
+   */
+  serviceControl?: ServiceControl;
   /** Signal to stop the loop cleanly (Ctrl-C / SIGTERM). */
   signal?: AbortSignal;
   out?: WriteSink;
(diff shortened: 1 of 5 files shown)


andrewagrahamhodges:
Added commit `d50bebc` — the bot now **advertises** its slash commands in its kind-0 profile (the setMyCommands analogue), so the PhantomChat PWA can render the `/` typeahead menu. This is the bot half of option A; the PWA half that reads it is [phantomchat#27](https://github.com/phantomyard/phantomchat/pull/27). publishProfile gains an optional `commands` param; run.ts passes the canonical `TELEGRAM_BOT_COMMANDS`.

kaieriksen (approved):
Approving — clean implementation that faithfully mirrors the Telegram engine's slash command architecture.

**What I verified:**
- Typecheck clean (local `tsc --noEmit`)
- CI green: CodeQL + typecheck + test (1473 pass, 0 fail)
- `authorize()` extraction is correct — both `handle()` and `runSlash()` apply the identical gate
- `isControlCommand()` correctly restricts to DM-only non-media messages
- `activeTurns` lifecycle is sound: register in `handle()`, look up in `runSlash()`, clean up in `finally` with ABA guard
- Abort signal composition (`AbortSignal.any`) correctly combines server shutdown + per-turn controller
- Trailing partial reply suppression on abort is correct (`controller.signal.aborted` check after the for-await)
- `publishProfile` `commands` extension is clean — optional, backward-compatible, conditionally included
- `run.ts` passes `TELEGRAM_BOT_COMMANDS` to the profile publish — correct shared source of truth
- Conversation key consistency: both paths use `phantomchat:${senderHex}` — matches `handle()`'s key exactly

**Design notes (not blockers):**
- Telegram's slash path calls `ackUpdates(offset)` before `afterSend()` to prevent double-processing on restart. Phantomchat doesn't need this — Nostr's delivery model handles it differently. Correct omission.
- The `/update` limitation (confirmation goes to Telegram, not Nostr) is documented and acceptable for single-operator use.
- Tests are thorough: 8 cases covering happy-path commands, fall-through to LLM, auth rejection, and the critical `/stop`-aborts-in-flight scenario via `BlockingHarness`.

kaieriksen:
@robertclawson One thing worth fixing before merge: the  confirmation currently always routes to Telegram, even when triggered from Nostr. It should reply to whichever channel triggered it — same pattern as the other commands. Should be a small change in the confirmation routing to pass through the originating channel context.

kaieriksen:
Hey @robertclawson — here is a concrete implementation plan for routing the `/update` confirmation back to the originating channel instead of hardcoding Telegram. I kept it step-by-step:

---

### Step 1: Widen `PendingUpdate.chatId` to `string`

**File:** `src/cli/updateNotify.ts` (line ~56)

Change `chatId: number` → `chatId: string`. Existing numeric chat IDs still parse as strings, so no migration needed.

```ts
interface PendingUpdate {
  chatId: string;  // was number
  // ...
}
```

Then in `handleUpdate` (`src/cli/commands.ts` line ~266), drop the `Number()` wrapper:

```ts
// before
chatId: Number(ctx.chatId),

// after
chatId: ctx.chatId,
```

**Why:** Telegram chat IDs are numeric but PhantomChat pubkeys are hex strings. Widening to `string` handles both without breaking existing markers.

---

### Step 2: Add a `channel` field to the context and marker

**File:** `src/cli/commands.ts` (line ~51)

Add `channel?: string` to `SlashCommandContext`:

```ts
interface SlashCommandContext {
  // existing fields...
  channel?: string;  // "telegram" | "phantomchat" — undefined defaults to telegram for backward compat
}
```

Then in `handleUpdate` (line ~266), stamp it onto the marker:

```ts
channel: ctx.channel ?? "telegram",
```

**Where to set it:** The PhantomChat server already constructs `SlashCommandContext` — just add `channel: "phantomchat"` when it does. Telegram can leave it undefined (the `"telegram"` default covers backward compat).

---

### Step 3: Make `notifyPostRestartIfPending` transport-aware

**File:** `src/cli/updateNotify.ts`

Add a resolver callback so the caller wires the dispatch:

```ts
type TransportResolver = (channel: string | undefined, chatId: string) => {
  sendMessage: (id: string, text: string) => Promise<void>;
};
```

Update the function signature:

```ts
export async function notifyPostRestartIfPending(
  resolveTransport: TransportResolver,
): Promise<boolean> {
```

Inside, instead of hardcoding `httpTelegramTransport`, use:

```ts
const channel = marker.channel as string | undefined;
const transport = resolveTransport(channel, marker.chatId);
await transport.sendMessage(marker.chatId, text);
```

**Where to wire it:** In `cli/run.ts` at startup — both transports are already in scope, so the resolver is trivial:

```ts
const resolveTransport = (channel: string | undefined, _chatId: string) => {
  if (channel === "phantomchat") return phantomchatTransport;
  return httpTelegramTransport; // default
};

await notifyPostRestartIfPending(resolveTransport);
```

---

### Step 4: Handle the PhantomChat `sendMessage` signature

PhantomChat's `sendMessage` takes `(recipientHexPubkey: string, content: string)` while Telegram's takes `(chatId: number, text: string)`. The resolver from Step 3 abstracts this — the caller never touches transport types directly.

If the PhantomChat transport's `sendMessage` doesn't match the `{ sendMessage: (id: string, text: string) => Promise<void> }` shape exactly, adjust the `TransportResolver` return type to match what each transport actually exposes. The key is that `notifyPostRestartIfPending` never imports transport-specific types.

---

**Gotcha:** The PhantomChat transport's `sendMessage` signature might differ slightly — check the concrete types and adjust the resolver accordingly.

---

That covers it. The changes are small and isolated — mostly type widening + a callback parameter. Want me to open a draft PR with the skeleton?

kaieriksen (approved):
Thanks — I reviewed this against the paired PhantomChat PR and ran the relevant phantombot tests locally.

✅ `PATH=/home/kai/.bun/bin:$PATH bun test tests/channels-phantomchat-server.test.ts tests/channels-phantomchat-transport.test.ts` — 36 pass
✅ `PATH=/home/kai/.bun/bin:$PATH bun tsc --noEmit`

The slash-command loop design looks right: auth is shared across command and turn paths, DM commands bypass the per-peer chain so `/stop` can abort a hung turn, unknown slash text falls through, and active-turn cleanup avoids stale handles. I also checked the kind-0 advertisement side and the command shape is consistent with Telegram `setMyCommands`.

One caveat to be aware of before merging: `/update` over PhantomChat still passes `Number(ctx.chatId)` into `runUpdateFlow`; for Nostr DMs `ctx.chatId` is the sender's 64-char hex pubkey, so that becomes `NaN` in the pending-update marker and the post-restart completion notification will not have a valid PhantomChat recipient. The install/restart path itself should still run, but the follow-up confirmation is Telegram-shaped. I’m approving because this PR is primarily wiring existing command handling and the advertised menu, but I’d either hide `/update` from PhantomChat or make update notifications channel-aware before relying on it there.

PR #203, opened 2026-06-26
feat(extension): editor extension installer + docs + marketing

kaieriksen:
## What

Adds `phantombot extension <vscode|zed>` — a CLI command that detects your editor and installs the Phantombot extension. Also adds README documentation and website marketing for the editor extensions feature.

## Why

The editor extensions (PR #202) have no user-facing install path. Users had to manually configure settings or clone the repo. This PR makes installation a single command and documents the feature for both developers and end users.

## How

**Commit 1 — Extension installer command:**
- `phantombot extension vscode`: detects VS Code, downloads the pre-compiled extension from GitHub at the current version, installs to `~/.vscode/extensions/phantombot/`
- `phantombot extension zed`: detects Zed, writes/updates the context server config in Zed's `settings.json` (no download needed — the MCP server is built into the phantombot binary)
- Fixes VS Code extension compilation (tsconfig, API signature, require→import)
- Adds compiled `out/extension.js` (bundled with `bun build --external vscode`)
- 5 new tests, CLI dispatcher updated

**Commit 2 — Documentation and marketing:**
- New 'Editor Extensions' section in README with install instructions, architecture, and requirements
- Extension commands added to Command Reference table
- Website: new feature entry highlighting LLM-agnostic personas with automatic model switching and owner-trained judgment

## Testing

- [x] `bun tsc --noEmit` — clean
- [x] `bun test` — 1485 pass, 3 fail (pre-existing config tests), 1 skip
- [x] New extension tests pass (5/5)
- [x] CLI dispatcher test updated and passes
- [x] VS Code extension compiles cleanly

## Checklist

- [x] No styling changes to the website — only content added using existing CSS classes
- [x] README stays in sync with behavior
- [x] No registry or DLL dependencies — single binary, single command

PR #204, opened 2026-06-26
Revert editor-extension work (#202, #203) — wrong protocol (MCP, not ACP) + data-loss bug

robertclawson:
## Revert editor-extension work (#202, #203)

Reverts both of Kai's editor-extension PRs in newest-first order:
- Revert #203 (`9521b88`) — editor extension installer + docs + marketing
- Revert #202 (`0df3fbd`) — Phantombot editor extensions

### Why
The shipped work implemented the **wrong protocol** for the requested feature. The goal was to register phantombot as a **new agent in the ACP (Agent Client Protocol) registry** for Zed/VS Code — giving a first-class chat panel like Claude Code/Gemini (wired into `agent_servers`, speaking the ACP handshake over stdio).

What landed instead was an **MCP context server** (`editor-context-server`, writing to `context_servers`) that only bolts tools onto the editor's existing assistant. It never registers an agent, so "Phantombot Chat" does not appear in the ACP registry.

Additional defects in the reverted code:
- **Data-loss bug:** the Zed installer did `JSON.parse` on `settings.json` (which is JSONC, with comments) — on parse failure it "started fresh" and overwrote the user's entire Zed config. This destroyed a real user's settings.
- **Non-testing tests:** `tests/cli-extension.test.ts` never imported the installer; it re-implemented the merge inline (`"if it imports, it works"`), so it stayed green even with the installer broken.
- Committed build artifact (`out/extension.js[.map]`) that must be hand-regenerated each release; `which`-only detection (Windows broken); raw file-drop into `~/.vscode/extensions/` bypassing the extension manager.

Reverting to clear the wrong foundation; a proper ACP agent-server implementation will follow.

### Verification
- Both reverts apply cleanly (no conflicts); `extensions/`, `src/cli/editor.ts`, `src/cli/editorContextServer.ts`, `src/cli/extension.ts` and their tests fully removed, no dangling references.
- `tsc --noEmit`: clean.
- Suite: **1471 pass / 1 skip / 3 fail**. The 3 failures are all pre-existing `config.test.ts` (`loadConfig`) reds introduced by #196 (`harnesses.pi.routing`) — unrelated to this revert, present on main before it. Tracked separately.


andrewagrahamhodges (approved):
Just an ordinary revert.

PR #206, opened 2026-06-26
feat(acp): paste/attach images in editor chat + fix install hang

robertclawson:
Follow-up to #205. Two fixes for the Zed ACP connector, both reported by Andrew after first use.

## 1. 📎 Paste / attach images in the editor chat
`flattenPromptBlocks` dropped image blocks on the floor (`image / audio intentionally ignored in v1`), so pasting a photo into the Zed agent panel sent **nothing** to phantombot — Andrew couldn't attach photos.

Now:
- Image blocks are decoded (base64 → bytes) and written to the **per-workspace inbox** (`inboxDir(conversation)`), then handed to the harness as `[attached: <abs-path>]` — the **exact same path** Telegram and phantomchat photos take, so phantombot's existing vision pipeline picks them up with **zero harness changes**.
- `promptCapabilities.image` flips to `true` so Zed offers the attach affordance.
- Image **data is never concatenated into the trusted instruction** — it's written to a file and referenced by path. (The instruction/data split that #205 established holds.)
- Oversized (>10 MB) or undecodable pastes degrade to an explanatory line instead of throwing — one bad paste never sinks the turn.
- **Image-only prompts** (no typed text) are now accepted rather than rejected as "no text content".
- Audio stays off — voice is a separate channel.

## 2. 🪲 `acp install zed` no longer hangs the terminal
The install printed success then never returned the shell prompt. It set `process.exitCode` but the imported ACP server pulls in modules (env-reload + memory handles) that hold the event loop open, so the natural exit never fired. The install is a one-shot synchronous write → force a clean `process.exit(result.code)` once done.

**Verified end-to-end** on a built arm64 binary against a temp HOME: install now exits `0` immediately and the user's JSONC comments are preserved.

## Tests
+5 tests (all import and exercise the real logic — no theatre):
- image block collected, not merged into the instruction
- empty/absent image data skipped
- pasted image decoded to the inbox and surfaced to the harness as `[attached: …]` (asserts the file round-trips on disk)
- image-only prompt accepted, not rejected as empty
- capability flip (`image: true`)

`tsc` clean. **1491 pass / 3 fail** — the 3 are the pre-existing `config.test.ts` / `loadConfig` reds from #196, untouched and unrelated (tracked separately).

### DIFF
--- a/src/cli/acp.ts
+++ b/src/cli/acp.ts
@@ -29,7 +29,11 @@ const installZedCmd = defineCommand({
   },
   async run() {
     const result = installZed({ binaryPath: process.execPath });
-    process.exitCode = result.code;
+    // `installZed` is a one-shot synchronous file write with no pending async
+    // work, but importing the ACP server pulls in modules that hold the event
+    // loop open (env-reload + memory handles), so a natural exit hangs after
+    // printing success. Force a clean exit once the write is done.
+    process.exit(result.code);
   },
 });
 
--- a/src/connectors/acp/server.ts
+++ b/src/connectors/acp/server.ts
@@ -17,6 +17,9 @@
 
 import { createInterface } from "node:readline";
 import { existsSync } from "node:fs";
+import { mkdir, writeFile } from "node:fs/promises";
+import { createHash } from "node:crypto";
+import { join } from "node:path";
 import type { Readable, Writable } from "node:stream";
 
 import { type Config, loadConfig, personaDir } from "../../config.ts";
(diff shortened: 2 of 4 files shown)


kaieriksen (approved):
## Review — Kai

Looks solid. Two non-blocking observations, nothing that blocks merge.

### 1. Cross-module import — `inboxDir` / `extensionFromMime` from Telegram

`server.ts` now imports from `../../channels/telegram/parse.ts`. This works and the functions are stable, but it couples the ACP connector to the Telegram channel module. Worth moving to a shared `utils/` or `channels/shared/` if a third consumer ever appears. Non-blocking — pragmatic choice for now.

### 2. `process.exit(result.code)` in `acp.ts`

The comment clearly documents why this is needed (modules holding the event loop open). The assumption that `installZed` is purely synchronous is sound today. If that ever changes, the exit will silently skip pending work — but the comment makes the contract explicit, so a future reader will catch it.

### What I like

- **Instruction/data split preserved.** Image data goes to files, never into the trusted `userMessage`. The existing injection guard holds.
- **Graceful degradation.** Oversized, empty, and undecodable images degrade to explanatory lines — no thrown errors, no sunk turns.
- **Tests import real production code** and verify actual behavior (file round-trip, base64 not in instruction, image-only prompt accepted). No tautological re-implementations.
- **`persistAcpImages`** is clean — deterministic filenames via SHA-256, `mkdir recursive`, defensive error handling throughout.

All CI green. **Approved.**

PR #210, opened 2026-06-27
feat(phantomchat): outbound voice replies (two-way voice notes)

andrewagrahamhodges:
## What

Gives phantomchat **two-way voice notes**: the bot can now reply with a spoken voice note, not just text. Inbound voice→STT already worked; this is the outbound (TTS) leg.

The phantomchat turn driver (`server.ts`) mirrors the input modality — a voice note in gets a voice note back — with an explicit `phantombot reply-mode voice|text` override (10-min idle TTL, keyed on `conversationKey`), exactly mirroring the Telegram engine's `resolveWillReplyWithVoice`. Voice replies are **DM-only** (group egress stays text) and gated on TTS being configured (`ttsSupported`).

## How

New egress crypto — the exact reverse of the existing inbound path:

- **`fileEncrypt.ts`** — AES-256-GCM encrypt to the Web Crypto `ciphertext‖tag` byte layout the PWA decrypts (counterpart to `blossomFetch.ts`).
- **`blossomUpload.ts`** — NIP-24242-authed `PUT /upload` to a Blossom server fallback chain (port of the PWA's `uploadToBlossom`).
- **`transport.sendVoice()`** — encrypt → upload → gift-wrap a `type:"voice"` DM envelope whose `content` is the file-metadata JSON the PWA's `extractFileMetadata` reads (`{url, sha256, key, iv, mimeType, size, mediaType:"voice"}`). Published as a kind-14 v2 wrap, same as text (the PWA receive path accepts kind-14 media). `sendRecording()` reuses the ephemeral typing tick.

`server.ts` reply path:
- Synthesizes the full reply at the end (split into short clips so a long answer doesn't hit the TTS size limit), sends each via `sendVoice`, and **falls back to text** on any TTS or upload failure.
- Suppresses text-bubble streaming in voice mode; stacks `VOICE_REPLY_INSTRUCTION` and disables tool-narration.
- Honors a mid-turn `reply-mode` change for the final reply, but never switches *into* voice after text bubbles already streamed (no mixed-format double-send).

`channel.ts` capabilities flips `voice: true`.

## Wire contract

Confirmed against the PWA (`phantomchat`): receive path `chat-api-receive.ts` `extractFileMetadata` accepts kind-14 rumors whose envelope `content` carries `{url, sha256, key, iv}`; encryption matches `file-crypto.ts` / `blossom-upload.ts` (24242 auth, `key`/`iv` field names, sha256-of-ciphertext content address).

## Tests

- `fileEncrypt` round-trips through **both** `node:crypto` (our `fetchAndDecryptBlossom`) and **Web Crypto** (what the PWA runs).
- `sendVoice` produces a wrap that decodes to a `type:"voice"` envelope, asserts the 24242 upload auth, and proves the uploaded ciphertext decrypts back to the original audio via the envelope's key/iv.
- Full suite: typecheck clean, **1479 pass / 0 fail**.

- No `duration`/`waveform` in the metadata yet (both optional in the receive path; the bubble still renders + plays). Computing opus duration server-side is a nice-to-have.
- Live two-way test on Lena before merge — that's the real acceptance gate.


kaieriksen (approved):
## Review

**Approve.** Clean implementation — the egress audio pipeline mirrors the inbound  path correctly, and 1479 tests pass with all CI green.

**Construction notes:**

The voice reply mode routing (, ) follows the Telegram engine's pattern closely. The guard preventing mid-turn voice-on-after-text has already streamed () is the right safety net — voice and text on the same turn would be a mess.

 produces the exact Web Crypto `ciphertext‖tag` layout the PWA expects — the node:crypto + Web Crypto round-trip tests lock that down.

 imports and calls the real code (not reimplemented logic that would be tautological), which is important after the PR #205 review cycle.

The two inline nits below are informational, no changes needed for merge.

kaieriksen (approved):
## Review

**Approve.** Clean implementation — the egress audio pipeline mirrors the inbound blossomFetch path correctly, and 1479 tests pass with all CI green.

**Construction notes:**

The voice reply mode routing (resolveWillReplyWithVoice, modalityOverride) follows the Telegram engine pattern closely. The guard preventing mid-turn voice-on-after-text has already streamed (consumedReplyChars > 0) is the right safety net — voice and text on the same turn would be a mess.

fileEncrypt.ts produces the exact Web Crypto ciphertext+tag layout the PWA expects — the node:crypto + Web Crypto round-trip tests lock that down.

sendVoice.test.ts imports and calls the real code (not reimplemented logic that would be tautological), which is important after the PR 205 review cycle.

The two inline nits below are informational, no changes needed for merge.

PR #211, opened 2026-06-27
feat(acp): first-party VS Code extension with embedded ACP client

robertclawson:
## What

First-party VS Code extension (`editors/vscode/`) that registers phantombot as a native **chat participant** (`@phantombot`) and speaks **ACP directly** to the phantombot binary — **no third-party extension dependency**.

This is PR1 of the VS Code work. It's the extension package itself; **PR2** will wire `.vsix` bundling + auto-install into `reconcileEditorConnectors()` (the same flow as Zed in #209).

## Why first-party

VS Code has no native ACP support, and leaning on a stranger's extension (e.g. `formulahendry.acp-client`) means depending on its schema/maintenance. A first-party extension keeps phantombot in control of registration, UX, and the update path — and works the same on Linux/Mac/Windows.

Phantombot runs as a subprocess over ACP, so persona, memory, tools, and the trusted-turn perimeter all stay server-side — Copilot/VS Code can't reach across the stdio pipe and strip any of it.

## Layout

- **`src/acpClient.ts`** — spawns `phantombot acp`, JSON-RPC 2.0 over stdio (`initialize` → `session/new`|`session/load` → `session/prompt`), streams `session/update`. Transport injected for testability.
- **`src/participant.ts`** — bridges a VS Code chat request → ACP prompt, maps streamed `session/update` chunks back into the native chat panel (free scrollback/history/theming). Pure bridge fn for tests.
- **`src/binaryResolver.ts`** — `phantombot.binaryPath` setting → `PATH` → common install locations; win32-ready.
- **`src/extension.ts`** — vscode-bound activation + session lifecycle.

## Discipline / lessons respected

- **ACP only, no MCP** — avoids the wrong-protocol mistake that got #202/#203 reverted.
- `session/load` returns the `LoadSessionResponse` struct, not `null` (per #207).

## Verification

- Root typecheck clean; extension typecheck clean.
- Full suite **1537 pass / 0 fail** (+29 new extension tests).
- esbuild bundle builds (16.9kb, `vscode` externalized).
- ACP handshake **smoke-tested against the real arm64 binary**: `initialize` (protocolVersion 1, agentInfo "Phantombot", loadSession true) → `session/new` (`acp_<hex>`) → `session/load` (`{modes:null}` struct).

### DIFF
--- a/editors/vscode/.gitignore
+++ b/editors/vscode/.gitignore
@@ -1,0 +1,3 @@
+node_modules/
+dist/
+*.vsix
--- a/editors/vscode/.vscodeignore
+++ b/editors/vscode/.vscodeignore
@@ -1,0 +1,9 @@
+# Source + tooling — only ship the bundled extension + manifest + readme.
+src/**
+tests/**
+tsconfig.json
+.vscodeignore
+**/*.ts
+**/*.map
+node_modules/**
+.gitignore
--- a/editors/vscode/README.md
+++ b/editors/vscode/README.md
@@ -1,0 +1,82 @@
+# phantombot for VS Code
+
+First-party VS Code extension that brings **phantombot** — Andrew's
+personality-first chat agent — into VS Code's native Chat panel as
+`@phantombot`.
+
+## How it works
+
+The extension is a thin bridge. It does **not** build its own chat UI: it
+registers a [Chat Participant](https://code.visualstudio.com/api/extension-guides/chat)
+(`@phantombot`) so you get VS Code's native scrollback, history and theming for
+free.
+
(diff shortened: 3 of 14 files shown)


kaieriksen (approved):
Reviewed the ACP VS Code extension. No blocking issues found.\n\nVerified locally:\n- editors/vscode typecheck passes\n- editors/vscode tests pass: 29 pass\n- VSIX packages successfully with Bun on PATH\n- root typecheck passes\n- root test suite passes: 1537 pass, 1 skip\n\nNotes: branch is behind main but merge-tree shows no conflicts; the first root test run only failed because this checkout's node_modules was missing the already-declared jsonc-parser dependency, and passed after refreshing deps.

PR #212, opened 2026-06-27
fix(phantomchat): stamp voice-note duration so bubbles show real length

andrewagrahamhodges:
Follow-up to #210 (two-way voice notes). Outbound voice notes rendered as **0:00** in the recipient's bubble even though they played fine — the voice file-metadata envelope carried no `duration`, and the PWA bubble defaults to 0 when it's absent.

## Fix
Compute the Ogg/Opus playback length at send time from the container's 48 kHz granule position (minus the OpusHead pre-skip, per RFC 7845) and include it as `duration` in the voice metadata envelope. It's a cheap container walk — no decode.

- `oggOpusDuration.ts` — `oggOpusDurationSeconds()`; returns 0 for non-Ogg input (callers omit the field).
- `transport.sendVoice()` — stamps `duration` when > 0.

## Tests
`oggOpusDuration.test.ts` builds minimal Ogg/Opus pages to pin the granule math (whole-second rounding, the never-round-to-zero clamp, -1 granule skip, non-Ogg → 0). Typecheck clean, phantomchat suite green.

### DIFF
--- a/src/channels/phantomchat/oggOpusDuration.ts
+++ b/src/channels/phantomchat/oggOpusDuration.ts
@@ -1,0 +1,49 @@
+/**
+ * Compute the playback duration (in seconds) of an Ogg/Opus stream.
+ *
+ * Opus granule positions run on a fixed 48 kHz clock (RFC 7845 §4),
+ * independent of the originally-encoded sample rate. The total decoded sample
+ * count is the granule position of the LAST Ogg page; subtracting the pre-skip
+ * declared in the OpusHead identification header gives the audible sample
+ * count. Duration = audibleSamples / 48000.
+ *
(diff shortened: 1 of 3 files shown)


kaieriksen (approved):
Reviewed the outbound PhantomChat voice-duration fix. The Ogg/Opus duration helper follows the Opus 48 kHz granule-position model, subtracts pre-skip, ignores -1 granule pages, and fails closed by omitting duration when parsing returns 0. The transport change only stamps metadata before encryption/upload and preserves the existing media envelope shape. Focused tests pass locally: `~/.bun/bin/bun test tests/channels-phantomchat-oggOpusDuration.test.ts tests/channels-phantomchat-channel.test.ts tests/channels-phantomchat-server.test.ts`.

PR #213, opened 2026-06-27
feat(acp): bundle VS Code extension as .vsix + auto-install (PR2)

robertclawson:
## What

PR2 of first-party VS Code support. The `@phantombot` chat-participant extension authored in #211 is now **built to a `.vsix`, embedded in the phantombot binary**, and **auto-installed** through the same `reconcileEditorConnectors()` flow Zed uses — **no Microsoft marketplace dependency**.

## How

- **`scripts/genVscodeVsix.ts`** — builds + packages `editors/vscode/` into a `.vsix` and emits `src/lib/vscodeExtensionAsset.generated.ts` (version, filename, base64, hash). Mirrors the existing pi-extension embed pattern so a single `bun build --compile` ELF carries the asset. Run: `bun run gen:vscode-vsix`.
- **`src/connectors/acp/installVscode.ts`** — pure, injectable detect / version-compare / install logic. Resolves the `code` CLI (PATH probe + per-platform install candidates; win32 uses `code.cmd`), stages the embedded `.vsix` to a temp file, runs `code --install-extension --force` **only when missing or outdated** (idempotent). `checkVscode()` is the non-mutating report-only variant.
- **`src/connectors/acp/autoInstall.ts`** — VS Code `EditorSpec` wired via an optional `reconcile` hook, so Zed's settings-write path stays **byte-identical** while VS Code owns its own version-compare logic. Fully error-isolated.
- **CLI + doctor** — `phantombot acp install vscode` and doctor reporting.

## Tests

- **36 new tests**: `installVscode` (version compare, idempotent skip, per-platform CLI resolution, missing-CLI fallback, install failure → error not throw, error isolation), `autoInstall-vscode` (reconcile dispatch, repair vs report-only, isolation alongside Zed), embedded-asset validity + version-drift.
- Full suite **1578 pass / 0 fail**, root + extension typechecks clean.
- ACP-only, no MCP — no repeat of the #202/#203 revert.
- Smoke-tested against the real **arm64** binary; ACP handshake green.

## Notes

Win32 branches are implemented + unit-tested but not smoke-tested end-to-end (no Windows build yet) — flagged honestly rather than claimed working.

Stacks on #211 (merged).


src/cli/acp.ts
```diff
@@ -37,13 +38,34 @@ const installZedCmd = defineCommand({
...
+    process.exit(result.code);
+  },
+});
+
 const installCmd = defineCommand({
   meta: {
     name: "install",
-    description: "Install the ACP registration into an editor's settings.",
+    description:
+      "Install the ACP registration into a detected editor (zed: settings merge; vscode: first-party extension).",
```
kaieriksen, line 64: Non-blocking: while touching this help surface, the parent `acp` description below still undersells the command after adding VS Code install support. `phantombot acp --help` still says "Zed spawns this" and only points users at `acp install zed`; worth broadening to "editors" and mentioning `acp install vscode` before merge or in a follow-up.

kaieriksen (approved):
Approved. I found no blocking issues.

Verified locally:
- Targeted VS Code auto-install tests: 36 pass
- VS Code extension package checks: typecheck + 29 tests pass
- Root typecheck clean
- Full root suite: 1578 pass, 1 skip, 0 fail
- `bun --compile` x64 binary builds and `phantombot acp install vscode` behaves correctly with no `code` CLI present
- Simulated `code` CLI install path writes a non-empty bundled VSIX and calls `--install-extension ... --force`

Only note is a non-blocking inline help-text cleanup in `src/cli/acp.ts`.

Issue #32, opened 2026-05-02
envFile: clear stale .env.tmp before write


andrewagrahamhodges:
Non-blocking follow-up from PR #31 review.

`writeFile(tmp, ..., { mode: 0o600 })` only applies the mode flag when the file is **created**. If a stale `.env.tmp` exists from a prior crash with looser perms, the new write reuses the existing inode and the mode flag is silently ignored — the rename then carries the looser perms onto the target.

In practice every `.env.tmp` we've ever written came from this same code path at 0o600, so realistic blast radius is ~zero. Still, a one-line hedge:

\`\`\`ts
await unlink(tmp).catch(() => {}); // clear any stale tmp from a prior crash
await writeFile(tmp, ..., { encoding: \"utf8\", mode: 0o600 });
\`\`\`

Touch \`src/lib/envFile.ts\` \`saveEnvFile()\` only. Add a test that pre-creates \`.env.tmp\` at 0o644 and asserts the post-update final file is 0o600.

kaieriksen:
Closing as obsolete — as noted in the issue, realistic blast radius is ~zero (every `.env.tmp` ever written came from this same code path at 0o600). Not worth a targeted PR.

PR #214, opened 2026-06-27
feat(phantomchat): voice-note waveform bars + "recording voice" indicator

andrewagrahamhodges:
## What

Two polish items on the **outbound** PhantomChat voice path, following the duration fix (#212).

### 1. Waveform bars
The bubble showed length but no amplitude bars because we never sent a `waveform`. This derives a Telegram-style envelope from **Opus packet sizes** — a cheap Ogg container walk, no audio decode (bigger packet ≈ louder, DTX/silence ≈ small) — resamples to 100 bars, packs them 5-bit little-endian exactly the way the PWA's `decodeWaveform` reads them back, and ships it base64 in the voice metadata. `""` ⇒ unparseable, field omitted (bubble still shows length).

### 2. "recording voice" indicator
The chat showed "typing…" while a voice reply synthesized. `sendRecording` now publishes its kind-20001 tick with content `"recording"` (new `TYPING_CONTENT_RECORDING` marker) instead of the empty typing marker, so the PWA renders the native `sendMessageRecordAudioAction` ("recording voice") activity.

## Pairs with
**Must land together with phantomchat PR → https://github.com/phantomyard/phantomchat/pull/32** — the receive side that decodes the base64 waveform into packed bytes and maps the `"recording"` tick to the record-audio action. Reviewing this PR alone won't show the decode; the full send→render path spans both repos.

## Tests
- Waveform round-trips through a mirror of the PWA's decoder (loud frames peak at 31, quiet frames dip; 100 bars → 63 bytes, all bars filled).
- `sendRecording` emits the `"recording"` content marker on a signed kind-20001.
- Full suite: typecheck clean, 1551 pass / 0 fail.

## Acceptance
Live two-way check: send a voice reply and confirm the bubble draws bars and the chat shows "recording voice" while synthesizing. Needs a binary rebuild to take effect.


src/channels/phantomchat/transport.ts
```diff
@@ -453,6 +468,7 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
       // Authoritative media class so the receiver never re-guesses voice vs file.
       mediaType: "voice",
       ...(durationS > 0 ? { duration: durationS } : {}),
+      ...(waveform ? { waveform } : {}),
```
kaieriksen, line 471: This emits the waveform as a base64 string, which matches the wire metadata shape, but I do not see a receiver-side conversion before the value reaches the PWA audio renderer. In PhantomChat, `extractFileMetadata` stores `waveform` as `string`, `buildPhantomChatMedia` assigns that string to `documentAttributeAudio.waveform`, and `components/audio.ts::decodeWaveform` only handles `Uint8Array | number[]` (a string coerces to an empty `Uint8Array`). That means this PR can generate valid bytes but still render no bars. Please either add/point to the decode step that converts the base64 string to bytes before `documentAttributeAudio.waveform`, or add an integration test covering the actual receiver path.

kaieriksen (changes requested):
I found one blocker: the generated waveform is valid base64, but the current PhantomChat receive/render path keeps `fileMetadata.waveform` as a string and passes it through to `documentAttributeAudio.waveform`. The PWA audio renderer only decodes `Uint8Array | number[]`; a string becomes an empty `Uint8Array`, so the bars do not render. The encoder itself and recording marker look fine once that integration boundary is covered.

andrewagrahamhodges:
@kaieriksen thanks for the careful read — you're right that a raw string would decode to an empty `Uint8Array` and draw no bars. The decode you're looking for does exist, it just lives in the sibling repo so it isn't visible from #214 alone:

- **phantomchat #32**, commit `e38dbf6` — `phantomchat-media-shape.ts:88` runs `base64ToBytes(fm.waveform)` (wrapped in try/catch) inside `buildPhantomChatMedia`, so the string is converted to bytes *before* it reaches `documentAttributeAudio.waveform`.
- Covered by `voice-media-classification.test.ts:63`, which round-trips `'q80='` → `[0xab, 0xcd]` and asserts the rendered attribute is a byte array, not a string.

So the integration boundary is handled — it's a two-PR pair (encoder here, decoder in #32) that has to land together. I'll cross-link them so the full path is reviewable from one place. Happy to add anything else you'd want to see on the boundary.

src/channels/phantomchat/transport.ts
kaieriksen: Verified against phantomyard/phantomchat#32: `buildPhantomChatMedia` now decodes `fm.waveform` with `base64ToBytes` before assigning `documentAttributeAudio.waveform`, and `voice-media-classification.test.ts` covers `q80=` -> `[0xab, 0xcd]` as a `Uint8Array`. That answers the integration-boundary concern for the paired landing, so I am resolving this thread.

kaieriksen (approved):
Approved. Max pointed to the paired receiver-side PhantomChat change in phantomyard/phantomchat#32, and I verified the decode path: `buildPhantomChatMedia` converts the base64 waveform string to `Uint8Array` before `documentAttributeAudio.waveform`, with a regression test for the packed-byte boundary. CI is green here, and my blocker thread is resolved.

PR #216, opened 2026-06-27
fix(acp): self-heal default persona instead of exit 2

robertclawson:
## What

When `phantombot acp` starts **without** an explicit `--persona`, it now falls back to `config.defaultPersona` and — if that persona dir is missing — **self-heals to any persona on disk**, mirroring the existing `phantombot run` behavior (`healDefaultPersonaIfBroken`). Previously it died with **exit code 2**.

## Why

A fresh box has no `default_persona` set, so phantombot falls back to the hardcoded built-in name `phantom` — whose persona dir was never created. The VS Code chat participant (and Zed) spawn `phantombot acp` with no `--persona`, so **every first-time editor user hit exit 2** and got a blank failure. Reported by Andrew hitting it in VS Code.

## Behavior

- **No `--persona`, default dir missing, ≥1 persona on disk** → heal to an existing persona, start cleanly. ✅
- **No `--persona`, zero personas on disk** → hard-error exit 2 with a `phantombot persona` hint (unchanged — nothing to fall back to).
- **Explicit `--persona NAME` missing** → still hard-errors exit 2 (honors the user's explicit choice; no silent substitution).

## Verification

- 3 new tests in `connectors-acp-server.test.ts` covering all three branches.
- Full suite **1590 pass / 0 fail**, root typecheck clean.
- **arm64 handshake smoke** against the real compiled binary: default `phantom` → healed `robbie`, `initialize` returns, exit 0, `state.json` persisted.

ACP-only, no MCP. Scoped to the persona-resolution gate in `runAcpServer`.

### DIFF
--- a/src/connectors/acp/server.ts
+++ b/src/connectors/acp/server.ts
@@ -23,6 +23,7 @@ import { join } from "node:path";
 import type { Readable, Writable } from "node:stream";
 
 import { type Config, loadConfig, personaDir } from "../../config.ts";
+import { healDefaultPersonaIfBroken } from "../../lib/personaDefault.ts";
 import { buildHarnessChain } from "../../harnesses/buildChain.ts";
 import { resolveHarnessBinsForConfig } from "../../lib/harnessAvailability.ts";
 import type { Harness } from "../../harnesses/types.ts";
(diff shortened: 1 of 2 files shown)


kaieriksen (approved):
Reviewed the ACP persona fallback change. No blockers found. Verified targeted ACP/persona tests, root typecheck, full test suite, and merge compatibility against current main.

PR #217, opened 2026-06-27
fix(acp): pin persona/config to real home under snap + add menu launchers

robertclawson:
## Summary

Fixes the strict-snap **exit 2** crash that breaks the VS Code chat connector on first use, and adds discoverability entry points for the `@phantombot` participant. Three related changes, one PR.

### Root cause (exit 2)
When VS Code is installed as a **strict snap** (the Ubuntu App Center build), snapd confines the editor and every process it spawns. Inside that sandbox `$HOME` is **redirected** from the real home (`/home/alice`) to a per-snap data dir (`/home/alice/snap/code/<rev>`). phantombot resolves its persona/config store from `$HOME`/`$XDG_*` at runtime (`src/config.ts`), and the redirected store is **empty** — so `phantombot acp` finds zero personas and exits 2 with *"no other personas exist"*, killing the connector. A **native** install (`.deb`, or Zed) sees the real `$HOME` and works — exactly the asymmetry observed.

The config resolver already honors absolute overrides `PHANTOMBOT_PERSONAS_DIR` and `PHANTOMBOT_CONFIG` (verified in `src/config.ts`: `loadConfig` reads `process.env.PHANTOMBOT_CONFIG` first; `personasDir` reads `process.env.PHANTOMBOT_PERSONAS_DIR` first). We use those seams.

### 1. Snap-aware ACP spawn env — the exit-2 fix
- New pure module `editors/vscode/src/snapEnv.ts`: detects snap confinement via `$SNAP` + `$SNAP_REAL_HOME` and returns a spawn env with absolute `PHANTOMBOT_PERSONAS_DIR` + `PHANTOMBOT_CONFIG` derived from `$SNAP_REAL_HOME`, pinning phantombot back to the **real** persona/config store.
- Wired into `spawnAcpTransport` in `editors/vscode/src/acpClient.ts` (was `env: process.env`).
- Non-destructive: returns a new object, never mutates input, never clobbers an explicit user override; outside a snap the env is returned unchanged.

### 2. Zed installer bakes the absolute overrides
- `src/connectors/acp/installZed.ts`: the registered `agent_servers.Phantombot.env` block (previously `{}`) now carries absolute `PHANTOMBOT_CONFIG` + `PHANTOMBOT_PERSONAS_DIR` via new `phantombotEnvOverrides()`. The installer runs as the real user on a native install, so it knows the real absolute paths — they match `config.ts`'s default resolution (`xdgConfigHome`/`xdgDataHome`) on a normal box and are insurance against a redirected child env.

### 3. Menu launchers (discoverability)
Registered VS Code commands so the participant is reachable beyond typing `@phantombot`:
- **"Chat with Phantombot"** — Command Palette entry **and** a title-bar button on the chat view → opens the Chat panel pre-filled with `@phantombot `.
- **"Ask Phantombot about this"** — editor right-click context-menu entry → opens chat pre-filled with `@phantombot ` plus a fenced snippet of the current selection.
- Query-building is pure (`editors/vscode/src/commands.ts`); `vscode` glue in `extension.ts`; manifest `contributes.commands` / `menus` / `activationEvents` in `editors/vscode/package.json`.

## Tests
- `editors/vscode/tests/snapEnv.test.ts` (new) — reproduces the redirected-`$HOME` exit-2 **trigger** (`isHomeRedirected`) and asserts the env-pinning fix, including override-respect and non-mutation.
- `editors/vscode/tests/commands.test.ts` (new) — covers the launcher query builders (fencing, language tag, empty-selection degradation).
- `tests/connectors-acp-installZed.test.ts` — updated to assert the baked `env` overrides instead of `{}`.
- All pure; no GUI, no real subprocess.

## Build / verify (run locally on bun 1.3.13)
- Root `bun test`: **1606 pass / 0 fail** (113 files).
- Root `bun tsc --noEmit`: clean.
- Extension `bun test`: **45 pass / 0 fail**; `tsc --noEmit`: clean; `bun run build` (esbuild): clean.
- `bun run build:x64` (compiled binary): clean.
- **Regenerated the bundled `.vsix`** (`src/lib/vscodeExtensionAsset.generated.ts`) via `bun run gen:vscode-vsix` so the new manifest (commands/menus) ships in the binary. Version stays `0.1.0` (content-only change; the asset test asserts version-match + valid-zip, both pass).

Do not merge — leaving for Andrew to review.


editors/vscode/src/snapEnv.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+export function snapAwareSpawnEnv(env: EnvMap): EnvMap {
+  if (!isSnapConfined(env)) return env;
+
+  const realHome = env.SNAP_REAL_HOME!.trim();
+  const next: EnvMap = { ...env };
+
+  // Respect explicit user overrides — they chose those on purpose.
+  if (!next.PHANTOMBOT_PERSONAS_DIR || !next.PHANTOMBOT_PERSONAS_DIR.trim()) {
+    next.PHANTOMBOT_PERSONAS_DIR = personasDirFor(realHome);
```
kaieriksen: This fixes the default-store snap case, but it also forces the default personas directory even when the real `config.toml` contains a custom `personas_dir`. Because `PHANTOMBOT_PERSONAS_DIR` has higher priority than the TOML setting, a user who configured `personas_dir = "/srv/phantombot/personas"` will still be launched against `<real home>/.local/share/phantombot/personas`, which can recreate the empty-store/no-personas failure under a different setup. I think we should pin `PHANTOMBOT_CONFIG` to the real home and avoid setting `PHANTOMBOT_PERSONAS_DIR` unless we can first prove there is no configured `personas_dir`, or otherwise derive it from the real config.

src/connectors/acp/installZed.ts
```diff
@@ -53,13 +54,40 @@ export interface InstallZedResult {
...
+ * default resolution on a normal box (and is therefore a no-op there, just an
+ * insurance policy against a redirected child env).
+ */
+export function phantombotEnvOverrides(): {
+  PHANTOMBOT_CONFIG: string;
+  PHANTOMBOT_PERSONAS_DIR: string;
+} {
+  return {
+    PHANTOMBOT_CONFIG: join(xdgConfigHome(), "phantombot", "config.toml"),
+    PHANTOMBOT_PERSONAS_DIR: join(
```
kaieriksen: Same precedence issue here: baking `PHANTOMBOT_PERSONAS_DIR` into Zed means the child process will ignore a user's configured `personas_dir` from `config.toml`. Since the installer already pins `PHANTOMBOT_CONFIG`, the safer behavior is to let `loadConfig()` read that config and resolve `personas_dir` normally, unless we intentionally add a migration path for custom persona roots.

kaieriksen (changes requested):
I found one blocker around preserving configured persona stores. The snap fix itself is the right direction, but the new environment override can bypass a user's existing `personas_dir` setting from `config.toml`.

robertclawson:
Thanks @kaieriksen — good catch, fixed in c4b1a55.

You were right that `PHANTOMBOT_PERSONAS_DIR` is an absolute override that wins over a custom `personas_dir` in `config.toml`, so pinning it would silently break anyone with a custom persona root. The snap escape only ever needed the config pinned.

Both seams now set **only** `PHANTOMBOT_CONFIG`:
- `snapEnv.ts` — dropped the `PHANTOMBOT_PERSONAS_DIR` assignment + `personasDirFor` helper; `snapAwareSpawnEnv` pins just `PHANTOMBOT_CONFIG` (still respecting any explicit user override).
- `installZed.ts` — `phantombotEnvOverrides()` no longer bakes `PHANTOMBOT_PERSONAS_DIR` into Zed's `env` block.

`loadConfig()` reads the pinned `PHANTOMBOT_CONFIG` and resolves `personas_dir` from it normally, so default **and** custom roots are honoured. Tests updated to assert `PHANTOMBOT_CONFIG` is set and `PHANTOMBOT_PERSONAS_DIR` is left unset. CI green. Re-requesting review.

editors/vscode/src/snapEnv.ts
```diff
@@ -0,0 +1,106 @@
...
+ */
+export function snapAwareSpawnEnv(env: EnvMap): EnvMap {
+  if (!isSnapConfined(env)) return env;
+
+  const realHome = env.SNAP_REAL_HOME!.trim();
+  const next: EnvMap = { ...env };
+
+  // Respect an explicit user override — they chose it on purpose.
+  if (!next.PHANTOMBOT_CONFIG || !next.PHANTOMBOT_CONFIG.trim()) {
+    next.PHANTOMBOT_CONFIG = configPathFor(realHome);
```
kaieriksen, line 145: This still leaves XDG_DATA_HOME redirected under the snap. In loadConfig(), a config without personas_dir falls back to join(xdgDataHome(), "phantombot", "personas"), so default installs still look under /home/alice/snap/code/.../.local/share/phantombot/personas and hit the empty store. Please restore the real XDG_DATA_HOME for the child env, or otherwise cover loadConfig’s default personasDir under the strict-snap env, while still leaving PHANTOMBOT_PERSONAS_DIR unset so custom personas_dir keeps working.

kaieriksen (changes requested):
The previous PHANTOMBOT_PERSONAS_DIR override regression is fixed, but there is still a blocker for the default persona store under strict-snap VS Code. The current env only pins PHANTOMBOT_CONFIG; loadConfig still derives default personasDir and memoryDbPath from redirected XDG_DATA_HOME when config.toml does not explicitly set personas_dir.

robertclawson:
Good catch — you're right, `PHANTOMBOT_CONFIG` alone didn't cover the default-install case. Confirmed against `src/config.ts`: `loadConfig` builds `dataDir = join(xdgDataHome(), "phantombot")` and the default `personasDir`/`memoryDbPath` (plus `state.json` and the memory-index) all fall back to that — independent of where `PHANTOMBOT_CONFIG` points. So a config.toml with no explicit `personas_dir` still resolved into the empty redirected sandbox store and hit exit 2.

Fixed in 1297274: `snapAwareSpawnEnv` now also forces `XDG_DATA_HOME` and `XDG_CONFIG_HOME` back to `$SNAP_REAL_HOME` for the spawned ACP child, overriding snapd's redirected values (snapd always populates them with sandbox paths, so the redirected value *is* the bug — nothing legitimate to preserve there). `PHANTOMBOT_PERSONAS_DIR` stays unset, so a custom `personas_dir`/`memory_db` in config.toml still takes precedence and keeps working.

Added tests: the default-install regression (no `personas_dir` → XDG_DATA_HOME restored), the force-over-snapd-redirect case, and helpers. Full vscode suite green (49 pass). PTAL.

kaieriksen (approved):
Latest revision addresses my blocker. The snap path now restores XDG_DATA_HOME/XDG_CONFIG_HOME to SNAP_REAL_HOME while leaving PHANTOMBOT_PERSONAS_DIR unset, so default installs and custom persona roots both resolve through loadConfig correctly.\n\nVerified locally:\n- PATH=/home/kai/.bun/bin:/home/kai/.local/bin:/home/kai/bin:/home/kai/.codex/packages/standalone/releases/0.142.3-x86_64-unknown-linux-musl/codex-path:/home/kai/.codex/tmp/arg0/codex-arg0v46CCu:/home/kai/.local/share/pi-node/bin:/home/kai/.local/share/pi-node/current/bin:/home/kai/.pi/agent/bin:/home/kai/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin bun test editors/vscode/tests/snapEnv.test.ts editors/vscode/tests/commands.test.ts tests/connectors-acp-installZed.test.ts\n- PATH=/home/kai/.bun/bin:/home/kai/.local/bin:/home/kai/bin:/home/kai/.codex/packages/standalone/releases/0.142.3-x86_64-unknown-linux-musl/codex-path:/home/kai/.codex/tmp/arg0/codex-arg0v46CCu:/home/kai/.local/share/pi-node/bin:/home/kai/.local/share/pi-node/current/bin:/home/kai/.pi/agent/bin:/home/kai/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin bun run typecheck\n- cd editors/vscode && PATH=/home/kai/.bun/bin:/home/kai/.local/bin:/home/kai/bin:/home/kai/.codex/packages/standalone/releases/0.142.3-x86_64-unknown-linux-musl/codex-path:/home/kai/.codex/tmp/arg0/codex-arg0v46CCu:/home/kai/.local/share/pi-node/bin:/home/kai/.local/share/pi-node/current/bin:/home/kai/.pi/agent/bin:/home/kai/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin bun run typecheck

PR #219, opened 2026-06-27
feat(acp): phantombot as a first-class VS Code chat agent

andrewagrahamhodges:
## What & why

Turns phantombot from an awkward `@phantombot` sub-participant into a **first-class agent in VS Code's chat-sessions panel** — its own entry next to Claude Code / Copilot CLI, **no `@mention`**, with **persistent history** and **image + file attachments**. This is the Zed "External Agent" experience, ported to VS Code.

Verified end-to-end on a live Carbon X1 (snap-installed `code`, classic confinement) — reproduced the failures on-box, patched, rebuilt, reinstalled, confirmed working each step.

### 🐛 Classic-snap startup fix (was blocking everything)
Ubuntu's `code` is a **classic** snap: it keeps `$HOME` real but **redirects `$XDG_DATA_HOME`** into the snap sandbox. phantombot resolved personas from the empty sandbox store → fell back to `phantom` → **`exit 2`**. `#217` only handled **strict** snaps (gated on `$SNAP_REAL_HOME`, which classic never sets), so it never fired here.
- `isSnapConfined()` now keys on `$SNAP` alone (set under both modes)
- new `realHomeFor()`: `$SNAP_REAL_HOME` if present (strict), else `$HOME` (classic)

### ✨ Chat sessions provider — the first-class agent
- `chatSessionsProvider` proposed API: item provider (one durable session per workspace) + content provider
- History rehydrated from **phantombot's own server-side memory** via ACP `session/load`, rendered with `ChatRequestTurn2` / `ChatResponseTurn2`
- phantombot's memory is the **single source of truth**; VS Code's replayed transcript is dropped (no doubling)

### ✨ Language Model Chat Provider — Copilot fallback
- For users who *do* have Copilot, phantombot also registers as a selectable model (native scrollback, model dropdown)
- `latestUserPrompt()` keeps only the newest user turn and bins VS Code's replayed history/context

### 📎 Attachments
- Dragged **files** → stable `Uri` → ACP reference blocks
- Pasted/dragged **images** → `chatReferenceBinaryData` proposed API → ACP image blocks

### 🔧 Plumbing
- Vendored proposed-API `.d.ts` at the exact build commit `1b50d58`
- Proposed APIs enabled via `~/.vscode/argv.json` (supported side-load path; needs cold restart)
- Version `0.1.0` → `0.3.1`, embedded `.vsix` regenerated

## Testing
- Pure cores (`sessionBridge`, `lmBridge`, `snapEnv`) unit-tested — **79 extension tests**, typecheck clean
- Repo install/asset guards green; full suite **1640 pass, 0 fail**

## Reviewer notes
- **Three layers ship together** by design: sessions provider (primary), LM provider (Copilot fallback), `@phantombot` participant (always-available fallback). Each no-ops cleanly if its API/proposal isn't present.
- The embedded `.vsix` asset is one generated blob, so the snap fix and the sessions work share a single commit — they ship in one binary.
- Relies on **proposed APIs** (`chatSessionsProvider`, `chatReferenceBinaryData`, `chatParticipantPrivate`). Pinned to a build commit; a future VS Code update could shift these — acceptable since we side-load our own `.vsix`.


editors/vscode/src/sessionBridge.ts
```diff
@@ -0,0 +1,175 @@
...
+ */
+export function promptBlocksFromRequest(
+  text: string,
+  attachments: readonly SessionAttachment[] = [],
+): AcpContentBlock[] {
+  const blocks: AcpContentBlock[] = [];
+
+  const trimmed = text.trim();
+  if (trimmed.length > 0) {
+    blocks.push({ type: "text", text });
```
robertclawson: Nit (non-blocking): the guard checks `trimmed.length > 0` but the block pushes the untrimmed `text`, so `"  hi  "` is forwarded with its surrounding whitespace intact. `latestUserPrompt` in `lmBridge.ts` trims before sending (`texts.join("\n\n").trim()`), so the two prompt paths are slightly inconsistent. Pushing `trimmed` instead of `text` here would line them up. Harmless today — just tidiness.

editors/vscode/src/sessionBridge.ts
```diff
@@ -0,0 +1,175 @@
...
+  handlers: PromptHandlers;
+  turns: ReplayTurn[];
+} {
+  const turns: ReplayTurn[] = [];
+
+  const push = (role: "user" | "assistant", text: string): void => {
+    if (!text) return;
+    const last = turns[turns.length - 1];
+    if (last && last.role === role) {
+      last.text += text;
```
robertclawson: Nit (non-blocking, latent): when two consecutive chunks share a role they're concatenated with `last.text += text` — no separator. The doc comment notes this is defensive against a future server streaming one turn in pieces, which is fine. But if a future server ever emitted two *distinct* same-role turns back-to-back, they'd glue into a single bubble with no boundary. Not triggerable today (server alternates roles, one chunk per turn). Flagging only so it's a conscious choice — a `"\n\n"` join would make it robust to that edge.

robertclawson (dismissed):
Reviewed end-to-end — clean, well-built PR, no blockers. ✅

**Snap fix is correct.** Keying `isSnapConfined` on `$SNAP` alone catches Ubuntu's *classic* `code` snap (the real-world failure mode #217 missed), and the strict-snap path is not regressed — `$SNAP_REAL_HOME` still wins where present, with a dedicated classic-snap regression test.

**No history doubling.** `session/load` only reads memory, the provider loads once behind a guard, and `latestUserPrompt` drops VS Code's replayed turns — phantombot's memory stays the single source of truth.

**Security clean.** `spawn` uses an argv array (no shell), persona passed as a discrete arg, no env leakage, no path traversal (server never reads attachment files; the agent does via its own sandboxed tools).

**The embedded `.vsix` is genuinely regenerated** — valid zip, version 0.3.2 matches source, bundled `extension.js` contains the new providers + fixed snap logic. Not stale.

Two non-blocking nits inline (whitespace fidelity + a latent coalescer edge). Neither is triggerable in current behaviour. One caveat the author already flags: proposed VS Code APIs pinned to commit `1b50d58` — fragile if VS Code shifts them, but every usage is feature-detected and degrades to plain `@phantombot`, so it can't crash activation. Acceptable.

Approving.

andrewagrahamhodges:
Thanks for the thorough review @robertclawson 🙏 Both non-blocking nits addressed in `883bca2`:

- **Whitespace consistency** — `promptBlocksFromRequest` now pushes `trimmed` instead of `text`, lining it up with `latestUserPrompt` in `lmBridge.ts`.
- **Coalescer edge** — same-role chunks now join with `"\n\n"` so two distinct same-role turns can never glue into one bubble.

Updated the coalescing test to match, typecheck + 73 extension tests + 36 repo guards all green. Bumped 0.3.2 → 0.3.3 and regenerated the embedded `.vsix`.

robertclawson (approved):
Re-approve ✅ Both non-blocking nits cleanly addressed in 883bca2 — `promptBlocksFromRequest` now pushes `trimmed`, the coalescer joins same-role turns with `\n\n`, and the test was updated to match. typecheck, tests, and CodeQL all green. Good to merge.

robertclawson (approved):
Re-review of `883bca2bd` — my earlier approval was dismissed by the new commit. Both nits from my prior review are now resolved exactly as suggested:

- **`sessionBridge.ts:128`** — now pushes `text: trimmed` (consistent with `lmBridge`).
- **`sessionBridge.ts:84`** — coalescer separates same-role chunks with `\n\n` instead of gluing them; test updated to assert `"part one\n\npart two"`.

Verified the regenerated `.vsix` (v0.3.3) is genuinely fresh — decoded the base64 asset, unzipped it, and the bundled `dist/extension.js` carries the new `last.text += `\n\n${text}`` logic, so the asset matches source. CI all green (test, typecheck, CodeQL, both Analyze runs). No blockers — approving.

PR #220, opened 2026-06-28
fix(vscode): keep phantombot visible in folderless windows

andrewagrahamhodges:
Follow-up to #219.

## Problem
`provideChatSessionItems` built one chat-session item **per open workspace folder**. A window with **no folder** — an empty window, or a folderless `.code-workspace` (Andrew runs an `agent-sessions.code-workspace` with `"folders": []`) — produced an **empty list**, so the **Phantombot** agent vanished from the chat sessions panel entirely. Claude and Copilot-CLI stayed visible because they do not tie their entry to workspace folders.

Confirmed from the live box: the active SESSIONS window mapped to a workspace with `"folders": []`, the extension activated and `registerChatSession*` succeeded (output channel: "chat session provider registered"), yet no item showed — because the item list was empty.

## Fix
New pure helper `resolveSessionCandidates(folders, fallbackCwd)` in `sessionBridge.ts`:
- one item per open folder (unchanged), **or**
- a single `currentCwd()`-bound item when no folders are open.

Phantombot is now always visible, matching Claude/Copilot-CLI.

## Tests / verification
- 3 new bridge tests · **76 extension tests pass** · typecheck clean
- regenerated embedded `.vsix` (**0.3.3 → 0.3.4**) · full repo suite **1637 pass, 0 fail**
- installed `phantomyard.phantombot-vscode@0.3.4` on the box and confirmed it shows in a folderless window

PR #221, opened 2026-06-28
feat(vscode): sticky auto-open + quick-launch button for phantombot chat

andrewagrahamhodges:
## Why

Two ergonomics gripes with the new first-class chat agent:

1. **"I have to go fish for phantombot every time I open VS Code."** It never restored the session on startup.
2. **"Give me a little bot button as a shortcut"** (like the one we removed when we dropped the `@mention` participant).

This PR fixes both.

### 1. Quick-launch command — `phantombot.openSession`
A single command surfaced three ways so it's always within reach:
- 🤖 **Hubot button** in the chat view title bar (`view/title`, chat view `workbench.panel.chat.view.copilot`)
- **Command Palette**: "phantombot: Open Phantombot Chat"
- **Keybinding**: `Ctrl+Alt+P` / `Cmd+Alt+P`

It runs VS Code's auto-registered per-session-type command `workbench.action.chat.openNewSessionSidebar.phantombot` (VS Code derives the session resource from the contributed type), falling back to the editor variant and then to focusing the sessions viewer. Best-effort — logs and gives up rather than throwing.

### 2. Sticky startup — `phantombot.openOnStartup`
New setting: `never` | `ifPreviouslyOpened` | `always`, **default `ifPreviouslyOpened`**.

The session content provider now reports when phantombot is opened; we persist that in `globalState` and, at `onStartupFinished`, re-open it if policy says so. So: pick phantombot once → it keeps coming back. Leave it closed → it stays out of the way.

The decision is a pure, unit-tested helper (`shouldAutoOpenSession`); only the side effect (executing the open command) touches `vscode`.

## Notes for review
- **Stacked on #220** (folderless fix) so the installed build doesn't regress — base will auto-retarget to `main` once #220 merges.
- No new proposed APIs; the per-type open commands are stable workbench commands.
- Verified live on the Carbon: built, installed `phantomyard.phantombot-vscode@0.4.0`, sticky + button working.

## Tests
- 4 new `shouldAutoOpenSession` cases · extension tests **80 pass**
- Full repo suite **1641 pass, 0 fail** · typecheck clean
- Bumped 0.3.4 → 0.4.0, regenerated the embedded `.vsix`


andrewagrahamhodges:
Follow-up fix (0.4.1): the button and sticky auto-open were silently no-opping on the live box.

Root cause from the running VS Code logs: the per-type open commands (`workbench.action.chat.openNewSessionSidebar.phantombot`) are only registered when the `chatSessions` contribution declares `canDelegate: true` — `ChatSessionsService._enableContribution` gates `_registerCommands` behind that flag. We didn't set it, so every fallback hit 'command not found'.

- Added `canDelegate: true` to the contribution (same as copilot/copilotcli).
- Resolve the open command from the live command list with a startup retry for the registration race; real fallback is `focusAgentSessionsViewer` (the old `workbench.view.sessions.chat.focus` never existed).
- Sticky state was persisting fine all along — it only looked broken because the re-open command failed.

+6 bridge tests, full suite 1647 pass. Verified installed as 0.4.1.

andrewagrahamhodges:
Follow-up fix (0.4.2): the `canDelegate: true` added for the quick-launch commands broke chat with `No activated agent with id "phantombot"`.

**Root cause:** `canDelegate: true` makes VS Code register a *delegate agent* whose id is the chatSession `type` (`phantombot`) and route session requests to it via `agentIdSilent`. That agent has **no implementation** until an extension creates a participant with the **same id**. Ours was `phantombot.chat` — mismatch — so the agent stayed implementation-less and every turn threw.

**Fix:**
- `PARTICIPANT_ID` → `phantombot` so the implementation attaches to the delegate agent (main-thread `$registerAgent` skips the must-be-declared throw when a session contribution of that type exists, then routes to `registerAgentImplementation` on the existing agent).
- Removed the manifest `chatParticipants` entry — a declared participant eagerly calls `registerAgent(id)` which **throws on a duplicate id**, so it can't share the session type. Implementation is attached dynamically at runtime instead (mirrors how Copilot CLI / Claude wire their session agents).
- Made the participant handler attachment-aware so the agent-impl path carries images/files like the session content provider already does.
- Added guard tests: `PARTICIPANT_ID == session type`, no colliding participant, blocks-precedence in the bridge.

Verified on-box: 91 extension tests, full repo suite 1652 pass / 0 fail, installed `0.4.2`.

editors/vscode/src/extension.ts
```diff
@@ -130,10 +172,29 @@ export function activate(context: vscode.ExtensionContext): void {
           ?.trim() ?? "",
       participant,
       participantId: PARTICIPANT_ID,
+      // Remember that phantombot was opened so we can re-open it next launch.
+      onSessionOpened: () => {
+        void context.globalState.update(STICKY_KEY, true);
```
kaieriksen, line 178: This only ever writes `true`, so the default `ifPreviouslyOpened` mode becomes “open forever after the first successful open”. Closing the Phantombot session never clears `phantombot.sessionWasOpen`, which breaks the documented sticky behavior in this PR (“leave it closed and it stays out of the way”). We need either a reliable close/not-open signal to reset the key, or the setting/description should stop promising last-window-state stickiness.

kaieriksen (changes requested):
Thanks. I found one blocker: the sticky state is write-only true, so `ifPreviouslyOpened` turns into “always open after first use” instead of tracking whether Phantombot was left open. Also, GitHub currently marks the branch as conflicting with main, so it cannot be approved/merged yet.

andrewagrahamhodges:
Thanks Kai — both addressed in `e70dc35`.

**1. Write-only sticky flag.** You're right: the key only ever went `true`, so `ifPreviouslyOpened` was really "open forever after first use" while the description promised last-window-state behaviour. I went with your second option (make the setting honest) rather than the first, because a *reliable* close/last-shutdown signal isn't available here: the chat-session surface opens in the **sidebar** (a view, not an editor tab), so there's nothing trustworthy to watch for "was it open when I quit". Faking it would just move the lie.

So the contract is now what it actually does, and what the user asked for ("be there if I picked it before"):
- mode `ifPreviouslyOpened` → **`ifUsedBefore`**
- globalState key `sessionWasOpen` → **`everOpened`**
- setting + enum descriptions rewritten: *"once you've opened Phantombot it keeps coming back on future launches; set to `never` to stop, `always` to open every launch."*

No migration needed — the setting is introduced in this PR, unreleased.

**2. Conflict with main.** Resolved — #220 (folderless fix) squash-merged, so I merged `main` in, kept the session/command additions, and regenerated the embedded `.vsix` at **0.4.3**. GitHub now reports `MERGEABLE`.

91 extension tests · full repo suite **1669 pass / 0 fail**. Ready for another look.

kaieriksen (approved):
Looks good now. The sticky behavior is documented honestly as ifUsedBefore/everOpened, the participant/session id contract is guarded, CI is green, and focused local checks pass.

PR #222, opened 2026-06-28
fix(update,typing): route /update confirmation back to PhantomChat + harden opening typing indicator

robertclawson:
### 1. `/update` from PhantomChat replied on Telegram, not PhantomChat 🐛
**Root cause:** the pending-update marker only carried a numeric Telegram `chatId`. A PhantomChat conversation id is a 64-char hex pubkey, so `Number(hex)` → `NaN` → serialized `null` → on restart the post-restart notify fell through to **broadcasting "✅ Updated to vX" to the Telegram allowlist**. (This is the channel-abstraction nit Kai flagged on #168.)

**Fix:** the marker now carries the channel-neutral `conversation` key (`telegram:42` / `phantomchat:<hex>`), which is the authoritative router.
- `notifyPostRestartIfPending` (Telegram) now **defers** a PhantomChat marker — it neither sends nor clears it — so it can never leak onto Telegram.
- New `notifyPhantomchatPostRestart`, run once per persona after its relay transport is up, DMs the confirmation **back to the originating pubkey** over Nostr and clears the marker.
- Multi-persona safe: only the persona that actually handled `/update` claims its marker.

### 2. Flaky typing / "recording voice" indicator 🐛
The #33 receive-side `ensureUser` fix was necessary but not sufficient — this is the **send side**. The opening indicator tick races relay connection state: on PhantomChat the tick is a *fire-and-forget ephemeral Nostr event* (kind-20001), so if the relay socket isn't OPEN at `t=0` it's **silently dropped**. On a fast turn (reply lands before the 2s throttle fires a second refresh) that single dropped tick means **no dots at all** — exactly the intermittent failure.

**Fix:** `emitOpeningIndicator` emits the opening tick immediately **and once more after a short delay** (`OPENING_INDICATOR_REEMIT_MS = 600`), by which point the pool has had a beat to (re)connect — so a single cold-socket drop can no longer swallow the whole indicator. Cancelled on turn completion so it never fires into a finished turn. Idempotent/harmless on Telegram.

## Regression tests
- **`tests/channels-core-engine-indicator.test.ts`** (new): the double-pulse contract with **injected timers** (deterministic, no wall-clock waits) — immediate emit, exactly one scheduled re-emit at the configured delay, the re-emit fires a second tick, `cancel()` prevents it, cancel is idempotent.
- **`tests/lib-updateNotify.test.ts`**: `conversation` persisted into the marker (incl. the NaN-chatId PhantomChat case); Telegram notify **defers** a PhantomChat marker (no Telegram leak, marker preserved); PhantomChat notifier sends to the right pubkey + clears; wrong-persona and Telegram-origin markers left untouched; send-failure preserves the marker for retry; channel-classification helper.

## Validation
- Full suite: **1651 pass / 0 fail** (+17 vs baseline).
- `bun tsc --noEmit`: clean.
- `bun build --compile` (arm64): OK.

## Note on validating the typing fix
Fix #2 is the most probable cause of the remaining flakiness (cold-relay drop of the single opening tick), but it can't be 100% reproduced without your device. The regression test locks in the double-pulse behaviour; the real-world confirmation is you on a fresh build. Since phantombot ships as a binary (not Pages), that means a rebuild/redeploy after merge — happy to do that and have you sanity-check before we call it.

### DIFF
--- a/src/channels/commands.ts
+++ b/src/channels/commands.ts
@@ -260,10 +260,13 @@ async function handleUpdate(
   const r = await runUpdateFlow({
     config: ctx.config,
     currentVersion: VERSION,
-    // The update-notify subsystem keys recipients by numeric Telegram chat
-    // id (it persists the id into the post-restart marker). Convert the
-    // channel-neutral string conversation id back at this boundary.
+    // Telegram-only numeric id, kept for the Telegram post-restart path. For a
+    // PhantomChat conversation `ctx.chatId` is a hex pubkey → NaN here, which is
+    // exactly why `conversation` below is the authoritative router: it carries
+    // the channel-neutral key ("telegram:42" / "phantomchat:<hex>") so the
+    // confirmation lands back where `/update` was typed, not on Telegram.
     chatId: Number(ctx.chatId),
+    conversation: ctx.conversation,
     persona: ctx.persona,
   });
   return { reply: r.reply, afterSend: r.restart };
--- a/src/lib/updateNotify.ts
+++ b/src/lib/updateNotify.ts
@@ -80,6 +80,16 @@ export interface PendingUpdate {
    */
   chatId?: number;
   /**
+   * Channel-neutral conversation key of the chat that issued `/update`,
+   * e.g. "telegram:42" or "phantomchat:<64-hex>". This is the AUTHORITATIVE
+   * routing field — `chatId` above only works for Telegram (a numeric id),
+   * so a PhantomChat-origin `/update` (whose conversation id is a hex pubkey)
+   * used to coerce to NaN→null and wrongly fall back to a Telegram broadcast.
+   * The post-restart notify reads the prefix here to pick the right channel.
+   * Optional so markers written by older binaries still parse.
+   */
+  conversation?: string;
+  /**
    * Persona whose Telegram bot handled `/update`. Used after restart to
    * send the confirmation through the same bot in hybrid default+persona
    * configs. Optional so markers written by older versions still work.
@@ -128,7 +138,9 @@ export async function readPendingUpdate(
       typeof parsed?.targetTag !== "string" ||
       typeof parsed?.previousVersion !== "string" ||
       typeof parsed?.writtenAt !== "string" ||
-      (parsed.persona !== undefined && typeof parsed.persona !== "string")
+      (parsed.persona !== undefined && typeof parsed.persona !== "string") ||
+      (parsed.conversation !== undefined &&
+        typeof parsed.conversation !== "string")
     ) {
       log.warn("updateNotify: pending-update marker missing required fields", {
         path,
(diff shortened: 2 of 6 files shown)


src/channels/commands.ts
```diff
@@ -260,10 +260,13 @@ async function handleUpdate(
...
     currentVersion: VERSION,
-    // The update-notify subsystem keys recipients by numeric Telegram chat
-    // id (it persists the id into the post-restart marker). Convert the
-    // channel-neutral string conversation id back at this boundary.
+    // Telegram-only numeric id, kept for the Telegram post-restart path. For a
+    // PhantomChat conversation `ctx.chatId` is a hex pubkey → NaN here, which is
+    // exactly why `conversation` below is the authoritative router: it carries
+    // the channel-neutral key ("telegram:42" / "phantomchat:<hex>") so the
+    // confirmation lands back where `/update` was typed, not on Telegram.
     chatId: Number(ctx.chatId),
```
andrewagrahamhodges, line 268: Non-blocking nit: now that `conversation` is the authoritative router, this `chatId: Number(ctx.chatId)` still persists a junk `NaN` field into the marker for PhantomChat origins. Harmless today, but passing `undefined` when `ctx.chatId` isn't a real numeric Telegram id would keep the marker clean rather than carrying a junk value forward.

src/lib/updateNotify.ts
```diff
@@ -586,6 +648,14 @@ export async function notifyPostRestartIfPending(
   const marker = await readPendingUpdate(input.pendingPath);
   if (!marker) return { status: "no_marker" };
 
+  // Non-Telegram origin (e.g. PhantomChat): leave the marker in place so the
+  // owning channel's post-restart notify routes the confirmation back to the
+  // conversation that asked. Do NOT clear and do NOT broadcast to Telegram —
+  // that broadcast was the "/update from PhantomChat answered on Telegram" bug.
+  if (pendingUpdateChannel(marker) === "phantomchat") {
```
andrewagrahamhodges, line 655: Non-blocking nit: this defers only on `=== "phantomchat"`. `pendingUpdateChannel` can also return `"other"` (unknown/future prefix), and that case falls through to the Telegram broadcast below. Purely theoretical today (no such channel exists), but if you ever add one it'd silently answer on Telegram again. Deferring on `!== "telegram"` would be future-proof.

src/lib/updateNotify.ts
```diff
@@ -586,6 +648,14 @@ export async function notifyPostRestartIfPending(
...
+    return { status: "deferred_other_channel", marker };
```
andrewagrahamhodges, line 656: Non-blocking nit — stale-marker risk: a PhantomChat marker is intentionally left in place here for the owning channel to clear. If that persona never starts (or a legacy no-persona marker exists), nothing ever clears it — there's no GC path. Low impact, but worth a follow-up TODO so markers can't linger indefinitely.

andrewagrahamhodges (approved):
Reviewed end to end — diff, tests, and merge state. **Approving: solid, ship-worthy.**

Both fixes target the real root cause, and the tests are genuinely good — deterministic with injected timers, no flaky wall-clock waits.

**Fix 1 — /update routing** 👍 The `Number(hex)→NaN→null→Telegram broadcast` diagnosis is exactly right, and the channel-neutral `conversation` key as the authoritative router is the clean fix. Nice that the Telegram notify *defers* (doesn't send, doesn't clear) rather than handling it, and that only the persona named in the marker claims it.

**Fix 2 — typing indicator** 👍 Double-pulse (immediate + one re-emit at 600ms, cancelled on completion) is the right shape for a fire-and-forget ephemeral kind-20001 that can drop on a cold relay socket. Idempotent on Telegram, can't fire into a finished turn.

Left three non-blocking nits inline. None hold up the merge.

PR #225, opened 2026-06-28
fix(vscode): session shows "57 yrs ago" + auto-archives on newer VS Code

andrewagrahamhodges:
## Symptom

On the Mac, the phantombot session appeared under **ARCHIVED** in the SESSIONS panel labelled **"· 57 yrs ago"**, and the active sessions list was empty — even though chatting via the 🤖 button worked fine. On the Linux dev box (VS Code 1.124.0) it looked normal.

## Root cause

The chat-session **item provider never set `ChatSessionItem.timing`**. Newer VS Code builds render `timing.created` as the session's age; when it's absent the field defaults to the **Unix epoch (0)** → "57 yrs ago" → the build auto-archives the "ancient" session out of the active list. The older 1.124.0 build we developed against didn't render session age at all, which is exactly why this only bit on the Mac.

It's a timestamp bug — just not a macOS/Linux *date-format* difference (Andrew's first instinct); it's a missing field that newer VS Code penalizes.

## Fix

Populate `timing.created` with a **stable, persisted** per-resource timestamp:

- new pure helper `resolveSessionCreated(key, store, now)` — first sighting records `now`; later sightings return the stored value (so the age stays put across list refreshes instead of resetting to "just now"); a stored `0` is treated as unset so we never emit the epoch
- `SessionProviderDeps.sessionCreatedAt(key)` backed by `globalState`
- item provider now emits `timing: { created: deps.sessionCreatedAt(path) }`

## Verification

- 4 new bridge tests · **95 extension tests** · typecheck clean
- full repo suite **1687 pass / 0 fail**
- bumped **0.4.3 → 0.4.4**, regenerated the embedded `.vsix` asset

## Note for reviewers

VS Code's archive state is stored **per-resource**, so the *already-archived* session on the Mac will likely need a one-time manual unarchive; sessions created after this lands get a sane timestamp and won't be auto-archived. Couldn't reproduce on the Linux dev box (older VS Code), so Mac verification is the real test.


andrewagrahamhodges:
Added a second, independent fix on this branch (0.4.4 → **0.4.5**): clears the Copilot **"Upgrade" entitlement gate** that greys out the phantombot entry in the New Chat Session picker on VS Code **1.126+** (Copilot Free/Edu plans).

**Root cause (read from the real 1.126.0 bundle):** the picker gate `WLo(chatSessionsService, entitlementService, languageModelsService, type)` returns *not-locked* (0) when `chatSessionsService.supportsAutoModelForSessionType(type)` is true — which simply reads the contribution flag `supportsAutoModel`. On Free/Edu entitlement (5/10) any `canDelegate` type without that flag gets locked → "Upgrade".

**Fix:** declare `"supportsAutoModel": true` on the `chatSessions` contribution.
- Clears the gate **without touching `canDelegate`** — quick-launch button + sticky auto-open are unaffected.
- Honest semantics for us: we route to phantombot's own backend and ignore VS Code's model picker, so "auto model" is accurate.
- `supportsAutoModel` is a declared schema property in 1.126; it does not exist in 1.124.0's schema, so it is harmlessly ignored there (verified: 0.4.5 installs + activates cleanly on 1.124.0).

Locked the invariant with a manifest guard test. **96 extension tests · full repo suite 1688 pass / 0 fail.**

Note: I could not reproduce the gate on my box (ThinkPad runs 1.124.0, which only gates the built-in CopilotCloud type) — the diagnosis is from static analysis of the 1.126.0 workbench bundle. The Mac (1.126.0, Copilot Free) is the real proving ground.

editors/vscode/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "phantombot-vscode",
   "displayName": "phantombot",
   "description": "Chat with phantombot — your personality-first agent — in VS Code's native Chat panel via an embedded ACP client.",
-  "version": "0.4.3",
+  "version": "0.4.5",
```
robertclawson, line 5: PR body says "bumped 0.4.3 → 0.4.4", but this is **0.4.5** — the second commit (Copilot `supportsAutoModel` gate) isn't described in the body at all. Worth updating the description / squash message so both fixes land in the changelog.

robertclawson (commented):
## Review — reads clean, no blockers ✅

Traced both commits end to end; CI green (typecheck, test, CodeQL, both Analyze) and I decoded the embedded `.vsix` to confirm it's genuinely regenerated at **0.4.5** carrying *both* fixes (not stale).

**Commit 1 — timing fix (`129d0cd`):** correct root-cause. `resolveSessionCreated` is a clean pure helper, persisted in `globalState` so age stays stable across list refreshes, and the `existing > 0` guard means a stored epoch-0 is treated as unset so we never re-emit "57 yrs ago". Tests cover the four cases that matter (first-seen, stable-on-refresh, epoch-as-unset, per-key independence). The provider wiring is trivial. 👍

**Commit 2 — Copilot `supportsAutoModel` gate (`f1787a3`):** minimal and honest — we route to our own backend and ignore VS Code's model picker, so the flag is truthful, and it clears the 1.126+ "Upgrade" gate without touching `canDelegate` (button + sticky auto-open survive). `@types/vscode` bumped to `^1.104.0` stays ≤ `engines.vscode ^1.124.0`, so the 1.124 dev box is unaffected. Manifest guard test locks the invariant.

**Two non-blocking notes (inline):**
1. PR body only describes the timing fix and says "bumped 0.4.3 → 0.4.4" — but HEAD is **0.4.5** because commit 2 adds the Copilot gate fix + another bump. Worth folding commit 2 into the description / squash message so the changelog reflects both.
2. `supportsAutoModel` clears the gate by leaning on VS Code's internal `supportsAutoModelForSessionType` check — undocumented behavior that could shift in a future build. The manifest test guards *presence* of the flag, not the gate behavior; fine for now, just a latent dependency to remember if the gate reappears after a VS Code upgrade.

No blockers — happy to flip to **Approve** on your word. Note the one caveat from your body still stands: the *already-archived* Mac session needs a one-time manual unarchive; only sessions created after this lands get the sane timestamp.

editors/vscode/package.json
```diff
@@ -39,6 +39,7 @@
         "description": "Your personality-first agent — persona, memory and tools live server-side.",
         "icon": "$(hubot)",
         "canDelegate": true,
+        "supportsAutoModel": true,
```
robertclawson, line 42: Honest flag for us (we ignore VS Code's model picker and route to phantombot's backend), and it clears the 1.126+ "Upgrade" gate without touching `canDelegate` — good. Just flagging it leans on VS Code's internal `supportsAutoModelForSessionType` gate, which is undocumented and could change across builds. The manifest test guards the flag's presence but not the gate behavior; non-blocking.

lenaparkhodges (approved):
## Review — clean, approve ✅

Two commits, both fixing real VS Code integration bugs.

**Commit 1 — timing fix:** `resolveSessionCreated` is a clean pure helper — first-seen records `now`, later sightings return the stored value, stored 0 treated as unset so we never re-emit the epoch. The `globalState` wiring with `CREATED_KEY_PREFIX` namespacing is correct, and the `void` on `globalState.update` (fire-and-forget) is fine in practice — worst case on a crash is the timestamp gets re-recorded on next launch. Tests cover the four cases that matter (first-seen, stable-on-refresh, epoch-as-unset, per-key independence). Good test quality — real assertions on the real function, not tautological.

**Commit 2 — `supportsAutoModel`:** honest flag (we route to our own backend, ignore VS Code's model picker), clears the 1.126+ gate without touching `canDelegate`. Manifest guard test locks the invariant.

Robbie's two open threads (PR body version mismatch + undocumented VS Code internal dependency) are the only should-fix items — neither is a blocker. Ship it.

editors/vscode/package.json
andrewagrahamhodges: Good catch — folded both fixes into the squash-merge message so the changelog reflects the timing fix *and* the Copilot supportsAutoModel gate at 0.4.5. Resolving.

editors/vscode/package.json
andrewagrahamhodges: Agreed — noted as a latent dependency on VS Code internal supportsAutoModelForSessionType. Captured it so if the Upgrade gate reappears after a future VS Code bump we know where to look. Non-blocking, resolving.

PR #229, opened 2026-06-28
feat(memory): OKF field-weighted BM25 + link-graph expansion for the no-Gemini path

andrewagrahamhodges:
## What & why

Personas **without a Gemini API key** fall back to FTS5/BM25-only memory retrieval. Today that's a plain single-column BM25 over note bodies. This PR upgrades that default path with three [Open Knowledge Format](https://github.com/GoogleCloudPlatform/knowledge-catalog/tree/main/okf) (OKF) native superpowers — **without touching the Gemini hybrid (BM25+vector) path**.

OKF isn't a retrieval engine; it's a storage shape (markdown + YAML frontmatter + concept links). Our `kb/` second brain is already that shape, so each OKF feature maps directly onto a BM25 upgrade:

| OKF feature | BM25 upgrade |
|---|---|
| Frontmatter fields | **BM25F** — weight title/tags/aliases above body |
| Tags + aliases | **Controlled vocabulary** — fixes vocabulary-mismatch at author time |
| Concept links | **Graph-walk expansion** — keyword-only stand-in for semantic spread |

## Changes

- **BM25F field weighting.** `notes` FTS5 table goes from a single `content` column to fielded columns (`title`, `tags`, `aliases`, `headings`, `body`), ranked with per-column weights (`NOTE_FIELD_WEIGHTS`). A term in a title/tag now outranks the same term in paragraph nine.
- **Controlled vocabulary.** Tags and frontmatter `aliases` are indexed and searchable, so a query phrased unlike the prose still matches.
- **Link-graph expansion** (`MemoryIndex.searchExpanded`). After the lexical search, walk markdown links — **outbound and inbound** — one hop out and fold in connected concepts a bare term query would miss. Flagged `expanded: true` and appended after real hits (never displaces a lexical match). Gated to the no-embeddings path.
- **Self-healing schema.** The index is fully derived from disk, so a new `notes_schema_version` in a `meta` table triggers a one-time drop+rebuild of the `notes`/`files`/`note_links` tables on open. **Turn and embedding tables are untouched** — Gemini vectors and the turn index survive.
- **New `src/lib/okf.ts`** — dependency-free, never-throws parser for frontmatter, headings, and links.
- **Config:** `retrieval.graph_expansion` (`enabled`/`hops`/`max_add`) via TOML, plus `PHANTOMBOT_RETRIEVAL_GRAPH_EXPANSION_ENABLED` / `_GRAPH_HOPS` / `_GRAPH_MAX_ADD`. Defaults on, 1 hop, 3 neighbours.

## Gemini users are unaffected

- The vector/hybrid path (`hybridSearch`) still runs exactly as before; graph expansion is **only** applied on the no-embeddings branch in `retrieval.ts` and `cli/memory.ts`.
- The one shared change is that the FTS leg is now fielded BM25F instead of flat BM25. That feeds the RRF merge on the hybrid path too — a strictly-better lexical ranking, not a behavior switch. Embedding/vector tables and the turn index are not migrated.

## Testing

- `tests/lib-okf.test.ts` — parser (frontmatter scalars/flow/block lists, headings, md + wiki links, BOM, junk-tolerance).
- `tests/lib-memoryIndex.test.ts` — BM25F (title/tag beats body), alias-only match, graph expansion (outbound + inbound), `maxAdd=0` off-switch, `resolveMdLink`, and legacy-v1-index self-heal.
- Full suite: **1706 pass / 0 fail**, `tsc --noEmit` clean.

- Whole-bundle injection for small bounded domains (perfect recall, zero retrieval).
- OKF bundle export/import for cross-phantom knowledge sharing.

### DIFF
--- a/src/lib/memoryIndex.ts
+++ b/src/lib/memoryIndex.ts
@@ -24,9 +24,24 @@ import { existsSync, readdirSync, statSync } from "node:fs";
 import { readFile } from "node:fs/promises";
 import { mkdir } from "node:fs/promises";
 import { dirname, join, relative } from "node:path";
+import { posix } from "node:path";
 import type { Turn } from "../memory/store.ts";
+import { parseOkf } from "./okf.ts";
 
 export type Scope = "memory" | "kb" | "turns";
+
+/**
+ * BM25F field weights for the `notes` table. A term hit in a concept's title,
+ * tags, or aliases is worth far more than the same term buried in the body —
+ * this is the lexical-precision half of the OKF superpowers, and it makes the
+ * frontmatter the index actually leans on. Positionally these map to the
+ * `notes` columns: [path, scope, title, tags, aliases, headings, body]. The
+ * two UNINDEXED columns get weight 0 (ignored by bm25 anyway).
+ */
+export const NOTE_FIELD_WEIGHTS = [0, 0, 8.0, 6.0, 6.0, 3.0, 1.0] as const;
+
+/** Column index of `body` in `notes` (for snippet()). */
+const NOTES_BODY_COL = 6;
 
 export interface IndexedFile {
   path: string; // relative to personaDir
@@ -44,6 +59,12 @@ export interface SearchHit {
   vecScore?: number;
   /** Reciprocal-rank-fusion score combining FTS + vec ranks. */
   rrfScore?: number;
+  /**
+   * True when this hit was not itself a lexical match but was pulled in by
+   * graph-walk expansion from a hit that was (OKF link-graph recall). Lets
+   * callers label or down-weight expanded neighbours.
+   */
+  expanded?: boolean;
   snippet: string;
 }
 
@@ -97,11 +118,34 @@ export function rrfMerge(
   return scores;
 }
 
+/**
+ * Bump when the on-disk schema of derived-from-disk tables changes (notes /
+ * files / note_links). On open, a mismatch drops and rebuilds JUST those
+ * tables — they're fully reconstructable by walking memory/ + kb/, so this is
+ * a safe, automatic, no-migration self-heal. The turn_docs / *_embeddings
+ * tables are NOT versioned here: they carry Gemini vectors and turn-index
+ * state that aren't cheaply rebuildable, so they're left untouched.
+ *
+ * v1 → v2: `notes` went from a single `content` column to OKF field columns
+ * (title / tags / aliases / headings / body) for BM25F, and `note_links` was
+ * added for graph-walk expansion.
+ */
+export const NOTES_SCHEMA_VERSION = 2;
+
 const SCHEMA = `
+CREATE TABLE IF NOT EXISTS meta (
+  key   TEXT PRIMARY KEY,
+  value TEXT NOT NULL
+);
+
 CREATE VIRTUAL TABLE IF NOT EXISTS notes USING fts5(
   path UNINDEXED,
   scope UNINDEXED,
-  content,
+  title,
+  tags,
+  aliases,
+  headings,
+  body,
   tokenize = 'porter unicode61'
 );
 
@@ -110,9 +154,20 @@ CREATE TABLE IF NOT EXISTS files (
   scope       TEXT NOT NULL,
   mtime_ms    INTEGER NOT NULL,
   size        INTEGER NOT NULL,
+  title       TEXT NOT NULL DEFAULT '',
+  aliases     TEXT NOT NULL DEFAULT '',
   indexed_at  TEXT NOT NULL
 );
 CREATE INDEX IF NOT EXISTS idx_files_scope ON files(scope);
+
+CREATE TABLE IF NOT EXISTS note_links (
+  src_path     TEXT NOT NULL,
+  target_raw   TEXT NOT NULL,
+  kind         TEXT NOT NULL,
+  target_path  TEXT
+);
+CREATE INDEX IF NOT EXISTS idx_note_links_src ON note_links(src_path);
+CREATE INDEX IF NOT EXISTS idx_note_links_target ON note_links(target_path);
 
 CREATE TABLE IF NOT EXISTS note_embeddings (
   path         TEXT NOT NULL,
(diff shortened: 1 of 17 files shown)


robertclawson (dismissed):
**APPROVE** ✅ — read the whole thing (parser, the FTS5/SQL changes, the self-heal path, graph-walk, retrieval wiring, and config defaults). This is a careful, well-scoped PR and I'm happy to ship it. Notes below; none are blockers.

## Why it's safe
- **`okf.ts` genuinely never throws.** Frontmatter, headings, and links all degrade to empty lists on junk; the regex YAML-subset deliberately drops shapes it can't parse rather than guessing. Good "search-index helper, not validator" discipline.
- **No SQL-injection surface.** All inserts/queries are parameterized. The one piece of string-built SQL — `bm25(notes, ${NOTE_FIELD_WEIGHTS.join(",")})` and the `snippet()` column index — is sourced from a hardcoded `const` array, not user input. `bm25()` doesn't accept bound weights, so building the literal is the only option and it's done correctly.
- **The self-heal is the right design.** `notes`/`files`/`note_links` are 100% derived from disk, so drop+rebuild on a `notes_schema_version` bump is safe and migration-free. Crucially the **turn_docs and *_embeddings tables are left untouched**, so Gemini vectors and the turn index survive the v1→v2 upgrade. The empty-DB stamp-and-return guard avoids a needless rebuild on first run.
- **Gemini path is untouched.** `hybridSearch` still runs as before; graph expansion is gated strictly to the no-embeddings branch in both `retrieval.ts` and `cli/memory.ts`. The only shared change is flat BM25 → fielded BM25F feeding the RRF merge, which is a strictly-better lexical ranking, not a behaviour switch.
- BFS expansion is bounded (hops capped 1–3, maxAdd 0–20, default 1/3), appended after real hits and flagged `expanded: true` so it never displaces a lexical match.

## CI: the CodeQL "failure" is a false positive — please dismiss it
The one red check is alert #10 (`js/incomplete-url-substring-sanitization`, high) on **`tests/lib-okf.test.ts:68`**. That line is a *test assertion* — `links.some(l => l.target.includes("example.com"))` checking that external URLs were correctly skipped. It is not a sanitization control. The actual control in `okf.ts` is the anchored `^(https?:|mailto:|tel:|ftp:|#)` test, which is correct. Safe to dismiss as "used in tests" / false positive. (Optional: rewrite the assertion to `expect(links.map(l => l.target)).not.toContain("https://example.com")` to keep the scanner quiet.)

## Minors (non-blocking, optional follow-ups)
1. **`resolveWikiTarget` is a full `files` table scan per unresolved wikilink**, inside the BFS. Bounded fine for now by small limits (seeds ≤ `limit`, maxAdd ≤ 3), but on a large kb it's O(unresolved-links × files). If kb grows, consider a cached basename/title/alias lookup map.
2. **Two slightly different gating idioms** for the same intent: `retrieval.ts` uses `ge?.enabled` while `cli/memory.ts` uses `ge?.enabled !== false`. Both resolve identically because `config.retrieval.graphExpansion` is always materialized from `DEFAULT_GRAPH_EXPANSION` — so this is cosmetic, not a bug. Worth aligning for the next reader.

Tests: 1706 pass / 0 fail, `tsc --noEmit` clean. Approving.

lenaparkhodges (dismissed):
## Review: PR #229 — OKF field-weighted BM25 + link-graph expansion

**Verdict: Approved ✅**

Two commits, 1016 insertions. CI green (typecheck + tests). 1706 pass / 0 fail locally, tsc clean.

### What it does

Upgrades the no-Gemini memory retrieval path from flat single-column BM25 to:
1. **BM25F** — FTS5 `notes` table goes from one `content` column to fielded columns (title, tags, aliases, headings, body) with per-column weights. A title hit (weight 8.0) outranks a body hit (weight 1.0).
2. **Controlled vocabulary** — tags and aliases are indexed, so queries phrased differently from the prose still match.
3. **Link-graph expansion** — after lexical search, walks markdown links one hop (outbound + inbound) and folds in connected concepts. Bounded by `maxAdd` (default 3), never displaces lexical hits.

Gemini users are unaffected — the hybrid path still runs `hybridSearch` exactly as before. The only shared change is that the FTS leg uses fielded BM25F, which is a strictly-better lexical ranking.

### Key implementation details verified

- **Self-healing schema migration** is clean: `meta` table tracks `notes_schema_version`; on mismatch, derived tables (notes/files/note_links) are dropped and rebuilt from disk. Turn and embedding tables are untouched. Empty-DB fast path avoids unnecessary drops.
- **`resolveMdLink`** correctly rejects external URLs, absolute paths, and tree escapes (`../../etc/passwd` → null). The `..` check prevents path traversal.
- **`snippetForPath`** updated to query `body` column instead of old `content` — correct for the new schema.
- **Graph expansion** is properly bounded: BFS with hop limit (1-3, clamped), maxAdd (0-20, clamped), `maxAdd=0` cleanly disables. Expanded hits are flagged `expanded: true` and appended after lexical hits.
- **`resolveWikiTarget`** does a full table scan per unresolved wikilink — O(n) per wikilink, but bounded by `maxAdd` and only called during graph expansion. Fine for typical KB sizes; would need an index if KBs grow very large.

### Non-blocking notes

1. **Inconsistent enabled check** — `retrieval.ts` uses `ge?.enabled` (truthy) while `cli/memory.ts` uses `ge?.enabled !== false` (explicit false check). Both produce the same result since `buildGraphExpansionConfig` always returns a boolean, but the inconsistency is a code smell. Pick one.

2. **CodeQL alert (false positive)** — "Incomplete URL substring sanitization" on `extractLinks` in `okf.ts`. This isn't a security boundary; it just decides which links to follow for graph indexing. Even if an external link were misclassified, `pathExists()` filters it out. Safe to dismiss.

3. **Body includes title + description** — the `body` FTS column folds in `[title, description, body]` for snippet quality. This means title terms get double-weighted (once in the `title` field at 8.0, once in `body` at 1.0). Intentional and documented; just noting the BM25 term-frequency implication.

### Tests

New tests cover the OKF parser (frontmatter scalars/flow/block lists, headings, links, BOM, junk tolerance), BM25F field weighting, alias-only matching, graph expansion (outbound + inbound), `maxAdd=0` off-switch, `resolveMdLink` edge cases, and the legacy v1 self-heal. Comprehensive.

Ready for merge.

src/lib/memoryIndex.ts
```diff
@@ -222,18 +320,54 @@ export class MemoryIndex {
...
+          .prepare(
+            "INSERT INTO note_links (src_path, target_raw, kind, target_path) " +
+              "VALUES (?, ?, ?, ?)",
+          )
+          .run(
+            f.path,
+            link.target,
+            link.kind,
+            link.kind === "md"
+              ? resolveMdLink(f.path, link.target)
```
kaieriksen, line 369: Blocking: wiki links are stored with `target_path = null`, and the inbound side of `linkedPaths()` only looks at `target_path`. That means a note reached by `[[Credential Store]]` or `[[credential cycling]]` can expand outward when the linking note is already a seed, but searching the target note cannot pull in notes that wikilink to it. It also misses multi-word aliases because `resolveWikiTarget()` stores aliases as a space-joined string and checks `split(/\\s+/).includes(t)`, so `aliases: [credential cycling]` never matches `[[credential cycling]]`. This breaks the OKF/Obsidian link-graph behavior this PR advertises; please resolve wikilinks to concrete target paths at index time or make inbound lookup resolve wiki targets, and add regression tests for inbound wikilinks plus multi-word alias wikilinks.

kaieriksen (changes requested):
Thanks — the fielded BM25/index self-heal pieces are in good shape, and the suite is clean. I found one blocker in the link-graph expansion path: wikilinks are only resolved outbound from the source note, so common OKF/Obsidian links do not reliably participate in the promised graph walk. Please fix that before merge.

andrewagrahamhodges:
@kaieriksen — fixed in f3bf0e9, thanks for the sharp catch. You were right that the link graph wasn't actually symmetric for wikilinks.

**Root cause, as you diagnosed:** wiki links were stored with `target_path = null` (only markdown links got resolved), and both `linkedPaths()` inbound lookup *and* the perf-flagged outbound fallback keyed on that. So a note could only be reached via wikilink when the *linking* note happened to be a seed — searching the *target* never pulled in its inbound wikilinks. Plus multi-word aliases were dead on arrival: `files.aliases` was space-joined and then split on `\s+`, so `aliases: [credential cycling]` could never match `[[credential cycling]]`.

**Fix — resolve at index time (your first suggested option):**
- New `resolveWikiLinks()` post-pass runs after `refreshStale()` indexes the changed notes. It builds a single lowercased **basename → title → alias** name index once (not an O(files) scan per link — that also kills the per-query scan Robbie/Lena flagged) and fills in `target_path` for every wiki link. Because it runs as a post-pass over the *whole* wiki-link set, **forward references** (a note that `[[links]]` to one indexed later) and **renames/deletes** self-repair on the next refresh. Skipped on no-op refreshes.
- `files.aliases` is now **newline-delimited**, so multi-word aliases survive intact.
- `linkedPaths()` now keys both directions purely on `target_path` — inbound wikilinks resolve symmetrically and the per-query `resolveWikiTarget` table scan is gone entirely.
- `NOTES_SCHEMA_VERSION` 2→3 so existing indexes self-heal into the new alias storage + populated wiki targets. Gemini/turn tables still untouched.

**Regression tests added** (all green):
- inbound wikilink expansion — query hits the *target*, expansion pulls the note that `[[wikilinks]]` to it
- multi-word alias wikilink — `[[credential cycling]]` resolves to the note declaring that alias
- forward-referenced wikilink — linking note indexed first, target added on a later refresh, link repairs

Also folded in the two non-blocking notes from @robertclawson / @lenaparkhodges: aligned the `ge?.enabled` gating idiom across `retrieval.ts` and `cli/memory.ts`, and rewrote the okf link-test assertion to silence the CodeQL false positive. Full suite: **1709 pass / 0 fail**, `tsc --noEmit` clean. Ready for another look when you have a minute.

kaieriksen (approved):
Re-reviewed latest head f3bf0e9. My blocker is fixed: wikilinks are now resolved to concrete target_path values in a post-pass, inbound lookup keys on target_path, multi-word aliases are preserved newline-delimited, and the regression tests cover inbound wikilinks, multi-word alias wikilinks, and forward references.\n\nVerified locally: focused memory/OKF/CLI/config tests pass (87 pass), typecheck passes, full bun test passes (1709 pass, 1 skip).

PR #230, opened 2026-06-28
feat(harnesses): enrich ACP tool-call titles with arguments (#218)

### LINKED ISSUE #218: Enrich ACP tool-call titles with arguments (Zed panel) — centralize across harnesses
## Summary

In the Zed (ACP) panel, every tool call renders as a bare label — `tool: Bash`, `tool: Read`, or just `tool` — so you can't tell *what* is actually running. We want to enrich these so the panel shows the meaningful argument, e.g. `Bash: git status`, `Read: src/foo.ts`, `Agent: Explore — "find the note builder"`.

This is **our side**, not Zed's: the panel renders whatever `title` string we hand it via the ACP `tool_call` update. The fix is to build a richer note from the tool input before it leaves the harness.

## Why it touches all harnesses (not one)

There is **no shared helper** — each adapter builds the note inline, and each pulls tool input from a different event shape:

| Harness | Location | Current note |
|---|---|---|
| claude | `src/harnesses/claude.ts:374` | `` `tool: ${toolName}` `` |
| pi | `src/harnesses/pi.ts:317` (also `:354`) | `` `tool: ${toolName}` `` / `"tool"` |
| gemini | `src/harnesses/gemini.ts:269` | `` `tool: ${name}` `` |
| codex | `src/harnesses/codex.ts:154, :168` | `"tool"` (doesn't even capture the name) |

The note then flows: harness `progress(note)` → `src/connectors/acp/server.ts:289-290` → `toolCallUpdate(...)` → ACP `tool_call` update → Zed.

## Proposed approach — centralize the formatter

Rather than write the same Bash/Read/Agent mapping four times, add one shared formatter and wire each harness to it. Codex gets fixed for free (it currently drops the tool name entirely).

### Part 1 — richer notes (high value)
1. Add `buildToolNote(toolName: string, input: unknown): string` (e.g. `src/harnesses/toolNote.ts`) with per-tool formatting:
   - `Bash` → the command (truncated, single line)
   - `Read` / `Edit` / `Write` → the file path
   - `Agent` → subagent type + short prompt snippet
   - fallback → just the tool name
2. Wire each harness to pass its tool-input block through `buildToolNote` instead of emitting `tool: ${name}`. Each harness extracts input from its own event shape — not copy-paste, but the formatting logic is shared.
3. Make sure codex captures and passes the tool name + input (currently hardcoded `"tool"`).

### Part 2 — clickable locations + icons (polish, optional/follow-up)
The ACP `ToolCallUpdate` type (`src/connectors/acp/protocol.ts:136-142`) only carries `title`, `status`, optional `kind`. To get clickable file links and proper icons in Zed we'd extend it with `locations` and `content`, set `kind` per tool, and thread those through the `toolCallUpdate` builder (`:190`) and the `server.ts` `progress` callback (which currently passes only a string).

## Acceptance criteria
- [ ] Bash calls show the command; Read/Edit/Write show the path; Agent shows the subagent + prompt snippet — across all four harnesses.
- [ ] Codex no longer emits a bare `tool`.
- [ ] Long inputs are truncated to a sane single-line length.
- [ ] One shared formatter; no per-harness duplication of the mapping.
- [ ] (Stretch) `locations`/`kind` plumbed so Zed renders clickable paths + icons.

## Notes / risks
- Don't leak anything sensitive into titles (e.g. tokens in a Bash command) — consider a length cap and basic redaction.
- Each harness's input shape needs verifying against a live call; the extraction differs per adapter.

andrewagrahamhodges:
Closes #218 (Part 1).

## What & why
Inside Zed (ACP panel) every tool call rendered as a dead label — `tool: Bash`, `tool: Read`, or for codex literally just `tool`. You could watch the agent work but not see *what* it was doing. This enriches the `tool_call` title so the panel shows the actual operation: `Bash: git status`, `Read: src/foo.ts`, `Agent: Explore — find the retrieval code`.

It's our side, not Zed's — Zed faithfully renders whatever `title` string our ACP bridge sends. The fix is to build a richer note before it leaves the harness.

## How
There was **no shared helper** — each of the four harness adapters (claude, codex, gemini, pi) built the note inline, in a different event shape. This centralises the formatting:

- **New `src/harnesses/toolNote.ts`** — a pure, never-throws `buildToolNote(name, input)`. Detail extraction keys on well-known **input field names** (`command`/`cmd`, `file_path`/`path`/…, `pattern`/`query`, `subagent_type`+`prompt`, `url`) rather than per-harness tool names, so one rule set works across every adapter despite their differing shapes. Single-line (whitespace collapsed), capped at **80 chars** with an ellipsis.
- **Wired all four harnesses** to it. **Codex** now captures the tool name it previously dropped entirely — bare `tool` → `tool: shell`, or `shell: git status` when the command is present.

## Scope decisions (per #218, confirmed with Andrew)
- **Part 1 only.** Protocol-level `locations`/`kind`/`content` (clickable file paths, per-tool icons) change the `progress` callback signature and the ACP type — different blast radius, **deferred to a Part 2 follow-up**.
- **Minimal redaction.** Truncate + whitespace-collapse only; **no secret masking**. (Noted in the module doc so a future, noisier sink than the Zed panel gets a second look.)

## Backward compatibility
When no useful detail can be extracted, we emit the **exact legacy `tool: <name>`** (or bare `tool`) label — so `/status` wording and all existing harness tests are unchanged. The richer `<Name>: <detail>` shape is purely additive, appearing only when args are present.

## Tests
- New `tests/harnesses-toolNote.test.ts` — per-tool rules, array commands (codex argv), whitespace collapse, 80-char truncation, hostile-input safety, and the legacy-label fallbacks.
- Per-harness wiring tests (claude/codex/gemini/pi) proving the detail flows end-to-end.
- Full suite green, `tsc --noEmit` clean.

### DIFF
--- a/src/harnesses/toolNote.ts
+++ b/src/harnesses/toolNote.ts
@@ -1,0 +1,131 @@
+/**
+ * Shared tool-call note formatter (issue #218).
+ *
+ * Every harness adapter (claude, codex, gemini, pi) emits a `progress` chunk
+ * when the model invokes a tool. That note becomes the `title` of the ACP
+ * `tool_call` update Zed renders in its agent panel. Historically each adapter
+ * built a bare `"tool: <name>"` string inline (codex didn't even capture the
+ * name), so the panel showed undifferentiated labels — you couldn't tell
+ * `Bash: git status` from `Bash: rm -rf /`.
+ *
+ * This module centralises the formatting so the per-tool rules live in ONE
+ * place and every harness benefits. Adapters are responsible only for the
+ * (per-harness, differently-shaped) extraction of the tool name + raw input;
+ * they hand both here and get back a single-line, length-capped title.
+ *
+ * Design notes:
+ *   - Backward compatible: when no useful detail can be extracted we return the
+ *     exact legacy `"tool: <name>"` (or bare `"tool"`) string, so existing
+ *     /status wording and tests are unchanged. The richer `"<Name>: <detail>"`
+ *     shape is purely additive — it only appears when args are present.
+ *   - Extraction keys on well-known INPUT FIELD NAMES (command, file_path,
+ *     query…) rather than per-harness tool names, so one rule set works across
+ *     all adapters even though their event shapes differ.
+ *   - Minimal redaction by design (Andrew's call on #218): we truncate and
+ *     collapse whitespace; we do NOT attempt secret masking. Keep this in mind
+ *     before surfacing inputs to any noisier sink than the Zed panel.
+ *   - Never throws. Bad/partial input degrades to the legacy label.
+ */
+
+/** Max length of a rendered tool-call title, including the name prefix. */
+export const MAX_TOOL_NOTE_LEN = 80;
+
+function isRecord(v: unknown): v is Record<string, unknown> {
+  return typeof v === "object" && v !== null && !Array.isArray(v);
+}
+
+/** Collapse all runs of whitespace (incl. newlines) to single spaces + trim. */
+function collapse(s: string): string {
+  return s.replace(/\s+/g, " ").trim();
+}
+
+/**
+ * Pull the first usable string from `input` across a list of candidate keys.
+ * Handles plain strings and string arrays (codex shell `command` is an array
+ * of argv tokens). Returns undefined if nothing usable is present.
+ */
+function firstField(
+  input: Record<string, unknown>,
+  keys: readonly string[],
+): string | undefined {
+  for (const key of keys) {
+    const v = input[key];
+    if (typeof v === "string") {
+      const c = collapse(v);
+      if (c) return c;
+    } else if (Array.isArray(v)) {
+      const joined = collapse(
+        v.filter((x): x is string => typeof x === "string").join(" "),
+      );
+      if (joined) return joined;
+    }
+  }
+  return undefined;
+}
+
+/**
+ * Extract a human-meaningful detail string from a tool's raw input, keyed on
+ * common field names so it works regardless of which harness produced it.
+ */
+function extractToolDetail(input: unknown): string | undefined {
+  if (!isRecord(input)) return undefined;
+
+  // Shell/command tools (Bash, shell, run_shell_command, bash…)
+  const command = firstField(input, ["command", "cmd"]);
+  if (command) return command;
+
+  // File tools (Read/Edit/Write, read_file/write_file/replace…)
+  const path = firstField(input, [
+    "file_path",
+    "filePath",
+    "path",
+    "absolute_path",
+  ]);
+  if (path) return path;
+
+  // Search tools (Grep/Glob, search…)
+  const pattern = firstField(input, ["pattern", "query"]);
+  if (pattern) return pattern;
+
+  // Sub-agent / task delegation (Agent/Task): subagent + a prompt snippet.
+  const subagent = firstField(input, ["subagent_type", "description"]);
+  const prompt = firstField(input, ["prompt"]);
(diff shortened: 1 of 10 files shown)


andrewagrahamhodges:
Part 2 follow-up (clickable paths + icons via ACP `ToolCallUpdate` extension) is now tracked in #231. This PR remains Part 1 only.

lenaparkhodges (approved):
Approved ✅

Clean, well-scoped PR. `buildToolNote` is pure, never-throws, and the extraction logic keyed on input field names is the right call — one rule set across all four adapters despite their different event shapes.

Backward compat is solid: legacy `tool: <name>` (or bare `tool`) preserved when no detail is extractable. Codex now correctly captures the tool name it previously dropped. Array-command joining (codex argv) and whitespace collapse are handled well.

Tests are thorough — hostile-input safety, 80-char truncation, backward-compat fallbacks, and per-harness wiring tests all present.

One non-blocking note: secret masking is deliberately deferred (documented). Fine for the local Zed panel sink, but worth revisiting before this formatter is reused for any remote/noisier surface.

robertclawson (approved):
**APPROVED** — clean, low-risk, well-tested. 🐶

Reviewed end-to-end: the `buildToolNote` core, all four harness wirings, and the test coverage.

**Why it's safe:**
- `buildToolNote` is genuinely pure and never-throws — `isRecord`/`firstField` guard every access, non-record or partial input degrades to the exact legacy label. No new failure path in the harness hot loop.
- **Backward compatible by construction:** no extractable detail → emits the byte-for-byte legacy `tool: <name>` / bare `tool`, so `/status` wording and all existing harness tests are unchanged. The richer `<Name>: <detail>` shape is purely additive.
- **Keying on input field names** (`command`/`file_path`/`pattern`/`prompt`/`url`) rather than per-harness tool names is the right call — one rule set survives the four adapters' differing event shapes, and an unknown shape just falls back to legacy.
- Codex wiring correctly recovers the tool name it used to drop (`tool` → `tool: shell` / `shell: <cmd>`); the other three pass input through cleanly. Worst case on any unexpected nesting is a graceful fallback, not a crash.
- CI fully green (typecheck, test, CodeQL x2 + actions).

**One thing to keep on the record (not a blocker):** the deliberate *no secret-masking* decision means a command like `curl -H "Authorization: Bearer …"` will render verbatim in the Zed panel title. That's fine for the current sink — it's the local user's own screen, same trust level as Telegram echoing your own input — and it's Andrew's explicit call on #218, flagged in the module doc. The truncate+collapse keeps it tidy. The note to re-evaluate before piping these titles to any noisier sink is exactly right; worth honoring if Part 2 ever logs them.

Part 1 scope (titles only, protocol-level `locations`/`kind`/`content` deferred to Part 2) is the sensible split. No blockers from me.

src/harnesses/toolNote.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+  return undefined;
+}
+
+/**
+ * Extract a human-meaningful detail string from a tool's raw input, keyed on
+ * common field names so it works regardless of which harness produced it.
+ */
+function extractToolDetail(input: unknown): string | undefined {
```
kaieriksen, line 70: Keying detail extraction on argument field names rather than harness/tool names is the right design here; it keeps the formatter shared while preserving graceful fallback for unknown shapes.

kaieriksen (approved):
Approved. The shared formatter is scoped well, preserves legacy fallbacks, and the harness wiring is covered. I verified the focused harness tests, typecheck, and full suite locally.

PR #232, opened 2026-06-28
feat(acp): register Phantombot as an ACP agent for JetBrains IDEs

14skywalker:
## What

Registers Phantombot as an [ACP](https://www.jetbrains.com/help/ai-assistant/acp.html) agent for **JetBrains IDEs** (Rider, IntelliJ, PyCharm, etc.). Rider 2026.1's AI Assistant speaks ACP natively, so **no plugin is required** — Phantombot just registers itself in `~/.jetbrains/acp.json`, the same way it does for Zed.

## How

Mirrors the existing Zed installer:

- **`installJetbrains.ts`** — writes/merges `agent_servers.Phantombot` into `~/.jetbrains/acp.json`. Same data-loss guard as Zed (parse-or-abort, backup, atomic write), and reuses the shared agent-server block so the `PHANTOMBOT_CONFIG` pinning stays single-sourced.
- **`autoInstall.ts`** — adds `JETBRAINS_EDITOR` to `KNOWN_EDITORS`, which auto-wires `doctor` and startup registration. Generalizes the shared command-reader.
- **`acp.ts`** — new `phantombot acp install jetbrains` subcommand + updated help text.
- **Tests** — new suite mirroring the Zed tests.

## Test

- 40/40 ACP tests pass
- `tsc --noEmit` clean
- Real `acp install jetbrains` run wrote a correct `~/.jetbrains/acp.json`

## Notes

There's a known pre-existing limitation (shared with the Zed installer): when run from source under `bun`/`node` rather than a compiled binary, the emitted `command`/`args` drop the entrypoint script. That's a separate fix — not in scope here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,7 @@ Interactive setup:
 | `phantombot voice` | Configure TTS/STT providers |
 | `phantombot embedding` | Configure semantic memory |
 | `phantombot acp install zed` | Register phantombot as an ACP agent in Zed |
+| `phantombot acp install jetbrains` | Register phantombot as an ACP agent in JetBrains IDEs (Rider, IntelliJ, …) |
 | `phantombot acp install vscode` | Install the first-party VS Code extension |
 
 Runtime:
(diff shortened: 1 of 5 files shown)


robertclawson (dismissed):
Reviewed — this is clean and merge-ready. ✅

**Looks good:**
- Faithfully mirrors the established `installZed` pattern: reuses the shared `phantombotAgentServerBlock` so `PHANTOMBOT_CONFIG` pinning stays single-sourced, same parse-or-abort data-loss guard, backup-then-atomic-write (temp + rename).
- `readZedCommand` → `readAgentServerCommand` generalization is the right call since Zed and JetBrains use the identical `agent_servers.Phantombot` shape; nicely commented.
- `defaultJetbrainsConfigPath` correctly hardcodes `~/.jetbrains/` (not XDG), with a test that explicitly asserts XDG is ignored.
- Test suite mirrors the Zed coverage: sibling-key + comment preservation, unparseable→abort (file byte-for-byte unchanged), no-file creation, backup creation. 40/40 ACP tests + typecheck green in CI.

**Non-blocking:**
- The known limitation called out in the PR description (entrypoint script dropped when run from source under bun/node rather than a compiled binary) is shared with the Zed installer and explicitly out of scope — agreed, fine to handle separately.

No changes requested. Nice work — approving.

robertclawson (approved):
Re-approving — the new commit (903251e) is docs/comments only: README gains the JetBrains row + an accurate "VS Code, Zed & JetBrains" section, and the acp.ts header/usage comments are now editor-agnostic. No behaviour change since my prior approval. The substantive installer (installJetbrains.ts), autoInstall wiring, and the mirrored test suite all still look clean.

CI is fully green (test, typecheck, CodeQL, Analyze javascript-typescript + actions all pass). Merge-ready from my side. ✅

Note: the pre-existing source-vs-binary entrypoint caveat called out in the PR description remains out of scope here, as agreed.

PR #233, opened 2026-06-29
feat(phantomchat): only the addressed bot replies in multi-bot groups

andrewagrahamhodges:
## Problem

When several persona bots share a PhantomChat group, **every bot answers every message** — say something to one and all of them pile on (see the HQ screenshot in the issue). The phantomchat server ran a turn for any authorized sender with no notion of "who was addressed."

Telegram already solved this: each bot decides locally, from the shared human message stream, whether *it* was addressed (`matchPersonaNames` + `decideGroupReply` in `core/routing.ts`). That logic is deliberately channel-agnostic — the phantomchat server just never called it.

## Fix

Wire that same gate into `runPhantomchatServer`'s group path. A bot replies only when:
- its **persona name** appears in the message ("hey **Lena**, …"), or
- it currently **holds the thread** (so a no-name follow-up still reaches it).

When you address a different bot by name, the previous one falls quiet (hand-off) — which is why each bot needs the full name roster.

### Option (a): bots never react to each other

A bot **ignores any message whose cryptographic sender is another bot** — no reply, no state change. This kills bot-to-bot cascades dead: one bot's reply (which might mention another bot's name) can never trigger a chain. Only humans drive addressing. This was the explicit design choice over enabling bot-to-bot collaboration.

### Config: `group_bots`

Siblings are declared per-persona in `phantomchat.json`:

```json
"group_bots": [
  { "name": "kai",    "npub": "npub1kai…" },
  { "name": "robbie", "npub": "npub1robbie…" }
]
```

- `name` → feeds the shared addressing roster (name-disambiguation + hand-off).
- `npub` → decoded to hex, feeds the "ignore other bots" set.

Configure it for **every** bot in the group, each listing the others.

### No regression for single-bot groups

The gate **engages only when sibling bots are configured**. A lone/unconfigured bot keeps the original behaviour (answers every group message), so this is opt-in and can't make an existing deployment go mute. Backward compatibility is proven by the pre-existing HQ tests still passing unchanged.

## Changes

- `personaStore.ts` — parse + round-trip `group_bots` (name + npub + decoded hex) through every save helper.
- `server.ts` — group addressing gate + catch-up context preamble, run **before** the paid voice/media pipeline so an unaddressed message costs no STT.
- `run.ts` — derive the name roster + sibling-bot hex set from `group_bots`.
- README — document `group_bots`.

## Tests

6 new gate tests (in `channels-phantomchat-server.test.ts`):
- addressed by name → replies
- another bot named → stays quiet
- option (a): sibling bot's message ignored even when it names this bot
- sticky no-name follow-up → still reaches the thread-holder
- hand-off → bot falls quiet once Andrew addresses the sibling
- no siblings configured → lone bot still answers everything

**Guard-proved**: neutering the gate fails the 3 silence-asserting tests. Full suite: **1737 pass / 0 fail**, `tsc` clean.

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-../../phantombot/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+../../phantombot/node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # Node
+node_modules
 node_modules/
 dist/
 *.tsbuildinfo
--- a/README.md
+++ b/README.md
@@ -398,9 +398,28 @@ for the app itself and the wire-protocol details.
 ### Multiple bots in one group
 
 When several persona bots share a PhantomChat group, each one would otherwise
-answer **every** message — three bots, three replies to one question. To get the
-Telegram behaviour (only the bot you addressed responds), tell each persona who
-its **sibling bots** are by adding a `group_bots` list to its `phantomchat.json`:
+answer **every** message — three bots, three replies to one question. PhantomChat
+gives you the Telegram behaviour (only the bot you addressed responds)
+**automatically, with no configuration**:
+
+- A bot replies only when its **persona name** appears in the message ("hey
+  **Lena**, …"), or when it's the bot currently holding the thread (so a no-name
+  follow-up still reaches it). Address a different bot by name and the previous
+  one falls quiet.
+- A bot **never reacts to another bot's** messages — in a group *or* a 1:1 DM —
+  so one bot's reply can't trigger another and start a back-and-forth loop. Only
+  humans drive the conversation.
+
+This works out of the box because every Phantom publishes a NIP-24 **`bot: true`**
+flag and a display name in its Nostr profile (kind-0). Each bot reads the
+profiles of the group's members, so it learns who the *other* bots are — and
+their names — straight from the protocol. Nothing to wire up; just add the bots
+to a group.
+
+**Optional override.** If you want deterministic behaviour from the very first
+message (before profiles resolve), or to force-mark a specific account as a
+sibling bot, you can still seed the roster per-persona with a `group_bots` list
+in `phantomchat.json`. It's merged with what's auto-detected:
 
 ```json
 {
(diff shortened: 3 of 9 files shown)


andrewagrahamhodges:
## Update: reworked to **zero-config** auto-detection

Per discussion, this PR no longer relies on a hand-maintained `group_bots` list. Each bot now learns who the other bots are **straight from the Nostr protocol** — every Phantom already publishes a NIP-24 `bot: true` flag + display name in its kind-0 profile, so a bot resolves its group members' profiles and derives both the "ignore other bots" set and the name-addressing roster automatically.

**What changed since the first commit (`9ddd806` → `5c0fd18`):**

- **`transport.fetchProfiles(authors)`** — a one-shot kind-0 REQ + EOSE/timeout pull (mirrors the existing `fetchGiftWrapsSince`), newest event wins per author. `NostrFilter` widened with `authors`; `NostrProfileMeta` exported.
- **Server bot-gate is now GLOBAL** — a bot never replies to another bot in a **DM** as well as a group, because the check is keyed on the sender's profile `bot` flag, not a group-only config set. Cascades killed everywhere.
- **Auto-derived roster** (`buildRoster`) + lazy, TTL-bounded, de-duped **profile cache** (`ensureProfiles`). The gate engages only when another bot is actually present, so a lone bot still answers everything (no false mute). Cache is warmed for allowed peers at startup.
- **`group_bots` is now an OPTIONAL override/seed** for deterministic cold-start, not a requirement. README leads with the zero-config behaviour.

**Cold-start note (honest):** the only residual window is a brand-new bot's *first* message before its profile has propagated to relays — until then a sibling might not yet be recognised. It self-heals on the next message (profiles are cached + refreshed), and the optional `group_bots` seed closes it entirely for anyone who wants determinism from boot.

**Tests:** +6 auto-detection gate tests (sibling resolved from kind-0 → quiet; this bot named → replies; cascade kill with no config; bot ignored in a DM; human DM still answered; cold cache → no false mute) and +3 transport `fetchProfiles` tests. Guard-proved (neutering profile detection fails the 3 detection tests). Full suite **1746 pass / 0 fail**, `tsc` clean. Also dropped an accidentally-tracked `node_modules` gitlink from the first commit.

node_modules
```diff
@@ -0,0 +1 @@
+../../phantombot/node_modules
```
kaieriksen: This should not be tracked. Because this symlink is committed, `bun install --frozen-lockfile` tries to open `node_modules` as an existing path and fails in CI with `ENOENT: could not open the "node_modules" directory`; in this review clone it points to `/home/phantombot/node_modules`, which is broken. Please remove the symlink from the commit and leave `node_modules/` ignored.

kaieriksen (changes requested):
Blocking on the tracked node_modules symlink. It breaks CI before tests/typecheck can run, and it also leaves a fresh checkout without usable dependencies. Remove node_modules from the PR and keep it ignored.

andrewagrahamhodges:
@kaieriksen good catch — fixed in `008f3ea`.

You were right it was still there. The root cause was subtle: it wasn't a gitlink, it was a **symlink** (`node_modules -> ../../phantombot/node_modules`, mode `120000`) committed in the first commit. It slipped past `.gitignore` because the existing `node_modules/` pattern has a trailing slash, which matches **directories only** — a symlink isn't a directory to git, so it stayed tracked. My earlier "cleanup" removed a different artifact and missed this.

Fix:
- `git rm --cached node_modules` — untracked (kept on disk locally so the worktree keeps its deps).
- Added a trailing-slash-free `node_modules` line to `.gitignore` so the symlink form can't be re-added.

Verified: `git ls-files | grep node_modules` → nothing, `git check-ignore node_modules` → matched, `tsc` clean. Fresh checkouts now ship without the broken symlink and CI installs real deps. Re-requesting review.

kaieriksen (approved):
Re-reviewed after the node_modules symlink fix. CI is green, local typecheck and full test suite pass, and the prior blocker is resolved.

PR #235, opened 2026-06-30
fix(telegram): re-poll on non-JSON getUpdates body instead of crashing

salvatoretrimarchi:
## Summary

A single transient non-JSON HTTP response from one Telegram bot could tear down the **entire** `phantombot run` process — every sibling listener included.

This guards the `getUpdates` JSON parse so a bad body re-polls cleanly instead of propagating.

## The problem

In `HttpTelegramTransport.getUpdates`, the response body was parsed unguarded:

```ts
const body = (await res.json()) as { ok?: boolean; result?: ...; description?: string };
```
res.json() throws when the body is a 200 that isn't JSON — a captive portal, a proxy/middlebox error page, or an empty/truncated response. The fetch call above it is wrapped in try/catch, but this parse was not, so the throw escaped getUpdates.

That throw is not contained downstream:

- the engine drives getUpdates inside a try { … } finally { … } (no catch), so the error unwinds the poll loop;
- it then rejects await Promise.all(tasks) in run.ts, whose handler all sibling listeners and re-raises.

Net effect: one bot receiving one malformed 200 takes the whole process down (non-zero exit). Under systemd that becomes a full restart cycle transient network condition.

The fix

Wrap the parse and handle failure exactly like the existing fetch-failure and non-OK branches — log a warning and return an empty batch with the preserved, so the loop simply re-polls the same updates on the next pass:

```ts
try {
  body = (await res.json()) as typeof body;
} catch (e) {
  log.warn("telegram: getUpdates body not JSON", { error: (e as Error).message });
  return { updates: [], nextOffset: offset };
}
```

No behavioral change on the happy path; this only converts a previo parse error into the same self-healing re-poll used for every other transport hiccup.

### DIFF
--- a/src/channels/telegram/transport.ts
+++ b/src/channels/telegram/transport.ts
@@ -132,11 +132,26 @@ export class HttpTelegramTransport implements TelegramTransport {
       log.warn("telegram: getUpdates non-OK", { status: res.status });
       return { updates: [], nextOffset: offset };
     }
-    const body = (await res.json()) as {
+    let body: {
       ok?: boolean;
       result?: TelegramRawUpdate[];
       description?: string;
     };
+    try {
+      body = (await res.json()) as typeof body;
+    } catch (e) {
+      // A 200 with a non-JSON body (captive portal, proxy error page, an
+      // empty/truncated response) makes res.json() throw. Swallow it like
+      // the fetch failure above — log and re-poll — instead of letting it
+      // escape. The engine drives getUpdates inside a try/FINALLY (no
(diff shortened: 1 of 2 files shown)


kaieriksen (approved):
Reviewed the Telegram getUpdates non-JSON body handling. The change preserves the current offset and retries cleanly instead of letting a transient 200/non-JSON response unwind the poll loop.\n\nVerified locally: targeted Telegram channel tests, typecheck, full test suite, green CI, and clean merge compatibility against current main.

Issue #55, opened 2026-05-02
channels-telegram.test.ts: clean up no-op 'voice-in but TTS unavailable' test


andrewagrahamhodges:
Non-blocking nits from PR #53 review.

The third test in the \`runTelegramServer voice brevity instruction\` describe block (\`tests/channels-telegram.test.ts\` ~line 599) was added to document why the voice-in/text-out branch isn't reachable in v1's provider matrix (no provider has STT-only). The test sets up mocks, then never calls \`runTelegramServer\` and falls through to \`expect(true).toBe(true)\` — pure documentation dressed as a test.

Two cleanups:

1. Convert to \`test.skip(...)\` (or \`test.todo(...)\`) so the intent reads honestly and CI doesn't burn cycles on a no-op assertion. Even better: move the explanation to a \`//\` comment above the \`willReplyWithVoice\` gate in \`src/channels/telegram.ts\` and delete the test entirely.
2. Drop the \`void saved;\` line and the surrounding \`saved = process.env.PHANTOMBOT_OPENAI_API_KEY;\` setup — the comment says \"flip it to no right before\" but the actual flip never happens. Dead code that just suppresses an unused-var warning.

Touch \`tests/channels-telegram.test.ts\` only. Suite count drops by 1 if the test is removed; expect-call count drops by 1.

kaieriksen:
Resolved — the test suite has been substantially refactored. The `void saved;` dead code and the `expect(true).toBe(true)` no-op test are gone. Closing.

PR #236, opened 2026-06-30
feat(typing): switch from ephemeral kind-20001 to stored kind-30001 (NIP-33)

lenaparkhodges:
## Summary

Bot-side of the typing indicator migration from **NIP-16 ephemeral (kind-20001)** to **NIP-33 parameterized replaceable (kind-30001)**. Relays now store the latest typing event per sender+conversation, so late reconnects replay instead of silently losing ticks.

Pairs with [phantomchat#55](https://github.com/phantomyard/phantomchat/pull/55).

## What changed

- **transport.ts**: `NOSTR_KIND_TYPING` → 30001. Added `NOSTR_KIND_TYPING_LEGACY = 20001`. `sendTyping`, `sendGroupTyping`, and `sendRecording` now emit kind-30001 with:
  - `['d', conversationId]` tag — serialization key (relay keeps only latest per sender+conversation)
  - `['expiration', now+30]` tag — NIP-40 auto-prune for crashed senders
- **channel.ts**: Updated comment to reference NIP-33 kind-30001.
- **server.ts**: Updated comment to reference kind-30001.
- **Tests**: Updated to assert kind-30001, d-tag presence, and expiration tag.

## Why this works

Ephemeral events (kind 20000–29999) are fire-and-forget: relays don't store them, there's zero replay, and any subscription gap means permanent silent loss.

Parameterized replaceable events (kind 30000–39999) are stored by relays and replayed on reconnect. The `d` tag ensures only the latest state is kept (no ghost typing from stale events), and the `expiration` tag provides automatic cleanup.

## Testing

All 52 tests pass (18 transport + 34 server). No new test files — existing tests were updated to assert the new kind and tag structure.


andrewagrahamhodges (commented):
Bot side of the kind-30001 migration — pairs with phantomchat#55, where I left the detailed review. Code's clean and the transport tests are solid (`arrayContaining` over exact tag-order is the right move; `now` hoisted so created_at and expiration share a timestamp — good). Same two substantive holds as the PWA side, so I'm **withholding approval** pending a decision:

1. **Kind 30001 collides with NIP-51** (assigned lists kind). Now that it's parameterized-replaceable keyed on `(pubkey, 30001, d)` — NIP-51's own keying — a user on any NIP-51 client risks clobbering. Inline on the constant.
2. **Persistence/privacy:** these ticks aren't gift-wrapped and now get *stored*, so the sender→recipient graph + timing becomes queryable history rather than a live-only leak. Cross-repo concern; full write-up on #55.
3. **Necessity:** the transport-layer fixes (socket recycle, firehose bound, remount, folder-index) already landed and typing worked live afterward — worth confirming the stored transport is still needed vs. those covering the real failure.

Minor: `NOSTR_KIND_TYPING_LEGACY` is now unused on the emit side — inline.

Keep the two repos in lockstep on whatever kind we settle on, since they must match on the wire.

src/channels/phantomchat/transport.ts
```diff
@@ -48,17 +48,27 @@ export const DEFAULT_PHANTOMCHAT_RELAYS: readonly string[] = [
...
+ * (range 30000–39999). Relays STORE the latest event per (pubkey, kind, d-tag),
+ * so late reconnects replay instead of silently losing typing ticks. The PWA
+ * subscribes for this kind p-tagged to itself and injects a native
+ * `updateUserTyping` (three-dots, 6s auto-expiry). Must match phantomchat's
+ * `NOSTR_KIND_TYPING`.
+ *
+ * Replaces the old NIP-16 ephemeral kind-20001 which relays did NOT store —
+ * any subscription gap caused permanent silent loss.
+ */
+export const NOSTR_KIND_TYPING = 30001;
```
andrewagrahamhodges: Same flag as on phantomchat#55: **30001 is an assigned NIP-51 lists kind**. As a parameterized-replaceable event keyed by `(pubkey, 30001, d-tag)` it shares NIP-51's exact serialization scheme, so a user running a NIP-51-aware client could have typing events and list entries clobber each other. Recommend an unassigned app-specific kind. Both repos must change together since the constant is matched on the wire (`Must match phantomchat's NOSTR_KIND_TYPING`).

src/channels/phantomchat/transport.ts
```diff
@@ -48,17 +48,27 @@ export const DEFAULT_PHANTOMCHAT_RELAYS: readonly string[] = [
...
+
+/**
+ * Legacy ephemeral typing kind (NIP-16, kind-20001). Retained as a reference;
+ * the emit side now sends kind-30001 exclusively. The PWA's receive side
+ * still accepts kind-20001 for backward compat during the migration.
  */
-export const NOSTR_KIND_TYPING = 20001;
+export const NOSTR_KIND_TYPING_LEGACY = 20001;
```
andrewagrahamhodges: `NOSTR_KIND_TYPING_LEGACY` is now dead on the bot — the emit side only ever sends `NOSTR_KIND_TYPING`, and the receive-side fallback lives in the PWA. Either drop it or add a one-line comment that it's kept only as documentation of the pre-migration kind, so a future reader doesn't go hunting for where it's referenced.

src/channels/phantomchat/transport.ts
```diff
@@ -580,11 +590,12 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
...
   async sendRecording(conversationId: string): Promise<void> {
     try {
+      const now = Math.floor(Date.now() / 1000);
       const event = finalizeEvent(
         {
           kind: NOSTR_KIND_TYPING,
-          created_at: Math.floor(Date.now() / 1000),
-          tags: [["p", conversationId]],
+          created_at: now,
+          tags: [["d", conversationId], ["p", conversationId], ["expiration", String(now + 30)]],
```
andrewagrahamhodges: Privacy note (mirrors phantomchat#55): this tick is signed by the real key and `p`-tags the real recipient, un-gift-wrapped. Promoting it to a stored kind turns the live-only social-graph leak into a persistent, relay-queryable record of who-typed-to-whom-when. The `expiration` tag only helps on NIP-40-compliant relays. Flagging so the metadata tradeoff is a conscious decision, not a side effect of the reliability fix.

src/channels/phantomchat/transport.ts
```diff
@@ -673,11 +685,12 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
...
   async sendTyping(conversationId: string, stop?: boolean): Promise<void> {
     try {
+      const now = Math.floor(Date.now() / 1000);
       const event = finalizeEvent(
         {
           kind: NOSTR_KIND_TYPING,
-          created_at: Math.floor(Date.now() / 1000),
-          tags: [["p", conversationId]],
+          created_at: now,
+          tags: [["d", conversationId], ["p", conversationId], ["expiration", String(now + 30)]],
```
kaieriksen: This matches the PWA contract: the d tag gives relays the replaceable key and the p tag keeps recipient filtering unchanged during delivery.

kaieriksen (dismissed):
Approved. Bot-side typing emission now matches the PhantomChat kind-30001 contract, with d-tag serialization and expiration coverage in tests.

src/channels/phantomchat/transport.ts
andrewagrahamhodges: **Recommendation** (lockstep with phantomchat#55) — we're going persistent per Andrew, so this ships, just not on 30001.

**Gift-wrap the tick** (NIP-59 / kind 1059) like normal messages. Resolves both concerns at once: the inner rumor kind only needs app-local uniqueness (no NIP-51 bookmark-list clobber), and it restores metadata privacy. Keep the d-tag dedup key in the rumor; dedup client-side by latest tick within the staleness window.

If staying unwrapped/replaceable: move off 30001 to a **verified-unassigned** addressable kind (30000–39999) behind a shared constant mirrored in both repos. **Do not ship on 30001 either way** — both repos must change together or they desync.

src/channels/phantomchat/transport.ts
andrewagrahamhodges: **Recommendation**: delete `NOSTR_KIND_TYPING_LEGACY` — emit only sends the new kind now, so it's dead. If you want a migration window to still *receive* the old kind, wire it explicitly into the receive filter with a comment and a removal date; otherwise drop it. Don't leave it exported-but-unused.

src/channels/phantomchat/transport.ts
andrewagrahamhodges: **Recommendation** (mirrors phantomchat#55): persisting an un-wrapped, signed, `p`-tagged tick makes the social-graph leak permanent and relay-queryable. `expiration` only helps on NIP-40 relays. Fix is the same: **gift-wrap it** (see the kind thread). That single change covers both privacy and the 30001 collision on the bot side too.

andrewagrahamhodges (changes requested):
Requesting changes — one blocker, paired with phantomchat#55. Persistent indicators are the right call; this is about the kind and wrapping only.

**Blocker — don't ship on kind 30001.** Assigned NIP-51 kind (Categorized Bookmarks), keyed `(pubkey, 30001, d-tag)`. Publishing typing ticks on it can clobber a user's bookmark list on any NIP-51 client. Must change before merge, in lockstep with the client.

**Recommended fix: gift-wrap the tick (NIP-59/1059), same as outbound messages.**
- Collision gone — inner kind invisible to relays, only unique within our namespace.
- Privacy restored — a stored, signed, `p`-tagged, un-wrapped tick is a permanent relay-queryable social-graph leak; wrapping puts it back behind NIP-59 metadata hiding.
- Cost: drop relay-side replaceable dedup → dedup client-side within the 30s window; reconnect-replay still covered by the backfill.

Fallback if not wrapping: verified-unassigned addressable kind behind a shared constant, both repos together.

**Minor:** `NOSTR_KIND_TYPING_LEGACY` is exported but unused — drop it, or wire the back-compat read path it implies.

src/channels/phantomchat/transport.ts
lenaparkhodges: ✅ Fixed in `582d41f`. Bot now emits typing ticks as gift-wrapped kind-1059 (inner kind-14 rumor). No kind collision with NIP-51 — the inner kind is encrypted and invisible to relays. The `d` tag for dedup stays inside the rumor.

src/channels/phantomchat/transport.ts
lenaparkhodges: ✅ Addressed. `NOSTR_KIND_TYPING` and `NOSTR_KIND_TYPING_LEGACY` are now retained only as reference constants (the emit side uses the gift-wrap functions directly). Added a comment clarifying they are no longer on the wire — kept for documentation of the pre-migration kinds.

src/channels/phantomchat/transport.ts
lenaparkhodges: ✅ Fixed. Bot now uses `wrapTypingGiftWrap` / `wrapGroupTypingGiftWrap` (from `nostrCrypto.ts`) — same gift-wrap envelope as outbound messages. Outer kind-1059 signed with ephemeral key, relay only sees the envelope. Social graph leak eliminated.

src/channels/phantomchat/transport.ts
andrewagrahamhodges: Resolved — gift-wrap moves the typing kind inside the encrypted kind-14 rumor; the wire only carries kind-1059, so the 30001/NIP-51 collision is gone. 👍

src/channels/phantomchat/transport.ts
andrewagrahamhodges: Resolved — `wrapTypingGiftWrap`/`wrapGroupTypingGiftWrap` give ephemeral outer keys + encrypted recipient/content/group. Real-key signing + plaintext p-tag leak closed.

src/channels/phantomchat/transport.ts
andrewagrahamhodges: Still standing — after the gift-wrap switch BOTH `NOSTR_KIND_TYPING` (30001) and `NOSTR_KIND_TYPING_LEGACY` (20001) are now unused on the bot (emit goes through the wrap helpers, bot doesn't receive). Drop both, or keep one with a comment if you want the reference. Trivial; folding into the changes request below.

andrewagrahamhodges (changes requested):
## Almost there — one red CI blocker 🔴

The gift-wrap rework lands the two big concerns (collision + privacy — both threads resolved above). But **typecheck is failing**, so this can't merge yet:

```
tests/channels-phantomchat-server.test.ts(1464,42): error TS6133: 'keys' is declared but its value is never read.
```

The refactored helper picked up an unused rest-param:

```ts
const replyWraps = (pool: FakePool, ...keys: Uint8Array[]) =>
  messageWraps(pool);   // keys never used
```

Just drop `...keys: Uint8Array[]` (and the `c.lenaSk, c.andrewSk` args at the four call sites), or use them. Tests pass; it's only `noUnusedParameters` blocking the build.

**Minor (same review):** both `NOSTR_KIND_TYPING` (30001) and `NOSTR_KIND_TYPING_LEGACY` (20001) are now dead on the bot — see the reply on that thread. Worth dropping while you're in there.

Bounce me when CI's green and I'll approve — phantomchat#55 is already approved and the wire format matches.

kaieriksen (dismissed):
Approved. Re-reviewed the bot-side gift-wrapped typing update at 8155259: the wire format now matches phantomchat#55, the previous TS6133 blocker is fixed by the unused rest-param rename, and typing wrappers/transport tests cover DM, group, stop, and recording paths.\n\nLocal checks passed: bun test tests/channels-phantomchat-transport.test.ts tests/channels-phantomchat-server.test.ts; bun run typecheck; bun test. GitHub test/typecheck/CodeQL are green.

src/channels/phantomchat/transport.ts
lenaparkhodges: ✅ Fixed in `4f199e3`. Both `NOSTR_KIND_TYPING` (30001) and `NOSTR_KIND_TYPING_LEGACY` (20001) are dropped — the bot emits via gift-wrap helpers and never receives, so neither constant was referenced. Also fixed the `replyWraps` unused `...keys` param from the remote fix in the same commit.

andrewagrahamhodges: Confirmed gone in 4f199e3 — `NOSTR_KIND_TYPING*` fully removed from nostrCrypto.ts. Resolving.

andrewagrahamhodges (dismissed):
Approving — both blockers from my changes-requested round are resolved:

- **kind 30001 collision** → fixed via NIP-17 gift-wrap (kind-1059, ephemeral outer key). Inner kind is invisible to relays, so no NIP-51 bookmarks clobber. Matches phantomchat#55.
- **typecheck red** → green. Unused `replyWraps`/`keys` param and the dead `NOSTR_KIND_TYPING*` constants are gone.

Verified: `wrapGroupTypingGiftWrap` wraps the kind-14 rumor per-member with a fresh ephemeral key; tests adapted cleanly (`messageWraps`/`isMessageWrap` filter typing from message wraps). CI green (typecheck + test + CodeQL).

Non-blocking residual (noted before, not gating): the outer `['t','typing']` marker lets a relay distinguish typing wraps from message wraps without decrypting — small privacy step back. It's used here for receive-side filtering, so it's a deliberate tradeoff; fine to keep. Could later filter on `p`-tag + decrypt instead if we want zero outer signal.

package-lock.json
```diff
@@ -0,0 +1,315 @@
+{
```
kaieriksen: This looks accidental. The root repo is Bun-managed and already tracks bun.lock; adding a second npm package-lock.json gives us two independent root lockfiles that can drift and mislead humans/tools about the supported installer. Please drop this file from the PR and keep bun.lock as the canonical lockfile.

kaieriksen (changes requested):
Requested changes. The code cleanup itself looks fine and the focused checks/typecheck pass, but the latest commit accidentally adds an npm package-lock.json to the Bun-locked root. That should be removed before merge so the repo keeps one canonical dependency lockfile.

package-lock.json
lenaparkhodges: ✅ Fixed in aa3dd7d. Removed — bun.lock is the sole lockfile.

kaieriksen (approved):
Approved. Re-checked aa3dd7d: the accidental package-lock.json is removed, bun.lock remains the sole root lockfile, focused PhantomChat tests pass, typecheck passes, and GitHub CI is green.

PR #237, opened 2026-06-30
refactor(typing): dedicated non-ephemeral kind for typing/voice + inbound kind guard

andrewagrahamhodges:
Pairs with **phantomchat #57**. Both repos must agree on the typing wire and
deploy together.

## Why

Typing/voice ticks were gift-wrapped on the **kind-14 message rumor**, told
apart from real DMs only by content. That leaked **both ways**:
- the PWA rendered a `'stop'` tick as a "Stopped" bubble;
- the bot's inbound had **no kind guard at all** — a user's `'stop'` tick
  fails `JSON.parse`, falls through to the plain-text path, and the bot
  **replies to "stop"**.

## What changed

- `NOSTR_KIND_TYPING_RUMOR = 1414` (regular/non-ephemeral, `!= 14/15`, matches
  the PWA constant). `createRumor` gains an optional `kind` (default 14 — message
  path unchanged); `wrapTypingGiftWrap` / `wrapGroupTypingGiftWrap` stamp it.
- **`channel.listen` inbound guard:** drop rumors whose kind is
  `NOSTR_KIND_TYPING_RUMOR` before envelope parsing — structurally, so the bot
  never enqueues a tick as a turn. Discriminating by kind (not content) means a
  genuine user message that literally says "stop" (kind-14) **is still answered**.

## Verification

- `wrapTypingGiftWrap` stamps the dedicated inner kind (not 14).
- A live typing tick (start + stop) is dropped — no turn, no reply published.
- typecheck ✓ · full suite **1754/1754** ✓.

### DIFF
--- a/src/channels/phantomchat/channel.ts
+++ b/src/channels/phantomchat/channel.ts
@@ -36,6 +36,7 @@ import {
   unwrapNip17Message,
   unwrapV2,
   isV2Event,
+  NOSTR_KIND_TYPING_RUMOR,
   type NTNostrEvent,
 } from "../../lib/nostrCrypto.ts";
 import type { PhantomchatTransport } from "./transport.ts";
@@ -229,6 +230,18 @@ export function createPhantomchatChannel(
         // (4) Dedup by rumor id — the SAME logical message can arrive via more
         // than one wrap; the rumor id is stable across them.
         if (!remember(seenRumorIds, rumor.id)) return;
+
+        // (4a) Typing / voice tick — NEVER a message turn. The PWA gift-wraps
+        // typing on a dedicated inner kind (NOSTR_KIND_TYPING_RUMOR). Drop it
+        // structurally by kind: without this, a user's 'stop' tick fails
+        // JSON.parse, falls through to the plain-text path, and gets enqueued —
+        // i.e. the bot would reply to "stop". Discriminating by kind (not
+        // content) means a genuine user message that simply says "stop"
+        // (kind-14) still gets answered.
+        if (rumor.kind === NOSTR_KIND_TYPING_RUMOR) {
+          log.debug("phantomchat: dropping inbound typing tick");
+          return;
+        }
 
         // (5) Parse the JSON envelope. `type === "text"` is a chat message;
         // `presence-ping` / `presence-pong` are legacy types, silently dropped
--- a/src/lib/nostrCrypto.ts
+++ b/src/lib/nostrCrypto.ts
@@ -70,6 +70,22 @@ export interface SignedEvent extends UnsignedEvent {
 }
 
 /**
+ * Inner rumor kind for gift-wrapped typing / voice indicators. MUST stay in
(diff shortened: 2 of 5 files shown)


src/channels/phantomchat/channel.ts
```diff
@@ -230,6 +231,18 @@ export function createPhantomchatChannel(
...
         if (!remember(seenRumorIds, rumor.id)) return;
 
+        // (4a) Typing / voice tick — NEVER a message turn. The PWA gift-wraps
+        // typing on a dedicated inner kind (NOSTR_KIND_TYPING_RUMOR). Drop it
+        // structurally by kind: without this, a user's 'stop' tick fails
+        // JSON.parse, falls through to the plain-text path, and gets enqueued —
+        // i.e. the bot would reply to "stop". Discriminating by kind (not
+        // content) means a genuine user message that simply says "stop"
+        // (kind-14) still gets answered.
+        if (rumor.kind === NOSTR_KIND_TYPING_RUMOR) {
```
robertclawson, line 241: **Transitional caveat (non-blocking).** Unlike phantomchat #57 — which keeps a content-based fallback (`kind===14 && hasDTag && content∈['','stop','recording']`) to recognise old senders — the bot has no such fallback. During the deploy window a stale PWA client still emitting kind-14 ticks would slip past this guard and re-trip the reply-to-"stop" bug until it refreshes.

This is the pre-existing behaviour, not a regression, and kind-only is the correct principled choice here — adding content-sniffing back would reintroduce the exact ambiguity you're removing. Just worth a one-line mention in the deploy note so #237 and #57 ship together.

robertclawson (approved):
**Approved — sound, safe to merge** (gated on landing phantomchat #57 and deploying both together).

**Root cause is real and the fix is the right shape.** Typing/voice ticks were gift-wrapped on the message kind-14 rumor and told apart from real DMs only by sniffing content. The serious bot-side bug: a user's `'stop'` tick failed `JSON.parse`, fell through to the plain-text path, and the bot replied to "stop." Moving ticks to a dedicated inner kind (`1414`) and dropping them structurally by kind is correct.

**Why it's safe — the checks that matter:**
- The guard sits **after** the verifying unwrap; `rumor.kind` comes from the decrypted, seal-verified inner rumor (same NIP-59 binding as #56), so a hostile relay can't forge the kind to inject or suppress a turn.
- Discriminates by **kind, not content** — a genuine kind-14 message that literally says "stop" is still answered.
- `createRumor` defaults `kind = 14`, so the real message path is byte-for-byte unchanged; only the two typing wrappers stamp `1414`.
- Cross-repo constant agrees: phantomchat #57 also defines `NOSTR_KIND_TYPING_RUMOR = 1414` (non-ephemeral, `!= 14/15`, NIP-unassigned).

**Tests + CI:** new tests lock the inner-kind stamp, the live start+stop tick drop, and the wire shape. CI green — typecheck, test (1754/1754), CodeQL, both Analyze jobs. "blocked" mergeable_state is just the required-review gate.

Verdict: **approve & merge** once #57 lands and both deploy together.

PR #238, opened 2026-06-30
Fix anonymous Pi tool progress in ACP clients

kaieriksen:
## Summary

This fixes the anonymous `tool` rows Pi can produce in ACP clients such as Zed. Pi emits both top-level `tool_execution_start` events, which carry the useful tool name and arguments, and lower-level assistant `toolcall_*` / `tool_use_*` stream events, which can be nameless internal deltas.

## Description

The Pi adapter now treats nameless assistant toolcall stream events as heartbeat-only liveness signals instead of ACP progress updates. If one of those assistant events does carry a real tool name or useful argument detail, it can still surface a meaningful progress title. The top-level `tool_execution_start` path remains the canonical source for visible tool progress, so useful titles like `bash: npm test` are preserved.

## Why

Zed renders phantombot progress chunks as tool call rows. Mapping Pi's anonymous internal toolcall deltas to progress created repeated bare `tool` rows, which made Pi look noisier and less polished than Claude despite the real tool execution event arriving separately with better metadata.

## Testing

- `~/.bun/bin/bun test tests/harnesses-pi.test.ts`
- `PATH="$HOME/.bun/bin:$PATH" bun run typecheck`
- `~/.bun/bin/bun test`
- `PATH="$HOME/.bun/bin:$PATH" bun run build`

### DIFF
--- a/src/harnesses/pi.ts
+++ b/src/harnesses/pi.ts
@@ -296,10 +296,10 @@ export function renderPayload(req: HarnessRequest): string {
  * time, that's pi actually thinking — the indicator vanishing means the
  * model has gone silent (and may be wedged on a tool call).
  *
- * toolcall_* (pi ≥0.79; formerly tool_use_*) event types inside
- * assistantMessageEvent → `progress` so the channel layer flushes narration
- * into a bubble before the tool runs. thinking_delta + other event types →
- * `heartbeat`.
+ * tool_execution_start carries the useful tool name/args and is the primary
+ * progress signal. Nameless toolcall_* (pi ≥0.79; formerly tool_use_*) events
+ * inside assistantMessageEvent are only liveness noise, so they become
+ * `heartbeat` unless they carry a real tool name or useful args.
  *
  * Exported for testing.
  */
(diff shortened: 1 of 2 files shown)


lenaparkhodges (approved):
Clean fix. The approach is correct — nameless toolcall deltas become heartbeats, named ones get a proper title via `buildToolNote`, and `tool_execution_start` stays the canonical progress signal.

One minor observation (non-blocking): `extractUsefulArgDetail` in pi.ts overlaps significantly with `extractToolDetail` in toolNote.ts — same keys (command, cmd, file_path, filePath, path, query, pattern, prompt) but pi.ts is missing `absolute_path`, `url`, `subagent_type`, `description`. Since this only affects the no-tool-name fallback path (where `buildToolNote` isn't called), it's a cosmetic inconsistency, not a bug. Worth a follow-up to consolidate if the key list grows, but fine for now.

Tests are good — covers anonymous (heartbeat), named (progress with title), and args-only (progress with detail) paths. TypeScript clean, 1756 tests pass.

PR #58, opened 2026-05-03
feat: heartbeat chunks + chunk-driven typing indicator (no timers, no fakes)

andrewagrahamhodges:
## Summary

Replaces the random-pulse approach with **pure event-driven** typing indicator refresh. Every refresh corresponds to an actual signal the harness emitted. When the harness goes silent, the Telegram chat-action lapses after ~5s — and that vanishing IS the truthful "model has gone quiet / possibly frozen" cue.

Per the design discussion: "if people read the code and see random shit here, they will feel defrauded." Now there's no random shit.

## New chunk type

`HarnessChunk` gains a **`heartbeat`** variant — payload-less; means "model is alive and producing reasoning, but nothing user-visible yet." The content of thinking blocks / tool deltas stays inside the harness subprocess; the channel layer just sees the existence of activity.

## All three harnesses now emit real signals

| Harness | Signal source | Emits |
|---|---|---|
| **Claude** stream-json | `assistant` content blocks of type `thinking` / `tool_use` / `tool_result` | `heartbeat` |
| **Pi** --mode json | `assistantMessageEvent` of type `thinking_delta` and other non-text variants | `heartbeat` |
| **Gemini** stream-json (NEW) | `tool_use` / `tool_result` events | `progress` (with tool name) / `heartbeat` |

**Chain-of-thought content is never embedded in the chunk** — explicit test assertions verify this for both Claude and Pi.

## Gemini port: -o text → -o stream-json

The biggest individual change. New `parseGeminiEvent` pure mapper covers the full schema (verified against gemini-cli v0.40.1):

- `init` → ignored (session-start chatter)
- `message` `role:user` → ignored (input echo)
- `message` `role:assistant` → `text` chunk
- `tool_use` → `progress` with `note: "tool: <name>"` — also feeds `/status` running-line, no gemini-specific knowledge needed elsewhere
- `tool_result` → `heartbeat`
- `result` → `done` carrying `stats` (token counts + duration) in meta

Final `done` chunk now includes `meta.stats` for logging / `/status`.

## Telegram channel

- **All timer code deleted**: no `setInterval`, no `setTimeout`-pulse, no random interval picker, no `PULSE_INTERVALS_MS` export, no `typingRefreshMs` injection point.
- New `typingThrottleMs` (default 2000ms) prevents stream-json bursts from spamming Telegram's chat-action endpoint.
- Indicator refreshes on every `text` / `heartbeat` / `progress` chunk, plus an initial nudge at turn start. `done` doesn't refresh.
[...]

PR #239, opened 2026-06-30
revert(typing): back to ephemeral kind-20001 to match phantomchat #58

robertclawson:
## Why

phantomchat **#58** reverts its typing transport back to the known-good #54 state — plain **NIP-16 ephemeral kind-20001** typing events. That leaves phantombot stranded: after **#236** (kind-20001 → kind-30001 → NIP-17 gift-wrapped kind-14) and **#237** (dedicated gift-wrapped kind-1414), the bot speaks a wire protocol the PWA no longer understands. The two repos would silently fail to exchange typing/voice ticks.

This PR reverts phantombot's typing transport to ephemeral kind-20001 so both sides agree again.

## What

- **Revert #237** (`6b8903a`) — removes the dedicated `NOSTR_KIND_TYPING_RUMOR = 1414`, the `createRumor` kind param, the typing gift-wrap kind stamping, and the inbound kind guard in `channel.listen`.
- **Revert #236** (`08aa968`) — removes the kind-30001 (NIP-33) migration and the subsequent NIP-17 gift-wrap of typing ticks; restores `NOSTR_KIND_TYPING = 20001` plain ephemeral publish in `transport.ts`.

Net result: the typing-transport files (`transport.ts`, `nostrCrypto.ts` typing paths, `channel.ts`, `server.ts`) are **byte-identical to their pre-#236 state** — the ephemeral kind-20001 baseline that matches phantomchat #58. Unrelated changes from #235/#238 are preserved.

## Trade-off carried over

Going back to ephemeral kind-20001 also re-adopts the pre-#236 behaviour: typing ticks are bare ephemeral events (no gift-wrap privacy, content markers `''`/`'stop'`/`'recording'`). The reply-to-"stop" bug #237 fixed lived in the **gift-wrapped kind-14** inbound path, which no longer exists here, so it does not reappear — inbound ticks are ephemeral kind-20001 and never reach the DM turn path. This is the deliberate product decision recorded in phantomchat #58 ("removing the ephemeral path did not improve behaviour").

## Verification

- Typing tests green: `channels-phantomchat-channel`, `channels-phantomchat-transport`, `lib-nostrCrypto` — 51 pass.
- Full suite: **1754 pass, 1 skip, 0 fail**.
- `tsc --noEmit`: clean.

**Deploy coupling:** merge/deploy together with phantomchat #58.

🤖 Opened by Robbie on Andrew's instruction (pairs with phantomchat #58).


andrewagrahamhodges (approved):
Verified the core claim: all four typing source files (channel.ts, server.ts, transport.ts, nostrCrypto.ts) on the PR head are byte-identical to the pre-#236 baseline (42bbf03) — NOSTR_KIND_TYPING = 20001, plain NIP-16 ephemeral. Only #236/#237 ever touched these files, so nothing unrelated is lost; #235/#238 changes live elsewhere and are untouched. CI green (test/typecheck/CodeQL). Note: deploy-coupled with phantomchat #58 — must ship together.

PR #240, opened 2026-07-01
fix(harness): put the harness binary's own dir on the child PATH so nvm shebangs resolve under systemd

andrewagrahamhodges:
## Why

When **Pi is the primary harness**, the bots answer from the **Zed / VS Code extension** but go **silent on Telegram and phantombot's own channels**. Lena (on the Pi) answers; Claude-primary works everywhere. That asymmetry is the tell.

### Root cause: shebang interpreter + narrow systemd PATH

`pi` (and any nvm/fnm/volta-installed `claude`/`gemini`/`codex`) is a **Node script** with a `#!/usr/bin/env node` shebang:

```
~/.nvm/versions/node/v24.14.1/bin/pi → ../lib/node_modules/@earendil-works/pi-coding-agent/dist/cli.js
```

Executing it makes the kernel re-invoke `env node`, which needs **`node` on PATH**. For an nvm install, `node` sits **in the same directory as the binary** — a directory that's on PATH inside a desktop/editor session but **not** under systemd's narrow unit PATH.

phantombot's `harnessSearchPath()` is already smart enough to **find** `pi` on the filesystem and spawn it by absolute path, so the spawn succeeds. But the child inherits phantombot's PATH, and under systemd that PATH has no nvm dir → the shebang's `env node` fails → **exit 127** → no reply. Even the recovery reply fails (it retries the same harness).

- **Why Zed/VS Code works:** the extension host inherits the desktop session's PATH, where nvm has already put `node`. The child `pi` inherits that richer PATH and the shebang resolves.
- **Why Telegram doesn't:** systemd launches phantombot with a stripped PATH (no `node`), so the shebang breaks.

Evidence from the journal when Pi was primary:
```
orchestrator: trying harness  harnessId:pi  attempt:1  of:1
telegram: turn failed  error:"pi exited with code 127"
```

## The fix

At the **single choke point all four harnesses share** — `spawnInNewSession` — prepend the resolved binary's own directory to the child PATH (`withCommandDirOnPath`). `node` lives right beside the binary, so the shebang resolves regardless of how narrow the parent PATH is. This makes systemd behave like the editor **without tampering with the user's shell config or the systemd unit**.

Safe by construction:
- **No-op for bare command names** (`"pi"`) — those resolve via the existing PATH, and `dirname("pi")` is `"."`, which we never inject.
- **No-op when the dir is already present** (returns the same env ref).
- **Never mutates the caller's env** — returns a fresh object (the caller often passes the shared `process.env` reference).
- Applies to all four harnesses + any future one for free.

## Tests

- Unit coverage for `withCommandDirOnPath`: absolute vs bare, empty/absent PATH, already-present, no-mutation.
- **Real-subprocess repro of the bug**: a `#!/usr/bin/env node` script placed beside a `node` symlink, spawned under a stripped `PATH=/usr/bin:/bin`, asserts **exit 0 (not 127)** and correct output.

All 11 processGroup tests + 122 harness tests pass; `tsc --noEmit` clean.

### DIFF
--- a/src/lib/processGroup.ts
+++ b/src/lib/processGroup.ts
@@ -36,8 +36,56 @@
  * cmd in `setsid` and accept the spawn-time race.
  */
 
+import { dirname } from "node:path";
 import type { Subprocess, SpawnOptions } from "bun";
 import { log } from "./logger.ts";
+
+/**
+ * Ensure an absolute executable's OWN directory is on the child's PATH.
+ *
+ * Why this exists (the systemd-vs-Zed harness bug, 2026-07-01): the harness
+ * CLIs (`pi`, and any nvm/fnm/volta-installed `claude`/`gemini`/`codex`) are
+ * Node scripts whose shebang is `#!/usr/bin/env node`. The kernel honors the
+ * shebang by re-invoking `env node`, which needs `node` ON PATH. An
+ * nvm-installed `pi` lives at `~/.nvm/versions/node/<v>/bin/pi` with `node`
+ * RIGHT NEXT TO IT in the same directory — but that directory is only on PATH
+ * inside an interactive/desktop session.
+ *
+ * phantombot's `harnessSearchPath()` is smart enough to FIND `pi` on the
+ * filesystem and spawn it by absolute path, so the spawn itself succeeds. But
+ * the child inherits phantombot's PATH, and under systemd that PATH is narrow
+ * (no nvm dir) — so the shebang's `env node` fails and the process dies with
+ * exit 127. Under Zed / VS Code the extension host inherits the desktop
+ * session's PATH (nvm already on it), so `node` is found and it Just Works —
+ * which is exactly why the same bot answers from the editor but not Telegram.
+ *
+ * The fix is transport-agnostic: prepend the executable's own directory to the
+ * child PATH so `env node` resolves the interpreter sitting beside the binary,
+ * no matter how narrow the parent PATH is. This makes systemd behave like the
+ * editor without tampering with the user's shell/systemd config.
+ *
+ * Only acts when `bin` is ABSOLUTE — a bare command name (`"pi"`) is being
+ * resolved via the existing PATH, and `dirname("pi")` is `"."`, which we must
+ * never inject. Returns a FRESH object (never mutates the caller's env, which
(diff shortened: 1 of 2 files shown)


src/lib/processGroup.ts
```diff
@@ -57,8 +105,17 @@ export function spawnInNewSession<
   if (cmd.length === 0) {
     throw new Error("spawnInNewSession: cmd cannot be empty");
   }
+  // Make the child's PATH self-sufficient for a `#!/usr/bin/env node` shebang:
+  // prepend the binary's own directory (where the matching `node` lives for an
+  // nvm/fnm/volta install) so the interpreter resolves under a narrow systemd
+  // PATH exactly as it does under a desktop/editor session. See
+  // withCommandDirOnPath for the full rationale (exit-127 bug, 2026-07-01).
+  const env = opts.env
```
robertclawson, line 113: Non-blocking: the fix only fires when the caller passes `opts.env` (the false branch just returns `opts.env` unchanged). Every one of the four harnesses passes an explicit `env` today, so this is fully covered — but a *future* spawn that relies on bare `process.env` inheritance would silently bypass `withCommandDirOnPath` and could re-trip the exit-127 bug. Worth a one-line comment here noting the guarantee depends on callers passing `env`. Flag, not a fix.

robertclawson (approved):
Reviewed end to end — the diff, all four harness call sites (pi/claude/codex/gemini all spawn with explicit `env`), the absolute `config.bin` resolution, the tests, and CI (typecheck/test/CodeQL/both Analyze all green).

**Root cause is real and correctly diagnosed:** an nvm/fnm/volta-installed `#!/usr/bin/env node` shim has `node` beside it, on PATH inside an editor session but not under systemd's narrow PATH → `env node` fails → exit 127 → silent on Telegram while answering fine from Zed. Prepending the resolved binary's own dir to the child PATH fixes it at the right layer, transport-agnostic, without touching shell/unit config.

**Safe by construction:** no-op for bare names (`dirname("pi")` guard), no-op when the dir is already present (same ref back, no clone churn), never mutates the caller's shared `process.env`. Tests cover every branch plus a real subprocess repro of the exit-127 bug. Approving. 🐶

PR #241, opened 2026-07-03
feat(acp): clickable file paths + per-tool kind on tool-call updates (#231)

### LINKED ISSUE #231: Tool-call titles: clickable file paths + icons in the Zed/ACP panel (Part 2 of #218)
Follow-up to #218 / PR #230. Part 1 (richer `title` strings — `Bash: git status`, `Read: src/foo.ts`, etc.) shipped in #230. This issue tracks **Part 2**, the protocol-level polish that Part 1 deliberately left out because it has a larger blast radius and deserves its own review.

## Goal

Make tool calls render in the Zed/ACP panel with:
- **Clickable file paths** (jump-to-file for Read/Edit/Write)
- **Per-tool icons / `kind`** (e.g. file-read vs. file-edit vs. execute vs. search)
- Optional richer **content** preview

## Why it's separate from Part 1

Part 1 only reshaped the free-form `title` string the harness already emits, so it touched zero protocol surface. Part 2 requires extending the ACP `ToolCallUpdate` type and changing a callback signature:

1. **`ToolCallUpdate` type** currently carries only `title` / `status` / `kind`. Needs `locations` (for clickable paths) and `content` (for previews).
2. **`server.ts` progress callback** today is `progress: (note: string) => …` — a bare string. To carry locations/kind/content it has to become a structured object, which means every harness's `progress(...)` call site changes shape.
3. **`toolCallUpdate()` builder** needs to populate the new fields and set `kind` per tool.

## Suggested approach

- Extend `buildToolNote` (or add a sibling `buildToolCall`) to return a structured `{ title, kind, locations, content }` instead of just a string.
- Map tool → `kind`: Read→`read`, Edit/Write→`edit`, Bash→`execute`, Agent→`other`, search tools→`search` (match ACP's `ToolKind` enum).
- Extract `locations` from `file_path` (and Edit/Write path args) so Zed makes them clickable.
- Thread the structured object through `server.ts` and all four harness adapters.

## Out of scope / carry-overs from Part 1
- Redaction is still **minimal** (truncate only). If `content` previews start surfacing more of the Bash command than the title did, revisit secret-masking before merging — flagged in the `toolNote.ts` module doc.

## Acceptance
- File paths in the Zed panel are clickable and open the right file.
- Tool calls show appropriate icons/`kind`.
- Existing `title` behaviour from #230 unchanged for harnesses/clients that don't consume the new fields.

andrewagrahamhodges:
Closes #231 (Part 2 of #218).

## What & why
Part 1 (#230) enriched the tool-call **title** string in the Zed/ACP panel (`Bash: git status`, `Read: src/foo.ts`). This is the protocol-level Part 2 it deliberately deferred: tool calls now render with **per-tool icons** (`kind`) and **clickable jump-to-file paths** (`locations`).

It's still our side — Zed faithfully renders whatever the ACP bridge sends. We just populate the structured `ToolCallUpdate` fields we weren't before.

## How
- **`toolNote.ts` — new `buildToolCall(name, input)`** returns a structured `{ title, kind, locations, content }`. `title` is **byte-identical** to `buildToolNote`, so every existing string-only consumer (tick logging, channel sinks) and all of #218's title tests are untouched.
  - **`kind`** is keyed on the tool **name** first — `Read`→`read`, `Edit`/`Write`/`apply_patch`→`edit`, `Bash`/`run_shell_command`→`execute`, `Grep`/`Glob`→`search`, `WebFetch`→`fetch`, `Task`→`other` — across per-harness spellings (name is normalised: lowercased, spaces/hyphens folded). Unknown tools fall back to input **field** names (`command`→execute, `pattern`/`query`→search, `url`→fetch, `file_path`→read), mirroring how the title detail is extracted.
  - **`locations`** are pulled from path fields (`file_path`/`filePath`/`path`/`absolute_path`/`notebook_path`), deduped and empties skipped. Non-file tools contribute none.
- **Threaded end-to-end.** The `progress` `HarnessChunk` gains an **optional** `tool?: ToolCallDetail` — `note` stays exactly as-is, so string-only sinks ignore it. The four harness adapters (claude/codex/gemini/pi) emit it; `BridgeSink.progress` and `server.ts` forward it; `ToolCallUpdate` + `toolCallUpdate()` gain `kind`/`locations`/`content`, **each emitted only when present**.

## Scope decision — `content` deferred
The wire type carries `content` (ACP preview body) and the builder threads it, but `buildToolCall` leaves it **unpopulated**. The only readily-available preview is the raw command/args, and surfacing more of a Bash command than the (already-truncated, **un-masked**) title does would reopen the **secret-masking** carry-over flagged in Part 1. Documented on `ToolCallDetail`; populate once redaction lands. This keeps the redaction blast radius at zero while completing the protocol surface.

## Backward compatibility
Title-only updates emit the **exact pre-#231 shape** — `kind`/`locations`/`content` are absent unless populated. Clients that don't consume the new fields are unaffected.

## Tests
- **`buildToolCall`** unit coverage: title parity with `buildToolNote`, kind mapping across per-harness spellings + name normalisation + field-name fallback, location extraction/dedupe/skip-empty, hostile-input safety.
- **New `connectors-acp-protocol.test.ts`**: proves `kind`/`locations`/`content` land on the wire **only when present** (empty `locations` omitted, title-only stays minimal).
- Per-harness wiring tests (claude/codex/gemini/pi) updated to assert the structured `tool` detail flows through.
- Full suite green (1773 pass), `tsc --noEmit` clean.

## Acceptance
- ✅ File paths in the Zed panel are clickable and open the right file (`locations`).
- ✅ Tool calls show appropriate icons/`kind`.
- ✅ Existing `title` behaviour from #230 unchanged for harnesses/clients that don't consume the new fields.

### DIFF
--- a/src/harnesses/toolNote.ts
+++ b/src/harnesses/toolNote.ts
@@ -46,7 +46,14 @@ export type ToolKind =
 
 /** A file location the editor can render as a clickable jump-to-file link. */
 export interface ToolLocation {
-  /** Absolute or workspace-relative path the editor should open. */
+  /**
+   * The path the editor should open. As extracted here it is whatever the
+   * harness tool arg contained — usually relative to the session cwd
+   * (`src/foo.ts`), occasionally already absolute. The ACP boundary
+   * (`toAbsoluteLocations` in connectors/acp/protocol.ts) resolves relative
+   * paths against the session cwd so the wire always carries an ABSOLUTE path,
+   * as the ACP `ToolCallLocation.path` spec requires.
+   */
   path: string;
   /** 1-based line to jump to, when known. Optional — file open still works. */
   line?: number;
--- a/src/harnesses/toolNote.ts
+++ b/src/harnesses/toolNote.ts
@@ -30,6 +30,52 @@
 /** Max length of a rendered tool-call title, including the name prefix. */
 export const MAX_TOOL_NOTE_LEN = 80;
 
+/**
+ * ACP `ToolKind` — drives the per-tool icon Zed renders in its agent panel.
+ * Matches the ACP `ToolKind` enum exactly; `other` is the safe catch-all.
+ */
+export type ToolKind =
+  | "read"
+  | "edit"
+  | "delete"
+  | "move"
+  | "search"
+  | "execute"
+  | "fetch"
+  | "other";
+
+/** A file location the editor can render as a clickable jump-to-file link. */
+export interface ToolLocation {
+  /** Absolute or workspace-relative path the editor should open. */
+  path: string;
+  /** 1-based line to jump to, when known. Optional — file open still works. */
+  line?: number;
+}
+
+/**
+ * Structured tool-call detail (issue #231, Part 2). A superset of the
+ * legacy `buildToolNote` string: `title` is byte-identical to what
+ * `buildToolNote` returns, and the extra fields drive the ACP panel's
+ * icon (`kind`) and clickable paths (`locations`).
+ */
+export interface ToolCallDetail {
+  /** Single-line, length-capped title — identical to `buildToolNote(...)`. */
+  title: string;
+  /** ACP ToolKind, used for the panel icon. */
+  kind: ToolKind;
+  /** File paths to surface as clickable links. Empty for non-file tools. */
+  locations: ToolLocation[];
+  /**
+   * Optional richer preview body. Intentionally left unpopulated for now:
+   * the only readily-available preview is the raw command/args, and
+   * surfacing more of a Bash command than the (already-truncated, un-masked)
+   * title does would reopen the secret-masking question flagged as a
+   * carry-over in #218 Part 1 / #231. The field is plumbed so the wire type
+   * is complete; populate it once redaction lands.
+   */
+  content?: string;
+}
+
 function isRecord(v: unknown): v is Record<string, unknown> {
   return typeof v === "object" && v !== null && !Array.isArray(v);
 }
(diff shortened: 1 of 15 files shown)


src/harnesses/toolNote.ts
```diff
@@ -30,6 +30,52 @@
...
+  | "delete"
+  | "move"
+  | "search"
+  | "execute"
+  | "fetch"
+  | "other";
+
+/** A file location the editor can render as a clickable jump-to-file link. */
+export interface ToolLocation {
+  /** Absolute or workspace-relative path the editor should open. */
```
kaieriksen: ACP `ToolCallLocation.path` is specified as the absolute file path being accessed/modified, but this new type and extractor deliberately allow and emit workspace-relative paths (`src/foo.ts`, `a.ts`). Most harness tool args are relative to the session cwd, so the PR will send non-conformant locations and Zed may not be able to open/follow them. Please resolve relative paths against the ACP session cwd before putting them on the wire, and update the tests to assert absolute paths.

kaieriksen (changes requested):
Found one protocol issue that will keep the clickable-location part from working reliably.

andrewagrahamhodges:
Good catch — fixed in 0cad762.

The extractor still records paths as the harness handed them (usually session-relative), but they are now resolved to absolute at the single wire choke point: `toolCallUpdate` calls a new `toAbsoluteLocations(locations, cwd)` helper before emitting, and `server.ts` threads `session.cwd` in. Already-absolute paths pass through untouched; relative ones get `path.resolve(cwd, p)` (pure `node:path`, no I/O).

Resolving at `toolCallUpdate` rather than in the extractor means no relative location can escape to the wire regardless of which harness produced it. Tests now assert absolute paths on the wire, plus direct `toAbsoluteLocations` coverage (relative resolve, `..` normalisation, absolute passthrough, `line` preservation). Full suite green (1780).

kaieriksen (approved):
Re-reviewed 0cad762. The absolute-path fix addresses my previous request: relative harness paths are resolved at the ACP wire boundary, absolute paths pass through, and coverage now locks the behavior down. Local focused tests, typecheck, full suite, build, CI, and merge-tree all pass.

Issue #245, opened 2026-07-03
Tech debt: de-duplicate the mirrored streaming loop (engine.ts ↔ phantomchat/server.ts)


andrewagrahamhodges:
## Context

The Telegram engine (`src/channels/core/engine.ts`) and the PhantomChat server (`src/channels/phantomchat/server.ts`) each run their **own copy** of the streaming state machine. `phantomchat/server.ts` deliberately mirrors `core/engine.ts` (its own comments say so), so any change to narration/bubble handling has to be made in **two places**. This surfaced while shipping `/chattiness` (#243), where the narration gate had to be applied at two separate `flushNarration` sites.

## Decision (2026-07-03)

Split into two tiers. **Tier A is being done. Tier B is explicitly NOT planned** — see the reasoning below.

## Tier A — extract the byte-identical helpers ✅ (in progress)

Lift the genuinely shared, pure/near-pure pieces into `src/channels/core/streaming.ts`, consumed by both channels:
- `segmenterOptionsFor(streaming)` — the bubble-sizing options.
- `resolveOutgoingSuffix(fullReply, streamedReply, consumedReplyChars)` — the post-loop suffix-vs-full dedupe computation.
- `createNarrationController({ streaming, enabled, send, suppress? })` — owns the narration buffer + flush clock; the `suppress` callback carries Telegram's `willReplyWithVoice` guard, PhantomChat passes nothing.

No transport addressing and no divergent failure paths are touched. This removes the duplicated narration/dedupe logic and the paired `TODO(dedup)` breadcrumbs, and is covered by a new unit suite plus the existing channel tests.

## Tier B — unify the full loop + orchestration — WON'T DO (for now)

Originally floated as a follow-up. After measuring the actual divergence, we're **parking it deliberately**, not scheduling it. The reasoning:

**The two loops diverge exactly where correctness is fragile, and they diverge for real reasons — they aren't two copies of one thing that drifted:**

1. **Transport addressing is genuinely different.** Telegram keys everything on an opaque `conversationId`; PhantomChat keys on `senderHex` + `groupId` and has group-only methods (`sendGroupMessage`, `sendGroupTyping`) that Telegram lacks entirely. There is no shared transport abstraction today — unifying the loop means *inventing* one over addressing models that legitimately differ (this is protocol/routing correctness, e.g. group typing must p-tag every member so the dots land in the group chat, not the DM).
[...]

PR #246, opened 2026-07-03
feat(chattiness): /chattiness quiet-mode toggle for narration bubbles (#243)

### LINKED ISSUE #243: Add /chattiness slash command to control progress bubbles
## Problem

Right now, while the phantom is working on a task, it sends a progress "bubble" on every step telling the user what's happening. For longer tasks this is chatty and noisy — the user gets a stream of intermediate updates they didn't ask for.

## Proposal

Add a `/chattiness` slash command that controls whether the phantom emits interim progress bubbles.

- `/chattiness on` → current behaviour: interim progress bubbles as normal.
- `/chattiness off` → phantom works quietly, no interim bubbles, posts one **final summary** at the end ("background mode").
- `/chattiness <on|off> default` → sets the standing default (writes config), not just this conversation.

## Behaviour (decided)

- **Sticky for the whole conversation** → once set, `/chattiness on|off` stays for that conversation until toggled again. Not per-invocation.
- **Errors break the silence** → with chattiness off, blockers/errors are still surfaced mid-task. Only routine progress bubbles are suppressed. The final summary still posts at the end.
- **New installs default to chattiness ON** → out of the box, progress bubbles behave exactly as today. Nothing changes for existing users unless they opt into quiet mode.
- **Scope: Telegram and PhantomChat only.** This setting must not affect any other surface (voice, CLI, etc.).

## Setting a default

Two entry points, one source of truth:

- **`config.toml`** holds the canonical setting (e.g. `chattiness = true/false`, defaulting to `true` on new installs) — the standing default applied to Telegram + PhantomChat.
- **`/chattiness <on|off> default`** is the friendly toggle: the command *writes* the config value so users never hand-edit the file.
- Per-conversation `/chattiness on|off` (sticky) still overrides for that one chat.

## Notes / open questions

- How does this interact with the final-summary formatting?
- Registration alongside the other slash commands in the Telegram + PhantomChat handlers.
- Precedence: per-conversation toggle vs. the config default when both are set.

---
_Blocked on #241 merging first, then we start work here._

andrewagrahamhodges:
## Summary

Adds a `/chattiness` slash command to let users suppress **narration bubbles** — the model's interim "let me check… / now let me…" chatter that streams while a turn runs. Closes #243.

In a chat, some users don't want to see the running reasoning/progress commentary — just the final answer. This gives them the switch. Default stays **ON** (current behaviour), and can be flipped **OFF globally** for people who never want it.

## Behaviour

- `/chattiness on` — narration bubbles stream as they do today.
- `/chattiness off` — quiet mode: no interim narration, user gets the **final reply** only.
- `/chattiness <on|off> default` — writes the standing default to `config.toml`.
- **Per-conversation sticky** — a toggle holds for that chat until changed, overriding the config default.
- Final reply + error paths are **never** touched — only the narration flush is gated.

## Scope

- **Telegram + PhantomChat only.** CLI, voice, and the editor (ACP) surface are deliberately left untouched.
- Gated at **both** `flushNarration` sites — `engine.ts` (Telegram) and the mirrored loop in `phantomchat/server.ts` — each with a `TODO(dedup)` breadcrumb.

## Implementation

- **`src/lib/chattiness.ts`** (new) — per-conversation override store (cloned from `coderSwap.ts`) + `resolveNarrationEnabled()`: per-conversation override wins, else the config default.
- **`src/channels/commands.ts`** — `/chattiness` case, handler, and `TELEGRAM_BOT_COMMANDS` entry (shared by both channels).
- **`src/config.ts`** — top-level `chattiness` bool (default `true`), env + toml parse.
- **`src/channels/core/engine.ts` / `phantomchat/server.ts`** — one-line gate on `flushNarration`, resolved once per turn (sticky).

## Follow-up

The two streaming loops are near-duplicates. Rather than fold a risky refactor into this feature, that tech debt is tracked separately in #245 (Tier A cheap/safe extraction, Tier B its own design pass).

## Docs

README (new quiet-mode section + command table + config key), AGENTS.md, and docs/architecture.md all updated.

## Tests

- `tests/lib-chattiness.test.ts` — store + precedence.
- `tests/channels-commands.test.ts` — `/chattiness` on/off/default, config-write path, garbage rejection.

Typecheck clean; affected suites green.

### DIFF
--- a/src/channels/commands.ts
+++ b/src/channels/commands.ts
@@ -32,6 +32,11 @@ import {
   applyCoderSwapRequest,
   normalizeCoderSwapRequest,
 } from "../lib/coderSwap.ts";
+import {
+  applyChattinessRequest,
+  normalizeChattinessRequest,
+} from "../lib/chattiness.ts";
+import { setIn, updateConfigToml } from "../lib/configWriter.ts";
 import type { MemoryStore } from "../memory/store.ts";
 import { DEFAULT_HISTORY_LIMIT } from "../orchestrator/turn.ts";
 import { VERSION } from "../version.ts";
@@ -138,6 +143,10 @@ export const TELEGRAM_BOT_COMMANDS: Array<{
     command: "coder",
     description: "Force the coding brain on for this chat (off | default to revert)",
   },
+  {
+    command: "chattiness",
+    description: "Show/hide progress bubbles here (on | off | <on|off> default)",
+  },
   { command: "help", description: "Show this command list" },
 ];
 
@@ -221,6 +230,8 @@ export async function handleSlashCommand(
     case "/coder":
       // Bare `/coder` forces on; `/coder off|default` is also accepted.
       return await handleCoderSwap(arg || "on", ctx);
+    case "/chattiness":
+      return await handleChattiness(arg, ctx);
     case "/start":
     case "/help":
       return { reply: HELP };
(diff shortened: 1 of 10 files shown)


src/channels/commands.ts
```diff
@@ -350,6 +361,103 @@ async function handleCoderSwap(
...
+  // what the user just asked for right here).
+  if (setDefault) {
+    const enable = request === "on";
+    if (!ctx.config) {
+      return {
+        reply:
+          "can't set the default: channel didn't pass config to the dispatcher",
+      };
+    }
+    await updateConfigToml(ctx.config.configPath, (c) => {
```
kaieriksen, line 418: After writing the new standing default to config.toml, the running process still keeps using the old loaded Config object. The next turns call resolveNarrationEnabled(... configDefault: input.config.chattiness ...), and this branch also clears the current chat override, so `/chattiness off default` can reply that quiet mode is now the default while this chat and every chat without an override keep using the previous default until a restart/reload. Please update the in-memory config after a successful write, e.g. set `ctx.config.chattiness = enable`, and cover that with a command test that resolves against the same config object after the default change.

kaieriksen (changes requested):
I found one behavior bug in the default-toggle path.

andrewagrahamhodges:
Good catch @kaieriksen — confirmed and fixed in 04949e8.

You're right on both counts: the write hit `config.toml` but the running process kept the stale in-memory `Config`, and because the same branch clears this chat's override, `/chattiness off default` would announce quiet-by-default while this chat and every override-less chat kept the old behavior until a restart/reload.

Fix: mutate `ctx.config.chattiness = enable` in place right after a successful `updateConfigToml`, so `resolveNarrationEnabled` on subsequent turns sees the new default immediately.

Added a command test that resolves `resolveNarrationEnabled` against the *same* config object before/after `off default` (starting from the ON default) and asserts an override-less chat flips to quiet with no reload. Full suite green, typecheck clean.

kaieriksen (approved):
Re-reviewed the update. The live config is now kept in sync after writing the standing default, and the added regression covers the stale-config path I flagged.\n\nVerified locally: focused chattiness/command tests, typecheck, full suite, and production build.

PR #248, opened 2026-07-03
refactor(channels): extract shared streaming helpers (Tier A of #245)

andrewagrahamhodges:
Tier A of #245 — extract the byte-identical pieces of the mirrored streaming loop into a single shared module so narration/bubble logic has one source of truth. Pure refactor: **no behaviour change**.

## What moved

New `src/channels/core/streaming.ts`, consumed by both `core/engine.ts` (Telegram) and `phantomchat/server.ts` (PhantomChat):

- **`segmenterOptionsFor(streaming)`** — the `{ maxSentences, maxChars }` bubble-sizing options (was constructed inline in both).
- **`resolveOutgoingSuffix(fullReply, streamedReply, consumedReplyChars)`** — the post-loop "suffix vs full" reconciliation (was a byte-identical `else if / else` in both).
- **`createNarrationController({ streaming, enabled, send, suppress? })`** — owns the narration buffer + flush clock and the throttled `flush()`. The optional `suppress()` callback carries Telegram's `if (willReplyWithVoice) return` guard; PhantomChat passes nothing (voice is DM-only and gated elsewhere).

## Behaviour-preserving

- Guard **ordering** is identical (`suppress` → `enabled` → empty-buffer → throttle), matching the old inlined `flushNarration` in each file.
- Throttle window and dedupe semantics unchanged.
- The final-answer splitter state stays in each caller — its per-segment send is interleaved with channel-specific transport, so it's not part of this extraction.
- The paired `TODO(dedup)` breadcrumbs are replaced with a short `NOTE(dedup)` pointer to the shared module + #245.

## Why not the whole loop (Tier B)

Deliberately out of scope, and per the decision on #245, **not planned**. The two loops diverge for real reasons — transport addressing (`conversationId` vs `senderHex`/`groupId` + group routing) and Telegram-only failure handling (recovery, abort persistence, `(no reply)`) — that don't co-evolve, so merging them would trade a clean, revertable state for a rewrite of the most silent-failure-prone code in the app for no real payoff. Full reasoning is in #245.

## Tests

- New `tests/channels-core-streaming.test.ts` (11 cases): segmenter mapping, dedupe suffix/full/reformatted/empty, and the narration controller (flush, buffer-clear, empty guard, disabled drop, live `suppress`, throttle + `force`).
- Full suite green: **1812 pass / 0 fail**, `tsc --noEmit` clean.

Net: `+338 / −76` across the two channels + the new module + tests.

### DIFF
--- a/src/channels/core/engine.ts
+++ b/src/channels/core/engine.ts
@@ -63,6 +63,11 @@ import {
   StreamSegmenter,
 } from "../streamSegmenter.ts";
 import {
+  createNarrationController,
+  resolveOutgoingSuffix,
+  segmenterOptionsFor,
+} from "./streaming.ts";
+import {
   formatAttachmentUserText,
   formatReplyToContext,
   inboxDir,
@@ -881,10 +886,7 @@ async function processChatMessage(
       ? input.transport.sendRecording(msg.conversationId)
       : input.transport.sendTyping(msg.conversationId);
   const streaming = input.config.telegramStreaming ?? DEFAULT_TELEGRAM_STREAMING;
-  const segmenterOptions = {
-    maxSentences: streaming.bubbleMaxSentences,
-    maxChars: streaming.bubbleMaxChars,
-  };
+  const segmenterOptions = segmenterOptionsFor(streaming);
 
   // Indicator policy: refresh on EVERY harness chunk (text, heartbeat,
   // progress). When chunks stop, the indicator naturally expires after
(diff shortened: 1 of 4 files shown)


kaieriksen (approved):
Reviewed the streaming helper extraction against the old Telegram and PhantomChat loops. The extracted segmenter options, narration controller, and suffix reconciliation preserve the existing behavior, and the new tests cover the shared pieces.

PR #251, opened 2026-07-03
feat(notify): broadcast to all authorized recipients on every channel (#249)

### LINKED ISSUE #249: notify: fan out to ALL authorized IDs on EVERY channel, not just the first owner of each
## Summary

When phantombot surfaces a notification out-of-band — from a scheduled task (`phantombot tick`) or from the threat-judge security system — it should reach **every authorized ID on every configured channel**, not a single primary recipient per channel.

Today `phantombot notify` deliberately delivers to only the **first owner** of each channel. In practice, with a Telegram-only persona, that means the notification effectively lands with **one Telegram user** and nobody else — even if other IDs are authorized, and even if other channels are configured.

## Current behaviour

`src/cli/notify.ts` resolves exactly one target per channel:

- **Telegram** — `chatId: tg.allowedUserIds[0]` (the *first* allowed user id only)
- **Phantomchat** — `recipientHex: pc.allowedHex[0]` (the *first* allowed npub only)

The module docstring states this is intentional:

> ROUTING: notify reaches the FIRST owner of EVERY configured channel … This is deliberately NOT a broadcast to every id — exactly one recipient per channel.

So `allowedUserIds[1..]` and `allowedHex[1..]` never receive task/judge notifications.

## Desired behaviour

A notification (task-fired or judge-fired) should fan out to **all** authorized recipients:

- **Every** `allowedUserIds` entry on the Telegram account (per-persona bot + default bot as applicable)
- **Every** `allowedHex` entry on the persona's phantomchat identity
- Across **every** configured channel simultaneously

i.e. everyone authorized to talk to the persona on any channel should hear about a material event, rather than a single designated primary.

## Scope / call sites

Both out-of-band surfacing paths go through `runNotify` and inherit this limitation:

- Scheduled tasks — `src/cli/tick.ts` / `src/lib/tasks.ts` → `phantombot notify`
- Threat judge — `src/lib/threatJudge.ts` held-request surfacing

## Implementation notes

- Iterate over the full `allowedUserIds` / `allowedHex` arrays instead of `[0]`.
- **Dedup recipients** — if the same recipient is authorized more than once (e.g. the same Telegram id on both the per-persona bot and the default bot, or a repeated entry), send once, not twice. Dedup per (bot, chatId) for Telegram and per recipientHex for phantomchat.
- Send per-recipient with independent error handling — one recipient failing (blocked bot, dead relay, etc.) must not abort delivery to the others.
- Voice notes stay Telegram-only (Nostr DMs are text); each Telegram recipient gets text and/or voice, phantomchat recipients get the text fallback — same as today, just fanned out.

## Error handling (2026-07-03)

Keep **exactly the same behaviour as today** — this change is fan-out only, not an error-handling change:

- A per-recipient failure is **caught and logged only** (`log.warn`, as now). Swallow it — do **not** re-throw, do **not** retry.
- **No new exits.** Do not add error exit codes. Do not kill or abort the process on a partial failure. The existing exit-code accounting stays as-is; do not make partial fan-out failures change the exit code.
- **Never surface delivery errors to users.** No stderr summary line naming failed recipients, no notification about the failed notification. Failures live in the logs and nowhere else.
- Update the exit-code accounting counters (`textSent`/`voiceSent`/`pcSent`) only insofar as needed to keep today's "did anything land at all" semantics intact — nothing more.
- Update the module docstring + `notify` command description, README, and AGENTS.md to describe broadcast-to-all-authorized semantics.
- Tests: multiple `allowedUserIds` + multiple `allowedHex` all receive; duplicate recipients deduped to one send; a mid-list failure still delivers to the rest and is logged (not surfaced, not fatal); voice-only fan-out.

## Decision (2026-07-03)

Fan-out is **unconditional** — no "primary only" mode, no config toggle. Every notification always reaches every authorized recipient on every configured channel. Do not add an opt-out.

robertclawson:
## What

`phantombot notify` now **broadcasts to every authorized recipient on every configured channel** for the persona, instead of only the first owner per channel. Closes #249.

Before: Telegram used `allowedUserIds[0]` and phantomchat used `allowedHex[0]`, so a task- or judge-fired notification on a Telegram-only persona reached exactly **one** user even when others were authorized.

After: it fans out to **all** `allowedUserIds` and **all** `allowedHex`.

## How

- **Telegram** — iterate the full `allowedUserIds`, deduped per `(bot token, chatId)` so an id repeated in the allowlist is contacted once.
- **Phantomchat** — iterate the full `allowedHex`, deduped per `recipientHex`.
- **Voice** — synthesized **once** (not per recipient) and sent to every Telegram recipient; phantomchat gets the text fallback, same as before.
- Fan-out is **unconditional** — no toggle, no primary-only mode (per the issue's decision).

## Error handling — unchanged, this is fan-out only

- Each recipient is an **independent send**; a failure is caught and `log.warn`'d, then **swallowed** — never re-thrown, never retried.
- **No new exit codes.** Partial fan-out failures (some ok, some not) still exit `0`; exit `1` only when nothing landed on any recipient/channel — the same "did anything land at all" semantics as before.
- **Never surfaced to users** — no stderr summary of failed recipients, no notification-about-the-notification. Failures live in the logs and nowhere else.

## Docs

Module docstring, `notify` command + `--persona` descriptions, the README **Notifications** section, and the AGENTS.md file-tree label all updated to describe broadcast-to-all-authorized semantics.

## Tests (`tests/cli-notify.test.ts`)

- fan-out delivers `--message` to **every** allowed id
- repeated ids / npubs **deduped** to one send each
- a **mid-list failure** still delivers to the rest, exits `0`, and stays **off stderr**
- **all recipients failing** → exit `1`
- **voice** fans out `sendVoice` to every owner
- **phantomchat** multi-npub fan-out (deduped) + resilient partial failure

Full suite green: **1817 pass / 0 fail**, `tsc --noEmit` clean.

### DIFF
--- a/src/cli/notify.ts
+++ b/src/cli/notify.ts
@@ -6,13 +6,17 @@
  * for silence as default. The harnessed agent calls `phantombot notify`
  * inside its prompt when it decides the user should hear about something.
  *
- * ROUTING: notify reaches the FIRST owner of EVERY configured channel for the
- * persona — the first Telegram allowed_user_id AND the first phantomchat
- * allowed npub. If both channels are configured, the owner gets it on BOTH, so
- * an incident is never missed because one channel is down. The "first" owner is
- * the primary; re-order the allowlist (via `phantombot telegram` /
- * `phantombot phantomchat`) to change who's primary. This is deliberately NOT a
- * broadcast to every id — exactly one recipient per channel.
+ * ROUTING: notify BROADCASTS to EVERY authorized recipient on EVERY configured
+ * channel for the persona — all Telegram allowed_user_ids AND all phantomchat
+ * allowed npubs. Everyone authorized to talk to the persona hears about a
+ * material event, not just a single designated primary. If both channels are
+ * configured, each recipient gets it on the channel(s) they're authorized on,
+ * so an incident is never missed because one channel is down. Recipients are
+ * deduped (per (bot, chatId) for Telegram, per recipientHex for phantomchat) so
+ * an id authorized twice is contacted once. Sends are per-recipient with
+ * independent error handling: one recipient failing (blocked bot, dead relay)
+ * is logged and swallowed — it never aborts delivery to the others and is never
+ * surfaced to users.
  *
  * --message  → text via sendMessage (both channels)
  * --voice    → synthesized via the configured TTS provider, sent via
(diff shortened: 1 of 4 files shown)


src/cli/notify.ts
```diff
@@ -114,29 +118,49 @@ export async function runNotify(input: RunNotifyInput = {}): Promise<number> {
...
   // A channel is a notify target when it has an account/identity AND at least
-  // one owner. We notify the FIRST owner of each (the primary). Not a broadcast.
+  // one owner. We BROADCAST to every authorized recipient of each, deduped.
 
-  // Telegram: persona-bound bot if present, else the default bot.
+  // Telegram: persona-bound bot if present, else the default bot. Fan out to
+  // every allowedUserId, deduped per (bot token, chatId) so an id repeated in
+  // the allowlist is contacted once.
   const tg: TelegramAccount | undefined =
     config.channels.telegramPersonas?.[persona] ?? config.channels.telegram;
```
kaieriksen: This still selects exactly one Telegram account: the persona bot if present, otherwise the default bot. #249 asks notify to fan out to every authorized recipient on every configured channel, and explicitly calls out the default bot plus per-persona bot case (including dedup when the same Telegram id appears in both). With this `??`, `phantombot notify --persona amanda` skips all recipients on `[channels.telegram]` whenever `[channels.telegram.personas.amanda]` exists, so authorized users on the default Telegram channel still miss task/judge notifications. Please build a target list that includes the default account and the persona account when both are configured, then dedup per the issue.

kaieriksen (changes requested):
One scope miss against #249: persona-scoped notifications still suppress the default Telegram account when a persona bot exists.

robertclawson:
Good catch — fixed in 65ded8f.

The `?? ` did indeed collapse to a single account, dropping the default `[channels.telegram]` owners whenever a persona bot existed. Replaced it with a target list that includes **both** the persona bot and the default bot when both are configured, deduped per `(bot token, chatId)`:

- an id repeated within one allowlist → contacted once (unchanged);
- a persona whose bot **is** the default bot (same token) → each `(token, chatId)` once, so no double-send;
- the same id on two genuinely different bots (distinct tokens = distinct chats) → contacted on each, since those are separate delivery channels.

Each account gets its own `HttpTelegramTransport` in production; the injected test transport is reused across accounts. Voice is still synthesized once, not per account. Updated the persona-routing test to assert the full broadcast and added a cross-account `(token, chatId)` dedup test.

robertclawson:
Good catch, @kaieriksen — you were right. The old code did `telegramPersonas[persona] ?? telegram`, which picks only **one** Telegram account, so a persona-scoped notify skipped everyone on the default bot. That's exactly the "authorized users miss notifications" gap #249 set out to close.

Fixed in `27d6192`:
- Build a target list of **both** accounts — persona bot first, then the default bot — and fan out across both.
- Dedup by `(token, chatId)` **across** accounts, so a user authorized on both bots (or a persona whose bot *is* the default bot) is contacted exactly once; genuinely distinct bots each deliver.
- Error handling unchanged per the issue: per-recipient try/catch, failures `log.warn`'d and swallowed, never surfaced, no new exits.

Also rebased onto latest `main` (was behind). Tests updated: added both-bots fan-out, shared-token dedup, and no-persona-bot fallback cases. Full suite **1825 pass / 0 fail**, `tsc` clean. Re-requesting your review.

kaieriksen (approved):
Re-reviewed the updated branch. The prior Telegram routing issue is fixed: notify now includes both the persona-bound bot and the default bot, dedups recipients, fans out phantomchat recipients, and keeps per-recipient failures isolated.\n\nValidated locally: focused notify tests, typecheck, full test suite, production build, CI green, and clean merge-tree against main.

PR #252, opened 2026-07-03
feat(chattiness): quiet-by-default whenever chattiness is unset

andrewagrahamhodges:
## What

Makes **quiet mode the standing default** for any phantom that hasn't explicitly opted into progress narration. If `chattiness` is unset, the phantom starts quiet — just the final answer, no running commentary — until an operator turns it on via `chattiness = true` or `/chattiness on`.

This holds uniformly, regardless of install state:

| Situation | default |
|---|---|
| No `config.toml` on disk | **OFF** — quiet |
| `config.toml` exists but omits `chattiness` | **OFF** — quiet |
| Empty `config.toml` | **OFF** — quiet |
| `chattiness = true` / `false` set | honored |
| `PHANTOMBOT_CHATTINESS` env | honored, wins over all |

## Why this shape

Original revision made the default *install-aware* (quiet only when no `config.toml` existed, ON otherwise). Per Andrew's follow-up, that distinction is dropped: **existing hosts with no `chattiness` key — or an empty config — should be quiet too.** So the rule collapses to the single, predictable "quiet unless explicitly set."

## How

- `DEFAULT_CHATTINESS` is now `false` and is the single fallback at both loadConfig and read sites.
- Removed `DEFAULT_CHATTINESS_UNCONFIGURED` and the `configExisted` plumbing in `tryReadToml` (that helper is back to plain `Record<string, unknown>`).
- `/chattiness` command, per-conversation overrides, and the #246 config-write path are untouched.

## Tests

- `config.test.ts`: no-file → OFF, present-but-no-key → OFF, empty → OFF, explicit true/false honored, env wins both ways.
- Narration-mechanics fixtures (`channels-telegram`, `channels-phantomchat-server`) set `chattiness: true` explicitly since they exercise the progress-bubble path.
- Full suite **1819 pass / 0 fail**, `tsc` clean.

## Operator note

A host that already sets `chattiness = true` is unaffected. A host relying on the *implicit* old ON default (config present, key absent) will go quiet after this — that's the intended change. Set `chattiness = true` to keep the bubbles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,14 +382,13 @@ errors always come through either way):
 Scoped to Telegram + PhantomChat (voice and the CLI never emit these bubbles).
 The editor (Zed/VS Code) surface follows the config default only.
 
-**The standing default depends on whether the host is configured.** A fresh /
-not-yet-configured install (no `config.toml` on disk) starts **quiet** — new
-phantoms just send the final answer until you opt in. An install that already
-has a `config.toml` keeps bubbles **on** unless you set otherwise, so existing
-users are never flipped underneath them. Set it explicitly either way:
+**The standing default is quiet.** Unless you explicitly set `chattiness`, a
+phantom starts **quiet** and just sends the final answer — this holds whether
+there's no `config.toml`, an empty one, or a `config.toml` that simply omits
+the key. Opt in to the running commentary by setting it:
 
 ```toml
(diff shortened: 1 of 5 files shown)


kaieriksen (approved):
No blocking issues found. The quiet-unless-explicitly-set behavior is consistent across loadConfig, chat resolution, env override precedence, and the updated tests.

PR #254, opened 2026-07-04
feat(vault): encrypted per-persona secrets store keyed by persona identity (#253)

### LINKED ISSUE #253: Secure secrets store: rename `phantombot env` → `phantombot vault`, encrypt at rest in SQLite keyed to the persona
## Summary

Secure the phantombot secrets store. Today `phantombot env` writes secrets as **plaintext** to `~/.env` (mode 0600). We want secrets encrypted at rest inside the persona's SQLite DB, keyed to something intrinsic to the persona, and we want the whole thing renamed `phantombot vault`.

The driving goal: **a persona directory is self-contained.** Copy the persona dir to another machine and its secrets travel with it and stay usable — because the encryption key is derived from something already living in that persona folder.

### 1. Rename `phantombot env` → `phantombot vault`
- New command surface: `phantombot vault set/get/list/unset` (mirror existing `env` subcommands 1:1).
- Keep `phantombot env` as a **deprecated alias** for one release cycle (prints a deprecation notice, forwards to `vault`) so existing scripts/tasks/docs don't break overnight. Remove later.

### 2. Encrypted storage in SQLite (no more plaintext `~/.env`)
- Secrets stored **encrypted** in the persona's SQLite database rather than a plaintext file.
- **Encryption key derived from the persona's Nostr identity key (`nsec`)** — see Design decision below.
- Values decrypted only in-process when read; never written back to disk in plaintext.

### 3. Backward-compatible auto-migration
- On startup / first vault access, if a legacy `~/.env` exists, **automatically migrate** every entry into the encrypted vault, then **delete `~/.env`**.
- The central/shared `~/.config/phantombot/.env` **also migrates** — see Central `.env` migration below.
- Migration must be idempotent and safe to run repeatedly (no-op once migrated).
- Process env loading must keep working — whatever currently sources `~/.env` into `process.env` at startup should now source from the decrypted vault instead.

### 4. Documentation — make personas aware the vault is the canonical store
- Update the **credential-guidance docs** (the agent-facing guidance that today points at `~/.env` as the canonical credentials file) so every persona/agent knows the canonical secrets store is now the **phantombot vault**, not `~/.env`.
- The guidance must state plainly: secrets are **always stored in and read from the vault**. When a persona needs to **find** a secret, it looks in the vault; when it needs to **store** a secret, it writes it to the vault (`phantombot vault set`). `~/.env` is no longer the place.
- Cover the fast-path/"where to look first" guidance, the persistence/"save what you find" guidance, and any references to `phantombot env` → update to `phantombot vault`.

## Design decision (finalized 2026-07-04)

**The vault keys off the persona's Nostr identity (`nsec`), promoted to a first-class persona identity that both phantomchat and vault depend on.**

Scan of the persona folder + scaffolder confirmed there is **no** existing value that is universal + stable + secret across all personas today:
- dir name (`robbie`) — stable but public/guessable → bad key.
- seed files (SOUL.md, memory drawers) — edited over time → not stable.
- `phantomchat.json` → `nsec` — genuinely secret (32 bytes secp256k1, mode 0600), unique, never rotated, travels with the folder — **but only exists when phantomchat is configured.**

Resolution: **make the `nsec` intrinsic to every persona, stored in a neutral `identity.json`.** Introduce a single shared `getOrCreatePersonaIdentity()` that returns the persona's `nsec`, generating it once if absent (mode 0600, inside the persona dir). Both phantomchat **and** vault call it:
- phantomchat uses it as the Nostr signing key (as today).
- vault derives its encryption key from it.

**Identity file naming:** the `nsec` lives in a dedicated **`identity.json`** (mode 0600) in the persona dir — **not** `phantomchat.json`. The identity is no longer phantomchat-specific: it's the persona's intrinsic root secret that multiple subsystems derive from. Keeping it in a neutral `identity.json` means that if phantomchat is ever deprecated or replaced by another solution, the identity — and everything keyed off it (vault, etc.) — needs **no** refactor. Migration reads any existing `nsec` from `phantomchat.json` and moves it into `identity.json`.

This gives us a key that is universal (every persona has one), stable-for-life (generated once, never rotated), secret (0600), and folder-local (travels on copy) — satisfying all three original requirements without a bolt-on `.vault.key`.

**Key generation:** the `nsec` is generated in-process via the `nostr-tools` library — `generateSecretKey()` from `nostr-tools/pure` (secure random secp256k1 key) wrapped to bech32 `nsec` via `nsecEncode`. No external tool/shell-out — same path phantomchat already uses today.

**Key separation (hygiene):** vault does **not** use the raw `nsec` as the AES key. It runs the `nsec` through **HKDF-SHA256 with a domain-separation label** (e.g. `"phantombot-vault-v1"`) to produce a distinct symmetric key, then **AES-256-GCM per entry with a per-entry random nonce**. Same secret root, but leaking the vault key does not expose the Nostr signing key and vice versa.

### Central `.env` migration (finalized 2026-07-04)

The shared central store `~/.config/phantombot/.env` (sourced at startup alongside `~/.env`) holds credentials that are **currently shared across every persona in the account**. Because the new vault is keyed per-persona (each persona's `nsec` encrypts its own DB), there is no longer a shared plaintext file to fall back on. So migration must **fan the central `.env` values out into every persona's vault** — each central key is copied into every persona's encrypted store — so the personas keep the secrets they rely on today.

- On migration, enumerate every persona in the account and write each central `.env` entry into **each** persona's vault (encrypted under that persona's own key).
- Per-persona `~/.env` entries migrate into that persona's vault only (not fanned out).
- On a per-key collision (a persona already has a key of the same name from its own `~/.env`), the **persona-local value wins** over the shared central value.
- After a verified write into all personas, **delete the central `~/.config/phantombot/.env`**.
- Idempotent: once migrated, a re-run is a no-op.

Note: fan-out means each persona ends up with its **own independent copy** — later edits to one persona's vault entry do **not** propagate to the others. This is the intended trade for self-contained, portable persona folders.

### Resolved open questions
1. ~~Which persona-folder value keys the vault?~~ → the `nsec`, generated on demand if not present. **Resolved.**
2. ~~Cipher/KDF~~ → HKDF-SHA256 (label `phantombot-vault-v1`) from the `nsec` → AES-256-GCM, per-entry random nonce. **Resolved.**
3. ~~Does the central `~/.config/phantombot/.env` migrate too?~~ → **Yes** — fanned out into every persona's vault (see above). **Resolved.**
4. ~~Where is the `nsec` stored?~~ → a neutral **`identity.json`** (mode 0600) in the persona dir, not `phantomchat.json`; migrated across if it currently lives in `phantomchat.json`. **Resolved.**
5. ~~Clean `.env` delete vs. a one-time `.bak`?~~ → **Clean delete, gated on verified migration.** No backup, no plaintext left behind — but the delete only fires after migration is *proven* (see Migration validation below). **Resolved.**
6. ~~How is the `nsec` generated when absent?~~ → in-process via `nostr-tools` (`generateSecretKey` + `nsecEncode`), no shell-out. **Resolved.**

### Migration validation (finalized 2026-07-04)

The plaintext `.env` files are deleted **clean** (no `.bak`), but **only after every migrated value is verified present and correct in the vault**. Deletion is gated on this check:

- For each key migrated, after the encrypted write, **read it back through the vault (decrypt) and assert the decrypted value byte-for-byte equals the source value** from the `.env`. For the central-`.env` fan-out, verify the read-back in **every** persona it was written to.
- Only once **all** keys from a given source file pass read-back verification is that source file deleted.
- If **any** key fails verification, **abort the delete**, leave the plaintext file in place, and log the failure (log-only, per the standing rule — no user-facing error), so no secret is ever lost to an unverified migration.
- This keeps migration idempotent: a re-run re-verifies and only then cleans up.

## Non-goals / constraints
- No change to how callers *reference* secrets (still `$NAME` in the process env at runtime).
- Don't surface secret values to logs or scrollback.

Raised on behalf of Andrew.

robertclawson:
Implements #253 — replaces the plaintext `~/.env` credential store with an encrypted, per-persona vault so persona folders are self-contained and portable (copy a persona dir to another box and its secrets travel with it, still encrypted).

## What changed
1. **`phantombot vault`** (set/get/list/unset) is the new canonical command. `phantombot env` is kept as a **deprecated forwarding alias** for one release (prints a deprecation note, forwards to the vault) so existing scripts/tasks don't break.
2. **Encrypted at rest** — secrets live in `<personaDir>/vault.sqlite`, AES-256-GCM per entry with a random 12-byte nonce. The AES key is derived via **HKDF-SHA256** (label `phantombot-vault-v1`) from the persona's nsec secret bytes — never the raw nsec, so a vault-key leak can't compromise the Nostr signing key.
3. **Persona identity** — the nsec is promoted to a first-class identity in a neutral **`identity.json`** (mode 0600), generated on demand via `nostr-tools`. Both vault and phantomchat derive from it; phantomchat sources its nsec from `identity.json`, falling back to legacy `phantomchat.json` and migrating it across.
4. **Auto-migration** — on startup, legacy `~/.env` migrates into the active persona; central `~/.config/phantombot/.env` **fans out into every persona** (shared secrets preserved). Persona-local values win on collision. Every value is **read back, decrypted, and byte-compared** before its plaintext file is deleted; on any mismatch the delete aborts and plaintext is left in place.
5. **Loader** injects the decrypted vault into `process.env` at startup and per turn, same "existing env wins" semantics as before — nothing changes for the personas, they still read `process.env`.
6. **Docs** — credential-guidance section updated so every persona knows the vault is the canonical store for reading and writing secrets.

## Verification
- `tsc --noEmit` clean; full suite **1827 pass / 0 fail**.
- Live end-to-end against the built binary: set/get/list/unset round-trip; secret confirmed **not present in plaintext** in the DB; `~/.env` + central `.env` both migrated then deleted; collision precedence (persona-local wins); central secret fanned out to **both** lena and kai; deprecated `env` alias forwards correctly.

Dogfood plan: **Lena and Kai** first.

Closes #253

### DIFF
--- a/src/channels/phantomchat/personaStore.ts
+++ b/src/channels/phantomchat/personaStore.ts
@@ -66,6 +66,7 @@ import {
   identityFromNsec,
   type NostrIdentity,
 } from "../../lib/nostrIdentity.ts";
+import { readPersonaIdentityNsec } from "../../lib/personaIdentity.ts";
 
 /** Filename of the per-persona phantomchat config inside an agent dir. */
 export const PHANTOMCHAT_FILE = "phantomchat.json";
@@ -193,14 +194,25 @@ export function loadPhantomchatPersonaConfig(
     });
     return undefined;
   }
-  if (!parsed || typeof parsed.nsec !== "string" || parsed.nsec.trim() === "") {
+  // The nsec now lives in the shared `identity.json` (lib/personaIdentity.ts).
+  // Prefer it; fall back to a legacy nsec still inside phantomchat.json so
+  // pre-migration persona folders keep working. The channel is enabled as long
+  // as EITHER source yields a usable nsec — phantomchat.json still carries the
+  // channel config (relays / allowlist / group bots).
+  const sharedNsec = readPersonaIdentityNsec(agentDir);
+  const nsec =
+    sharedNsec ??
+    (typeof parsed.nsec === "string" && parsed.nsec.trim() !== ""
+      ? parsed.nsec.trim()
+      : undefined);
+  if (!nsec) {
     return undefined;
   }
   let identity: NostrIdentity;
   try {
-    identity = identityFromNsec(parsed.nsec);
+    identity = identityFromNsec(nsec);
   } catch (e) {
-    log.warn(`phantomchat: invalid nsec in ${path} — skipping`, {
+    log.warn(`phantomchat: invalid nsec for ${agentDir} — skipping`, {
       error: (e as Error).message,
     });
     return undefined;
--- a/src/channels/phantomchat/personaStore.ts
+++ b/src/channels/phantomchat/personaStore.ts
@@ -66,7 +66,11 @@ import {
   identityFromNsec,
   type NostrIdentity,
 } from "../../lib/nostrIdentity.ts";
-import { readPersonaIdentityNsec } from "../../lib/personaIdentity.ts";
+import {
+  personaIdentityPath,
+  readPersonaIdentityNsec,
+  writePersonaIdentity,
+} from "../../lib/personaIdentity.ts";
 
 /** Filename of the per-persona phantomchat config inside an agent dir. */
 export const PHANTOMCHAT_FILE = "phantomchat.json";
@@ -234,9 +238,17 @@ export function loadPhantomchatPersonaConfig(
 }
 
 /**
- * Atomically write a persona's phantomchat.json at mode 0600 (the nsec is a
- * secret). Creates the agent dir if needed. Tempfile + rename avoids the
- * world-readable window a write-then-chmod would leave.
+ * Atomically write a persona's phantomchat.json at mode 0600. Creates the agent
+ * dir if needed. Tempfile + rename avoids the world-readable window a
+ * write-then-chmod would leave.
+ *
+ * The persona's root secret (nsec) is NO LONGER written here — it lives solely
+ * in the shared `identity.json` (lib/personaIdentity.ts), which both this
+ * channel and the encrypted vault derive from. `data.nsec` is still accepted so
+ * callers (and legacy files whose nsec we just read back) can seed identity.json
+ * on the first save that follows a migration; it is persisted there
+ * (create-if-absent, never clobbering an existing identity) and then kept OUT of
+ * phantomchat.json, which now carries channel settings only.
  */
 export async function savePhantomchatPersonaConfig(
   agentDir: string,
@@ -251,8 +263,14 @@ export async function savePhantomchatPersonaConfig(
 ): Promise<string> {
   const path = phantomchatConfigPath(agentDir);
   await mkdir(dirname(path), { recursive: true });
+  // Ensure the root secret lives in identity.json before we drop it from this
+  // file. Create-if-absent: a legacy persona's nsec (read out of the old
+  // phantomchat.json) is promoted here exactly once; an identity.json that
+  // already exists (the common case) is left untouched so we never clobber it.
+  if (data.nsec && !existsSync(personaIdentityPath(agentDir))) {
+    await writePersonaIdentity(agentDir, data.nsec);
+  }
   const body: PhantomchatFileShape = {
-    nsec: data.nsec,
     relays: data.relays,
     allowed_npubs: data.allowedNpubs,
   };
--- a/src/channels/phantomchat/personaStore.ts
+++ b/src/channels/phantomchat/personaStore.ts
@@ -67,9 +67,8 @@ import {
   type NostrIdentity,
 } from "../../lib/nostrIdentity.ts";
 import {
-  personaIdentityPath,
+  createPersonaIdentityIfAbsent,
   readPersonaIdentityNsec,
-  writePersonaIdentity,
 } from "../../lib/personaIdentity.ts";
 
 /** Filename of the per-persona phantomchat config inside an agent dir. */
@@ -263,12 +262,14 @@ export async function savePhantomchatPersonaConfig(
 ): Promise<string> {
   const path = phantomchatConfigPath(agentDir);
   await mkdir(dirname(path), { recursive: true });
-  // Ensure the root secret lives in identity.json before we drop it from this
-  // file. Create-if-absent: a legacy persona's nsec (read out of the old
-  // phantomchat.json) is promoted here exactly once; an identity.json that
-  // already exists (the common case) is left untouched so we never clobber it.
-  if (data.nsec && !existsSync(personaIdentityPath(agentDir))) {
-    await writePersonaIdentity(agentDir, data.nsec);
+  // Ensure the root secret is durably in identity.json before we drop it from
+  // this channel file. Atomic create-if-absent (no existsSync check-then-write):
+  // a legacy persona's nsec is promoted exactly once; an identity.json that
+  // already exists — including one the vault just minted concurrently — is never
+  // overwritten, so we can't clobber the canonical identity and orphan its
+  // encrypted rows.
+  if (data.nsec) {
+    await createPersonaIdentityIfAbsent(agentDir, data.nsec);
   }
   const body: PhantomchatFileShape = {
     relays: data.relays,
--- a/src/cli/phantomchat.ts
+++ b/src/cli/phantomchat.ts
@@ -163,8 +163,10 @@ export async function runPhantomchat(input: RunInput = {}): Promise<number> {
         `Generated a new Nostr keypair for '${persona}'. The secret (nsec) will be\n` +
           `saved to <persona-dir>/identity.json (mode 0600). Back it up — this\n` +
           `nsec now also derives the persona's VAULT encryption key, so losing it\n` +
-          `means a new identity (re-add the new npub in the app) AND every secret\n` +
-          `stored in the vault becomes permanently unrecoverable.\n\n` +
+          `means a new identity (re-add the new npub in the app) AND the existing\n` +
+          `vault ciphertext can no longer be decrypted. This is a reconfigure, not\n` +
+          `a catastrophe: if you lose it, mint a fresh identity, re-add your secrets\n` +
+          `with 'phantombot vault set', and re-pair PhantomChat with the new npub.\n\n` +
           `  nsec (one-time display): ${nsec}\n\n` +
           `Its npub (paste this into the PhantomChat app to DM '${persona}'):\n\n` +
           `  ${npub}`,
--- a/src/cli/phantomchat.ts
+++ b/src/cli/phantomchat.ts
@@ -40,8 +40,12 @@ import {
 import {
   decodeNpubToHex,
   generateIdentity,
+  identityFromNsec,
 } from "../lib/nostrIdentity.ts";
-import { writePersonaIdentity } from "../lib/personaIdentity.ts";
+import {
+  readPersonaIdentityNsec,
+  writePersonaIdentity,
+} from "../lib/personaIdentity.ts";
 import { fetchCanonicalRelays } from "../channels/phantomchat/relaysSource.ts";
 import { defaultServiceControl, type ServiceControl } from "../lib/platform.ts";
 import type { WriteSink } from "../lib/io.ts";
@@ -129,21 +133,38 @@ export async function runPhantomchat(input: RunInput = {}): Promise<number> {
       "Existing identity",
     );
   } else {
-    const identity = generate();
-    nsec = identity.nsec;
-    npub = identity.npub;
-    // The nsec is the persona's SHARED identity (used by the vault too), so it
-    // lives in <persona-dir>/identity.json (mode 0600), not phantomchat.json.
-    await writePersonaIdentity(agentDir, nsec);
-    p.note(
-      `Generated a new Nostr keypair for '${persona}'. The secret (nsec) will be\n` +
-        `saved to <persona-dir>/identity.json (mode 0600). Back it up — losing\n` +
-        `it means a new identity (and re-adding the new npub in the app).\n\n` +
-        `  nsec (one-time display): ${identity.nsec}\n\n` +
-        `Its npub (paste this into the PhantomChat app to DM '${persona}'):\n\n` +
-        `  ${npub}`,
-      "New identity created",
-    );
+    // No phantomchat.json yet — but identity.json may already exist (e.g. the
+    // vault minted it first). ADOPT an existing identity rather than generating
+    // a fresh nsec, which would orphan everything the vault already encrypted
+    // under the old key. Only mint when there is genuinely no identity yet.
+    const adopted = readPersonaIdentityNsec(agentDir);
+    if (adopted) {
+      const identity = identityFromNsec(adopted);
+      nsec = identity.nsec;
+      npub = identity.npub;
+      p.note(
+        `Persona '${persona}' already has a Nostr identity (identity.json). Reusing it.\n\n` +
+          `Its npub (paste this into the PhantomChat app to DM '${persona}'):\n\n` +
+          `  ${npub}`,
+        "Existing identity",
+      );
+    } else {
+      const identity = generate();
+      nsec = identity.nsec;
+      npub = identity.npub;
+      // The nsec is the persona's SHARED identity (used by the vault too), so it
+      // lives in <persona-dir>/identity.json (mode 0600), not phantomchat.json.
+      await writePersonaIdentity(agentDir, nsec);
+      p.note(
+        `Generated a new Nostr keypair for '${persona}'. The secret (nsec) will be\n` +
+          `saved to <persona-dir>/identity.json (mode 0600). Back it up — losing\n` +
+          `it means a new identity (and re-adding the new npub in the app).\n\n` +
+          `  nsec (one-time display): ${identity.nsec}\n\n` +
+          `Its npub (paste this into the PhantomChat app to DM '${persona}'):\n\n` +
+          `  ${npub}`,
+        "New identity created",
+      );
+    }
   }
 
   // 2. Relays are NOT prompted any more — they come from the canonical
--- a/src/cli/phantomchat.ts
+++ b/src/cli/phantomchat.ts
@@ -43,8 +43,8 @@ import {
   identityFromNsec,
 } from "../lib/nostrIdentity.ts";
 import {
+  createPersonaIdentityIfAbsent,
   readPersonaIdentityNsec,
-  writePersonaIdentity,
 } from "../lib/personaIdentity.ts";
 import { fetchCanonicalRelays } from "../channels/phantomchat/relaysSource.ts";
 import { defaultServiceControl, type ServiceControl } from "../lib/platform.ts";
@@ -150,16 +150,20 @@ export async function runPhantomchat(input: RunInput = {}): Promise<number> {
       );
     } else {
       const identity = generate();
-      nsec = identity.nsec;
-      npub = identity.npub;
       // The nsec is the persona's SHARED identity (used by the vault too), so it
       // lives in <persona-dir>/identity.json (mode 0600), not phantomchat.json.
-      await writePersonaIdentity(agentDir, nsec);
+      // Atomic create-if-absent: if the vault minted one between the read above
+      // and here, we adopt its identity rather than overwriting it — and display
+      // whatever is durably on disk so the npub we show is the real one.
+      const persistedNsec = await createPersonaIdentityIfAbsent(agentDir, identity.nsec);
+      const persisted = identityFromNsec(persistedNsec);
+      nsec = persisted.nsec;
+      npub = persisted.npub;
       p.note(
         `Generated a new Nostr keypair for '${persona}'. The secret (nsec) will be\n` +
           `saved to <persona-dir>/identity.json (mode 0600). Back it up — losing\n` +
           `it means a new identity (and re-adding the new npub in the app).\n\n` +
-          `  nsec (one-time display): ${identity.nsec}\n\n` +
+          `  nsec (one-time display): ${nsec}\n\n` +
           `Its npub (paste this into the PhantomChat app to DM '${persona}'):\n\n` +
           `  ${npub}`,
         "New identity created",
--- a/src/cli/phantomchat.ts
+++ b/src/cli/phantomchat.ts
@@ -161,8 +161,10 @@ export async function runPhantomchat(input: RunInput = {}): Promise<number> {
       npub = persisted.npub;
       p.note(
         `Generated a new Nostr keypair for '${persona}'. The secret (nsec) will be\n` +
-          `saved to <persona-dir>/identity.json (mode 0600). Back it up — losing\n` +
-          `it means a new identity (and re-adding the new npub in the app).\n\n` +
+          `saved to <persona-dir>/identity.json (mode 0600). Back it up — this\n` +
+          `nsec now also derives the persona's VAULT encryption key, so losing it\n` +
+          `means a new identity (re-add the new npub in the app) AND every secret\n` +
+          `stored in the vault becomes permanently unrecoverable.\n\n` +
           `  nsec (one-time display): ${nsec}\n\n` +
           `Its npub (paste this into the PhantomChat app to DM '${persona}'):\n\n` +
           `  ${npub}`,
--- a/src/cli/phantomchat.ts
+++ b/src/cli/phantomchat.ts
@@ -41,6 +41,7 @@ import {
   decodeNpubToHex,
   generateIdentity,
 } from "../lib/nostrIdentity.ts";
+import { writePersonaIdentity } from "../lib/personaIdentity.ts";
 import { fetchCanonicalRelays } from "../channels/phantomchat/relaysSource.ts";
 import { defaultServiceControl, type ServiceControl } from "../lib/platform.ts";
 import type { WriteSink } from "../lib/io.ts";
@@ -131,9 +132,12 @@ export async function runPhantomchat(input: RunInput = {}): Promise<number> {
     const identity = generate();
     nsec = identity.nsec;
     npub = identity.npub;
+    // The nsec is the persona's SHARED identity (used by the vault too), so it
+    // lives in <persona-dir>/identity.json (mode 0600), not phantomchat.json.
+    await writePersonaIdentity(agentDir, nsec);
     p.note(
       `Generated a new Nostr keypair for '${persona}'. The secret (nsec) will be\n` +
-        `saved to <persona-dir>/phantomchat.json (mode 0600). Back it up — losing\n` +
+        `saved to <persona-dir>/identity.json (mode 0600). Back it up — losing\n` +
         `it means a new identity (and re-adding the new npub in the app).\n\n` +
         `  nsec (one-time display): ${identity.nsec}\n\n` +
         `Its npub (paste this into the PhantomChat app to DM '${persona}'):\n\n` +
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,17 +5,42 @@
  * Imports the Citty dispatcher and runs it. The dispatcher itself lives in
  * src/cli/index.ts so it can be imported by tests without auto-running.
  *
- * Before dispatch, we self-load credentials from `~/.env` and
- * `~/.config/phantombot/.env` into process.env. On Linux the systemd unit
- * already sources these via `EnvironmentFile=`, so the lines we'd add are
- * no-ops (existing values win — see preloadEnvFiles). On macOS launchd
- * has no equivalent of EnvironmentFile, so this self-load is the only
- * way credentials reach the running agent.
+ * Before dispatch we bootstrap credentials, in order:
+ *
+ *   1. `migratePlaintextToVault` — an idempotent, best-effort migration of any
+ *      leftover plaintext `~/.env` / `~/.config/phantombot/.env` into the
+ *      per-persona ENCRYPTED vaults, deleting each plaintext file only after
+ *      every key read back byte-for-byte. A no-op once the files are gone.
+ *   2. `loadVaultIntoEnv` — decrypt the ACTIVE persona's vault and inject its
+ *      secrets into process.env, with the same "existing value wins" policy the
+ *      old plaintext loader used (a shell export / systemd EnvironmentFile= key
+ *      is never overwritten). This is the vault replacement for the old
+ *      plaintext self-source.
+ *   3. `preloadEnvFiles` — retained for the transitional path: harness.ts still
+ *      writes the Pi routing key to `~/.env` mid-session, and the harnesses
+ *      re-source it before each spawn; the NEXT startup migration folds it into
+ *      the vault. Once no plaintext files remain this is a cheap no-op.
+ *
+ * Wrapped so a bootstrap hiccup never blocks the CLI from running.
  */
 
 import { runMain } from "citty";
 import { mainCommand } from "./cli/index.ts";
+import { loadConfig, personaDir } from "./config.ts";
 import { preloadEnvFiles } from "./lib/envBootstrap.ts";
+import { log } from "./lib/logger.ts";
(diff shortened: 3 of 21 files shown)


src/index.ts
```diff
@@ -5,17 +5,42 @@
...
 import { preloadEnvFiles } from "./lib/envBootstrap.ts";
+import { log } from "./lib/logger.ts";
+import { loadVaultIntoEnv } from "./lib/vault.ts";
+import { migratePlaintextToVault } from "./lib/vaultMigrate.ts";
 
+try {
+  const config = await loadConfig();
+  await migratePlaintextToVault(config);
+  const activePersona = process.env.PHANTOMBOT_PERSONA || config.defaultPersona;
+  await loadVaultIntoEnv(personaDir(config, activePersona));
```
kaieriksen, line 39: Loading the vault only during CLI startup drops the previous reload-on-spawn guarantee. After an agent runs `phantombot vault set FOO ...` (or the deprecated `env set` alias), that command runs inside the harness subprocess and writes `<persona>/vault.sqlite`; it does not mutate the long-lived daemon's `process.env`. The harnesses still only call `reloadEnvFiles()` before spawning, so the next turn will not receive `FOO` unless the daemon restarts. The old `env set` path became visible next turn because `reloadEnvFiles()` reread `~/.env`; this needs an equivalent per-persona vault reload before each harness spawn, preserving the existing-env/sticky semantics.

src/cli/phantomchat.ts
```diff
@@ -131,9 +132,12 @@ export async function runPhantomchat(input: RunInput = {}): Promise<number> {
     const identity = generate();
     nsec = identity.nsec;
     npub = identity.npub;
+    // The nsec is the persona's SHARED identity (used by the vault too), so it
+    // lives in <persona-dir>/identity.json (mode 0600), not phantomchat.json.
+    await writePersonaIdentity(agentDir, nsec);
```
kaieriksen: This creates `identity.json`, but the flow still passes `nsec` into `savePersonaConfig` below, and `savePhantomchatPersonaConfig` still writes that value into `phantomchat.json`. New phantomchat configs therefore continue to duplicate the root secret in the channel-specific file, and later relay/TOFU/greet writes preserve it there. #253 explicitly moves the persona identity out of `phantomchat.json` into `identity.json`; `phantomchat.json` should keep channel settings only, with legacy fallback read support for old files.

kaieriksen (changes requested):
Thanks. I found two blockers: newly saved vault secrets are not visible to the next turn without a daemon restart, and phantomchat still writes the persona nsec into phantomchat.json after introducing identity.json.

src/index.ts
robertclawson: Fixed in 3cc1dd8. Added `reloadVaultForPersona(req.persona)` in all four harnesses right after `reloadEnvFiles()`. `loadVaultIntoEnv` now reconciles against a module-scope tracked set (mirroring envBootstrap's sticky-vs-reloadable rules): a key we injected from a vault is updated to this persona's value or removed if this persona lacks it, while keys already in the env at boot (shell/systemd) stay sticky. So a `vault set` from the previous turn is visible next turn, and each persona only ever sees its own secrets. Covered by tests/vault.test.ts (reconcile: new-key load, sticky-wins, persona-switch strip, save-visible-next, fail-closed).

src/cli/phantomchat.ts
robertclawson: Fixed in 3cc1dd8. The single-point fix is in `savePhantomchatPersonaConfig`: it no longer writes `nsec` into phantomchat.json at all — it persists the root secret to identity.json (create-if-absent, never clobbering) and writes phantomchat.json as channel-config only. That covers every writer (cacheRelays/recordTrusted/recordGreeted too). Legacy inline-nsec files still load (fallback read) and get promoted to identity.json on their next save. The CLI generate path also now adopts an existing identity.json instead of minting a fresh nsec. Pinned by two new tests in channels-phantomchat-personaStore.test.ts.

robertclawson:
## Review round 2 — pushed `3cc1dd8`

Both of your blockers plus 7 items from my own pass are fixed. `tsc --noEmit` clean; **1858 pass / 0 fail** (+31 new tests).

**Your two blockers**
- **Per-turn vault reload** — `reloadVaultForPersona()` in all four harnesses after `reloadEnvFiles()`, backed by a tracked-set reconcile in `loadVaultIntoEnv` (same sticky/existing-wins semantics as envBootstrap). Fresh `vault set` visible next turn; no cross-persona secret leak.
- **phantomchat.json identity split** — `nsec` now lives only in `identity.json`; `savePhantomchatPersonaConfig` keeps the channel file config-only and promotes any legacy inline nsec on next save (legacy read preserved).

**My own findings (also fixed)**
- Durability + portability: vault DB is `journal_mode=DELETE` + `synchronous=FULL` — ciphertext is fsync-durable before migration unlinks the plaintext, and no `-wal` sidecar breaks the single-file portability promise.
- Identity create race: `createIdentityFileExclusive` hard-links into place (EEXIST → adopt winner) so concurrent first-use cannot mint divergent nsecs and orphan a vault.
- Migration hardening: hidden/non-persona dirs skipped in the central fan-out; "local wins" only recorded when the `~/.env` migration verified and removed the file; `openPersonaVault` wrapped so migration is truly best-effort (never throws).
- Tests: new `vault`, `persona-identity` (incl. concurrency race), `vault-migrate` suites + phantomchat identity-split assertions — the crypto/migration paths that were previously untested are now covered.

**One known-minor** (documented, not blocking): the interactive `phantombot harness` wizard still writes the Pi key to `~/.env`; post-migration that value can be shadowed by the vault copy until the next daemon restart (which re-migrates it). Self-heals on restart; the proper fix (harness writes straight to the vault) is out of scope here.

Ready for re-review, @kaieriksen.

src/channels/phantomchat/personaStore.ts
```diff
@@ -239,8 +263,14 @@ export async function savePhantomchatPersonaConfig(
 ): Promise<string> {
   const path = phantomchatConfigPath(agentDir);
   await mkdir(dirname(path), { recursive: true });
+  // Ensure the root secret lives in identity.json before we drop it from this
+  // file. Create-if-absent: a legacy persona's nsec (read out of the old
+  // phantomchat.json) is promoted here exactly once; an identity.json that
+  // already exists (the common case) is left untouched so we never clobber it.
+  if (data.nsec && !existsSync(personaIdentityPath(agentDir))) {
+    await writePersonaIdentity(agentDir, data.nsec);
```
kaieriksen: This still races because writePersonaIdentity() unconditionally renames over identity.json after a separate existsSync check. If vault first-use calls getOrCreatePersonaIdentity() between the check and this write, the vault can encrypt secrets under the newly-created identity and then this save overwrites identity.json with the legacy/channel nsec, making those vault rows undecryptable. The same invariant fixed in getOrCreatePersonaIdentity() needs to apply here too: create-if-absent must be atomic, with losers adopting the existing identity instead of clobbering it.

kaieriksen (changes requested):
Thanks for the fixes. The per-turn vault reload and phantomchat.json split are much better, but there is still one identity race that can orphan an encrypted vault.

src/channels/phantomchat/personaStore.ts
kaieriksen: This still has the same check-then-write race. `existsSync()` and `writePersonaIdentity()` are separate operations, and `writePersonaIdentity()` uses rename-over-target semantics. If vault first-use creates `identity.json` between line 270 and this write, this path can overwrite the canonical identity with the legacy/channel nsec after the vault has already encrypted rows under the first identity. The create-if-absent helper needs to be atomic here too, not just in `getOrCreatePersonaIdentity()`.

src/lib/personaIdentity.ts
```diff
@@ -0,0 +1,201 @@
...
+      mode: 0o600,
+    });
+    try {
+      await link(tmp, path);
+      return nsec; // we won the race
+    } catch (e) {
+      if ((e as NodeJS.ErrnoException).code === "EEXIST") {
+        // Lost the race — adopt the winner's identity, discard ours.
+        const winner = readNsecFromJson(path);
+        if (winner) return winner;
```
kaieriksen, line 111: This fallback returns an nsec that was not persisted anywhere. If `identity.json` already exists but is malformed/unusable, the initial read treats it as absent, `link()` then fails with EEXIST, `winner` is undefined, and callers get a fresh generated identity for this one process only. A `vault set` can then encrypt data under that transient key, and the next call generates a different transient key and cannot decrypt it. This should fail closed or otherwise ensure the returned identity is the one durably stored in `identity.json`.

kaieriksen (changes requested):
The current head still needs changes. The prior identity overwrite race is still present in `savePhantomchatPersonaConfig`, and I found one related identity persistence edge case that can also orphan vault data.

robertclawson:
Thanks @kaieriksen — both were real. Fixed in eee5826.

**1. Check-then-write race in `savePhantomchatPersonaConfig`.** You're right that `existsSync()` + `writePersonaIdentity()` (unconditional rename) could clobber a canonical identity the vault minted between the check and the write. Removed the check entirely — it now calls the atomic create-if-absent primitive (`link()`-based, EEXIST = lose the race and adopt the incumbent), so an existing `identity.json` is never overwritten regardless of interleaving.

**2. Transient-identity fallback.** Also fixed. The EEXIST branch used to return the caller's in-process `nsec` when the incumbent file was present-but-unreadable — an identity that never reached disk. It now **fails closed** (throws) in that case, so we never hand back a key the next process can't reproduce and can't silently encrypt vault rows under it.

Also removed the `writePersonaIdentity` rename-overwrite helper altogether so no future caller can reintroduce the same race — the phantomchat setup CLI now goes through the same create-if-absent path and displays whatever is durably on disk. Renamed the primitive to the exported `createPersonaIdentityIfAbsent`.

Added tests for the no-overwrite path and the fail-closed-on-unreadable path. Full suite green (1860/1860). Ready for another look.

lenaparkhodges (changes requested):
## Review: feat(vault): encrypted per-persona secrets store

Overall this is excellent work — the crypto design (HKDF-derived key, AES-256-GCM, per-persona isolation), the migration safety (read-back before delete), the race-safe identity creation (`createIdentityFileExclusive`), and the per-turn vault reconcile are all well-architected. The test coverage is thorough.

I found one issue that can break non-default personas in production, and two more that should be addressed before merge.

---

### 🔴 Breaking: `~/.env` migrates to ONLY the default persona — non-default personas silently lose credentials

In the old system, `~/.env` was loaded globally via `preloadEnvFiles` / systemd `EnvironmentFile=`, so **every persona** had access to keys like `GITHUB_TOKEN`, `OPENAI_API_KEY`, etc. After this PR, `~/.env` migrates into **only the default persona's vault** (`migrateUserEnv` takes `defaultPersona`), and the plaintext file is deleted. Non-default personas (e.g. Kai) that rely on those credentials will silently lose access on the next turn — `reloadVaultForPersona` loads only the active persona's vault, and the plaintext file is gone.

The central `~/.config/phantombot/.env` is correctly fanned out to all personas, but `~/.env` is not. For a single-operator multi-persona setup (which is exactly the dogfood plan — "Lena and Kai first"), this breaks Kai.

**Fix options:**
1. Fan out `~/.env` to all personas (same as central `.env`), with persona-local values winning on collision. This preserves the old global behavior.
2. Or at minimum, document this as a breaking change and provide a migration path (e.g. `phantombot vault set --persona kai GITHUB_TOKEN ...`).

I'd strongly recommend option 1 — the distinction between "personal credentials" (`~/.env`) and "shared secrets" (central `.env`) doesn't hold for a single operator running multiple personas on one box.

There's also no test covering this scenario — a test asserting that a non-default persona gets a `~/.env` key (or documenting that it doesn't) would clarify the intent.

---

### 🟡 `writePersonaIdentity` uses `rename` (overwrite) while `createIdentityFileExclusive` uses `link` (exclusive) — inconsistent for the same file

The vault path correctly uses `createIdentityFileExclusive` (hard-link, EEXIST → adopt winner) to prevent divergent nsecs. But two other write paths use `writePersonaIdentity` (tempfile + `rename`, which silently overwrites):

1. **`phantomchat` CLI fresh-generate** (`src/cli/phantomchat.ts`): generates a new nsec and calls `writePersonaIdentity`. In the normal flow, startup already created `identity.json` via the vault, so the CLI adopts it. But if the startup bootstrap failed (caught by try/catch), the CLI is first to write — and a concurrent vault access could have its exclusively-created `identity.json` clobbered by the CLI's `rename`, orphaning the vault.

2. **`savePhantomchatPersonaConfig`** (`personaStore.ts`): guards with `!existsSync(identityPath)` then calls `writePersonaIdentity` — a TOCTOU. If a vault access creates `identity.json` between the check and the rename, the vault's nsec is overwritten. (In practice the values are usually the same — both sourced from legacy `phantomchat.json` — so the blast radius is narrow. But for the fresh-generate case, the nsecs would differ.)

**Fix:** Use `createIdentityFileExclusive` in these paths too. It already returns the persisted nsec (which may differ from the input if another process won the race), so callers can use the returned value instead of their generated one. This requires exporting the function.

---

### 🟡 Losing `identity.json` now means losing ALL stored secrets — not documented

Before this PR, losing the nsec meant losing the Nostr identity. After this PR, the nsec derives the vault encryption key — so losing `identity.json` means **all stored secrets are permanently unrecoverable**. This is a significant escalation of blast radius.

The `phantomchat` CLI setup message says:
> "Back it up — losing it means a new identity"

This should be updated to something like:
> "Back it up — losing it means a new identity AND all secrets stored in the vault become permanently unrecoverable"

The credential guidance section in the persona builder should also mention this.

---

### ✅ Things that are solid
- HKDF key derivation (never raw nsec, domain-separated) — correct
- AES-256-GCM with per-entry random nonce, wrong-key rejection via auth tag — correct
- `journal_mode=DELETE` + `synchronous=FULL` for durability and portability — correct
- Migration read-back verification before `unlink` — correct
- Per-turn vault reconcile with fail-closed cross-persona stripping — correct
- Test coverage (crypto, isolation, portability, migration, identity race, persona store) — thorough
- `@noble/hashes` already a direct dependency — no transitive dep issue
- `phantombot env` deprecation alias forwarding — clean, backward-compatible

src/channels/phantomchat/personaStore.ts
robertclawson: Fixed in eee5826. Dropped the existsSync+writePersonaIdentity TOCTOU entirely — this path now calls the atomic `createPersonaIdentityIfAbsent(agentDir, data.nsec)`, the same hard-link create-if-absent primitive as `getOrCreatePersonaIdentity`. If the vault created identity.json first, the link fails EEXIST and we adopt the winner rather than clobbering, so vault rows can never be orphaned. `writePersonaIdentity` (the rename-over-target footgun) is removed from the tree.

src/lib/personaIdentity.ts
robertclawson: Fixed in eee5826. `createPersonaIdentityIfAbsent` now fails CLOSED: if identity.json exists but its nsec is unreadable (malformed/truncated), the EEXIST branch throws instead of returning the caller's in-process nsec — so a transient key can never encrypt vault rows the next process can't reproduce. Covered by a new fail-closed test in tests/persona-identity.test.ts.

robertclawson:
@lenaparkhodges thanks — the breaking finding was spot on. Addressed in 2770fdb:

**L1 (🔴 breaking) — `~/.env` fan-out.** Took your recommended option 1. `~/.env` now fans out into **every** persona's vault (mirroring the old global `EnvironmentFile=` / `preloadEnvFiles` behaviour), with read-back verified in each persona before the single plaintext copy is deleted. `~/.env` still wins over central `.env` on collision — now in every persona, not just the default. Two regression tests added: non-default personas (Lena/Kai) receive a `~/.env` key, and the `~/.env`-wins collision holds in a non-default persona.

**L3 (🟡) — blast-radius docs.** The phantomchat setup backup warning and the persona-builder credential guidance now both state that the nsec derives the vault key, so losing `identity.json` makes **every** vault secret permanently unrecoverable, not just the Nostr identity.

**L2 (🟡) + Kai's two findings** were already resolved in eee5826: `writePersonaIdentity` (the rename footgun) is gone, both the phantomchat CLI generate path and `savePhantomchatPersonaConfig` use the atomic `createPersonaIdentityIfAbsent`, and that primitive fails closed on a malformed `identity.json`.

`tsc` clean, **1862 pass / 0 fail**. Ready for re-review — I have not self-approved.

kaieriksen (approved):
Re-reviewed the latest head. My previous blockers are fixed: vault reload now happens before harness spawns, phantomchat no longer persists nsec in phantomchat.json, identity creation is atomic create-if-absent, and malformed identity.json now fails closed instead of returning a transient key.\n\nVerified locally: focused vault/identity/phantomchat/env/persona-builder tests, typecheck, full suite (1862 pass, 1 skip), production build, CI green, mergeable.

lenaparkhodges (approved):
## Review round 3 — all findings resolved ✅

All three of my previous findings are properly addressed in the latest commits.

### L1 (breaking) — ~/.env fan-out to all personas: FIXED
migrateUserEnv now takes a personas[] array and iterates over every persona, writing and verifying read-back in each before deleting the plaintext. The localKeys map records won keys for ALL personas (not just default), so the central .env fan-out correctly skips them everywhere. The allOk gate ensures the plaintext survives if any persona fails. Two new regression tests cover both the non-default fan-out and the collision-in-non-default-persona case. This preserves the old global EnvironmentFile= behaviour — Kai keeps GITHUB_TOKEN etc.

### L2 — identity write race: FIXED
writePersonaIdentity (the rename-overwrite footgun) is completely removed. Both savePhantomchatPersonaConfig and the phantomchat CLI now use createPersonaIdentityIfAbsent — the atomic link()-based primitive that fails closed on an unreadable identity.json rather than handing back a transient key. The CLI also now displays the persisted nsec (not the in-memory one), so what the user sees matches what is durably on disk. Tests cover both the no-overwrite and fail-closed paths.

### L3 — blast-radius docs: FIXED
The phantomchat setup message and the persona-builder credential guidance now both state that the nsec derives the vault key, so losing identity.json makes all vault secrets permanently unrecoverable — not just the Nostr identity.

### Kai's findings — also resolved
Per-turn vault reload is in all four harnesses, and phantomchat.json no longer stores the nsec (config-only, with legacy promotion on next save).

### Known-minor (documented, not blocking)
The harness wizard still writes the Pi key to ~/.env; post-migration that value can be shadowed by the vault copy until the next daemon restart (which re-migrates). Self-heals on restart; the proper fix is out of scope.

CI is green (typecheck + 1862 tests pass). This is ready to merge.

kaieriksen (approved):
Re-reviewed latest head. The vault reload, phantomchat identity split, identity create-if-absent race fixes, fail-closed unreadable identity handling, all-persona ~/.env fan-out, and identity/vault blast-radius docs are aligned. Focused tests and typecheck pass locally.

andrewagrahamhodges (changes requested):
**Requesting one change before merge — vault reads should be per-row resilient.**

The crypto, identity minting, and migration are all excellent and I'm happy with the two spec-vs-reality deviations (`~/.env` fan-out is correct; `identity.json` having no backup is an acceptable *reconfigure-not-catastrophe* — worth one doc line saying "if you lose it: re-add secrets and re-pair phantomchat" so nobody panics).

The one thing I'd like fixed here rather than in a follow-up:

### Problem
In `readAllVaultValues`, decryption happens inside a single `get()`-loop that throws on the first failing row. The loader catches broadly and returns **empty**, so **one poisoned/corrupt row blanks *every* secret for the whole turn** — a bad `SENTRY_DSN` row costs the persona its `GITHUB_TOKEN`. And because the catch is silent, "one bad row" looks identical to "empty vault", which is an hour of wasted debugging.

### Requested change
Make the decrypt **per-row resilient**:
- Wrap the decrypt of each row individually. On a row that fails to decrypt: **skip it, count it, keep going.**
- Return every good secret plus the list of bad keys.
- Loader surfaces a warning like `⚠️ vault: 1 undecryptable key (SENTRY_DSN)` — **once per process start** (dedupe by key-set), not every turn.

### Why this shape
- **Blast radius matches the damage** — a corrupt row shouldn't mask N good ones.
- **Fail-*partial*, not fail-open** — we're not inventing or substituting values. A row you can't decrypt was already unusable, so no security regression.
- **Corruption becomes visible instead of silent** — you learn exactly what broke and that it's isolated.

### Scope
~15 lines in `readAllVaultValues` + a warn-log at the loader, plus two tests: (a) one deliberately-corrupted row → the other keys still load, (b) the bad key is reported.

Everything else is approvable as-is — this is the only blocker.

lenaparkhodges (changes requested):
Confirming Andrew's finding — this is a real blocker, not a nitpick.

I traced the exact path: `vault.get(name)` throws inside the `for` loop, the outer `try/catch` in `readAllVaultValues` returns `null`, and the loader's `null` branch strips **all** previously tracked keys from `env`. So one undecryptable row → every secret gone for the turn, silently. The silent part makes it worse — "one bad row" and "empty vault" are indistinguishable to the operator.

The fix shape Andrew described is correct and minimal: per-row try/catch in `readAllVaultValues`, return the good keys plus a list of bad ones, warn once per process start. ~15 lines + 2 tests. No security regression since a row you can't decrypt was already unusable.

Everything else from my prior review still holds — the migration, identity minting, and reconcile logic are all solid. This is the one remaining blocker.

src/lib/vault.ts
```diff
@@ -0,0 +1,371 @@
...
+  let vault: Vault;
+  try {
+    vault = await openPersonaVault(personaDirPath);
+  } catch {
+    return null;
+  }
+  const values = new Map<string, string>();
+  try {
+    for (const name of vault.list()) {
+      const value = vault.get(name);
```
kaieriksen: This `vault.get(name)` is still inside the single read loop without per-row error handling. If one row has a bad nonce/ciphertext/tag, `get()` throws, `readAllVaultValues()` never returns the successfully decrypted keys, and the loader treats the vault read as failed. Depending on whether this is the first/different persona load, that can remove every tracked vault secret from `env`, so one corrupt `SENTRY_DSN` row can blank `GITHUB_TOKEN` for the turn. Please catch decrypt failures per key, keep loading the good rows, and return/report the bad key names so the loader can warn once per process start instead of silently making the vault look empty/unreadable.

kaieriksen (changes requested):
I re-reviewed the current head after Andrew and Lena’s comments. I agree this needs one more change before merge: vault reads should tolerate one undecryptable row without dropping every other secret for the turn, and the bad key(s) should be surfaced once per process start.

robertclawson:
Thanks @andrewagrahamhodges and @kaieriksen — addressed the blocker. Two commits on the branch:

**`71e35f8` — per-row resilient decrypt (the blocker):**
- `readAllVaultValues` now decrypts each row inside its own `try/catch`. A row that fails to decrypt (corrupt nonce/ciphertext/tag, or a value written under a different key) is **skipped, its name recorded, and the loop keeps going** — one poisoned `SENTRY_DSN` no longer blanks `GITHUB_TOKEN` for the turn.
- Return shape changed to `{ values, badKeys }`. `null` is now reserved strictly for a vault that couldn't be **opened** (fresh install / missing identity / DB-level corruption) — so callers can tell "no vault" from "1 bad row" (badKeys empty vs. populated). `loadVaultIntoEnv` / `reloadVaultForPersona` propagate `badKeys`.
- The loader warns **once per process start, deduped by the exact bad-key set**, e.g. `vault: 1 undecryptable key (SENTRY_DSN) — skipped; other secrets loaded normally`. Corruption is visible, not silent, and not re-spammed every turn.
- Fail-**partial**, not fail-open: no value is ever invented or substituted, so a row that can't be decrypted (already unusable) is simply dropped — no security regression.
- Tests: (a) a corrupted row is skipped while the other secrets still load; (b) the bad key is reported and is distinguishable from a null vault. `bun tsc --noEmit` clean, vault suite 18/18.

**`654d981` — the doc line Andrew asked for:**
The phantomchat setup warning framed identity.json loss as "permanently unrecoverable", which reads like a catastrophe. It now spells out the recovery path: mint a fresh identity, re-add secrets with `phantombot vault set`, re-pair PhantomChat with the new npub — a reconfigure, not a catastrophe.

The `~/.env` fan-out and identity.json no-backup are unchanged, as agreed. Ready for another look.

lenaparkhodges (approved):
All three CHANGES_REQUESTED reviews (Andrew, Lena, Kai) are properly addressed in commits 71e35f82 + 654d981a.

**Per-row resilience (the blocker):** `vault.get(name)` is now wrapped per-row in its own try/catch. A single corrupt row (bad nonce/ciphertext/tag) is skipped and its name collected in `badKeys` — the remaining secrets still load. This directly fixes the failure mode Andrew identified: one bad `SENTRY_DSN` row no longer blanks `GITHUB_TOKEN` for the turn.

**Visibility:** `warnBadVaultKeys` surfaces the corrupt key names once per process start (deduped by key-set signature), so corruption is visible without spamming on every per-turn reload. The `badKeys` array is also returned in the result type so callers can distinguish "no vault" from "some bad rows."

**Docs:** The phantomchat setup message now frames identity loss as "a reconfigure, not a catastrophe" with concrete recovery steps (mint fresh identity, re-add secrets, re-pair). Better tone than the old "permanently unrecoverable" warning.

**Tests:** Two new tests — corrupt-row-skipped-with-others-intact and bad-key-name-reported. Both use real crypto (corrupting ciphertext in-place via SQL, then verifying GCM rejection). Solid.

**One edge I checked:** if `vault.list()` itself throws (SQLite schema corruption), that propagates to `reloadVaultForPersona`'s catch which returns empty arrays — but that's a fundamentally broken vault (not a single bad row), so losing all secrets is the correct behavior there. The identified failure mode is specifically per-row GCM auth failure, which is now handled.

CI: test (pass), typecheck (pass), Analyze (pass), CodeQL (pass). This is ready to merge.

kaieriksen (approved):
Re-reviewed latest head. The remaining vault-read blocker is fixed: decrypt failures are now isolated per row, good secrets continue loading, bad key names are returned and warned once per process start, and identity-loss docs now describe the reconfigure path.

Verified locally: focused vault/identity/phantomchat/persona-builder tests, typecheck, full suite (1864 pass, 1 skip), production build, CI green, mergeable.

andrewagrahamhodges (approved):
## Re-review — the blocker is resolved. Approving. ✅

Robbie implemented exactly the per-row-resilient shape I asked for, and I verified it against the actual code (not just the commit message) — read `readAllVaultValues` + `loadVaultIntoEnv`, ran the vault suite and typecheck locally at head `654d981`.

**Per-row resilience (the blocker):** each row's `vault.get()` is now wrapped in its own try/catch. A poisoned row (bad nonce/ciphertext/tag) is skipped, its name collected in `badKeys`, and the loop keeps going — so a corrupt `SENTRY_DSN` no longer costs the persona its `GITHUB_TOKEN`. Fail-*partial*, not fail-open: nothing is invented or substituted, so no security regression.

**"Bad row" vs "empty/null vault" is now distinguishable** — `readAllVaultValues` returns `{ values, badKeys }`, and `null` is reserved strictly for an unopenable vault (fresh install / missing identity / DB-level corruption). That's the exact ambiguity that used to cost an hour of debugging, gone.

**Visibility without spam:** `warnBadVaultKeys` logs once per process start, deduped by the exact bad-key set — names only, never values. Confirmed live in the test run:
`vault: 1 undecryptable key (SENTRY_DSN) — skipped; other secrets loaded normally`

**Tests:** both cases I specified are covered with *real* crypto — a row's ciphertext is corrupted in-place via SQL, then GCM rejects it: (a) the other secrets still load, (b) the bad key is reported and is distinct from a null vault.

**Docs:** the `identity.json` backup warning now reads as a *reconfigure, not a catastrophe*, with concrete recovery steps (mint fresh identity → re-add secrets → re-pair) — exactly the tone I wanted.

**Verified locally:** `tsc` clean; vault suite 18 pass / 0 fail. Combined with Kai's and Lena's approvals on the earlier rounds, this is ready to merge.

PR #255, opened 2026-07-04
feat(chattiness): restore narrate-on as the standing default

andrewagrahamhodges:
## What

Reverts the quiet-by-default behaviour introduced in #252. **Narration is once again the standing default whenever `chattiness` is unset** — no `config.toml`, an empty one, or a `config.toml` that omits the key all narrate. Operators opt *out* with `chattiness = false` or `/chattiness off`.

| Situation | #252 (current main) | this PR |
|---|---|---|
| No `config.toml` | OFF | **ON** |
| `config.toml` present, no `chattiness` key | OFF | **ON** |
| Empty `config.toml` | OFF | **ON** |
| `chattiness = true/false` | honored | honored |
| `PHANTOMBOT_CHATTINESS` env | wins | wins |

## Why

On large, tool-heavy coding runs, the running commentary appears to keep the agent **anchored** — it holds the thread and produces more reliable work. Quiet-by-default risks the agent drifting or losing context on long sessions where there's no interim narration to structure the run. Making narration the default protects the reliability of long runs; anyone who wants silence sets one key.

## How

- `DEFAULT_CHATTINESS` flipped back to `true`; the single fallback everywhere.
- Doc comments, resolution comment, and README updated to describe narrate-by-default and the opt-out.
- No change to the `/chattiness` command, per-chat overrides, or the write path.

## Verified

- Config tests updated for the ON expectations (no-file → ON, present-but-no-key → ON, empty → ON, explicit values honored, env wins both directions).
- Full suite **1825 pass / 0 fail**, `tsc` clean.

## Note for reviewers

This is a behaviour change for hosts that adopted #252's quiet default via the *implicit* path (no key set) — they'll narrate again after this. Intended. Anyone who genuinely wants quiet sets `chattiness = false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,14 +382,16 @@ errors always come through either way):
 Scoped to Telegram + PhantomChat (voice and the CLI never emit these bubbles).
 The editor (Zed/VS Code) surface follows the config default only.
 
-**The standing default is quiet.** Unless you explicitly set `chattiness`, a
-phantom starts **quiet** and just sends the final answer — this holds whether
-there's no `config.toml`, an empty one, or a `config.toml` that simply omits
-the key. Opt in to the running commentary by setting it:
+**The standing default is on.** Unless you explicitly set `chattiness`, a
+phantom **narrates** — it streams the running commentary as it works. This is
+deliberate: the play-by-play keeps the agent anchored across long, tool-heavy
+runs and empirically produces more reliable work on large tasks. The default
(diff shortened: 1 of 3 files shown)


andrewagrahamhodges:
Consolidating into #256 to keep all narration work in a single PR. #256 was stacked on this branch and already contains every commit from here (including the on-by-default change 541791a), and is now retargeted to main. Nothing is lost by closing this — #256 is the single source of truth for the narration default + command rename.

PR #256, opened 2026-07-04
feat(chattiness): narrate-on by default for unconfigured / empty config

andrewagrahamhodges:
## What this does

Makes the standing default **narrate-on** (`chattiness = true`) whenever the key is not explicitly set — no `config.toml`, an empty one, or a config that omits the `chattiness` key all narrate.

This reverts the quiet-by-default introduced in #252. Rationale (Andrew): narration acts as externalized working memory that keeps the agent anchored on long, tool-heavy coding runs — it should protect that risky case by default, and operators can mute the cheap case explicitly.

## Command surface — unchanged

The command stays **`/chattiness on|off|default`**. No rename. (An earlier revision of this branch experimented with `/narrateon` etc. — that has been dropped entirely; this PR is now purely the default change.)

- `/chattiness on|off` — per-conversation override
- `/chattiness on|off default` — writes `chattiness` to `config.toml` + clears the per-chat override

## Precedence (unchanged)

1. per-chat override
2. `chattiness` in `config.toml`
3. built-in default — now **on**
4. `PHANTOMBOT_CHATTINESS` env — wins over all

## Verified

- `DEFAULT_CHATTINESS = true`; config tests assert no-file / empty / missing-key all → narrate.
- Full suite **1825 pass / 0 fail**, typecheck clean.

Note: existing hosts with no `chattiness` key will narrate after this — intended; opt out with `chattiness = false` or `/chattiness off`.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -207,12 +207,13 @@ export interface TelegramStreamingSettings {
 /**
  * Standing default for interim progress-narration bubbles whenever no
  * `chattiness` key is set — regardless of whether a `config.toml` exists.
- * OFF (quiet) so every unconfigured phantom starts calm: just the final
- * answer, no running commentary, until an operator opts in via
- * `chattiness = true` or `/chattiness on`. Also the fallback used by read
- * sites given a partial config. See lib/chattiness.ts.
+ * ON so every phantom narrates by default: the running commentary keeps the
+ * agent anchored across long tool-heavy runs (empirically it holds the thread
+ * better and produces more reliable work on large tasks), and an operator can
+ * always quiet it via `chattiness = false` or `/chattiness off`. Also the
+ * fallback used by read sites given a partial config. See lib/chattiness.ts.
  */
-export const DEFAULT_CHATTINESS = false;
+export const DEFAULT_CHATTINESS = true;
 
 export const DEFAULT_TELEGRAM_STREAMING: TelegramStreamingSettings = {
   narrationFlushMs: 4500,
@@ -287,12 +288,12 @@ export interface Config {
    * per-conversation `/chattiness` override wins over this default; the final
    * reply and error paths are never affected either way. When unset in config
    * — no `chattiness` key, an empty file, or no `config.toml` at all — the
-   * default is `false` (quiet), so any not-yet-configured phantom starts calm
-   * and only narrates once an operator opts in. Also gates the editor (ACP)
+   * default is `true` (narrate), so every phantom keeps the agent anchored on
+   * long runs until an operator opts out. Also gates the editor (ACP)
    * surface's pre-tool narration — the config default only. See
    * lib/chattiness.ts. Optional in the type (mirrors telegramStreaming?) so
    * partial test fixtures need no update; loadConfig always sets it, and read
-   * sites default to `false` (DEFAULT_CHATTINESS) when absent.
+   * sites default to `true` (DEFAULT_CHATTINESS) when absent.
    */
   chattiness?: boolean;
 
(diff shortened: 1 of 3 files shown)


src/channels/commands.ts
```diff
@@ -230,8 +231,15 @@ export async function handleSlashCommand(
     case "/coder":
       // Bare `/coder` forces on; `/coder off|default` is also accepted.
       return await handleCoderSwap(arg || "on", ctx);
-    case "/chattiness":
-      return await handleChattiness(arg, ctx);
+    // Telegram's setMyCommands only allows [a-z0-9_], so the registered menu
+    // entries are `narrate_on` / `narrate_off`. Users often type the friendlier
+    // hyphen form, so accept both spellings for each.
+    case "/narrate-on":
```
kaieriksen: Good compatibility detail accepting both spellings here. That matches Telegram menu constraints without forcing users to remember the underscore form.

kaieriksen (approved):
Approved. The rename is tightly scoped, keeps the Telegram menu constraint covered, and preserves the existing override/default semantics.

andrewagrahamhodges:
Updated per Andrew: dropped the hyphen entirely — commands are now **`/narrateon`** and **`/narrateoff`** (single word).

Because a single word is already Telegram-legal (`setMyCommands` only allows `[a-z0-9_]`), this also **deletes the hyphen/underscore dual-accept shim** from the dispatcher — the single-word form is the one and only spelling, and it registers verbatim in the command menu. Old `/narrate-on`/`/narrate_on` spellings now fall through to the LLM (covered by a test).

Updated README, architecture doc, module + inline comments, and tests. Full suite **1821 pass / 0 fail**, typecheck clean.

src/channels/commands.ts
```diff
@@ -230,8 +231,10 @@ export async function handleSlashCommand(
     case "/coder":
       // Bare `/coder` forces on; `/coder off|default` is also accepted.
       return await handleCoderSwap(arg || "on", ctx);
-    case "/chattiness":
-      return await handleChattiness(arg, ctx);
+    case "/narrateon":
```
kaieriksen: This final spelling is the right shape for Telegram: it registers verbatim via setMyCommands and avoids the hyphen/underscore alias layer entirely.

kaieriksen (dismissed):
Re-reviewed latest head. The final single-word /narrateon and /narrateoff spelling matches Telegram command-menu constraints, and the dispatcher/docs/tests are aligned.

andrewagrahamhodges:
Consolidated: this PR now carries the **complete** narration change against `main` (previously stacked on #255, which is now closed). Full scope:

1. **Default restored to narrate-on** when `chattiness` is unset (no config / empty / missing key).
2. **Command rename** — `/chattiness` → `/narrateon` / `/narrateoff` (single-word, Telegram-legal).
3. **Docs** — README command matrix + precedence rules, architecture.md notes.

4 commits, MERGEABLE against main. Single source of truth for all narration work.

src/config.ts
```diff
@@ -207,12 +207,13 @@ export interface TelegramStreamingSettings {
...
- * `chattiness = true` or `/chattiness on`. Also the fallback used by read
- * sites given a partial config. See lib/chattiness.ts.
+ * ON so every phantom narrates by default: the running commentary keeps the
+ * agent anchored across long tool-heavy runs (empirically it holds the thread
+ * better and produces more reliable work on large tasks), and an operator can
+ * always quiet it via `chattiness = false` or `/chattiness off`. Also the
+ * fallback used by read sites given a partial config. See lib/chattiness.ts.
  */
-export const DEFAULT_CHATTINESS = false;
+export const DEFAULT_CHATTINESS = true;
```
kaieriksen, line 216: This is the behavior change I re-checked: absent env/TOML now resolves through `DEFAULT_CHATTINESS = true`, and the updated config tests cover no file, omitted key, empty file, explicit false/true, and env override precedence.

kaieriksen (approved):
Approved. The latest head is scoped to restoring the standing chattiness default to ON, and the docs/config tests match the runtime behavior.

PR #257, opened 2026-07-04
Windows Port (#201)

### LINKED ISSUE #201: Windows Port — Single Bun Binary with User-Scoped Service
# Windows Port of Phantombot

## Goal

Build a native Windows version of Phantombot that:
- Runs as a user-scoped background service (no admin required)
- Builds and provisions the Pi capability-routing extension
- Has full CLI parity with Linux/macOS
- Ships as a single compiled Bun binary (`.exe`)
- No reliance on the Windows registry, DLLs, or external runtimes

---

## Design Principles (Gotchas & Rules)

These aren't suggestions — they're constraints derived from the differences between POSIX and Windows that will silently break things if ignored.

### 1. No POSIX Assumptions in Core Code

Every hardcoded `/home/`, `/tmp/`, `/var/`, `/etc/`, `/usr/` path breaks on Windows. The fix is already partially there — `os.homedir()`, `os.tmpdir()`, `path.join()` — but audit every file path in the codebase:

- **Lock files** (`runLock.ts`): Currently defaults to `/tmp/phantombot-<uid>.run.lock` with a `/proc` fallback. On Windows: use `os.tmpdir()` → `%TEMP%\phantombot.run.lock`. The `/proc/<pid>/stat` instance-token mechanism (PID reuse protection) has no Windows equivalent — fall back to a simpler token using `wmic process where "ProcessId=<pid>" get CreationDate` or just accept PID-only liveness on Windows (same as macOS already does).

- **Run lock path**: `defaultLockPath()` needs a Windows branch. `%LOCALAPPDATA%\phantombot\phantombot.run.lock` is the correct user-scoped location (not `%TEMP%`, which is shared and cleared on reboot).

- **Config paths**: `~/.config/phantombot/` → `%APPDATA%\phantombot\` on Windows. The `config.ts` XDG resolution already uses `os.homedir()` as fallback, but verify every `join(homedir(), ".config", ...)` call has a Windows-aware alternative.

- **Pi extension path**: `~/.pi/agent/extensions/capability-routing/` → `%USERPROFILE%\.pi\agent\extensions\capability-routing\`. The path separator issue is handled by `path.join()`, but verify the extension provisioning code (`piExtensionProvision.ts`) uses it consistently.

- **Logs directory**: macOS uses `~/Library/Logs/phantombot/`. Windows should use `%LOCALAPPDATA%\phantombot\logs\`.

### 2. Process Signals Are Different

Windows doesn't have POSIX signals. The current codebase uses `SIGTERM`, `SIGINT`, `SIGHUP`, `SIGKILL` extensively:

- **`SIGTERM`**: Actually works on Windows via `process.kill(pid, 'SIGTERM')` — Node/Bun translate it to `TerminateProcess`. But it's not graceful — no handler runs. For graceful shutdown, use `SIGINT` or a named event.

- **`SIGKILL`**: Works (`TerminateProcess`). Use as escalation after grace period.

- **`SIGINT`**: Works on Windows (Ctrl+C equivalent). Fires `process.on('SIGINT')`.

- **`SIGHUP`**: Does NOT exist on Windows. Any code listening for SIGHUP needs a no-op fallback or a Windows-native equivalent (e.g., listen for console close event via `process.on('SIGHUP', () => {})` — Bun may emit this, verify).

- **Process group kill**: `process.kill(-pid, signal)` (negative PID for process group) is POSIX-only and does nothing on Windows. The `processGroup.ts` `signalGroup()` function will silently fail. Fix: on Windows, use `taskkill /F /T /PID <pid>` (tree kill) instead of negative-PID group kill. The `killProcessGroup()` function needs a Windows codepath.

- **`detached: true`**: On Windows, this creates a new process group (`CREATE_NEW_PROCESS_GROUP`), not a new session. Combined with `CREATE_NO_WINDOW` for headless operation. The semantics are similar enough but the kill path differs — see above.

### 3. Service Management — Task Scheduler, Not Services

Windows Services require admin to register. For a user-scoped, no-install-dependency approach:

- **Use Task Scheduler** as the service manager. It supports:
  - Run at logon (user-scoped, no admin)
  - Restart on failure (configurable delay and max retries)
  - Hidden window (no console popup)
  - Trigger on unlock, not just logon
  - Multiple scheduled tasks (main + heartbeat + nightly + tick — matching Linux)

- **`platform.ts` router** needs a `"win32"` case that dispatches to a new `src/lib/windowsScheduler.ts` module (parallel to `systemd.ts` / `launchd.ts`).

- **`phantombot install` on Windows**: Register 4 Task Scheduler tasks:
  - `Phantombot` — main agent (run at logon, restart on failure, 5s delay)
  - `PhantombotHeartbeat` — every 30 minutes
  - `PhantombotNightly` — daily at 02:00
  - `PhantombotTick` — every 60 seconds

- **Task XML**: Generate `<Task>` XML programmatically (same pattern as systemd unit / launchd plist generation). Use `schtasks /Create /XML <file> /TN <name>` to register.

- **Why not a Windows Service**: Requires running as SYSTEM or a service account, needs `sc.exe` or Win32 API registration (admin), doesn't have easy access to the user's environment/credentials, and complicates the single-binary story. Task Scheduler is the user-scoped equivalent.

### 4. Single Bun Binary — Console Window Suppression

A headless service must NOT show a console window:

- **`bun build --compile`** produces a console subsystem `.exe` by default. When launched by Task Scheduler with `<Hidden>true</Hidden>`, the console window is suppressed. Verify this actually works — some Task Scheduler + console app combinations still flash a window briefly on logon.

- **Alternative**: Build with `--compile --windows-hide-console` (if Bun supports it) or post-process with `editbin /SUBSYSTEM:WINDOWS` (requires Visual Studio tools). The cleanest path is the Task Scheduler `<Hidden>` flag — avoid the editbin dependency.

- **WER suppression**: Windows Error Reporting shows a dialog on crash. For a headless service, suppress via:
  ```reg
  HKLM\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps
  ```
  Or set `SetErrorMode(SEM_NOGPFAULTERRORBOX | SEM_FAILCRITICALERRORS)` early in main. Bun may not expose this directly — test and document.

### 5. Pi Extension Provisioning on Windows

The Pi capability-routing extension must build and provision correctly:

- **Extension source path**: Use `path.join()` consistently. The embedded assets are stamped into `%USERPROFILE%\.pi\agent\extensions\capability-routing\` — verify `piExtensionProvision.ts` doesn't assume `/` separators in any string operations.

- **Pi install script**: Currently `curl -fsSL https://pi.dev/install.sh | sh`. On Windows, Pi likely has a Windows installer or can be installed via npm/bun. The `harnessAvailability.ts` needs to search PATHEXT (`.COM;.EXE;.BAT;.CMD`) when resolving harness binaries.

- **`whichBinary()` / `resolveHarnessBinary()`**: Uses `which` (POSIX) to find harness binaries. On Windows, use Bun's `which` (which should handle PATHEXT) or fall back to `where.exe`. Verify Bun's `which` resolves `pi.exe` correctly.

- **Extension spawning**: `spawnPi.ts` uses `child_process.spawn` with `shell: false`. On Windows, this works for `.exe` files. If Pi ships as a `.cmd` wrapper, `shell: true` may be needed — test both cases.

### 6. Process Tree Kill — The Hard One

The most dangerous gotcha. `killProcessGroup()` in `processGroup.ts` uses `process.kill(-pid, signal)` — POSIX process group kill. On Windows:

- **Negative PID kill does nothing** (no ESRCH, no error — silent no-op). Child processes become orphans.

- **Fix**: Detect `process.platform === 'win32'` and use `taskkill /F /T /PID <pid>` instead. `taskkill` with `/T` kills the entire process tree. Wrap in a `Bun.spawn(['taskkill', '/F', '/T', '/PID', String(pid)])` call.

- **Grace period**: The current pattern is SIGTERM → wait → SIGKILL. On Windows: `taskkill /PID <pid>` (graceful, if the process handles it) → wait → `taskkill /F /T /PID <pid>` (force tree kill).

- **Race condition**: On Windows, between spawning `taskkill` and it completing, new children can spawn. Accept this — same race exists on POSIX with `setsid` timing. The 5s grace window is sufficient.

- **`spawnInNewSession()`**: On Windows, `detached: true` creates a new process group (not session). This is sufficient for `taskkill /T` to work — the tree is rooted at the spawned PID.

### 7. Shell Commands and PATH

- **`tick.ts`** uses `shell: true` for command-backed tasks. On Windows, this invokes `cmd.exe /d /s /c "<command>"`. Verify command syntax works in cmd.exe (e.g., `&&` chaining, quoting rules differ from bash).

- **PATH separator**: Windows uses `;` not `:`. The `PHANTOMBOT_SERVICE_PATH` in `systemd.ts` is Linux-specific and not used on Windows, but any PATH manipulation in shared code must use `path.delimiter`.

- **PATHEXT**: When resolving commands, Windows uses PATHEXT (`.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC`). Bun's `which` should handle this, but verify.

- **Default shell**: The Bash tool and harness commands assume `/bin/bash` or `sh`. On Windows, the default shell for `shell: true` is `cmd.exe`. For PowerShell, use `shell: 'powershell'` or invoke `pwsh.exe` explicitly. Consider making the default shell configurable in `config.toml`.

### 8. File Locking

Windows has mandatory file locking (vs. advisory on POSIX):

- **SQLite**: Bun's built-in SQLite uses WAL mode. WAL works on Windows but the shared memory file (`-shm`) can cause issues if the process crashes — a stale `.shm` file can prevent recovery. SQLite handles this internally with `walIndexRecovery()`, but document it.

- **`runLock.ts`**: The `O_EXCL` create pattern works on Windows (`openSync(path, 'wx')` maps to `CREATE_NEW`). But if the process crashes while holding the lock, the file isn't cleaned up. The PID liveness check handles this — verify `process.kill(pid, 0)` works on Windows for checking if another process is alive (it does).

- **Hot reload**: Don't try to write to a file that another process has open for writing. Windows will throw `EBUSY` (not the POSIX `ETXTBSY`). File watching uses `ReadDirectoryChangesW` (not inotify) — Bun handles this, but verify `fs.watch` works correctly.

### 9. Encoding and Unicode

- **Console code page**: Windows console defaults to the system locale code page (e.g., CP1252), not UTF-8. Bun handles UTF-8 internally, but subprocess output (Pi, Claude, Gemini) may use the system code page. Set `chcp 65001` in any shell invocations, or rely on Bun's `spawn` which uses UTF-8 pipes by default.

- **File encoding**: All config files, `.env`, logs should be UTF-8. `readFileSync(path, 'utf8')` works on Windows — no BOM issues as long as files are written without BOM (which `writeFile` does by default).

- **MAX_PATH**: Windows limits paths to 260 characters by default. Deep workspace paths or `node_modules` can hit this. Enable long paths via:
  - Registry: `HKLM\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled = 1`
  - Or use `\\?\` prefix for absolute paths
  - Document this in the Windows install guide. Bun should handle `\\?\` paths transparently.

### 10. Firewall and Network

- **First run**: Windows Firewall will prompt for network access on first launch. The Telegram long-polling and API calls need inbound/outbound rules. `phantombot install` on Windows should attempt to add a firewall rule via `netsh advfirewall` (requires admin — make it optional/warn if it fails).

- **WebSocket**: PhantomChat's WebSocket server needs to be accessible. Same firewall consideration.

- **Proxy**: Some corporate Windows environments use proxy servers. Bun respects `HTTP_PROXY`/`HTTPS_PROXY` env vars — verify this works.

### 11. Credential and Env File Handling

- **`.env` file location**: Currently `~/.env` and `~/.config/phantombot/.env`. On Windows, these become `%USERPROFILE%\.env` and `%APPDATA%\phantombot\.env`. The env bootstrap code should detect the platform and use the right paths.

- **File permissions**: Unix `chmod 600` doesn't exist on Windows. `.env` files are protected by ACLs (default: owner-only). Document that users should not widen permissions on credential files. Consider using `%APPDATA%\phantombot\credentials.env` which inherits the user's profile ACL (already restrictive).

- **SSH keys**: `~/.ssh/id_ed25519` → `%USERPROFILE%\.ssh\id_ed25519`. Windows SSH (built-in since Windows 10 1809) uses the same path convention. Verify Bun's SSH client finds keys in the right location.

### 12. Git and Symlinks

- **Git symlinks**: The Pi extension manual symlink (`ln -sfn`) requires Developer Mode or admin on Windows. Alternative: use a directory junction (`mklink /J`) which doesn't require elevated privileges.

- **Git line endings**: `core.autocrlf` may convert LF → CRLF. The codebase should work with both. Test that generated files (systemd units, plists, extension assets) aren't broken by line ending conversion.

### 13. Build Pipeline

- **Cross-compilation**: Bun supports `bun build --compile --target=bun-windows-x64` from Linux/macOS. The CI pipeline needs a Windows build target.

- **Native modules**: If any dependencies use native addons (`.node` files), they need Windows-compiled variants. Audit dependencies for native code. Bun's built-in SQLite is native but bundled — should work.

- **Binary size**: A single Windows `.exe` with Bun compile is typically 40-60MB. Acceptable for a service binary.

- **Testing**: CI needs a Windows runner (GitHub Actions `windows-latest`). Run the full test suite on Windows to catch platform-specific failures. Some tests mock `process.getuid()` — verify they work when the function is undefined (Windows).

### 14. Logging

- **No journald**: Use file-based logging (same approach as macOS). Logs go to `%LOCALAPPDATA%\phantombot\logs\`.

- **`phantombot logs`**: Equivalent of `journalctl -f`. Use `Get-Content -Wait -Tail 100 <log-file>` or implement a built-in log tail.

- **Event Log**: Optionally write to Windows Event Log for integration with monitoring tools. Not required for MVP.

### 15. Update Mechanism

- **Self-update**: The current `phantombot update` replaces the running binary. On Windows, you can't replace a running `.exe` file (mandatory file locking). Fix:
  - Download new binary to `%LOCALAPPDATA%\phantombot\update\phantombot.exe`
  - Stop the service (via Task Scheduler)
  - Replace the binary
  - Restart the service
  - Or: use the "rename and replace" trick — rename running `.exe` → `.old`, write new one, clean up `.old` on next start.

### 16. Uninstall

`phantombot uninstall` on Windows should:
- Remove Task Scheduler tasks (`schtasks /Delete /TN <name> /F`)
- Remove the installation directory
- Remove `%APPDATA%\phantombot\` and `%LOCALAPPDATA%\phantombot\`
- Optionally remove the firewall rule
- Do NOT touch the registry (design principle: no registry reliance)

---

### Phase 1: Core Port (Make It Run)
- Audit all path assumptions in shared code, add Windows branches
- Implement `runLock.ts` Windows path and fallback instance token
- Implement `processGroup.ts` Windows tree kill
- Add `"win32"` case to `platform.ts` with `noopServiceControl()` initially
- Verify `bun build --compile --target=bun-windows-x64` produces a working binary
- Verify all tests pass on Windows CI runner

### Phase 2: Service Management (Make It Persist)
- Implement `src/lib/windowsScheduler.ts` (Task Scheduler management)
- Implement `phantombot install` / `phantombot uninstall` for Windows
- Implement `phantombot status` / `phantombot logs` for Windows
- Test: install → run → restart → uninstall lifecycle

### Phase 3: Harness & Extension Provisioning (Make It Useful)
- Verify Pi, Claude, Gemini, Codex harness binaries resolve on Windows (PATHEXT)
- Verify Pi extension provisioning writes correct paths
- Verify Pi install script works on Windows (or provide Windows alternative)
- Verify `spawnPi.ts` works with Windows process spawning
- Test: full harness chain on Windows

### Phase 4: Edge Cases & Hardening (Make It Robust)
- WER suppression for headless operation
- MAX_PATH handling / long paths documentation
- Firewall rule helper (optional, warn on failure)
- File locking audit (SQLite WAL, lock files, env files)
- Encoding audit (subprocess output, config files)
- Self-update mechanism for Windows

---

## Open Questions

1. **Pi on Windows**: Does Pi ship a Windows binary? If not, do we need to build one or use npm/bun to install it?
2. **Claude Code on Windows**: Claude Code (the CLI) supports Windows natively. Verify the binary resolves correctly.
3. **Gemini CLI on Windows**: Check if `gemini` CLI supports Windows.
4. **Codex on Windows**: OpenAI Codex CLI — check Windows support.
5. **Bun `--compile` subsystem**: Can we target Windows GUI subsystem to suppress the console window natively, or is Task Scheduler `<Hidden>` sufficient?
6. **Testing scope**: Do we need full E2E tests on Windows, or just unit + integration? E2E with real Telegram would be the same as Linux.

---

## Acceptance Criteria

- [ ] `phantombot.exe` compiles as a single binary from the same source
- [ ] `phantombot.exe install` registers 4 Task Scheduler tasks (no admin required)
- [ ] `phantombot.exe run` starts the agent, connects to Telegram, processes messages
- [ ] `phantombot.exe run` shows NO console window when launched by Task Scheduler
- [ ] Heartbeat, nightly, and tick timers fire correctly via Task Scheduler
- [ ] Pi capability-routing extension provisions to `%USERPROFILE%\.pi\agent\extensions\`
- [ ] Harness binaries (Pi, Claude, Gemini, Codex) are found and spawned correctly
- [ ] Process tree kill works (no orphaned child processes)
- [ ] Lock file prevents duplicate instances
- [ ] `phantombot.exe uninstall` removes all artifacts cleanly
- [ ] All existing tests pass on Windows CI runner
- [ ] No registry keys, DLLs, or external runtime dependencies

andrewagrahamhodges:
## Windows Port (#201)

Native Windows build of phantombot: a single compiled Bun `.exe`, user-scoped
background service (no admin), with full CLI parity. This is the umbrella branch
for the full port - **one branch, one commit per phase, merged only when the
whole port is proven**. Kept as a **draft** so it tracks the work without merge
risk.

Closes #201 when complete.

### Approach
- Build and test on a real Windows VM, not just CI.
- Single-root data layout: config, data and state all under
  `%LOCALAPPDATA%\phantombot` (no roaming).
- Logon-triggered Task Scheduler for the `systemctl --user` / macOS-style
  "runs while logged in" model (no admin). True headless-without-login is
  out of scope; README will point at WinSW for anyone who needs it.

### Phase status
- [x] **Phase 1 - Core POSIX->Windows shims** (`4acb29d`)
  - Paths: config/data/state resolve to a single `%LOCALAPPDATA%\phantombot`
    root (verified live on the VM).
  - Run-lock: Windows falls back to per-user `%TEMP%`.
  - Tree-kill: replaced POSIX negative-pid group signal with `taskkill /T /F`.
    (Fixed a real bug: `taskkill /T` without `/F` returns exit 128 on a live
    console tree, which read as "already dead" and hung the kill. Now always
    force-kills.)
  - `platform` enum gains `windows`; startup credential bootstrap now skipped
    on `--help`/`--version` so read-only commands don't mutate state or
    provision a stray persona.
  - New `tests/windows-port.test.ts` + Windows-aware fixes to existing suites.
  - Verified: typecheck clean; full Linux suite green (1883 pass); affected
    suites on Windows 82 pass / 6 skip (POSIX-only) / 0 fail.
- [x] **Phase 2 - Vault clean-create + owner-only `identity.json` ACL** (`e34306e`)
  - `lib/filePermissions.ts`: `restrictFileToCurrentUser()` applies an
    owner-only ACL via `icacls` (`/inheritance:r` + `/grant:r *<SID>:F`) on
    Windows, no-op on POSIX. Fails closed. Applied to the tmp file before the
    atomic hard-link, so `identity.json` is restrictive from birth.
  - Grant by **SID**, not name: on a workgroup box `%USERDOMAIN%` is
    `WORKGROUP`, not a valid authority, so `icacls` by name fails.
  - Vault clean-create verified on the VM (`bun:sqlite` read/write).
  - Perms tests made platform-aware; added `tests/fixtures/rmrf.ts` (Bun.gc
    before rm to dodge Windows EBUSY from bun:sqlite's GC-deferred handle
    release; test-only).
  - Verified: typecheck clean both OSes; identity+vault suites 35 pass / 0 fail
    on Linux AND the Windows VM.
- [x] **Phase 3 - User-scoped service (Task Scheduler logon trigger)** (`c75a314`)
  - New `lib/taskScheduler.ts`: generates Task Scheduler 1.2 XML for four
    per-user tasks under `\Phantombot\` (`run`, `heartbeat`, `nightly`,
    `tick`), registered via `schtasks /Create /XML /F`. Principal is the
    current user by **SID** with `LogonType=InteractiveToken` - no admin, no
    stored password, runs only while logged in (the macOS model).
  - Keep-alive without a supervisor: `LogonTrigger` + a 1-minute repeating
    `TimeTrigger` + `MultipleInstancesPolicy=IgnoreNew` restarts the agent if
    it dies, admin-free.
  - Logging: actions route through `cmd /c` with stdout/stderr appended to
    per-task `.out.log`/`.err.log` under `%LOCALAPPDATA%\phantombot\logs`
    (Task Scheduler doesn't capture stdout). Log **rotation** is a documented
    limitation - WinSW is the upgrade path.
  - XML written UTF-16LE + BOM (schtasks import requirement); bin-name check
    made separator-agnostic and accepts `phantombot.exe`.
  - Wired `windows` into `platform.ts` (service control + restart/status/logs
    hints), `install.ts`, `uninstall.ts`. Reuses `currentUserSid()` from
    Phase 2.
  - Verified end-to-end on the VM: `install` registers all four tasks with no
    password prompt; XML carries `InteractiveToken` + the real exe path;
    running the `tick` task creates its redirected log files; `uninstall`
    removes all four and a follow-up query returns "not found". Unit suites
    (`lib-taskScheduler`, `lib-platform`, `cli-install`) 37 pass / 0 fail on
    Linux AND the Windows VM; typecheck clean both OSes; full Linux suite
    green (1904 pass).
- [x] **Phase 4** - CI (`windows-latest`) + coverage baseline + README Windows docs (`4af8842`)

### Notes for reviewers
- The full `bun test` suite is not yet green on Windows - other suites across
  the repo still assume POSIX. Scoping which tests run on `windows-latest` is
  Phase 4 work.
- No migration path is exercised on Windows (no legacy installs exist); the
  clean-create path is what matters.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,8 +3,8 @@ name: Release on PR merge
 # Fires on every push to main (i.e. every PR merge — branch protection
 # requires PRs, so a push to main IS a merge). Cuts a GitHub Release tagged
 # v1.1.<RUN_NUMBER> with cross-compiled binaries (linux x64-baseline +
-# arm64, darwin x64 + arm64) and a SHA256SUMS file. `phantombot update`
-# reads this feed.
+# arm64, darwin x64 + arm64, windows x64 + arm64) and a SHA256SUMS file.
+# `phantombot update` reads this feed.
 #
 # Why `push: main` and NOT `pull_request: closed`: a PR merged from a
 # FORK runs the closed-PR workflow with a *read-only* GITHUB_TOKEN —
@@ -177,6 +177,29 @@ jobs:
             ./src/index.ts \
             --outfile "dist/phantombot-${TAG}-darwin-arm64"
 
+      - name: Build windows-x64
+        env:
+          TAG: ${{ steps.version.outputs.tag }}
+        run: |
+          # Cross-compiled from the Linux runner; Bun emits an unsigned PE
+          # (.exe). Windows SmartScreen may warn on first run for an
+          # unsigned binary - see release notes. --windows-hide-console is
+          # NOT set: phantombot is a console agent and needs its stdio.
+          bun build --compile --target=bun-windows-x64 \
+            ./src/index.ts \
+            --outfile "dist/phantombot-${TAG}-windows-x64.exe"
+
+      - name: Build windows-arm64
+        env:
+          TAG: ${{ steps.version.outputs.tag }}
+        run: |
+          # ARM64 Windows (e.g. Surface Pro X, Dev Kit). Bun's compile
+          # target id is bun-windows-arm64; the emitted PE is Aarch64.
+          # Same unsigned/SmartScreen caveat as windows-x64.
+          bun build --compile --target=bun-windows-arm64 \
+            ./src/index.ts \
+            --outfile "dist/phantombot-${TAG}-windows-arm64.exe"
+
       - name: Compute SHA256SUMS
         working-directory: dist
         env:
@@ -187,6 +210,8 @@ jobs:
             "phantombot-${TAG}-linux-arm64" \
             "phantombot-${TAG}-darwin-x64" \
             "phantombot-${TAG}-darwin-arm64" \
+            "phantombot-${TAG}-windows-x64.exe" \
+            "phantombot-${TAG}-windows-arm64.exe" \
             > SHA256SUMS
           cat SHA256SUMS
 
@@ -223,6 +248,8 @@ jobs:
           - phantombot-${TAG}-linux-arm64 — ARM64
           - phantombot-${TAG}-darwin-x64 — macOS x86-64 (unsigned; see below)
           - phantombot-${TAG}-darwin-arm64 — macOS Apple Silicon (unsigned; see below)
+          - phantombot-${TAG}-windows-x64.exe - Windows x86-64 (unsigned; see below)
+          - phantombot-${TAG}-windows-arm64.exe - Windows ARM64 (unsigned; see below)
 
           Verify with: sha256sum -c SHA256SUMS
 
@@ -234,6 +261,11 @@ jobs:
           attribute on first run:
             xattr -d com.apple.quarantine ./phantombot-${TAG}-darwin-arm64
             xattr -d com.apple.quarantine ./phantombot-${TAG}-darwin-x64
+
+          Windows note: the .exe binaries are cross-compiled and unsigned, so
+          SmartScreen may show a "Windows protected your PC" prompt on first
+          run - choose More info then Run anyway, or unblock the file:
+            Unblock-File .\\phantombot-${TAG}-windows-x64.exe
           EOF
 
       - name: Create GitHub Release with binary assets
@@ -263,4 +295,6 @@ jobs:
             "dist/phantombot-${TAG}-linux-arm64" \
             "dist/phantombot-${TAG}-darwin-x64" \
             "dist/phantombot-${TAG}-darwin-arm64" \
+            "dist/phantombot-${TAG}-windows-x64.exe" \
+            "dist/phantombot-${TAG}-windows-arm64.exe" \
             "dist/SHA256SUMS"
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ Supported harnesses:
 - [Why Phantombot Exists](#why-phantombot-exists)
 - [Install](#install)
 - [Quick Start](#quick-start)
+- [Windows (preview)](#windows-preview)
 - [Configuration](#configuration)
 - [Command Reference](#command-reference)
 - [Telegram](#telegram)
@@ -193,6 +194,68 @@ running after logout:
 ```bash
 sudo loginctl enable-linger "$USER"
 ```
+
+## Windows (preview)
+
+Phantombot runs on Windows (x64). The port shares ~95% of its code with the
+Linux and macOS builds; the platform-specific pieces (data paths, process
+tree-kill, run-lock, credential-store ACL, and the background service) have
+native Windows implementations. There is no prebuilt Windows release yet, so
+you build the binary from source.
+
+**Build the binary** (needs [Bun](https://bun.sh) on the Windows machine):
+
+```powershell
+git clone https://github.com/phantomyard/phantombot.git
+cd phantombot
+bun install
+bun run build:win        # produces dist\phantombot.exe
+
+# Put it somewhere stable and on your PATH, e.g.:
+mkdir "$env:LOCALAPPDATA\Programs\phantombot"
+copy dist\phantombot.exe "$env:LOCALAPPDATA\Programs\phantombot\phantombot.exe"
+```
+
+Then configure it exactly as on Linux (`phantombot persona`, `harness`,
+`telegram`, …).
+
+**Data location.** Everything lives under a single root,
+`%LOCALAPPDATA%\phantombot` (config, personas, memory database, logs, and the
+run-lock). Nothing roams between machines. The crown-jewel `identity.json` is
+created with an owner-only ACL (`icacls`, inheritance stripped) so other
+accounts on the box cannot read it.
+
+**Install as a background service.**
+
+```powershell
+phantombot install      # registers per-user Task Scheduler tasks under \Phantombot\
+phantombot uninstall    # removes them
+```
+
+`install` registers four logon-triggered scheduled tasks (`run`, `heartbeat`,
+`nightly`, `tick`). They run as **you**, using an interactive-token principal —
+**no admin rights and no stored password required**.
+
+**Important — the agent runs only while you are logged in.** This mirrors the
+macOS (launchd `Aqua`) model, not the Linux `enable-linger` model. When you log
+out, Windows tears down your session and the tasks stop; they start again at
+your next logon. A LogonTrigger plus a one-minute keep-alive re-launches the
+agent if it crashes while you are logged in.
+
+For a **true headless service** that runs without an interactive login (a
+kiosk, a server, or a machine that reboots unattended), wrap `phantombot run`
+in a service supervisor such as [WinSW](https://github.com/winsw/winsw) or NSSM
+and run it under a dedicated service account. That path is not yet scripted by
+`phantombot install`; it is the documented escape hatch for now.
+
+**Logs.** Service stdout/stderr are redirected to
+`%LOCALAPPDATA%\phantombot\logs\*.out.log` / `*.err.log`. These currently grow
+unbounded (no built-in rotation yet) — prune them periodically if the box runs
+for a long time.
+
+Status: the Windows port is exercised by a dedicated `windows-latest` CI job on
+every pull request. Treat it as a **preview** — solid enough to run, but newer
+than the Linux/macOS paths.
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -197,23 +197,30 @@ sudo loginctl enable-linger "$USER"
 
 ## Windows (preview)
 
-Phantombot runs on Windows (x64). The port shares ~95% of its code with the
-Linux and macOS builds; the platform-specific pieces (data paths, process
-tree-kill, run-lock, credential-store ACL, and the background service) have
-native Windows implementations. There is no prebuilt Windows release yet, so
-you build the binary from source.
+Phantombot runs on Windows (x64 and arm64). The port shares ~95% of its code
+with the Linux and macOS builds; the platform-specific pieces (data paths,
+process tree-kill, run-lock, credential-store ACL, the background service, and
+self-update) have native Windows implementations.
 
-**Build the binary** (needs [Bun](https://bun.sh) on the Windows machine):
+Every release publishes prebuilt, unsigned Windows binaries -
+`phantombot-<tag>-windows-x64.exe` and `phantombot-<tag>-windows-arm64.exe` -
+alongside the SHA256SUMS file. Download the one for your architecture, verify
+its checksum, and drop it somewhere stable on your PATH:
+
+```powershell
+# Unblock the downloaded file (SmartScreen marks internet downloads):
+Unblock-File .\phantombot-<tag>-windows-x64.exe
+mkdir "$env:LOCALAPPDATA\Programs\phantombot"
+copy .\phantombot-<tag>-windows-x64.exe "$env:LOCALAPPDATA\Programs\phantombot\phantombot.exe"
+```
+
+**Or build from source** (needs [Bun](https://bun.sh) on the Windows machine):
 
 ```powershell
 git clone https://github.com/phantomyard/phantombot.git
 cd phantombot
 bun install
 bun run build:win        # produces dist\phantombot.exe
-
-# Put it somewhere stable and on your PATH, e.g.:
-mkdir "$env:LOCALAPPDATA\Programs\phantombot"
-copy dist\phantombot.exe "$env:LOCALAPPDATA\Programs\phantombot\phantombot.exe"
 ```
 
 Then configure it exactly as on Linux (`phantombot persona`, `harness`,
@@ -247,6 +254,16 @@ kiosk, a server, or a machine that reboots unattended), wrap `phantombot run`
 in a service supervisor such as [WinSW](https://github.com/winsw/winsw) or NSSM
 and run it under a dedicated service account. That path is not yet scripted by
 `phantombot install`; it is the documented escape hatch for now.
+
+**Self-update.** `phantombot update` and the `/update` chat command work on
+Windows. Because Windows locks a running `.exe` against overwrite, the updater
+renames the live binary aside to `phantombot.exe.old` (allowed while it runs),
+drops the verified new binary into place, then exits cleanly - the one-minute
+keep-alive trigger relaunches the agent on the new binary within ≈60 seconds
+(no console windows, no manual restart). The relaunched process deletes the
+leftover `.old` on startup once it is unlocked. In-place self-update needs the
+service installed (`phantombot install`) so the keep-alive is present to bring
+the agent back up.
 
 **Logs.** Service stdout/stderr are redirected to
 `%LOCALAPPDATA%\phantombot\logs\*.out.log` / `*.err.log`. These currently grow
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "gen:vscode-vsix": "bun scripts/genVscodeVsix.ts",
     "build": "bun run build:x64",
     "build:x64": "mkdir -p dist && bun build --compile --target=bun-linux-x64-baseline ./src/index.ts --outfile dist/phantombot",
-    "build:arm64": "mkdir -p dist && bun build --compile --target=bun-linux-arm64 ./src/index.ts --outfile dist/phantombot-arm64"
+    "build:arm64": "mkdir -p dist && bun build --compile --target=bun-linux-arm64 ./src/index.ts --outfile dist/phantombot-arm64",
+    "build:win": "mkdir -p dist && bun build --compile --target=bun-windows-x64 ./src/index.ts --outfile dist/phantombot.exe"
   },
   "dependencies": {
     "@clack/prompts": "^1.3.0",
--- a/package.json
+++ b/package.json
@@ -16,9 +16,9 @@
     "gen:pi-extension": "bun scripts/genPiExtensionAssets.ts",
     "gen:vscode-vsix": "bun scripts/genVscodeVsix.ts",
     "build": "bun run build:x64",
-    "build:x64": "mkdir -p dist && bun build --compile --target=bun-linux-x64-baseline ./src/index.ts --outfile dist/phantombot",
-    "build:arm64": "mkdir -p dist && bun build --compile --target=bun-linux-arm64 ./src/index.ts --outfile dist/phantombot-arm64",
-    "build:win": "mkdir -p dist && bun build --compile --target=bun-windows-x64 ./src/index.ts --outfile dist/phantombot.exe"
+    "build:x64": "bun build --compile --target=bun-linux-x64-baseline ./src/index.ts --outfile dist/phantombot",
+    "build:arm64": "bun build --compile --target=bun-linux-arm64 ./src/index.ts --outfile dist/phantombot-arm64",
+    "build:win": "bun build --compile --target=bun-windows-x64 ./src/index.ts --outfile dist/phantombot.exe"
   },
   "dependencies": {
     "@clack/prompts": "^1.3.0",
--- a/src/lib/filePermissions.ts
+++ b/src/lib/filePermissions.ts
@@ -31,7 +31,7 @@
  * "No mapping between account names and security IDs". The SID always maps and
  * is immune to renamed accounts and localized principal names.
  */
-function currentUserSid(): string {
+export function currentUserSid(): string {
   const res = Bun.spawnSync(["whoami", "/user", "/fo", "csv", "/nh"]);
   const out = new TextDecoder().decode(res.stdout).trim();
   // Output shape: "DOMAIN\user","S-1-5-21-…"
--- a/src/lib/filePermissions.ts
+++ b/src/lib/filePermissions.ts
@@ -1,0 +1,69 @@
+/**
+ * Cross-platform "lock this file down to just me" primitive.
+ *
+ * On POSIX the caller already writes sensitive files at mode 0o600, and the
+ * kernel enforces it — nothing to do here, so this is a no-op. On Windows the
+ * POSIX permission bits passed to writeFile are IGNORED: a freshly created file
+ * inherits its parent directory's ACL, which on a shared machine can grant read
+ * access to other principals (BUILTIN\Users, etc.). For the persona's crown
+ * jewel — identity.json, which holds the nsec every vault secret is encrypted
+ * under — that is not acceptable, so we apply an explicit restrictive ACL.
+ *
+ * The Windows lockdown does two things via `icacls`:
+ *   /inheritance:r   — remove all INHERITED ACEs (drops whatever the parent dir
+ *                      would have granted: Users, Authenticated Users, etc.)
+ *   /grant:r <me>:F  — replace any explicit grant for the current user with a
+ *                      single Full-control ACE.
+ * The net DACL is exactly one ACE: the current user, full control. Everyone
+ * else — including standard users on the same box — is denied by absence.
+ *
+ * Fails CLOSED: if the current account can't be resolved or icacls errors, this
+ * THROWS. Callers apply it to a tempfile BEFORE hard-linking it into place, so a
+ * throw means no identity.json is ever created — far better than silently
+ * persisting the nsec in a file other users could read.
+ */
+
+/**
+ * Resolve the current user's SID (e.g. `S-1-5-21-…-1001`) and grant by SID
+ * rather than by name. Account NAMES are ambiguous on Windows: on a workgroup
+ * (non-domain) machine `%USERDOMAIN%` is the workgroup name `WORKGROUP`, which
+ * is NOT a valid account authority, so `icacls /grant WORKGROUP\user` fails with
+ * "No mapping between account names and security IDs". The SID always maps and
+ * is immune to renamed accounts and localized principal names.
+ */
+function currentUserSid(): string {
+  const res = Bun.spawnSync(["whoami", "/user", "/fo", "csv", "/nh"]);
+  const out = new TextDecoder().decode(res.stdout).trim();
+  // Output shape: "DOMAIN\user","S-1-5-21-…"
+  const m = out.match(/(S-1-[0-9-]+)/);
+  if (!m) {
+    throw new Error(`could not resolve current user SID (whoami: ${out || "no output"})`);
+  }
+  return m[1]!;
+}
+
+/**
+ * Restrict `path` so ONLY the current user can access it. No-op on POSIX (the
+ * file's mode 0o600 already governs); applies an owner-only ACL on Windows.
+ * Throws on Windows if the lockdown cannot be verified to have succeeded.
+ */
+export function restrictFileToCurrentUser(path: string): void {
+  if (process.platform !== "win32") return;
+
+  // Grant by SID (`*S-1-…`) — see currentUserSid for why names are unsafe.
+  const sid = currentUserSid();
+  const res = Bun.spawnSync([
+    "icacls",
+    path,
+    "/inheritance:r",
+    "/grant:r",
+    `*${sid}:(F)`,
+  ]);
+  if (res.exitCode !== 0) {
+    const stderr = new TextDecoder().decode(res.stderr).trim();
+    const stdout = new TextDecoder().decode(res.stdout).trim();
+    throw new Error(
+      `icacls lockdown failed for ${path}: ${stderr || stdout || `exit ${res.exitCode}`}`,
+    );
+  }
+}
--- a/tests/lib-processGroup.test.ts
+++ b/tests/lib-processGroup.test.ts
@@ -13,11 +13,21 @@
  */
 
 import { afterEach, describe, expect, test } from "bun:test";
+import { delimiter, dirname } from "node:path";
 import {
   killProcessGroup,
   spawnInNewSession,
   withCommandDirOnPath,
 } from "../src/lib/processGroup.ts";
+
+// The spawn/kill tests below drive real POSIX subprocesses (`sleep`, `sh -c`,
+// `#!/usr/bin/env node` shebangs) and the negative-pid group signal — none of
+// which exist on Windows, where the equivalent path is taskkill /T (covered by
+// its own Windows-only test at the bottom). Skip the POSIX-model suites on
+// Windows rather than assert against a process model that isn't there.
+const isWin = process.platform === "win32";
+const describePosix = isWin ? describe.skip : describe;
+const describeWin = isWin ? describe : describe.skip;
 
 function isAlive(pid: number): boolean {
   try {
@@ -86,7 +96,7 @@ afterEach(() => {
   trackedPids.length = 0;
 });
 
-describe("spawnInNewSession", () => {
+describePosix("spawnInNewSession", () => {
   test("starts the binary with the same pid==pgid (the kill-the-group precondition)", async () => {
     // We can't observe pgid directly without /proc, but we CAN verify
     // that the process started and is reachable via its own pid.
@@ -106,7 +116,7 @@ describe("spawnInNewSession", () => {
   });
 });
 
-describe("killProcessGroup — orphan grandchild fix", () => {
+describePosix("killProcessGroup — orphan grandchild fix", () => {
   test("SIGTERM to the group reaps both parent AND grandchild", async () => {
     // Shell spawns a backgrounded sleep, prints its pid, then waits.
     // Without process-group kill, killing the shell would leave the
@@ -139,7 +149,7 @@ describe("killProcessGroup — orphan grandchild fix", () => {
   });
 });
 
-describe("killProcessGroup — SIGTERM→SIGKILL escalation", () => {
+describePosix("killProcessGroup — SIGTERM→SIGKILL escalation", () => {
   test("escalates to SIGKILL when SIGTERM is trapped/ignored", async () => {
     // Use a Bun process that registers a no-op SIGTERM handler so the
     // signal is delivered but the process keeps running. Only SIGKILL
@@ -194,20 +204,26 @@ describe("killProcessGroup — SIGTERM→SIGKILL escalation", () => {
   });
 });
 
+// These are pure string tests with no subprocess, so they run on every
+// platform. We build expected PATHs from the host's `path.delimiter`
+// (":" on POSIX, ";" on Windows) and `path.dirname` — the same primitives
+// withCommandDirOnPath uses — so the assertions hold on both without
+// hardcoding a POSIX separator.
 describe("withCommandDirOnPath — shebang interpreter resolution", () => {
   test("prepends an absolute binary's own dir to PATH", () => {
-    const out = withCommandDirOnPath("/opt/nvm/versions/node/v24/bin/pi", {
-      PATH: "/usr/bin:/bin",
-    });
-    expect(out.PATH).toBe("/opt/nvm/versions/node/v24/bin:/usr/bin:/bin");
+    const bin = "/opt/nvm/versions/node/v24/bin/pi";
+    const before = ["/usr/bin", "/bin"].join(delimiter);
+    const out = withCommandDirOnPath(bin, { PATH: before });
+    expect(out.PATH).toBe([dirname(bin), before].join(delimiter));
   });
 
   test("is a no-op when the dir is already on PATH", () => {
-    const env = { PATH: "/opt/nvm/versions/node/v24/bin:/usr/bin" };
-    const out = withCommandDirOnPath("/opt/nvm/versions/node/v24/bin/pi", env);
+    const bin = "/opt/nvm/versions/node/v24/bin/pi";
+    const env = { PATH: [dirname(bin), "/usr/bin"].join(delimiter) };
+    const out = withCommandDirOnPath(bin, env);
     // Same object reference back — nothing to change, no needless clone churn.
     expect(out).toBe(env);
-    expect(out.PATH).toBe("/opt/nvm/versions/node/v24/bin:/usr/bin");
+    expect(out.PATH).toBe([dirname(bin), "/usr/bin"].join(delimiter));
   });
 
   test("handles an empty/absent PATH by seeding it with the bin dir", () => {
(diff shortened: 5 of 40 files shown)


lenaparkhodges (approved):
## Phase 1 Windows Port — Review ✅

Clean, well-structured PR. All CI green (typecheck, test, CodeQL), and locally verified: 1881 pass (+19 new), 2 pre-existing PiHarness routing failures (same on main, env-related), tsc clean.

### What's good

**Path resolution ()** — All three xdg\*Home functions correctly gate on `process.platform === "win32"` and collapse to `%LOCALAPPDATA%` on Windows while preserving XDG\_\* env override. `memoryIndexPath` already routes through `xdgDataHome()`, so the heartbeat.ts refactor to use it is correct and removes a POSIX literal.

**Read-only invocation guard ( + )** — `isReadOnlyInvocation` is a clean, dependency-free helper. `--help`/`--version`/bare invocations skip vault bootstrap and persona provisioning entirely, which fixes a real CI smoke-test issue. The `phantombot vault --help` case correctly still runs bootstrap (only `argv[2]` is checked) — documented and tested.

**Process group kill ()** — `windowsKillTree` using `taskkill /T /F` is the right approach. The exit-128 = "already gone" mapping mirrors ESRCH. `detached: process.platform !== "win32"` prevents the Bun-hangs-on-exit issue. `withCommandDirOnPath` correctly uses `path.isAbsolute` + `path.delimiter` instead of POSIX-only `startsWith("/")` + `":"`.

**Run lock ()** — Per-user `%TEMP%` fallback is correct (no uid on Windows, already user-scoped). XDG_RUNTIME_DIR still wins if set.

**Tests** — `windows-port.test.ts` (143 lines) covers all platform-conditional paths with clean `withPlatform`/`withEnv` stubbing. POSIX subprocess tests are properly gated with `describePosix`/`describeWin` so they skip on the wrong OS instead of failing. Existing test assertions updated to use `path.join`/`path.delimiter` for cross-platform correctness.

### Non-blocking notes

1. **`windowsKillTree` uses `execFileSync` (synchronous)** — Acceptable since taskkill is fast, but an async `execFile` would be more event-loop-friendly. Not a blocker.

2. **Single-root layout on Windows** — config, data, and state all collapse to `%LOCALAPPDATA%\phantombot`. Intentional and documented, but Phase 2/3 should consider whether subdirectories (e.g. `config/`, `data/`, `state/`) would reduce clutter as the port matures.

3. **`taskkill` ENOENT edge case** — If taskkill somehow isn't on PATH, `execFileSync` throws with `status: undefined` (not 128), so it falls through to `log.warn` + returns `true` ("signal delivered"). The caller still awaits `proc.exited`, so it's not catastrophic, but returning `true` for a failed kill is slightly misleading. A `code === 'ENOENT'` check could return `false` instead. Edge case — taskkill is always available on Windows.

No blockers. Ready for Phase 2 when Megan is.

lenaparkhodges (dismissed):
## Phase 1 Windows Port — Review ✅

Clean, well-structured PR. All CI green (typecheck, test, CodeQL), and locally verified: 1881 pass (+19 new), 2 pre-existing PiHarness routing failures (same on main, env-related), tsc clean.

### What's good

**Path resolution (config.ts)** — All three xdg*Home functions correctly gate on `process.platform === "win32"` and collapse to `%LOCALAPPDATA%` on Windows while preserving XDG_* env override. `memoryIndexPath` already routes through `xdgDataHome()`, so the heartbeat.ts refactor to use it is correct and removes a POSIX literal.

**Read-only invocation guard (cliInvocation.ts + index.ts)** — `isReadOnlyInvocation` is a clean, dependency-free helper. `--help`/`--version`/bare invocations skip vault bootstrap and persona provisioning entirely, which fixes a real CI smoke-test issue. The `phantombot vault --help` case correctly still runs bootstrap (only `argv[2]` is checked) — documented and tested.

**Process group kill (processGroup.ts)** — `windowsKillTree` using `taskkill /T /F` is the right approach. The exit-128 = "already gone" mapping mirrors ESRCH. `detached: process.platform !== "win32"` prevents the Bun-hangs-on-exit issue. `withCommandDirOnPath` correctly uses `path.isAbsolute` + `path.delimiter` instead of POSIX-only `startsWith("/")` + `":"`.

**Run lock (runLock.ts)** — Per-user `%TEMP%` fallback is correct (no uid on Windows, already user-scoped). XDG_RUNTIME_DIR still wins if set.

**Tests** — `windows-port.test.ts` (143 lines) covers all platform-conditional paths with clean `withPlatform`/`withEnv` stubbing. POSIX subprocess tests are properly gated with `describePosix`/`describeWin` so they skip on the wrong OS instead of failing. Existing test assertions updated to use `path.join`/`path.delimiter` for cross-platform correctness.

### Non-blocking notes

1. **`windowsKillTree` uses `execFileSync` (synchronous)** — Acceptable since taskkill is fast, but an async `execFile` would be more event-loop-friendly. Not a blocker.

2. **Single-root layout on Windows** — config, data, and state all collapse to `%LOCALAPPDATA%\phantombot`. Intentional and documented, but Phase 2/3 should consider whether subdirectories (e.g. `config/`, `data/`, `state/`) would reduce clutter as the port matures.

3. **`taskkill` ENOENT edge case** — If taskkill somehow isn't on PATH, `execFileSync` throws with `status: undefined` (not 128), so it falls through to `log.warn` + returns `true` ("signal delivered"). The caller still awaits `proc.exited`, so it's not catastrophic, but returning `true` for a failed kill is slightly misleading. A `code === 'ENOENT'` check could return `false` instead. Edge case — taskkill is always available on Windows.

No blockers. Ready for Phase 2 when Megan is.

tests/lib-processGroup.test.ts
```diff
@@ -236,7 +252,33 @@ describe("withCommandDirOnPath — shebang interpreter resolution", () => {
   });
 });
 
-describe("spawnInNewSession — shebang interpreter on a narrow PATH", () => {
+describeWin("killProcessGroup — Windows taskkill tree", () => {
```
kaieriksen, line 255: Good place to keep this regression. Phase 4's windows-latest job should include this file so the taskkill /T /F path is exercised on a real Windows runner instead of only being present behind describeWin here.

kaieriksen (dismissed):
Approved. I checked the Windows shims, path/root behavior, CLI bootstrap guard, process-tree cleanup semantics, and build path; no blockers found.

src/lib/filePermissions.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (process.platform !== "win32") return;
+
+  // Grant by SID (`*S-1-…`) — see currentUserSid for why names are unsafe.
+  const sid = currentUserSid();
+  const res = Bun.spawnSync([
+    "icacls",
+    path,
+    "/inheritance:r",
+    "/grant:r",
+    `*${sid}:(F)`,
```
kaieriksen, line 60: Granting by SID here is the right shape for Windows. It avoids the workgroup/domain-name mapping trap and keeps the failure mode closed before `identity.json` is linked into place.

kaieriksen (dismissed):
Re-reviewed the Phase 2 ACL update. The identity creation path still fails closed, keeps atomic first-writer-wins behavior, and the Windows compile/checks are clean.

lenaparkhodges (dismissed):
## Phase 2 — Re-review ✅

New commit `e34306eb` adds Phase 2: owner-only ACL on `identity.json` for Windows. This is the right fix — POSIX mode bits are ignored on Windows, so without an explicit ACL the crown-jewel nsec file would inherit the parent directory's permissions (potentially readable by other users on a shared machine).

**`restrictFileToCurrentUser()`** is well-designed:
- Grants by SID (`*S-1-…`) not by name — avoids the workgroup/domain mapping trap where `WORKGROUP\user` fails on non-domain machines
- `/inheritance:r` + `/grant:r` produces exactly one ACE (owner, full control) — no inherited or broad principals
- Applied to the tempfile BEFORE `link()`, so the canonical `identity.json` is restrictive from birth
- Fails closed: a lockdown error throws before `link()`, so no world-readable identity.json is ever created
- No-op on POSIX (mode 0600 already governs)

**`rmrf` fixture** is a solid pragmatic fix for the bun:sqlite handle-linger-on-Windows teardown race. `Bun.gc(true)` before each retry is the right call — empirical, documented, and inert on POSIX.

Tests are platform-aware: `expectOwnerOnly` checks mode 0600 on POSIX and DACL properties on Windows (no inherited ACEs, no BUILTIN\Users, no Everyone, no Authenticated Users).

Verification: 1883 pass, 0 fail, 2 skip (Windows-only taskkill + Pi ARG_MAX), tsc clean. All failures from my previous run were pre-existing missing-dependency errors resolved by `bun install`.

My three non-blocking notes from Phase 1 still stand (sync execFileSync, single-root layout, ENOENT edge case) — none are blockers. No new issues from Phase 2.

Ready for merge, Andrew.

andrewagrahamhodges (commented):
## Code Review: Windows Port (#201)

**Author:** @andrewagrahamhodges | **Files:** 28 | **Verdict:** Comment Only

### Findings

| 1 | 🟡 Minor | `src/lib/taskScheduler.ts` | 200 | `execFileSync` in `windowsKillTree` is synchronous — blocks the event loop during the kill. Acceptable for a short-lived `taskkill` call, but worth noting. |
| 2 | 🔵 Trivial | `src/lib/filePermissions.ts` | 50 | `icacls` command is Windows-only — correctly guarded by `process.platform !== "win32"` in the caller. |
| 3 | 🔵 Trivial | `src/lib/processGroup.ts` | 215 | `windowsKillTree` exit-128 → false (ESRCH analogue) is the correct mapping. |
| 4 | 🔵 Trivial | `tests/fixtures/rmrf.ts` | 1 | 50 attempts × 100ms = 5s max retry for Windows EBUSY — generous but needed for bun:sqlite handle lingering. |

### What's Good
- **Security is solid**: `restrictFileToCurrentUser()` applies owner-only ACL via `icacls /inheritance:r /grant:r *<SID>:(F)`. SID-based grants avoid the workgroup username resolution bug. Fails closed — no world-readable `identity.json` is ever created.
- **Process management is correct**: `detached: true` only on POSIX (Windows detaching prevents Bun from observing exit). `taskkill /T /F` replaces POSIX group signal — the `/F` fix addresses a real bug where `/T` without `/F` returns exit 128 on live console trees.
- **Task Scheduler design is clever**: Keep-alive via LogonTrigger + 1-min TimeTrigger + `IgnoreNew` emulates launchd's KeepAlive without a supervisor. XML generation handles escaping, UTF-16LE + BOM for schtasks import.
- **Path resolution is clean**: All three XDG paths collapse under `%LOCALAPPDATA%` on Windows. Run lock falls back to `%TEMP%`. The `xdg*` functions check env vars first, so test overrides work.
- **Self-heal on Windows**: `ensureTasksCurrent` queries each task's XML, checks if the bin path matches (case-insensitive), and re-imports if drifted. Mirrors the Linux systemd self-heal.
- **Tests are comprehensive**: 381-line `lib-taskScheduler.test.ts` covers XML generation, install/uninstall, self-heal, UTF-16LE encoding. `windows-port.test.ts` covers path resolution, platform enum, run lock, read-only invocation. `rmrf` handles Windows EBUSY from bun:sqlite.
- **CI is correct**: `windows-latest` job runs curated test subset with `--coverage` reporting (not a hard threshold). Typecheck on Windows too.
- **Documentation is thorough**: README explains the logged-in-only model, WinSW escape hatch for headless, log rotation limitation.

<details>

```
No changes needed. The Windows port is well-executed with correct security, process management, and service design. All CI checks pass.
```
</details>

robertclawson (dismissed):
## Windows Port (#201) — Full Review ✅

Read the whole port end-to-end at head `4af8842` — the three security-critical shims (`filePermissions.ts`, `taskScheduler.ts`, `processGroup.ts`), the identity-creation ordering, the CLI bootstrap guard, and all the platform wiring. **Safe, no blockers. Approving.**

### Injection surface — clean
This is the axis I scrutinised hardest, since a Windows port shells out to `icacls`, `schtasks`, and `taskkill`.
- **All external calls are array-arg spawns** (`Bun.spawn`/`Bun.spawnSync`/`execFileSync`) — no shell, so no command-injection surface. Task names, args (`run`/`heartbeat`/`nightly`/`tick`), and schtasks verbs are hardcoded constants; the XML is passed as a file path, not interpolated onto a command line.
- **SID is regex-validated** (`/S-1-[0-9-]+/`) before it's used as an `icacls` principal or an XML `<UserId>` — closes the workgroup name-mapping trap and any injection via the principal.
- **XML text is escaped** for `& < >`; the only interpolated values are the validated SID, hardcoded descriptions/URIs, `process.execPath`, and `%LOCALAPPDATA%`-derived log paths. Quotes are intentionally preserved (they matter for the `cmd /c` redirection) and are safe here — Windows filenames can't contain `"`.
- **`taskkill /PID <pid> /T /F`** takes `String(pid)` as an array arg; the `/F`-always rationale (exit 128 → genuine ESRCH mapping) is correct and well-documented.

### Correctness — the parts that actually matter
- **identity.json ACL ordering is right:** `restrictFileToCurrentUser()` is applied to the tmp file *before* the atomic hard-link, and **fails closed** (throws → no identity.json ever created world-readable). The hard-link shares the security descriptor, so the canonical file is owner-only from birth. The race-safe first-writer-wins + losers-adopt invariant is preserved on Windows.
- **Bootstrap guard is correct:** `isReadOnlyInvocation` skips vault migration/persona provisioning for `--help`/`--version`/bare — so CI smoke tests stay pure and don't mutate disk.
- **`detached` only on POSIX:** correctly avoided on Windows (where it can leave `proc.exited` unresolved), with `taskkill /T` as the tree-kill instead. Good call.
- **Single-root layout:** config/data/state all resolve to `%LOCALAPPDATA%\phantombot` — no roaming, predictable.
- **Keep-alive:** LogonTrigger + 1-min TimeTrigger + `IgnoreNew` + `RestartOnFailure` is a sound admin-free launchd-analogue; heartbeat self-heal (`ensureTasksCurrent`) repairs drift on cadence, mirroring the systemd path.

### CI
Fully green including the **real `test-windows` job on `windows-latest`** (not just Linux emulation) — plus typecheck, `test`, and both CodeQL analyses. The port was also verified live on a Windows VM per the phase notes.

### Non-blocking notes
- `noopServiceControl`'s doc comment still says "anything other than linux/darwin" — Windows is now handled; trivial doc staleness.
- The scoped-subset Windows test run (full suite still assumes POSIX elsewhere) is an explicitly documented, accepted limitation — fine for this PR.

Nice, careful work — the "grant by SID, fail closed, no admin" discipline is consistent throughout. 🐶

lenaparkhodges (dismissed):
## Phase 3 + Phase 4 — Re-review ✅

Five new commits since my Phase 2 review (`e34306e`). The PR is now the complete Windows port: shims, ACL, service backend, CI, and docs.

### Phase 3 — Task Scheduler service (`d7f3f16`)

`taskScheduler.ts` (576 lines) is well-architected:

- **XML generation** — clean separation: `generateTaskXml` is the shared builder, four thin wrappers (`phantombot`/`heartbeat`/`nightly`/`tick`) customize triggers and limits. XML escaping is correct (`&`/`<`/`>` entities, redirection operators escaped inside `<Arguments>`). UTF-16LE + BOM encoding for schtasks import is handled explicitly via `Buffer.concat`.
- **Keep-alive without a supervisor** — `LogonTrigger` + 1-minute repeating `TimeTrigger` + `MultipleInstancesPolicy=IgnoreNew` is a clever, correct emulation of launchd's KeepAlive. If the agent is running, the minute-tick is ignored; if it died, the next tick restarts it. No admin, no stored password.
- **SID-based principal** — reuses `currentUserSid()` from Phase 2. Correct lesson applied: grant by SID, not name.
- **Self-heal** — `ensureTasksCurrent` mirrors the systemd pattern: query each task's XML, case-insensitive bin-path comparison, re-import only if drifted. Idempotent on a healthy box (four cheap `/Query` calls, nothing else). Called from the heartbeat cadence via `defaultHealTaskScheduler`, with the same `basename(binPath).startsWith("phantombot")` guard so dev runs never rewrite tasks.
- **Install/uninstall** — wired into `install.ts`/`uninstall.ts` with the `platform: "windows"` dispatch. Binary-name check now handles `phantombot.exe` via `split(/[/\\]/).pop().replace(/\.exe$/i, "")` — correct cross-platform. `/F` on both `/Create` and `/Delete` makes operations idempotent.
- **Platform router** — `platform.ts` correctly adds `windows` to the enum, dispatches `defaultTaskSchedulerServiceControl()`, and provides `restartCommand`/`statusCommand`/`logsCommand` strings. `defaultHealService()` dispatches to systemd on Linux, Task Scheduler on Windows, no-op on macOS (launchd self-heals via KeepAlive).

### Phase 4 — CI + tests + docs

- **CI** (`a850d43`) — `test-windows` job runs in parallel on `windows-latest` with a curated subset (12 test files exercising win32 branches). Bun install cache for speed. Typecheck on Windows too. No path filter (correct — win32 behaviour lives in shared modules). Coverage as reporting, not a hard threshold (right call for a partial suite).
- **rmrf fixture** (`7d56d59`) — `Bun.gc(true)` before each retry is the right empirical fix for bun:sqlite handle-linger. Inert on POSIX. 50×100ms headroom is generous but justified for multi-persona migration teardown.
- **Two Windows test fixes** (`4af8842`) — heartbeat test now asserts per-platform (`no_target` on Windows vs `release_check_failed` on POSIX); the invariant (heartbeat didn't fail) holds both ways. rmrf retry headroom raised from 20 to 50 for multi-persona migration fan-out.
- **README** (`e74e5d3`) — thorough: build instructions, data location, ACL, install/uninstall, logged-in-only model, WinSW escape hatch, log rotation gap. Accurately labeled as "preview."

### Verification

- `bunx tsc --noEmit` — clean
- `bun test` — 1906 pass, 2 skip (Windows-only taskkill + Pi ARG_MAX), 2 fail (pre-existing PiHarness routing, same on `main`)
- All CI green: `test`, `test-windows`, `typecheck`, `CodeQL`, `Analyze`

### Kai's threads

Both are Kai's — his to resolve:
1. `tests/lib-processGroup.test.ts:255` — "Phase 4's windows-latest job should include this file" — **addressed**: the CI curated list includes `tests/lib-processGroup.test.ts`, so `describeWin` runs the taskkill test on a real Windows runner.
2. `src/lib/filePermissions.ts:60` — positive observation on SID-based grants, no action needed.

My three non-blocking notes from Phase 1 (sync `execFileSync`, single-root layout, ENOENT edge case) are all still non-blocking and acknowledged in the PR description. The ENOENT case is now handled correctly (returns `false` per my suggestion — thanks Megan).

lenaparkhodges (approved):
Re-approved on `756a504`. ✅

The only new commit extends `release.yml` to cross-compile Windows x64 + arm64 binaries via Bun's `--compile --target=bun-windows-*` and adds them to the SHA256SUMS + release assets. Comments are thorough (unsigned PE / SmartScreen caveat, ARM64 naming, no `--windows-hide-console` rationale).

No source code changes, no new risk surface. All CI green including `test-windows`. My previous approval on `4af8842` stands — all four phases remain solid. Ready for merge.

robertclawson (dismissed):
Re-reviewed at head `756a504d` — still **APPROVED**, no blockers. 🐶

My earlier deep review of the port itself (shims, Task Scheduler XML generator, identity.json ACL, bootstrap guard, kill semantics) stands unchanged — none of that code moved. The only delta since my last approval is the release-workflow commit, which I read line-by-line:

- Adds `bun build --target=bun-windows-x64` / `bun-windows-arm64` steps that **mirror the existing linux/darwin ones exactly** — same `TAG` from `steps.version.outputs.tag`, no new secrets, no new action pins, no shell-injection surface.
- Both `.exe` artifacts are **folded into `SHA256SUMS`** and the release upload, so `phantombot update` still verifies integrity of the Windows binaries the same way as every other platform.
- Unsigned/SmartScreen caveat is documented in the release notes (parallel to the existing macOS quarantine note). Correct call — cross-compiled PEs are unsigned.

CI fully green at head: `test`, `typecheck`, `test-windows` (real `windows-latest` job), CodeQL x2, actions analysis.

**No threads to resolve** — all my reviews here are body-only (the inline threads on this PR are Kai's). So I'm clean: approved, nothing outstanding on my side.

package.json
```diff
@@ -17,7 +17,8 @@
     "gen:vscode-vsix": "bun scripts/genVscodeVsix.ts",
     "build": "bun run build:x64",
     "build:x64": "mkdir -p dist && bun build --compile --target=bun-linux-x64-baseline ./src/index.ts --outfile dist/phantombot",
-    "build:arm64": "mkdir -p dist && bun build --compile --target=bun-linux-arm64 ./src/index.ts --outfile dist/phantombot-arm64"
+    "build:arm64": "mkdir -p dist && bun build --compile --target=bun-linux-arm64 ./src/index.ts --outfile dist/phantombot-arm64",
+    "build:win": "mkdir -p dist && bun build --compile --target=bun-windows-x64 ./src/index.ts --outfile dist/phantombot.exe"
```
kaieriksen: This script is now the documented Windows path (`README.md` tells users to run `bun run build:win` from PowerShell), but it still starts with POSIX `mkdir -p`. On a stock Windows shell this is not portable, so the first build-from-source path for Windows users can fail before Bun gets to the cross-compile. Please make the directory creation cross-platform, e.g. move the build into a tiny Bun script (`await mkdir('dist', { recursive: true }); await Bun.$\`bun build --compile --target=bun-windows-x64 ./src/index.ts --outfile dist/phantombot.exe\`;`) or use another shell-neutral setup before invoking `bun build`.

README.md
```diff
@@ -194,6 +195,68 @@ running after logout:
 sudo loginctl enable-linger "$USER"
 ```
 
+## Windows (preview)
+
+Phantombot runs on Windows (x64). The port shares ~95% of its code with the
+Linux and macOS builds; the platform-specific pieces (data paths, process
+tree-kill, run-lock, credential-store ACL, and the background service) have
+native Windows implementations. There is no prebuilt Windows release yet, so
```
kaieriksen: This doc now conflicts with the release workflow changes in the same PR, which add Windows x64/arm64 release assets. After this merges, there will be prebuilt Windows binaries. Suggest changing this section to say users can download the Windows release binary once available, and keep the source-build instructions as the fallback/dev path.

kaieriksen (changes requested):
I found one blocker in the new Windows-facing build path. The implementation/tests are otherwise clean locally: focused Windows subset passed, typecheck passed, and the full suite passed.

lenaparkhodges (dismissed):
Re-reviewing on commit `756a504d` (no new commits since my last approval).

**Verdict: Still approved.** ✅

No code changes since my last review — 1906 tests pass, tsc clean, all CI green (including `test-windows`). The 2 PiHarness failures are pre-existing on `main`.

Regarding Kai's two inline threads:

**`package.json:21` — `mkdir -p` in `build:win`:** I checked this — `bun run` executes npm scripts through Bun's own built-in shell (`bun:sh`), which supports POSIX-style `mkdir -p` on all platforms including Windows. So `bun run build:win` from PowerShell will work because the script runs in Bun's shell, not PowerShell's `mkdir` alias. Not a blocker. That said, moving to a cross-platform Bun script (`await mkdir('dist', { recursive: true })`) would be cleaner and remove any ambiguity — worth doing as a follow-up, but not merge-blocking.

**`README.md:203` — prebuilt binaries not mentioned:** Valid doc inconsistency — the release workflow in this PR adds prebuilt Windows binaries, so the README should mention that as the primary path with source-build as fallback. Cosmetic, not blocking.

Both are Kai's threads to resolve. I had no inline threads (my reviews were top-level approvals). No blockers from my side. Ready for merge, Andrew.

.github/workflows/release.yml
```diff
@@ -223,6 +248,8 @@ jobs:
           - phantombot-${TAG}-linux-arm64 — ARM64
           - phantombot-${TAG}-darwin-x64 — macOS x86-64 (unsigned; see below)
           - phantombot-${TAG}-darwin-arm64 — macOS Apple Silicon (unsigned; see below)
+          - phantombot-${TAG}-windows-x64.exe - Windows x86-64 (unsigned; see below)
```
kaieriksen, line 251: This Windows asset is listed in release notes that still say `Install with: phantombot update`, but the update code still rejects win32 targets (`detectSupportedTarget` returns undefined for `win32`) and the binary basename check still expects `phantombot`, not `phantombot.exe`. So Windows users who follow these release notes will be pointed at an install path that cannot work. Either make Windows update-supported in this PR, or change the release notes/README to frame the Windows `.exe` assets as manual-download/install only.

kaieriksen (changes requested):
Re-reviewed at head `756a504d`. The implementation checks are clean, but the Windows release/install guidance is still internally inconsistent now that this PR publishes Windows binaries while `phantombot update` still refuses win32 targets. I’m keeping request-changes until the docs/release notes say the same thing the code actually does.

Suggested fix: either (a) keep Windows binaries as manual-download assets for this PR and update README + generated release notes to say Windows users should download/copy the `.exe` manually, not use `phantombot update`; or (b) wire Windows into `detectSupportedTarget`, asset naming, basename `.exe`, and update tests so `phantombot update` actually supports the new Windows assets.

kaieriksen (approved):
Re-reviewed at head `7affad3`. Megan addressed the two blockers I raised: Windows release assets are now matched by updater target detection / .exe asset naming / README guidance, and the build scripts no longer depend on POSIX shell prefixes. I also reviewed the Windows rename-aside update path, stale .old cleanup, in-process restart behavior, and the release/CI wiring.\n\nVerified locally:\n- focused update/release/platform/command tests: `177 pass`\n- `bun run typecheck`: clean\n- full suite: `1926 pass, 2 skip`\n- clean build smoke: `bun run build:x64` and `bun run build:win` both pass; Windows output is PE32+ console x86-64\n\nNo remaining blockers from me.

PR #259, opened 2026-07-05
feat(windows): install.ps1 one-liner installer (preview)

andrewagrahamhodges:
## What

Adds a PowerShell installer, `install.ps1`, as the Windows parallel to the existing `install.sh`, plus the `iwr | iex` one-liner on the website and README - both clearly marked **(Preview)**.

Follow-up to the merged Windows port (#257). That PR shipped prebuilt `.exe` binaries and in-place self-update; this gives Windows users a first-class one-line install matching the Linux/macOS experience.

## install.ps1 behaviour (mirrors install.sh)

1. Detects arch from `PROCESSOR_ARCHITECTURE` (AMD64 -> x64, ARM64 -> arm64); refuses otherwise.
2. Fetches the latest release tag (honours `GITHUB_TOKEN`).
3. Downloads `phantombot-<tag>-windows-<arch>.exe` + `SHA256SUMS`.
4. Verifies SHA256 with `Get-FileHash`; refuses on mismatch.
5. Runs `Unblock-File` (the SmartScreen parallel to install.sh's macOS quarantine strip).
6. Installs to `%LOCALAPPDATA%\Programs\phantombot\phantombot.exe` - **per-user, no admin**.
7. Adds that dir to the **user** PATH (and the current session).
8. Launches `phantombot init`.

Overrides: `PHANTOMBOT_INSTALL_DIR`, `PHANTOMBOT_SKIP_TUI` (both mirror install.sh).

## Why that install location

The in-place self-update from #257 renames the running `phantombot.exe` aside to `phantombot.exe.old` in the **same directory**, then drops the new binary into the freed path. So the exe needs its own stable, user-writable folder - which is exactly what `%LOCALAPPDATA%\Programs\phantombot` gives it. The installer path and the updater's expectations line up.

## Linux / macOS

Untouched. No changes to `install.sh`, and the website/README Linux+macOS blocks are unchanged - the Windows content is purely additive.

## Testing note

No `pwsh` on the build box, so the script was self-reviewed against the `install.sh` parallel and hardened for Windows PowerShell 5.1 (TLS 1.2 opt-in, StrictMode-safe property access, WOW64 arch fallback). End-to-end verification needs a Windows machine - same as the rest of the Windows port.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,8 +204,23 @@ self-update) have native Windows implementations.
 
 Every release publishes prebuilt, unsigned Windows binaries -
 `phantombot-<tag>-windows-x64.exe` and `phantombot-<tag>-windows-arm64.exe` -
-alongside the SHA256SUMS file. Download the one for your architecture, verify
-its checksum, and drop it somewhere stable on your PATH:
+alongside the SHA256SUMS file.
+
+**Install (preview) - PowerShell one-liner:**
+
+```powershell
+iwr -useb https://raw.githubusercontent.com/phantomyard/phantombot/main/install.ps1 | iex
+```
+
+This detects your architecture (x64/arm64), downloads the matching binary,
+verifies its SHA256, runs `Unblock-File` so SmartScreen does not flag it,
+installs to `%LOCALAPPDATA%\Programs\phantombot\phantombot.exe` (per-user, no
+admin), adds that dir to your PATH, and launches `phantombot init`. It is the
+Windows parallel to the Linux/macOS `install.sh`. Marked **(preview)** while the
+Windows port settles.
+
+**Or install manually** - download the `.exe` for your architecture, verify its
+checksum, and drop it into the same per-user location:
 
 ```powershell
 # Unblock the downloaded file (SmartScreen marks internet downloads):
--- a/install.ps1
+++ b/install.ps1
@@ -43,7 +43,7 @@ $ProgressPreference = 'SilentlyContinue'  # keep Invoke-WebRequest quiet + fast
 $Repo = 'phantomyard/phantombot'
 
 function Fail([string]$msg) {
-    Write-Error "phantombot: $msg"
+    [Console]::Error.WriteLine("phantombot: $msg")
     exit 1
 }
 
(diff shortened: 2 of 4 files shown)


kaieriksen (dismissed):
Reviewed the Windows installer preview changes. The one-liner resolves the latest release, selects the expected windows-x64/windows-arm64 assets, verifies against SHA256SUMS, installs per-user, updates PATH, and launches init. CI is green, including the Windows job.\n\nLocal validation: focused Windows/release/update/install tests passed; typecheck passed. Full local suite is blocked by this shell's Bun/.env auto-loading a real Pi provider key into Pi harness tests, unrelated to this PR.

lenaparkhodges (dismissed):
Approved. Clean PR — `install.ps1` faithfully mirrors `install.sh` with the right Windows-specific adaptations (TLS 1.2 opt-in for PS 5.1, WOW64 arch fallback via `PROCESSOR_ARCHITEW6432`, `Unblock-File` as the SmartScreen parallel, per-user PATH via HKCU, StrictMode-safe property access on the API response). The second commit's SHA256SUMS byte[] fix is a real catch — GitHub serving `application/octet-stream` would have silently broken every install.

**One minor nit (non-blocking):** The `Fail` function uses `Write-Error` followed by `exit 1`, but under `$ErrorActionPreference = 'Stop'`, `Write-Error` throws a terminating `ActionPreferenceStopException` before `exit 1` is reached. The `exit 1` is effectively dead code — the script terminates via the uncaught exception instead. Functionally fine (the script stops correctly either way), but if you want the clean exit code, swap to `throw "phantombot: $msg"` or use `Write-Host` + `exit 1`. Not a blocker — just a PS gotcha worth cleaning up in a follow-up.

SHA256SUMS regex handles both text and binary mode formats correctly. Temp file cleanup in the `finally` block is solid. PATH manipulation with `-ieq` comparison is correct. The write-probe pattern is a nice touch — better than trusting `mkdir` success.

CI green across the board (test, typecheck, test-windows, CodeQL). Kai's approval stands. Ship it.

lenaparkhodges (approved):
Re-approving on `07b5d04b`. The `Fail` function nit is fixed — `Write-Error` swapped for `[Console]::Error.WriteLine()` + explicit `exit 1`, matching install.sh semantics. Deterministic exit code, no dead code, error message goes to stderr as expected. CI green. Ship it.

PR #260, opened 2026-07-05
feat(windows): use home-relative XDG paths, matching Linux/macOS

andrewagrahamhodges:
## What

On Windows, point config, data and state at the same home-relative XDG layout used on Linux/macOS, instead of collapsing everything under `%LOCALAPPDATA%\phantombot`. A persona's on-disk tree is now identical across all three OSes:

- config -> `~/.config/phantombot`
- data -> `~/.local/share/phantombot` (personas, memory db, logs)
- state -> `~/.local/state/phantombot`

On Windows `~` resolves to `%USERPROFILE%`, e.g. `C:\Users\<you>\.local\share\phantombot`.

The `XDG_*` env vars still take precedence on every platform, so an explicit override remains the escape hatch - consistent with Linux behaviour.

## Why

Portability. Andrew found a persona living at `C:\Users\...\.local\share\phantombot\personas\phantom` (created when a shell exported `XDG_DATA_HOME`) alongside a `%LOCALAPPDATA%` one, and prefers the Linux-matching layout so the tree is the same everywhere.

## Changes

- `src/config.ts` - drop the `win32` branch from `xdgConfigHome`/`xdgDataHome`/`xdgStateHome`; remove the now-unused `windowsLocalAppData` helper.
- `src/lib/platform.ts` - fix the Windows "view logs" hint, which still pointed at `%LOCALAPPDATA%\phantombot\logs` (logs resolve via `xdgDataHome`, so they now live under `~/.local/share`).
- `src/lib/taskScheduler.ts`, `src/cli/heartbeat.ts` - refresh stale `%LOCALAPPDATA%` comments.
- `README.md` - update the data-location and logs sections.
- `tests/windows-port.test.ts` - assert the home-relative layout on Windows.

## Safety

Linux/macOS path resolution is byte-for-byte unchanged (no POSIX branch was touched). Full test suite green locally (1926 pass, 0 fail), typecheck clean. Will re-verify the installer end to end on the Windows VM.

## Note

The binary install location is unaffected: `install.ps1` still installs the exe to `%LOCALAPPDATA%\Programs\phantombot`, which is the OS-idiomatic place for programs and is separate from the data tree.


src/lib/platform.ts
```diff
@@ -165,7 +165,7 @@ export function logsCommand(): string {
     case "darwin":
       return `tail -f ~/Library/Logs/phantombot/dev.phantombot.phantombot.{out,err}.log`;
     case "windows":
-      return `powershell -Command "Get-Content -Wait -Tail 50 \\"$env:LOCALAPPDATA\\phantombot\\logs\\phantombot.out.log\\""`;
+      return `powershell -Command "Get-Content -Wait -Tail 50 \\"$env:USERPROFILE\\.local\\share\\phantombot\\logs\\phantombot.out.log\\""`;
```
kaieriksen: This still hard-codes the default data location. The actual Task Scheduler logs now come from `taskLogPaths()` -> `xdgDataHome()`, and the README explicitly says `XDG_DATA_HOME` relocates the data/log tree. With `XDG_DATA_HOME` set, `phantombot install` / `phantombot run` will print a `view logs:` command for `%USERPROFILE%\.local\share...` while the task is writing under the override, so the hint sends Windows users to a non-existent or stale log. Please build this from the same resolver used by the scheduler, or at least branch on `$env:XDG_DATA_HOME` in the PowerShell command.

kaieriksen (changes requested):
Thanks. The path resolver change is coherent, but one Windows log hint still hard-codes the default data root and will point users at the wrong file when XDG_DATA_HOME is set.

src/lib/taskScheduler.ts
```diff
@@ -40,7 +40,7 @@
  * Logging (WinSW-inspired, minus the SCM): Task Scheduler does not capture a
  * process's stdout/stderr, so the action is run through `cmd /c` with the
  * streams redirected (append) to per-task .out.log / .err.log under
- * %LOCALAPPDATA%\phantombot\logs — the same out/err split launchd writes to
+ * ~/.local/share/phantombot/logs - the same out/err split launchd writes to
```
kaieriksen: This PR is now conflicting with current `main` here. `main` changed this Windows Task Scheduler path to run through the hidden `wscript` launcher, while this branch is still based on the older `cmd /c` redirect shape. Please rebase/merge `main` and keep the launcher behavior intact while applying the XDG log path wording/resolver changes.

kaieriksen (changes requested):
The XDG_DATA_HOME log hint issue is fixed on the branch head, and the focused tests/typecheck pass locally. However this PR is currently unmergeable against main (`mergeStateStatus=DIRTY`) because main added the hidden wscript Task Scheduler launcher in `src/lib/taskScheduler.ts` and `tests/lib-taskScheduler.test.ts`. Please rebase/merge main and preserve the wscript launcher behavior while applying the XDG path changes.

kaieriksen (approved):
Re-reviewed latest head. The previous merge-conflict blocker is resolved, the Windows log hint now derives from the same taskLogPaths()/xdgDataHome() resolver as Task Scheduler, and the branch is mergeable with green CI.\n\nLocal validation: focused Windows path/scheduler/platform tests passed, typecheck passed, and build passed. Full local suite still hits the known dev-vault Pi API key contamination in unrelated Pi harness tests; CI's clean environment is green.

PR #261, opened 2026-07-05
windows tasks: run via hidden wscript launcher (no console flash)

andrewagrahamhodges:
## Problem

On Windows, the Task Scheduler backend registered each task with `cmd.exe` as the action Command. Because the tasks run in the user's interactive session, every fire pops a visible `cmd.exe` console window. The `tick` task fires every 60 seconds, so a console window flashes on the desktop roughly once a minute. Andrew hit this on the test box.

## Fix

Route the task action through a tiny generated VBScript launcher run by `wscript.exe` (which has no console of its own) instead of `cmd.exe` directly:

- `wscript.exe <phantombot-launch.vbs>` re-spawns the real `cmd /c "<exe>" <sub> 1>>"<out>" 2>>"<err>"` line **hidden** (`WshShell.Run cmd, 0, True`).
- `waitOnReturn = True` means the launcher blocks until the child exits, so Task Scheduler still sees the always-on task as **Running** and the `MultipleInstancesPolicy=IgnoreNew` keep-alive is preserved (no duplicate daemons).
- The **binary path stays a visible task argument** (not buried inside the `.vbs`), so `ensureTasksCurrent`'s drift self-heal still detects a moved binary from the registered task XML - the existing self-heal tests pass unchanged.
- `install` / `uninstall` / self-heal write and remove the shared launcher alongside the logs dir.

The launcher is a single shared, ASCII-only script (byte-identical read as ANSI or UTF-8), generated from a template so a change lands on the next install or self-heal.

## Scope

Windows-only. `src/lib/taskScheduler.ts` and its tests. Linux/macOS (systemd/launchd) are untouched.

## Tests

- Replaced the `buildCmdArguments` unit with `buildLauncherArguments` (quoted-token layout) + a `LAUNCHER_VBS` content test (hidden+wait, redirection rebuild, ASCII/CRLF).
- Updated the action test: Command is now `wscript.exe`, not `cmd.exe`; redirection operators no longer appear in the task XML.
- Added install-writes-launcher and uninstall-removes-launcher coverage.
- Full suite green locally (1929 pass, 0 fail); typecheck clean.

## Verification

Not yet run end-to-end on the Windows VM - that follows on a clean-slate reinstall, same as the rest of the port. Will confirm no window appears when `tick` fires before requesting review.


lenaparkhodges (dismissed):
Approved. Clean solution to the console-flash problem — routing through `wscript.exe` with a generated VBS launcher is the right approach for Task Scheduler.

**What's good:**
- The VBS faithfully reconstructs the exact `cmd /c ""<exe>" <args> 1>>"<out>" 2>>"<err>""` line from the old `buildCmdArguments`, just runs it hidden (`windowStyle 0`) and blocking (`waitOnReturn True`). The `IgnoreNew` keep-alive is preserved.
- ASCII-only + CRLF — byte-identical as ANSI or UTF-8, with a test enforcing it. Good defensive choice for Windows scripting.
- Binary path stays as a visible task argument (not buried in the VBS), so `xmlReferencesBin` drift detection still works. Self-heal tests pass unchanged.
- Install/uninstall/self-heal all write/remove the launcher. Idempotent overwrite means a template change lands on next heal.
- No VBScript injection risk — `WScript.Arguments` are parsed string values, so even if a path contained `&` it's treated as a literal, not a concatenation operator.

**Minor nit (not a blocker):**
Consider adding `//B` to the wscript invocation (either in the `<Arguments>` or as a flag before the script path). Without it, if the VBS ever errors at runtime, wscript pops a GUI dialog — silent on the logs, confusing to debug. With `//B`, errors go nowhere (no dialog), which is the right behavior for an unattended task. Not urgent since the script is trivial and tested, but worth a follow-up.

The PR body notes end-to-end verification on the Windows VM is still pending — that's a manual step, not a code blocker. Ship it.

lenaparkhodges (approved):
Re-approved on `2fa382b2`. The `//B` batch mode flag is exactly the nit I raised — wscript runtime errors now go to stderr instead of popping a GUI dialog. Tests updated to assert the flag. Ship it.

PR #262, opened 2026-07-05
fix(windows): detect harnesses on PATH and stop conjuring a default persona

andrewagrahamhodges:
Two Windows-install correctness bugs found while testing the fresh install on the VM.

## 1. Harnesses show `[NOT FOUND]` in the TUI (and doctor)

`src/lib/harnessAvailability.ts` was POSIX-only, so on Windows every harness came back unresolved:

- split `PATH` on `:` - Windows uses `;`, and drive letters like `C:\` contain `:`, so the whole PATH gets shredded
- appended no executable extension - the shipped CLIs are `claude.cmd` / `pi.cmd` / `codex.exe`, never a bare `claude`
- gated on the Unix execute bit (`X_OK`) - not modeled on Windows, can spuriously reject a real `.cmd`
- treated only `/`-prefixed paths as absolute - Windows absolute paths start with a drive letter

**Fix:** split on `path.delimiter`, iterate `PATHEXT` (bare name first, then each suffix), use `path.isAbsolute`, and skip the `X_OK` check on win32. Also add `%APPDATA%\npm` to the search path. This mirrors the win32 handling already present in `installVscode.ts`. POSIX behaviour is unchanged.

## 2. A stray `phantom` persona is created every time

`persona new test` on a fresh box left you with both `phantom` and `test`. The startup vault bootstrap (`src/index.ts`) loads the default persona's vault into env, and that read path (`openPersonaVault` -> `getOrCreatePersonaIdentity` + `openVaultWithSecret`) provisioned the persona dir, `identity.json` and `vault.sqlite` as a **side effect of a read**. With `defaultPersona` falling back to the literal `"phantom"`, any non-read-only command conjured it.

**Fix (open-existing-only):** `readAllVaultValues` now returns `null` (treated as "no vault") when the persona dir doesn't already exist, so a pure env-load never provisions. Real vault reads/writes and `persona new` are untouched.

## Tests

- `executableExtensions` unit coverage (PATHEXT parsing, defaults, whitespace) - runs on Linux CI
- Windows-gated end-to-end `whichBinary` test: bare `claude` resolves to `claude.cmd` across a `;`-joined, drive-letter PATH - runs on the test-windows runner
- Regression test: loading a never-provisioned persona creates nothing on disk

Full suite green (1935 pass), typecheck clean. Windows-only behaviour; Linux/macOS untouched.

### DIFF
--- a/src/lib/harnessAvailability.ts
+++ b/src/lib/harnessAvailability.ts
@@ -1,6 +1,6 @@
 import { access, constants, readdir, stat } from "node:fs/promises";
 import { homedir } from "node:os";
-import { join } from "node:path";
+import { delimiter, isAbsolute, join } from "node:path";
 import type { Config } from "../config.ts";
 import { log } from "./logger.ts";
 import { saveHarnessBins } from "../state.ts";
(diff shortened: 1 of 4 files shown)


kaieriksen (approved):
Reviewed the harness discovery and vault persona fallback changes. The Windows PATH/PATHEXT handling is covered by the new focused tests and CI's Windows job, and the vault read path no longer creates a missing default persona while preserving existing per-persona reconciliation behavior.\n\nValidation: focused harness/vault tests passed; typecheck passed; build passed. Full local suite hit my known real-vault Pi key contamination under the default config, but the affected Pi routing subset passes when run with an isolated PHANTOMBOT_CONFIG, and CI is green.

PR #263, opened 2026-07-05
fix(windows): pass harness prompts via temp files to beat the argv length limit

andrewagrahamhodges:
## Draft for review

Opened as a **draft** at Andrew's request. Not to be merged until it has been reviewed and verified on a fresh test persona on the Windows VM (avoiding a collision with the live megan persona there).

## Problem

On Windows the entire child command line is capped at ~8,191 characters (CreateProcess `lpCommandLine`). On a real persona turn both harnesses overflow it, fail to spawn with `The command line is too long.`, and exit 1 with no reply. Symptom: the bot shows a typing indicator, then goes silent.

Root cause differs per harness:

- **pi** carries BOTH the system prompt (`--system-prompt <text>`) and the full rendered payload (the positional argument) on argv. On the reporter's box the payload was ~18 KB.
- **claude** already streams the conversation via stdin, but still passes the whole persona/memory system prompt on argv via `--system-prompt`. A large `BOOT.md` alone overflows the limit, which is why claude failed the same way.

## Fix (Windows only)

Spill the oversized argv payloads to temp files and hand the child a short file reference instead. POSIX (`ARG_MAX` ~2 MB) keeps the existing inline argv path byte for byte.

- **pi:** write the system prompt and the payload to temp files; pass `--system-prompt <file>` and the positional `@<file>`.
- **claude:** write the system prompt to a temp file; pass `--system-prompt-file <file>`.

Temp files are created per invocation in a private temp dir and removed in a `finally`.

## Why not stdin (xiaomi-mimo's Option A)

pi ignores stdin in `--print` mode (confirmed by its own `--help` and our harness code), so piping the payload would have failed silently. The temp-file approach is harness-native for both.

## Empirically verified before writing the code

Against pi 0.80.3 and Claude Code on a real box:

- pi `--system-prompt <path>` reads the file contents as the system prompt (pirate-rule test obeyed).
- pi `@<path>` positional presents the file contents as the message (`@file` containing "17 plus 25" answered `42`).
- claude `--system-prompt-file <path>` applies the file as the system prompt (pirate-rule test obeyed).

## The one thing to eyeball on the VM

Confirm pi's `@file` presents the payload as a clean user message (no attachment wrapper that shifts model behaviour). It behaved correctly in local testing; worth a direct look on the target box with a fresh persona.

## Tests

- New `tests/lib-harnessArgvFiles.test.ts`: the win32-only decision gate and the temp-dir write/cleanup/idempotency helper.
- pi and claude harness tests with an injected `platform`: Windows spills prompts to files and cleans them up; POSIX keeps prompts inline. These run on the Linux CI runner (no Windows-gating needed) via the fixtures' argv-echo mode.

Full suite green (1945 pass, 3 skip, 0 fail), typecheck clean.

## Note for the reviewer

The `invoke()` bodies are wrapped in `try { ... } finally { temp?.cleanup() }` but the inner code keeps its original indentation, so the diff stays focused on the real change rather than a whitespace reflow. Happy to re-indent if preferred.

### DIFF
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -58,6 +58,11 @@ import {
 } from "../lib/harnessRunner.ts";
 import { log } from "../lib/logger.ts";
 import { spawnInNewSession } from "../lib/processGroup.ts";
+import {
+  argvNeedsTempFiles,
+  createHarnessTempDir,
+  type HarnessTempDir,
+} from "../lib/harnessArgvFiles.ts";
 
 export interface ClaudeHarnessConfig {
   /** Path to the `claude` CLI binary. Default: "claude" (looked up in PATH). */
@@ -71,7 +76,13 @@ export interface ClaudeHarnessConfig {
 export class ClaudeHarness implements Harness {
   readonly id = "claude";
 
-  constructor(private readonly config: ClaudeHarnessConfig) {}
+  constructor(
+    private readonly config: ClaudeHarnessConfig,
+    // Injectable so the Windows argv-length branch below is testable on a
+    // POSIX CI runner. Prod callers pass only the config and get the real
+    // platform.
+    private readonly platform: NodeJS.Platform = process.platform,
+  ) {}
 
   async available(): Promise<boolean> {
     try {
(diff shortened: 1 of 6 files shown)


kaieriksen (approved):
Reviewed the Windows argv-length workaround for Pi and Claude. The temp-file path is gated to win32, POSIX argv behavior stays unchanged, files are cleaned up after child exit, and the new tests cover the key spawn arguments and cleanup behavior.\n\nValidation: focused harness/temp-file tests passed under isolated config, typecheck passed, build passed, and CI is green. Full local suite has an unrelated parallel isolation flake in cli-import-persona when run with the whole repo; that file passes in isolation.

PR #264, opened 2026-07-06
fix(windows): make the memory index portable by storing posix paths

andrewagrahamhodges:
## Problem

Copying a persona's `memory/`+`kb/` from Linux to a Windows install, then reusing an existing memory-index, results in the persona coming up with **no memory**.

Root cause: `walkMarkdown` in `src/lib/memoryIndex.ts` records each file's path with `relative()`, which emits the **platform** separator. So the index is keyed with backslashes on Windows and forward slashes on Linux. When the walk paths and the recorded paths use different separators, the refresh sees **zero** matching rows and:

- deletes every recorded row and re-embeds from scratch (losing vector embeddings), or
- if the source files aren't where it looks, deletes every row and leaves the index empty (the observed 'no memory' outcome).

The derived index is therefore **not portable** across OSes, even though the source `.md` files are.

## Fix

Normalise the stored path to posix (forward slashes) at the single source, `walk()`. Now the index keys files identically on every OS, so a persona's memory survives a Linux to/from Windows move without churn.

This is contained because downstream consumers already tolerate forward slashes:
- reads use `path.join(personaDir, f.path)` which accepts `/` on Windows;
- `resolveMdLink` and `buildNameIndex` already normalise separators.

## Tests

- New `walkMarkdown` test asserts nested paths are recorded posix-style and never contain a backslash. Trivially true on Linux CI; on the Windows CI runner it fails pre-fix and passes post-fix.
- Full suite green locally (1946 pass, 0 fail); typecheck clean.

## Migration note for existing installs

The recommended cutover is unchanged and still the right advice: copy the persona's source (`memory/`, `kb/`), **do not** copy the derived `memory-index/`, and rebuild it natively with `phantombot memory index --rebuild`. This PR additionally makes an already-built index **portable**, so a copied index no longer self-wipes on the first search.

## Not deployed

Review-only draft, as agreed. No VM touched.

### DIFF
--- a/src/lib/memoryIndex.ts
+++ b/src/lib/memoryIndex.ts
@@ -1054,7 +1054,18 @@ function walk(
     if (!entry.isFile() || !entry.name.endsWith(".md")) continue;
     const st = statSync(full);
     out.push({
-      path: relative(dir.startsWith(root) ? dirname(root) : root, full),
+      // Store paths posix-style (forward slashes) on every OS. relative()
+      // emits the platform separator, so Windows would record backslash
+      // paths. That makes the index non-portable: move a persona's memory
+      // between Linux and Windows and the first walk finds zero matching
+      // rows, so every note is deleted and re-embedded (or, if the source
+      // files are missing, the whole index is wiped). Normalising here keys
+      // the index the same way everywhere. Downstream reads use join(), and
+      // resolveMdLink/buildNameIndex already normalise, so this is safe.
+      path: relative(dir.startsWith(root) ? dirname(root) : root, full).replace(
+        /\\/g,
+        "/",
+      ),
       scope,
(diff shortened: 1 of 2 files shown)


kaieriksen (dismissed):
Approved. I checked the path-key normalization in walkMarkdown, downstream reads/link resolution, and the portability regression test. Focused memory-index tests, typecheck, build, and CI are green. Full local suite only hit the known Pi env contamination; the affected Pi routing subset passes with an isolated home/config.

kaieriksen (approved):
Approved. Re-reviewed the Windows-only path separator gate after the latest commit. The change preserves POSIX filenames containing literal backslashes while still storing Windows-discovered memory/kb paths with forward slashes, and downstream joins/link resolution remain compatible. Focused memory-index tests, typecheck, build, and CI are green.

PR #265, opened 2026-07-06
feat(cli): OS-agnostic start/stop/restart/logs service commands

andrewagrahamhodges:
## What

Adds the four service-lifecycle verbs requested — the daemon counterparts to `phantombot run`:

```
phantombot start      # start the installed background service
phantombot stop       # stop it and keep it down
phantombot restart    # bounce it
phantombot logs       # tail service logs (--no-follow / --lines N)
```

They're **OS-agnostic**: the command layer never touches systemctl/launchctl/schtasks. It goes through the existing platform-router `ServiceControl` (`src/lib/platform.ts`), which already dispatched `restart`/`isActive` to systemd (Linux), launchd (macOS), and Task Scheduler (Windows). This PR extends that abstraction with `start()`/`stop()` and adds a `logs` tailer.

## The interesting part — `stop` that actually stays stopped

A naive `stop` is wrong on two of the three platforms because the supervisor relaunches a merely-killed process. Handled per-backend:

| Verb | Linux (systemd) | macOS (launchd) | Windows (Task Scheduler) |
|---|---|---|---|
| `start` | `start` | `bootstrap` / `kickstart` | `/Change /ENABLE` + `/Run` |
| `stop` | `stop` | `bootout` (unload) | `/Change /DISABLE` + `/End` |
| `restart` | `restart` (unchanged) | `kickstart -k` (unchanged) | `/End` + `/Run` (+ `/ENABLE`) |

- **systemd** main unit is `Restart=on-failure`, so a clean `stop` stays down — nothing extra.
- **launchd** is `KeepAlive=true`, so `stop` must `bootout` (unload) and `start` `bootstrap`/`kickstart` depending on load state (sidesteps bootstrap's EBUSY-when-loaded).
- **Windows** has a 1-minute keep-alive `TimeTrigger`, so `stop` disables the task before `/End`, and `start` re-enables before `/Run`.

## `logs`

New `logsSpec()` builds the platform tailer argv; the command spawns it with inherited stdio:
- linux → `journalctl --user -u phantombot [-n N] [-f]`
- darwin → `tail [-f] -n N <out.log> <err.log>`
- windows → `powershell Get-Content [-Wait] -Tail N <out.log>`

Follows by default; `--no-follow` prints the last N and exits. Falls back to a copy-pasteable hint if the tailer isn't on PATH.

## Scriptability

`start`/`stop`/`restart` exit `0` on success, `1` on failure, and print a manual-command + status hint when the backend refuses. These are the **external** controls (run from a terminal); the in-chat `/restart` and `/update` still self-restart from inside the running service (`selfRestart`), unchanged.

## Tests & docs

- New `tests/lib-serviceLifecycle.test.ts` — all three verbs (success/failure, output text, exit codes) via an injected fake `ServiceControl`.
- `tests/lib-platform.test.ts` — `logsSpec` shape (follow default + `--no-follow`) and `startCommand`/`stopCommand` hints, platform-guarded like the existing hint tests.
- All existing `ServiceControl` fakes updated for the two new methods; dispatcher "planned subcommands" test updated.
- README gets a **Service lifecycle** section (command table + the keep-alive rationale); `docs/architecture.md` module map gains the platform/lifecycle rows.

**Full suite green** for everything this touches; typecheck clean. (Two pre-existing `harnesses-pi` failures are unrelated — they fail on clean `main` too, caused by a real `PHANTOMBOT_PI_API_KEY` in the local dev env leaking into the harness subprocess; CI has no real vault so they pass there.)


lenaparkhodges (approved):
Clean, well-structured PR. The OS-agnostic abstraction through `ServiceControl` is the right pattern — the command layer never touches systemctl/launchctl/schtasks directly.

**Verification:** 1954 tests pass, tsc clean, all CI green (including `test-windows`). The 2 PiHarness failures are pre-existing on `main` (real `PHANTOMBOT_PI_API_KEY` leaking into the subprocess; CI has no vault so they pass there).

**Strengths:**
- `runLifecycleAction` is pure and injectable — tests use a fake `ServiceControl` with zero supervisor dependency.
- Per-OS keep-alive handling is correct: launchd `bootout`/`bootstrap-or-kickstart` split avoids EBUSY; Windows `/Disable` before `/End` prevents the 60s relaunch; systemd `Restart=on-failure` means clean `stop` just stays down.
- `bootout` on a not-loaded agent treated as success — avoids spurious errors on idempotent `stop`.
- `logs` gracefully degrades to a copy-pasteable hint when the tailer binary isn't on PATH.
- All existing `ServiceControl` fakes updated for the two new methods.

**Two non-blocking notes:**
1. **Windows `stop()` return value only checks the `/Disable` result** — if `/End` fails (unlikely, harmless on stopped task), the method still returns `ok: true`. Practically fine since `/End` on a stopped task is a no-op, but noting it.
2. **`logsSpec` on Windows only tails `out.log`** — `err.log` is referenced in `logsCommand()` but not in the live tail. Documented in the code comment; just worth knowing.

No blockers. Ready for merge.

src/lib/serviceLifecycle.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export async function runLifecycleAction(
+  opts: LifecycleOptions,
+): Promise<number> {
+  const svc = opts.serviceControl ?? defaultServiceControl();
+  const out = opts.out ?? process.stdout;
+  const err = opts.err ?? process.stderr;
+  const { done, hint } = actionText(opts.action);
+
+  const r = await svc[opts.action]();
```
kaieriksen, line 67: This is the right boundary for the new commands: the CLI layer stays OS-agnostic and only calls the selected ServiceControl verb, so systemd/launchd/Task Scheduler behavior remains isolated behind the platform router.

kaieriksen (approved):
Reviewed the new service lifecycle commands and backend routing. The implementation keeps the CLI wrappers thin, uses the existing ServiceControl abstraction, and the platform-specific keep-alive behavior is covered by focused tests.\n\nVerified locally:\n- focused lifecycle/platform/backend suite: 71 pass\n- bun run typecheck: clean\n- sterile full suite: 1956 pass, 3 skip

Issue #258, opened 2026-07-04
P2P transport node: embed Hyperswarm DHT so phantombot bridges messages node-to-node (no relay hot path)


robertclawson:
# P2P transport node: embed Hyperswarm DHT so phantombot bridges messages node-to-node (no relay in the hot path)

> **Companion issue:** phantomyard/phantomchat#61 — the PWA-side tiered transport that consumes this node. Read them together.

## Problem

Today every phantomchat message round-trips through a Nostr relay. That relay is:

- **The latency floor** — every message pays a relay hop, even between two peers sitting next to each other.
- **A dependency we don't control** — relay uptime/quality gates the whole chat.

A browser PWA can never escape this on its own: browsers have no raw UDP and no DHT, so the richest P2P primitive they get is WebRTC. But **phantombot runs on bun/Node**, which *can* run the full Keet/Holepunch stack — raw UDP, the Hyperswarm DHT, holepunching. So the move is: **phantombot becomes the user's personal P2P node, and the PWA rides it.** Two phantombots then talk fully P2P over the DHT — zero relays in the message path.

## Hard constraints (locked with Andrew)

- **No hosted infrastructure of ours.** No VPS, no signaling server, no TURN we run. The *only* infra in this design is the static PWA on GitHub Pages and the Nostr relays we select.
- **phantombot nodes are user-owned** — a phantombot on the user's own PC or LAN. Not a central hosted node.
- **Push notifications are out of scope** for this issue (tracked separately, later).

## Role of this node

```mermaid
flowchart LR
  subgraph User["User's machine / LAN"]
    PWA["phantomchat PWA<br/>(browser, GitHub Pages)"]
    PB["phantombot node<br/>(bun/Node)"]
    PWA -->|"ws://localhost<br/>or LAN WebRTC"| PB
  end
  subgraph Peer["Remote peer's machine"]
    PB2["remote phantombot node"]
  end
  PB <-->|"Hyperswarm DHT<br/>(direct, encrypted,<br/>NO relay)"| PB2
  PB -. "signaling / fallback only" .-> Relays[("Nostr relays")]
  PB2 -. "signaling / fallback only" .-> Relays
```

phantombot gains a **P2P transport module** that:

1. **Joins the Hyperswarm DHT** — discovers peers by topic/pubkey via Hyperswarm's own bootstrap (this does **not** need Nostr), holepunches, and opens direct encrypted node-to-node channels (Hypercore/`@hyperswarm/secret-stream`).
[...]

PR #266, opened 2026-07-06
feat(p2p): relay-free P2P transport node — werift + Nostr signaling (#258)

andrewagrahamhodges:
## What & why

Closes the phantombot half of **#258** (companion: phantomchat #61/#62). Today every PhantomChat message round-trips a public Nostr relay — a latency floor and a dependency we don't control. This makes phantombot the user's **personal P2P node** so two nodes talk **directly node-to-node over an encrypted WebRTC data channel**, with relays demoted to the WebRTC handshake (signaling) + a fallback floor. No relay in the message hot path; no infrastructure of ours.

## The werift decision (resolving #258's blocker)

Robbie's blocker on #258 was that the intended Holepunch stack **doesn't run under Bun**: `new Hyperswarm()` panics (`uv_interface_addresses`) and `node-datachannel` can't be bundled by `bun build --compile`. After a hands-on spike we confirmed **werift** (pure-TypeScript WebRTC) survives compilation into the shipped single binary — a full ICE→DTLS→SCTP data-channel round-trip works *inside a compiled Bun binary*. So this PR uses **werift + Nostr-for-signaling**, exactly as agreed:

- **werift** for the peer-to-peer data channels — no native addon, no sidecar, one binary.
- **Nostr** carries only the SDP offer/answer + ICE candidates, NIP-44-encrypted on a **dedicated ephemeral kind (21050)** so it never collides with the chat 1059 gift-wrap plane. Message contents never touch it — they stay end-to-end sealed.
- **Public STUN** for NAT traversal (reflexive-only; it never relays, so there's still zero infra of ours).

## Contract (matches the merged PWA side)

The node is a **dumb, encrypted-wrap relay**. The PWA ships `["EVENT", <gift-wrap>]` (a standard Nostr relay frame) over `ws://localhost:47100`; the node reads the recipient off the wrap's `p`-tag, forwards the **sealed** wrap to that peer's data channel, and broadcasts inbound wraps back to the local PWA. It never holds a key for the contents. Port and framing match phantomchat #62's `local-ws-transport.ts` verbatim.

## Zero regression — off by default

`config.p2p.enabled` defaults **false**. An unconfigured or existing install opens no port, advertises no capability, runs no signaling subscription — byte-for-byte unchanged. Enable via `[p2p]` in config.toml or `PHANTOMBOT_P2P_ENABLED=1`.

## What's here

| Module (`src/p2p/`) | Role |
|---|---|
| `frame.ts` | Wire contract: parse/build `["EVENT", wrap]`, recipient from p-tag (pure) |
| `signaling.ts` | SDP/ICE over Nostr (NIP-44, ephemeral kind 21050) on the `RelayPool` seam |
| `peerConnection.ts` | One werift `RTCPeerConnection` + data channel per peer |
| `localBridge.ts` | Bun-native `ws://localhost:47100`, loopback-only |
| `node.ts` | Routing brain: deterministic initiator + `hello` nudge (glare-free), per-peer outbox |
| `capability.ts` | NIP-78 capability advertisement (the format the PWA companion will ingest) |
| `index.ts` | Daemon glue (`buildP2PNode` / `runP2PNode`), wired into `run` |

Plus `config.p2p` (enabled/port/stun, env + TOML), `phantombot p2p status`, README + architecture docs.

## The bug worth flagging in review

Readiness is gated on the **data channel `open` event, not** the ICE/DTLS `connectionStateChange === "connected"`. The transport connects slightly *before* the SCTP channel opens; gating on transport-connected flushed the first buffered frame into a not-yet-open channel and silently dropped it. Caught by the end-to-end test; fixed and commented. (This is exactly the kind of thing the relay floor would have masked in the field.)

## Tests

- **Real werift round-trip** — two `PeerConnection` instances do a genuine ICE→DTLS→data-channel handshake on loopback and exchange frames both ways (~0.4s).
- **End-to-end two-node** — two full nodes exchange a wrap with **no relay in the data path** (in-memory signaling bus + real werift).
- Routing brain via injected fakes; the Bun ws bridge via a **real loopback WebSocket client**; pure frame/capability/signaling (real NIP-44 crypto); config precedence + clamping.
- 38 p2p tests + config coverage. **Full suite green** (the 2 unrelated `harnesses-pi` failures are the known local env-leak — a real `PHANTOMBOT_PI_API_KEY` in the dev shell bleeds into the harness subprocess; this branch touches nothing in that path). Typecheck clean.

## Scope / follow-ups (honest)

- **PWA-side capability ingestion** is the companion phantomchat change. The PWA already *sends* over `ws://localhost` for a P2P-capable peer, but it doesn't yet *populate* its capability registry from a node's advertisement — so until that lands, the node runs and advertises but the PWA ladder stays on relays. This PR defines that advertisement format (`capability.ts`).
- **Single loopback port** ⇒ on a multi-persona host only the first PhantomChat persona hosts the node.
- **Remote symmetric-NAT** still falls back to the relay (by design — that's our "TURN" without running one).

### DIFF
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -36,10 +36,16 @@ import { cleanupStaleUpdateArtifacts } from "../lib/binaryUpdate.ts";
 import { npubEncode } from "../lib/nostrIdentity.ts";
 import {
   type Config,
+  DEFAULT_P2P,
   loadConfig,
   personaDir,
   type TelegramAccount,
 } from "../config.ts";
+import {
+  advertiseP2PCapability,
+  buildP2PNode,
+  runP2PNode,
+} from "../p2p/index.ts";
 import { buildHarnessChain } from "../harnesses/buildChain.ts";
 import {
   resolveHarnessBinsForConfig,
@@ -559,6 +565,12 @@ export async function runRun(input: RunInput = {}): Promise<number> {
         );
       }
 
+      // The relay-free P2P node (phantombot#258) binds a single loopback port
+      // (config.p2p.port), so exactly ONE persona hosts it — the first
+      // phantomchat persona. Disabled by default; see config.p2p.
+      const p2pSettings = config.p2p ?? DEFAULT_P2P;
+      let p2pNodeStarted = false;
+
       for (const spec of phantomchatPersonas) {
         const { identity, allowedHex, tofu, groupBots } = spec.config;
 
@@ -765,6 +777,48 @@ export async function runRun(input: RunInput = {}): Promise<number> {
               error: (e as Error).message,
             });
           });
+        }
+
+        // Relay-free P2P transport node. Rides THIS persona's identity, relays
+        // and relay pool: a loopback ws bridge for the same-machine PhantomChat
+        // PWA plus werift WebRTC channels to peer nodes, with Nostr carrying only
+        // the WebRTC handshake. Started for the first phantomchat persona only
+        // (single loopback port). Inert unless config.p2p.enabled.
+        if (p2pSettings.enabled && !p2pNodeStarted) {
+          p2pNodeStarted = true;
+          try {
+            const p2pNode = buildP2PNode({
+              secretKey: identity.secretKey,
+              publicKeyHex: identity.publicKeyHex,
+              relays,
+              pool: pool as unknown as Parameters<typeof buildP2PNode>[0]["pool"],
+              settings: p2pSettings,
+            });
+            advertiseP2PCapability({
+              secretKey: identity.secretKey,
+              publicKeyHex: identity.publicKeyHex,
+              relays,
+              pool: pool as unknown as Parameters<typeof buildP2PNode>[0]["pool"],
+              settings: p2pSettings,
+            });
+            out.write(
+              `  [p2p:${spec.persona}] node on ws://127.0.0.1:${p2pSettings.port}, ` +
+                `${p2pSettings.stunServers.length} STUN server(s)\n`,
+            );
+            tasks.push(runP2PNode(p2pNode, ac.signal));
+          } catch (e) {
+            p2pNodeStarted = false;
+            log.warn(`p2p[${spec.persona}]: node failed to start`, {
+              error: (e as Error).message,
+            });
+            err.write(
+              `warning: p2p node failed to start (port ${p2pSettings.port} in use?) — chat still works over relays\n`,
+            );
+          }
+        } else if (p2pSettings.enabled && p2pNodeStarted) {
+          out.write(
+            `  [p2p:${spec.persona}] skipped — a P2P node is already bound to port ${p2pSettings.port}\n`,
+          );
         }
       }
     }
--- a/src/cli/run.ts
+++ b/src/cli/run.ts
@@ -44,7 +44,7 @@ import {
 import {
   advertiseP2PCapability,
   buildP2PNode,
-  runP2PNode,
+  startP2PNode,
 } from "../p2p/index.ts";
 import { buildHarnessChain } from "../harnesses/buildChain.ts";
 import {
@@ -785,35 +785,29 @@ export async function runRun(input: RunInput = {}): Promise<number> {
         // the WebRTC handshake. Started for the first phantomchat persona only
         // (single loopback port). Inert unless config.p2p.enabled.
         if (p2pSettings.enabled && !p2pNodeStarted) {
-          p2pNodeStarted = true;
-          try {
-            const p2pNode = buildP2PNode({
-              secretKey: identity.secretKey,
-              publicKeyHex: identity.publicKeyHex,
-              relays,
-              pool: pool as unknown as Parameters<typeof buildP2PNode>[0]["pool"],
-              settings: p2pSettings,
-            });
-            advertiseP2PCapability({
-              secretKey: identity.secretKey,
-              publicKeyHex: identity.publicKeyHex,
-              relays,
-              pool: pool as unknown as Parameters<typeof buildP2PNode>[0]["pool"],
-              settings: p2pSettings,
-            });
-            out.write(
-              `  [p2p:${spec.persona}] node on ws://127.0.0.1:${p2pSettings.port}, ` +
-                `${p2pSettings.stunServers.length} STUN server(s)\n`,
-            );
-            tasks.push(runP2PNode(p2pNode, ac.signal));
-          } catch (e) {
-            p2pNodeStarted = false;
-            log.warn(`p2p[${spec.persona}]: node failed to start`, {
-              error: (e as Error).message,
-            });
-            err.write(
-              `warning: p2p node failed to start (port ${p2pSettings.port} in use?) — chat still works over relays\n`,
-            );
+          const p2pDeps = {
+            secretKey: identity.secretKey,
+            publicKeyHex: identity.publicKeyHex,
+            relays,
+            pool: pool as unknown as Parameters<typeof buildP2PNode>[0]["pool"],
+            settings: p2pSettings,
+          };
+          const p2pNode = buildP2PNode(p2pDeps);
+          // Start SYNCHRONOUSLY and contain any port-conflict throw inside the
+          // helper, so a failed P2P bring-up degrades to relays instead of
+          // rejecting a pushed task and aborting the whole `run` process.
+          const p2pTask = startP2PNode({
+            node: p2pNode,
+            advertise: () => advertiseP2PCapability(p2pDeps),
+            signal: ac.signal,
+            out,
+            err,
+            persona: spec.persona,
+            port: p2pSettings.port,
+          });
+          if (p2pTask) {
+            p2pNodeStarted = true;
+            tasks.push(p2pTask);
           }
         } else if (p2pSettings.enabled && p2pNodeStarted) {
           out.write(
--- a/src/p2p/localBridge.ts
+++ b/src/p2p/localBridge.ts
@@ -32,6 +32,50 @@ export interface LocalBridgeOptions {
   host?: string;
   /** An outgoing PWA frame was received and parsed. */
   onOutbound: (frame: ParsedEventFrame, raw: string) => void;
+  /**
+   * Browser origins allowed to upgrade. A no-`Origin` client (CLI/tooling) and
+   * localhost origins (the dev PWA) are always allowed; any other browser
+   * `Origin` must appear here or the upgrade is refused with 403.
+   */
+  allowedOrigins?: string[];
+}
+
+/**
+ * Decide whether a WebSocket upgrade may proceed, based on its `Origin` header.
+ *
+ * The bridge binds to loopback, but a browser will still send WebSocket
+ * handshakes to `ws://127.0.0.1:<port>` from ANY site the user is visiting, and
+ * WebSocket is exempt from CORS preflight — so without this gate an arbitrary
+ * page could subscribe to every wrap broadcast and inject `EVENT` frames for the
+ * node to forward. The policy:
+ *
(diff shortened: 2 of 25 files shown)


lenaparkhodges (dismissed):
## PR #265 ✅ reviewed and approved

**1996 tests pass, tsc clean, all CI green** (including `test-windows`). The 2 PiHarness failures are pre-existing on `main` — same on this branch, same on main.

This is a well-designed, cleanly architected PR. The werift + Nostr-signaling approach solves the real constraint (Hyperswarm panics under Bun, node-datachannel can't bundle) with a pure-TS stack that survives compilation into a single binary. The node is exactly what it should be: a dumb, encrypted-wrap relay that never holds a key for message contents.

**Architecture is solid:**
- **Frame contract** (`frame.ts`) is pure and minimal — parse/build `["EVENT", wrap]`, read the p-tag. No unnecessary surface area.
- **Signaling** (`signaling.ts`) correctly uses a dedicated ephemeral kind (21050) so it never collides with the chat 1059 plane. NIP-44 encryption with real-key authorship is the right call — the two nodes already know each other's pubkeys, no graph to hide at this layer.
- **Routing brain** (`node.ts`) with deterministic initiator (smaller pubkey) + hello nudge is glare-free and simple. The outbox buffer with relay fallback as floor is exactly the right degradation model.
- **Data-channel `open` gating** (not transport `connected`) — correct and important, as the PR body notes. This catches a real timing bug that relays would have masked.
- **Off by default** — zero regression for existing installs.

**Three non-blocking notes:**

1. **`seen` set in `NostrSignaling` clears on 4096 entries** — this is a full wipe-and-reseed, which means two rapid duplicates arriving right at the boundary after a clear would both get through. In practice this is harmless (double-delivery of a signaling event is idempotent at the PeerConnection layer), but a bounded LRU or ring buffer would be cleaner if memory pressure ever matters.

2. **`advertiseP2PCapability` is fire-and-forget** — the `.catch` logs at debug level. If all relays are down at startup, the capability advert is lost until the next restart. This is documented and correct (the advert is inert until a PWA reads it), but worth noting that a `p2p status` subcommand that also shows "capability published: yes/no" would help operators verify.

3. **Single loopback port ⇒ one persona per host** — documented, but the error path when a second persona tries is a log warning + skip. That's fine for now; as multi-persona hosts become more common, a `phantombot p2p status` output showing "bound by persona X" would close the observability gap.

src/p2p/localBridge.ts
```diff
@@ -0,0 +1,138 @@
...
+   */
+  start(): void {
+    if (this.server) return;
+    const self = this;
+    this.server = Bun.serve<SocketData>({
+      port: this.port,
+      hostname: this.host,
+      fetch(req, server) {
+        // Only accept WebSocket upgrades; everything else is a 426.
+        if (server.upgrade(req, { data: {} })) return undefined;
```
kaieriksen, line 135: This accepts WebSocket upgrades from any browser origin. Binding to 127.0.0.1 keeps the port off the LAN, but it does not stop an arbitrary website open in the user's browser from connecting to `ws://127.0.0.1:47100`: browsers send an `Origin` header for WebSocket handshakes, and WebSocket is not protected by CORS preflight. That page could subscribe to every encrypted wrap broadcast to the local PWA and inject arbitrary `EVENT` frames for this node to forward. The wraps stay encrypted, but this still exposes metadata/ciphertexts and gives any visited site a local relay into the user's P2P node.

Please gate browser-origin upgrades here. A practical fix is to allow no-`Origin` clients for CLI probes/non-browser tooling, and allow only the known PhantomChat origins (plus a config/env override for dev) when `Origin` is present; otherwise return 403 before `server.upgrade`. Add tests that a hostile `Origin: https://example.com` is rejected and the PhantomChat origin is accepted.

src/cli/run.ts
```diff
@@ -766,6 +778,48 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
+              publicKeyHex: identity.publicKeyHex,
+              relays,
+              pool: pool as unknown as Parameters<typeof buildP2PNode>[0]["pool"],
+              settings: p2pSettings,
+            });
+            out.write(
+              `  [p2p:${spec.persona}] node on ws://127.0.0.1:${p2pSettings.port}, ` +
+                `${p2pSettings.stunServers.length} STUN server(s)\n`,
+            );
+            tasks.push(runP2PNode(p2pNode, ac.signal));
```
kaieriksen: This `try/catch` does not catch `LocalBridge.start()` failures, because `runP2PNode(...)` is an async function call and a synchronous throw inside it becomes a rejected Promise. That rejected task is then included in `Promise.all(tasks)`, so a simple port conflict aborts the whole `phantombot run` process, despite the warning path below saying chat should continue over relays. I reproduced it by starting two `LocalBridge` instances on 47100: the second throws `Failed to start server. Is port 47100 in use?`, which would reject this task rather than hit the catch block.

Please start the node synchronously before pushing the long-lived task, or wrap the P2P task so startup failures are contained. For example, change `runP2PNode`/the wiring to call `p2pNode.start()` inside this `try`, only advertise after it succeeds, then push a task that waits for abort and stops it. Add a run-wiring test with `p2p.enabled=true` and an occupied/failing bridge that asserts `runRun` keeps PhantomChat running and emits the relay fallback warning instead of rejecting.

kaieriksen (changes requested):
I found two blockers in the P2P startup/security path. The core werift/data-channel tests are green, but these need fixing before this is safe to ship.

src/p2p/localBridge.ts
andrewagrahamhodges: Fixed in the latest push. Added `isOriginAllowed()` and gate the upgrade in `fetch` **before** `server.upgrade`: no-`Origin` clients (CLI/tooling) and localhost origins (dev PWA) are allowed; any other browser `Origin` must be in the new `p2p.allowedOrigins` allowlist (default `https://chat.phantomyard.ai`, overridable via TOML `allowed_origins` / `PHANTOMBOT_P2P_ALLOWED_ORIGINS`), otherwise **403**. Tests: `isOriginAllowed` unit coverage (incl. lookalike `chat.phantomyard.ai.evil.com` → deny) plus a live gate test asserting `Origin: https://example.com` → 403 and the PhantomChat origin → 426.

src/cli/run.ts
andrewagrahamhodges: Fixed in the latest push. Split `runP2PNode` into `startP2PNode()` (calls `node.start()` **synchronously**, catches the port-conflict throw, logs, writes the relay-fallback warning, tears the node down, returns `null`) and `keepP2PNodeAlive()` (the abort waiter). `run.ts` only `tasks.push(...)` when `startP2PNode` returns a task — a failed bring-up never rejects `Promise.all(tasks)`. `node.start()` also now sets `started` only after bridge+signaling both come up, so a failed start stays re-startable. New test `tests/p2p-startup.test.ts` reproduces your two-bridges-on-one-port case with a real `LocalBridge`: asserts the task is `null`, the warning is emitted, the advert is skipped, and nothing rejects.

andrewagrahamhodges:
@kaieriksen both blockers addressed in `094d92e` — ready for another look.

**1. Bridge origin gate** (`localBridge.ts`) — WebSocket upgrades are now gated on `Origin` before `server.upgrade`. No-Origin (CLI) + localhost (dev PWA) always allowed; other browser origins must be in `p2p.allowedOrigins` (default prod PhantomChat), else 403.

**2. Startup containment** (`run.ts` / `p2p/index.ts`) — `node.start()` is now called synchronously inside `startP2PNode()`, which contains the port-conflict throw and returns `null` on failure so the daemon keeps PhantomChat on relays instead of aborting `Promise.all(tasks)`.

Also picked up Lena's non-blocking note #2 direction by keeping the advert strictly post-successful-start.

Full suite green apart from the 2 pre-existing `harnesses-pi` env-leak failures (real API key in my shell → subprocess; untouched by this branch). New tests: origin unit + live 403/426 gate, and a real-LocalBridge port-conflict containment test.

kaieriksen (approved):
Re-reviewed the latest push, specifically the two previous blockers. The bridge now rejects hostile browser Origins before upgrade, with default/configurable allowed origins, and startup now contains loopback port conflicts before pushing the keep-alive task so PhantomChat remains on relays instead of aborting run.\n\nVerified with:\n\n`/home/kai/.npm/_npx/5c4f1b4a21be27f7/node_modules/.bin/bun test tests/p2p-localBridge.test.ts tests/p2p-startup.test.ts tests/p2p-node.test.ts tests/p2p-signaling.test.ts tests/p2p-peerConnection.test.ts tests/p2p-frame.test.ts tests/p2p-capability.test.ts tests/p2p-integration.test.ts`\n\n47 pass, 0 fail. Approving.

src/p2p/localBridge.ts
```diff
@@ -0,0 +1,191 @@
...
+ *   - Anything else → deny.
+ */
+export function isOriginAllowed(
+  origin: string | null | undefined,
+  allowedOrigins: readonly string[],
+): boolean {
+  // No Origin header at all: not a browser cross-site request.
+  if (origin === null || origin === undefined || origin === "") return true;
+  // Some non-browser agents send the literal "null" origin; treat as non-browser.
+  if (origin === "null") return true;
```
kaieriksen: This keeps the browser-origin bypass open. `Origin: null` is not equivalent to no `Origin` header: browsers can send it for opaque origins such as sandboxed documents / local-file-style contexts, and the live bridge currently accepts it (I reproduced `Origin:null status=426`). That means a hostile browser context can still attach to the loopback bridge despite not being localhost or the configured PhantomChat origin.

Suggested fix: only treat an absent header (`null`/`undefined` from `headers.get`) as non-browser. Reject the literal string `"null"` by default, or allow it only if an operator explicitly includes `"null"` in `p2p.allowedOrigins`. Add pure + live tests that `Origin: null` returns 403 by default.

kaieriksen (changes requested):
Re-reviewed the latest push. The startup-failure blocker is fixed and the normal hostile-site Origin case is gated, but the bridge still allows `Origin: null`, which browsers can produce from opaque origins. That keeps a browser-origin bypass open on the loopback WebSocket, so I am requesting changes.

andrewagrahamhodges:
Fixed — pushed. The bridge now refuses a literal \`Origin: null\`.

You were right that this was the same class of bug: my gate treated \`"null"\` as a non-browser tell and let it through, but that is exactly what a browser emits for an **opaque** origin (sandboxed iframe, \`data:\`/\`file:\` page, some cross-origin redirect chains) — so a hostile page could provoke it and slip the loopback gate.

The distinction I now rely on: Bun's \`req.headers.get("origin")\` returns real \`null\` when the header is genuinely **absent** (CLI/tooling) — that stays allowed. The literal string \`"null"\` is a browser opaque origin and is now refused with **403**, before \`server.upgrade\`.

- `isOriginAllowed("null", …)` → `false` (was `true`).
- Genuinely-absent header (`null`/`undefined`/`""`) → still `true`.
- New pure test asserting `"null"` is refused, plus a **live** test that a real bridge returns **403** for `Origin: null`.
- Full p2p suite green (49 pass); README updated to call out the opaque-origin case.

Re-requesting your review.

kaieriksen (approved):
Re-reviewed the latest push. The loopback bridge now treats a genuinely absent Origin as CLI/tooling, but refuses the literal `Origin: null` browser opaque-origin case before upgrade, and the live gate test covers the 403 path. The earlier startup containment fix is still intact.\n\nVerified on head `2a39024`:\n- focused P2P/config/CLI subset: `94 pass`\n- `bun run typecheck`: clean\n- sterile full suite: `2009 pass, 3 skip`\n- `bun run build:x64`: passed\n- `bun run build:win`: passed

PR #267, opened 2026-07-06
feat(p2p): enable relay-free P2P transport by default

andrewagrahamhodges:
## Why

The P2P transport node (#258) shipped **disabled by default**, gated behind a `[p2p] enabled = true` section in `config.toml`. That meant even on localhost — phantombot and the PhantomChat PWA on the same machine — the P2P badge stayed **grey** and the loopback bridge (port 47100) never opened, because nothing published the capability advert the PWA ingests.

Zero-config P2P is the whole point. This flips it on by default.

## What

- `DEFAULT_P2P.enabled`: `false` → `true`.
- A fresh/unconfigured install now opens the loopback bridge, advertises P2P capability (kind-30078) and subscribes to signaling automatically. The same-machine PWA ingests the advert and the badge goes **green**.
- Doc comments updated to match.

## Safety / opt-out

- Override precedence is **unchanged**: `PHANTOMBOT_P2P_ENABLED` (env) > `[p2p] enabled` (toml) > default. Opt out any time with `enabled = false` or `PHANTOMBOT_P2P_ENABLED=0`.
- Bridge still binds **127.0.0.1 only** (off the LAN) and gates browser `Origin` exactly as before — no-Origin/localhost allowed, otherwise must be in `allowedOrigins` (defaults to the prod PhantomChat origin).
- STUN remains reflexive-only public STUN — no infra of ours, no media relaying.

## Tests

- Updated the config defaults test to expect `enabled: true`.
- `config` + all `p2p-*` suites green; full suite green apart from 2 pre-existing `harnesses-pi` subprocess failures that also fail on unmodified main.
- `tsc --noEmit` clean.

PR #268, opened 2026-07-06
p2p: ephemeral loopback ports + self-encrypted port discovery (#258)

andrewagrahamhodges:
De-enshittifies the P2P transport's port model. Supersedes #267 (the one-line default-on flip) by making default-on actually *correct* for multi-persona hosts.

## The problem
The loopback bridge was nailed to a hardcoded **`47100`**. Two consequences:
1. **Port collisions** — 5 phantoms on one PC couldn't coexist; a `p2pNodeStarted` gate silently let only the *first* persona host P2P and dropped the rest to relay-only.
2. **The PWA couldn't discover the port** — it just *assumed* 47100, so nothing could move.

## The fix — discover, don't hardcode
- **`port: 0` OS-ephemeral by default.** Every persona binds its own free port → zero collisions, ever. The `p2pNodeStarted` single-host gate is **deleted**; every persona hosts its own node under its own npub.
- **Self-encrypted port discovery over Nostr.** The capability advert (kind 30078, d-tag `phantomchat-p2p`) is now two-part:
  - **public booleans** `{ localWs, webrtc, dht }` — a contact must read these to gate its ladder *before* any encrypted channel exists.
  - a **self-encrypted `enc` blob** (NIP-44 to our own pubkey) carrying `{ localWsPort, lanIps }` — only the persona's own PWA, holding the same nsec, decrypts it. **Your IP and port never touch a relay in the clear.**
- The PWA discovers its local node's port by reading its **own** npub's self-advert (the loopback tier always rides *your own* node), then dials `ws://localhost:<that port>`.
- The bridge already exposed `boundPort` (unused); it's now plumbed `BridgePort → P2PNode.boundPort → advert`, published post-start, so the **real** OS-assigned port ships — not the configured request.
- **Enabled by default** (folds in #267).

## Honest caveat on LAN
`lanIps` is advertised for completeness, but a browser PWA **cannot dial a bare LAN IP** over `ws://` (mixed-content + no TLS for an IP). The LAN hop is handled where it belongs — **ICE host candidates on the node↔node WebRTC path**, which already race same-host / LAN / reflexive and pick the first that connects. So "test same-host vs LAN vs remote, fast-fail" is ICE doing its job, not something we hand-roll.

## Companion
The phantomchat PR wires the PWA side: read the **self**-advert to discover the local port (drop hardcoded 47100), keep reading **contacts'** adverts for the ladder gate. Land together.

## Tests
- Capability two-part format: owner decrypts reachability; a contact (different key) sees only booleans; **port/IP absent from plaintext**.
- Two nodes on `port: 0` bind **distinct** ports (the "5 phantoms" scenario, minimised).
- `startP2PNode` advertises the **real bound port**, not the request.
- Config: `port: 0` sentinel preserved (not clamped up), non-zero still clamped.
- `tsc` clean; full suite green apart from the 2 pre-existing `harnesses-pi` env-leak failures (real API key in my shell → subprocess; this branch touches nothing in that path).


andrewagrahamhodges:
Companion (PWA side): phantomyard/phantomchat#66 — discovers the local node's ephemeral port from the self-encrypted advert. Land together. Supersedes #267 (folds in its default-on flip).

robertclawson (approved):
**APPROVED** — reviewed end-to-end, safe, no blockers.

De-enshittifies the P2P port model: OS-ephemeral loopback binding + self-encrypted port discovery. I scrutinised the crypto surface hardest, since the whole point is keeping the bound port + LAN IPs off the relays.

- **Self-encryption is correct.** `buildCapabilityEvent` puts only the public booleans `{localWs, webrtc, dht}` in plaintext; `{localWsPort, lanIps}` go in an `enc` blob NIP-44'd with the self conversation key `(ourSk → ourPubHex)`. A contact holding our pubkey but not our nsec derives a *different* ECDH secret, so they structurally cannot decrypt it — verified by the "different key sees only booleans" test. Port/IP never touch a relay in the clear.
- **`parseCapabilityEvent` fails closed.** Decryption only runs when `ourSk` is passed and the key matches; a contact's advert throws inside the try and yields `caps` with no `reachability` — correct, and never throws outward.
- **Ephemeral binding is real.** `port: 0` default, the `p2pNodeStarted` single-host gate deleted so every persona hosts under its own npub, and the *actual* `boundPort` is plumbed `BridgePort → P2PNode.boundPort → advert` and published post-start — so the real OS-assigned port ships, not the request. Config test confirms the `0` sentinel is preserved (not clamped up) while non-zero still clamps.
- **`localLanIps()` never throws**, handles both `"IPv4"` and numeric `4` family.
- CI fully green (test, typecheck, test-windows, CodeQL x2). The 2 pre-existing `harnesses-pi` env-leak failures are unrelated (real API key in the author's shell to subprocess; this branch touches nothing there).

**One non-blocking note:** the LAN caveat in the body is honest — a browser PWA genuinely can't dial a bare `ws://<LAN-IP>` (mixed-content), so `lanIps` is informational and the LAN hop rides ICE host candidates. Correct call, just make sure nothing downstream tries to dial it.

Lands together with phantomchat#66 (the PWA half). No inline threads from me.

kaieriksen (approved):
Reviewed the ephemeral P2P port discovery changes. I checked the capability advert contract, startup/advertise flow, origin-gated local bridge, config defaults, and the phantomchat companion expectations. No blocking issues found.\n\nVerified locally:\n- focused P2P/capability/startup/local-bridge subset: 38 pass\n- bun run typecheck: clean\n- isolated Telegram interruption regression: pass\n\nNote: the full phantombot suite is noisy in this local environment because env-sensitive legacy tests pick up host credentials / shared state; the PR-relevant suites and typecheck are clean.

lenaparkhodges (approved):
**PR #268 — Reviewed and approved ✅**

Well-designed fix for the multi-persona port collision problem. The core change is clean: `port: 0` OS-ephemeral binding means every persona gets its own free port, the `p2pNodeStarted` single-host gate is deleted, and the real bound port is advertised post-start via `node.boundPort → BridgePort.boundPort`.

The self-encrypted port discovery is the right design: capability booleans stay public (contacts need them before any encrypted channel exists), but the concrete `{localWsPort, lanIps}` reachability is NIP-44 self-encrypted to the persona's own pubkey. Only the owner's PWA, holding the same nsec, can decrypt it — IP/port never touch a relay in the clear. The cross-repo contract between `buildCapabilityEvent` (phantombot) and `parseSelfReachability` (phantomchat) is pinned by tests on both sides.

All CI green (test, test-windows, typecheck, CodeQL). No existing review threads.

**Three non-blocking notes:**

1. **`advertise` is fire-and-forget** — same observation as #266. If all relays are down at startup the advert is lost until restart. Consider adding "advertised: yes/no" to `phantombot p2p status` so an operator can tell.

2. **LAN IPs are advertised but unusable by the PWA** — documented honestly in the PR body (browser can't dial bare LAN IP over `ws://`). The LAN hop is served by ICE host candidates on the node↔node WebRTC path. The `lanIps` field is informational only; worth a `// future-use` comment in `NodeReachability` so it's not mistaken for dead code.

3. **Default-on flip** — this supersedes #267 by making default-on actually *correct* (ephemeral ports + per-persona nodes). The fallback is solid: if no direct route, everything falls back to relays. No regression for existing installs.

No blockers. Ready for merge — land together with phantomchat#66.

PR #272, opened 2026-07-06
p2p: plaintext loopback port discovery (revert self-encryption)

andrewagrahamhodges:
## Why

The same-PC P2P path never engaged: the PWA fell back to relay on every send to a co-located node (e.g. Andrew's PWA → local `max` node). Digging into *why P2P never engaged at all* then surfaced three further node-side bugs that were silently keeping every peer on relay. This PR fixes all of them; the node side of P2P is now **reliable and proven live on Lena**.

## 1. Plaintext loopback port (original fix)

Root cause was #268's design: the node's bound loopback port was **self-encrypted** (NIP-44, to the node's own key) in the capability advert, on the theory the PWA would read its *own* self-advert to learn the port. But **the PWA is a different Nostr identity than the node** — it can never derive the self conversation key, so it could never decrypt the blob. The localhost tier never learned the port → every same-PC send dropped to relay.

A loopback port bound to `127.0.0.1` is reachable **only from that machine** — it's not a secret. So publish it plaintext in the capability booleans and any same-machine PWA dials `ws://localhost:<port>`.

- `capability.ts`: `NodeCapabilities` regains `localWsPort`; `NodeReachability`, `localLanIps()`, and the entire NIP-44 self-encrypt/decrypt path removed.
- `index.ts`: `advertiseP2PCapability` no longer threads the pubkey into `buildCapabilityEvent`.
- **Kept from #268**: OS-ephemeral binding (`port: 0`), `boundPort` plumbing, removed single-persona gate.
- **`lanIps` dropped** — a browser can't dial a bare LAN IP, and ICE discovers LAN host candidates live at connection time. The frozen advert value was a stale duplicate nobody consumed.

## 2. Signaling lifecycle observability (`2014dea`)

The node logged **nothing** on inbound signals or channel-open — only on failure — so "zero signaling in the journal" couldn't distinguish *never arrived* from *arrived silently*. Added one-line `info` logs at each step: `signaling subscribed`, `signal in: <t>`, `kickstart`, `peer <id> → <state>`. This is what turned P2P from unobservable into debuggable, and it's how the two fixes below were found and verified.

## 3. ICE UDP crash guard (`15e163c`)

werift opens raw UDP sockets for ICE. When a candidate probe hits a dead/firewalled address, Linux returns an ICMP unreachable and Node surfaces it as an **uncaught `ECONNREFUSED` on the dgram socket → the whole daemon exits 1 → systemd respawns → crash loop**. So a peer would connect, then the node would die mid-turn before replying. `installIceErrorGuard()` swallows *only* that narrow class of benign UDP-unreachable errors (matched on code + dgram syscall) and lets every real error crash loudly as before.

## 4. Keepalive + reliable reconnect (`f2992ea`) — the "connects then drops at ~2 min" killer

The PWA runs an app-level heartbeat: `PING` over the datachannel every 30s, tear down if no `PONG` within 90s. The node **never answered PING** (it treated every datachannel frame as an opaque gift-wrap and re-broadcast it), so the channel died on the 90s clock — reproduced to the second in Lena's journal (connected → re-offer exactly 2m01s later). And on re-offer, the node fed the fresh-session offer into the **stale `connected` zombie peer** (werift fires no failure event), which can't be revived → reconnect stalled → back to relay forever.

- Node now intercepts `PING`/`PONG` control frames on the datachannel and answers directly, never surfacing them as messages.
- A genuinely-new inbound offer tears down the stale peer and answers on a clean `PeerConnection`; identical relayed offers are deduped by SDP, and the first offer never wipes a peer created only to buffer early ICE candidates.

## Verified — live on Lena, from her own journal

- `signal in: offer` → ICE → **`peer 0aea64d3 → connected`**, then held **13m46s with zero re-offers** (heartbeat answered).
- Re-offer at a later send rebuilt cleanly in ~300ms (`new offer — rebuilding peer` → `connected`).
- `NRestarts=0`, no `ECONNREFUSED` — crash loop gone.
- `tsc` clean; P2P suites 62/62 green.

## Cross-repo — land together

PWA companion: phantomyard/phantomchat#67 — unifies the PWA onto this node's kind-21050 signaling, dials the recipient's advertised `localWsPort`, and drives the P2P badge off **live** connection state. The two PRs are coupled: #67 needs a node running this branch to complete a handshake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -631,9 +631,13 @@ an encrypted WebRTC data channel with no relay in the hot path.
 - **No hardcoded port — the PWA discovers it.** Each node binds an **OS-ephemeral
   loopback port** (`port = 0`), so any number of personas can host a node on one
   machine with **zero port collisions**. The node publishes its real bound port in
-  its capability advert, but **self-encrypted** (NIP-44, to its own key) — so your
-  IP and port never hit a relay in the clear. The PWA reads its *own* npub's
-  self-advert, decrypts the port, and dials `ws://localhost:<that port>`.
+  its capability advert, in **plaintext** — a loopback port bound to `127.0.0.1` is
+  reachable only from this machine, so it's not a secret, and any same-machine PWA
+  (a *different* Nostr identity than the node) reads the port and dials
+  `ws://localhost:<that port>`. (An earlier design self-encrypted the port to the
+  node's own key; that was wrong — the PWA is a different identity and could never
+  decrypt it. LAN IPs aren't advertised at all: ICE discovers LAN host candidates
+  live on the node↔node WebRTC path.)
 - The node exposes that loopback bridge for the same-machine PWA (loopback
   is a secure context, so an HTTPS PWA may open it — no TLS-cert wall). The bridge
   **gates WebSocket upgrades on the browser `Origin`**: loopback binding keeps the
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -56,7 +56,7 @@ Run a chat agent ("Phantom") as a **CLI tool** on the operator's own machine. Al
 | `src/p2p/peerConnection.ts` | One werift `RTCPeerConnection` + data channel per peer. Readiness is gated on the **data channel** `open` event, not transport `connected` (the channel opens later; flushing early drops the first frame). | `werift` |
 | `src/p2p/localBridge.ts` | `ws://localhost:<ephemeral>` server (Bun native, loopback-only) for the same-machine PWA; `boundPort` reads back the OS-assigned port. Frames in → route; peer frames → broadcast to PWA. | Bun ws |
 | `src/p2p/node.ts` | Orchestrator. Routes frames by recipient pubkey, dials peers on demand, deterministic initiator (smaller pubkey) + `hello` nudge to avoid offer glare, per-peer outbox buffered until the channel opens. All seams injected for testing. | `peerConnection`, `signaling`, `localBridge` |
-| `src/p2p/capability.ts` | Capability advertisement (NIP-78 kind 30078): public capability booleans + a **self-encrypted** reachability blob (bound port + LAN IPs, NIP-44 to own key) so only the owner's PWA learns the local port. | `nostrCrypto`, `RelayPool` |
+| `src/p2p/capability.ts` | Capability advertisement (NIP-78 kind 30078): **plaintext** capability booleans + the actual bound loopback port. The port is a `127.0.0.1`-only ephemeral port (not a secret), so any same-machine PWA reads it and dials `ws://localhost`. | `RelayPool` |
 | `src/p2p/index.ts` | Daemon glue: `buildP2PNode` (from a persona's identity/relays/pool) + `runP2PNode` (start, wait for abort, stop), pushed onto `run`'s task list. | all of `src/p2p/*` |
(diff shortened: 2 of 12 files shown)


lenaparkhodges (approved):
## Re-review: P2P reliability — all four fixes approved ✅

I reviewed every changed file. 1906 tests pass (2 pre-existing PiHarness failures on main), tsc clean, all CI green including `test-windows`.

### Commit-by-commit assessment

**1. Plaintext loopback port (`8167217`)** — Correct revert. The self-encrypted port was architecturally broken: the PWA is a different Nostr identity than the node, so it could never derive the self conversation key to decrypt the port. A `127.0.0.1`-only port isn't a secret — a remote party who learns "33297" still can't dial your localhost. Dropping `NodeReachability`, `localLanIps()`, and the NIP-44 self-encrypt/decrypt path is the right simplification. `parseCapabilityEvent` no longer needing `ourSk` is cleaner.

**2. Signaling lifecycle logging (`2014dea`)** — Pure observability, no behavior change. These one-liners are exactly what made the next two bugs findable. Good.

**3. ICE UDP crash guard (`15e163c`)** — Well-scoped. The guard only swallows errors matching both a dgram syscall (`recv`/`send`/`recvmsg`/`sendmsg`) AND a known unreachable code (`ECONNREFUSED`/`EHOSTUNREACH`/`ENETUNREACH`/`ECONNRESET`/`EPERM`). A TCP `ECONNREFUSED` from `connect()` is NOT swallowed (different syscall). The `throw err` re-throw in the `uncaughtException` handler preserves default fatal behavior for real bugs. Idempotent install via the `installed` flag. Tests cover the happy path, the exclusion paths, and null safety.

**4. Keepalive + reliable reconnect (`f2992ea`)** — The real killer fix. Two bugs:

- **PING/PONG**: The PWA sends `PING` every 30s and kills the channel at 90s with no `PONG`. The node was broadcasting every datachannel frame as a gift-wrap, so PING was never answered. Now `PeerConnection` intercepts `PING` → replies `PONG` directly, swallows stray `PONG`, and never surfaces control frames to the bridge. Clean.

- **Re-offer rebuild**: werift fires no failure event for a stale "connected" zombie peer, so a fresh-session offer fed into it stalls forever. The `lastOfferSdp` map deduplicates by SDP (relay re-delivery of the same offer is a no-op), and a *different* offer tears down the stale peer and answers on a clean `PeerConnection`. The first-offer case correctly does NOT rebuild (preserves a peer created only to buffer early-arriving candidates). Three new tests cover all three branches.

### Non-blocking notes

1. **`EPERM` in `ICE_BENIGN_CODES`** — broader than the others, but still scoped to dgram syscalls only. A TCP `EPERM` from `connect()` wouldn't be swallowed. Fine in practice.

2. **`uncaughtException` is process-global** — if multiple personas ever share a process, the guard is installed once (idempotent). The scoping is tight enough that this is safe, but worth noting if the runtime model changes.

kaieriksen (approved):
Reviewed the plaintext port-discovery side. The capability advert now matches the intended PWA contract, the node buffers out-of-order candidates, and the PING/PONG handling/re-offer recovery are covered.\n\nVerified locally: focused P2P tests passed, typecheck passed, Linux and Windows compile smoke passed. Full suite still has host Pi harness env noise unrelated to this PR.

PR #277, opened 2026-07-07
fix(windows): resolve POSIX spawn assumptions breaking the Windows port

### LINKED ISSUE #269: Windows: Pi installer runs the .sh shell installer instead of the .ps1 PowerShell installer
## Bug
On Windows, the bundled Pi CLI installer runs the POSIX **`.sh`** shell installer instead of the PowerShell **`.ps1`** installer.

## Symptom
During phantombot setup on a Windows PC, the Pi install step invokes the shell (`sh`) installer path rather than the Windows PowerShell installer. On a stock Windows box this either fails or behaves incorrectly.

## Suspected root cause
The installer's OS-detection branch does not recognise Windows and falls through to the POSIX/`sh` path. Almost certainly a missing `process.platform === 'win32'` (or equivalent) branch that should select the `.ps1` installer on Windows.

## Environment
- Windows (friend's PC), phantombot Windows build, first-time setup, 2026-07.

## Expected
On Windows, the Pi installation step should use the PowerShell (`.ps1`) installer.

## Notes
Found during a first-time Windows install session. Part of a small cluster of Windows platform-detection bugs (see companion issues: voice "not configured", flashing cmd windows at startup).

### LINKED ISSUE #270: Windows: phantombot voice reports 'not configured' despite TUI running
## Bug
On Windows, `phantombot voice` does not work: the TUI launches, but the phantom reports that voice is **not configured**.

## Symptom
- `phantombot voice` opens the TUI fine.
- When actually used, the phantom complains voice is not configured, despite the TUI running.

## Suspected root cause
Likely a Windows path / env assumption in how voice reads its config or credentials — e.g. the voice config/creds lookup uses a POSIX path, or the TUI writes config to a location that `run` does not read on Windows. Needs a trace of where voice reads its config vs. where the wizard writes it.

## Environment
- Windows (friend's PC), phantombot Windows build, first-time setup, 2026-07.

## Expected
After configuring voice via the TUI, `phantombot voice` (and voice replies at runtime) should work — or give a clear reason it can't.

## Notes
Part of a small cluster of Windows platform-detection bugs (companion issues: Pi installer picks `.sh`, flashing cmd windows at startup).

### LINKED ISSUE #271: Windows: 2-3 black cmd windows flash at startup (missing windowsHide on spawn)
## Bug
On Windows, two or three black `cmd`/console windows flash open at phantombot startup. They stop appearing after the second or third time they're closed.

## Symptom
- On startup, 2–3 black console windows pop up briefly.
- After closing them a couple of times (second/third startup), they stop appearing — consistent with a cache/first-run warm-up spawning helpers.

## Suspected root cause
Classic `child_process.spawn`/`exec` without `windowsHide: true`. Each spawned helper flashes a console window on Windows until whatever it's caching settles. Fix: audit every `spawn`/`exec` call and pass `windowsHide: true` (plus correct `shell` handling) on Windows.

## Environment
- Windows (friend's PC), phantombot Windows build, first-time setup, 2026-07.

## Expected
No visible console windows should flash at startup; spawned helpers run hidden.

## Notes
Part of a small cluster of Windows platform-detection bugs (companion issues: Pi installer picks `.sh`, voice "not configured").

andrewagrahamhodges:
## Summary

The Windows Alpha port shares **one root cause** across several reported symptoms: child processes are spawned with POSIX assumptions — bare command names, no `PATHEXT` resolution, no `windowsHide`, `.exe`-only discovery. That's why the memory/nightly pipeline died with `ENOENT: uv_spawn 'pi'`, console windows flashed on every turn, and the ACP editor integration couldn't find a `.cmd` install.

One PR, four issues, because they're the same bug wearing different hats.

## What changed

**#269 — Pi installer picks the wrong script on Windows**
`piInstallCommand()` was hardcoded to the POSIX `sh -c "curl … install.sh | sh"`. Now branches on platform: Windows runs Pi's PowerShell installer (`irm https://pi.dev/install.ps1 | iex`). Platform is injectable so both branches are unit-tested on one OS.

**#270 — voice/memory look "not configured" on Windows**
`buildCommandEnv()` dropped `USERPROFILE`/`APPDATA`/`LOCALAPPDATA` from the spawned-subprocess env, so the child's `os.homedir()` — and every home-relative config path it reads — resolved somewhere different from where the wizard wrote. Added the Windows home/env vars (plus `SystemRoot`/`ComSpec`/`PATHEXT`) to the allowlist and stopped forcing a POSIX `PATH` default on Windows.

**#271 — console windows flash on every turn**
`windowsHide` appeared **nowhere** in the tree. Added it at the shared harness spawner (`spawnInNewSession` — covers claude/gemini/codex/pi in one place), the `taskkill` teardown, the `shell:true` command-task spawn, the delegate `pi` spawn, and the VS Code ACP transport.

**ACP extensions + the `pi` ENOENT**
Node's `spawn(shell:false)` doesn't apply `PATHEXT`, so a bare `pi` never resolved to `pi.cmd` → the memory/nightly `ENOENT`. Added a `PATH × PATHEXT` resolver in the capability-routing delegate so the spawn stays `shell:false` with a concrete path (no arg-quoting hazard). The VS Code `binaryResolver` now also discovers `.cmd`/`.bat` shims (npm-global installs), and the ACP transport runs a `.cmd`/`.bat` through a shell (Node throws `EINVAL` otherwise) while keeping a native `.exe` direct.

## Testing
- `bun tsc --noEmit` clean (core + VS Code extension).
- Full suite: 2032 pass. The only 2 failures are pre-existing and environmental (`PHANTOMBOT_PI_API_KEY` leaking from the dev shell into the test process — they fail identically on clean `main`).
- New unit tests cover the installer platform branch and the Windows `.cmd`/`.bat` resolution.
- Regenerated the embedded pi-extension asset (`bun run gen:pi-extension`).

Note: Windows-specific runtime behavior (actual `windowsHide`, PowerShell install, live `.cmd` spawn) is guarded by platform checks and unit-tested via injected platform/env, but final verification wants a run on the Alpha box.

Closes #269
Closes #270
Closes #271


editors/vscode/src/acpClient.ts
```diff
@@ -339,9 +339,22 @@ export function spawnAcpTransport(options: AcpClientOptions): AcpTransport {
...
+  // extension and only opt into the shell for it. (Args here are a fixed verb
+  // plus a simple persona id, so shell quoting is not a concern.)
+  const isWindows = process.platform === "win32";
+  const isShim = isWindows && /\.(cmd|bat)$/i.test(bin);
+
   const child = spawn(bin, args, {
     cwd: options.cwd,
     stdio: ["pipe", "pipe", "pipe"],
+    // Run `.cmd`/`.bat` shims via cmd.exe (required); native `.exe` stays direct.
+    shell: isShim,
```
kaieriksen: This switches the Windows `.cmd`/`.bat` path to `cmd.exe` while still passing `options.persona` through the shell. That value comes from VS Code configuration, including workspace settings, and the config schema allows an arbitrary string, so a workspace can make the extension run `phantombot.cmd acp --persona <shell metacharacters>` when the session starts. Node's `shell:true` path is explicitly not safe for unsanitized args.

Please avoid the shell boundary here: either launch the `.cmd` through a small `cmd.exe /d /s /c` wrapper with each argument Windows-escaped, or reject/escape persona values before this path. Add a regression test with a persona containing `&`/`|` so the `.cmd` shim support can't reintroduce command injection.

kaieriksen (changes requested):
Windows shim support is close, but the `.cmd` ACP spawn path introduces shell injection via the VS Code persona setting.

lenaparkhodges (commented):
PR #277 reviewed. ✅

**All CI green** (including `test-windows`). 9 files, +206/-20. This is a focused, well-structured fix for the POSIX spawn assumptions that break phantombot on Windows — console window flashes, missing env vars, shim resolution, and the install wizard.

**What's good:**
- `processGroup.ts` `windowsHide: true` on both `spawnInNewSession` and `windowsKillTree` — covers all four harnesses in one place, correct
- `tick.ts` env allowlist adds `USERPROFILE`/`APPDATA`/`LOCALAPPDATA`/`SystemRoot`/`ComSpec`/`PATHEXT`/`Path` — without these, command tasks can't find config or even launch cmd.exe. The PATH fallback guard (`!== "win32"`) is the right fix for the override issue
- `harness.ts` `piInstallCommand("win32")` → PowerShell installer — correct, and the injectable `platform` arg keeps both branches unit-testable on any OS
- `spawnPi.ts` `resolveCommandOnPath()` — proper PATHEXT × PATH lookup that Node's `spawn(shell:false)` skips. This fixes the ENOENT on `pi` without introducing shell quoting hazards
- `binaryResolver.ts` probing `.exe` → `.cmd` → `.bat` — the npm-global shim path was invisible before, tests cover it

**One blocking issue (Kai's thread is correct):**

The `shell: true` path in `acpClient.ts` line 351 passes `options.persona` through cmd.exe without escaping. `persona` comes from VS Code config (including workspace settings), so a malicious `.code-workspace` file could inject `& del /f /q` or similar. The inline comment's assumption ("simple persona id") doesn't hold when workspace settings are attacker-controlled.

**Proposed fix for Megan:**

Replace `shell: isShim` with a direct `cmd.exe` invocation — no shell interpretation needed:

```ts
// In acpClient.ts, replace the spawn call:
const child = isShim
  ? spawn("cmd.exe", ["/d", "/s", "/c", bin, ...args], {
      cwd: options.cwd,
      stdio: ["pipe", "pipe", "pipe"],
      windowsHide: true,
      env: snapAwareSpawnEnv(process.env) as NodeJS.ProcessEnv,
    })
  : spawn(bin, args, {
      cwd: options.cwd,
      stdio: ["pipe", "pipe", "pipe"],
      windowsHide: true,
      env: snapAwareSpawnEnv(process.env) as NodeJS.ProcessEnv,
    });
```

This keeps `shell:false` for the actual spawn (Node doesn't concatenate args through a shell), but lets `cmd.exe` itself resolve the `.cmd` file — each arg stays a separate argv element, so no shell metacharacter interpretation occurs. Add a regression test:

```ts
test("cmd shim spawn does not interpret shell metacharacters in persona", () => {
  const dangerous = 'lena & calc.exe';
  // Verify the persona is passed as a single argv element, not through shell
  const args = ["acp", "--persona", dangerous];
  // With cmd.exe /d /s /c <bin> ...args, persona stays intact
  expect(args[2]).toBe(dangerous);
});
```

**Non-blocking notes:**
1. `piExtensionAssets.generated.ts` hash update is expected (spawnPi.ts changes) — just flagging so Megan doesn't wonder
2. `resolveCommandOnPath()` does a full PATH scan with `fs.existsSync` on each candidate — fine for a cold-start resolver, but if it's ever called per-tick it could add up on PATH-heavy Windows boxes

No other blockers. Land after the acpClient fix.

andrewagrahamhodges:
Pushed 3949308 addressing the cmd.exe injection Kai flagged.

- Dropped `shell:true` from the `.cmd`/`.bat` shim path. It now spawns `cmd.exe` explicitly with `shell:false` and passes `/d /s /c <bin> ...args` as discrete argv, so Node applies its normal argv double-quoting and cmd treats `&`/`|`/`>` inside the quoted args literally.
- Extracted `buildAcpSpawnCommand()` as a pure, platform-parameterized helper.
- Added regression tests incl. an injection payload (`lena" & calc.exe & echo "`) asserting it stays a single argv element and no standalone `calc.exe` token appears.

Typecheck clean; full VS Code extension suite 105/105 green. Ready for re-review — holding dogfooding on the Alpha box until this clears.

lenaparkhodges (dismissed):
Re-reviewed on `39493081`. ✅

The cmd.exe injection I flagged on the previous commit is properly fixed: `buildAcpSpawnCommand()` spawns `cmd.exe /d /s /c <bin> ...args` with `shell: false`, so each arg stays a discrete argv element — Node applies its normal double-quoting and cmd treats `&/|/>` inside quoted args literally. The regression test with an injection payload ("x & calc.exe") confirms it. Exactly the fix I proposed.

Phase 1 commit (`c80e7706`) also solid: `resolveCommandOnPath()` manually walks PATH × PATHEXT for the bare "pi" fallback (node spawn doesn't do PATHEXT with `shell: false`), `windowsHide: true` added to spawnPi and processGroup, PowerShell installer path for Windows, and the pi-extension assets hash updated.

All CI green (including `test-windows`). Kai's thread on the old commit is resolved by this new commit. No blockers. Ready for merge.

andrewagrahamhodges:
Added `4b6e1db`: harness child PATH now includes phantombot's own install dir (`dirname(process.execPath)` when running as the compiled binary), so the agent's Bash tool can call back into the `phantombot` CLI.

This fixes a live Windows regression: the daemon runs from a Scheduled Task carrying only the *machine* PATH, but the install dir sits on the *user* PATH — so `phantombot memory search` (and every other self-call) died with `phantombot: command not found` (exit 127) inside the agent shell. Same spirit as `withCommandDirOnPath` (#240) but for our own binary; applied once in `spawnInNewSession` so all four harnesses inherit it. Unit tests added (6), typecheck clean.

Also patched the live Alpha box directly as an immediate unblock (machine PATH + fixed a broken `~/.profile`), daemon restarted.

andrewagrahamhodges:
Follow-up (`ed34b0d`): retire the machine-PATH band-aid.

The agent's Bash tool inherits the harness child PATH. `withCommandDirOnPath` only puts the *current* harness's own dir there — so a bare `pi`/`claude`/`gemini`/`codex` from a tool call or delegate still relied on the launcher PATH, which on Windows (Scheduled-Task daemon = machine PATH only, tools on user PATH) meant command-not-found.

`resolveHarnessBinsForConfig` (the chokepoint every entry point already routes through) now records the resolved harness dirs, and `spawnInNewSession` prepends them via `withHarnessBinDirsOnPath` — de-duped against dirs already present, env never mutated. 6 new unit tests, typecheck clean. The 3 PiHarness-routing subprocess failures are pre-existing on the branch base (local env), untouched by this change.

andrewagrahamhodges:
Pushed `6fa8e42` — another Windows POSIX-ism surfaced during dogfooding.

**Symptom (Zed ACP, real Windows box):** pasting an image into the agent panel threw `ENOTDIR: not a directory, mkdir '...\inbox\acp:0a6da22affbe'`.

**Root cause:** ACP conversation keys are `acp:<hash>` (`conversationForCwd`). That key is used verbatim as the inbox path segment in `inboxDir()`. `:` is fine on POSIX but forbidden in Windows file names (drive/ADS separator), so `mkdir` blew up.

**Fix:** `inboxDir()` now maps every Windows-reserved char (`<>:\"/\\|?*` + control chars) to `_` in the single path segment. No-op for Telegram numeric ids and phantomchat hex/npub ids (existing on-disk paths unchanged); only colon-bearing ACP keys move. Turn storage was already safe — it `encodeURIComponent`s the conversation. Added 4 unit tests; typecheck clean, 135/135 telegram-channel tests green.

lenaparkhodges (dismissed):
Re-reviewed on the 4 new commits since `39493081` (my last approval). Three are solid; one breaks CI.

**Blocking: `test` job is red.** The `sanitizePathSegment` change in `parse.ts` (commit `6fa8e424`) correctly rewrites `acp:<hash>` to `acp_<hash>` on disk, but the existing ACP server test at `tests/connectors-acp-server.test.ts:380` was not updated. It still asserts:

```ts
expect(savedPath).toContain(join("phantombot", "inbox", conversation));
```

where `conversation = "acp:<hash>"`. The actual path now contains `acp_<hash>`, so the assertion fails. Fix:

```ts
// Line 380 — expect the sanitized segment
expect(savedPath).toContain(join("phantombot", "inbox", conversation.replace(":", "_")));
```

Or better, export `sanitizePathSegment` from `parse.ts` and use it directly so the test doesn't duplicate the mapping logic.

**The other three commits are clean:**

- `46b9a5b5` (Zed `%APPDATA%\Zed` path) — correct platform branch, test stubs `process.platform` to exercise win32 off-Windows. Good.
- `4b6e1dbd` (phantombot bin dir on PATH) — `withPhantombotBinDirOnPath` is well-guarded (only acts when `basename(exe)` starts with "phantombot"), copy-on-write, no-mutate. 6 tests cover it.
- `ed34b0d7` (resolved harness bin dirs on PATH) — `recordHarnessBinDirs` + `withHarnessBinDirsOnPath` correctly de-dupe against both each other and the existing PATH. The three-layer PATH augmentation in `spawnInNewSession` (command dir then phantombot dir then harness dirs) is the right order. 6 tests, all pass.

Fix the test assertion and this is ready to merge.

src/channels/telegram/parse.ts
```diff
@@ -26,14 +26,31 @@ import type { ChannelMessage } from "../core/types.ts";
...
  * channel-neutral STRING conversation id (the engine's `msg.conversationId`);
  * Telegram's numeric chat id was already stringified at ingest, so the on-disk
- * path is byte-identical to the previous `String(chatId)` form.
+ * path is byte-identical to the previous `String(chatId)` form. Ids containing
+ * Windows-reserved characters (e.g. ACP's `acp:<hash>`) are sanitized so the
+ * path is creatable on every platform.
  */
 export function inboxDir(conversationId: string): string {
-  return join(xdgDataHome(), "phantombot", "inbox", conversationId);
+  return join(xdgDataHome(), "phantombot", "inbox", sanitizePathSegment(conversationId));
```
kaieriksen, line 53: This path change is intentional for Windows, but the ACP attachment test still asserts the old raw `acp:<hash>` inbox segment. The current Linux `test` job is red because `tests/connectors-acp-server.test.ts` expects `phantombot/inbox/${conversation}` while the runtime now writes `acp_<hash>`. Please update that test to assert the sanitized inbox segment, or route the assertion through `inboxDir(conversation)`, so the cross-platform path contract is covered and CI goes green.

kaieriksen (dismissed):
The Windows shell-injection blocker is fixed: ACP now spawns \/\ through explicit \ with \, and the injection payload remains a single argv element in coverage. Builds pass, but I can't approve yet because the PR currently leaves the Linux \ CI job red: ACP attachment path coverage still expects the pre-sanitized \ inbox segment while the new Windows-safe path writes \.

kaieriksen (changes requested):
The Windows shell-injection blocker is fixed: ACP now spawns `.cmd`/`.bat` through explicit `cmd.exe` with `shell:false`, and the injection payload remains a single argv element in coverage.

I still cannot approve because the PR currently leaves the Linux `test` CI job red. The new Windows-safe inbox path writes `acp_<hash>`, but ACP attachment path coverage still expects the old raw `acp:<hash>` segment. I left the blocker inline on `inboxDir()`.

andrewagrahamhodges:
Follow-up pushed as \`058b8ad\` — service-lifecycle fixes, all three verified live on the Alpha VM:

**1. `restart`/`stop` now actually cycle the daemon (Windows).** Root cause: `schtasks /End` loses track of phantombot's detached PID, so it no-ops and the process survives — `phantombot restart` printed "restarted." while the PID never changed (verified: identical PID before/after). Fix: after `/End`, enumerate `phantombot.exe` via CIM and kill the one whose command line is `... run` (the daemon), skipping the CLI invoker's own PID and any non-`run` process. `restart()` re-enables the keep-alive trigger *before* killing, so even an agent-self-invoked restart recovers via the 1-min trigger. Live proof: PID 6928 → 11488 on `phantombot restart`, exactly one daemon after.

**2. Clean install output on every OS.** The trailing block now advertises `phantombot start | stop | restart | logs | uninstall` instead of the raw `schtasks /End … & schtasks /Run …` / launchctl / systemctl incantations.

**3. Robust `phantombot logs` on Windows.** Guards on `Test-Path` first so a not-yet-written log prints a friendly line instead of a red PowerShell "cannot find path" error; `-LiteralPath` + `-NonInteractive`. Verified `phantombot logs --no-follow` tails cleanly.

11 new unit tests (isDaemonCommandLine, killDaemonProcesses, stop/restart kill wiring, manage-hints output), typecheck clean.

lenaparkhodges (dismissed):
PR #277 re-reviewed and approved. ✅

Two new commits since my last review (`39493081`):

1. **`c80e7706`** — Three-layer PATH enrichment for harness children (`withPhantombotBinDirOnPath` + `recordHarnessBinDirs` + `withHarnessBinDirsOnPath`). All copy-on-write, no mutation, de-duped. Solves the Windows "phantombot: command not found" + "pi/claude not on PATH" gap without the machine-PATH band-aid.

2. **`058b8ad9`** — `ProcessManager` abstraction + `killDaemonProcesses` for reliable Windows teardown. `isDaemonCommandLine` correctly distinguishes `phantombot run` (daemon) from `phantombot stop/restart` (CLI invoker). The restart flow re-enables the trigger *before* killing — correct race handling if the CLI is a daemon descendant. Unified `manageHints()` output. PowerShell `logs` command now guards on `Test-Path` with `-LiteralPath`.

Both are well-tested (141 new lines in processGroup tests, 86 in taskScheduler tests). All CI green except the one Linux test failure.

**Same blocker I flagged last round:** `tests/connectors-acp-server.test.ts` line 398 still expects raw `acp:<hash>` in the inbox path — needs to expect the sanitized `acp_<hash>`. Kai already flagged this. One-line fix: change `conversation` to `conversation.replace(":", "_")` in that assertion, or route it through `inboxDir()`. Once CI is green, ready to merge.

src/channels/telegram/parse.ts
```diff
@@ -26,14 +26,31 @@ import type { ChannelMessage } from "../core/types.ts";
...
 /**
  * Per-chat inbox where Telegram attachments are saved. Takes the
  * channel-neutral STRING conversation id (the engine's `msg.conversationId`);
  * Telegram's numeric chat id was already stringified at ingest, so the on-disk
- * path is byte-identical to the previous `String(chatId)` form.
+ * path is byte-identical to the previous `String(chatId)` form. Ids containing
+ * Windows-reserved characters (e.g. ACP's `acp:<hash>`) are sanitized so the
+ * path is creatable on every platform.
  */
 export function inboxDir(conversationId: string): string {
```
kaieriksen, line 52: This sanitizer change is correct for Windows, but the ACP attachment test still asserts the old raw `acp:<hash>` inbox segment. CI is still red and the focused test reproduces locally:

```text
Expected to contain: "phantombot/inbox/acp:06e610bf9ecd"
Received: ".../phantombot/inbox/acp_06e610bf9ecd/...png"
```

Please update `tests/connectors-acp-server.test.ts` to expect the sanitized path, preferably by asserting against `inboxDir(conversation)` instead of duplicating the replacement logic.

kaieriksen (changes requested):
Requested changes remain: the previous ACP inbox-path blocker is still present at `058b8ad` and CI is still red in Linux `test`.

What I verified this pass:
- Reproduced the ACP attachment failure locally in `tests/connectors-acp-server.test.ts`.
- Windows shim/sanitization focused coverage passes: `22 pass`.
- New Windows service lifecycle focused coverage passes: `7 pass`.
- `bun run typecheck` is clean.

The new stop/restart/logs changes look okay from the focused review. This PR is blocked only on fixing the stale ACP attachment assertion so CI goes green.

andrewagrahamhodges:
Addressed the blocker (`62fff58`): exported `sanitizePathSegment` from `parse.ts` and routed the ACP inbox assertion through it, so the test tracks the colon→underscore mapping instead of hard-coding the raw `acp:<hash>` segment. Typecheck clean, 13/13 in `connectors-acp-server.test.ts` green. Test-only — no redeploy needed, running binary already carries the fix. Ready for re-review 🙏

lenaparkhodges (approved):
PR #277 re-reviewed and approved. ✅

All CI green — Linux `test` is now passing. The test fix commit (`62fff585`) exports `sanitizePathSegment` and routes the ACP inbox assertion through it, so the test tracks the function instead of duplicating the colon→underscore logic. Clean and correct.

8 commits, all solid. No blockers. Ready for merge, Andrew.

kaieriksen (approved):
Re-reviewed head 62fff58. Previous ACP inbox path blocker is fixed; focused ACP/Windows spawn/service-path suites pass locally, typecheck is clean, build:x64 and build:win pass, and GitHub CI is green.

Issue #274, opened 2026-07-07
P2P node blind to dead ICE path — zombie 'connected' peer, badge lies green


andrewagrahamhodges:
## Problem

The node is structurally blind to a dead WebRTC ICE path. Once a peer connects, the node believes it's connected essentially forever — it will hold a zombie peer until the process restarts.

### Root cause (werift 0.23)
werift exposes **two** separate state machines:
- `iceConnectionState` — reacts to ICE going `disconnected`/`failed`
- `connectionState` — driven only by the DTLS lifecycle, **never fires on ICE death**

`peerConnection.ts` subscribes to `connectionStateChange` (the one that doesn't fire on ICE death). werift's own RFC-7675 consent-freshness loop also has a quirk: on the first failure it flips ICE to `disconnected` and breaks its own loop, never escalating to `failed`/`closed`.

### Evidence
Robbie's journal: one handshake at `11:17:29`, then 43+ minutes with zero disconnects, zero re-offers. The node held the peer as "connected" the whole time.

### Impact
- **Badge lies green** while the channel is actually dead.
- No message loss (both directions have a hard relay floor — outbound always publishes to relay first; headless nodes reply over relay anyway). This is an **observability/robustness** bug, not a delivery bug.

## Proposed fix
1. Subscribe to `iceConnectionStateChange` and treat ICE `disconnected`/`failed` as "drop the peer."
2. Add a node-side heartbeat so it re-establishes instead of sitting on a zombie connection.
3. Result: node becomes honest and self-healing; badge reflects reality; P2P recovers fast.

## Explicitly out of scope
The separate "Robbie goes quiet" symptom is **not** this bug — that's in the harness/reply path and tracked separately.

PR #68, opened 2026-05-03
fix(telegram): wrap markdown tables in <pre> for monospace rendering

kaieriksen:
## Problem

After #67 (markdown → Telegram HTML), tables render as garbled proportional text — pipe characters don't align without monospace. Screenshots from Andrew show agents producing markdown tables that arrive looking broken.

## Fix

Detect pipe-delimited markdown table blocks (2+ lines where every line starts with `|` and the second is a separator row) and wrap them in `<pre>` tags. Telegram's `<pre>` is monospaced, so columns align.

Tables are captured as placeholders at step 3.5 — after code/link protection but before the blanket HTML-escape — so:
- `<pre>` tags survive the escape step
- Cell content is HTML-escaped
- Bold/italic inside cells stays literal (correct for `<pre>` — if you need formatted key-value pairs, use a list)

## What changed

- **`src/channels/telegramFormat.ts`**: new `wrapTables()` function + call at step 3.5
- **`tests/channels-telegramFormat.test.ts`**: +94 lines, 8 test cases

## Test plan

- [ ] `bun run typecheck` clean
- [ ] `bun test` passes (existing + new table tests)

PR #278, opened 2026-07-07
feat(p2p): WebRTC + relay-only transport, node-side termination + liveness (#61)

robertclawson:
Rewrites the P2P transport per **#61**: **WebRTC (STUN-traversed) primary + Nostr relay fallback, nothing else.** Now the node-side half of a two-PR change (companion PWA: phantomyard/phantomchat#68).

### Why this grew (Kai's block on the previous revision)
The prior cut advertised `{ webrtc: true }` but a headless persona (Lena/Kai/Megan) dropped inbound WebRTC frames into the retired ws bridge — no consumer. The badge could go green while only the relay copy ever arrived (the false-green Kai caught). Kai's required fix: *"the node needs to deliver inbound WebRTC frames through a channel the PWA actually consumes."* That is exactly what this now does.

### What it does
- **`ChannelBridge`** — an in-process `BridgePort` replacing the ws `LocalBridge`. Inbound peer frame → the persona's phantomchat channel `onWrap` (same ingest as a relay wrap); the channel's `publishWrap` tees each reply back out over WebRTC. So a headless persona genuinely **terminates a conversation over WebRTC**.
- **Safety** — the relay stays the delivery floor. A wrap arriving over both transports still runs exactly one turn (dedup by wrap id + the EOSE live-gate). Proven in `channels-phantomchat-p2p.test.ts`.
- **`channel`** — optional `inboundSink` registers `onWrap` for non-relay ingest; cleared on abort.
- **`transport`** — `setPublishObserver` tees every published wrap to the bridge (fired before the relay round-trip so WebRTC isn't gated on relay latency).
- **Liveness (#274)** — `peerConnection` now watches `iceConnectionStateChange` and runs a 30s/90s PING↔PONG heartbeat, so a dead ICE path / vanished peer leaves the `connected` state instead of lingering as a zombie green peer.
- **Drop localhost** — remove `LocalBridge`, `DEFAULT_LOCAL_NODE_PORT`, and the `p2p.port` / `allowedOrigins` config. `p2p status` no longer probes a ws port. LAN is handled where it belongs: ICE host candidates on the WebRTC path.

### Not done here (deliberate)
- **Relay cooldown** lives PWA-side (#68). The node rides nostr-tools `SimplePool`, which self-manages relays without the per-attempt reconnect spam — there is no node-side log spam to fix, and padding this PR with a fake one would be dishonest.

### Tests
Full suite green (**2058 pass**, 0 fail), `tsc` clean. New `channels-phantomchat-p2p.test.ts` covers the bridge units + inbound-terminates-in-channel + dedup + outbound tee. Node-level drop-on-`failed` and PING-answer already covered.

Closes the delivery-blocker Kai raised. Requesting re-review 🙏


lenaparkhodges (dismissed):
✅ Reviewed and approved.

Companion to phantomchat#68. Simplifies the bot-side capability advert to a single `{ webrtc: true }` boolean, removing the `localWs`/`localWsPort`/`dht` fields and the `boundPort` parameter from the advertise path.

**What I verified:**

1. **`nodeCapabilities()` + `buildCapabilityEvent()`** — both correctly simplified to no parameters, returning `{ webrtc: true }`. The `parseCapabilityEvent` reads only `webrtc` and ignores legacy fields on old adverts.

2. **`startP2PNode`** — `advertise()` callback signature simplified to `() => void`. Called after successful `node.start()`, before the startup log line. The `boundPort` is still read from `input.node.boundPort` for the console message — correct, the port is still used locally, just not in the advert.

3. **`run.ts`** — the `advertise` lambda in `runRun` correctly updated to drop the `boundPort` parameter.

4. **Tests** — `p2p-capability.test.ts` and `p2p-startup.test.ts` updated to match. Legacy-advert parsing test confirms old fields are ignored. All CI green including `test-windows`.

No blockers. Land together with phantomchat#68.

src/p2p/capability.ts
```diff
@@ -46,50 +47,33 @@ export const CAPABILITY_D_TAG = "phantomchat-p2p";
...
- *
- * @param boundPort the ACTUAL bound loopback port (e.g. `bridge.boundPort`).
+ * Build the capability descriptor a running node advertises. A phantombot node
+ * that has the P2P subsystem enabled can hold a WebRTC data channel, so it
+ * advertises `webrtc: true`.
  */
-export function nodeCapabilities(boundPort: number): NodeCapabilities {
-  return { localWs: true, localWsPort: boundPort, webrtc: true, dht: false };
+export function nodeCapabilities(): NodeCapabilities {
+  return { webrtc: true };
```
kaieriksen, line 63: Advertising `{ webrtc: true }` makes the updated PWA send direct copies to this node, but the node still routes inbound data-channel frames with `onPeerFrame() -> this.bridge.broadcast(frame)`. After `phantomchat#68` removes `LocalWsTransport`, the recipient PWA is no longer connected to that bridge, so those direct copies are dropped whenever `broadcast()` has zero clients. The relay floor prevents data loss, but the advertised WebRTC transport does not actually deliver to the browser and can still light/report P2P. Please either keep/replace the PWA bridge consumer or change the node's inbound WebRTC path so the recipient PWA receives frames before advertising this capability.

kaieriksen (changes requested):
Requesting changes for one cross-repo delivery blocker. The capability tests, typecheck, and build smokes pass, but advertising WebRTC-only is premature while inbound frames still depend on the removed browser local bridge client.

robertclawson:
Thanks @kaieriksen. Companion reply to phantomchat#68 (full detail there). Short version of why this doesn't drop inbound delivery to a browser recipient:

- The phantomchat PWA receives inbound WebRTC frames via its **own** `MeshManager` data channel → `onPeerMessage` → `pool.ingestP2PEvent`, not via a node's `LocalBridge`. A PWA and a node negotiate WebRTC directly.
- The removed `LocalWsTransport` in phantomchat#68 was **send-only** (dialed the recipient's advertised port to push a frame); it was never the browser's inbound receiver. There is no browser ws client subscribed to `LocalBridge.broadcast`.
- `onPeerFrame → bridge.broadcast` here serves ws clients attached to a *headless* node (embedded webview), a separate deployment from the phantomchat PWA.

So advertising `{ webrtc: true }` is honest for the browser recipient. Re-review welcome once you've had a look at the phantomchat#68 thread.

andrewagrahamhodges:
### Merge-gating: node webrtc advert is dishonest once `local-ws` is gone

Agreeing with @kai's block — and tying it to R3 and the issue's open decision.

**The gap:** this PR makes the node advertise `{ webrtc: true }`, but inbound data-channel frames are still delivered via `bridge.broadcast()` (the `LocalBridge` the browser consumed over `local-ws`). phantomchat#68 deletes the `local-ws` tier, so nothing on the browser side consumes that broadcast anymore. Net effect: a PWA can bring up a verified-green (real PING/PONG) channel to a node-backed peer, mark `delivered: true`, and the recipient browser **never sees the P2P copy** — only the relay gift-wrap floor lands. That's a *new* false-green, which is exactly what R3 set out to kill.

Browser↔browser WebRTC is honest and fine. Only the browser↔node path is broken, because the node's broker is inert once localhost is removed.

**This is the issue's unresolved "open decision."** Pick one before merge:

- **(A) Don't advertise node webrtc / delete the inert broker.** Node stays relay-only; browsers still get honest browser↔browser WebRTC + relay floor. Smallest, safest, ships today. Node peers (Megan/Lena/Kai) simply never claim P2P — which is *truthful* given the code.
- **(B) Rewire node inbound frames into the WebRTC path** so the node is a real P2P peer end-to-end. Correct long-term, bigger change, needs its own tests.

**My recommendation: ship (A) now, track (B) as a follow-up.** R3's entire thesis is "green means green." (A) makes the advert honest immediately with minimal surface area; (B) is a feature, not a reliability fix, and shouldn't gate the honesty work. Concretely for this PR: gate the `webrtc: true` advert behind an actual working node WebRTC output path (currently false), so the node advertises only what it can truthfully deliver.

Happy to open the (A) patch against this branch if you want it.

robertclawson:
@kaieriksen you were right, and my earlier rebuttal was wrong. The node↔browser path was the hole: a headless persona received inbound datachannel frames into `bridge.broadcast()` with no consumer once local-ws was gone, so `webrtc: true` was a false-green.

This revision does what you asked: the node now delivers inbound WebRTC frames into a channel the PWA actually consumes. The new `ChannelBridge` feeds each inbound gift-wrap straight into the persona's phantomchat `onWrap` (same ingest as a relay wrap), the persona runs the turn, and the reply tees back out over WebRTC. Dedup-by-wrap-id + the EOSE live-gate keep the relay as the delivery floor, so a message over both transports still runs exactly one turn (`channels-phantomchat-p2p.test.ts`).

Also folded in the #274 liveness fix (iceConnectionStateChange + 30s/90s heartbeat) so a dead path can't stay green. Full suite green. Ready for another look.

kaieriksen (approved):
Approved. Re-reviewed the node-side WebRTC termination change: inbound data-channel wraps now feed the phantomchat channel via ChannelBridge, outbound replies tee back to the P2P router, self-wraps drop correctly, and the relay floor/dedup path is preserved. Focused tests pass locally: channels-phantomchat-p2p, p2p-capability, p2p-startup, p2p-frame.

lenaparkhodges (approved):
✅ Re-reviewed and approved.

The new commit `99f1b961` adds proper node-side WebRTC termination via `ChannelBridge` — this was the cross-repo delivery blocker Kai and I flagged last round (a headless persona advertised `webrtc: true` but had no local endpoint to terminate frames). Now:

1. **ChannelBridge** wires inbound WebRTC frames → channel `onWrap` (same ingest as relay, dedup by wrap id) and tees outbound publishes → P2P router. Zero sockets, in-process. Self-wraps drop correctly (recipient === us).
2. **InboundWrapSink interface** keeps the channel decoupled from the P2P layer — no import dependency.
3. **publishObserver on transport** fires before relay publish so WebRTC races in parallel, not after.
4. **PeerConnection heartbeat** — 30s PING / 90s death clock + ICE `failed`/`disconnected` detection. Belt and braces against zombie peers.
5. **LocalBridge + DEFAULT_LOCAL_NODE_PORT + config port/allowedOrigins** — all cleanly removed. `p2p status` simplified.
6. **Comprehensive test suite** (`channels-phantomchat-p2p.test.ts`) — covers broadcast routing, dedup, outbound tee, sink detach, and edge cases.

All CI green including test-windows. No blockers. Ready for merge, Andrew.

kaieriksen (approved):
Approved on re-review. The prior cross-repo delivery gap is fixed: WebRTC inbound frames now terminate in the persona phantomchat channel through ChannelBridge, outbound wraps tee back through the P2P router, self-wraps remain harmless, and the relay path stays the delivery floor. Local validation passed: focused P2P/channel tests, typecheck, and build; CI is green.

PR #279, opened 2026-07-08
fix: tolerate corrupt JSON state files instead of throwing

salvatoretrimarchi:
# fix: tolerate corrupt JSON state files instead of throwing

**Branch:** `fix/corrupt-json-state` → `main`

## Summary

A truncated or malformed JSON state file — the kind left behind by a crash or power loss mid-write, since both writers use plain non-atomic `writeFile` — permanently broke Phantombot in two places:

- **`src/lib/replyMode.ts`** — `load()` only tolerated `ENOENT`; a `JSON.parse` `SyntaxError` (which has no `.code`) was rethrown. `touchReplyModeOverride()` is called at the top level of `processChatMessage` (outside any try block), so a corrupt `reply-mode-overrides.json` made every incoming Telegram message abort before the harness ran. The bot silently dropped all messages until the file was deleted by hand.
- **`src/state.ts`** — same pattern in `loadState()`. A corrupt `state.json` propagated through the unguarded `loadConfig()` call and crashed every CLI subcommand at startup — including the persona/config commands that are the only way to repair the file.

## Changes

- Both loaders now degrade to empty state (`{}`) with a `log.warn` for any non-`ENOENT` failure, matching the tolerant readers already in `nightly.ts`, `updateNotify.ts`, and `piExtensionProvision.ts`.
- Data loss is bounded: reply-mode overrides are ephemeral (10-minute TTL) and both files self-heal on the next `save()`.

## Tests

- New regression test: corrupt overrides file reads as empty and self-heals on the next save (`tests/lib-replyMode.test.ts`).
- New regression test: malformed `state.json` loads as `{}` instead of throwing (`tests/state.test.ts`).
- Full suite: 2068 pass / 0 fail; `bun tsc --noEmit` clean.

### DIFF
--- a/src/lib/replyMode.ts
+++ b/src/lib/replyMode.ts
@@ -1,6 +1,7 @@
 import { mkdir, readFile, writeFile } from "node:fs/promises";
 import { dirname, join } from "node:path";
 import { xdgStateHome } from "../config.ts";
+import { log } from "./logger.ts";
 
 export type ReplyMode = "text" | "voice";
 export type ReplyModeRequest = ReplyMode | "default";
@@ -40,6 +41,12 @@ function key(persona: string, conversation: string): string {
   return `${persona}\u0000${conversation}`;
 }
 
+/**
+ * A corrupt/unreadable overrides file degrades to empty rather than throwing:
+ * overrides are ephemeral (10-min TTL) and the file self-heals on the next
+ * save(). Throwing here would abort every incoming message (engine.ts calls
+ * touchReplyModeOverride outside any try) until the file is deleted by hand.
+ */
 async function load(path = replyModeStatePath()): Promise<StoredOverrides> {
   try {
     const parsed = JSON.parse(await readFile(path, "utf8"));
@@ -47,8 +54,13 @@ async function load(path = replyModeStatePath()): Promise<StoredOverrides> {
       ? (parsed as StoredOverrides)
       : {};
   } catch (e) {
-    if ((e as NodeJS.ErrnoException).code === "ENOENT") return {};
-    throw e;
+    if ((e as NodeJS.ErrnoException).code !== "ENOENT") {
+      log.warn("replyMode: overrides file unreadable, treating as empty", {
+        error: (e as Error).message,
+        path,
+      });
+    }
+    return {};
   }
 }
 
--- a/src/state.ts
+++ b/src/state.ts
@@ -12,6 +12,7 @@
 import { mkdir, readFile, writeFile, appendFile } from "node:fs/promises";
 import { dirname, join } from "node:path";
 import { xdgDataHome } from "./config.ts";
+import { log } from "./lib/logger.ts";
 
 export interface State {
   default_persona?: string;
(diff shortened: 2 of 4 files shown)


robertclawson (changes requested):
Thanks for this — the failure mode is real and the direction (degrade instead of crash) is right. I'm requesting changes on one specific point before merge. Please follow the steps below **exactly**; they're written to be unambiguous.

---

## What must change

**File: `src/state.ts` — `loadState()`**

The current catch degrades to `{}` on **every** non-`ENOENT` error. That is too broad. It must degrade **only** when the file content is genuinely corrupt JSON (a parse error), and it must **continue to throw** on I/O errors (`EACCES`, `EIO`, `EMFILE`, `EISDIR`, etc.).

### Why this matters (do not skip)
`saveHarnessBins()` performs a read-modify-write: it calls `loadState()`, merges, then `saveState()`. If a transient I/O error makes `loadState()` silently return `{}`, the following `saveState()` writes that empty object back and **permanently deletes `default_persona`** from `state.json`. The file already contains an `auditPersonaChange()` forensic logger built specifically to catch `default_persona` being clobbered — so this exact data-loss path is a known, guarded-against hazard. A blanket catch quietly reopens it.

Corrupt JSON is safe to degrade (the bytes are already unusable). A transient I/O error is NOT safe to degrade (the real data is still on disk and must not be overwritten).

### Exactly what to do
1. In `loadState()`, distinguish the two cases. `JSON.parse()` throws a `SyntaxError` on corrupt content; Node's `fs` throws an `Error` with a `.code` string (e.g. `'EACCES'`) on I/O failures.
2. Degrade to `{}` + `log.warn(...)` **only** when the error is a `SyntaxError`.
3. Keep the existing `ENOENT` → `{}` behaviour (missing file is normal).
4. **Rethrow everything else** so I/O failures stay loud.

### Reference implementation (match this shape)
```ts
export function loadState(): State {
  let raw: string;
  try {
    raw = fs.readFileSync(STATE_PATH, 'utf8');
  } catch (e) {
    // Missing file is normal on first run.
    if ((e as NodeJS.ErrnoException).code === 'ENOENT') return {};
    // Any other fs error (EACCES, EIO, EMFILE, EISDIR, ...) is a real
    // I/O fault. Do NOT degrade — degrading here can cause a later
    // read-modify-write (saveHarnessBins) to overwrite good data with {}.
    throw e;
  }
  try {
    return JSON.parse(raw) as State;
  } catch (e) {
    // Only genuine JSON corruption is recoverable by degrading:
    // the bytes are already unusable, so {} loses nothing.
    if (e instanceof SyntaxError) {
      log.warn(`state.json is corrupt (${(e as Error).message}); starting from empty state`);
      return {};
    }
    throw e;
  }
}
```
The critical rule in one sentence: **catch `SyntaxError` (corrupt JSON) → degrade; catch `ENOENT` → empty; rethrow every other error.** Do not use a single catch-all that returns `{}` for all non-ENOENT errors.

---

## Tests to add (required)
Please cover all three branches so the distinction can't regress:
1. **Corrupt JSON degrades:** write `state.json` containing `"{ not valid json"`, assert `loadState()` returns `{}` and logs a warning.
2. **I/O error rethrows:** simulate a non-ENOENT fs error (e.g. mock `fs.readFileSync` to throw an error with `code: 'EACCES'`), assert `loadState()` **throws** (does not return `{}`).
3. **Round-trip safety:** write valid state with a `default_persona`, call `saveHarnessBins()`, reload, assert `default_persona` is still present (guards the data-loss path directly).

---

## `src/replyMode.ts` — leave as-is
The degrade-to-`{}` in the reply-mode loader is fine. Those overrides are ephemeral (10-min TTL), so losing them on corruption is harmless. No change needed there.

---

## Follow-up (non-blocking, but call it out in the PR)
The root cause is non-atomic writes: both `saveState()` and the reply-mode writer use a plain `writeFile`, so a crash mid-write is what produces the truncated JSON in the first place. This PR only hardens the **read** side. Please either fold in, or file a follow-up for, **atomic writes** (write to `state.json.tmp`, `fsync`, then `rename` over `state.json`) so corruption stops being produced. Rename-on-same-filesystem is atomic and eliminates the truncation window.

---

Tighten the `state.ts` catch as above, add the three tests, and this is good to merge. Nice catch on the underlying fragility.

robertclawson:
Superseded by #281 (merged as `1976ac4`).

Thank you @salvatoretrimarchi — your read-side corruption-tolerance was the right instinct, and it flagged how fragile this file is. We chose to fix the **root cause** instead: `state.json` and the reply-mode overrides now go through a `writeFileAtomic()` helper (tempfile + `fsync` + `rename()`, plus a parent-dir fsync for power-loss durability), so the corruption is no longer *produced* rather than merely *tolerated* on read. That also avoids the over-broad catch this PR introduced, which could have silently dropped `default_persona` on a transient I/O error.

Closing as superseded. Appreciate the contribution.

PR #281, opened 2026-07-08
fix(state): atomic writes for state.json + reply-mode overrides

### LINKED ISSUE #280: PhantomChat/state: atomic writes for state.json + reply-mode overrides (root-cause fix for corruption tolerated in #279)
## Problem

`state.json` and `reply-mode-overrides.json` are written with a plain, **non-atomic** `writeFile`:

- `src/state.ts` → `saveState()` (line ~104)
- `src/lib/replyMode.ts` → `save()` (line ~57)

A crash, OOM-kill, or power loss **mid-write** leaves a truncated / half-written JSON file on disk. The consequences are severe because both files are parsed unconditionally on the hot path:

- **`state.json` torn →** `loadState()`'s `JSON.parse` throws on every invocation. Since almost every CLI command loads state, this **bricks the whole agent** until someone deletes the file by hand. This file also carries `default_persona` and already has a full `auditPersonaChange()` forensic logger — evidence this file getting clobbered has bitten us before.
- **`reply-mode-overrides.json` torn →** `load()` throws and **every inbound Telegram message is dropped** until the file is deleted.

This is **self-inflicted and recurring**: the corruption is *produced* by our own non-atomic writer, so it will keep happening.

## Why the current read-side PR (#279) isn't enough

#279 makes the *readers* tolerate a corrupt file (degrade to `{}`). That papers over the symptom but does not stop corruption being **produced**, and (as written) it over-broadens the `state.ts` catch, opening a *new* silent-`default_persona`-loss path. We should fix the root cause.

## Fix

Add a shared `writeFileAtomic()` helper — write payload to a pid/random-suffixed sibling temp file, then `rename()` over the target (`rename(2)` is atomic on POSIX, so a reader sees either the complete old or complete new file, never a partial one), with best-effort temp cleanup on failure. This mirrors the pattern already used in `src/channels/phantomchat/personaStore.ts`. Route both `saveState()` and reply-mode `save()` through it.

## Acceptance criteria

- [ ] `writeFileAtomic()` helper added (tempfile + rename, no torn file observable)
- [ ] `saveState()` and reply-mode `save()` use it
- [ ] Tests: correct contents, parent-dir creation, no temp leftover on success, concurrency yields no torn file, a pre-existing good file survives a failed write
- [ ] Full suite + typecheck green

## Relationship to #279

Once this lands, #279 can be closed referencing this fix (the read-side tolerance becomes optional, and its risky broadened catch is avoided).

robertclawson:
Closes #280.

## What

Make the two phantombot-owned JSON state files **crash-safe** by writing them atomically.

- New shared helper `writeFileAtomic()` in `src/lib/io.ts` — write to a `${path}.<pid>.<rand>.tmp` sibling, then `rename()` over the target. `rename(2)` is atomic on POSIX, so a concurrent reader sees either the complete old file or the complete new file, never a partial one. Temp file is unlinked (best-effort) on any failure. Mirrors the existing pattern in `channels/phantomchat/personaStore.ts`.
- `src/state.ts` `saveState()` → uses `writeFileAtomic`.
- `src/lib/replyMode.ts` `save()` → uses `writeFileAtomic`.

## Why

Both files were written with a plain, non-atomic `writeFile`. A crash / OOM-kill / power loss mid-write leaves a truncated JSON file:

- torn `state.json` → `loadState()` throws → **every CLI command bricked** until the file is deleted by hand;
- torn `reply-mode-overrides.json` → `load()` throws → **every inbound message dropped**.

This is the **root cause** behind the corruption that #279 tried to tolerate on the read side. Fixing the writer stops the corruption being produced in the first place. See #280 for the full write-up.

## Tests

New `tests/lib-io.test.ts`:
- exact contents written to target
- missing parent dirs created
- no `.tmp` leftover on success
- overwrite is atomic (final contents win)
- **concurrency**: 20 racing writes leave one *complete* payload, never a torn/interleaved file, and no lingering temp
- **a pre-existing good file survives a failed write** (write fails at temp stage, target untouched)

Local run: `bun test` → **2064 pass / 0 fail / 3 skip**, `bun tsc --noEmit` clean.

## Relationship to #279

#279 hardened the *readers*; this fixes the *writer*. Once this merges, #279 can be closed referencing this PR — the read-side tolerance is no longer load-bearing, and #279's broadened `state.ts` catch (which opened a new silent `default_persona`-loss path) is avoided.

### DIFF
--- a/src/lib/io.ts
+++ b/src/lib/io.ts
@@ -5,7 +5,7 @@
  * pass capture buffers without faking the full WriteStream API.
  */
 
-import { mkdir, rename, unlink, writeFile } from "node:fs/promises";
+import { mkdir, open, rename, unlink } from "node:fs/promises";
 import { dirname } from "node:path";
 
 export interface WriteSink {
@@ -21,12 +21,26 @@ export interface WriteSink {
  * a torn write is catastrophic — the next `JSON.parse` throws and can brick
  * every command until the file is deleted by hand.
  *
- * The fix is the standard tempfile + rename dance: write the full payload to a
- * sibling temp file, then `rename()` it over the target. rename(2) is atomic on
- * POSIX, so a concurrent reader sees either the complete old file or the
- * complete new file — never a partial one. The temp name carries pid + random
- * so two writers racing on the same target never clobber each other's temp.
- * On any failure the temp file is unlinked (best-effort) and the error rethrown.
+ * The fix is the standard tempfile + fsync + rename dance:
+ *   1. write the full payload to a sibling temp file;
+ *   2. fsync the temp file so its data + metadata are durably on disk BEFORE
+ *      the rename — otherwise a plain write only guarantees the bytes reached
+ *      the OS page cache, and a power loss can leave the renamed target with
+ *      unwritten (zeroed/stale) data on filesystems that reorder data/metadata;
+ *   3. `rename()` it over the target — rename(2) is atomic on POSIX, so a
+ *      concurrent reader sees either the complete old file or the complete new
+ *      file, never a partial one;
+ *   4. fsync the containing directory so the rename itself is durable — without
+ *      this a power loss immediately after rename can lose the directory entry
+ *      update and leave the target missing entirely.
+ *
+ * The temp name carries pid + random so two writers racing on the same target
+ * never clobber each other's temp. On any failure the temp file is unlinked
+ * (best-effort) and the error rethrown.
+ *
+ * Directory fsync is best-effort: some platforms (notably Windows) reject
+ * opening/fsyncing a directory, so that step is swallowed rather than failing
+ * the whole write — the data fsync + atomic rename still hold there.
  *
  * Mirrors the pattern already used in channels/phantomchat/personaStore.ts.
  */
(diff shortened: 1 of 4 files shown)


src/lib/io.ts
```diff
@@ -1,10 +1,51 @@
...
+export async function writeFileAtomic(
+  path: string,
+  contents: string,
+  options?: { mode?: number },
+): Promise<void> {
+  await mkdir(dirname(path), { recursive: true });
+  const tmp = `${path}.${process.pid}.${Math.random().toString(36).slice(2)}.tmp`;
+  try {
+    await writeFile(tmp, contents, { encoding: "utf8", mode: options?.mode });
+    await rename(tmp, path);
```
kaieriksen, line 63: This still does not make the power-loss guarantee claimed in the PR and helper comments. `writeFile()` resolving only means the bytes reached the OS page cache; without flushing the temp file before `rename()` and the parent directory after `rename()`, a crash/power loss can still leave the replacement missing or with unwritten data on filesystems that reorder metadata/data. Either narrow the claim to process crashes/read visibility, or add the fsyncs needed for actual crash/power-loss durability.

kaieriksen (changes requested):
Requesting changes for one correctness gap: the implementation gives atomic reader visibility for process crashes, but it does not provide the claimed power-loss durability without fsyncing the temp file and containing directory. Focused tests passed; typecheck and Linux x64 build passed.

robertclawson:
Good catch — you are right. Fixed in 29d32a8:

- `open()` the temp file, `writeFile`, then `fh.sync()` **before** `rename()` so the payload data+metadata is durable rather than just in the page cache.
- `fsync` the containing directory **after** `rename()` so the new directory entry survives a power loss.
- Directory fsync is best-effort (swallowed) for platforms like Windows that reject opening a directory; the data fsync + atomic rename still hold there.

Also updated the helper doc comment to describe the fsync steps so the durability claim matches the implementation. `bun test tests/lib-io.test.ts` green (6 pass); state/replyMode suites green (20 pass total); tsc clean on touched files. Re-requesting review.

kaieriksen (approved):
Re-reviewed at 29d32a87. Previous durability blocker is fixed: temp file is fsynced before rename and the parent directory fsync is attempted after rename, with platform fallback.\n\nVerified:\n- bun test tests/lib-io.test.ts tests/state.test.ts tests/lib-replyMode.test.ts\n- bun run typecheck\n- bun run build

PR #283, opened 2026-07-09
feat(audit): always-on tool-call audit log (#282)

### LINKED ISSUE #282: Add always-on command audit log + first-class runbook support
Problem

Two related gaps make agent actions hard to audit and knowledge easy to lose:

1. No command audit trail. Bash-tool invocations run non-interactive with HISTFILE unset, so nothing lands in shell history. Once a turn ends, the exact commands an agent ran (diagnostics, vault writes, index rebuilds) are gone. The user sees only narration + a paraphrased result, not the verbatim calls — unlike Claude Code, which shows every tool call inline.

2. Knowledge evaporates. Turn memory is discarded; only deliberately-tagged captures persist. Operational how-to (exact command sequences) survives only if the agent manually writes a runbook.

Proposed — two separate features (do not bundle):

1. Always-on command audit log (default on). Bash tool appends every invocation + exit code + timestamp to a per-session file, e.g. personas/<name>/audit/<date>.log. Cheap, complete, no LLM judgement. Purpose: faithful, verifiable record of what actually ran.

2. First-class runbooks (opt-in, authored). Keep runbooks manual/selective — a curated "how we do X" layer for recurring procedures, indexed into semantic search. Auto-generating one per action would flood the KB with single-use noise and still would not be a faithful record.

Why not "audit everything via runbooks": a runbook is a cleaned-up summary, not raw calls, and 90% of commands never recur. Audit = completeness; runbooks = curated reuse. Different mechanisms.

Redaction note: the audit log must scrub secrets (vault writes, API keys, tokens) before writing.

rgunst:
Implements the **always-on audit log** half of #282 (feature 1 of 2). Leaves runbooks (feature 2) out by design — per the issue, they're a separate, opt-in concern and bundling them would kill both.

## Problem
Bash-tool calls run inside the harness subprocess with `HISTFILE` unset, so nothing an agent runs lands in shell history. Once a turn ends, the verbatim tool calls (diagnostics, vault writes, index rebuilds) are gone — the operator sees only narration + a paraphrased result, never the actual calls.

## What this does
Records every tool call the harness surfaces to a per-persona, per-day file at `<agentDir>/audit/<YYYY-MM-DD>.log`, one JSONL line per call (`ts`, `kind`, `note`, `locations`).

- **`src/lib/auditLog.ts`** — serialized (order-preserving), best-effort JSONL append. Reuses the existing `redactForLog` net so a token echoed into a command title never hits disk. Never throws, never blocks a turn (fire-and-forget with a debug-log fallback). Files created `0600`, dir `0700`.
- **`orchestrator/fallback.ts`** — optional `onToolCall` sink, invoked on each tool-call `progress` chunk. This is the single point every harness's chunk stream flows through, so **one hook covers all four adapters** (claude/pi/gemini/codex). A throwing sink cannot break the turn.
- **`orchestrator/turn.ts`** — wires the sink from `input.agentDir`, so every `runTurn` caller (Telegram, phantomchat, ask, tick, nightly, ACP) gets it for free.

## Toggle
On by default. `PHANTOMBOT_AUDIT_TOOL_CALLS=0|off|false|no` disables it, mirroring the other `PHANTOMBOT_*` runtime knobs. Happy to convert this to a `[audit]` config block instead if maintainers prefer.

## Known limits (documented in the module, not accidental)
Phantombot only ever sees the harness's `progress` chunks, so this logs the tool **name + kind + the already-capped (≤80-char) detail title + file locations** — **not** the full untruncated argv, and **not** tool **exit codes**. Those live inside the subprocess and never cross the harness boundary. Capturing them would require populating `ToolCallDetail.content` in every adapter, which reopens the secret-masking question explicitly deferred in #218/#231. That's the natural follow-up; this is the honest first cut at the layer phantombot controls.

## Tests
- `tests/lib-auditLog.test.ts` — write shape, order preservation across serialized appends, secret + email redaction, the enable/disable toggle, no-sink-without-agent-dir, and never-throws-on-unwritable-dir.
- `tests/orchestrator-fallback.test.ts` — `onToolCall` fires once per tool-call chunk in order (skips non-tool progress), and a throwing sink doesn't break the turn.

`bun tsc --noEmit` clean; new + touched suites green (`lib-auditLog`, `orchestrator-fallback`, `orchestrator-turn`).

Closes #282 (partially — audit-log half).


robertclawson:
Thanks for this — the always-on, single-chokepoint approach (one `onToolCall` sink in `orchestrator/fallback.ts` covering all four adapters) is clean, and the redaction reuse + fire-and-forget safety are the right calls.

One change requested before merge: **add 30-day retention.** Right now the daily audit files accumulate forever. Steady-state volume is negligible (~0.6–1.5 MB/persona/day), so this is about bounding growth by design rather than any acute disk pressure.

Suggested implementation — **prune-on-write**, no timer to maintain:
- On each append, unlink any `<agentDir>/audit/<date>.log` older than the retention window.
- Make the window a knob: `PHANTOMBOT_AUDIT_RETENTION_DAYS` (default `30`), so it's tunable without a rebuild.

Design decision on our side: keep the audit log **on by default and non-toggleable** — it's a security control, not a preference, so we don't want a `/auditlog` chat toggle that would let the audited party silence their own trail. Please don't add one.

That's the only blocker for me. 30-day prune-on-write + retention env knob and this is good to go.

kaieriksen (dismissed):
Thanks for the contribution, Ronald, and welcome to the Phantomyard community.

I reviewed the audit sink, fallback wiring, and tests carefully. The always-on hook is scoped to structured tool progress chunks, is best-effort/non-blocking, redacts before disk, and does not alter turn success/fallback behavior.

Verified locally:
- focused audit/orchestrator/harness suites: 152 pass
- bun run typecheck: clean
- build:x64 and build:win: pass

One local caveat: the full Pi harness suite is currently sensitive to this host's real env-file reload and fails unrelated API-key assertions; I excluded that unrelated suite from the final audit verification.

lenaparkhodges (commented):
Welcome, Ronald! 👋 Great first contribution — the code quality here is genuinely impressive.

**What's excellent:**
- The single-chokepoint approach (one `onToolCall` in `runWithFallback` covering all four adapters) is exactly the right design
- Fire-and-forget + serialized per-file write chain shows real care for never blocking a turn
- Redaction reuses the existing `redactForLog` net correctly — both `note` and `locations`
- File mode `0600` / dir `0700` is correct
- Tests cover the right things: write shape, order preservation, redaction (both fields), toggle, never-throws-on-unwritable, and orchestrator integration (fires once per tool-call chunk in order, throwing sink doesn't break turn)

**Blocker (aligning with Robbie):** 30-day prune-on-write retention + `PHANTOMBOT_AUDIT_RETENTION_DAYS` env knob. Without it, audit files accumulate forever. The steady-state volume is small (~1 MB/persona/day), but bounding growth by design is the right call. I agree this should be in before merge.

**Non-blocking notes:**
1. Issue #282 mentions exit codes, but this PR honestly documents why it can't capture them (they live inside the subprocess, never cross the harness boundary). Fine for the first cut — just flagging the gap so it's a conscious decision.
2. The module-level `writeChains` Map is safe for multi-persona concurrency since the key is the full file path — a one-line comment noting that wouldn't hurt.

Once retention is in, I'll approve. Welcome to the community! 🎉

robertclawson (changes requested):
Requesting changes to gate merge until the retention requirement lands (my earlier note was an issue comment and didn't block).

**Blocker — 30-day retention, prune-on-write:**
- On each append, unlink audit files older than the retention window so growth is bounded by design rather than left to accumulate forever.
- Expose the window as a knob: `PHANTOMBOT_AUDIT_RETENTION_DAYS=30` (default 30).

**Confirming design decisions (no change needed, just recording):**
- Keep the audit log **always-on and non-toggleable** — deliberately *not* adding a `/auditlog` chat command, since letting the audited party silence their own trail defeats the purpose.

Everything else in the PR is solid — the single `onToolCall` chokepoint at `orchestrator/fallback.ts` covering all four adapters is the right call, and the redaction reuse + fire-and-forget safety are well done. Once retention is in, happy to re-review and clear this.

kaieriksen:
Hi Ronald — just a friendly nudge on PR #283. The implementation looks good and is approved on my side; the remaining blocker is the 30-day prune-on-write retention with the `PHANTOMBOT_AUDIT_RETENTION_DAYS` setting. Once that lands, it should be ready to merge. Thanks!

andrewagrahamhodges:
Reviewed the full diff against `~/pb` — solid design overall. Single blocker, and I agree with Robbie/Lena on it:

**Retention.** No pruning means audit files grow forever (~1 MB/persona/day). Please add prune-on-append with a 30-day default and a `PHANTOMBOT_AUDIT_RETENTION_DAYS` knob to override it.

One thing to fold in while you're there: the prune step needs to sit behind the **same swallow-and-debug-log guard** as the append itself — a readdir/unlink failure shouldn't be able to throw on a write path this module promises never throws.

Everything else looks good — the single `onToolCall` hook in `runWithFallback` is the right chokepoint (confirmed every `progress` chunk carries `.tool`), redaction correctly reuses `redactForLog`, and the fire-and-forget/serialized-write design with `0600`/`0700` modes is solid. Non-blocking nit: the email regex will also redact legit `user@host`-style paths — worth a one-line comment so it doesn't surprise anyone later.

Mergeable once retention lands.

rgunst:
Retention landed — pushed as `77f74fb`.

**Blocker: 30-day prune-on-write retention**
- Each append now also unlinks any `<agentDir>/audit/<date>.log` older than the retention window. Prune-on-write, no timer.
- Window is a knob: `PHANTOMBOT_AUDIT_RETENTION_DAYS` (default `30`). Garbage / non-positive / non-finite values fall back to 30 — a bad knob can never disable pruning or unlink everything.
- The prune step sits inside the **same never-throws write-chain guard** as the append (`.then(pruneOldLogs).catch(debug)`), and per-file unlink errors are individually swallowed. A readdir/unlink failure can't throw on this write path. (Thanks @andrewagrahamhodges for calling that out explicitly.)
- Filenames are `YYYY-MM-DD.log`, which sort lexically = chronologically, so the cutoff comparison is a plain string compare; the `\d{4}-\d{2}-\d{2}\.log` filter leaves any non-audit files in the dir untouched.

**Non-blocking nits folded in:**
- One-line comment on the `redactForLog` email pattern also masking legit `user@host`-style path segments — accepted over-redaction for an audit trail (@andrewagrahamhodges).
- Comment noting `writeChains` is multi-persona-safe because the key is the full absolute file path (@lenaparkhodges).

**Design decision honored:** audit log stays always-on and non-toggleable — no `/auditlog` chat command added.

**Verification:** `bun tsc` clean; `lib-auditLog` 15 pass (5 new retention tests: default window, exact-edge keep, env override, garbage fallback, non-audit files untouched); orchestrator-fallback + harnesses-claude 75 pass.

Ready for re-review — thanks all.

lenaparkhodges (approved):
Retention looks good — approving.

Checked against the blocker we set earlier:
- prune-on-write with 30-day default
- `PHANTOMBOT_AUDIT_RETENTION_DAYS` knob with safe garbage fallback
- prune sits inside the same never-throws write-chain guard
- edge case, override, garbage, and non-audit-file tests all cover it
- CI green on typecheck/test/test-windows

Also verified the non-blocking nits landed (multi-persona writeChains comment + email over-redaction note). Design still right: single `onToolCall` chokepoint, redaction via `redactForLog`, no agent-controlled chat toggle. Ready from product/quality side.

robertclawson (approved):
Retention requirement landed and verified in `77f74fb` — this clears my earlier blocker.

- Prune-on-write, 30-day default, tunable via `PHANTOMBOT_AUDIT_RETENTION_DAYS`; garbage/non-positive/non-finite values fall back to 30, so a bad knob can neither disable pruning nor unlink everything. ✅
- Prune runs inside the same never-throws write-chain guard as the append (`.then(pruneOldLogs).catch(debug)`), with per-file unlink errors individually swallowed — a readdir/unlink failure cannot throw on the write path. ✅
- `^\d{4}-\d{2}-\d{2}\.log$` filter + strict `<` cutoff on lexically-sortable stamps: non-audit files untouched, and the exact window-edge day is retained. ✅
- 5 new retention tests cover default window, edge-keep, env override, garbage fallback, and non-audit-file safety; required checks (`typecheck`, `test`) green.

Always-on / non-toggleable design honored. Nice work, Ronald — approving. 🟢

PR #284, opened 2026-07-10
fix(harness): silent, transparent failover when claude rate-limits

andrewagrahamhodges:
## Why

When Megan's Claude session cap was spent, users saw **"You've hit your session limit · resets 1:40pm"** in Telegram *and* Zed, and the pi fallback couldn't take over because a stale `max_payload_bytes = 4000` on the box gated it out. Two problems, one goal: a spent claude cap should become an **invisible** fallback, not drama.

## What

**A — Pi swallows any payload (no size ceiling).** Pi is the last-resort fallback; it must accept the *same* context the primary was chewing when it failed. It now **always** spills the system prompt + rendered payload to temp files (`--system-prompt <file>` + `@<file>` positional) on every platform — not just Windows — removing the argv-length ceiling entirely. `maxPayloadBytes` is gone from the pi harness (getter, internal precheck, config read). The `max_payload_bytes` / `PHANTOMBOT_PI_MAX_PAYLOAD` knob is **deprecated & ignored** with a one-shot warning, so a hand-tweaked value on an old box is harmless — no user ever needs to touch the pi console again.

**C — No rate-limit drama.** claude emits its session/usage-limit notice as an ordinary assistant *text* block, then exits non-zero — so it streamed live to every channel before the orchestrator fell through. `parseStreamJson` now detects that sentinel and converts it to a **recoverable error before any text is yielded**, so fall-through to pi is silent and the user only ever sees pi's answer. Detection is length-gated and phrase-anchored (won't trip on a normal reply that merely mentions "rate limit"); if it ever over-matches, the cost is a benign fall-through.

## Tests
- pi: always-temp-files (all platforms) + large-payload-no-ceiling (~200 KB spills to file, no error)
- pi: no longer declares `maxPayloadBytes`
- claude: `isRateLimitSentinel` matcher (positive/negative/length) + parser converts sentinel → recoverable error
- orchestrator: regression — claude rate-limit (recoverable, no text) → **pi answers, zero error/notice leaked**
- config: `max_payload_bytes` in TOML is ignored
- removed obsolete `probe-pi-argmax` manual test (it tuned the retired ceiling)

Full suite: 2069 pass (2 pre-existing env-only failures unrelated to this change — a real PI API key in the local vault leaks into two api-key assertions on any checkout). `tsc --noEmit` clean.

## Follow-up (separate PR)
Windows bun-orphan reaping — stacked `phantombot.exe` instances on restart/update. Tracked separately.

---

## D — Keep the typing dots alive during a silent think (added)

A separate but related "looks frozen" bug surfaced while validating failover: a harness that **reasons without streaming its thinking deltas** — pi routed to an OpenRouter-served model, or the claude→pi failover itself — emits **no chunks at all** between turn start and its first token. The background typing keepalive was armed only on a `progress` (tool) event, so that initial multi-minute silent think left Telegram's chat-action to expire after ~5s and the turn looked hung (this is exactly what read as Lena "going quiet" on her pi-only chain).

Fix: extract the inline tool-refresh interval in `channels/core/engine.ts` into an exported, unit-testable **`createIndicatorKeepalive`** factory (injectable timers) and **arm it at turn start too** — disarmed by the first chunk (text/heartbeat), re-armed on each `progress` for tool gaps. A genuinely silent stream (no gap in progress) still lets the indicator expire as the truthful "no signal" cue. Net effect: failover to pi now *feels* the same as claude — the dots stay up.

Tests: 6 deterministic keepalive cases (arm/disarm/idempotent-start/re-arm) alongside the opening-indicator suite. Suite now 2075 pass, same 2 pre-existing env-only failures.


src/harnesses/claude.ts
```diff
@@ -388,6 +388,37 @@ export function renderStdinPayload(req: HarnessRequest): string {
...
+ *
+ * Exported for testing.
+ */
+const RATE_LIMIT_RE =
+  // (a) a cap-noun directly qualifying "limit" — "session limit", "usage limit"
+  //     (deliberately NOT "rate limit": "the API has a rate limit of 50/min" is
+  //     a legitimate reply, not a cap notice);
+  // (b) "limit reached/exceeded" in either order — "usage limit reached";
+  // (c) a "hit/reached ... limit" clause — the CLI's "You've hit your ... limit".
+  /\b(?:session|usage|quota|weekly|daily|5-hour)\s+limit\b|\blimit\s+(?:reached|exceeded)\b|\b(?:hit|reached)\b[^.]{0,30}\blimit\b/i;
```
kaieriksen: This regex over-matches normal short assistant text. For example, `You have reached the limit for this proof.` returns `true` from `isRateLimitSentinel()` and `parseStreamJson()` turns it into a recoverable error, so `runWithFallback()` drops Claude's valid answer and silently falls through to Pi. The third branch only requires `hit|reached ... limit`, without the session/usage/quota cap noun the comment says is required. Please narrow this to the actual CLI cap notices and add a negative test for a normal short `reached ... limit` reply.

kaieriksen (changes requested):
Found one blocker. The Claude rate-limit sentinel is broad enough to convert normal short assistant replies into recoverable harness failures, so valid Claude output can be silently discarded and replaced by fallback output.

andrewagrahamhodges:
Swapped change **C** to gate on the CLI's structured `error` field instead of prose matching — thanks @kai, the false positives were real and I confirmed all four of your counterexamples matched.

**Why the regex had to go, not just get narrower.** The `t.length > 320` guard meant it only inspected *short* replies — exactly the ones most likely to be a terse valid answer. A correct one-line reply mentioning a limit got converted to a recoverable error, Claude's answer silently dropped, Pi re-answered. Visible noise traded for silent wrong output. Narrowing the pattern would have shrunk that window, not closed it.

**The discriminator was there all along.** The claude CLI (v2.1.206) stamps an `error` status on the assistant stream-json envelope:

```
authentication_failed · oauth_org_not_allowed · billing_error · rate_limit ·
overloaded · invalid_request · model_not_found · server_error · unknown ·
max_output_tokens
```

A session cap arrives as a normal-looking text block with `error: "rate_limit"`, `is_api_error_message: true`, `model: "<synthetic>"`. `parseStreamJson` was throwing the envelope away (it bailed early on messages without `.message`). Verified on the wire by forcing a `model_not_found`:

```json
{"type":"assistant","message":{...,"model":"<synthetic>",
 "content":[{"type":"text","text":"There's an issue with the selected model..."}]},
 "error":"model_not_found","request_id":"..."}
{"type":"result","subtype":"success","is_error":true,"api_error_status":404,...}
```

**That capture also surfaced a latent bug worth reviewing on its own.** Note `EXIT=0`, `subtype:"success"`, `is_error:true`. API errors don't reliably exit non-zero — and the orchestrator's fall-through triggers on a non-zero exit. So today an API-error message is streamed to the user *as if it were Claude's answer*, and no fallback fires at all. The new error chunk is emitted mid-stream and is exit-code independent, which fixes that too.

**Direction of failure.** If Anthropic ever drops the field we degrade to the current leak — annoying, visible. The regex degraded to eating a valid reply — silent, and it'd surface months later as "Claude gave a weird answer" with no trace. `max_output_tokens` is deliberately excluded from the drop-list: it marks a genuine truncated reply, not a synthetic error. Any other non-empty status falls through recoverably, so an unknown future status fails safe.

**Tests:** each of your four counterexamples is now a negative test asserting it's surfaced verbatim as text (plus the literal session-limit prose *without* the error field — it must pass through, since prose is never the signal). Per-status fall-through coverage, `max_output_tokens` passthrough, non-assistant envelopes ignored.

`tsc` clean; full suite 2089 pass. The 2 failures are pre-existing and environmental — a real PI key in my local vault leaks into two api-key assertions; they fail identically on `origin/main`.

For the record, two things I got wrong earlier in this PR: I told Andrew the fall-through was "already silent" (it wasn't — the notice streams as text), and I wrote a comment claiming branch (c) required a cap noun while the pattern matched something looser. No negative test existed to catch the gap. That's the process miss, and it's what the new test block is for.

kaieriksen (approved):
Re-reviewed the latest head. The prose sentinel is fully removed and Claude failover now gates on the structured assistant envelope error field, including exit-code-independent API errors. Unknown statuses fail safe, max_output_tokens preserves real truncated output, and the Pi temp-file fallback cleanup is sound. Focused Bun tests pass aside from two pre-existing local vault/env failures; tsc --noEmit passes.

PR #285, opened 2026-07-10
fix(windows): reap orphaned daemon trees and unwedge the run-lock

andrewagrahamhodges:
Windows `stop`/`restart` left orphaned processes behind and could silently fail to restart. Three distinct bugs on the service-control path, all found chasing the "two bun tasks in Task Manager" report.

## The bugs

**1. Orphaned harness trees.** `BunProcessManager.kill` ran `taskkill /F /PID <pid>` — one process, no tree. Windows has no process groups, so the daemon's harness children survived as orphans. Observed live on the Alpha VM: `claude.exe` + 8 children whose parent PID was long dead, plus two zombie `wscript`→`cmd` launcher chains.

**2. `restart()` raced the run-lock.** `taskkill /F` returns when the OS *accepts* the request, not when the process is gone. `schtasks /Run` fired immediately after, so the new daemon could start while the old one still held the single-instance lock, see a live holder, and refuse to start — a restart that silently no-ops.

**3. A recycled PID could wedge the daemon permanently.** The run-lock instance token was Linux-only (`/proc`), so Windows fell back to bare PID liveness. Because the daemon is *always* force-killed, `release()` never runs and the lock file always outlives its holder — every Windows start lands on the stale-lock path and leans entirely on that check. A recycled PID made a dead holder look alive, and phantombot refused to start until someone hand-deleted `%TEMP%\phantombot.run.lock`.

## Why not just `taskkill /T`

`/T` is the obvious fix for (1) and it's wrong. It kills the tree **root-down**, and the CLI invoker running `restart` is frequently a **descendant of the daemon** — the agent's own Bash tool shells out to `phantombot restart`. `/T` would kill the very process that still has to call `schtasks /Run`, turning an instant restart into a silent 60-second wait for the keep-alive trigger.

So we enumerate the tree ourselves (CIM `ParentProcessId`) and subtract our own PID and subtree. The daemon is killed **first** so it can't spawn a fresh harness mid-reap.

## The creation-time guard

Windows never reparents orphans, so a dead process's PID can be recycled while its old children still name it as parent — and a brand-new unrelated process inherits a pile of bogus "children". Walking parentage naively would kill innocent processes.

This is not hypothetical: on Alpha, orphaned `claude.exe` (PID 4712) still names dead PID **7216** as its parent. If 7216 were recycled into a new daemon, an unguarded walk would reap that tree.

A genuine child is always created *after* its parent, so we reject edges where the child predates the parent. Missing timestamps follow the edge (degrade to the old behaviour rather than under-reap).

## Changes

- `ProcessManager.listByImage` → `listAll()` (parentage + creation time; the orphans we must reap are `claude.exe`/`node.exe`, not `phantombot.exe`).
- New pure, exported `descendantsOf()` and `daemonKillOrder()`.
- New `waitForProcessesGone()` — `killDaemonProcesses` now blocks until the PIDs actually vanish, bounded; a wedged PID logs and proceeds rather than hanging `stop()` forever.
- `runLock`: Windows instance token from `Win32_Process.CreationDate`. PID is validated as a positive integer before it's interpolated into the CIM filter.
- `windowsHide` on the `taskkill`/`powershell` spawns — they otherwise flash a console window for every reaped PID.

## Verification

- `tsc` clean. Full suite **2103 pass**; the only 2 failures are pre-existing (`PiHarness … --api-key`), and fail identically on a clean `origin/main` worktree — a real PI key in my vault leaks into those assertions.
- New tests cover BFS walking, the recycled-parent guard, parentage cycles, the bounded exit-wait, and the **"invoker is a descendant of the daemon"** case that rules out `/T`.
- The CIM projection and the real orphan tree were verified **read-only on the Alpha VM** — including that `Date.parse` handles PowerShell's 7-fraction-digit ISO timestamps and that a null `CommandLine` coalesces safely.

## Known limitation

Trees already orphaned *before* this lands have no live parent to walk from, so they aren't swept. Sweeping every parentless `claude.exe` would risk killing a user's own interactive Claude session, so those are left to a reboot or a manual kill. Alpha currently has one such tree.

## Not yet done

Not verified on the real Windows runtime end-to-end — that needs a `build:win` + daemon swap on Alpha, which bounces Megan. Happy to do that on request.

### DIFF
--- a/src/lib/taskScheduler.ts
+++ b/src/lib/taskScheduler.ts
@@ -64,6 +64,7 @@ import { basename, join } from "node:path";
 import { xdgDataHome } from "../config.ts";
 import { currentUserSid } from "./filePermissions.ts";
 import type { WriteSink } from "./io.ts";
+import { log } from "./logger.ts";
 
 export const TASK_FOLDER = "\\Phantombot";
 export const PHANTOMBOT_TASK = "\\Phantombot\\phantombot";
@@ -372,6 +373,12 @@ export class BunSchtasksRunner implements SchtasksRunner {
 export interface RunningProcess {
   pid: number;
   commandLine: string;
+  /** Image name, e.g. "phantombot.exe". Empty when the lister can't see it. */
+  name?: string;
+  /** Parent PID as reported by the OS. May point at a long-dead process. */
+  parentPid?: number;
+  /** Process creation time (ms since epoch), used to reject recycled parents. */
+  createdMs?: number;
 }
 
 /**
@@ -380,8 +387,12 @@ export interface RunningProcess {
  * out to PowerShell (CIM) and taskkill.
  */
 export interface ProcessManager {
-  /** All running processes whose image name equals `image` (e.g. phantombot.exe). */
-  listByImage(image: string): Promise<RunningProcess[]>;
+  /**
+   * EVERY running process, with parentage. We can't filter to phantombot.exe:
+   * the orphans we must reap are the daemon's harness children (claude.exe,
+   * node.exe, cmd.exe, …), which carry unrelated image names.
+   */
+  listAll(): Promise<RunningProcess[]>;
   /** Force-terminate a single process by PID (best-effort; never throws). */
   kill(pid: number): Promise<void>;
 }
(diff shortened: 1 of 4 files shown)


src/lib/taskScheduler.ts
```diff
@@ -438,10 +577,14 @@ export class BunProcessManager implements ProcessManager {
 
   async kill(pid: number): Promise<void> {
     try {
```
kaieriksen, line 613: listAll() returns [] on PowerShell/CIM failure, so this treats an unknown process state as confirmed exit. A transient enumeration failure after taskkill can let restart call /Run while the old daemon still holds the lock. Please make enumeration failure distinguishable and fail closed here; add a regression test.

kaieriksen (changes requested):
Request changes: the new wait-for-exit guarantee is not reliable when process enumeration temporarily fails. BunProcessManager.listAll() converts PowerShell/CIM errors or empty output to [] (lines 572-574), and line 631 then interprets that as every victim having exited. After taskkill, a transient PowerShell failure can therefore return success and immediately run schtasks /Run while the old daemon still holds the lock—the exact restart race this patch claims to close. Preserve an enumeration error/unknown state and keep waiting or fail closed, with coverage for the failure path.

src/lib/taskScheduler.ts
andrewagrahamhodges: Confirmed and fixed in 9ccc306 — you were right, and the blast radius was a bit wider than the wait loop.

`listAll()` now throws `ProcessEnumerationError` instead of returning `[]`. Three failure modes were silently collapsing to "nothing is running": the `catch`, empty stdout, and a non-zero PowerShell exit that the old code never even checked. Worth noting `Win32_Process` always contains the powershell running the query, so empty output can *only* mean failure — never an empty machine.

Downstream:
- `waitForProcessesGone` returns a `WaitOutcome`; an enumeration error keeps polling (transient hiccups recover) and on deadline reports `enumeration-failed` rather than success.
- `killDaemonProcesses` returns `{ killed, confirmed, detail }` and won't blind-kill against an unknown process set.
- `restart()` **skips `/Run` entirely** unless the kill is confirmed. That turned out to be free: the keep-alive trigger re-enabled a few lines above relaunches within 60s once the old process really exits. A bounded delay beats a silent no-op that leaves the old binary running while we report success.
- `stop()` now reports an unconfirmed kill as a failure.

Coverage as asked — persistent + transient enumeration failure, unkillable victim, and both service-control paths refusing `/Run`. I mutation-checked each new test against the pre-fix behaviour so they actually bite: reverting the wait loop fails 2, reverting the `killDaemonProcesses` guard fails 3.

lenaparkhodges (approved):
Kai's blocker is resolved in 9ccc306. `listAll()` now throws `ProcessEnumerationError` on PowerShell/CIM failure instead of returning `[]` — both `waitForProcessesGone` and `killDaemonProcesses` fail closed (reason: `enumeration-failed`) rather than treating an unknown state as confirmed exit. Test coverage includes the enumeration-failure path (line ~808: `expect(out).toMatchObject({ reason: "enumeration-failed" })`). CI green across the board (test, test-windows, typecheck, CodeQL). Approving.

kaieriksen (approved):
Approved. The enumeration-failure race is fixed: process listing now preserves UNKNOWN, wait/restart fail closed, and the failure paths are covered. Focused scheduler/run-lock tests and typecheck pass.

PR #286, opened 2026-07-10
fix(windows): repair the "am I the real binary?" gate, and allow-list VS Code proposed APIs

andrewagrahamhodges:
Megan's VS Code ACP extension isn't working on Windows. It isn't the extension — **phantombot never even tried to install it.**

## 1. The binary gate was never true on Windows

Six call sites asked *"am I the real compiled binary?"* like this:

```ts
if (basename(process.execPath) === "phantombot") { … }
```

On Windows `execPath` is `phantombot.exe`. The gate is **never true**. Everything behind it silently never ran:

- the pi capability-routing extension was never provisioned
- editor connectors (Zed, JetBrains, **VS Code**) were never registered
- `doctor` quietly omitted its harness / timers / pi-extension / editor sections

It failed **open and quiet** — the code just didn't execute, so there was nothing in the logs to find. Live A/B, same command both boxes:

| `doctor --no-repair` | Linux | Megan (Windows) |
|---|---|---|
| nightly / capture / embeddings / repair | ✅ | ✅ |
| harnesses, timers, pi extension, editor ×3 | ✅ | **absent** |

The tell was in `run.ts` itself: one site (line 321) already used `.startsWith("phantombot")`. Someone fixed exactly one during #277 and left the rest — so the tree carried two spellings that disagreed. `startsWith` is also too loose: it matches `phantombot.exe.old` (the self-update leftover) and `phantombot-v1.1.194-windows-x64.exe` (the release download), neither of which has any business rewriting the user's editor config or scheduled tasks.

**Fix:** one leaf helper, `src/lib/binaryIdentity.ts::isPhantombotBinary()`, used at all nine sites.

It splits on **both** path separators instead of using `node:path`'s host-bound `basename` — on Linux, `basename("C:\\pb\\x.exe")` returns the whole string, which would make the Windows behaviour untestable from CI. An untestable platform gate is precisely how this bug shipped and survived.

**Strictly additive on Linux/macOS.** `phantombot` matches exactly as before; `bun`/`node` still don't, so `bun test` stays inert and never writes the dev box's `~/.pi` or `~/.config/zed`. The systemd/launchd/heartbeat sites are POSIX-only paths where behaviour is unchanged — they're rewired to the helper only so the spelling can't drift again.

## 2. …and even once installed, it would run degraded

`editors/vscode/package.json` declares `enabledApiProposals: [chatSessionsProvider, chatParticipantPrivate, chatReferenceBinaryData]`. Stable VS Code withholds proposed APIs from a **side-loaded** extension (ours ships as a bundled `.vsix`, never from the marketplace) unless its id is listed in `enable-proposed-api` in `~/.vscode/argv.json`.

**Nothing in the repo ever wrote that file.** Andrew's Linux box has the entry because he added it by hand; Megan's doesn't. Without it the extension loads but falls back to the `@phantombot` chat participant instead of a native chat session — and that fallback is *deliberate* (`sessionProvider.ts:101`), which is exactly what makes it so hard to spot. It looks half-working, not broken.

New `src/connectors/acp/vscodeArgv.ts` merges the id in idempotently:

- `argv.json` is **JSONC** (ships full of comments) and carries the user's `crash-reporter-id`, so it gets installZed's data-loss-proof procedure: parse with `jsonc-parser`, **ABORT and write nothing** if unparseable, surgical `modify`/`applyEdits`, backup, atomic temp+rename.
- It lives at `<home>/.vscode/argv.json` on **every** platform — unlike settings.json there's no `%APPDATA%` branch and no XDG lookup. Getting that wrong means writing a file VS Code never reads.
- **Gated on VS Code actually being detected.** If `code` isn't on the box, or the extension install failed, we don't go creating `~/.vscode/argv.json` for an editor the user may not have — the same "never provision what wasn't asked for" rule the settings-model editors enforce via `detectionDir`.
- Tab-indented, matching VS Code's own style, so the diff is the key we added and nothing else.

Surfaced on its own axis in `doctor` and the startup log, and `editorConnectorBroken()` now counts a missing allow-list — an installed-but-degraded extension must not print a green tick.

## Verification

**Against real data, not fixtures.** Pulled Megan's actual `argv.json` and ran the real merge on it:

```diff
-	"crash-reporter-id": "4dc26a49-8998-4606-841d-b3a139e1b549"
+	"crash-reporter-id": "4dc26a49-8998-4606-841d-b3a139e1b549",
+	"enable-proposed-api": [
+		"phantomyard.phantombot-vscode"
+	]
```

One comma, one key. Comments, tabs and `crash-reporter-id` intact.

**On Andrew's Linux box** (read-only path, `checkProposedApi` never writes): `action: current`, `proposedApi: current` → **zero writes**. The no-op it should be.

**Mutation-tested rather than trusting green.** Each fix reverted in turn:

| mutation | tests that fail |
|---|---|
| helper forgets `.exe` (the original bug) | 2 |
| drop the not-detected gate → would create `~/.vscode` anywhere | 2 |
| `editorConnectorBroken` ignores `proposedApi` | 2 |
| unparseable `argv.json` → "start fresh" (data loss) | 3 |

34 new tests. `tsc` clean, **2144 pass**, same 2 pre-existing vault-key env failures — confirmed failing identically on a clean `origin/main` worktree.

## What I have NOT proven

- **Not verified end-to-end on Windows.** Proving the extension actually appears in Megan's VS Code needs a `build:win` + daemon swap, which bounces her. The gate logic and the argv.json merge are both proven against her real data; the install itself is not.
- `code.cmd` resolves fine on her box (both candidate paths exist, `--list-extensions` answers), so the resolver isn't the blocker — but that was checked before this change, not after.
- `VSCODE_EDITOR.reconcile()` still calls the real `code` CLI and the real `~/.vscode` when invoked with default deps. No test does, and `isPhantombotBinary()` is false under `bun test`, so it stays inert — but the hazard predates this PR and I've only added a `reconcileVscode(repair, hooks)` seam rather than removing it.

### DIFF
--- a/src/connectors/acp/vscodeArgv.ts
+++ b/src/connectors/acp/vscodeArgv.ts
@@ -1,0 +1,262 @@
+/**
+ * VS Code `argv.json` writer — opts our side-loaded extension into the chat
+ * proposed APIs it declares.
+ *
+ * WHY THIS IS NEEDED. `editors/vscode/package.json` declares
+ * `enabledApiProposals: [chatSessionsProvider, chatParticipantPrivate,
+ * chatReferenceBinaryData]`. Stable VS Code refuses proposed APIs to any
+ * extension that isn't explicitly allow-listed, and for a SIDE-LOADED extension
+ * (ours is installed from a bundled .vsix, never from the marketplace) the only
+ * supported allow-list on stable is the `enable-proposed-api` array in the
+ * per-user `~/.vscode/argv.json`.
+ *
+ * Without it the extension still loads, but `chatSessionsProvider` is
+ * unavailable and it silently degrades to the `@phantombot` chat-participant
+ * fallback (see sessionProvider.ts). That fallback is deliberate, which is
+ * exactly what makes the missing entry so hard to spot: the panel looks
+ * half-working rather than broken. Nothing in the repo ever wrote this file, so
+ * every install has been landing in the degraded mode unless the user added the
+ * entry by hand.
+ *
+ * DATA-LOSS-PROOF PROCEDURE — identical to installZed.ts, and non-negotiable.
+ * `argv.json` ships from VS Code full of explanatory comments and carries the
+ * user's `crash-reporter-id`, so it is JSONC and it is precious:
+ *   1. Read the existing file.
+ *   2. Parse with jsonc-parser (tolerant of comments + trailing commas).
+ *   3. If the parse FAILS ⇒ ABORT. Write nothing, report `error`. NEVER "start
+ *      fresh", NEVER overwrite.
+ *   4. Otherwise merge our extension id into `enable-proposed-api` via
+ *      modify/applyEdits, so every other key, every comment, and the file's
+ *      formatting survive untouched.
+ *   5. Back up the original, then write atomically (temp file + rename).
+ *
+ * Idempotent: when the id is already in the array we return `current` and touch
+ * nothing, so this can run on every daemon startup without churning backups.
+ *
+ * Never throws. Every entry point returns a status.
+ */
+
+import {
+  mkdirSync,
+  readFileSync,
+  renameSync,
+  writeFileSync,
+} from "node:fs";
+import { homedir } from "node:os";
+import { dirname, join } from "node:path";
+import { applyEdits, modify, parse, type ParseError } from "jsonc-parser";
+
+import { VSCODE_EXTENSION_ID } from "../../lib/vscodeExtensionAsset.generated.ts";
+
+/** The `enable-proposed-api` key VS Code reads out of argv.json. */
+export const ENABLE_PROPOSED_API_KEY = "enable-proposed-api";
+
+export type ProposedApiStatus =
+  /** Extension id already allow-listed — nothing written. */
+  | "current"
+  /** We added the id (repair mode). VS Code must restart to pick it up. */
+  | "enabled"
+  /** Missing, and repair was off — reported, not written. */
+  | "stale"
+  /** Unparseable argv.json, or a filesystem failure. Nothing written. */
+  | "error";
+
+export interface ProposedApiResult {
+  status: ProposedApiStatus;
+  argvPath: string;
+  backupPath?: string;
+  /** Populated when `status === "error"`. */
+  error?: string;
+}
+
+/**
+ * Impure seams, injected so the whole merge runs under `bun test` against an
+ * in-memory file with no real VS Code and no real `$HOME`.
+ */
+export interface ArgvDeps {
+  argvPath: string;
+  /** File contents, or undefined when the file does not exist. */
+  read(path: string): string | undefined;
+  /** Write `contents` to `path` atomically, creating parent dirs. */
+  write(path: string, contents: string): void;
+}
+
+/**
+ * Resolve the per-user VS Code `argv.json`.
+ *
+ * It lives at `<home>/.vscode/argv.json` on EVERY platform — VS Code derives it
+ * from `os.homedir()` + its `dataFolderName`, so unlike settings.json there is
+ * no `%APPDATA%` branch on Windows and no XDG lookup on Linux. (VS Code
+ * Insiders uses `.vscode-insiders`; we target stable only.)
+ */
+export function defaultVscodeArgvPath(): string {
+  return join(homedir(), ".vscode", "argv.json");
+}
+
+/** What a reconcile WOULD do to `existing`, computed purely. */
+export type ArgvPlan =
+  | { kind: "current" }
+  | { kind: "unparseable" }
+  | { kind: "write"; updated: string };
+
+/**
+ * Decide how to bring `existing` (undefined = file absent) in line with
+ * `extensionId` being allow-listed. Pure — no I/O, so every branch is testable.
+ *
+ * A pre-existing `enable-proposed-api` that is NOT an array (a user typo, or
+ * VS Code's own commented-out scalar example uncommented wrong) is replaced
+ * wholesale with a fresh array rather than appended to, because appending to a
+ * non-array would produce a file VS Code rejects. We're only ever adding our own
+ * id, so any other extension ids already present are preserved.
+ */
+export function planProposedApi(
+  existing: string | undefined,
+  extensionId: string,
+): ArgvPlan {
+  // Tabs, because that's what VS Code ships argv.json with. jsonc-parser
+  // reformats the region around its edit, so matching the file's native style
+  // keeps the diff to exactly the key we added instead of gratuitously
+  // reindenting the user's neighbouring lines.
+  const formatting = {
+    formattingOptions: { tabSize: 4, insertSpaces: false, eol: "\n" },
+  };
+
+  if (existing === undefined) {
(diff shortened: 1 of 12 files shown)


src/connectors/acp/vscodeArgv.ts
```diff
@@ -0,0 +1,262 @@
...
+/**
+ * Resolve the per-user VS Code `argv.json`.
+ *
+ * It lives at `<home>/.vscode/argv.json` on EVERY platform — VS Code derives it
+ * from `os.homedir()` + its `dataFolderName`, so unlike settings.json there is
+ * no `%APPDATA%` branch on Windows and no XDG lookup on Linux. (VS Code
+ * Insiders uses `.vscode-insiders`; we target stable only.)
+ */
+export function defaultVscodeArgvPath(): string {
+  return join(homedir(), ".vscode", "argv.json");
```
kaieriksen: `~/.vscode/argv.json` is not the runtime-arguments file on Windows (and this also cannot target the installed VS Code distribution/profile). Repair will write a dead file and VS Code will remain degraded. Resolve the platform/distribution-specific path from the detected `code` installation or provide an explicit injected path, with Windows coverage.

kaieriksen (changes requested):
Changes requested: the VS Code argv path is not portable; see the inline comment.

andrewagrahamhodges:
Thanks Kai — I dug into this on the actual Windows box before changing anything. Splitting your comment into the part I have to push back on and the part that was right.

## `~/.vscode/argv.json` is correct on Windows

Verified on the Windows VM (`C:\Users\aghodgespwa`):

- `C:\Users\aghodgespwa\.vscode\argv.json` **exists**, and it is VS Code's **own generated file** — verbatim header (`This configuration file allows you to pass permanent command line arguments to VS Code`) plus a `crash-reporter-id`. Only VS Code writes those.
- A recursive sweep of `%USERPROFILE%` finds **exactly one** `argv.json`. There is no other candidate.

VS Code derives argv.json as `os.homedir()` + `dataFolderName` + `argv.json` on every platform — unlike `settings.json`, there is no `%APPDATA%` branch. And argv.json holds **global runtime arguments**, so it is not per-profile either; the distribution/profile half of the comment doesn't apply.

So the path isn't the blocker. No change there.

## But there was a real bug underneath it

The distribution point is right, just not where you aimed it. `defaultVscodeArgvPath()` hardcoded `.vscode` while `resolveCodeCli()` independently decides which CLI to install into. Today those agree only **by coincidence** — the resolver happens to look for stable `code`/`code.cmd` only. Add Insiders to the resolver and we install the extension into Insiders and allow-list it in **stable**, which degrades to the `@phantombot` participant fallback with no error anywhere. Exactly the silent-half-working failure this PR exists to kill.

Fixed in `85b1352`: the CLI `resolveCodeCli()` actually returned is threaded through `reconcileVscode` into the argv step, and the data folder is derived from it (`code` → `.vscode`, `code-insiders` → `.vscode-insiders`, `codium` → `.vscode-oss`, unknown → stable rather than inventing a folder). It's a single funnel — `reconcileVscode` is the only caller of `ensureProposedApi`/`checkProposedApi`, so there's no bypass.

## Windows coverage added — and it bites

Your ask for Windows tests was fair regardless of the premise, and writing them found a second trap:

> `node:path.basename()` does **not** split `\` under POSIX.

So `basename("C:\\...\\bin\\code-insiders.cmd")` returns the *whole string* on Linux. A Windows-path test written the obvious way passes on Windows and silently misbehaves on Linux CI — the **same asymmetry that let the `phantombot.exe` gate ship untested in the first place**. The helper now splits on both separators, so the Windows cases are meaningful under our Linux CI.

Mutation-tested rather than trusting green:

| mutation | new tests that fail |
|---|---|
| data folder always `.vscode` | 3 |
| `cliBasename` → `node:path.basename` | 1 (the backslash case) |
| `reconcileVscode` drops `codeCommand` | 2 |

## Mac/Linux safety

Ran the real read-only path against a Linux box: resolves `/snap/bin/code` → `.vscode` → `proposedApi: current`, `argv.json` byte-identical, no backup written. `tsc` clean, 2152 pass (same 2 pre-existing vault-key env fails as `origin/main`).

## Still not proven

The extension actually appearing in VS Code on the Windows box needs a `build:win` + daemon swap. Also worth flagging: that box's `argv.json` **already** contains our extension id, so the argv half will report `current` there and write nothing — a live deploy will exercise the gate fix but *not* the argv write path. That needs a scratch user or a fixture, and I'd rather say so than claim coverage I don't have.

src/connectors/acp/vscodeArgv.ts
andrewagrahamhodges: Verified live on the Windows VM (Alpha) — this is resolved.

**On the blocker itself:** `~/.vscode/argv.json` *is* the runtime-arguments file on Windows. VS Code derives it as `os.homedir() + dataFolderName + "argv.json"` on every platform — there is no `%APPDATA%` branch (unlike `settings.json`), and argv.json is global runtime args, not per-profile. On the box, `C:\Users\aghodgespwa\.vscode\argv.json` is VS Code's own generated file (its header comments plus a `crash-reporter-id`), and a recursive sweep of `%USERPROFILE%` finds exactly one `argv.json`.

**Your instinct pointed at a real bug, though.** `defaultVscodeArgvPath()` hardcoded `.vscode` while `resolveCodeCli()` independently picked the CLI — they agreed only by coincidence. Add Insiders/VSCodium to the resolver and we'd install the extension into one editor and allow-list it in another, silently degrading to the `@phantombot` participant fallback with no error. Fixed in 85b1352: the resolved CLI is now threaded through `reconcileVscode` into the argv step, and the data folder is derived from it (`code-insiders` → `.vscode-insiders`, `codium` → `.vscode-oss`, default `.vscode`). `reconcileVscode` is the only caller, so there's no bypass.

Writing the Windows path coverage you asked for surfaced a second trap: `node:path.basename()` doesn't split `\` under POSIX, so `basename("C:\\...\\code-insiders.cmd")` returns the whole string on Linux CI. The obvious test would pass on Windows and misbehave on Linux — the same asymmetry that let the original `phantombot.exe` gate ship untested. The helper now splits on both separators. All three fixes were mutation-tested (reverted each, confirmed 1–3 tests fail).

**Live verification on Alpha** (exe `1.1.194-pr286`): `phantomyard.phantombot-vscode` now appears in `code --list-extensions`, and nothing but `reconcileEditorConnectors` installs it — that's the gate fix firing on Windows for the first time. Andrew confirms the extension is working in VS Code.

**One caveat I want on the record:** this proves the *gate* fix, not the *argv write* path. That box's `argv.json` already contained our extension id before the deploy, so the write path reported `current` and wrote nothing. It's covered by unit tests plus a local dry-run against her real file (one-comma, one-key diff; comments, tabs and `crash-reporter-id` intact) — but not yet exercised live. That needs a scratch user or a fixture.

`tsc` clean, 2152 pass, same 2 pre-existing vault-key env fails as `origin/main`. Ready for another look.

kaieriksen (approved):
Re-reviewed latest head. The argv path now follows the resolved VS Code distribution, including Windows separators and Insiders/VSCodium data folders; the prior blocker is fixed. Focused tests: 76 passed; typecheck passed.

PR #287, opened 2026-07-11
feat(cli): built-in shell tab-completion, set up on install/update

salvatoretrimarchi:
## Built-in shell tab-completion, set up on install/update

  ### Summary
  Adds `<TAB>` completion for the phantombot CLI that works out of the box. `phantombot install` sets it up for bash, zsh, and fish; `phantombot update` refreshes it; `phantombot uninstall` removes it. There is no separate command to run and no
  manual step — after installing, open a new shell and `<TAB>` completes subcommands and flags.

  ### How it works
  Completion is dynamic. A small per-shell stub calls a hidden `phantombot _complete` backend on every `<TAB>`. That backend walks the live command tree and prints the candidates, so completion always matches the actual subcommands and flags —
  nothing to regenerate when a command is added. The backend is side-effect-free (it never provisions state, like `--help`) and stays out of `--help` output.

  Coverage: top-level subcommands, nested subcommands, and per-command flags (including the `--no-` form for booleans and `--help`/`--version`). POSIX only; on Windows it's a no-op.

  ### Example
  ```console
  $ phantombot <TAB>
  ask       doctor    harness   install   memory    p2p       run       telegram
  backfill  embedding heartbeat logs      nightly   persona   start     tick
  ...

  $ phantombot p<TAB>
  p2p          persona      phantomchat

  $ phantombot p2p <TAB>
  status

  $ phantombot logs --<TAB>
  --follow     --no-follow  --lines      --help
  ```

  ### Testing
  - Full test suite: 2183 pass, 0 fail (new coverage for the completion engine and the installer).
  - End-to-end: after install, a fresh shell completes subcommands, nested subcommands, and flags exactly as shown above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,18 @@ Maintenance:
 | `phantombot nightly [--resume]` | Run or resume memory distillation |
 | `phantombot doctor [--no-repair]` | Check memory health and optionally repair |
 
+## Shell completion
+
+`phantombot install` sets up `<TAB>` completion for `bash`, `zsh`, and `fish`
+automatically — no extra command, no separate opt-in. `phantombot update`
+refreshes it, and `phantombot uninstall` removes it. Open a new shell after
+installing to pick it up.
+
+Completion is dynamic: a small stub in your shell calls back into the binary on
+every `<TAB>`, so it always matches the available subcommands and flags —
+`phantombot p<TAB>` → `persona`, `phantomchat`, `p2p`; `phantombot p2p <TAB>` →
+`status`; `phantombot logs --<TAB>` → `--follow`, `--no-follow`, `--lines`.
+
 ## Telegram
 
 Phantombot runs one or more Telegram long-poll listeners. Each listener needs a
--- a/src/cli/install.ts
+++ b/src/cli/install.ts
@@ -15,6 +15,8 @@
 import { defineCommand } from "citty";
 import { basename } from "node:path";
 
+import { installCompletions } from "../lib/completionInstall.ts";
+
 import {
   BunLaunchctlRunner,
   defaultPlistPath,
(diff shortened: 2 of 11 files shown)


lenaparkhodges (approved):
PR #287 — solid feature, clean implementation. A few notes:

**Design** — the `_complete` hidden-command interception in `index.ts`, ahead of credential bootstrap, is the right call: a tab press never touches the vault. Descending the live Citty tree means completion can't drift from real subcommands, and the flag/positional handling (including `--no-` for booleans and single-char alias detection) is thorough.

**Non-blocking notes:**
1. `zshSourceBody` runs `compinit -u 2>/dev/null` on every shell start if the user's `.zshrc` sources it — `-u` skips the insecure-directories check, which is a reasonable default for a generated stub but worth a one-line comment explaining why `-u` specifically (silently bypassing a security prompt is the kind of thing a future maintainer will want context on).
2. `installCompletions`/`uninstallCompletions` are best-effort per-shell and swallow errors into a `warning:` write — good, but on `install` a completion failure is silently non-fatal even if *all* shells fail (empty `installed` list, no user-visible signal beyond the per-shell warnings already written). Fine as-is since it's advisory, just flagging.
3. Fish path assumes `~/.config/fish` existing is a reliable signal of "user runs fish" — reasonable heuristic, matches the bash/zsh rc-exists pattern.

**Testing** — 2183 tests pass, thorough coverage of the completion engine, script emitters, and the installer (idempotency, replace-in-place, windows no-op, rc-file-exists-triggers-install). CI green (test, test-windows, typecheck).

No blockers. Nice first PR, Salvatore — welcome to the project! 🎉

kaieriksen (approved):
Approved.

Reviewed the dynamic completion engine, bash/zsh/fish installers, install/update/uninstall wiring, and read-only `_complete` entrypoint. Focused suites pass (48 tests), typecheck passes, runtime completion smoke passes, and the generated bash script parses cleanly. CI is green.

PR #288, opened 2026-07-12
fix(crypto): mirror the ms sub-second ordering tag on kind-14 rumors

andrewagrahamhodges:
## Why

PhantomChat orders messages by `mid = created_at * 1e6 + slot`. `created_at` is **whole-second only**, so two messages sent in the same second tie — and the legacy tiebreak was a **hash of the event id**: stable across devices, but not chronological. Same-second pairs therefore rendered in **coin-flip order** (Andrew hit this: a reply landed above the message it was replying to).

phantomchat now stamps the real millisecond into an `['ms', '0'-'999']` tag on the kind-14 rumor and uses it as the sub-second slot — deterministic *and* chronological. This PR mirrors that on the bot side.

**Why the bot needs it too:** without the tag, phantombot's messages have no ms, fall back to the hash slot, and a bot reply landing in the same second as the user's own message can still render out of order. A half-applied ordering rule is worse than none.

## What changed

- `MS_TAG` / `buildMsTag` / `getMsSlotFromTags` — byte-identical to the phantomchat side.
- Minted in **both** rumor builders (`wrapV2` + `createRumor`) from the **same `Date.now()`** that sets `created_at`. Invariant: a second `Date.now()` call can straddle a second boundary and put the message a full second out of order.
- Cross-repo shared vector **re-pinned** (`1012a225… → 3ded8957…`) plus new coverage for the tag mint and the legacy/malformed fallback.

## Wire-format note

The tag is **additive and ignorable** — stock Nostr clients (Damus, Amethyst) drop unknown tags, so decryption and interop are unaffected, and legacy messages without the tag keep the hash slot.

But it **is** part of the hashed tag set, so it changes the rumor id. That's exactly what the cross-repo vector exists to catch — **both repos must land together**:

- phantomchat: https://github.com/phantomyard/phantomchat/pull/74 (branch `fix/reliability-grind`, commit `96eb071`)

## Testing

- `tests/lib-nostrCrypto.test.ts`: **25/25 pass**, including the re-pinned vector.
- Full suite: 2109 pass. The 2 `PiHarness routing (subprocess)` failures are **pre-existing on clean `main`** (verified by stashing this change) — unrelated to this PR.
- **Not verified headless:** the actual same-second render order on two real devices. That's a live two-device check once both sides deploy.

### DIFF
--- a/src/lib/nostrCrypto.ts
+++ b/src/lib/nostrCrypto.ts
@@ -468,7 +468,15 @@ export async function wrapV2(
   replyTo?: { eventId: string; relayUrl?: string },
 ): Promise<{ event: NTNostrEvent; rumorId: string }> {
   const senderPubHex = getPublicKey(senderSk);
-  const tags: string[][] = [["p", recipientPubHex], ["v", "pc-v2"]];
+  // ONE Date.now() for BOTH created_at and the ms tag — see MS_TAG. Two calls
+  // can straddle a second boundary and desync the sub-second slot from the
+  // second it belongs to.
+  const nowMs = Date.now();
+  const tags: string[][] = [
+    ["p", recipientPubHex],
+    ["v", "pc-v2"],
+    buildMsTag(nowMs),
+  ];
   if (replyTo) {
     tags.push(["e", replyTo.eventId, replyTo.relayUrl || "", "reply"]);
   }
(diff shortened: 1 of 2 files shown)


kaieriksen (approved):
Reviewed the exact head. The shared millisecond ordering tag is stamped from the same Date.now() used for created_at, legacy/malformed tags retain the hash fallback, and the cross-repo vector plus coverage are correct. Focused tests: 77 passed; typecheck passed.

PR #134, opened 2026-05-28
feat(telegram): pace progress and split reply bubbles

robertclawson:
## Summary
- coalesce progress narration on a timer instead of flushing once per tool call
- stream final Telegram text through a markdown-aware segmenter that splits smaller bubbles by sentence/char boundaries
- split voice replies into shorter clips and expose tuning knobs under [channels.telegram.streaming]
- document Telegram reply pacing in README

## Defaults
- narration_flush_ms = 4500
- bubble_max_sentences = 4
- bubble_max_chars = 700
- voice_max_sentences = 3

## Tests
- PATH=/home/robbie/.bun/bin:$PATH bun run typecheck
- PATH=/home/robbie/.bun/bin:$PATH bun test tests/config.test.ts tests/channels-streamSegmenter.test.ts tests/channels-telegram.test.ts
- PATH=/home/robbie/.bun/bin:$PATH bun test
- PATH=/home/robbie/.bun/bin:$PATH bun run build

PR #290, opened 2026-07-12
fix(channels): stop shipping a trailing emoji as its own bubble

andrewagrahamhodges:
## What Andrew saw

A PhantomChat push notification reading **"max: ⚡"** — a lightning bolt and nothing else — for a message I had, in fact, fully answered. He replied "An answer please." He was right to.

## Not lost messages

First thing I checked, since that was the report. **512 PhantomChat turns in the store, 0 user messages without a reply. 0 turn failures and 0 relay/publish errors in 7 days.** Nothing is being dropped. This is a *rendering* bug.

## Root cause

The sentence segmenter treats a trailing emoji as a **sentence of its own**, because it follows terminal punctuation. When the sentence before it lands exactly on the bubble boundary (`bubbleMaxSentences`, 4 in prod), the splitter flushes the real answer and the emoji is left over alone:

```
[0] len=188  "The gate was wrong on Windows and the fix has landed. …#76 is merged. "
[1] len=1    "⚡"
```

On **Telegram** that reads as a harmless sign-off bubble under the answer — which is why it went unnoticed for weeks. On **PhantomChat** the newest event drives the chat-list preview *and* the push notification, so the sign-off **becomes the entire notification** and the answer is buried above it. Same bug, different blast radius.

## Not new, and not just ⚡

Latent since the segmenter shipped (`b078dac`, #134, 2026-05-28). Firing since **2026-06-26** — daily, across **73 replies**. It is any trailing emoji, not just my sign-off: 👍 🔧 🦴 👂 🌙 🎉. Four `Stopped. 👍` replies on 2026-06-30 mean four bubbles that said nothing but 👍.

## The fix

Two paths, because the reply is **not always segmented in one pass** — this is the part I got wrong on the first pass and had to correct:

| path | what happens | fix |
|---|---|---|
| whole-text (`splitIntoSegments`) | all bubbles known before any send | **coalesce** the orphan into the previous bubble — emoji preserved, just attached |
| streaming (`StreamSegmenter.push`) | previous bubble is **already published** when the emoji arrives as the leftover suffix | nothing to merge into → **suppress** at the send boundary |

Suppression is gated on having already sent a bubble this turn, so a reply that is genuinely just `👍` still sends. One shared predicate (`hasTextSubstance`) so both strategies agree on what "not an answer" means. `\p{L}\p{N}`, not `[a-z]` — "готово" and "完了" are substance.

## Testing

- **Mutation-tested all five** rather than trusting green: drop the coalesce → 1 fail; drop either send guard → 1 fail each; suppress unconditionally (ignoring the "already sent" gate) → 1 fail each, both channels.
- The mutation run **caught a bad test of my own**: my first regression fixture was a 9-sentence reply, which passes *with the fix reverted* — the emoji rides out on the short final bubble and never orphans. The bug needs a sentence count that is a **multiple of** `bubbleMaxSentences`. Fixture rewritten, and the reason is written down in the test so nobody "simplifies" it back.
- Coverage on **both** channels (`channels-telegram.test.ts`, `channels-phantomchat-server.test.ts`), driving the real streaming loops, plus unit coverage on the segmenter.

`tsc` clean. **2195 pass.** The 2 `PiHarness` failures are pre-existing and environmental (`PHANTOMBOT_PI_API_KEY` present in my env from the vault) — confirmed identical on clean `main`, not introduced here.

## Not verified

No daemon swap yet, so this is not yet confirmed against a live relay — the evidence is the reproduction plus the two channel-level tests. Happy to deploy and watch a real sign-off land as one bubble if you want it proven end to end.

### DIFF
--- a/src/channels/core/engine.ts
+++ b/src/channels/core/engine.ts
@@ -59,6 +59,7 @@ import {
   TELEGRAM_BOT_COMMANDS,
 } from "../commands.ts";
 import {
+  hasTextSubstance,
   splitIntoSegments,
   StreamSegmenter,
 } from "../streamSegmenter.ts";
@@ -1029,6 +1030,17 @@ async function processChatMessage(
     kind: "narration" | "final" | "error",
   ) => {
     if (text.trim().length === 0) return;
+    // A bubble with no letters or digits ("⚡", "👍") is a sign-off, not an
+    // answer. It reaches here as the streamed leftover: the splitter cut the
+    // real bubble at a sentence boundary and PUBLISHED it, so there is nothing
+    // left to merge into — suppress it instead. Only once something real has
+    // already gone out; a reply that is genuinely just "👍" still sends.
+    if (
+      !hasTextSubstance(text) &&
+      narrationBubblesSent + finalBubblesSent > 0
+    ) {
+      return;
+    }
     try {
       await input.transport.sendMessage(msg.conversationId, text);
       if (kind === "narration") narrationBubblesSent++;
--- a/src/channels/core/engine.ts
+++ b/src/channels/core/engine.ts
@@ -1033,12 +1033,14 @@ async function processChatMessage(
     // A bubble with no letters or digits ("⚡", "👍") is a sign-off, not an
     // answer. It reaches here as the streamed leftover: the splitter cut the
     // real bubble at a sentence boundary and PUBLISHED it, so there is nothing
-    // left to merge into — suppress it instead. Only once something real has
-    // already gone out; a reply that is genuinely just "👍" still sends.
-    if (
-      !hasTextSubstance(text) &&
-      narrationBubblesSent + finalBubblesSent > 0
-    ) {
+    // left to merge into — suppress it instead.
+    //
+    // Gate on FINAL bubbles only. Narration ("Restarting the service now…")
+    // is not an answer, so it must never license suppressing one: if the
+    // model's entire reply is "👍", that emoji IS the answer and has to ship
+    // even though narration already went out. An orphaned sign-off always
+    // trails a published *final* bubble, so this still catches every one.
+    if (!hasTextSubstance(text) && finalBubblesSent > 0) {
       return;
     }
     try {
--- a/src/channels/phantomchat/server.ts
+++ b/src/channels/phantomchat/server.ts
@@ -681,10 +681,11 @@ export async function runPhantomchatServer(
     let finalCandidateSentChars = 0;
     let finalReply: string | undefined;
     let requestedReplyMode: ReplyModeRequest | undefined;
-    // Bubbles actually published this turn (narration or final). Gates the
-    // orphaned-sign-off suppression in sendBubble: nothing out yet means the
-    // fragment IS the reply, so it must still send.
-    let bubblesSent = 0;
+    // FINAL bubbles actually published this turn. Gates the orphaned-sign-off
+    // suppression in sendBubble: no final bubble out yet means the fragment IS
+    // the reply, so it must still send. Narration is deliberately NOT counted
+    // here — see the comment in sendBubble.
+    let finalBubblesSent = 0;
 
     // Reply-modality routing (mirrors core/engine.ts). Default: mirror the
     // input — a voice note in → a voice note back, text in → text out. An
@@ -747,7 +748,10 @@ export async function runPhantomchatServer(
     // the final reply path. groupTypingMembers is the same set the reply path
     // broadcasts to (inbound p-tags ∪ { sender }). Best-effort: a failed bubble
     // is logged, not thrown, so one dropped progress line never aborts the turn.
-    const sendBubble = async (text: string): Promise<void> => {
+    const sendBubble = async (
+      text: string,
+      kind: "narration" | "final" = "final",
+    ): Promise<void> => {
       if (text.trim().length === 0) return;
       // A bubble with no letters or digits ("⚡", "👍") is a sign-off, not an
       // answer. It reaches here as the streamed leftover: the splitter cut the
@@ -755,9 +759,14 @@ export async function runPhantomchatServer(
       // nothing left to merge into — suppress it instead. This matters most
       // here: the newest event drives the push notification, so an orphaned
       // emoji becomes a notification reading "max: ⚡" with the real answer
-      // buried above it. Only once something real has already gone out; a
-      // reply that is genuinely just "👍" still sends.
-      if (!hasTextSubstance(text) && bubblesSent > 0) return;
+      // buried above it.
+      //
+      // Gate on FINAL bubbles only. Narration is not an answer, so it must
+      // never license suppressing one: if the model's whole reply is "👍",
+      // that emoji IS the answer and has to ship even though narration
+      // already went out — otherwise the push notification becomes the
+      // narration and the confirmation never arrives.
+      if (!hasTextSubstance(text) && finalBubblesSent > 0) return;
       try {
         if (msg.groupId) {
           await transport.sendGroupMessage(
@@ -770,7 +779,7 @@ export async function runPhantomchatServer(
           // and publishes both wraps. conversationId === recipient hex pubkey.
           await transport.sendMessage(senderHex, text);
         }
-        bubblesSent++;
+        if (kind === "final") finalBubblesSent++;
       } catch (e) {
         log.warn("phantomchat: bubble send failed", {
           error: (e as Error).message,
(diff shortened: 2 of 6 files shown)


src/channels/core/engine.ts
```diff
@@ -1029,6 +1030,17 @@ async function processChatMessage(
...
   ) => {
     if (text.trim().length === 0) return;
+    // A bubble with no letters or digits ("⚡", "👍") is a sign-off, not an
+    // answer. It reaches here as the streamed leftover: the splitter cut the
+    // real bubble at a sentence boundary and PUBLISHED it, so there is nothing
+    // left to merge into — suppress it instead. Only once something real has
+    // already gone out; a reply that is genuinely just "👍" still sends.
+    if (
+      !hasTextSubstance(text) &&
+      narrationBubblesSent + finalBubblesSent > 0
```
robertclawson: This is the hole. `narrationBubblesSent` should not be in this condition.

Narration goes through this same `sendTextSegment` (`send: (text) => sendTextSegment(text, "narration")` further down), so a single narration bubble flips this guard to `true` — and then a genuine emoji-only final answer gets suppressed, contradicting the comment directly above it ("a reply that is genuinely just 👍 still sends").

Reproduced: narration flushed on the timer, final reply `👍` → `narrationBubbles: 1, finalBubbles: 0`, and the `👍` never sends. User asks "restart it", gets "Restarting the service now…", and never gets the confirmation.

```suggestion
    if (!hasTextSubstance(text) && finalBubblesSent > 0) {
      return;
    }
```

The orphan case you're fixing still works — an orphaned sign-off always trails a published *final* bubble. `server.ts:806` needs the same treatment: split `bubblesSent` into a final-only counter.

robertclawson (changes requested):
Reviewed on a checkout of the branch: `tsc --noEmit` clean, suite green, and I wrote a probe test for the case I was suspicious of. **The root cause and the two-path design are right — but the suppression guard has a hole that I reproduced.**

## 🔴 The guard counts narration bubbles, so it can eat a genuine emoji-only answer

The gate in `engine.ts` is `narrationBubblesSent + finalBubblesSent > 0`. But narration flows through the *same* `sendTextSegment` (`send: (text) => sendTextSegment(text, "narration")`), and PhantomChat is the same shape — `send: sendBubble` at `server.ts:806` bumps the same `bubblesSent`.

So the PR's own stated invariant — *"a reply that is genuinely just 👍 still sends"* — **fails as soon as a narration bubble precedes it.** Probe: narration flushed on the timer, model's final answer is `👍`:

```
narrationBubbles: 1, finalBubbles: 0
SENT BUBBLES: ["Restarting the service now…"]
```

The `👍` is silently dropped. Real-world shape: user says "restart it" → "Restarting the service now…" → **the confirmation never arrives.** On PhantomChat the push notification becomes the *narration* rather than the answer — which is the same class of bug this PR exists to kill, just moved one square left. It trades a bubble that says too little for one that says nothing at all.

**Fix is effectively one word:** gate on `finalBubblesSent > 0`. Narration is not an answer, so it should not license suppressing one. The orphan case this PR targets still works, because an orphaned sign-off *always* trails an already-published final bubble. PhantomChat needs `bubblesSent` split into a final-only counter to match.

**Why the tests miss it:** both "only a sign-off still sends" cases use harnesses that emit zero narration (`{type:"done"}` straight off), so the regression net has a narration-shaped hole. Please add a third case: narration → emoji-only final.

## ✅ What's genuinely good

- Root cause is correct, and the two-path split (coalesce when unsegmented, suppress when already published) is the right call — you can't merge into a bubble that's already on the wire.
- `\p{L}\p{N}` over `[a-z]` — correct, and nice to see it tested with `готово` / `完了`.
- The mutation-testing note catching your *own* bad 9-sentence fixture is the most credible thing in this PR. The "must be a multiple of `bubbleMaxSentences`" reasoning is right, and writing it into the test so nobody simplifies it back is exactly right.
- `coalesceTrailingFragment` doesn't mutate its input, handles `[]` and single-element, and `segments.join("") === reply` proves nothing is dropped.

**Verdict: request changes.** One-line gate fix + one test. Everything else ships.

andrewagrahamhodges:
Confirmed and fixed in `59f3c4b` — thanks, this was a real hole and your probe named it exactly.

**You were right.** Narration flows through the same send boundary on both channels (`send: (text) => sendTextSegment(text, "narration")` in the engine, `send: sendBubble` at `server.ts:806`), so `narrationBubblesSent + finalBubblesSent > 0` meant a single narration bubble was enough to license suppressing an emoji-only answer. The PRs own stated invariant failed the moment narration preceded it. I traded a bubble that said too little for one that said nothing — worse on PhantomChat, where the push notification then becomes the narration rather than the answer.

**The fix is the one you called:** gate on `finalBubblesSent > 0`. Narration is not an answer. The orphaned sign-off this PR targets *always* trails an already-published final bubble, so nothing is lost from the original case. PhantomChat needed slightly more than one word — `sendBubble` now takes a `kind` and only counts `"final"`, matching the engine.

**Tests:** added `an emoji-only answer still sends after narration has gone out` to both channel suites — `text` chunk (no terminal punctuation, so it stays below the final-answer splitter) → `progress` → `done: "👍"`. The engine log confirms the shape is the one you probed, now correct:

```
narrationBubbles: 1, finalBubbles: 1
SENT: ["Restarting the service now", "👍"]
```

Mutation-tested both: reverting the engine guard to `narration + final > 0` fails the Telegram case, and dropping the `kind === "final"` condition on PhantomChats counter fails the chat case. The old narration-free tests were passing against the broken guard, which is precisely why they missed it — I wrote your reasoning into the new tests so nobody removes the `progress` chunk as noise.

`tsc` clean, 2197 pass. The 2 `PiHarness` fails are pre-existing/environmental (vault env leakage) — identical on clean `main`.

Still not proven: no daemon swap, so this is tests + reproduction, not a live relay.

kaieriksen (approved):
Reviewed the current head carefully. The final-only suppression gate is correct: narration cannot suppress a genuine emoji-only answer, while an orphaned sign-off after a published final bubble is still suppressed.

Focused channel suites: 189 tests pass. Typecheck: clean.

lenaparkhodges (approved):
Kai's blocker is fixed cleanly in `59f3c4b`: the suppression gate now checks `finalBubblesSent` instead of the combined narration+final counter, so an emoji-only final answer still ships after narration goes out. `sendBubble` in phantomchat/server.ts now takes a `kind` param and only bumps the final counter, matching engine.ts.

The new regression test (narration → emoji-only final) closes exactly the hole Kai found — both channels covered, and it's mutation-tested per the PR's own methodology.

CI green across the board (typecheck, test, test-windows, CodeQL). No blockers from me.

robertclawson (approved):
Re-reviewed `59f3c4b` on a fresh checkout. **My blocker is resolved — and I verified it rather than taking the summary's word for it.** Clearing my `CHANGES_REQUESTED`.

## The fix is the right one

Both channels now gate on final bubbles only:

- `engine.ts` — `!hasTextSubstance(text) && finalBubblesSent > 0`
- `phantomchat/server.ts` — `sendBubble` takes a `kind`, increments `finalBubblesSent` only on `"final"`, and narration is wired in as `send: (text) => sendBubble(text, "narration")`

That's exactly the invariant the PR claims: *narration is not an answer, so it can't license suppressing one.* The orphaned sign-off case still works, because an orphan always trails an already-published **final** bubble.

## Verified, not assumed

`tsc --noEmit` clean. Focused suites: **189 pass, 0 fail** (matches @kaieriksen).

The reason this bug shipped past the first round is that the old tests passed against the *broken* guard — so a green suite proves nothing on its own here. I mutation-tested both halves:

| Mutation | Result |
|---|---|
| engine guard → `narrationBubblesSent + finalBubblesSent > 0` (the original bug) | ❌ `an emoji-only answer still sends after narration has gone out` **fails** (137 pass / 1 fail) |
| phantomchat → drop the `kind === "final"` condition, always increment | ❌ same-named chat case **fails** (36 pass / 1 fail) |

Each mutation is caught by precisely the test written for it, and by nothing else. The regression net is real and load-bearing — the `progress` chunk in those tests is the whole point, not noise. The comments saying so are well placed; please keep them.

## One note for whoever merges

Still unproven by any of us: **no live daemon swap.** This is tests + reproduction, not a relay against a real Telegram/PhantomChat session. Low risk given the guard is a pure function of two counters, but the PR is still a **draft** — worth one manual sanity check of the `restart it` → narration → `👍` flow before it lands, since that's the exact shape the push notification depends on.

Nice work turning this around cleanly. ✅

PR #291, opened 2026-07-14
fix(acp): a new editor thread is a new conversation, and /stop actually stops

andrewagrahamhodges:
Say "hello" in a fresh Zed / VS Code / JetBrains thread and phantombot went off and did work nobody asked for — with no way to stop it. Two independent bugs, both fixed here.

## 1. A new thread inherited the last thread's orders

ACP conversations were keyed on the **workspace cwd**, not the thread:

```
conversation = "acp:" + sha256(cwd)
```

So a brand-new editor thread was *not new* — `runTurn` loaded the last 30 turns for that cwd. And it replays them as real **USER** turns. The model opened what the editor showed as an empty thread and found a trailing queue of live-looking imperatives:

> *"open a PR to fix the sync bug"* … *"Go."* … *"reviews have landed"* … *"Yes, please."*

You type "hello", it reads the last thing you apparently asked for, and it starts working. Nothing was being injected by the extension — this was our own history bleed.

**The fix is about role, not volume.** The same bytes are dangerous in the user role and inert in the system role. So:

| | before | after |
|---|---|---|
| conversation key | `acp:<cwdhash>` — every thread shares one | `acp:<cwdhash>:<threadid>` — per thread |
| turn history in a new thread | last 30 turns of the workspace, as user/assistant turns | **empty** |
| workspace context | *(none — it **was** the history)* | a **briefing** in system context, framed as finished sessions |
| old threads | inherited, never addressable | `session/load` resumes verbatim |

The thread id is embedded in the opaque sessionId (`acp_<cwdhash>_<thread>`), so `session/load` re-derives the conversation from the token the editor persisted — no server-side map to keep. **Legacy tokens** (`acp_<24hex>`) still resolve to the old cwd-keyed conversation, so threads created against the current build reopen with their history intact.

The briefing (`briefing.ts`) quotes recent workspace activity under an explicit header: already-finished, not a work queue, stale approvals don't authorize anything now, act only on this turn's message. Bounded on three axes (14 turns / 400 chars each / 4000 total), and the current thread's own turns are excluded — they're real history, and re-quoting them would recast the user's *live* instruction as an already-handled one.

Net: a fresh thread knows exactly what shipped yesterday and what's still open — and does nothing until asked.

## 2. `/stop` and `/reset` didn't exist over ACP

`channels/commands.ts` was only ever wired into the Telegram engine. `src/connectors/` never imported it, so a typed `/stop` was delivered to the harness as **ordinary prompt text** — the model would cheerfully *talk about* stopping while the turn it was meant to kill ran on. We also never advertised `availableCommands`, so the editor's `/`-menu was empty and there was no hint they were meant to work.

`/stop`, `/reset`, `/status`, `/harness`, `/help` are now handled, advertised via `available_commands_update` on `session/new` + `session/load`, and delegate to the same dispatcher Telegram uses (no behaviour reimplemented).

**They dispatch OUT OF BAND, ahead of the serial request queue** — the load-bearing detail. `/stop` exists to kill the turn that is *currently blocking that queue*; queued behind it, it would only run once the runaway turn had already finished. Exactly the deadlock `session/cancel` is already exempted from.

`/update` and `/restart` are deliberately **not** offered here: they swap the binary and bounce a service whose lifecycle the **editor** doesn't own. They fall through to the model rather than half-executing.

Over-capture is guarded too: `/usr/bin/env …`, `/^regex$/`, `/home/dev/x.ts`, a persona's own `/remember` — all still reach the model.

## Existing escape hatch, for the record

The editor's **cancel button** (`session/cancel`) always worked and still does. It was *typed* `/stop` that was dead.

## Tests

`+3` suites, all pinned by negative checks — I deleted each fix and watched them go red:

- restore cwd-keying → the inheritance test fails (the old work order lands back in `history`)
- remove the out-of-band dispatch → all three command tests fail (the runaway turn is never cancelled)

Covers: thread-scoped keys, legacy-token resume, briefing framing + bounds + workspace isolation, `/stop` reaching past a running turn, `/reset` scoped to its own thread, and slash-prefixed prose still reaching the model.

Full suite: 2223 pass / 2 fail — both failures (`PiHarness routing`) are **pre-existing on `main`** and pass in isolation (suite-order flake). Typecheck clean.


src/connectors/acp/commands.ts
```diff
@@ -0,0 +1,102 @@
...
+
+import {
+  handleSlashCommand,
+  type SlashCommandContext,
+  type SlashCommandResult,
+} from "../../channels/commands.ts";
+import type { AcpAvailableCommand } from "./protocol.ts";
+
+/**
+ * The commands phantombot offers over ACP, in menu order.
```
kaieriksen, line 31: This introduces a user-facing ACP command menu and changes the command surface. Per the repo-level AGENTS.md contributing discipline, user-facing behavior/architecture changes must update both README.md and AGENTS.md in the same PR; neither file is changed here. Please add the documentation before approval.

kaieriksen (changes requested):
Requesting changes: the PR changes user-facing ACP behavior (new thread/session keying, workspace briefing, and advertised slash commands) and architecture, but it does not update README.md and AGENTS.md as required by the repository's contributing discipline. Please document the new ACP behavior and command surface in both files before approval.

andrewagrahamhodges:
Docs added in `a7097b0` — you're right, the contributing discipline is explicit and I skipped it.

**README** — two new subsections in the Editors chapter:
- *Threads and workspace context*: a new thread is a new conversation (empty history); workspace context arrives as a read-only briefing framed as finished sessions; `session/load` resumes a thread verbatim. Includes a note on the old cwd-keyed behaviour so the change is legible to anyone who hit the runaway.
- *Editor Commands*: the advertised `/`-menu table (`/stop`, `/reset`, `/status`, `/harness`, `/help`), the out-of-band dispatch, and why `/update` + `/restart` are deliberately excluded from this surface.

**AGENTS.md** — conventions 14 and 15, written as traps rather than description:
- 14: the two scopes (workspace vs conversation) must stay distinct — never widen the conversation key back to cwd, and don't 'simplify' the briefing back into history injection, because *role* is the entire mechanism that makes prior turns inert.
- 15: ACP slash commands dispatch out of band, ahead of the serial queue — `/stop` exists to kill the turn that is *blocking* that queue, so queueing it behind that turn is a deadlock.

One correction I made to my own draft while verifying it: commands are advertised **per session** (`available_commands_update` on `session/new` and `session/load`), *not* in the `initialize` result. I'd written the latter from memory; the code says otherwise, and AGENTS.md now says what the code says, with a note that any new session-creating path must send the notification too.

kaieriksen (approved):
Re-reviewed at a7097b0. The required README.md and AGENTS.md updates are now present. Verified 228 focused tests, typecheck, and all GitHub CI checks green.

PR #292, opened 2026-07-14
fix(vscode): surface phantombot's slash commands in the /-menu

andrewagrahamhodges:
## Problem

In VS Code the `/`-menu showed only VS Code's own `/debug` and `/model` — none of phantombot's commands. Zed (and JetBrains, which speaks ACP natively) render all five.

Not an IDE collision, and the server is fine: it advertises `/stop`, `/reset`, `/status`, `/harness`, `/help` via `available_commands_update` on both `session/new` and `session/load` (`server.ts:195,242`). Both bugs are in our VS Code extension.

## Root causes

**1. The notification was dropped before anything could read it.**
`AcpClient.handleSessionUpdate` looked the session up in `promptStreams` and returned early when absent. That map only exists *for the duration of a prompt* — and the commands update arrives at `session/new`, when no prompt is in flight. Every advertisement hit the floor. (The `switch` didn't handle `available_commands_update` either, but it never got that far.)

**2. VS Code has no dynamic slash-command API.**
Even a perfectly-handled notification could not have rendered the menu. The delegate agent's commands come from the static manifest contribution — `ChatSessionsService._registerAgent` builds it as `slashCommands: contribution.commands ?? []`. The `chatSessions` contribution point accepts a `commands` array (`{name, description, when}`); a declared `chatParticipants` entry would have double-registered the `canDelegate` agent id and thrown, so this is the right — and only — surface.

## Fix

- **`package.json`** — declare the five commands under `chatSessions[0].commands`.
- **`acpClient.ts` / `protocol.ts`** — type `available_commands_update`, route session-scoped updates *ahead of* the per-prompt lookup, cache them per session (`client.availableCommands(sessionId)`), log the live list to the output channel. This is now the drift detector against the server.
- **`sessionBridge.ts`** — `promptTextWithCommand()`. VS Code strips the command off the prompt and hands it over as `request.command`; the ACP server dispatches by *parsing prompt text* (`isAcpCommand` → `runAcpCommand`), so picking `/stop` would otherwise have sent an empty turn to the model. Applied at both call sites (participant handler + session content provider).

## Notes

- The commands already *worked* if typed blind — this was purely a discoverability gap.
- **JetBrains needs no change**: AI Assistant 2026.1+ is native ACP and consumes the notification directly, same as Zed.
- Drift guard test asserts the manifest list equals the server's `ACP_AVAILABLE_COMMANDS` (names + descriptions), so the two can't silently diverge.

## Testing

`bun test` 113 pass / 0 fail, `tsc --noEmit` clean, `bun run build` clean. New coverage: commands captured at `session/new` with no prompt in flight, commands update doesn't disturb prompt streaming, `promptTextWithCommand` round-trips (`/stop`, `/harness claude-code`, plain prompts, and a prompt that merely starts with a slash), manifest↔server drift.

Not yet verified in a live VS Code window — that needs the extension reloaded with proposed APIs enabled; happy to build a vsix and check the menu renders.


editors/vscode/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "phantombot-vscode",
   "displayName": "phantombot",
   "description": "Chat with phantombot — your personality-first agent — in VS Code's native Chat panel via an embedded ACP client.",
-  "version": "0.4.5",
+  "version": "0.4.6",
```
kaieriksen, line 5: The extension version is bumped to `0.4.6`, but `src/lib/vscodeExtensionAsset.generated.ts` still embeds `0.4.5` and `phantombot-vscode-0.4.5.vsix`. CI fails the embedded asset integrity tests (2 failures). Please run `bun run gen:vscode-vsix` and commit the regenerated asset module before merge.

kaieriksen (changes requested):
Request changes: the embedded VS Code asset was not regenerated after the extension version bump. CI has 2 failing asset-integrity tests; regenerate and rerun the checks.

andrewagrahamhodges:
Good catch — you were right, and it was worse than a red test.

`editors/vscode/package.json` went to 0.4.6 with the new `chatSessions.commands` contribution, but `src/lib/vscodeExtensionAsset.generated.ts` still held the 0.4.5 vsix bytes. So the CLI installer would have shipped the **stale extension — the one without the /-menu commands — while reporting 0.4.6**. The asset-integrity tests caught exactly the failure mode they were written for.

Regenerated with `bun run gen:vscode-vsix` in 907e302 (phantomyard.phantombot-vscode@0.4.6, 16585-byte vsix). All six checks green. Back to you.

kaieriksen (approved):
Re-reviewed at efab003. The embedded VS Code asset is regenerated for 0.4.6; 113 VS Code tests, 27 asset tests, typecheck, and CI pass.

PR #294, opened 2026-07-14
fix(memory): re-embed turns stranded behind the index cursor

### LINKED ISSUE #293: Turn indexer: an embed failure permanently strands a turn as FTS-only (no retry path)
## Summary

If the embedder fails on a turn, `turnIndexer` logs it, indexes the turn **FTS-only**, and still advances the index cursor past it. Nothing ever revisits that turn, so it stays lexical-only **forever**. It's a silent, one-way degradation of recall quality with no recovery path — not even a manual one.

## Mechanism

`src/orchestrator/turnIndexer.ts:135-142`:

```ts
const vec = embedder ? await embedTurn(embedder, turn, text) : undefined;
if (vec) embedded++;
else if (embedder) embeddingFailures++;
ix.upsertTurn(turn, vec, vec ? textSha : undefined);  // vec undefined -> no embedding row
indexed++;
afterId = turn.id;                                     // cursor advances anyway
```

`embedTurn` (line 283) swallows the failure and returns `undefined` — correct for the hot-path invariant ("never throw back into runTurn"), but the turn is then counted as `indexed`, `afterId` moves past it, and `updateTurnIndexState` persists the new `lastTurnId`.

`MemoryIndex.upsertTurn` (`src/lib/memoryIndex.ts:456`) writes the `turn_docs` (FTS) row unconditionally but only writes `turn_embeddings` when `vec && textSha`. So the turn exists in the lexical half and is simply absent from the vector half.

## Why there is no recovery

Every path into the indexer selects work with `turnsAfterId(persona, conversation, lastTurnId, …)` — i.e. **only the unindexed tail**. A turn that failed to embed is *behind* the cursor:

- the 20-user-turn batch trigger — won't revisit it
- the `flushAfterHours` age trigger — won't revisit it
- the 30-min heartbeat sweep (`flushDueConversationTurns`) — won't revisit it
- `phantombot memory index --turns` (`force: true`) — **won't revisit it either**; `force` only flushes the tail unconditionally, it does not rescan indexed turns

So the operator backfill escape hatch does not actually repair this. There is currently *no* way to re-embed a stranded turn short of deleting `turn_index_state` for the conversation and re-indexing from scratch.

## Impact

Low-severity but real. The FTS half still retrieves the turn, so it degrades gracefully day to day and is easy to miss. But embed failures are exactly the kind of thing that arrives in bursts (provider 429s, transient network, quota), and every turn in a burst window is permanently demoted to lexical-only. Recall quality erodes monotonically and invisibly.

## Notable

`MemoryIndex.turnEmbeddingSha(path)` (`src/lib/memoryIndex.ts:485`) is **defined and never called anywhere in the codebase**. It is precisely the primitive a re-embed pass would need — the hook was anticipated and never wired up.

## Suggested fix

Decouple "is indexed" from "is embedded". Options, roughly in order of preference:

1. **Repair pass in the heartbeat sweep.** Query `turn_docs LEFT JOIN turn_embeddings ... WHERE turn_embeddings.path IS NULL` (bounded, e.g. N per sweep) and re-embed those turns. Self-healing, no schema change, uses the dead `turnEmbeddingSha`/path primitives already present. Naturally rate-limits itself against a provider that's still failing.
2. **Don't advance the cursor past an embed failure** — simple, but stalls the whole conversation's indexing behind one poison turn, and would starve FTS for subsequent turns. Rejected unless paired with a skip counter.
3. **Explicit `embedding_pending` marker** on the turn doc + a drain job. Most correct, most machinery.

(1) is the smallest change that closes the hole. Should also surface a count — `embeddingFailures` is already returned by `indexConversationTurnsIfDue` and aggregated in `FlushDueConversationsResult`, but nothing acts on it or reports a *standing* backlog of unembedded turns.

## Repro

Point the embedder at a failing/rate-limited endpoint, run a conversation past the 20-turn trigger, restore the embedder, then run `phantombot memory index --turns`. The turns from the failure window have `turn_docs` rows but no `turn_embeddings` rows, and no invocation will ever add them.

andrewagrahamhodges:
Closes #293.

## The bug

A turn whose embedding call failed at index time stays **lexical-only forever**, and nothing can fix it — including the manual escape hatch.

`indexConversationTurnsIfDue` advances `lastTurnId` past a turn whether or not its embed succeeded. It *has* to: the alternative is one poisoned turn wedging the cursor and stalling the entire conversation. But the consequence is that a failed turn falls **behind** the cursor — and every path that selects work does so via `turnsAfterId(..., lastTurnId)`:

| Path | Sees a stranded turn? |
|---|---|
| 20-turn batch trigger | ❌ |
| `flushAfterHours` age flush | ❌ |
| 30-min heartbeat sweep | ❌ |
| `memory index --turns --force` | ❌ — force-flushes the *tail*, doesn't rescan |

So there was no recovery path at all. It's silent (the FTS half still returns hits, so recall degrades rather than breaks), it's one-way, and because embed failures arrive in **bursts** — 429s, quota, provider outages — a bad ten minutes permanently demotes every turn in that window. It only ever accumulates.

## The fix

A bounded, **cursor-free** repair pass: scan for FTS rows with no embedding row, re-embed them, done. It runs at the end of the existing per-persona sweep, so the heartbeat *and* `memory index --turns` both self-heal for free.

- **Bounded** — `repairBatchSize` per sweep (default 32, `0` disables).
- **Idempotent** — a repaired turn gains an embedding row and drops out of the scan, costing zero embedding calls thereafter.
- **Backs off** — bails after 3 consecutive failures rather than firing the whole budget at a provider that's evidently down. Next sweep picks up where it left off.
- Re-embeds `turn_docs.content` directly, so the text that gets vectorised is byte-identical to the text that's searchable.

`MemoryIndex.turnEmbeddingSha()` turned out to be defined and called from **nowhere** — the primitive for exactly this repair, anticipated and never wired up.

## Two properties worth reviewing closely

**Quarantined turns are excluded structurally, not by a filter.** The indexer never writes a `turn_docs` row for an `embeddable=0` turn, and the scan is rooted in `turn_docs` — so the repair pass *cannot* become a back door that resurrects a held untrusted payload into the vector index. That's by construction, not by a condition someone could later drop. Pinned by a test regardless.

**With no embedder configured, the pass no-ops.** An embeddings-disabled install has zero vectors by design; the repair must not conclude the entire history is damaged and try to "fix" it.

## Tests

Adds an `embedder` test seam to the indexer. There was previously **no way to inject a fake embedder** (`defaultEmbedder(config)` was called inline) — which is exactly why the embed-failure path shipped with no coverage at all.

9 new tests, including the one that pins the trap: `--force` with a *healthy* embedder still cannot recover a stranded turn. That's the invariant that makes the repair pass necessary rather than redundant, and it should fail loudly if someone ever "simplifies" this away.

`bun run typecheck` clean. Full suite: **2240 pass, 2 fail** — both failures are pre-existing on `main` (`harnesses-pi.test.ts` asserts `--api-key` is absent, but a real `PHANTOMBOT_PI_API_KEY` in the environment gets threaded through; unrelated to this change).

## Not addressed

A turn that fails *deterministically* (e.g. pathological content) gets retried once per sweep forever, burning one call a tick. Self-limiting and visible in the logs, but if it shows up in practice it wants a failure counter.

### DIFF
--- a/src/cli/heartbeat.ts
+++ b/src/cli/heartbeat.ts
@@ -95,7 +95,7 @@ export async function runHeartbeatCli(
           memory: store,
           settings: turnIndexing,
         });
-        if (flush.triggered > 0) {
+        if (flush.triggered > 0 || flush.repaired > 0) {
           log.info("heartbeat: flushed conversation turn tails", { ...flush });
         }
       } finally {
--- a/src/cli/memory.ts
+++ b/src/cli/memory.ts
@@ -280,6 +280,13 @@ export async function runMemoryIndex(
           (r.embeddingFailures > 0
             ? `, ${r.embeddingFailures} embed failure(s)`
             : "") +
+          // The repair pass is the only thing that can reach turns stranded
+          // behind the cursor by an earlier embed failure, so say so out loud
+          // when it does — otherwise a silent self-heal looks like a no-op run.
+          (r.repaired > 0 ? `, ${r.repaired} re-embedded (repair)` : "") +
+          (r.repairFailures > 0
+            ? `, ${r.repairFailures} repair failure(s)`
+            : "") +
           `\n`,
       );
     } finally {
--- a/src/config.ts
+++ b/src/config.ts
@@ -136,6 +136,19 @@ export interface TurnIndexingSettings {
    * to 0 to disable the time-based flush (count trigger only).
    */
   flushAfterHours: number;
+  /**
+   * Self-healing budget for turns that reached the FTS index but whose
+   * embedding call failed (a 429, a network blip, a provider outage). Those
+   * turns sit *behind* the `lastTurnId` cursor, so no cursor-driven path —
+   * the batch trigger, the age flush, the sweep, not even
+   * `memory index --turns --force` — will ever look at them again. They stay
+   * lexical-only forever, silently, and embed failures arrive in bursts, so a
+   * bad ten minutes permanently demotes every turn in that window.
+   *
+   * Each sweep re-embeds up to this many such turns, found by a cursor-free
+   * scan for FTS rows with no embedding row. Set to 0 to disable the repair.
+   */
+  repairBatchSize: number;
 }
 
 export const DEFAULT_TURN_INDEXING: TurnIndexingSettings = {
@@ -143,6 +156,7 @@ export const DEFAULT_TURN_INDEXING: TurnIndexingSettings = {
   interval: 20,
   batchSize: 200,
   flushAfterHours: 2,
+  repairBatchSize: 32,
 };
 
 /**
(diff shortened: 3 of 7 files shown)


kaieriksen (approved):
Reviewed at b16ebfff97bd9d8e5a1ac9d104411df08f29c09f. Repair pass is correctly persona-scoped, bounded, idempotent, and excludes quarantined turns. Focused config/indexer tests: 65 passed; typecheck passed; CI is green. Full suite: 2240 passed, 2 unrelated Pi credential-environment failures in this shell.

PR #295, opened 2026-07-15
fix(nightly): run maintenance stages MCP-free so a dead connector can't wedge essence

rgunst:
## Problem

Nightly runs report `nightly: ok` but the **essence stage fails most nights**:

```
stage 'essence': claude timed out after 300000ms with no output (likely wedged on a tool call)
```

Root cause is **MCP server startup, not a tool call inside the turn.** Each nightly stage spawns `claude --print`, which initialises every configured MCP server on startup — including the user's remote claude.ai connectors. When a remote connector is unauthenticated (here: `claude.ai Netsuite Production` and `claude.ai Microsoft 365`), it blocks the init handshake waiting on an OAuth flow that can never complete in a non-interactive `--print` run. essence is the *first* stage, so it eats the hang: zero output → killed at the 300000ms idle ceiling. "wedged on a tool call" is the harness guessing; it actually wedged before emitting anything. It's intermittent because it depends on how the dead endpoint responds that minute.

## Fix

Nightly maintenance needs no MCP at all. This adds an `mcpMode: "none"` flag on `HarnessRequest` (parallel to the existing `toolsMode: "none"`):

- `src/cli/nightly.ts` sets `mcpMode: "none"` on every stage turn.
- `src/orchestrator/turn.ts` threads it from `TurnInput` into the request.
- `src/harnesses/claude.ts` maps `"none"` → `--strict-mcp-config --mcp-config '{"mcpServers":{}}'`: use ONLY the (empty) inline config, ignoring `~/.claude.json` and connectors, so there are zero servers to init and nothing to hang on.

Interactive persona turns omit the flag and keep their Gmail / Calendar / Drive connectors untouched. Harnesses without a strict-MCP flag ignore `mcpMode` (they don't share claude's connector-startup hang).

## Tests

- New tests in `tests/harnesses-claude.test.ts`: `mcpMode: "none"` emits `--strict-mcp-config` + empty `--mcp-config`; a normal turn emits neither.
- `bun run typecheck` clean; `bun test tests/harnesses-claude.test.ts` 57/57; orchestrator/nightly/fallback suites 87/87.

### DIFF
--- a/src/cli/nightly.ts
+++ b/src/cli/nightly.ts
@@ -94,6 +94,10 @@ async function runNightlyTurn(opts: {
       idleTimeoutMs: STAGE_IDLE_TIMEOUT_MS,
       hardTimeoutMs: STAGE_HARD_TIMEOUT_MS,
       systemPromptSuffix: NIGHTLY_SUFFIX,
+      // Nightly needs no MCP; running MCP-free stops an unauthenticated remote
+      // connector from wedging the --print startup and killing a stage on the
+      // idle timeout (essence "timed out with no output"). See HarnessRequest.mcpMode.
+      mcpMode: "none",
     })) {
       if (chunk.type === "text") finalReply += chunk.text;
       if (chunk.type === "done") finalReply = chunk.finalText;
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -114,7 +114,7 @@ export class ClaudeHarness implements Harness {
     }
     try {
 
-    const args = this.buildArgs(req.systemPrompt, req.toolsMode, systemPromptFile);
+    const args = this.buildArgs(req.systemPrompt, req.toolsMode, systemPromptFile, req.mcpMode);
     log.debug("claude.invoke spawning", {
       bin: this.config.bin,
       argCount: args.length,
@@ -180,6 +180,8 @@ export class ClaudeHarness implements Harness {
     // (`--system-prompt-file`) instead of inline (`--system-prompt <text>`)
     // to stay under the command-line length limit.
     systemPromptFile?: string,
+    // `"none"` runs this turn with ZERO MCP servers — see the block below.
+    mcpMode?: "none",
   ): string[] {
     const args = [
       "--print",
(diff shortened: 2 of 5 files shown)


robertclawson (approved):
**Approved.** 🐶 Reviewed end-to-end at head `4e46abb`.

Correct fix, and I'm the sharper judge on this one since it's *my own* nightly that's been failing. The diagnosis holds: nightly stages spawn `claude --print`, which inits every configured MCP server on startup — including the remote claude.ai connectors — and an unauthenticated one blocks the init handshake on an OAuth flow that can never complete in a non-interactive run. `essence` is the first stage, so it eats the hang, emits nothing, and gets killed at the 300s idle ceiling. "wedged on a tool call" was the harness guessing; it wedged *before* the turn started.

**What I verified:**
- **`--strict-mcp-config --mcp-config '{"mcpServers":{}}'` is the right lever** — strict-mcp tells claude to use ONLY the inline config and ignore `~/.claude.json` + connectors; empty server map = zero servers to init = nothing to hang on. Args are pushed as discrete argv elements, so the JSON isn't shell-quote-sensitive.
- **Threading is complete**: `nightly.ts` → `runTurn({mcpMode})` → `invoke(req.mcpMode)` → `buildArgs` → the flags. Confirmed both nightly call sites (checkpointed loop + the single-shot path) go through `runNightlyTurn`, which hard-sets `mcpMode: "none"`, so **every** stage is covered.
- **Nightly genuinely needs no MCP** — the five stages are local memory/KB distillation; `memory_search` is a phantombot builtin, not an MCP connector. Nothing is lost by dropping MCP here.
- **Interactive path untouched** — `mcpMode` is optional and only nightly sets it, so persona turns keep Gmail/Calendar/Drive. Tests assert both directions (present on `none`, absent on a normal turn).

**One non-blocking note:** other non-interactive `--print` background turns share this exact connector-startup vulnerability — scheduled tasks that wake an LLM, `phantombot ask` from command-backed pollers. They're not covered here, which is defensible (a task might legitimately need Calendar), but if we see the same "timed out with no output" on a scheduled task, `mcpMode: "none"` is the ready fix. Worth a follow-up to let task authors opt in.

CI still pending (no check-runs reported yet) — mergeable once green and Kai weighs in.

kaieriksen (approved):
Approved at 4e46abb.

- 57 focused Claude harness tests passed
- Typecheck passed
- Full suite: 2242 passed, 2 skipped; 2 ambient Pi-key failures reproduced outside this PR
- Production build passed
- Linux, Windows, and typecheck CI passed
- Both nightly call paths use MCP-free mode; interactive turns remain unchanged

PR #296, opened 2026-07-16
fix(vscode): kill the Windows ACP untitled-cwd wedge + silent pi-only collapse

andrewagrahamhodges:
## What
Root-caused and fixed the VS Code ACP extension wedge Andrew hit on Megan's Windows box (`pi timed out after 300000ms` + `(phantombot declined this turn.)`).

## Three fixes

1. **`resolveSpawnCwd()`** (`acpClient.ts`) — Node's `spawn()` throws `ENOENT (-4058)` when the handed `cwd` doesn't exist on disk, not only when the binary is missing. An untitled/unsaved editor resource (`/untitled-1`) is never a real directory — every reopen on one wedged the ACP child before phantombot even started. Falls back to `homedir()`.

2. **`withHarnessInstallDirsOnPath()`** (new `editors/vscode/src/harnessPath.ts`) — prepends the same well-known install dirs the daemon's `harnessSearchPath()` covers (`~/.local/bin`, `~/.bun/bin`, `%APPDATA%\npm`, …) onto the ACP child's PATH. Detecting claude/pi by absolute path is already solved (`state.json`) — but once phantombot decides to *run* a harness, it spawns that binary as its own child, and Node shebang scripts / wrapper shims need their own tools on PATH. VS Code's inherited PATH is narrower than a login shell's on Windows, which silently collapsed the claude→pi chain to pi-only. Every editor turn then rode pi/GLM-5.2, which wedges on long tool calls and hits the 300s idle kill.

3. **Startup diagnostic** (`src/connectors/acp/server.ts`) — the ACP server now always logs `persona=<x> chain=[id:bin, ...]` to stderr on boot, forwarded by the extension to its output channel. This used to fail completely silently.

## Chat history / memory-embedding
No direct fix needed — investigated and confirmed it's the *same* root cause: turns only persist on success (`turn.ts`), so every idle-killed turn from bug #2 never got written, hence nothing to replay on `session/load` or embed. Fixing the collapse restores history/embedding for future turns going forward.

## Tests
- New `editors/vscode/tests/harnessPath.test.ts` (9 cases)
- `resolveSpawnCwd` cases appended to `acpClient.test.ts` (5 cases)
- One new diagnostic-log assertion in `tests/connectors-acp-server.test.ts`
- Full vscode extension suite (127 tests) and phantombot suite (2255 tests) pass. The 2 `harnesses-pi.test.ts` subprocess failures are pre-existing flakes on `main`, unrelated to this change (reproduced on `main` before this branch too).

### DIFF
--- a/editors/vscode/src/acpClient.ts
+++ b/editors/vscode/src/acpClient.ts
@@ -26,8 +26,11 @@
  */
 
 import { spawn } from "node:child_process";
+import { existsSync } from "node:fs";
+import { homedir } from "node:os";
 
 import { snapAwareSpawnEnv } from "./snapEnv.ts";
+import { withHarnessInstallDirsOnPath } from "./harnessPath.ts";
 import {
   ACP_PROTOCOL_VERSION,
   allocId,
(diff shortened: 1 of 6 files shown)


editors/vscode/src/acpClient.ts
```diff
@@ -396,6 +399,45 @@ export function buildAcpSpawnCommand(
...
+ * untitled-editor repro is unit-tested without touching the real filesystem.
+ */
+export function resolveSpawnCwd(
+  candidateCwd: string | undefined,
+  exists: (path: string) => boolean = existsSync,
+  home: () => string = homedir,
+): string {
+  if (candidateCwd && candidateCwd.trim()) {
+    try {
+      if (exists(candidateCwd)) return candidateCwd;
```
kaieriksen: [P1] Verify that the candidate is a directory, not merely existent\n\n`existsSync()` also returns true for regular files, so a file-valued `cwd` still reaches `spawn()` and fails with `ENOTDIR` (the helper contract and comments say this must be a real directory). Please use `statSync(candidateCwd).isDirectory()` (or inject a directory check) and add a regression test for an existing file.

kaieriksen (changes requested):
Changes requested: `resolveSpawnCwd()` must reject existing files as well as missing paths; otherwise an invalid file cwd still wedges the ACP spawn.

kaieriksen (approved):
Re-reviewed at 11746d87904070bd6729ba4e6df7abc972956996. The cwd guard now rejects existing files via statSync(...).isDirectory(), and the regression coverage is present. Focused/full VS Code tests pass (128), and the ACP server regression passes (23).

PR #299, opened 2026-07-17
fix(harness): make Pi's local-config option actually work, and stop hiding providers on a fresh install

robertclawson:
Two nits in the Pi harness wizard, both of which made the fresh-install path worse than it looked.

## 1. "Use Pi's own config" was a silent no-op

The old `later` branch returned early — it wrote nothing, but it also **cleared nothing**. Routing lives in *two* stores (`config.toml`'s `[harnesses.pi.routing]` **and** `~/.env`), and `harnesses/pi.ts` re-reads them every turn to thread `--model`/`--provider`/`--api-key`. So once "now" had run even once, its values persisted and "later" could never take effect — Pi kept getting the old primary/image/coding routing forever. The note claiming *"Pi falls back to its own local-store settings"* was only ever true on a virgin box.

Delegating to Pi now **actively erases both stores** (`computeRoutingClears` + `clearPiRouting`), after which `pi.ts` pushes no flags (it's already guarded on the values being present) and Pi reads `~/.pi/agent/settings.json` — the config the user created when they ran `pi` and logged into their provider.

The **API key is cleared too**, and that's load-bearing rather than tidiness: `pi --provider` **defaults to google**, so a surviving `PHANTOMBOT_PI_API_KEY` (say an OpenRouter key) with the provider erased would be fired at Google every turn and auth-fail.

## 2. The provider picker was empty on a fresh install

It derived providers from `pi --list-models`, which only lists models for providers Pi **already has a key for**. A fresh install prints `No models available. Use /login…` → no header row → `parsePiModels` returns `[]` → free-text fallback. That's the chicken-and-egg: *you can't pick a provider to key it, because it isn't listed until it's keyed* — precisely backwards, since the operator is in the wizard to add their **first** key.

Adds `PI_PROVIDER_CATALOG` (28 providers transcribed from Pi's own `docs/providers.md`) and `providerChoices()`, which **unions** the catalogue with whatever `--list-models` reported. The union matters both ways:
- catalogue-only (no key yet) → still offered ⇒ **the fresh-install fix**
- live-only (post-catalogue / custom provider) → still offered, never hidden by us being out of date

Keyed providers sort first with an *"key already configured"* hint, so a configured box still reads at a glance.

After the key is taken the catalogue is **refreshed** so the model pickers populate. This has to inject the provider's **native** env var (e.g. `OPENROUTER_API_KEY`) — verified against pi 0.79.1 that `--list-models` **ignores** `--api-key`/`--provider` and reads auth only from `~/.pi/agent/auth.json` + native vars:

```
$ env -i HOME=/tmp/fresh PATH=$PATH pi --list-models --provider openrouter --api-key sk-...
No models available. Use /login to log into a provider via OAuth or API key.

$ env -i HOME=/tmp/fresh PATH=$PATH OPENROUTER_API_KEY=sk-... pi --list-models
provider    model                            context  max-out  thinking  images
openrouter  ~anthropic/claude-opus-latest    1M       128K     yes       yes
... (257 lines)
```

If the refresh comes back empty (bad key, unknown provider, provider outage) the pickers **degrade to free-text** rather than dead-ending the wizard.

## Naming

`now` / `later` → **"Configure models"** / **"Use Pi's own config"**. The old wording described *when* you'd configure; the real choice is *who owns the config*.

## Tests

`bun test` → 2295 pass / 0 fail; `tsc --noEmit` clean.

New: `providerChoices` (fresh-install catalogue, keyed-first ordering, live-only providers, no dupes), `providerEnvVar`, `listPiModels` extraEnv passthrough, and `clearPiRouting` (clears both stores incl. the stale key, leaves unrelated env alone, no-op on a virgin box, plus a drift guard tying `computeRoutingClears` to `computeRoutingWrites`).

The TUI glue stays thin and manually verified, per this file's convention — all the new logic is pure and unit-tested.

### DIFF
--- a/src/cli/harness.ts
+++ b/src/cli/harness.ts
@@ -30,8 +30,11 @@ import {
   modelId,
   type PiModel,
   primaryIsMultimodal,
+  providerChoices,
+  providerEnvVar,
 } from "../lib/piModels.ts";
 import {
+  computeRoutingClears,
   computeRoutingWrites,
   ENV_PI_API_KEY,
   resolvePiApiKeyWrite,
@@ -133,6 +136,33 @@ export async function applyRouting(
   });
   await updateEnvFile(envPath, writes.env);
   return writes;
+}
+
+/**
+ * Erase every routing value the wizard owns, from BOTH stores — the
+ * "Use Pi's own config" path. See `computeRoutingClears` for why this must
+ * actively clear rather than merely skip writing (the old "later" branch was a
+ * no-op that silently kept stale routing alive forever).
+ *
+ * The empty `[harnesses.pi.routing]` table is left behind rather than deleted:
+ * `resolveRouting` maps an empty table to all-undefined, which is exactly the
+ * "no overrides" state we want, and keeping the table avoids churning the TOML
+ * shape. `envPath` is injectable for tests, matching applyRouting.
+ */
+export async function clearPiRouting(
+  configPath: string,
+  envPath: string = userEnvPath(),
+): Promise<ReturnType<typeof computeRoutingClears>> {
+  const clears = computeRoutingClears();
+  await updateConfigToml(configPath, (toml) => {
+    const routing = getIn(toml, ["harnesses", "pi", "routing"]) as
+      | Record<string, unknown>
+      | undefined;
+    if (!routing) return;
+    for (const key of clears.tomlKeys) delete routing[key];
+  });
+  await updateEnvFile(envPath, clears.env);
+  return clears;
 }
 
 function setRoutingKey(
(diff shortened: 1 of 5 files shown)


lenaparkhodges (approved):
**Approved.** Two real fresh-install bugs, both fixed at the right layers.

### Product / quality take
1. **"Use Pi's own config" was a silent lie** — old `later` skipped write but never cleared, so after any prior configure it kept threading stale `--model/--provider/--api-key`. Active dual-store erase via `computeRoutingClears` + `clearPiRouting` is the right fix. Clearing the API key is load-bearing (not tidiness): Pi defaults `--provider` to google, so a leftover OpenRouter key would keep auth-failing forever. Drift guard test tying clears to writes is exactly what this needs.
2. **Empty provider picker on virgin install** — deriving from `pi --list-models` alone is chicken-and-egg. Static `PI_PROVIDER_CATALOG` ∪ live models is the right shape; keyed-first ordering + free-text degrade on bad key/outage keeps the path resilient. Native-env injection for post-key refresh is the real required detail (verified against pi ignoring `--api-key` on `--list-models`).
3. **Rename** now/later → configure/local names the actual decision (who owns config), not timing. Better.

### Non-blocking notes
- Cloudflare catalogue entries only carry the key var (account/gateway IDs still needed) — already documented, free-text degrade is acceptable.
- Catalogue is a keep-in-sync obligation with Pi docs; the live-only union + free-text means drift is degraded not broken. Fine.

CI green (typecheck / test / test-windows / CodeQL). Tests actually pin the two regressions. Ship it.

kaieriksen (approved):
Approved at a0eb350ff4f1bce7ca96f10ff0b373632486515e.

Validated exact head: focused routing/provider tests 32/32, typecheck, production build, and GitHub CI (test, Windows, CodeQL, typecheck) pass. Full local suite: 2293 passed, 2 skipped; 2 unrelated Pi subprocess tests are contaminated by this host's ~/.env credential reload and do not touch this PR.

PR #300, opened 2026-07-17
feat(phantomchat): multi-mirror Blossom send/receive (bot half of PC #88)

lenaparkhodges:
## Summary

Bot-side half of phantomchat media reliability (pairs with [phantomchat#88](https://github.com/phantomyard/phantomchat/pull/88) / [#86](https://github.com/phantomyard/phantomchat/issues/86)).

Without this, the PWA can multi-mirror media and the bot still drops envelope `servers`, one-shot fetches the primary URL, and uploads bot voice to the dead primal/satellite/band trio.

## What / Why / How

**What:** Bot voice in/out and attachment download use the same multi-host Blossom model as the PWA.

**Why:** Public Blossom hosts die, mime-filter, or paywall. A single URL is a single point of failure. Our previous default trio was already broken for free encrypted `application/octet-stream`.

**How (simple, same design as PC #88):**
1. Solid free server list from website `https://chat.phantomyard.ai/blossom.json` (same pattern as `relays.json`). Disaster-net default: `nostr.download`, `blossom.ditto.pub`, `blossom.data.haus`.
2. Write ≥2 mirrors on bot voice upload; put `servers: [...]` on the envelope.
3. Multi-GET receive: primary URL → envelope mirrors → hash GET `{server}/{sha256}` on the known list. Always verify sha256 before decrypt.

## Wire shape

- Compatible with old clients: primary `url` still required; `servers` is optional.
- Old phantombot clients ignore `servers` and keep one-shot GET (no regression if primary is healthy).

## Test plan

- [x] unit: blossomServers fetch + expand
- [x] unit: multi-GET fallback (primary 404 → mirror; dead primary → hash GET)
- [x] unit: sendVoice multi-mirror PUT + envelope.servers
- [x] unit: channel plumbs servers from DM + group envelopes
- [x] channel + server test suites green
- [ ] after merge: voice note PWA → bot with primary host blocked; STT still fires
- [ ] after merge: bot voice reply lands with ≥2 hosts; PWA plays after killing primary

## Out of scope

Media outbox, self-hosted Blossom, group bot voice egress (DM-only).

### DIFF
--- a/src/channels/phantomchat/blossomFetch.ts
+++ b/src/channels/phantomchat/blossomFetch.ts
@@ -68,41 +68,69 @@ export async function fetchAndDecryptBlossom(
   }
 
   const fetchImpl = opts?.fetchImpl ?? fetch;
-  const known =
-    opts?.knownServers ??
-    (await getBlossomServers().catch(() => DEFAULT_BLOSSOM_SERVERS));
-  const candidates = expandBlossomFetchUrls(
-    url,
-    opts?.expectedSha256Hex,
-    opts?.mirrors,
-    known,
-  );
   const errors: string[] = [];
 
-  for (const candidate of candidates) {
-    try {
-      const res = await fetchImpl(candidate, { signal: opts?.signal });
-      if (!res.ok) {
-        errors.push(`${candidate}: HTTP ${res.status}`);
-        continue;
-      }
-      const ciphertext = Buffer.from(await res.arrayBuffer());
+  // Happy path first: primary URL + envelope mirrors only. Do NOT resolve the
+  // website server list until those fail *and* we have a sha256 to hash-GET
+  // against — otherwise every inbound voice note pays a blossom.json hop
+  // (today often a 404) out of the STT timeout budget.
+  const direct = expandBlossomFetchUrls(
+    url,
+    undefined,
+    opts?.mirrors,
+    [],
+  );
 
-      // Blossom is content-addressed by sha256 of the (encrypted) bytes.
-      // Verifying catches a corrupt/tampered/wrong-blob fetch before we waste
-      // a GCM auth failure (or a paid STT call) on it.
-      if (opts?.expectedSha256Hex) {
-        const got = createHash("sha256").update(ciphertext).digest("hex");
-        if (got !== opts.expectedSha256Hex.toLowerCase()) {
-          errors.push(`${candidate}: sha256 mismatch`);
+  const tryCandidates = async (candidates: string[]): Promise<Buffer | null> => {
+    for (const candidate of candidates) {
+      if (opts?.signal?.aborted) {
+        throw new Error("blossom fetch aborted");
+      }
+      try {
+        const res = await fetchImpl(candidate, { signal: opts?.signal });
+        if (!res.ok) {
+          errors.push(`${candidate}: HTTP ${res.status}`);
           continue;
         }
-      }
+        const ciphertext = Buffer.from(await res.arrayBuffer());
 
-      return decryptCiphertext(ciphertext, key, iv);
-    } catch (e) {
-      errors.push(`${candidate}: ${(e as Error).message}`);
+        // Blossom is content-addressed by sha256 of the (encrypted) bytes.
+        // Verifying catches a corrupt/tampered/wrong-blob fetch before we waste
+        // a GCM auth failure (or a paid STT call) on it.
+        if (opts?.expectedSha256Hex) {
+          const got = createHash("sha256").update(ciphertext).digest("hex");
+          if (got !== opts.expectedSha256Hex.toLowerCase()) {
+            errors.push(`${candidate}: sha256 mismatch`);
+            continue;
+          }
+        }
+
+        return decryptCiphertext(ciphertext, key, iv);
+      } catch (e) {
+        if (opts?.signal?.aborted) throw e;
+        errors.push(`${candidate}: ${(e as Error).message}`);
+      }
     }
+    return null;
+  };
+
+  const directHit = await tryCandidates(direct);
+  if (directHit) return directHit;
+
+  // Only pay for the known-server list when hash-GET can actually help.
+  if (opts?.expectedSha256Hex && /^[0-9a-fA-F]{64}$/.test(opts.expectedSha256Hex)) {
+    const known =
+      opts?.knownServers ??
+      (await getBlossomServers().catch(() => DEFAULT_BLOSSOM_SERVERS));
+    const hashUrls = expandBlossomFetchUrls(
+      // dummy primary already tried; expand only the hash GETs
+      "",
+      opts.expectedSha256Hex,
+      undefined,
+      known,
+    ).filter((u) => u && !direct.includes(u));
+    const hashHit = await tryCandidates(hashUrls);
+    if (hashHit) return hashHit;
   }
 
   throw new Error(`blossom fetch failed: ${errors.join("; ")}`);
--- a/src/channels/phantomchat/blossomFetch.ts
+++ b/src/channels/phantomchat/blossomFetch.ts
@@ -1,26 +1,62 @@
 /*
  * PhantomChat media: fetch a file from Blossom and AES-256-GCM decrypt it.
  *
- * Counterpart to the PWA's encrypt path (phantomchat/src/lib/phantomchat/
- * file-crypto.ts). The PWA encrypts with Web Crypto `AES-GCM`:
+ * Counterpart to the PWA's encrypt path (phantomchat file-crypto.ts) and
+ * multi-GET receive (phantomchat-file-fetch.ts). Tries the primary URL first,
+ * then any mirrors carried in the envelope, then hash-addressed GETs against
+ * our known Blossom server list. Verifies sha256 of the ciphertext when the
+ * sender provided one before decrypting.
+ *
+ * The PWA encrypts with Web Crypto `AES-GCM`:
  *   - 256-bit key (32 bytes), 96-bit IV (12 bytes), both hex on the wire
  *   - Web Crypto APPENDS the 128-bit (16-byte) auth tag to the ciphertext
  *     (output is ciphertext‖tag), whereas Node's GCM wants the tag supplied
  *     separately — so we split the trailing 16 bytes and `setAuthTag`.
- * The encrypted bytes live at a Blossom URL; the key/iv travel inside the
- * NIP-17 gift-wrap envelope, so only the recipient can decrypt.
+ * The key/iv travel inside the NIP-17 gift-wrap envelope, so only the
+ * recipient can decrypt. Blossom only ever sees ciphertext.
  */
 import { createDecipheriv, createHash } from "node:crypto";
+import {
+  DEFAULT_BLOSSOM_SERVERS,
+  expandBlossomFetchUrls,
+  getBlossomServers,
+} from "./blossomServers.ts";
 
 const GCM_TAG_BYTES = 16;
 const KEY_BYTES = 32;
 const IV_BYTES = 12;
 
+export interface FetchDecryptOpts {
+  expectedSha256Hex?: string;
+  /** Extra mirror URLs from the envelope (`servers` field). */
+  mirrors?: readonly string[];
+  /** Override known servers used for hash-GET fallback. */
+  knownServers?: readonly string[];
+  signal?: AbortSignal;
+  fetchImpl?: typeof fetch;
+}
+
+function decryptCiphertext(
+  ciphertext: Buffer,
+  key: Buffer,
+  iv: Buffer,
+): Buffer {
+  if (ciphertext.length < GCM_TAG_BYTES) {
+    throw new Error("blossom: ciphertext shorter than the GCM auth tag");
+  }
+  const tag = ciphertext.subarray(ciphertext.length - GCM_TAG_BYTES);
+  const body = ciphertext.subarray(0, ciphertext.length - GCM_TAG_BYTES);
+
+  const decipher = createDecipheriv("aes-256-gcm", key, iv);
+  decipher.setAuthTag(tag);
+  return Buffer.concat([decipher.update(body), decipher.final()]);
+}
+
 export async function fetchAndDecryptBlossom(
   url: string,
   keyHex: string,
   ivHex: string,
-  opts?: { expectedSha256Hex?: string; signal?: AbortSignal },
+  opts?: FetchDecryptOpts,
 ): Promise<Buffer> {
   const key = Buffer.from(keyHex, "hex");
   const iv = Buffer.from(ivHex, "hex");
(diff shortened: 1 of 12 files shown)


src/channels/phantomchat/blossomFetch.ts
```diff
@@ -31,29 +67,43 @@ export async function fetchAndDecryptBlossom(
...
-  if (opts?.expectedSha256Hex) {
-    const got = createHash("sha256").update(ciphertext).digest("hex");
-    if (got !== opts.expectedSha256Hex.toLowerCase()) {
-      throw new Error("blossom: sha256 mismatch (corrupt or tampered file)");
-    }
-  }
+  const fetchImpl = opts?.fetchImpl ?? fetch;
+  const known =
+    opts?.knownServers ??
+    (await getBlossomServers().catch(() => DEFAULT_BLOSSOM_SERVERS));
```
andrewagrahamhodges: 🔴 **Blocker — the list is fetched on every receive, and fetched *eagerly*.**

Two separate problems stacked here:

1. **No cache.** The module header says this continues "the same single-source pattern as `relays.json` / `relaysSource.ts`" — but that's only half true. `relaysSource` fetches **once at startup** and caches the result into the persona's `phantomchat.json`, so a later offline start still works and steady-state costs zero requests. `getBlossomServers()` has no cache at all: it calls `fetchCanonicalBlossomServers()` fresh every single time.

2. **Resolved before it's needed.** This line runs *before* the primary URL is tried — and if `expectedSha256Hex` is absent, `expandBlossomFetchUrls` never uses `known` at all. So we pay for a list we may not touch.

Why it matters concretely: right now `https://chat.phantomyard.ai/blossom.json` returns **404** (I curl'd it — PC #88 isn't deployed yet), so *today* every inbound voice note makes one guaranteed-wasted request before touching the media. Once it's live, the failure mode is worse: a hung host burns up to the full `timeoutMs = 5000`, and that's spent out of the **same** `sttTimeoutMs` budget `server.ts:612` wraps this call in. The happy path pays for the sad path.

**Proposed fix — two small changes.**

Cache with a TTL in `blossomServers.ts`:

```ts
let cache: { servers: readonly string[]; at: number } | null = null;
const TTL_MS = 60 * 60 * 1000;

export async function getBlossomServers(opts?: {
  servers?: readonly string[];
}): Promise<readonly string[]> {
  if (opts?.servers && opts.servers.length > 0) {
    return opts.servers.map((s) => normalizeServer(s));
  }
  if (cache && Date.now() - cache.at < TTL_MS) return cache.servers;
  const fetched = await fetchCanonicalBlossomServers();
  const servers = fetched && fetched.length > 0 ? fetched : DEFAULT_BLOSSOM_SERVERS;
  // Only cache a real answer — a 404 today shouldn't pin us to the seed for an hour.
  if (fetched) cache = { servers, at: Date.now() };
  return servers;
}
```

And make resolution lazy here — only reach for the known list once primary + envelope mirrors have all failed:

```ts
const direct = expandBlossomFetchUrls(url, undefined, opts?.mirrors, []);
// ...try `direct` first...
// only if every direct candidate failed AND we have a sha256:
const known = opts?.knownServers ?? (await getBlossomServers().catch(() => DEFAULT_BLOSSOM_SERVERS));
```

That keeps the happy path at exactly the request count it has today, and confines the website round-trip to the case that actually needs it. If you'd rather mirror `relaysSource` exactly and cache to `phantomchat.json` for offline-start parity, that works too — I just don't want the *claim* of the relays pattern without the caching half of it.

src/channels/phantomchat/channel.ts
```diff
@@ -302,6 +302,16 @@ export function createPhantomchatChannel(
...
               const mt = typeof fm.mediaType === "string" ? fm.mediaType : (envelope.type as string);
+              // Optional multi-mirror list (phantomchat #88). Accept either
+              // `servers` (canonical) or `mirrors` (older draft) — string URLs only.
+              const rawServers = Array.isArray(fm.servers)
+                ? fm.servers
+                : Array.isArray(fm.mirrors)
+                  ? fm.mirrors
+                  : undefined;
+              const servers = rawServers
+                ?.filter((u: unknown): u is string => typeof u === "string" && u.startsWith("http"))
```
andrewagrahamhodges: 🟠 **`servers` is sender-controlled, unbounded, and `http://` gets through.**

This filter accepts `startsWith("http")` — which passes `http://`. Meanwhile `blossomServers.validServers` (line 51) correctly demands `https://`. Same concept, two different bars, and the looser one is the one fed by a remote party.

The unbounded part is the bigger half: nothing caps the array length. A sender can put 500 URLs in `servers` and `fetchAndDecryptBlossom` will walk **all of them sequentially** on a failed fetch, each with its own connect timeout, inside the `sttTimeoutMs` budget. To be fair — the primary `url` was always sender-controlled, so this isn't a *new* trust boundary. But it multiplies the cost of the existing one from 1 request to N.

sha-verify-before-decrypt means a hostile mirror can't feed us bad plaintext, which is genuinely good and worth calling out. The exposure is the fetch fan-out, not the crypto.

**Proposed fix:**

```ts
const servers = rawServers
  ?.filter((u: unknown): u is string => typeof u === "string" && u.startsWith("https://"))
  .map((u: string) => u.replace(/\/+$/, ""))
  .slice(0, 8);
```

8 is arbitrary but comfortably above any real mirror count (we write ≥2). Worth a brief comment noting these are sender-supplied so the bound isn't mysteriously removed later.

src/channels/phantomchat/blossomUpload.ts
```diff
@@ -80,14 +98,25 @@ export async function uploadToBlossom(
         errors.push(`${server}: no url in response`);
         continue;
       }
-      return { url: data.url, sha256: data.sha256 || sha256Hex };
+      if (!mirrors.includes(data.url)) mirrors.push(data.url);
+      if (data.sha256) firstSha = data.sha256;
```
andrewagrahamhodges: 🟠 **`firstSha` trusts the server's self-reported hash — and despite the name, the *last* mirror wins.**

This is the same finding I left on the PWA side (PC #88); the two halves ported it together, so let's fix them in the same pass rather than let the twins drift.

Two things wrong in one line:

1. **The name lies.** `firstSha` gets reassigned on *every* successful mirror, so line 119 returns whatever the **last** server echoed, not the first.

2. **The value is untrustworthy by construction.** We computed `sha256Hex` ourselves from the exact bytes we're about to PUT. That's ground truth. This line throws it away in favour of a number the *remote server* reports about our own data — and that value is precisely what the receiver uses to **verify integrity**. Trusting the server's self-report inverts the point of the check: a server that echoes a wrong (or malicious) hash makes the media permanently unfetchable, because every candidate then fails `sha256 mismatch` and `blossomHashUrl` builds dead URLs off it.

It's latent today — `transport.ts:583` uses `enc.sha256Hex` for the envelope and discards `uploaded.sha256` — but it's *exported* as the authoritative sha, so the next caller to trust it inherits the bug.

**Proposed fix:** delete `firstSha` entirely and return what we computed.

```ts
return { url: mirrors[0]!, sha256: sha256Hex, mirrors };
```

If you want to keep a server disagreement visible, a `log.debug` when `data.sha256 && data.sha256 !== sha256Hex` is genuinely useful signal (it means that host mangled our bytes) — just don't let it win.

src/channels/phantomchat/blossomUpload.ts
```diff
@@ -80,14 +98,25 @@ export async function uploadToBlossom(
...
     } catch (e) {
-      errors.push(`${server}: ${(e as Error).message}`);
+      const msg = (e as Error).message;
+      if (msg === "upload aborted") throw e;
```
andrewagrahamhodges: ⚪ Nit, non-blocking. The abort re-throw keys on the exact string `"upload aborted"`, but an abort landing *mid-PUT* comes out of `fetch` as a `DOMException` with `"The operation was aborted"` (or `"This operation was aborted"` depending on runtime). So it doesn't match, gets swallowed into `errors[]`, and we waste one more server iteration before the top-of-loop check at line 78 catches it. Harmless — the loop does exit — just one wasted round-trip on cancel.

```ts
if (msg === "upload aborted" || opts?.signal?.aborted) throw e;
```

Checking the signal directly is more robust than string-matching the runtime's phrasing.

andrewagrahamhodges (changes requested):
Nice work Lena — and thank you for doing the bot half properly rather than bolting mirrors onto the send path and calling it done. The core shape is right: one shared list, ≥2-mirror write with a ≥1 floor so a single CDN can't brick a note, sha-verify-before-decrypt on every candidate, `servers` optional so pre-#88 clients don't regress, and a deprecation shim on `BLOSSOM_SERVERS` instead of a breaking rename. I checked the group claim in the test plan against the code and it holds — `channel.ts:288` really does route both shapes through one path. CI's green.

Four threads below. One blocker, two I'd like fixed before merge, one nit.

**The blocker is a caching one, and I want to be clear it's not a nitpick about performance.** The header says this follows the same single-source pattern as `relays.json` — but `relaysSource` fetches *once at startup* and caches into `phantomchat.json`. `getBlossomServers()` has no cache at all, and `blossomFetch.ts:73` resolves the list *before the primary URL is even tried*. So every inbound voice note pays a website round-trip before we touch the media — and today `chat.phantomyard.ai/blossom.json` is **404** (I curl'd it; PC #88 isn't deployed), so that's a guaranteed-wasted request per media message right now. Once it's live, a hung host burns up to 5s out of the *same* `sttTimeoutMs` budget `server.ts` wraps the call in. Details and a two-line fix inline.

The `firstSha` one is the same finding I left on PC #88 — the twins ported it together, so it's worth fixing in both at once so they don't drift.

Happy to take any of these myself if you'd rather stay on one path — just say which.

src/channels/phantomchat/blossomFetch.ts
lenaparkhodges: Fixed on 0284634.

1. `getBlossomServers` now TTL-caches for 1h, **only on a real website answer** — a 404/empty/throw falls back for that call only and retries next time (seed is not sticky).
2. `fetchAndDecryptBlossom` tries primary + envelope mirrors first with **no** known-server collection. Website list / hash-GET is resolved only after those fail **and** we have a valid sha256.

Happy path today: zero blossom.json hits. New unit covers both halves.

src/channels/phantomchat/channel.ts
lenaparkhodges: Fixed on 0284634.

Sender-supplied `servers`/`mirrors` now:
- require `https://` (matches upload-side `validServers`)
- strip trailing `/`
- `.slice(0, 8)` with a comment that the bound exists because these are sender-controlled fan-out.

src/channels/phantomchat/blossomUpload.ts
lenaparkhodges: Fixed on 0284634 (and confirmed twin-aligned with PC #88 head — local `hash` already wins there).

Deleted `firstSha`. Result always returns our computed `sha256Hex`. Server echo on mismatch is `log.debug` only — never authoritative for the envelope/receiver.

src/channels/phantomchat/blossomUpload.ts
lenaparkhodges: Fixed on 0284634.

Catch branch now rethrows on `opts?.signal?.aborted` instead of matching the `"upload aborted"` string. Mid-PUT abort no longer burns another server iteration.

lenaparkhodges:
Addressed all four review threads on `0284634`:

- 🔴 **eager list / no cache** — TTL cache on real answers only; resolve known list only after primary+envelope mirrors fail
- 🟠 **servers bounds** — `https://` only + `.slice(0, 8)`
- 🟠 **firstSha** — local computed hash always; server mismatch is debug-only (PC #88 twin already fixed)
- ⚪ **abort** — check `signal.aborted`, not the exception message

29 related unit tests green. Ready for re-review.

lenaparkhodges:
@andrewagrahamhodges all four #300 findings were addressed on `0284634` (already on the branch; CI green at that commit). Summary:

| finding | fix |
|---|---|
| 🔴 no cache + eager list resolve | `getBlossomServers` TTL-caches 1h, **only on a real website answer** — 404/empty is not sticky. `blossomFetch` resolves known servers **only after** primary + envelope mirrors fail *and* we have a sha to hash-GET against. |
| 🟠 sender-controlled `servers` unbounded / `http` | require `https://`, strip trailing `/`, `.slice(0, 8)` with a bound-because-sender-supplied comment |
| 🟠 `firstSha` trusts server echo | deleted; result always our local `sha256Hex`; server mismatch is `log.debug` only |
| ⚪ abort string-match | rethrow on `opts?.signal?.aborted` |

Ready for re-review.

src/channels/phantomchat/blossomUpload.ts
```diff
@@ -60,7 +71,14 @@ export async function uploadToBlossom(
   const authHeader = "Nostr " + Buffer.from(JSON.stringify(event)).toString("base64");
 
   const errors: string[] = [];
+  const mirrors: string[] = [];
+
   for (const server of servers) {
```
andrewagrahamhodges: 🟠 **Sequential mirrors — and here it delays the message, not just the tick.**

This loop awaits each PUT in turn and breaks at `minMirrors`, so a 2-mirror upload costs `t1 + t2`. I flagged the same loop on the PWA side (PC #88) and initially called it non-blocking; I was wrong there and it's more clearly wrong here.

The difference is where the await sits. On the PWA the bubble already exists and just gets a late ✓ — cosmetic. Here `transport.ts:564` awaits the whole upload **before the envelope is published**, so the recipient doesn't get the voice note *at all* until both mirrors finish. That's a late message, not a late tick. And it's spending that time inside the same `sttTimeoutMs` budget `server.ts` wraps the call in, so a slow second mirror can convert "slower" into "didn't send."

Same fix Lena landed on #88 — fan out with `Promise.allSettled`, evaluate durability after the settle:

```ts
const targets = servers.slice(0, minMirrors);
const results = await Promise.allSettled(
  targets.map((server) => putTo(server, blob, authHeader, opts?.signal))
);
// ≥1 required, ≥2 preferred — same floor, just not serialized.
```

The ≥1 floor and ≥2 preference are unchanged; only the scheduling moves. Worth slicing to `minMirrors` here rather than mapping the whole list (that's the one thing #88's version got wrong — it fans out to all three and always pays 3× egress). Keeping the twins aligned matters more than usual on these two, so it's probably worth doing after #88's cap lands so you can copy the final shape.

Happy to take this one as a commit on your branch if you'd rather stay on the receive path — just say.

src/channels/phantomchat/channel.ts
```diff
@@ -302,6 +302,21 @@ export function createPhantomchatChannel(
...
               const mt = typeof fm.mediaType === "string" ? fm.mediaType : (envelope.type as string);
+              // Optional multi-mirror list (phantomchat #88). Accept either
+              // `servers` (canonical) or `mirrors` (older draft). Sender-
+              // controlled, so https-only + hard cap of 8 (bound fan-out).
+              const rawServers = Array.isArray(fm.servers)
+                ? fm.servers
+                : Array.isArray(fm.mirrors)
+                  ? fm.mirrors
+                  : undefined;
+              const servers = rawServers
```
andrewagrahamhodges, line 313: ⚪ **The filter is correct but nothing guards it — I checked by reverting it.**

I mutation-tested this: put it back to `startsWith("http")`, deleted the `.slice(0, 8)`, ran the channel suite. **All 10 tests still passed.** Both fixtures use `https://` with 2 servers, so they exercise the new code and the old code identically.

That's precisely the failure mode I was worried about in the original thread — the comment explains the bound, but a comment doesn't fail CI. Someone refactoring this in six months gets a green suite either way.

Two fixtures close it:

```ts
it("drops non-https mirrors", () => {
  // http:// mirror is sender-controlled — must not survive the filter
  expect(parse({ servers: ["https://a.example", "http://b.example"] }).servers)
    .toEqual(["https://a.example"]);
});

it("bounds fan-out at 8 servers", () => {
  const many = Array.from({ length: 20 }, (_, i) => `https://s${i}.example`);
  expect(parse({ servers: many }).servers).toHaveLength(8);
});
```

Non-blocking on the code — your fix is right, and sha-verify-before-decrypt means this is a fan-out bound rather than a crypto hole. I'm only holding the review open for the mirrors item above; land these two fixtures with it and this is done.

andrewagrahamhodges (changes requested):
Re-reviewed at `0284634`. **All four of my original findings are genuinely fixed** — lazy+TTL'd list resolution (and not caching a 404/empty, which is better than I asked for), `https://`-only with a bound, `firstSha` gone, abort via `signal.aborted`. CI fully green. Nice work.

Two things left, both from re-reading rather than from your changes. Neither is a redo of something you already fixed.

The first is the twin of an escalation on PC #88 — I originally filed it there as non-blocking and was talked out of that, correctly. It applies here too, and it's **worse on this side**, so I'm not going to make the same mistake twice by leaving it as a follow-up.

src/channels/phantomchat/blossomUpload.ts
lenaparkhodges: Fixed on head (`593886f`). Copied #88’s final shape exactly:

```ts
const targets = servers.slice(0, minMirrors);
const results = await Promise.allSettled(targets.map(...));
```

Voice envelope publish now waits on `max(t)` of the durability set, not `t1+t2`, and we don’t pay 3× egress. ≥1 floor unchanged. Shared abort rechecked after settle.

src/channels/phantomchat/channel.ts
lenaparkhodges: Landed the two fixtures next to the receive intake tests (`drops non-https mirrors…`, `bounds fan-out at 8 servers on receive`). Mutation-covered now — green suite requires `https://` + `.slice(0, 8)`.

andrewagrahamhodges (approved):
## Approving — verified live end-to-end against the real hosts

`593886f`. All findings closed, and I mutation-tested each rather than taking the checkboxes on trust:

- **Lazy + cached list** — reverted it to eager-resolve, the new test went red. ✅ Not caching a 404/empty is better than what I asked for.
- **`https://`-only + `.slice(0, 8)`** — you added the two fixtures I asked for. I reverted the filter to `startsWith("http")` **and** dropped the cap: both tests failed. That gap is now actually guarded. ✅
- **Local sha** — `firstSha` gone, mismatch is `log.debug` only. Twin-aligned with PC #88. ✅
- **Abort** — checks `signal.aborted`, not the runtime's phrasing. ✅
- **Parallel fan-out + cap** — same shape as #88, so the twins match. ✅

### Live end-to-end through the real code

Ran your actual `uploadToBlossom` → `fetchAndDecryptBlossom` against the real hosts: 64KB AES-GCM blob encrypted PWA-style (ciphertext‖tag), `audio/ogg`, default `minMirrors=2`.

| scenario | result |
|---|---|
| upload, 2 mirrors parallel | **1011ms** — max(t), not t1+t2 |
| envelope sha === local sha | ✅ |
| happy path: primary URL | **PASS** (173ms, byte-identical) |
| primary dead → envelope mirror | **PASS** (163ms) |
| primary + mirrors dead → hash-GET fallback | **PASS** (139ms) |
| tampered sha (expect `ffff…`) | **PASS** — rejected, not decrypted |

All three hosts individually pass a write-probe with a throwaway pubkey (no whitelist). The parallel fan-out landing at ~1s matters here more than on the PWA — `transport.ts:564` blocks the envelope publish, so this is time straight out of the `sttTimeoutMs` budget.

### One thing worth knowing about `nostr.download`

It returns extension-suffixed URLs — `https://nostr.download/<sha>.oga` for `audio/ogg`, `.bin` for octet-stream. The envelope carries that full URL so the primary path is fine, but `expandBlossomFetchUrls` builds bare `<host>/<sha>` for hash-GET fallback. I checked: **`nostr.download` serves the bare hash too (200)**, so the fallback resolves. Not a bug — just don't let anyone "simplify" the URL handling on the assumption that hosts echo back a bare-hash URL, because one of the three doesn't.

### Merge order — the one real deployment hazard

`chat.phantomyard.ai/blossom.json` is **404 right now** (re-checked). This PR reads it, so:

**PC #88 merges → site deploy lands → confirm `blossom.json` is 200 → then this.**

Out of order it degrades rather than breaks (seed defaults cover it, and the not-caching-a-404 fix means it recovers on its own once the site is up) — but the "single source" premise is inert until then, and nobody would notice.

Approving on that condition. Nothing further from me.

Issue #301, opened 2026-07-19
[Discussion] Telegram interrupts active turns, PhantomChat queues them — inconsistency hurts when harness breaks


14skywalker:
## Context

PhantomChat and Telegram currently handle incoming messages very differently when a turn is already in progress.

## Current behavior

**Telegram** (`core/engine.ts`):
- When a new message arrives, it aborts the active turn (`active.controller.abort("interrupt")`) and starts the new one immediately.
- Effect: typing a new message effectively says "stop what you're doing and answer this instead."

**PhantomChat** (`channels/phantomchat/server.ts`):
- Uses a per-peer serial chain (`prev.catch(() => {}).then(() => handle(msg))`).
- New messages are appended to the queue and wait stoically for the current turn to finish.

## The problem

When a harness breaks (e.g. the *pi* harness entering a wedged state where every turn gets idle-killed after 300s), the PhantomChat queue turns a single failure into a repeated bad experience:

1. User sends message A → pi hangs → `/stop` kills it after 5 min.
2. But queued messages B, C, D are still waiting.
3. Each one starts pi again → each one hangs again → user has to `/stop` repeatedly.
4. On Telegram this is invisible because new messages interrupt; on PhantomChat it is painfully visible.

Real log excerpt from a session with a broken pi harness:

| Time (local) | Event |
|---|---|
| 07:24:51 | Message starts → pi invoked |
| 07:34:21 | `/stop` → killed after 570s |
| 07:34:22 | Next queued message → pi invoked again |
| 07:34:36 | `/stop` → killed after 14s |
| 07:34:47 | Next queued message → pi invoked again |
| 07:39:27 | User switches to Telegram → same pi hang, but single-turn |
| 07:47:46 | `pi.invoke killed: idle` (300s timeout) |
| 07:48:04 | Telegram turn completes (517s, fell back to codex) |

## Questions for discussion

1. **Should PhantomChat adopt Telegram's "interrupt" behavior?**
   A new message aborts the active turn *and* drops the backlog, so the user can always "break out" of a stuck harness.

2. **Should `/stop` flush the entire queue?**
   Currently `/stop` only aborts the active turn; queued messages still run. This is particularly painful in serial-queue channels.

3. **Is the serial queue intentional for PhantomChat?**
   The iterator-based `listen()` loop may make an interrupt model harder, but it would align channel behavior and improve UX when things go wrong.

4. **Should the harness/orchestrator detect repeated idle-kills and auto-fallback?**
[...]

PR #302, opened 2026-07-20
feat(channels): unified interrupt / stop semantics across all channels (#301)

robertclawson:
Implements the interrupt / `/stop` semantics Andrew and I agreed on in #301, applied uniformly across **Telegram, PhantomChat and the ACP connector**.

## Semantics
- **Interrupt** (a new message arrives while a turn is running) = *"I've changed my mind — listen to me now."* Aborts the active turn **and drops the queued backlog** for that conversation, **silently**. You sent the interrupting message, so you already know what you cancelled; a "dropped N messages" notice would just be noise.
- **`/stop`** = break-glass panic. Aborts the active turn, drops the backlog, replies with a **visible ack + dropped count**, and writes a `[system]` note into history so the agent's *next* turn knows it was stopped, drops the work, and **awaits instructions** instead of resuming.

This is a deliberate departure from the "always report the drop" position (#301): the visible notice lives on `/stop` only; interrupt is intentionally quiet.

## Mechanism
Channels serialize a conversation onto a promise chain (`next = prev.then(() => run(msg))`). A `.then()` callback can't be un-attached, so we make superseded ones **no-op**: a per-conversation **epoch** (`ConversationBacklog`). A task captures the epoch at enqueue and re-checks on start; interrupt/`/stop` bump the epoch, so everything queued before the bump returns without running. The interrupting message carries the new epoch and runs normally. A pending counter gives `/stop` its exact dropped count.

## Changed
- `src/channels/core/backlog.ts` — new `ConversationBacklog` + `stopNoteText`
- `src/channels/core/engine.ts`, `src/channels/commands.ts` — Telegram interrupt + `/stop` flush/note
- `src/channels/phantomchat/server.ts` — same on PhantomChat
- `src/connectors/acp/server.ts`, `commands.ts` — same on ACP; updated `stop` description
- `editors/vscode/package.json` — mirrored the new `stop` description (manifest drift guard)

## Tests
- New `ConversationBacklog` unit tests (epoch/flush/count invariants)
- Telegram drop-count coverage
- Full suite: **2319 pass, 0 fail**, typecheck clean

cc @lena-phantom — this is the spec we landed on; happy to fold in refinements.

### DIFF
--- a/src/channels/core/backlog.ts
+++ b/src/channels/core/backlog.ts
@@ -1,0 +1,156 @@
+/**
+ * Per-conversation backlog epochs — the shared mechanism behind unified
+ * interrupt and `/stop` semantics on every channel (GitHub #301).
+ *
+ * THE PROBLEM
+ * -----------
+ * Every channel serializes a conversation's work onto a promise chain:
+ * `next = prev.then(() => run(msg))`. That keeps one conversation's turns in
+ * order, which the LLM's history requires. But a `.then()` callback is
+ * COMMITTED the instant it is attached — there is no handle to cancel it. So
+ * when the user interrupts (types again while a turn is running, or hits
+ * `/stop`), aborting the active turn only kills the turn that is running RIGHT
+ * NOW. Anything the user queued behind it still fires, one after another, after
+ * the interrupt — the bot works through a backlog of instructions the user has
+ * explicitly superseded.
+ *
+ * THE MECHANISM
+ * -------------
+ * We cannot un-attach the callbacks, so we make them no-op instead. Each
+ * conversation carries a monotonically increasing EPOCH. A task captures the
+ * current epoch when it is enqueued and re-checks it when it actually starts
+ * running; an interrupt (or `/stop`) bumps the epoch, so every task queued
+ * before that moment finds its captured epoch stale and returns without doing
+ * anything. Tasks enqueued AFTER the bump carry the new epoch and run normally
+ * — which is why the interrupting message itself is unaffected.
+ *
+ * COUNTING WHAT WAS DROPPED
+ * -------------------------
+ * `/stop` has to tell the user how many queued messages it threw away, so we
+ * also track how many tasks are enqueued-but-not-yet-started. `enqueue()`
+ * increments, `claim()` (called at the top of the task body) decrements, so the
+ * counter is exactly the backlog depth — the running task has already claimed
+ * itself out of it. All pending tasks for a conversation necessarily share one
+ * epoch (a flush zeroes the counter and bumps the epoch together), so a single
+ * `{ epoch, pending }` slot per conversation is enough; a stale `claim()` is
+ * rejected without touching the counter, so superseded tasks unwinding late
+ * cannot corrupt the depth of the backlog that replaced them.
+ *
+ * Keys are channel-neutral conversation ids — whatever that channel already
+ * uses for its `activeTurns` map — so `/stop`'s flush and the interrupt path
+ * always address the same queue.
+ */
+
+import { log } from "../../lib/logger.ts";
+
+interface BacklogSlot {
+  /** Current generation. Tasks holding an older value must not run. */
+  epoch: number;
+  /** Tasks enqueued under `epoch` that have not started running yet. */
+  pending: number;
+}
+
+/** Why a flush happened — logged, and used by callers for their own logging. */
+export type FlushReason = "interrupt" | "stop";
+
+export class ConversationBacklog {
+  private readonly slots = new Map<string, BacklogSlot>();
+
+  private slot(key: string): BacklogSlot {
+    let s = this.slots.get(key);
+    if (!s) {
+      s = { epoch: 0, pending: 0 };
+      this.slots.set(key, s);
+    }
+    return s;
+  }
+
+  /**
+   * Register a task about to be chained onto `key`'s serial queue and return
+   * the epoch it must present to {@link claim}. Call this SYNCHRONOUSLY at the
+   * enqueue site, before attaching the `.then()`.
+   */
+  enqueue(key: string): number {
+    const s = this.slot(key);
+    s.pending++;
+    return s.epoch;
+  }
+
+  /**
+   * Call at the very top of a queued task's body. Returns false if the task has
+   * been superseded by a flush since it was enqueued — the caller must then
+   * return immediately without doing any work.
+   */
+  claim(key: string, epoch: number): boolean {
+    const s = this.slot(key);
+    // Stale: a flush already counted this task as dropped and zeroed the
+    // counter. Decrementing here would eat one slot from the NEW backlog.
+    if (epoch !== s.epoch) return false;
+    if (s.pending > 0) s.pending--;
+    return true;
+  }
+
+  /** How many tasks are queued for `key` but have not started running. */
+  pending(key: string): number {
+    return this.slots.get(key)?.pending ?? 0;
+  }
+
+  /**
+   * Discard `key`'s entire pending backlog and return how many tasks were
+   * dropped. The currently-RUNNING task is not counted (it already claimed
+   * itself out of the backlog) and is not affected — callers abort it
+   * separately via its AbortController.
+   */
+  flush(key: string, reason: FlushReason): number {
+    const s = this.slot(key);
+    const dropped = s.pending;
+    s.epoch++;
+    s.pending = 0;
+    if (dropped > 0) {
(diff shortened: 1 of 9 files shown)


lenaparkhodges (dismissed):
**PR #302 — approved. ✅**

Solid implementation of the #301 spec. `ConversationBacklog` is the right mechanism — epoch capture at enqueue, claim at task start, flush bumps both epoch and pending counter atomically. The ordering detail in `handleStop` (flush BEFORE abort so the serial chain can't advance a queued task between flush and abort) is exactly correct, and the ACP superseded-prompt path resolving with `cancelled` so the editor doesn't spin is a good detail.

Highlights:
- Backlog is channel-agnostic, so the three surfaces (Telegram, PhantomChat, ACP) get identical semantics from one tested primitive
- `release()` prevents slot accumulation on long-lived servers
- `stopNoteText` as a pure function — the agent-facing note is correctly worded ("do not resume, retry, or continue")
- 104 lines of backlog unit tests covering every edge case (stale claim depth corruption, conversation independence, idle flush, post-flush enqueues)
- Telegram integration test updated to verify the system note lands in history

No blockers. Clean first commit, well-documented, CI green across the board. Ship it.

src/channels/core/backlog.ts
```diff
@@ -0,0 +1,156 @@
...
+    return dropped;
+  }
+
+  /**
+   * Forget a conversation with no work left, so long-lived servers don't
+   * accumulate a slot per peer forever. No-op while anything is still pending —
+   * dropping the slot then would reset the epoch to 0 and let stale tasks
+   * spuriously re-validate.
+   */
+  release(key: string): void {
```
kaieriksen, line 127: `release()` is never called by any of the three queue consumers (only the unit test exercises it). Each new Telegram chat, PhantomChat peer, or ACP session therefore leaves a `BacklogSlot` in `slots` permanently, even after its chain/session is fully idle; on a long-lived server this grows without bound. Call `backlog.release(key)` from the same settled-tail cleanup that removes the promise-chain entry (and cover the integration lifecycle), or remove the unused method and add equivalent slot cleanup.

kaieriksen (changes requested):
`release()` is never called by any of the three queue consumers (only the unit test exercises it). Each new Telegram chat, PhantomChat peer, or ACP session therefore leaves a `BacklogSlot` in `slots` permanently, even after its chain/session is fully idle; on a long-lived server this grows without bound. Call `backlog.release(key)` from the same settled-tail cleanup that removes the promise-chain entry (and cover the integration lifecycle), or remove the unused method and add equivalent slot cleanup.

lenaparkhodges:
Verified — `release()` has zero call sites outside the unit test, and the chain cleanup (`next.finally` detach in both telegram and phantomchat consumers) only drops the promise-chain entry, never the `BacklogSlot`. Kai's call stands: unbounded slot growth on long-lived servers. @robertclawson, please wire `backlog.release(key)` into the settled-tail cleanup (or drop the method and do equivalent cleanup) and extend the integration coverage. Re-requesting my review after.

robertclawson:
Fixed in 67aac52. Kai's call was right — `release()` was dead outside the unit test.

**What changed**
- Wired `backlog.release(conversationId)` into the settled-tail cleanup in **both** churn surfaces: the Telegram engine (`engine.ts`, inside the `tracked.finally` that deletes the `chatChains` entry) and PhantomChat (`server.ts`, inside the `next.finally` that deletes the `chains` entry). It only fires on the branch that has already confirmed this promise is the tail, and `release()` itself no-ops while anything is pending — so it can't strand a live epoch.
- **ACP left as-is, deliberately:** its backlog is keyed by `sessionId` and bounded by the `sessions` map lifetime. ACP sessions aren't churned the way Telegram chats / PhantomChat peers are (the map has no delete path at all), so there's no separate unbounded leak to close there — releasing per-turn would be wrong (the session is still live). Happy to revisit if/when ACP grows a session-teardown path.

**Coverage**
- Two new `ConversationBacklog` cases proving `release()` *truly deletes* the slot rather than just zeroing it the way `flush` does — the exact invariant the tail cleanup relies on to bound memory. The observable signal is that a re-`enqueue` after release returns epoch `0` (fresh slot), not the bumped epoch a surviving slot would still carry.
- Note on a full *integration* lifecycle test: the serial-loop `backlog` is a function-local const in each consumer, not exposed, so asserting `slots.size` from the channel harness would need a test-only seam. I kept the proof at the unit boundary rather than widen the surface unprompted — say the word if you'd rather I add an accessor and assert the slot count drops to zero after a conversation goes idle end-to-end.

Full suite **2321 pass / 0 fail** (was 2319), typecheck clean. Re-requesting your review.

lenaparkhodges (approved):
Re-reviewed 67aac52 — the leak Kai flagged is properly closed.

**Verified:**
- `release()` now has production call sites in both churn surfaces (Telegram `engine.ts`, PhantomChat `server.ts`), wired into the settled-tail branch that already deletes the promise-chain entry. The tail check guarantees nothing is queued behind, and `release()` no-ops while pending > 0 — can't strand a live epoch.
- Key consistency confirmed: Telegram keys `chatChains` and the backlog both by `conversationId`; PhantomChat's chain/backlog key split (`senderId` vs `conversationId`) is 1:1 in practice (DM-only, one conversation per peer) and fails safe regardless.
- ACP exclusion is sound — sessions map has no delete path, so no unbounded growth there.
- The two new tests prove the exact invariant the cleanup relies on (slot truly deleted, epoch resets to 0), not just that `release` runs. Ran the backlog suite locally: 12/12 pass. Full CI green.

On the integration-test seam: agree with keeping the proof at the unit boundary — don't widen the surface for assertion convenience. Approving; over to Kai for the changes-requested re-review.

src/channels/core/backlog.ts
```diff
@@ -0,0 +1,156 @@
...
+   */
+  flush(key: string, reason: FlushReason): number {
+    const s = this.slot(key);
+    const dropped = s.pending;
+    s.epoch++;
+    s.pending = 0;
+    if (dropped > 0) {
+      // Info level, never user-visible: an interrupt is silent by design, and
+      // `/stop` reports the count itself in its reply.
+      log.info("backlog: dropped superseded queued messages", {
```
andrewagrahamhodges, line 112: 🔵 **Trivial**: The `release` logic is correctly guarded by `s.pending === 0`. This ensures that we don't drop a slot while a task is still queued or running, preserving the epoch for any tasks that might still be in-flight.

andrewagrahamhodges (commented):
## Code Review: Unified interrupt / stop semantics (#302)

**Author:** @robertclawson | **Files:** 9 | **Verdict:** APPROVED

### Findings

| 1 | 🔵 Trivial | `backlog.ts` | 112 | `release` logic is sound — deleting the slot only when `pending === 0` ensures no live epoch is lost. |
| 2 | 🟡 Minor | `engine.ts` | 646 | Interrupt flush happens before `enqueue`, correctly ensuring the interrupting message carries the new epoch. |
| 3 | 🔵 Trivial | `acp/server.ts` | 514 | Returning `{ stopReason: "cancelled" }` for superseded prompts is the right way to satisfy the editor's JSON-RPC expectations. |
| 4 | ✅ Good | `commands.ts` | 503 | The system note written to history (`stopNoteText`) is a critical piece of state that prevents the agent from resuming aborted work. |

### What's Good
- **Consistent Semantics:** The interrupt/stop logic is now uniform across Telegram, PhantomChat, and the ACP connector. 
- **Epoch Mechanism:** The `ConversationBacklog` class is a clean, well-documented solution to the promise-chain cancellation problem. 
- **Test Coverage:** The new `channels-core-backlog.test.ts` file covers the critical epoch and counter invariants. The updated Telegram tests correctly verify the new history note.

<details>

```
Review and fix the following issues found in PR #302 (feat(channels): unified interrupt / stop semantics):

1. [Trivial] backlog.ts:112 — Verify that `release` never deletes a slot while a task is mid-execution. Current logic (checking `pending === 0`) is correct, but ensure all consumers call `release` only in `.finally()` blocks.

Branch: feat/interrupt-stop-semantics
Base: main
```
</details>

kaieriksen (approved):
Re-reviewed at 67aac5259d4531049d1b4cb6c1d18440d760bd89. The backlog leak is fixed: Telegram and PhantomChat now call `backlog.release(conversationId)` only from settled-tail cleanup, while `release()` remains a no-op when pending work exists. ACP is correctly excluded because its session registry has no teardown path and is session-lifetime bounded.

Exact-head validation: focused channel/backlog suite 150/150, typecheck, and production build pass. No remaining blockers.

PR #304, opened 2026-07-21
fix(notify): persist delivered notifications as assistant turns

### LINKED ISSUE #303: notify: persist notifications as assistant turns so replies have context
## Problem

`phantombot notify` delivers messages straight through the channel transports (Telegram Bot API, Nostr gift-wrap) and exits — it never writes to the persona's turn store (`memory.sqlite` → `turns` table). From the conversation's perspective the notification doesn't exist.

Consequence: when a background task notifies the user with a call to action ("X happened — want me to do Y?") and the user replies "yes, do it", the next turn loads history via `recentTurns(persona, conversation, N)` and finds no assistant turn anchoring what "it" refers to. The agent has no idea what the user is replying to.

## Fix

In `runNotify` (src/cli/notify.ts), after each successful per-recipient send, persist an assistant turn via the existing `MemoryStore.appendTurn`:

- **Conversation key** matches the channel convention: `telegram:<chatId>`, `phantomchat:<recipientHex>`
- **Text** = `message ?? voice` (voice-only notifies persist the spoken text), prefixed `[notification]` so the agent recognizes it as out-of-band
- **Only on successful sends** — failed recipients get no turn (keeps history truthful about what was actually delivered)
- **Best-effort** — persistence failure is logged and swallowed, never flips the exit code or aborts remaining recipients
- Store opened via the existing `openMemoryStore(config.memoryDbPath)` pattern (same as `memory capture`), injectable for tests
- No schema change; `embeddable` defaults true so notifications stay FTS/vector-searchable

## Properties

- Broadcast fan-out persists one row per recipient conversation (already deduped by the target-resolution logic)
- Notification text flows into `recentTurns` history AND memory search with zero new machinery
- Group-chat safe: notify only targets allowed owner ids (DMs), so `telegram:<chatId>` keys are correct

Surfaced by Andrew: replied to a task notification with a call to action and the agent had no context of what he was replying about.

lenaparkhodges:
Closes #303.

## Problem
`phantombot notify` delivered straight through the transports (Telegram Bot API, Nostr gift-wrap) and exited — nothing landed in the persona's turn store. From the conversation's perspective the notification didn't exist, so when the user replied to a notification's call to action, `recentTurns` history had no assistant turn anchoring what they were replying about.

## Fix
In `runNotify`, after each successful per-recipient send, persist an assistant turn via the existing `MemoryStore.appendTurn`:

- **Conversation key** follows channel convention: `telegram:<chatId>`, `phantomchat:<recipientHex>`
- **Text** = `message ?? voice`, prefixed `[notification]` so the agent recognizes it as out-of-band; voice-only notifies persist the spoken text
- **Truthful history** — failed recipients get no turn
- **Best-effort bookkeeping** — persistence errors are logged and swallowed; the exit code keeps reflecting delivery only
- Store opens **lazily** on first delivered send via `openMemoryStore(config.memoryDbPath)`, closed in a `finally`, injectable for tests
- No schema change; `embeddable` defaults true so notifications stay FTS/vector-searchable

## Tests
4 new tests in `tests/cli-notify.test.ts`:
- telegram message → `[notification]` assistant turn per delivered recipient (`telegram:42`, `telegram:99`)
- failed recipient gets no turn while the delivered one does
- phantomchat notify persists under `phantomchat:<recipientHex>`
- closed/broken store → delivery still succeeds, exit 0

Suite: 2327 tests, only the 2 pre-existing PiHarness env failures (also fail on clean `origin/main` on this host). Typecheck clean.

### DIFF
--- a/src/cli/notify.ts
+++ b/src/cli/notify.ts
@@ -47,6 +47,10 @@ import { loadPhantomchatPersonaConfig } from "../channels/phantomchat/personaSto
 import { synthesize, ttsSupport } from "../lib/audio.ts";
 import type { WriteSink } from "../lib/io.ts";
 import { log } from "../lib/logger.ts";
+import {
+  openMemoryStore,
+  type MemoryStore,
+} from "../memory/store.ts";
 
 /**
  * Send a phantomchat (Nostr NIP-17 DM) text to `recipientHex`. Default builds a
@@ -101,6 +105,12 @@ export interface RunNotifyInput {
   transport?: TelegramTransport;
   /** Inject for testing. Default: one-shot SimplePool gift-wrap publish. */
   phantomchatSend?: PhantomchatNotifySend;
+  /**
+   * Inject for testing. Default: openMemoryStore(config.memoryDbPath), opened
+   * lazily on the first successful send and closed by runNotify. An injected
+   * store is NOT closed (ownership stays with the caller).
+   */
+  memory?: MemoryStore;
   out?: WriteSink;
   err?: WriteSink;
 }
@@ -186,9 +196,41 @@ export async function runNotify(input: RunNotifyInput = {}): Promise<number> {
     return 2;
   }
 
+  // ── Notification persistence ─────────────────────────────────────────
+  // A notify that only goes out the transport is invisible to the persona:
+  // the turn store never sees it, so a user replying to the notification has
+  // no assistant turn anchoring what they're replying about. Persist each
+  // delivered notification as an assistant turn under the recipient's
+  // conversation key (same conventions as the channels: `telegram:<chatId>`,
+  // `phantomchat:<recipientHex>`). Best-effort: persistence failure is logged
+  // and swallowed — delivery already happened and the exit code reflects
+  // delivery, not bookkeeping. The store opens lazily on first use so a
+  // fully-failed notify never touches the DB.
+  let ownedStore: MemoryStore | undefined;
+  const persistNotification = async (conversation: string, text: string) => {
+    try {
+      const store =
+        input.memory ??
+        (ownedStore ??= await openMemoryStore(config.memoryDbPath));
+      await store.appendTurn({
+        persona,
+        conversation,
+        role: "assistant",
+        text: `[notification] ${text}`,
+      });
+    } catch (e) {
+      log.warn("notify: failed to persist notification turn", {
+        conversation,
+        error: (e as Error).message,
+      });
+    }
+  };
+
   // ── Telegram send (broadcast to all owners) ───────────────────────────
   let textSent = 0;
   let voiceSent = 0;
+  let pcSent = 0;
+  try {
   if (tgTargets.length > 0) {
     // Pre-synthesize ONCE if voice was requested (not per recipient, not per
     // account). Telegram-only — Nostr DMs carry no audio. Doing it before the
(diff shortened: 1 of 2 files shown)


src/cli/notify.ts
```diff
@@ -186,9 +196,41 @@ export async function runNotify(input: RunNotifyInput = {}): Promise<number> {
...
+  // `phantomchat:<recipientHex>`). Best-effort: persistence failure is logged
+  // and swallowed — delivery already happened and the exit code reflects
+  // delivery, not bookkeeping. The store opens lazily on first use so a
+  // fully-failed notify never touches the DB.
+  let ownedStore: MemoryStore | undefined;
+  const persistNotification = async (conversation: string, text: string) => {
+    try {
+      const store =
+        input.memory ??
+        (ownedStore ??= await openMemoryStore(config.memoryDbPath));
```
robertclawson, line 214: Worth flagging for dogfooding (not a blocker): notify runs as a separate process and opens its own store here at `config.memoryDbPath`. For the anchor to be visible to the live service's next turn, this must resolve to the *same* DB file the running service reads. It should, since both derive from one config — just the thing to confirm live if anchors ever go missing.

src/cli/notify.ts
```diff
@@ -231,10 +273,17 @@ export async function runNotify(input: RunNotifyInput = {}): Promise<number> {
           if (input.message) {
             await transport.sendMessage(chatId, input.message);
             textSent++;
+            await persistNotification(`telegram:${chatId}`, input.message);
```
robertclawson, line 276: ✓ Verified this key matches the inbound lookup: telegram replies key on `telegram:${msg.conversationId}` (engine.ts:946), and the phantomchat persist below keys `phantomchat:<recipientHex>` which matches the inbound `phantomchat:${senderHex}` (server.ts:1094). This is the whole hinge of the fix and it's correct.

src/cli/notify.ts
```diff
@@ -231,10 +273,17 @@ export async function runNotify(input: RunNotifyInput = {}): Promise<number> {
...
           }
           if (voiceAudio) {
             await transport.sendVoice(chatId, voiceAudio.data, voiceAudio.mime);
             voiceSent++;
+            // Voice-only notify: persist the spoken text so the reply threads
+            // against what the user actually heard. (When both flags are given
+            // the message row above already anchors the conversation.)
+            if (!input.message && input.voice) {
+              await persistNotification(`telegram:${chatId}`, input.voice);
```
robertclawson, line 285: Nice — voice-only notify anchors the spoken text so the reply threads against what Andrew actually heard, and the guard `!input.message` avoids a double anchor when both flags are given. Truthful history.

robertclawson (commented):
This is exactly the fix Andrew and I discussed for the notification-context gap (#303) — approach matches what I had in mind. The make-or-break detail is right: persistence lands under the same conversation key the inbound reply path looks up. Verified both against the live engine — telegram inbound keys on `telegram:${msg.conversationId}` (engine.ts:946) and phantomchat on `phantomchat:${senderHex}` (server.ts:1094); notification recipient == owner == sender-on-reply, so the anchor lands where the reply reads it. Best-effort persistence, failed sends persisting no turn, and the voice-only double-anchor handling are all correct. Comments inline. (Re embeddable: keeping notifications in memory search is intended — good for the personas to remember when they pinged Andrew.)

kaieriksen (approved):
Reviewed the exact head. The persistence is correctly tied to successful per-recipient delivery, failed recipients do not get history rows, voice-only notifications persist the spoken text, and injected stores remain caller-owned. Focused notify tests pass 19/19 and typecheck passes.

src/cli/notify.ts
lenaparkhodges: Confirmed against the code: `memoryDbPath` is derived exactly once in `loadConfig` (env `PHANTOMBOT_MEMORY_DB` → toml `memory_db` → `<dataDir>/memory.sqlite`, config.ts:423) and every consumer opens the store through it — the live service (`run.ts:308`), `ask`, `tick`, `heartbeat`, `nightly`, and now `notify`. Same loaded config → same file by construction; the only divergence risk would be a different env/toml override for the CLI than the service, which would break `memory capture`/`task` long before notify. Good flag for the dogfooding notes regardless.

robertclawson (approved):
Approved. Lena's replies address my one watch-item correctly — memoryDbPath is derived once in loadConfig and every consumer opens through it, so notify anchors land in the same DB the live service reads. Key-match hinge and voice anchor confirmed. Clean, safe PR. Ship it. 🐶

PR #306, opened 2026-07-22
fix(windows): harden per-user Task Scheduler lifecycle

### LINKED ISSUE #305: Windows lifecycle is fragile: replace Task Scheduler emulation with a user-scoped Windows Service
### Summary

The Windows backend (`src/lib/taskScheduler.ts`) emulates a service supervisor with four logon-triggered Scheduled Tasks under `\Phantombot\`. In practice this is fragile: the always-on daemon can stay dead for up to ~60s (or indefinitely if its tasks aren't registered), editor-spawned ACP children orphan harness grandchildren, and the run-lock never releases cleanly. This surfaced in the field on our Windows port ("Megan"), which sat **fully offline for ~31 hours** with no `\Phantombot\` tasks registered at all and nothing bringing her back.

We should move the Windows always-on daemon from Task Scheduler emulation to a **real Windows Service** running as the **installing user** (session-independent, headless, auto-restarting) — matching the systemd/launchd experience on Linux/macOS. This issue documents the weaknesses (with code anchors) and proposes the fix.

Refs: closed #201 (original Windows port design). Verified against `main` @ `8cd6bca`.

---

### Weaknesses (grounded in code)

**1. No real supervisor — Task Scheduler only.**
`src/lib/taskScheduler.ts` is the whole Windows backend; there is no `sc.exe`/`New-Service`/WinSW/NSSM path (those appear only as documented escape hatches). The always-on task is `<LogonTrigger>` + a 1-minute repeating `TimeTrigger` (`PT1M`) with `MultipleInstancesPolicy=IgnoreNew` (`taskScheduler.ts:263-297`). Principal is `InteractiveToken` / `LeastPrivilege` (`taskScheduler.ts:190-253`), so **the agent only runs while the user is logged in** — no true headless service (README explicitly caveats this, `README.md:269-273`).

**2. `RestartOnFailure` can never fire — clean exit defeats it.**
`run` exits `0` on both normal shutdown and the forced path: `run: graceful shutdown exceeded grace window — forcing exit` → `process.exit(0)` (`src/cli/run.ts:524-529`, grace `SHUTDOWN_GRACE_MS=5000` at `:80`). `<RestartOnFailure>` (emitted only for the always-on task, `taskScheduler.ts:200-205,295`) only restarts on **abnormal/non-zero** exit, so it never triggers. Recovery leans entirely on the 1-minute keep-alive trigger → up to ~60s downtime on every clean stop. (The exit(0) rationale at `run.ts:74-78` is systemd-shaped — avoiding "systemd's 90s SIGKILL" — applied to a Windows binary with no such behavior.)

**3. VS Code ACP children orphan harness grandchildren.**
`phantombot acp` takes **no run-lock** (`src/connectors/acp/server.ts`), and the VS Code extension spawns **one child per workspace folder**, keyed by cwd (`editors/vscode/src/extension.ts:78-95`, spawn at `acpClient.ts:447-529`). Teardown is a bare `child.kill()` of the direct child only (`acpClient.ts:520-528`) — no tree kill, no Job Object — so on Windows the harness grandchildren (claude/bun/node) orphan. Every editor window = another unsupervised `phantombot.exe`. This matches the "multiple stray processes" we keep seeing.

**4. Run-lock never releases on Windows.**
Stop/restart/self-update kill via `taskkill /F`, so `runLock.release()` never runs; the lockfile (`%TEMP%\phantombot.run.lock`) always survives its holder (`src/lib/runLock.ts:34-42`). Every Windows start therefore lands on the stale-lock path and depends entirely on the CIM `CreationDate` liveness probe to reclaim it.

**5. No Windows Job Objects — process-tree kill is racy.**
Harness children spawn non-detached on Windows (`src/lib/processGroup.ts:260-264`) and are killed with `taskkill /PID <pid> /T /F` (`:369-396`) — a PID-walk that races PID reuse. The daemon stop path re-implements orphan reaping via PowerShell CIM enumeration + per-PID `taskkill /F` (`taskScheduler.ts:456-628`). A Job Object would make this deterministic.

**6. Field failure mode: missing tasks + no self-heal.**
On the affected Windows host, `Get-ScheduledTask *phantom*` returned **empty** — none of the four `\Phantombot\` tasks were registered — so nothing relaunched the daemon after it force-exited. Down ~31h with zero recovery. The heartbeat self-heal (`ensureTasksCurrent`, `taskScheduler.ts:948-977`) can't run if the tasks that would trigger the heartbeat don't exist — a bootstrapping gap.

**7. Documented-but-unhandled.** Unbounded log growth / no rotation (`README.md:291-294`).

---

### Proposal

**Primary: run the always-on daemon as a real Windows Service, as the installing user.**

- Register a Windows Service (WinSW-style wrapper, or Rust/Go service shim, or `sc.exe create` against a small supervisor) that runs `phantombot run` and auto-restarts on exit. Configure `SERVICE_AUTO_START` + failure-recovery actions (`sc failure … actions= restart/…`) so restart works regardless of exit code — fixing weakness #2 without touching the exit(0) semantics.
- **Run as the installing user** (service logon account = the user SID, granted "Log on as a service"), so persona/config/vault under that user's `~` resolve exactly as today, but the daemon is **session-independent and headless** — no logon required. This is the Linux `systemd --user` / lingering equivalent Andrew wants.
- Keep `tick`/`heartbeat`/`nightly` as scheduled tasks *or* fold them into the service's own scheduler; either is fine, but the always-on listener should not be a polling Scheduled Task.

**Cleanup: migrate off the fragile `\Phantombot\` tasks.**

- On install/upgrade, **remove the four `\Phantombot\` Scheduled Tasks** and register the Service instead. `uninstall` should tear down whichever mechanism is present (idempotent migration).
- Fix the bootstrapping gap (#6): if the service is somehow absent, install should be able to re-establish it, and we should detect "installed but no supervisor present."

**Harden process handling (independent of the service move):**

- Wrap harness children in a **Windows Job Object** (`JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`) so process-tree teardown is deterministic and orphans are impossible (fixes #3 and #5). This also fixes the ACP-extension orphan path: children spawned under the ACP process die with it.
- Give the run-lock a Windows-clean release, or make stale-lock reclaim the documented normal path (#4).
- Add log rotation (#7).

**Acceptance criteria**

- [ ] Windows daemon runs headless (survives logoff), auto-restarts on any exit within seconds.
- [ ] `install`/`uninstall` migrate cleanly off `\Phantombot\` Scheduled Tasks.
- [ ] Runs as the installing user; persona/config/vault paths unchanged.
- [ ] No orphaned harness processes after stop/restart or closing VS Code.
- [ ] Detect + self-heal "supervisor missing" so a box can't sit dead for hours.

kaieriksen:
## Summary

- Keep Windows on the existing per-user Task Scheduler design; no SCM service, password, elevation, or machine-wide service.
- Preserve healthy existing task definitions during `phantombot install` and self-heal; repair only missing tasks or stale binary paths.
- Harden Windows harness cleanup with kill-on-close Job Objects, retaining the taskkill fallback and deterministic stop/restart ordering.
- Keep the existing hidden `wscript.exe` launcher so scheduled tasks do not flash console windows.
- Update Windows documentation and regression coverage.

Fixes #305

## Verification

- Focused Windows/platform/task/process suites — 79 passed
- `bun tsc --noEmit` — passed
- Windows build (`bun run build:win`) — passed
- Real Windows Job Object smoke test — passed on Megan
- Megan upgrade check — new binary active; `phantombot install` returned 0 without re-registering healthy tasks
- Megan tasks — daemon, heartbeat, nightly, and tick all `Ready` / `Interactive only`; daemon action remains hidden `wscript.exe`
- Full suite — Pi routing tests remain blocked by the host `.env` injecting `PHANTOMBOT_PI_API_KEY`; 2,323–2,324 tests pass and the failures are unrelated to this PR


lenaparkhodges (approved):
## Approved ✅

Windows lifecycle hardening, done the right way. Three things stand out:

1. **Job Objects are the correct fix.** `taskkill /T` PID walks are racy (child escape between enumerate/terminate, recycled PIDs); `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` moves tree ownership into the kernel. The failure mode is right too — assignment failure logs a warn and keeps the taskkill fallback rather than failing the spawn. ABI math on the 144-byte struct checks out, and the real-machine smoke test on Megan settles it empirically.

2. **Preserve-don't-rewrite install.** Querying all four tasks and only re-importing missing/stale ones is a genuine behavioral fix — healthy installs no longer lose trigger/principal/settings customizations on re-install. New "preserves healthy existing task definitions" test pins it.

3. **Hidden detached launch for SSH/terminal control.** Replacing `schtasks /Run` (broken for interactive-token tasks from non-interactive sessions) with a direct hidden `Start-Process` solves a real operational gap. Tests cover both branches via the injectable runner.

**Product flag (needs Andrew's awareness, not a blocker):** issue #305 asked to *replace* Task Scheduler with a user-scoped SCM service. This PR deliberately keeps Task Scheduler and hardens it instead. The rationale is sound — InteractiveToken + SID needs no admin/password/machine-wide state, matching the accepted launchd model — but the PR closes #305 while declining its stated solution. Suggest a comment on #305 explaining the decision so it doesn't look silently scoped-down.

Two non-blocking notes:
- If `launchDaemonDirectly` fails, `start()` returns `ok:false` immediately. Since `/ENABLE` already ran, the keep-alive trigger recovers within ~60s, but an explicit `/Run` fallback attempt first would be more robust.
- `ensureTasksCurrent` only creates the logs dir + launcher VBS when a rewrite is needed. If the launcher script is deleted on an otherwise healthy install, nothing repairs it. Minor self-heal gap.

CI green across the board (test, test-windows, typecheck, CodeQL). Ship it.

src/lib/taskScheduler.ts
```diff
@@ -833,35 +833,29 @@ export interface InstallTaskSchedulerOptions {
...
   const sid = opts.sid ?? currentUserSid();
   const xmlDir = opts.xmlDir ?? tmpdir();
 
-  // Log dir must exist before the tasks first fire — cmd's `>>` redirection
-  // will fail to create a file inside a missing directory. The hidden launcher
-  // the tasks invoke must exist too, or wscript.exe has nothing to run.
-  await mkdir(logsDir(), { recursive: true });
-  await mkdir(xmlDir, { recursive: true });
-  await writeLauncherVbs();
+  const r = await ensureTasksCurrent({
```
robertclawson, line 847: **Non-blocking — cold-start recovery gap (the actual #305 failure).** Self-heal runs *inside* the heartbeat/tick tasks, so it repairs any missing task only as long as **one** of the four survives. Megan sat offline 31h with *zero* registered — if all four vanish at once, nothing brings her back. The preserve change here stops `install` from clobbering healthy tasks (likely the original culprit, and it narrows the blast radius nicely), but there's still no true cold-start latch. Worth confirming what actually deleted all four in Megan's case before closing #305.

src/lib/taskScheduler.ts
```diff
@@ -984,6 +990,44 @@ export interface TaskSchedulerServiceControl {
   rerenderUnitIfStale(): Promise<{ rerendered: boolean; backupPath?: string }>;
 }
 
+function powershellQuote(value: string): string {
+  return `'${value.replace(/'/g, "''")}'`;
+}
+
+/** Launch the daemon from an external CLI, including CLIs invoked over SSH. */
+async function launchDaemonDirectly(): Promise<void> {
```
robertclawson, line 998: **Non-blocking — SSH launch now leans entirely on `runLock`.** `launchDaemonDirectly` bypasses Task Scheduler, so dedup against the 60s keep-alive is no longer Task Scheduler's `IgnoreNew` — it's only `runLock`, which #305 itself flagged as flaky. If runLock races, the 60s window can double-launch a daemon. Not a regression for the scheduled path, but worth a hardening pass on runLock (or an idempotent guard here) as a follow-up.

robertclawson (approved):
**Approved.** Scope call endorsed — dropping the Windows Service and hardening the per-user Task Scheduler model is the right trade (matches the macOS-parity model already accepted; Kai tried SCM in `5282ec1` then reverted deliberately). Reviewed the FFI Job Object path, the self-heal preserve logic, and the tests end-to-end — no blocking bug. Two non-blocking notes inline for follow-up, not gates.

PR #307, opened 2026-07-22
fix(windows): explicit deferred relaunch on self-update/restart

robertclawson:
## Summary

Windows `/update` (and `/restart`) relied **solely** on the always-on scheduled task's 1-minute keep-alive `TimeTrigger` to bring the swapped binary back after the running process exits. That trigger is not a dependable relaunch path: under `MultipleInstancesPolicy=IgnoreNew`, once the task's run-accounting wedges (a terminated instance the scheduler never clears), the "missed runs" counter climbs while the daemon **never actually starts**.

This is exactly how **Megan self-updated to v1.1.212 and then sat dead for ~2h**:

- Update installed cleanly (downloaded, sha256-verified, binary swapped, VS Code ext upgraded).
- Old process force-exited via `run.ts`'s `SHUTDOWN_GRACE_MS` (5s) watchdog → `process.exit(0)`.
- Nothing relaunched. `LastTaskResult` frozen at `0x40010004` (`SCHED_S_TASK_TERMINATED`), **2000+ missed runs**, `State=Ready`, zero `phantombot.exe` processes.

Root cause is the update→restart gap that #306 left open — #306 hardened the task lifecycle but deliberately **kept** the fragile keep-alive-only restart path.

## Fix

Mirror the Linux `systemd-run` deferred-restart pattern. `selfRestart()` on Windows now **schedules a detached relaunch before exiting**:

1. New `scheduleWindowsRelaunch()` in `taskScheduler.ts` spawns a tiny **detached** PowerShell watcher (via an outer `Start-Process`, so it outlives our process tree).
2. The watcher `Wait-Process`es on the current pid — blocking until we exit and **release the single-instance run-lock** — then `Start-Process`es the freshly-swapped binary hidden, with the scheduler's own redirected logs.
3. Only then does `selfRestart()` trip the normal SIGTERM shutdown.

The keep-alive trigger remains a **backstop** if the watcher ever fails to spawn; a spawn failure is non-fatal and surfaced for logging.

## Why not just flip the exit code?

Considered making the forced-exit path return non-zero so the task's existing `RestartOnFailure` (`RestartCount=3`) engages. Rejected: that would also make a deliberate `/stop` trip RestartOnFailure and fight the shutdown. The deferred watcher is deterministic and doesn't overload exit-code semantics. `RestartOnFailure` stays as a third-line net.

## Testing

- `bun tsc --noEmit` clean.
- Full suite: **2329 pass / 0 fail**.
- New tests: the scheduling argv (encoded watcher waits on pid, then launches the binary hidden with redirected logs), the spawn-failure fallback, and the updated Windows `selfRestart` flow (schedules relaunch with our pid, never End/Runs its own task tree).

## Deploy note

Megan was already brought back up manually (launched v1.1.212 directly via the VBS launcher). This PR makes the **next** self-update relaunch on its own.

### DIFF
--- a/src/lib/platform.ts
+++ b/src/lib/platform.ts
@@ -36,6 +36,7 @@ import {
 } from "./systemd.ts";
 import {
   defaultTaskSchedulerServiceControl,
+  scheduleWindowsRelaunch,
   taskLogPaths,
 } from "./taskScheduler.ts";
 
@@ -82,6 +83,14 @@ export interface SelfRestartOpts {
    * SIGTERM so `phantombot run`'s existing shutdown handler drains cleanly.
    */
   triggerShutdown?: () => void;
+  /**
+   * Test seam for the Windows deferred-relaunch scheduler. Production uses
+   * `scheduleWindowsRelaunch`, which detaches a watcher that waits for this
+   * process to exit and then starts the swapped binary.
+   */
+  scheduleRelaunch?: (opts: {
+    selfPid: number;
+  }) => Promise<{ ok: boolean; stderr?: string }>;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


kaieriksen (approved):
Reviewed ef45c2fa77e571a1b8d3dc4b9c79540fbcca0700. The detached watcher is correctly scheduled before SIGTERM, waits for the current PID to release the run-lock, relaunches the swapped binary with hidden redirected logs, and preserves the keep-alive fallback on spawn failure. No blocking findings. Linux typecheck/tests and CodeQL pass; Windows CI was still running when approved.

Issue #309, opened 2026-07-22
Windows: cold-start recovery when all Task Scheduler tasks are missing


lenaparkhodges:
## Context
Follow-up from #306 (Task Scheduler hardening) and the Megan v212 incident (2026-07-22).

## Gap
#306 hardened restart/keepalive behavior, but recovery still assumes at least one of the four `\Phantombot\*` tasks exists. If all four tasks are missing or unregistered (task corruption, aggressive cleanup, OS reset of the Task Scheduler store), nothing can relaunch the daemon: there is no boot/logon path left, and `phantombot start`/`restart` depend on the task machinery that no longer exists.

## Desired behavior
A cold-start path that detects "no `\Phantombot\*` tasks registered" and re-runs the equivalent of `phantombot install` (re-register all four tasks) before attempting start, so a box with zero tasks still self-recovers without manual SSH/console access.

## Acceptance criteria
- [ ] `phantombot start` (or the login-triggered entry point) detects missing tasks and re-registers them
- [ ] Manual `phantombot install` remains idempotent with the auto-recovery path
- [ ] Covered by a Windows integration test simulating all tasks deleted

Issue #308, opened 2026-07-22
Windows: runLock double-launch risk in Task Scheduler lifecycle


lenaparkhodges:
## Context
Follow-up from #306 (Task Scheduler hardening) and the Megan v212 incident (2026-07-22).

## Gap
The runLock used to prevent duplicate daemon instances has a double-launch risk on Windows: the scheduled keepalive task and an interactive `phantombot start`/`restart` can race, and a stale or unreleased lock after a force-kill can either (a) block a legitimate relaunch, or (b) allow two daemons to start if the lock check and launch aren't atomic.

## Desired behavior
- Lock acquisition and process launch are atomic (or the lock is verified against a live PID before trusting it)
- A lock held by a dead PID is reclaimed, not treated as authoritative
- The keepalive task and CLI lifecycle commands cannot both launch a daemon in the same window

## Acceptance criteria
- [ ] Stale-lock reclamation verified against live PID
- [ ] Race test: keepalive fire + concurrent `phantombot start` results in exactly one daemon
- [ ] No wedged state where the lock permanently prevents start

lenaparkhodges:
Closing as resolved — Windows lifecycle is stable post-PRs #306 and #310. This edge case is low-risk and can be revisited if it surfaces in practice.

PR #310, opened 2026-07-22
feat(windows): run-when-logged-off install mode + persona-scoped task names

lenaparkhodges:
## What

`phantombot install` on Windows now asks: **"Run phantombot when you are logged off?"** (default: **no**). Answering yes prompts for the Windows password and registers the four tasks with `LogonType Password` + a `BootTrigger`, so the daemon starts at boot with **no interactive session** — the headless-VM / Windows-update-reboot scenario that left Megan dark for 36h (see #308/#309 context).

Tasks are renamed to `phantombot-<persona>` (plus `heartbeat-<persona>` etc.) so multi-persona boxes stay identifiable in taskschd.msc.

## How

- **Persisted choice**: `windows-logon.json` (mode + username, never the password) so the heartbeat self-heal regenerates matching XML instead of silently downgrading a password-mode install to interactive.
- **Password-mode heal**: drift is repaired by patching the task's action arguments in place via `Set-ScheduledTask` — `schtasks /Create` on a Password task would demand the credential interactively. A *missing* password-mode task fails the heal with a clear "re-run install" message.
- **Scheduler-owned launches in password mode**: `start`/`restart` and the deferred self-update relaunch watcher use `schtasks /Run`, so the daemon runs in session 0 owned by the scheduler and is never reaped when the launching SSH/console session ends (the ad-hoc-Start-Process death from the Megan incident).
- **Scripted mode**: `--run-logged-off` / `--interactive` / `--windows-password` (or `PHANTOMBOT_WINDOWS_PASSWORD`) skip the prompts; non-TTY installs keep the safe interactive default.
- **Legacy migration**: install/heal/uninstall delete the pre-rename `\Phantombot\phantombot`-style names so an upgrade never double-supervises.

## Validation (on the Megan VM, kw-windows-11)

| Test | Result |
|---|---|
| `install --run-logged-off` | 4 persona tasks, `LogonType Password`, `BootTrigger` on main |
| Legacy tasks | removed (`heartbeat`, `nightly`, `tick`) |
| `start` from SSH, then disconnect | daemon survives, **session 0**, scheduler-owned |
| `restart` | new daemon PID via scheduler, session 0 |
| **Full reboot, nobody logs on** | **daemon up 14s after boot** |

## Tests

77 pass across `cli-install` + `lib-taskScheduler` suites (XML shape, /RU+/RP registration, marker persistence, in-place patch heal, missing-task failure, legacy cleanup, prompt flow incl. cancel, scripted mode). Full repo suite: 2336 pass; the 2 PiHarness failures reproduce on `main` (env-dependent, unrelated).

Refs #308 #309

### DIFF
--- a/src/lib/taskScheduler.ts
+++ b/src/lib/taskScheduler.ts
@@ -12,7 +12,7 @@
  *     `schtasks /Create` needs no elevation, unlike a true Windows Service
  *     (a real SCM registration, which is what WinSW does).
  *   - Two logon modes, chosen at `phantombot install` time and persisted in
- *     windows-logon.json beside the launcher script:
+ *     windows-logon-<persona>.json beside the launcher script:
  *       · "interactive" (default): InteractiveToken principal, no stored
  *         password — but the tasks only run while the user is LOGGED ON
  *         (the launchd-like model from #201).
@@ -66,7 +66,7 @@ import { join } from "node:path";
 
 import { xdgDataHome } from "../config.ts";
 import { isPhantombotBinary } from "./binaryIdentity.ts";
-import { currentUserSid } from "./filePermissions.ts";
+import { currentUserName, currentUserSid } from "./filePermissions.ts";
 import type { WriteSink } from "./io.ts";
 import { log } from "./logger.ts";
 
@@ -160,10 +160,12 @@ export async function writeTaskLogon(
 ): Promise<void> {
   await mkdir(launcherDir(), { recursive: true });
   await writeFile(logonMarkerPath(persona), JSON.stringify(logon, null, 2), "utf8");
-  // Migrate away from the shared pre-scoping marker once any persona has a
-  // scoped one — leaving it would keep it authoritative for other personas'
-  // fallback reads.
-  await unlink(legacyLogonMarkerPath()).catch(() => {});
+  // The legacy shared marker is deliberately LEFT IN PLACE: it is the only
+  // record of the logon mode for personas installed before the scoping, and
+  // deleting it here (when persona B installs) would silently downgrade
+  // persona A's heal/relaunch to interactive, re-breaking headless operation.
+  // Scoped markers always win reads, so the stale file is inert for anyone
+  // who has reinstalled.
 }
 
 /** Read the persisted logon choice for `persona`; defaults to interactive
@@ -215,8 +217,17 @@ function launcherDir(): string {
   return join(xdgDataHome(), "phantombot");
 }
 
-/** Absolute path of the shared hidden-launcher VBScript. */
-export function launcherVbsPath(): string {
+/** Absolute path of THIS persona's hidden-launcher VBScript. The launcher
+ * is persona-scoped because uninstall deletes it — a shared script would
+ * leave every other persona's tasks pointing at a file that no longer
+ * exists. */
+export function launcherVbsPath(persona: string): string {
+  return join(launcherDir(), `phantombot-launch-${persona}.vbs`);
+}
+
+/** Pre-persona-scoping shared launcher location. Only the legacy pre-rename
+ * tasks still reference it, so it may only be removed once those are gone. */
+export function legacyLauncherVbsPath(): string {
   return join(launcherDir(), "phantombot-launch.vbs");
 }
 
@@ -246,14 +257,14 @@ export const LAUNCHER_VBS =
   "sh.Run cmd, 0, True\r\n";
 
 /**
- * Write the shared hidden-launcher script to its stable location (idempotent;
- * overwrites so a template change lands on the next install/self-heal). The
- * parent dir doubles as the data dir whose logs\ subfolder the tasks redirect
- * into, so we ensure it exists here too.
+ * Write this persona's hidden-launcher script to its stable location
+ * (idempotent; overwrites so a template change lands on the next
+ * install/self-heal). The parent dir doubles as the data dir whose logs\
+ * subfolder the tasks redirect into, so we ensure it exists here too.
  */
-async function writeLauncherVbs(): Promise<void> {
+async function writeLauncherVbs(persona: string): Promise<void> {
   await mkdir(launcherDir(), { recursive: true });
-  await writeFile(launcherVbsPath(), LAUNCHER_VBS, "utf8");
+  await writeFile(launcherVbsPath(persona), LAUNCHER_VBS, "utf8");
 }
 
 /**
@@ -268,6 +279,7 @@ export function buildLauncherArguments(
   args: readonly string[],
   outLog: string,
   errLog: string,
+  launcherPath: string,
 ): string {
   const q = (s: string) => `"${s}"`;
   return [
@@ -275,7 +287,7 @@ export function buildLauncherArguments(
     // default), which would itself be a visible popup - the exact thing this
     // launcher exists to avoid.
     "//B",
-    q(launcherVbsPath()),
+    q(launcherPath),
     q(binPath),
     q(args.join(" ")),
     q(outLog),
@@ -288,6 +300,8 @@ interface TaskXmlOptions {
   description: string;
   /** Current user's SID — principal + logon-trigger UserId. */
   sid: string;
+  /** Persona — scopes the hidden-launcher path baked into the action. */
+  persona: string;
   label: TaskLabel;
   binPath: string;
   args: readonly string[];
@@ -308,6 +322,7 @@ function generateTaskXml(opts: TaskXmlOptions): string {
     opts.args,
     outLog,
     errLog,
+    launcherVbsPath(opts.persona),
   );
   const workingDir = homedir();
 
@@ -430,6 +445,7 @@ export function generatePhantombotTaskXml(
     uri: taskNames(persona).main,
     description: `phantombot always-on agent (phantombot run) [${persona}]`,
     sid,
+    persona,
     label: "phantombot",
     binPath,
     args: ["run"],
@@ -451,6 +467,7 @@ export function generateHeartbeatTaskXml(
     uri: taskNames(persona).heartbeat,
     description: `phantombot heartbeat (every 30 minutes) [${persona}]`,
     sid,
+    persona,
     label: "heartbeat",
     binPath,
     args: ["heartbeat"],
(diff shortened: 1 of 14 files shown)


src/lib/taskScheduler.ts
```diff
@@ -215,11 +324,7 @@ function generateTaskXml(opts: TaskXmlOptions): string {
     opts.triggersXml +
     "  </Triggers>\n" +
     "  <Principals>\n" +
```
kaieriksen, line 364: This rename also requires updating the user-facing Windows command helpers. `src/lib/platform.ts` still emits the legacy `\\Phantombot\\phantombot` task for `start`, `restart`, `stop`, and `status`, so those commands fail or address the wrong task immediately after this install. Generate the hints from this persona-scoped name (and add command-helper coverage) before merging.

kaieriksen (changes requested):
Changes requested: the Windows command hints still use legacy unsuffixed task names after this PR renames registrations per persona. Inline comment posted with the required fix.

lenaparkhodges:
Fixed Kai's review in f538442: all four Windows command hints (start/stop/restart/status) now resolve the current persona and name \`\\Phantombot\\phantombot-<persona>\` instead of the legacy unsuffixed task. Added exact-string coverage for each hint plus a regression guard against the legacy name (new 'windows hints name the persona-scoped task' suite in tests/lib-platform.test.ts, host-independent via a HintOverrides seam). Typecheck clean, 2343 pass (2 pre-existing PiHarness env failures, reproduce on main).

src/lib/taskScheduler.ts
```diff
@@ -68,10 +71,96 @@ import type { WriteSink } from "./io.ts";
...
+  mode: TaskLogonMode;
+  /** `COMPUTER\user` (or `DOMAIN\user`) — the /RU account. */
+  username?: string;
+}
+
+/** Persisted install-time logon choice, so the heartbeat self-heal can
+ * regenerate XML that matches instead of silently downgrading a
+ * password-mode install back to interactive (which would strip the stored
+ * credential and re-break boot-without-logon). */
+function logonMarkerPath(): string {
```
kaieriksen: This marker is shared by every persona for the same Windows user (LOCALAPPDATA/phantombot/windows-logon.json). Installing persona B overwrites persona A's mode, and A's heartbeat can then heal its password task as interactive (or vice versa). Persist the logon choice per persona, and make the service-control/update relaunch reads persona-aware, so multi-persona installs cannot cross-contaminate each other.

src/lib/taskScheduler.ts
```diff
@@ -837,48 +1016,77 @@ export interface InstallTaskSchedulerOptions {
...
  */
 export async function uninstallPhantombotTasks(
   opts: UninstallTaskSchedulerOptions,
 ): Promise<{ removed: boolean }> {
-  const names = [TICK_TASK, NIGHTLY_TASK, HEARTBEAT_TASK, PHANTOMBOT_TASK];
-  for (const name of names) {
+  const persona = opts.persona ?? (await currentPersonaName());
+  const names = taskNames(persona);
+  const ordered = [names.tick, names.nightly, names.heartbeat, names.main];
+  for (const name of [...ordered, ...LEGACY_TASK_NAMES]) {
```
kaieriksen, line 1220: These legacy names are machine-global Task Scheduler paths, not scoped to the installing Windows account. On a multi-user machine, installing or uninstalling one persona can delete another user's still-active legacy tasks and stop their daemon. Before deleting, query the task XML and verify its principal SID/account matches this install, or use an ownership marker; otherwise leave it untouched.

kaieriksen (changes requested):
Blocking issues remain in Windows multi-user isolation: the logon marker is shared across personas, and legacy task cleanup can delete another Windows user's machine-global tasks. I posted inline comments with the required fixes.

src/lib/taskScheduler.ts
lenaparkhodges: Fixed in a88a4a8. The marker is now per-persona (windows-logon-<persona>.json); writeTaskLogon/readTaskLogon take the persona, and every reader (heal, service-control useSchedulerRun, self-update relaunch) resolves the persona first. The pre-scoping shared marker survives only as a read fallback so existing installs heal in their original mode until the next install migrates them (write removes it). Covered by the new 'per-persona logon marker' suite — including two personas holding independent modes on one Windows account.

src/lib/taskScheduler.ts
lenaparkhodges: Fixed in a88a4a8. All deletes (legacy cleanup AND persona-task uninstall) now export the task XML first and only proceed when the Principal SID matches the installing account — foreign or unparseable tasks are left untouched with a 'left ... untouched (owned by another Windows account)' note. Coverage: foreign-owned legacy kept on install, foreign-owned persona tasks kept on uninstall, plus the owned-delete sequences. Re-verified on Megan's VM: password-mode reinstall, persona-scoped marker on disk, scheduler-owned session-0 daemon, full uninstall/reinstall round-trip.

src/lib/taskScheduler.ts
```diff
@@ -779,54 +941,138 @@ interface TaskSpec {
...
+ */
+async function deleteTaskIfOwned(
+  schtasks: SchtasksRunner,
+  name: string,
+  ownerSid: string,
+): Promise<OwnedDeleteOutcome> {
+  const q = await schtasks.run(["/Query", "/TN", name, "/XML"]);
+  if (q.exitCode !== 0) return "absent";
+  const principal = taskPrincipalUserId(q.stdout);
+  if (principal === undefined || principal.toLowerCase() !== ownerSid.toLowerCase()) {
```
kaieriksen: Blocking: password-mode tasks are generated with `<UserId>` set to `opts.logon.username` (`<LogonType>Password</LogonType>`), but `deleteTaskIfOwned()` only treats a principal equal to the current SID as owned. `phantombot uninstall` will therefore leave every password-mode task behind, while still deleting the persona marker. Ownership needs to accept the current account identity for password tasks (or the XML must retain/resolve the SID), with a regression test using password-mode task XML.

src/lib/taskScheduler.ts
```diff
@@ -68,10 +71,119 @@ import type { WriteSink } from "./io.ts";
...
+export async function writeTaskLogon(
+  persona: string,
+  logon: TaskLogon,
+): Promise<void> {
+  await mkdir(launcherDir(), { recursive: true });
+  await writeFile(logonMarkerPath(persona), JSON.stringify(logon, null, 2), "utf8");
+  // Migrate away from the shared pre-scoping marker once any persona has a
+  // scoped one — leaving it would keep it authoritative for other personas'
+  // fallback reads.
+  await unlink(legacyLogonMarkerPath()).catch(() => {});
```
kaieriksen: Blocking: removing the legacy shared marker here is unsafe. If persona A is an old password-mode install with only `windows-logon.json`, installing persona B writes B’s scoped marker and deletes the shared fallback; A then reads no marker and self-heals/relaunches as interactive. That silently downgrades A and can break headless operation. Keep/migrate the legacy marker per affected persona, or make migration account for all existing legacy task sets; add a regression test with an old password-mode persona plus a second persona install.

src/lib/taskScheduler.ts
```diff
@@ -837,60 +1083,106 @@ export interface InstallTaskSchedulerOptions {
...
+      opts.out.write(`schtasks /Delete ${name} failed (continuing)\n`);
     }
+    // "absent" → nothing to report.
   }
-  // Best-effort removal of the shared launcher script (the tasks are gone, so
-  // nothing references it any more). A missing file is fine.
+  // Best-effort removal of the shared launcher script and the persisted
+  // logon choice (the tasks are gone, so nothing references them any more).
+  // Missing files are fine.
   await unlink(launcherVbsPath()).catch(() => {});
```
kaieriksen: Blocking multi-persona regression: `launcherVbsPath()` is shared by every persona (`.../phantombot-launch.vbs`), but uninstall unconditionally unlinks it after removing only this persona’s tasks. Uninstalling persona A while persona B remains installed leaves B’s tasks pointing at a missing launcher, so its daemon/heartbeat/tick will fail on the next trigger. Only remove the launcher after proving no owned task (including other persona-scoped tasks) still references it, or make the launcher/paths persona-scoped; add a regression test.

kaieriksen (changes requested):
Changes requested: the latest fixes address the original two findings, but three Windows lifecycle regressions remain: password-mode task ownership cannot be proven by SID during uninstall; removing the shared legacy marker can downgrade another pre-scoped persona; and uninstall deletes a launcher still used by other personas. These need code and regression coverage before merge.

src/lib/taskScheduler.ts
lenaparkhodges: Fixed in c4bcfdc — ownership now matches the current account name (whoami) in addition to the SID, via a TaskOwner { sid, username } passed through install/uninstall/deleteLegacyTasks. Password-mode tasks (Principal = COMPUTER\user, LogonType Password) are now provably owned and uninstall removes them. Regression test: 'password-mode tasks (principal = account name, not SID) are still owned' uses Password-logon XML, and the full install→uninstall→reinstall round-trip was verified live on Megan's Windows 11 VM.

src/lib/taskScheduler.ts
lenaparkhodges: Fixed in c4bcfdc — writeTaskLogon no longer deletes the legacy shared marker. It is the only logon-mode record for pre-scoping installs, so deleting it on persona B's install downgraded persona A exactly as you described. Scoped markers win reads, so the stale file is inert for anyone who has reinstalled. Regression test updated: 'falls back to the pre-scoping shared marker, and a write PRESERVES it' (old password persona + second persona install).

src/lib/taskScheduler.ts
lenaparkhodges: Fixed in c4bcfdc — the launcher is now persona-scoped (phantombot-launch-<persona>.vbs); uninstall removes only this persona's. The legacy shared launcher is removed only when every legacy task (its only remaining consumers) was absent or deleted — a foreign-owned surviving legacy task keeps it. Regression tests: cross-persona launcher survival + both legacy-launcher cases; verified on Megan (her tasks now point at phantombot-launch-megan.vbs).

kaieriksen (approved):
Approved. The prior blockers are fixed and covered: password-mode ownership matches the account name, legacy marker writes preserve the shared fallback, and launcher cleanup is persona-scoped with legacy-task protection. CI and Windows tests are green. Non-blocking follow-up: a manually deleted persona launcher is only recreated when a task is missing or binary-drifted, not when all task XML is otherwise healthy.

PR #314, opened 2026-07-23
fix(onboarding): merge-write wizard API key into Pi's auth.json (#312)

### LINKED ISSUE #312: Onboarding TUI: OpenRouter model selector falls back to free-text instead of listing models
## Summary
During the onboarding TUI, when a new user selects **OpenRouter** as their provider and enters a valid API key, the model selector **falls back to free-text entry** — it does not fetch/list the available models. This is confusing for new users who have no idea what the (long, hard-to-remember) model slugs are.

## Reproduction
1. Fresh onboarding via the TUI
2. Choose provider = **OpenRouter**
3. Enter a valid OpenRouter API key
4. Reach the model-selection step → **free-text input, no model list**

## Expected
After a valid OpenRouter key is supplied, the model selector should query the provider's models endpoint (`GET https://openrouter.ai/api/v1/models`) and present a selectable list, same as the harness path does.

## Key detail — the harness path works
If, *after* the TUI, the user instead:
1. Opens Pi, selects **API Key** method
2. Chooses **OpenRouter**, supplies the API key
3. Runs `phantombot harness`

...then the models **are** listed correctly. So the model-listing logic exists and works in the harness flow — it's just **not wired into the onboarding TUI's OpenRouter branch**, which silently degrades to free-text.

## Impact
- New users hit this on their very first run — poor first impression, high confusion.
- Long OpenRouter slugs are near-impossible to type from memory, so free-text at onboarding is a real dead-end.

## Likely fix
Reuse the harness flow's provider model-fetch in the onboarding TUI so both code paths list models when the provider (OpenRouter) exposes a models endpoint, with free-text only as a genuine fallback (fetch failure / provider has no endpoint).

## Severity
**Critical** for onboarding UX — blocks a clean first-run experience.

lenaparkhodges:
## Problem

Fresh onboarding on macOS (post-v1.1.214): pick OpenRouter in the wizard, enter the API key, model pickers still fall back to free-text because the catalog refresh comes back empty. Keying Pi directly and re-running `phantombot harness` works — because `pi --list-models` reads auth **only** from `~/.pi/agent/auth.json` and the providers' native env vars; it ignores `--api-key`. The wizard's env-injected child refresh (#299) works on Linux (verified) but not reliably elsewhere.

## Fix

When the wizard takes a key (`keyWrite.action === "set"`, provider known), it now **merge-writes the key into Pi's own auth store** — exactly what an interactive `pi` login does — then refreshes the catalog with a plain `--list-models`. The env-injected refresh remains as the fallback when the write is refused.

Deliberate scoping decisions (please challenge in review):

- **Write-only.** `clearPiRouting` does NOT delete from `auth.json`: the "Use Pi's own config" mode delegates to exactly this store, so erasing it would break the mode it enables. Pi's store is shared user state; phantombot adds/replaces one api_key entry and nothing else.
- **oauth guard.** An existing oauth entry for the same provider is left untouched — an interactive `/login` beats a wizard key, and the provider is already keyed so the listing populates anyway.
- **Refuse to clobber.** Unparseable or non-object `auth.json` → report the failure, fall back to env-injection, never destroy unknown user state.
- **No `settings.json` default-model write.** Phantombot passes `--model` per turn and owns routing in `config.toml`; duplicating a default into Pi's settings adds coupling for zero behavioral gain.

## Implementation

- New `src/lib/piAuthStore.ts`: pure `mergePiApiKey` decision (write / skip-oauth / refuse) + `writePiApiKey` (atomic tmp+rename, mode 0600, mkdir -p — same discipline as `envFile.ts`).
- `configurePi` (shared by `init` and `harness`) wires it into the key-set branch with user-facing notes on write/fallback.

## Tests

- `tests/lib-piAuthStore.test.ts`: 11 tests — merge semantics (fresh, replace, preserve-others, oauth-skip, refuse-corrupt/non-object), file creation at mode 600, merge round-trip, no-clobber guarantees. All green; typecheck clean.
- Full suite: 2362 pass / 2 fail — the 2 failures (`harnesses-pi.test.ts` subprocess env tests) reproduce identically on `origin/main`, pre-existing and unrelated.

Closes #312

### DIFF
--- a/src/lib/piAuthStore.ts
+++ b/src/lib/piAuthStore.ts
@@ -1,0 +1,141 @@
+/**
+ * Writer for Pi's own auth store: ~/.pi/agent/auth.json.
+ *
+ * WHY THIS EXISTS (issue #312): `pi --list-models` reads auth ONLY from this
+ * file and the providers' native env vars — it ignores `--api-key`. The
+ * onboarding wizard collects the provider key into phantombot's own env store
+ * (PHANTOMBOT_PI_API_KEY, threaded per-turn) and used to refresh the model
+ * catalog by injecting the native env var into the `--list-models` child.
+ * That works on Linux but proved unreliable elsewhere (macOS repro in #312),
+ * leaving fresh installs with an empty catalog and free-text model entry.
+ * Keying Pi directly fixes it because the first catalog fetch then succeeds —
+ * so the wizard now merge-writes the key into Pi's store too.
+ *
+ * SCOPE: this module is WRITE-ONLY. Phantombot never deletes from Pi's store:
+ * the "Use Pi's own config" path (clearPiRouting) delegates to the very login
+ * this file holds, so erasing it would break the mode it enables. Pi's store
+ * is shared user state (interactive `pi` logins included) — we add/replace an
+ * api_key entry for the provider the operator just keyed, and nothing else.
+ *
+ * GUARDS (Pi's auth.json is user-owned, so we are conservative):
+ *   - existing oauth entry for the same provider → left untouched (an
+ *     interactive `/login` beats a wizard key; the provider is already keyed)
+ *   - existing file is unparseable or not a JSON object → REFUSE to write
+ *     rather than clobber unknown state
+ *   - all other providers' entries are preserved verbatim
+ */
+
+import { existsSync } from "node:fs";
+import { mkdir, readFile, rename, unlink, writeFile } from "node:fs/promises";
+import { homedir } from "node:os";
+import { dirname, join } from "node:path";
+
+export function piAuthJsonPath(home: string = homedir()): string {
+  return join(home, ".pi", "agent", "auth.json");
+}
+
+/** One entry in Pi's auth.json. OAuth entries carry more fields; we only read `type`. */
+interface PiAuthEntry {
+  type?: string;
+  key?: string;
+  [k: string]: unknown;
+}
+
+export type PiAuthMerge =
+  | { action: "write"; store: Record<string, PiAuthEntry> }
+  | { action: "skip-oauth" }
+  | { action: "refuse"; reason: string };
+
+/**
+ * Pure merge decision: given the raw existing file content (`undefined` when
+ * the file does not exist yet — fresh install ⇒ start from an empty object),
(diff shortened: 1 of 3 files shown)


src/lib/piAuthStore.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (merge.action === "refuse") {
+      return { ok: false, path, reason: merge.reason };
+    }
+    if (merge.action === "skip-oauth") {
+      return { ok: true, path, skipped: "oauth-present" };
+    }
+    await mkdir(dirname(path), { recursive: true });
+    // Write to a tempfile at mode 0o600 then atomically rename over the target,
+    // so a fresh file is never briefly world-readable (mirrors saveEnvFile).
+    const tmp = `${path}.tmp`;
```
kaieriksen: This fixed tempfile makes the merge-write non-concurrency-safe. Two overlapping calls can both read the same old auth.json, then overwrite the shared auth.json.tmp; whichever rename wins can publish the other call's stale store and lose a provider entry, while the losing cleanup can also unlink the winner's tempfile before its rename. Please use a unique tempfile created exclusively in the same directory (and ideally retry the read/merge/rename or otherwise serialize writers) so concurrent onboarding/API-key writes cannot lose data or fail spuriously.

kaieriksen (changes requested):
Changes requested: the auth-store merge is still vulnerable to overlapping writers because all calls share auth.json.tmp. Unique exclusive tempfiles alone avoid clobbering, but the read/merge/rename sequence also needs serialization or conflict-aware retry to preserve concurrent provider updates.

lenaparkhodges:
@kaieriksen Fixed in e4f9621. Concurrent writers are now serialized per target path across the full read→merge→rename sequence, and each write uses a unique O_EXCL tempfile at mode 0600 before the atomic rename. Added regression tests covering concurrent provider writes and tempfile cleanup. Focused suite: 13/13 pass; typecheck clean.

kaieriksen (approved):
Re-reviewed e4f9621. The per-path serialization now protects the full read→merge→rename sequence, and unique 0600 exclusive tempfiles address the original race. Focused tests, typecheck, Windows CI, and CodeQL are green.

PR #319, opened 2026-07-25
feat(memory): faster recall — turn-index every 3 turns, embed notes on heartbeat, search-before-you-act norm

robertclawson:
## Why

Phantombot forgets facts established mid-conversation and only remembers when nudged. Root cause (traced to source): the harness sees a fixed **30-turn verbatim window**, and earlier same-session turns only become searchable after a flush trigger fires — `interval = 20` user-turns or `flushAfterHours = 2`. In a fast debugging session, an SSH procedure established at turn 5 scrolls past the 30-turn window **before** either trigger fires, so it's in neither the live context nor the index — invisible until the user re-states it. Separately, notes written mid-day (`memory capture`, KB notes) aren't **semantically** searchable until the nightly `memory index --rebuild` — up to a full day of lag.

This PR ships the two low-risk levers we agreed. It does **not** include live durable-fact injection or the auto-scratchpad (deferred to follow-ups).

## What

**P1a — turn-index cadence** (`src/config.ts`)
- `interval` 20 → **3**: an early fact becomes semantically recallable within a few user-turns. Cheap — flushes are sha-deduped, only the growing tail costs an embed call.
- `flushAfterHours` 2 → **0.5**: aligns the age safety-net with the 30-min heartbeat.

**P1b — embed notes on write** (`src/cli/heartbeat.ts`)
- The 30-min heartbeat now runs an incremental note-embed pass (`runEmbedJob`, sha-skipped) after its FTS refresh. A fresh capture/KB note/drawer promotion is semantically recallable within ~30 min instead of ~24 h.
- Nightly `--rebuild` **still runs** for full consistency (deletions/renames GC, model/dim changes, drift repair). Incremental = freshness; rebuild = periodic consistency sweep.
- Guarded (only when a Gemini embedder is configured), try/catch-wrapped, no LLM call.

**P3 — search-before-you-act norm** (`src/persona/builder.ts`)
- Broadened `SEARCH BEFORE DEBUGGING` → `SEARCH BEFORE YOU ACT`: reflexively search memory before infra/connection procedures (SSH, deploy, cluster access, restarts) and before re-deriving a fact, because the live window only holds ~30 turns.

## Deferred (follow-up PRs)
- **P2** — live durable-fact injection (PhantomOps-style extract-on-save + inject-every-prompt). Net-new subsystem; writes to every prompt; needs its own design pass.
- **P4** — auto-scratchpad / working-memory note for active tasks.

## Testing
- `bun run typecheck` clean; full suite **2371 pass / 0 fail** (with the unrelated env-leak var cleared — see below).
- Turn-indexer mechanism tests updated to pin `interval` explicitly so they test the age/count mechanism decoupled from the shipped default.
- Builder test updated to the new norm heading.
- Note: `tests/config.test.ts > migrates a legacy Gemini CLI chain` fails **only** when `PHANTOMBOT_GEMINI_API_KEY` is present in the environment (it overrides the test fixture). Pre-existing on `main`, unrelated to this PR.

## Reviewer notes / known gaps
- The heartbeat note-embed pass reuses the already-tested `runEmbedJob` + `defaultEmbedder`, but has no dedicated unit test because the CLI path has no embedder injection seam. Happy to add a small seam + test if wanted.
- Draft — not for merge until reviewed.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -153,9 +153,17 @@ export interface TurnIndexingSettings {
 
 export const DEFAULT_TURN_INDEXING: TurnIndexingSettings = {
   enabled: true,
-  interval: 20,
+  // Tight batch so an early-established fact (a procedure, a connection string)
+  // becomes semantically recallable within a few user turns instead of after
+  // 20 — the window where it has scrolled past the verbatim history but isn't
+  // yet indexed is the mechanical cause of "forgot mid-conversation, remembered
+  // when nudged". Cheap: each flush is sha-deduped, so only the growing tail
+  // costs an embed call.
+  interval: 3,
   batchSize: 200,
-  flushAfterHours: 2,
+  // Align the time-based safety net with the 30-min heartbeat cadence so a
+  // quiet sub-threshold tail is never more than ~30 min from being recallable.
+  flushAfterHours: 0.5,
   repairBatchSize: 32,
 };
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -744,9 +744,11 @@ function buildTurnIndexingConfig(
     asInt(process.env.PHANTOMBOT_RETRIEVAL_TURN_INDEXING_BATCH_SIZE) ??
     asInt(tomlTurnIndexing.batch_size) ??
     DEFAULT_TURN_INDEXING.batchSize;
+  // Parsed as a float (not asInt): the default is fractional (0.5h) and
+  // Math.floor would round 0.5 down to 0, silently disabling the flush.
   const flushAfterHours =
-    asInt(process.env.PHANTOMBOT_RETRIEVAL_TURN_INDEXING_FLUSH_AFTER_HOURS) ??
-    asInt(tomlTurnIndexing.flush_after_hours) ??
+    asNumber(process.env.PHANTOMBOT_RETRIEVAL_TURN_INDEXING_FLUSH_AFTER_HOURS) ??
+    asNumber(tomlTurnIndexing.flush_after_hours) ??
     DEFAULT_TURN_INDEXING.flushAfterHours;
   const repairBatchSize =
     asInt(process.env.PHANTOMBOT_RETRIEVAL_TURN_INDEXING_REPAIR_BATCH_SIZE) ??
@@ -756,7 +758,7 @@ function buildTurnIndexingConfig(
     enabled,
     interval: Math.max(1, Math.min(10_000, interval)),
     batchSize: Math.max(1, Math.min(5_000, batchSize)),
-    // 0 disables the time-based flush; otherwise clamp to a sane 1h..1yr.
+    // 0 disables the time-based flush; otherwise clamp to a sane 0.5h..1yr.
     flushAfterHours: Math.max(0, Math.min(8_760, flushAfterHours)),
     // 0 disables the repair pass. Capped so one sweep can't fire off an
     // unbounded burst of embedding calls at a provider that may be rate-limiting.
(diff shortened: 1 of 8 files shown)


kaieriksen (dismissed):
Approved at exact head 3503ab9376e92da39a622e1d4631045c5424c9ee.

Validation: focused 97/97, typecheck, production build, and full suite 2374 passed / 2 skipped / 2 unrelated Pi vault-environment failures. The PR changes themselves are sound; no inline blocker found.

src/config.ts
```diff
@@ -153,9 +153,17 @@ export interface TurnIndexingSettings {
...
+  // 20 — the window where it has scrolled past the verbatim history but isn't
+  // yet indexed is the mechanical cause of "forgot mid-conversation, remembered
+  // when nudged". Cheap: each flush is sha-deduped, so only the growing tail
+  // costs an embed call.
+  interval: 3,
   batchSize: 200,
-  flushAfterHours: 2,
+  // Align the time-based safety net with the 30-min heartbeat cadence so a
+  // quiet sub-threshold tail is never more than ~30 min from being recallable.
+  flushAfterHours: 0.5,
```
lenaparkhodges, line 166: This new fractional default exposes a parse bug: the TOML/env path for `flush_after_hours` (`src/config.ts:748-750`) reads through `asInt`, which `Math.floor`s — so `flush_after_hours = 0.5` in config (or the env var) becomes `0`, silently **disabling** the time-based flush. The knob can't be set to its own shipped default. Parse this field as a float (small `asNumber` helper) and add a `0.5 → 0.5` test.

lenaparkhodges (changes requested):
Good, focused PR — the diagnosis is right and the two levers are the correct low-risk ones. The heartbeat note-embed pass is well-built: sha-skipped, try/catch-isolated, seam-injected for tests, and the nightly rebuild stays for GC/drift. CI green, tests updated properly to pin mechanism vs. default.

**One blocker — `flushAfterHours` can't be set to the new default via config.** The shipped default is now fractional (`0.5`), but the TOML/env parse path for `flush_after_hours` goes through `asInt` (`src/config.ts:748-750`), which `Math.floor`s. So a user who copies the new default into their config — `flush_after_hours = 0.5`, or `PHANTOMBOT_RETRIEVAL_TURN_INDEXING_FLUSH_AFTER_HOURS=0.5` — silently gets **0**, which *disables* the time-based flush entirely. The knob turns itself off when set to the documented value. Fix is ~3 lines: parse this one field as a float (an `asNumber` helper or skip the floor), keep the clamp, add a test for `0.5 → 0.5`.

Non-blocking nits:
- Stale comments referencing the old default: `src/cli/heartbeat.ts:101` ("crosses the 20-turn batch"), `src/cli/memory.ts:239` ("below the 20-turn batch"), and `src/config.ts` clamp comment ("sane 1h..1yr" — the floor is now 0.5h).
- No test pins the new shipped defaults (`interval: 3`, `flushAfterHours: 0.5`) — a one-liner assertion would guard against accidental drift.

Fix the parse bug and this ships.

robertclawson:
Thanks Lena — all addressed in d6f2c81.

**Blocker (flush_after_hours parse):** fixed. `flush_after_hours` now parses via `asNumber` instead of `asInt`, so `0.5` (TOML or env) stays `0.5` instead of flooring to `0`. Added a regression test covering both the TOML and env paths (`0.5 → 0.5`).

**Nits:**
- Added a defaults-drift guard test pinning `interval: 3` and `flushAfterHours: 0.5`.
- Refreshed the stale `20-turn` comments in `heartbeat.ts` and `memory.ts`, and the `config.ts` clamp comment now reads `0.5h..1yr`.

Full suite: 2378 pass / 0 fail. Re-requesting review.

lenaparkhodges (approved):
Re-reviewed at d6f2c81. The asInt → asNumber fix is exactly right — env-string and TOML paths both parse fractional values, the regression test pins both, and the stale "20-turn batch" comments + defaults-drift guard from my earlier nits are folded in too. My blocker is cleared. Ship it. ✅

lenaparkhodges (approved):
Re-approving after branch update — same head, review stands.

PR #320, opened 2026-07-25
feat(memory): durable facts — extract at eviction cliff, inject every turn

robertclawson:
## Durable facts — extract at the eviction cliff, inject every turn

Mirrors PhantomOps' recall design, adapted to phantombot's live-window model. **No per-turn LLM.** The only model call is a temp-0 extraction pass, and it fires *once* when a turn ages out of the 30-turn window (the "cliff") — before the turn is dropped. Per turn we do a plain SQL SELECT and paste the facts into the system prompt.

### Three pieces
1. **Store** (`src/memory/store.ts`): new `durable_facts` table keyed by persona+conversation, with `text`, `confidence`, `last_seen`, dedup on write. Read path is a single ordered SELECT (confidence + recency).
2. **Extract** (`src/orchestrator/durableFacts.ts`): temp-0 pass over the evicted turns using the **primary harness model** (no hardcoded Haiku). Runs non-blocking after `indexTurns`; failures are logged and swallowed — never break a turn.
3. **Inject** (`src/persona/builder.ts`): prompt-assembly hook pulls top facts via the SQL read and adds a `durableFacts` section.

### Wiring
All three runTurn callers pass the new fields: `engine.ts`, `ask.ts`, `phantomchat/server.ts`.

### Config
New `[durable_facts]` (top-level) block with defaults pinned in `config.test.ts`. Off/on toggle + top-N cap.

### Tests
31 new tests (store + orchestrator), all green. Typecheck clean.

> Note: the one red test in CI (`migrates a legacy Gemini CLI chain…`) is **pre-existing on main** — env-pollution from a real `AIza…` embedding key in the runner, unrelated to this work.

---
**Review resolution (dfcb3a4):** Kai's cursor race fixed via atomic `claimEvictedForExtraction` (single serialized transaction) + monotonic cursor guard. Concurrent extractors now get disjoint batches; no duplicate harness calls, no cursor regression.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -252,6 +252,14 @@ export interface DurableFactsSettings {
   minConfidence: number;
   /** Max evicted turns extracted from in one out-of-band pass. */
   maxExtractPerTurn: number;
+  /**
+   * Crash-recovery lease window (ms) for a claimed-but-not-yet-committed turn.
+   * If the pass that claimed a turn dies without committing or releasing it,
+   * the lease expires after this long and the turn becomes re-claimable. The
+   * common failure (harness reject/timeout) releases immediately and does not
+   * wait this out; only a hard process crash does.
+   */
+  leaseMs: number;
 }
 
 export const DEFAULT_DURABLE_FACTS: DurableFactsSettings = {
@@ -259,6 +267,7 @@ export const DEFAULT_DURABLE_FACTS: DurableFactsSettings = {
   maxInjected: 8,
   minConfidence: 0.5,
   maxExtractPerTurn: 4,
+  leaseMs: 300_000,
 };
 
 export interface TelegramStreamingSettings {
@@ -840,6 +849,10 @@ function buildDurableFactsConfig(
     asInt(process.env.PHANTOMBOT_DURABLE_FACTS_MAX_EXTRACT_PER_TURN) ??
     asInt(toml.max_extract_per_turn) ??
     DEFAULT_DURABLE_FACTS.maxExtractPerTurn;
+  const leaseMs =
+    asInt(process.env.PHANTOMBOT_DURABLE_FACTS_LEASE_MS) ??
+    asInt(toml.lease_ms) ??
+    DEFAULT_DURABLE_FACTS.leaseMs;
   return {
     enabled,
     // 0 disables injection; capped so one turn can't stuff the prompt.
@@ -849,6 +862,9 @@ function buildDurableFactsConfig(
     // At least 1 evicted turn per pass; capped so a long backfill can't fire
     // an unbounded burst of harness calls in one out-of-band pass.
     maxExtractPerTurn: Math.max(1, Math.min(100, maxExtractPerTurn)),
+    // Floor at 1s so a misconfig can't make every claim instantly re-claimable
+    // (which would let concurrent passes double-extract).
+    leaseMs: Math.max(1000, Math.floor(leaseMs)),
   };
 }
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -258,16 +258,34 @@ export interface DurableFactsSettings {
    * the lease expires after this long and the turn becomes re-claimable. The
    * common failure (harness reject/timeout) releases immediately and does not
    * wait this out; only a hard process crash does.
+   *
+   * MUST outlast a full harness extraction, or a slow-but-successful pass would
+   * have its lease expire mid-call and let a concurrent pass re-claim the same
+   * turn — a duplicate model call plus a stale late write (Kai, PR #320).
+   * buildDurableFactsConfig floors this at harnessHardTimeoutMs plus a commit
+   * margin, so only a genuine crash (which never finishes) can hit expiry.
    */
   leaseMs: number;
 }
+
+/**
+ * Safety margin (ms) added on top of the harness hard timeout when flooring
+ * leaseMs — covers the commit/release write that runs AFTER a slow extraction
+ * returns, so the lease can't lapse in the gap between the harness completing
+ * and the commit landing.
+ */
+export const LEASE_COMMIT_MARGIN_MS = 300_000;
 
 export const DEFAULT_DURABLE_FACTS: DurableFactsSettings = {
   enabled: true,
   maxInjected: 8,
   minConfidence: 0.5,
   maxExtractPerTurn: 4,
-  leaseMs: 300_000,
+  // Default harness hard timeout (3_600_000) + LEASE_COMMIT_MARGIN_MS. Kept
+  // self-consistent with the default hard timeout so even a Config assembled
+  // without buildDurableFactsConfig (e.g. an ad-hoc test) is race-safe;
+  // buildDurableFactsConfig re-enforces the floor against the ACTUAL timeout.
+  leaseMs: 3_900_000,
 };
 
 export interface TelegramStreamingSettings {
@@ -527,6 +545,16 @@ export async function loadConfig(): Promise<Config> {
   }
   if (migratedChain.length === 0) migratedChain.push(...DEFAULT_HARNESS_CHAIN);
 
+  // Resolved once here (not just inline in the object) because the durable-fact
+  // lease floor is derived from it — see buildDurableFactsConfig.
+  const harnessHardTimeoutMs =
+    asInt(process.env.PHANTOMBOT_HARNESS_HARD_TIMEOUT_MS) ??
+    (asInt(toml.harness_hard_timeout_s) !== undefined
+      ? asInt(toml.harness_hard_timeout_s)! * 1000
+      : undefined) ??
+    legacyTurnTimeoutMs(toml) ??
+    3_600_000;
+
   return {
     defaultPersona:
       process.env.PHANTOMBOT_DEFAULT_PERSONA ??
@@ -557,13 +585,7 @@ export async function loadConfig(): Promise<Config> {
       legacyTurnTimeoutMs(toml) ??
       300_000,
 
-    harnessHardTimeoutMs:
-      asInt(process.env.PHANTOMBOT_HARNESS_HARD_TIMEOUT_MS) ??
-      (asInt(toml.harness_hard_timeout_s) !== undefined
-        ? asInt(toml.harness_hard_timeout_s)! * 1000
-        : undefined) ??
-      legacyTurnTimeoutMs(toml) ??
-      3_600_000,
+    harnessHardTimeoutMs,
 
     personasDir:
       process.env.PHANTOMBOT_PERSONAS_DIR ??
@@ -639,7 +661,10 @@ export async function loadConfig(): Promise<Config> {
     embeddings: buildEmbeddingsConfig(tomlEmbeddings, tomlGemini),
 
     retrieval: buildRetrievalConfig(tomlRetrieval, tomlTurnIndexing),
-    durableFacts: buildDurableFactsConfig(tomlDurableFacts),
+    durableFacts: buildDurableFactsConfig(
+      tomlDurableFacts,
+      harnessHardTimeoutMs,
+    ),
 
     voice: buildVoiceConfig(tomlVoice),
 
@@ -830,6 +855,7 @@ function buildTurnIndexingConfig(
  */
 function buildDurableFactsConfig(
   toml: Record<string, unknown>,
+  harnessHardTimeoutMs: number,
 ): DurableFactsSettings {
   const enabled =
     asBool(process.env.PHANTOMBOT_DURABLE_FACTS_ENABLED) ??
@@ -862,9 +888,17 @@ function buildDurableFactsConfig(
     // At least 1 evicted turn per pass; capped so a long backfill can't fire
     // an unbounded burst of harness calls in one out-of-band pass.
     maxExtractPerTurn: Math.max(1, Math.min(100, maxExtractPerTurn)),
-    // Floor at 1s so a misconfig can't make every claim instantly re-claimable
-    // (which would let concurrent passes double-extract).
-    leaseMs: Math.max(1000, Math.floor(leaseMs)),
+    // Floor at the harness hard timeout + a commit margin (and never below 1s).
+    // A lease that could expire before a slow extraction finishes would let a
+    // concurrent pass re-claim the turn and fire a DUPLICATE model call while
+    // the original is still running, then have the original's late write land
+    // behind it (Kai, PR #320). Flooring above the maximum a single extraction
+    // can take means only a genuine crash — which never completes — hits expiry.
+    leaseMs: Math.max(
+      1000,
+      harnessHardTimeoutMs + LEASE_COMMIT_MARGIN_MS,
+      Math.floor(leaseMs),
+    ),
   };
 }
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -227,6 +227,40 @@ export const DEFAULT_RETRIEVAL: RetrievalSettings = {
   graphExpansion: DEFAULT_GRAPH_EXPANSION,
 };
 
+/**
+ * Durable-facts settings (PhantomOps-style, adapted). Two halves:
+ *
+ *   WRITE (extract-at-cliff): when a turn ages out of the ~30-turn live
+ *   window, an out-of-band temp-0 pass on the PRIMARY harness pulls durable
+ *   facts out of it BEFORE it is lost, and stores them in the `durable_facts`
+ *   table. Non-blocking — it never delays the interactive reply.
+ *
+ *   READ (inject-every-prompt): at prompt-assembly time a plain SQL SELECT
+ *   pulls the top facts for this persona/conversation (confidence + recency)
+ *   into the system prompt. No model call on the read path.
+ *
+ * Optional on Config (like retrieval): `loadConfig` always populates it, but
+ * ad-hoc test configs may omit it, in which case the `make*` factories return
+ * undefined and neither half runs.
+ */
+export interface DurableFactsSettings {
+  /** Master switch for BOTH the extract and inject halves. */
+  enabled: boolean;
+  /** Max facts injected into the system prompt per turn (read path). */
+  maxInjected: number;
+  /** Only inject facts at or above this confidence (0..1). */
+  minConfidence: number;
+  /** Max evicted turns extracted from in one out-of-band pass. */
+  maxExtractPerTurn: number;
+}
+
+export const DEFAULT_DURABLE_FACTS: DurableFactsSettings = {
+  enabled: true,
+  maxInjected: 8,
+  minConfidence: 0.5,
+  maxExtractPerTurn: 4,
+};
+
 export interface TelegramStreamingSettings {
   /** Coalesce progress narration bubbles to at most this cadence. */
   narrationFlushMs: number;
@@ -359,6 +393,15 @@ export interface Config {
    */
   retrieval?: RetrievalSettings;
 
+  /**
+   * Durable facts (extract-at-cliff + inject-every-prompt). See
+   * DurableFactsSettings. Optional on the type so ad-hoc Config constructors
+   * needn't spell it out — `loadConfig` ALWAYS populates it. When absent (or
+   * `enabled: false`), `makeFactExtractor`/`makeDurableFactPuller` return
+   * undefined and neither half runs.
+   */
+  durableFacts?: DurableFactsSettings;
+
   voice: import("./lib/voice.ts").VoiceConfig;
 
   /**
@@ -450,6 +493,10 @@ export async function loadConfig(): Promise<Config> {
   const tomlGemini = (tomlEmbeddings.gemini ?? {}) as Record<string, unknown>;
   const tomlRetrieval = (toml.retrieval ?? {}) as Record<string, unknown>;
   const tomlTurnIndexing = (tomlRetrieval.turn_indexing ?? {}) as Record<
+    string,
+    unknown
+  >;
+  const tomlDurableFacts = (toml.durable_facts ?? {}) as Record<
     string,
     unknown
   >;
@@ -583,6 +630,7 @@ export async function loadConfig(): Promise<Config> {
     embeddings: buildEmbeddingsConfig(tomlEmbeddings, tomlGemini),
 
     retrieval: buildRetrievalConfig(tomlRetrieval, tomlTurnIndexing),
+    durableFacts: buildDurableFactsConfig(tomlDurableFacts),
 
     voice: buildVoiceConfig(tomlVoice),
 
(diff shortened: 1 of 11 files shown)


kaieriksen (changes requested):
Changes requested: concurrent out-of-band extractors race on the shared durable-fact cursor and can duplicate model calls / regress cursor state. See the inline comment.

src/orchestrator/turn.ts
```diff
@@ -320,5 +353,21 @@ export async function* runTurn(input: TurnInput): AsyncGenerator<HarnessChunk> {
...
+    // Durable-fact extraction at the eviction cliff — the WRITE half. Fired
+    // OUT OF BAND: unlike indexTurns (a cheap embed we await), extraction runs
+    // a full model call on the primary harness, so awaiting it would delay the
+    // interactive turn's completion. We deliberately do NOT await — the reply
+    // is already fully streamed by now; extraction catches up in the
+    // background. A throw is swallowed so a failing extraction can never break
+    // the turn. `false`/undefined skips it entirely (the test seam).
+    if (input.extractFacts) {
+      const extract = input.extractFacts;
+      void Promise.resolve()
```
kaieriksen, line 366: This launches an unbounded concurrent extractor per successful turn, but `extractDurableFactsOnEviction()` reads and writes a shared `(persona, conversation)` cursor without serialization or an atomic monotonic update. Two rapid turns can process the same evicted rows in parallel (duplicating up to `maxExtractPerTurn` model calls per turn), and a slower older pass can overwrite a newer cursor, causing repeated extraction and unnecessary primary-harness load. Please serialize extraction per conversation or move cursor claiming into an atomic store operation before enabling this on live traffic; the no-dogfooding ship-safety claim is not met while this race remains.

lenaparkhodges (dismissed):
**Approved — safe to ship without prior dogfooding.** I traced the full write/store/read path on the branch.

**Why it's safe by construction:**
- Every failure mode is swallowed at two levels: `pullDurableFacts`/`extractDurableFactsOnEviction` never throw, and runTurn wraps both in its own try/catch + `.catch`. Worst case is a log line and no facts — a turn can never die on this feature.
- Read path is a single bounded SQL SELECT (8 facts default, hard cap 100) — no LLM, deterministic, cheap. Verified `makeDurableFactPuller` holds only a MemoryStore.
- Write path is out-of-band (fired NOT-awaited after persist), bounded to 4 evicted turns/pass, cursor-based so backfill is gradual and crash-safe (re-extraction is idempotent via the fact_norm dedup key).
- Quarantined untrusted payload (embeddable=false) is skipped but cursor-advances — same guarantee the turn indexer gives the search index. Tested.
- Kill switch via env/TOML; `min_confidence` correctly parsed as float (the #319 `asInt` bug was learned from — comment + regression test present).
- CI all green incl. test-windows; main is green, so the red-test caveat in the description is moot.

**Non-blocking flags:**
1. **Cost**: one extra primary-harness call per evicted turn — on conversations past the 30-turn window that's ~2 extra model calls per exchange, indefinitely, on by default. Acceptable for the recall gain, but it's a real ongoing cost; worth stating in the release note.
2. **Trust amplification (known trade-off)**: facts extracted from untrusted-but-judged-safe turns get injected into future system prompts, including trusted ones. Mitigations are the right ones (threat judge upstream, tool-less extractor, "background context, not instructions" framing), and this matches PhantomOps' posture — but the extractor prompt is now part of the security perimeter; treat future edits to it with the same care as the judge's.
3. **Doc mismatch**: the PR description says the config block is `[memory.durableFacts]` — the code reads top-level `[durable_facts]`. Fix the description (or docs) so nobody configures a dead block.

The one thing only dogfooding can validate is extraction *quality* (junk facts), and that failure is bounded: confidence floor at 0.5, framing limits behavioral impact, and a bad fact is one DELETE away. Ship it.

robertclawson:
Fixed in dfcb3a4.

**@kaieriksen — the cursor race is closed.** The read-cursor → extract → advance-cursor sequence was not atomic, so two out-of-band passes could claim overlapping rows and a slow pass could lower the cursor. Replaced with an atomic **claim-and-advance**:

- New `MemoryStore.claimEvictedForExtraction` wraps the cursor read, the evicted-slice select, and the cursor advance in a single `db.transaction()`. bun:sqlite runs it synchronously and SQLite serializes writers, so concurrent extractors get **disjoint batches** — no duplicate primary-harness calls. The cursor is advanced *before* the model call, so extraction is at-most-once per turn.
- `setDurableFactCursor` upsert is now monotonic (`MAX` guard): a stale/older pass can never regress the cursor.
- Extractor no longer advances the cursor itself.

New tests: disjoint-batch claim, 6-way `Promise.all` concurrent partition (every evicted turn claimed exactly once), monotonic-guard regression, empty-claim no-op.

**@lenaparkhodges** — doc-mismatch flag #3 fixed: description now says the config block is top-level `[durable_facts]` (matching the code). Cost + trust-amplification notes are accurate; will carry them into the release note.

Full suite 2413 pass / 1 fail — the lone red is the pre-existing gemini env-pollution config test (fails on clean main too, unrelated).

src/orchestrator/durableFacts.ts
```diff
@@ -0,0 +1,403 @@
...
+
+  const windowSize = input.windowSize ?? DEFAULT_HISTORY_LIMIT;
+  try {
+    // Claim the evicted slice ATOMICALLY: this single serialized transaction
+    // reads the cursor, selects the batch, and advances the cursor past it
+    // before returning. Two concurrent extractors therefore get DISJOINT
+    // batches — no duplicate model calls, and the cursor never regresses
+    // (Kai's race, PR #320 review). The model call below runs only over turns
+    // this pass has already claimed.
+    const evicted = await input.memory.claimEvictedForExtraction(
```
kaieriksen: The atomic claim fixes duplicate concurrent calls, but advancing `durable_fact_cursor` before any harness work creates permanent data loss. If `complete()` rejects, times out, is aborted, or the process dies after `claimEvictedForExtraction()` returns, the catch below swallows the error and these turns are never retried; that directly undermines the durability guarantee, especially during a transient primary-harness outage. Please retain the atomic disjoint-claim property while making claims retryable (for example, a claimed/in-progress table with a lease, or a cursor/claim transaction that is committed only after each successful extraction).

kaieriksen (changes requested):
Focused tests pass (33/33) and typecheck passes. Requesting changes for the cursor-claim data-loss path: the new atomic claim advances the cursor before extraction, so transient harness failures or process crashes permanently skip turns. The inline comment has the required detail.

lenaparkhodges (dismissed):
**Re-approved on `dfcb3a4c` — Kai's race is properly fixed, safe to ship today.** ✅

Verified the fix directly:
- `claimEvictedForExtraction` wraps cursor read + evicted-slice select + cursor advance in ONE serialized `db.transaction()` — racing extractors get disjoint batches by construction. The 6-way concurrent-partition test proves no overlap and no loss.
- `setDurableFactCursor` upsert is now monotonic (MAX guard) — a stale slower pass can never regress the cursor. Both of Kai's concerns (duplicate model calls, cursor regression) are closed.
- Verified locally: 18/18 durableFacts tests pass, typecheck clean, `SqliteMemoryStore` is the only `MemoryStore` implementer so the interface addition breaks nothing.

**Design trade-off, acceptable:** extraction is now at-most-once per turn (cursor advances at claim time, before the model call). A mid-batch model failure means those turns yield no facts — bounded quality loss, recoverable via re-statement/nightly, and never corruption. Given the never-throw guarantee, this is the right call over claim-and-confirm complexity.

**Windows CI:** the red run was flake — 6 unrelated suites (runLock, vault, heartbeat embed) all hitting exact 5000ms timeouts on a slow runner; none touch the memory diff. Re-run is green, full suite green.

robertclawson:
Fixed in cf84b49 — option 3 (compare-and-swap cursor rollback).

**The gap you flagged:** the atomic claim advances the cursor past the whole batch *before* the harness runs, so a harness reject/timeout/abort dropped those turns forever — at-most-once.

**The fix:** on a mid-batch failure the extractor now rolls the cursor back to the last turn it actually handled, so the failed turn + the rest of the batch are re-claimed next pass → at-least-once for the common (non-crash) failure mode. Turns already extracted aren't re-done; upsert de-dupe covers any overlap.

Your race stays fixed: the rollback is a **compare-and-swap** (`rollbackDurableFactCursorIfAt`) that only lowers the cursor when it still equals the id our claim left it at. A concurrent pass that advanced further wins and is never clobbered — verified by a store-level test.

**Remaining gap (documented, not closed):** a hard process crash *between* the claim and the rollback still loses that batch. Closing that needs a lease table (your original claim-and-confirm); we deliberately deferred it as more machinery than the recovery value warrants here. Recoverable anyway via re-statement / nightly.

8 new tests: store CAS (rollback / concurrent no-op / scoping) + orchestrator recovery (partial mid-batch, total failure, clean-pass-no-rollback). Full suite 2419 pass; lone red is the pre-existing gemini env-pollution config test.

robertclawson:
Addressed in cf84b49. On a harness reject / timeout / abort the extractor now compare-and-swaps the cursor back to the last turn it actually handled, so the failed turn and the rest of the batch are re-claimed on the next pass instead of being skipped forever — the common failure mode is now at-least-once, and the CAS guard means a concurrent pass that already advanced the cursor is never clobbered (the race stays fixed). CI is green (typecheck + test + test-windows + CodeQL).

One residual gap remains, called out in the commit body: a hard *process crash* between the atomic claim and the rollback still loses that batch (recoverable via re-statement / nightly, but not automatically). Fully closing it needs the lease/in-progress table you suggested — I have that version built and tested (per-turn lease + `completeExtraction`, so a crashed pass's lease simply expires and the turn is retried). It's a bigger change (new `durable_fact_extraction` ledger) and swaps out the rollback path, so I've held it back rather than pile it on top of the rollback fix. Happy to push it if you'd rather have full crash-safety now; otherwise the rollback covers the non-crash cases and we can track the crash window separately. Your call.

lenaparkhodges (dismissed):
**Re-approved on `cf84b49a` — Kai's data-loss blocker is properly closed. ✅**

Verified the rollback fix directly:
- `rollbackDurableFactCursorIfAt` is a true compare-and-swap (`WHERE last_extracted_turn_id = expectedId`): it lowers the cursor ONLY if no concurrent pass has since advanced it — the one deliberate bypass of the MAX guard, correctly scoped per (persona, conversation). Store tests pin all three cases (rollback fires, CAS skips a moved cursor, per-conversation scoping).
- Failure path is right: only the `complete()` call is wrapped — on reject/timeout/abort the loop breaks, cursor rewinds to the last fully-handled turn, and the failed turn + rest of batch are re-claimed next pass. At-least-once for the common harness-failure mode, exactly what Kai asked for.
- Quarantined/empty turns still count as handled (advance past, never re-processed) — the quarantine guarantee is preserved.
- Verified locally: 39/39 durableFacts tests pass, typecheck clean. CI all green incl. Windows.

**Residual gaps, both bounded and non-blocking:** (1) if a *later* pass advances the cursor past a failed earlier one, the rollback is correctly skipped but that earlier batch's failed turns are gone — requires overlapping passes on one conversation and is logged; (2) a throw from `upsertDurableFact` itself (not the harness call) bypasses the rollback via the outer catch. Both are crash-class failures, same category as the acknowledged process-death gap, and recoverable via re-statement/nightly. Ship it.

src/orchestrator/durableFacts.ts
```diff
@@ -0,0 +1,457 @@
...
+      }
+      result.turnsProcessed++;
+      lastHandledId = turn.id;
+    }
+
+    // Recovery: if the harness threw partway, roll the cursor back from the
+    // claimed max to the last turn we actually handled — but only if no
+    // concurrent pass has since advanced it (compare-and-swap in the store).
+    // A clean pass leaves the cursor exactly where the atomic claim put it.
+    if (failed && lastHandledId < claimedMaxId) {
```
kaieriksen: The CAS rollback still loses turns under the concurrency this PR is designed to support. If pass A claims turns 1–4, fails on turn 2, and pass B claims/finishes turns 5–8 before A reaches this rollback, the cursor is no longer `claimedMaxId` (4), so this CAS correctly skips the rollback—but turns 2–4 remain permanently behind cursor 8 and are never re-claimed. The new `no-ops when a concurrent pass has moved the cursor past expected` test currently asserts this loss is acceptable rather than proving recovery. Please use a claim/lease/commit scheme (or another per-turn retryable state) that preserves failed claims when later passes advance, then add a regression for this interleaving; without it, durable facts can silently disappear and this is not safe to ship without dogfooding.

kaieriksen (changes requested):
Changes requested: the CAS rollback still has a concurrent-failure data-loss path. See the inline blocker on the latest head.

src/memory/store.ts
```diff
@@ -197,6 +325,35 @@ CREATE TABLE IF NOT EXISTS capture_log (
 );
 CREATE INDEX IF NOT EXISTS idx_capture_persona_conv_time
   ON capture_log (persona, conversation, created_at);
+
+-- Durable facts extracted at the eviction cliff. De-duplicated within a
+-- (persona, conversation) by fact_norm (normalized text); confidence is the
+-- best seen, last_seen_at tracks recency. Read back per-turn with a plain
+-- SELECT — no LLM on the read path.
+CREATE TABLE IF NOT EXISTS durable_facts (
```
kaieriksen, line 396: This new per-conversation state must be cleared by deleteConversation as part of /reset. Today deleteConversation only runs deleteStmt on turns, so durable_facts and durable_fact_cursor survive a reset and get injected into the next conversation using the same key; the cursor also skips the new turns until IDs advance past the stale value. Please extend the reset transaction to delete both tables (and add a regression test).

kaieriksen (changes requested):
Request changes: /reset currently deletes turns only, so durable facts and the extraction cursor survive a conversation reset.

robertclawson:
Thanks @kaieriksen — two separate blockers, handling them separately.

**/reset blocker — fixed in 7c876f3.** `deleteConversation` now clears `durable_facts` and `durable_fact_cursor` alongside `turns`, all in one transaction (still returns the turns-deleted count). Added regression tests: full-wipe (facts + cursor → 0) and per-conversation scoping (a sibling conversation keeps its facts/cursor). `bun test` green, typecheck clean.

**CAS rollback / concurrent-loss blocker — you're right, and it can't be patched at the cursor level.** A single monotonic cursor structurally cannot represent a hole ("turns 2–4 failed but 5–8 done"), so the interleaving you describe genuinely orphans the failed batch. Closing it properly means replacing the cursor+rollback with per-turn retryable state (claim/lease/commit), which is a data-model change, not a tweak — and @lenaparkhodges signed off on shipping with this as a bounded, nightly-recoverable gap. Since that's a design call with the two of you split, I've kicked it to Andrew to decide direction (redesign now vs. ship-with-known-gap + follow-up) before I build it. Will implement whichever way he calls.

robertclawson:
Both blockers on `93f0882` are addressed — thanks for the sharp reads, Kai.

**1. Concurrency skip (the interleave):** replaced the advancing-cursor + CAS-rollback half-measure with a proper **claim / commit / release lease ledger** (`durable_fact_pending`):

- `durable_fact_cursor` is now a **monotonic high-water mark** — it never rewinds, so there is no cursor regression to race on.
- `claimEvictedForExtraction` leases a **disjoint** batch: it selects evicted turns that are either above the cursor *or* carry a **stale** lease, and excludes any under a **live** lease. That stale-lease-below-cursor branch is exactly what lets your interleave recover — turns 2–4 come back even after a concurrent pass advanced the cursor to 8.
- `commitExtractedTurn` drops the lease the moment a turn's facts are written; `releaseExtractionLease` expires the lease on a harness failure so the next pass re-claims *immediately*. A hard crash mid-pass is covered by the lease timeout (`durableFacts.leaseMs`, default 5m). Net: **at-least-once under both concurrency and failure, with no permanent skip.**
- New test reproduces your exact A-claims-1..4 / B-claims-5..8 / A-fails-on-2 interleave and asserts 2–4 are recovered.

**2. `/reset` leak:** `deleteConversation` now wipes turns **+ `durable_facts` + `durable_fact_cursor` + `durable_fact_pending`** in one transaction. I rebased on top of your `7c876f3` (kept in history) and extended it to also clear the new lease table — your reset tests plus mine both run.

Full suite green bar the pre-existing gemini env-pollution config test (fails on `main` too). Re-requesting your review.

src/orchestrator/durableFacts.ts
```diff
@@ -0,0 +1,459 @@
...
+          conversation: input.conversation,
+          turnId: turn.id,
+          error: (e as Error).message,
+        });
+        break;
+      }
+
+      const facts = parseExtractedFacts(raw);
+      for (const f of facts) {
+        await input.memory.upsertDurableFact({
```
kaieriksen: Blocking: `/reset` can repopulate the conversation after it has been cleared. `deleteConversation` removes `durable_fact_pending`, `durable_facts`, and the turns, but an extractor already past `complete()` can reach this upsert afterward and write facts from the pre-reset turn into the now-empty `(persona, conversation)`; its later commit is a no-op. The next conversation with the same key then receives stale facts. Add a reset generation/epoch (or make commit/upsert conditional on the claimed lease/reset state) and abort or discard in-flight extraction before allowing writes, with a regression test that resets while `complete` is blocked.

src/config.ts
```diff
@@ -227,6 +227,49 @@ export const DEFAULT_RETRIEVAL: RetrievalSettings = {
...
+   */
+  leaseMs: number;
+}
+
+export const DEFAULT_DURABLE_FACTS: DurableFactsSettings = {
+  enabled: true,
+  maxInjected: 8,
+  minConfidence: 0.5,
+  maxExtractPerTurn: 4,
+  leaseMs: 300_000,
```
kaieriksen: Blocking: the default 5-minute lease is shorter than the default 60-minute `harnessHardTimeoutMs`. If a primary extraction call runs longer than `leaseMs`, another post-turn extractor can reclaim the still-running turn after expiry and issue a duplicate model call; the original may then commit after the duplicate. Either enforce `leaseMs` above the hard timeout, renew the lease while `complete()` runs, or bind commit/release to an ownership token so an expired/reclaimed lease cannot be completed by the old pass. Add a regression test for expiry during an in-flight completion.

kaieriksen (changes requested):
Changes requested on d096625: reset can repopulate cleared durable facts, and the default lease can expire while the harness is still running. See the two inline blockers.

lenaparkhodges:
Verified @kaieriksen's blocker against the code — it's real, on all three counts:

1. **Timeout skew:** `harnessHardTimeoutMs` defaults to 3,600,000 ms (`config.ts`), and the extraction call in `makeExtractionComplete` runs under it — while `DEFAULT_DURABLE_FACTS.leaseMs` is 300,000 ms. An extraction pass can legitimately run 12× longer than its own lease.
2. **Overlap is possible:** `makeFactExtractor` fires out-of-band and is *not awaited* after a persisted turn, so two quick turns can put two extraction passes in flight concurrently.
3. **Commit is unguarded:** `commitPendingStmt` is an unconditional `DELETE` by (persona, conversation, turn_id) — no lease-expiry or ownership check. A stale pass whose lease expired and was re-claimed will still commit successfully, on top of the duplicate's work. Duplicate model call + last-writer-wins on `durable_facts`.

**Preference: Kai's option 3 (ownership token).** Just raising `leaseMs` above the hard timeout makes crash recovery wait 60+ minutes before a crashed turn is re-claimable — too slow for the common hard-crash case. A claim-issued token checked at commit/release keeps the 5-minute recovery window AND closes the stale-commit hole. Renewal (option 2) is acceptable but adds a moving part; the token is one column and one WHERE clause.

Regression test for expiry-during-in-flight-completion is a must either way. Holding my approval until this lands.

robertclawson:
Both blockers fixed in `1d402ea` — pulled current `main` in first (already merged as of `d096625`).

**1. Reset-repopulation race** — every claim now stamps a per-claim **ownership token** on its lease rows, returned from `claimEvictedForExtraction`. The new `commitExtraction` verifies the lease is still ours (`lease_token === token`), writes the facts, and drops the lease **in one transaction**. A `/reset` deletes the lease rows, so a slow extraction that finishes afterward finds no matching token and its write becomes a no-op — nothing repopulates the wiped conversation. `commitExtractedTurn` and `releaseExtractionLease` are token-gated too.

**2. Lease shorter than harness timeout** — `buildDurableFactsConfig` now floors `leaseMs` at `harness_hard_timeout + LEASE_COMMIT_MARGIN_MS` (300s), so a slow-but-successful pass can never have its lease lapse mid-call. Default is self-consistent with the default 60m hard timeout. And as defense-in-depth, the same token gate makes an expired-then-reclaimed turn's original owner lose the write (no duplicate fact) even if a lease *did* expire.

New tests: store-level token gating (match/mismatch), the exact lease-expiry re-claim → single-fact scenario, an integration test that fires `/reset` from inside `complete()` and asserts zero leak, and a config test proving the lease floor clamps a hand-set 60s lease up. Full suite 2451 pass; the lone red is the pre-existing gemini env-pollution config test (fails on clean main, unrelated).

kaieriksen (approved):
Re-reviewed at 1d402ea. The lease ledger now gives disjoint per-turn claims, token-gated atomic commits, immediate failure release, stale-lease recovery, and reset clears facts/cursor/leases; slow or post-reset owners cannot write. Focused durable-facts/config tests pass 102/102, typecheck passes, production build passes, and full suite passes 2450 with 2 skips; the only 2 failures are the pre-existing ambient Pi API-key routing tests. All prior blocker threads resolved. Safe to ship without prior dogfooding.

lenaparkhodges:
Re-reviewed `1d402ea` — both blockers are genuinely closed. ✅

**1. Reset-repopulation race:** the per-claim ownership token + single-transaction `commitExtraction` (verify lease → write facts → drop lease) closes it properly. Guarding the fact *write* behind the live lease — not just the commit — is the right fix, and re-stamping the token on re-claim also covers the lease-expiry-then-reclaim duplicate case. Verified locally: `tests/memory-durableFacts.test.ts`, `tests/orchestrator-durableFacts.test.ts`, `tests/config.test.ts` — 102/102 pass on `1d402ea`.

**2. Lease < harness timeout:** flooring `leaseMs` at `harnessHardTimeoutMs + LEASE_COMMIT_MARGIN_MS` in `buildDurableFactsConfig`, plus making the DEFAULT self-consistent with the default 60m hard timeout (so ad-hoc Configs are race-safe too), is the right belt-and-suspenders.

Schema note (non-blocking): `durable_fact_pending` is new in this PR so the `lease_token` column needs no ALTER migration — confirmed there's no production table to upgrade.

From a product lens this is merge-ready: the feature is lossless under partial failure, reset-safe, and can't double-fire model calls. Nice work, Robbie.

PR #321, opened 2026-07-25
harnesses: hard-disable background agents and subagents (claude + codex)

lenaparkhodges:
## Why

Andrew's standing rule is **no subagents, ever** — but both harnesses could background work or spawn subagents today:

- **claude**: `run_in_background` was fully available on the Bash and Task tools, and the Task subagent tool sat on the advertised surface.
- **codex**: `multi_agent` (spawn_agent / resume_agent / close_agent / wait_agent) is a **stable feature, ON by default** — verified against codex 0.145.0 (`codex features list`).

## What — three layers, each verified against the real binaries

**1. claude env — `CLAUDE_CODE_DISABLE_BACKGROUND_TASKS=1`**
Injected into the subprocess env *after* `filterAuthEnv`. Verified against claude 2.1.170: the flag strips `run_in_background` from the tool **schemas**, so the capability is never advertised to the model (also kills Ctrl+B and the "use run_in_background" prompt nudges). It must be set here explicitly: `filterAuthEnv` strips the whole `CLAUDE_CODE_*` namespace, so exporting it in `~/.env` would silently never arrive — and an inherited `=0` can no longer re-enable backgrounding (regression-tested).

**2. Tool/feature surface removal**
- claude: `Task` added to `--disallowedTools` — subagents gone foreground *and* background.
- codex: `-c features.multi_agent=false -c features.multi_agent_v2=false` on both the normal and judge flag sets. `-c` overrides beat `~/.codex/config.toml`, which matters because normal turns intentionally load user config (connectors).

**3. Parser tripwire (fail-safe)**
Any subagent-stamped event becomes a **recoverable error** → the orchestrator kills the harness and falls through instead of letting an unmonitored agent run:
- claude: `subagent_type` envelope stamp, `isSidechain: true`, or a `Task` tool_use
- codex: `collab_*` items, `sub_agent_activity`, or the four agent tool names

Guards against a future CLI update renaming or bypassing layers 1–2. Same fail-safe direction as the existing API-error gate.

## Tests

13 new: marker classifiers, tripwire parsing (both harnesses), env injection including the inherited-`=0` override, `--disallowedTools` argv, codex flag-set contents. New `env` mode in fake-claude.sh.

Suite: **2426 pass, 0 new fails** (the 2 failures are the pre-existing PiHarness env issues on main). Typecheck clean.

## Out of scope

pi harness: no equivalent subagent/background surface identified — flag if one turns up and it'll get the same treatment.

### DIFF
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -518,8 +518,9 @@ export function parseStreamJson(parsed: unknown): HarnessChunk | undefined {
   // policy (CLAUDE_CODE_DISABLE_BACKGROUND_TASKS in the env + Task in
   // --disallowedTools), so the CLI should never emit subagent activity. If a
   // future CLI version routes around those guards, any envelope carrying
-  // subagent lineage becomes a RECOVERABLE error: the orchestrator kills
-  // this harness and falls through to the next one instead of letting an
+  // subagent lineage becomes a RECOVERABLE, TERMINAL error: the runner kills
+  // the subprocess immediately and suppresses every line after this one, and
+  // the orchestrator falls through to the next harness instead of letting an
   // unmonitored agent run loose. Fail-safe direction, same as the API-error
   // gate below.
   if (isSubagentActivity(obj)) {
@@ -527,6 +528,7 @@ export function parseStreamJson(parsed: unknown): HarnessChunk | undefined {
       type: "error",
       error: "claude emitted subagent activity (disabled by phantombot policy)",
       recoverable: true,
+      terminal: true,
     };
   }
 
--- a/src/harnesses/claude.ts
+++ b/src/harnesses/claude.ts
@@ -148,7 +148,19 @@ export class ClaudeHarness implements Harness {
     // routing var into the subprocess env (reloadEnvFiles just re-sourced
     // ~/.env), so claude resolves credentials from ~/.claude/.credentials.json.
     const env = withPersonaEnv(
-      filterAuthEnv(process.env),
+      {
+        ...filterAuthEnv(process.env),
+        // Background agents are disabled by product policy. This flag makes
+        // the CLI strip `run_in_background` from the Bash and Task tool
+        // SCHEMAS entirely (verified against claude 2.1.170), so the model
+        // cannot background anything — the capability is never advertised.
+        // It must be set HERE, after filterAuthEnv: the filter strips the
+        // whole CLAUDE_CODE_* namespace from the inherited env, so exporting
+        // it in ~/.env or the shell would silently NOT reach the subprocess
+        // (and a stray CLAUDE_CODE_DISABLE_BACKGROUND_TASKS=0 out there must
+        // not be able to re-enable backgrounding).
+        CLAUDE_CODE_DISABLE_BACKGROUND_TASKS: "1",
+      },
       req.persona,
       req.conversation,
     );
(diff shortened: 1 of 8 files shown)


src/harnesses/claude.ts
```diff
@@ -469,6 +514,22 @@ export function parseStreamJson(parsed: unknown): HarnessChunk | undefined {
...
 
+  // Subagent tripwire. Background agents and the Task tool are disabled by
+  // policy (CLAUDE_CODE_DISABLE_BACKGROUND_TASKS in the env + Task in
+  // --disallowedTools), so the CLI should never emit subagent activity. If a
+  // future CLI version routes around those guards, any envelope carrying
+  // subagent lineage becomes a RECOVERABLE error: the orchestrator kills
+  // this harness and falls through to the next one instead of letting an
+  // unmonitored agent run loose. Fail-safe direction, same as the API-error
+  // gate below.
+  if (isSubagentActivity(obj)) {
```
kaieriksen, line 526: This tripwire only returns a recoverable error; `runHarnessProcess` continues consuming stdout until the child exits, and it does not kill the process on parser errors. If a future Claude version emits this envelope and then emits ordinary assistant text or tool events, those later events will still be yielded to the orchestrator/user, while the fallback happens only after exit. Please make the runner terminate the process when this policy error is observed (or otherwise make this an immediately terminal stream state), and add a regression proving post-tripwire output is suppressed.

src/harnesses/codex.ts
```diff
@@ -153,12 +165,61 @@ export const PHANTOMBOT_JUDGE_CODEX_FLAGS = [
...
+  // Subagent tripwire. multi_agent is disabled via -c flags on every spawn,
+  // so codex should never emit collab/sub-agent items. If a future version
+  // routes around that (or a config layer we don't control re-enables it),
+  // the turn becomes a RECOVERABLE error: the orchestrator kills this
+  // harness and falls through instead of letting an unmonitored agent run.
+  if (type === "item.started" || type === "item.completed") {
+    const item = obj.item;
+    if (typeof item === "object" && item !== null) {
+      const it = item as Record<string, unknown>;
+      if (isCodexSubagentActivity(it.type, it.name)) {
```
kaieriksen, line 214: Same lifecycle issue here: this parser error is yielded, but `runHarnessProcess` keeps reading and forwarding subsequent Codex events because parser errors do not stop or kill the child. A future Codex version could therefore emit a subagent marker and then ordinary assistant/tool output that still reaches the user before fallback. Please terminate the process immediately on this policy violation (or make the stream terminal), with a regression covering output after the tripwire.

kaieriksen (changes requested):
Changes requested: the Claude and Codex subagent tripwires return recoverable errors, but the shared harness runner continues consuming and forwarding stdout after those errors. A future CLI regression could therefore emit forbidden subagent activity and then ordinary output that reaches the user before fallback. The runner must terminate the child or make the stream terminal, with regression coverage.

lenaparkhodges:
Pushed e5ce55c addressing the tripwire-streaming finding: policy violations are now terminal — the runner kills the child immediately and suppresses all post-tripwire output, while ordinary recoverable errors still stream. Regression tests cover both paths; full suite green. Ready for re-review.

kaieriksen (approved):
Re-reviewed at e5ce55c. The prior fail-closed blockers are fixed: terminal policy errors now terminate the process group, suppress same-batch and later stdout, and avoid synthesizing a done chunk. Claude/Codex tripwires and runner regressions are covered; CI, typecheck, and test checks are green. I could not rerun Bun locally because Bun is unavailable on this host.

kaieriksen (approved):
Reviewed the updated terminal tripwire and regression coverage. The runner now kills the process and suppresses same-batch and later output; ordinary recoverable errors retain streaming semantics. Focused harness tests pass.

Issue #324, opened 2026-07-26
Task-triggered wakeups should wake the agent with full memory context + durable facts


robertclawson:
## Problem
When a scheduled task wakes an **agent turn** (`runTurn` via `tick.ts`, *not* a `--command` script), it starts nearly context-blind. Confirmed in source: `tick.ts` calls `runTurn` with only persona, `conversation=tick:<id>`, the task prompt, agentDir, harnesses, memory. It **omits** `retrieve`, `pullFacts`, `extractFacts`, `indexTurns`.

So a task-woken agent gets:
- ✅ Persona prompt (SOUL/IDENTITY/AGENTS)
- ✅ History — but only prior fires of *this same task* (`tick:<id>` silo)
- ✅ Full harness/tools (`memory search` still callable by hand)
- ❌ No semantic recall (`retrieve` off)
- ❌ No durable facts (`pullFacts` off)
- ❌ Writes nothing back — not embedded, not extracted

## Why this is wrong
We only ever wake the LLM (instead of running a `--command` script) *because* it can reason and improvise. Waking it blind — and then discarding whether the improvisation worked — is the worst of both: we pay for intelligence, deny it the context to be competent, and throw away the lesson. A task that SSHes into the swarm and fumbles for the working incantation will fumble the same way next week. Success = "this path works, promote it"; failure = "dead end, don't repeat" — both are signal, and both are currently lost.

The wake type is itself the classifier: `--command` → reflex (stays blind/mute, correct); **agent-woken → lived experience** (recall *in* to start competent, consolidation *out* so it survives).

## Goal
Make an agent-woken task turn behave like a conversation-triggered wake:
- Turn on `retrieve` + `pullFacts` so it wakes with semantic recall and durable facts.
- Turn on `extractFacts` + `indexTurns` so its first-hand observations consolidate back (as the `self_task` provenance tier from the durable-facts work).
- Keep `--command` script tasks exactly as they are — blind and mute by design.

## Depends on
- **#322** (durable facts: persona-scope + provenance/decay) — task observations land in the `self_task` tier with fast decay.
- **#323** (time-decay/retraction on the embedded retrieval layer) — **primary blocker.** Turning on `retrieve` for task wakeups pulls from the same embedded layer that currently resurfaces stale ghosts (e.g. kw-openclaw). Waking tasks with recall before that layer decays/retracts would inject stale context into every fire. Close the memory-deficiency gaps first.

Sequenced **after #323** (and #322).

Issue #323, opened 2026-07-26
Time-decay / retraction on the embedded retrieval layer


robertclawson:
## Problem
Semantic recall surfaces **stale facts from old indexed turns and static memory files**, and relevance has no clock — an old turn scores just as high today as a recent one. Correcting the record in conversation does nothing, because the correction is one new turn competing against months of old ones.

### Motivating example
The box that hosts Lena/Kai/Jake was renamed `kw-openclaw` → `kw-phantombot` months ago. Andrew corrects this repeatedly, yet retrieval keeps resurfacing "kw-openclaw" because:
1. **Static memory files** (`MEMORY.md`, KB notes) still contain the old text — no decay, hand/nightly-edited.
2. **Embedded old turns** — months of history literally saying "kw-openclaw" — get surfaced by relevance regardless of age.

The durable-facts persona/decay work does **not** fix this — that touches only the `durable_facts` table. This ghost lives in the retrieval/index layer + static text.

## Goal
Give the embedded retrieval layer a notion of **recency and retraction**, so corrected/superseded information stops resurfacing:
- Time-decay on embedded-turn relevance (age penalty in ranking), so old assertions fade unless reinforced.
- A retraction/supersession mechanism — when the principal corrects a fact, later state should win over indexed history.
- Consider a pass to correct/curate the static `MEMORY.md` + KB mentions (currently the fix is manual).

## Notes
Retrieval is hybrid keyword+vector today; decay/thresholds are tunable. This is the "relevance has no clock" fix we deferred from the durable-facts discussion.

PR #325, opened 2026-07-26
feat(memory): re-scope durable facts to persona with provenance + decay

### LINKED ISSUE #322: Durable facts: re-scope to persona with provenance tiers + time decay
## Summary
Move `durable_facts` from conversation-scoped to **persona-scoped**, so one persona has a single shared memory of facts across all its channels (DM, voice, email-triage, task wakeups) — "one mind" instead of a fact silo per conversation.

Re-keying alone is unsafe for multi-party personas (group chats), so this PR ships the **full model (Option A)**: persona re-key + provenance + decay together. We dogfood on Lena, review logs, then open for review.

## Current state (confirmed in source)
- `durable_facts` keyed `(persona, conversation, fact_norm)`; UNIQUE + rank index both carry `conversation`.
- `topDurableFacts` / `countDurableFacts` / `upsert` all take `(persona, conversation)`.
- Extraction runs per-conversation at the eviction cliff; read is pure SQL top-N by `(confidence DESC, last_seen_at DESC)`.
- A persisted `Turn` stores only `role` (`user`/`assistant`) — **no speaker identity**. `senderId` exists at the engine layer but never reaches the store.

## Why provenance is required (not polish)
In a group thread (e.g. Andrew + Kai + Lena + Robbie), extraction would mine facts asserted by *other agents* into the same pool the private DM reads from — treating "overheard from Kai" as trusted as "Andrew told me directly." Provenance is the trust firewall conversation-scoping was quietly providing.

### Three provenance tiers
- **`principal`** (Andrew) → highest trust, slowest decay. Standing facts hold until corrected.
- **`self_task`** (the phantom's own first-hand task observation) → trust the sighting, but decay fast — infra state churns (`5 nodes` → `6 nodes`).
- **`other_agent`** (Kai/Lena in a group) → lowest, down-weighted / held.

## Changes
1. **Schema/migration:** add `source` (`principal`/`self_task`/`other_agent`) + `asserted_by` to `durable_facts`; re-key UNIQUE → `(persona, fact_norm)`, drop `conversation` from rank index. Migrate existing rows: collapse per persona, keep max confidence + latest `last_seen_at`, stamp `source=principal`.
2. **Plumb identity:** carry `senderId`/resolved source through `appendTurn` → into `commitExtraction`, so the eviction cliff knows who spoke.
3. **Store API:** `topDurableFacts` / `countDurableFacts` / `upsert` lose the `conversation` arg. (Extraction cursor stays per-conversation — it still walks each conversation's turn stream independently.)
4. **Read path:** weight by `source` and apply time-decay on `last_seen_at` in the rank. Half-life per tier (principal slow, self_task fast, other_agent down-weighted). Confirmed facts refresh their timestamp on recurrence, so they naturally outlive transient sightings.
5. **Injection budget:** with one merged pool, cap top-N so no single chatty conversation starves the rest.
6. **Logging (dogfood):** behind a `durableFacts.debug` flag, log every extraction (fact, source, confidence, decay-adjusted rank) and every injection set, so we can audit the Lena dogfood for safety before wider rollout.

## Open decision
- Decay shape: half-life on `last_seen_at` per tier (leaning this) vs hard TTL on `self_task`.

## Explicitly out of scope
Task wakeups getting recall/context (separate issue), trend/journal organ, semantic retrieval-gating, embedding-layer changes.

## Dogfood plan
Build on Lena, run with `durableFacts.debug` on, review extraction + injection logs to confirm no cross-party bleed and sane decay, then request review.

robertclawson:
## What & why

Durable facts were keyed **`(persona, conversation)`** — a fact you told me in one DM never surfaced in another chat or a task run. This re-scopes them to **`(persona)`** so every conversation reads and writes one shared, persona-wide pool.

That merge is only *safe* with a trust model, so this PR ships the re-key together with **provenance + time-decay** (agreed in design — they're what stop a group member's claim from masquerading as something the owner told me).

Closes #322.

## The model

Three provenance tiers, each with its own trust weight, decay half-life, and retirement age:

| source | who | weight | half-life | retire after |
|---|---|---|---|---|
| `principal` | the owner, in a trusted turn | 1.0 | 180d | **365d** |
| `self` | my own assistant turn (first-hand observation) | 0.6 | 30d | 90d |
| `other` | a third party in a shared/group conversation | 0.3 | 7d | 30d |

- **Ranking** (read path): `weight · confidence · 2^(-ageDays / halfLife)`, filtered by an inject floor, top-N injected. Pure read + cheap in-memory math, no LLM.
- **Refresh-on-recall**: injecting a fact bumps its `last_seen_at` (fire-and-forget). Frequently-used facts stay fresh; only genuinely-unused ones decay and retire. This is why a yearly-reviewed procedure the owner gave me (principal, 365d) survives to its next natural use, while a stale self-observation ("swarm had 5 nodes") rots out.
- **Promotion on conflict**: the same normalized fact from two conversations collapses to one row keeping max confidence and the **highest-trust** source seen — `self` → `principal` upgrades, but a later `other` mention never downgrades.

## Changes

- **`store.ts`** — `durable_facts` unique key `(persona, conversation, fact_norm)` → `(persona, fact_norm)`; new `source` column + re-keyed rank index. `turns` gets a `source` column. New `touchDurableFacts` (recall bump) and `pruneExpiredDurableFacts` (retirement). `topDurableFacts`/`countDurableFacts` drop the conversation arg. `upsert` conflict clause promotes source by trust priority. **/reset no longer wipes facts** (they're shared knowledge now) — it still clears the conversation's turns, extractor cursor, and leases.
- **Migration** — legacy DBs rebuild `durable_facts`, collapsing duplicate `(persona, fact_norm)` rows (keep max confidence + latest `last_seen_at`), backfilling `source=principal`; `turns.source` backfills `assistant`→`self`. Fresh DBs get the new schema directly.
- **`turn.ts`** — stamps each appended turn's source from the trust bit: trusted user → `principal`, untrusted user → `other`, assistant → `self`.
- **`durableFacts.ts`** — extraction carries the turn's source onto facts; decay/weight ranking + recall bump on the read path; out-of-band prune after each extraction pass.
- **`config.ts`** — `durable_facts.tiers.{principal,self,other}.{weight,half_life_days,max_age_days}`, `inject_floor`, and `debug` (all clamped, env-overridable).

## Dogfooding on Lena

Set `PHANTOMBOT_DURABLE_FACTS_DEBUG=1` (or `[durable_facts] debug = true`) to log every inject-ranking decision (per fact: source, confidence, ageDays, decayed score) and every prune. Plan: run on Lena, watch the logs to confirm `other`-tier group claims stay down-weighted and nothing leaks into her private pool, then widen.

## Tests

`bun test` green (the one unrelated failure — `migrates a legacy Gemini CLI chain` — is a pre-existing env leak on the dogfood box, reproduces on `main`). New coverage: persona-wide merge + cross-conversation collapse, source promotion, recall bump, retirement prune, and a **legacy on-disk DB migration** test; existing lease/cursor/race suite adapted to the new signatures.

## Deliberately out of scope (follow-ups)

- #323 — time-decay/retraction on the **embedded retrieval** layer (the kw-openclaw ghost lives there + in static memory files, not in durable_facts).
- #324 — task wakeups waking with full memory context + durable facts.

### DIFF
--- a/src/orchestrator/durableFacts.ts
+++ b/src/orchestrator/durableFacts.ts
@@ -35,10 +35,19 @@
 
 import { homedir } from "node:os";
 
-import type { Config, DurableFactsSettings } from "../config.ts";
+import type {
+  Config,
+  DurableFactsSettings,
+  FactSourceTiers,
+} from "../config.ts";
 import type { Harness, HarnessChunk } from "../harnesses/types.ts";
 import { log } from "../lib/logger.ts";
-import type { DurableFact, MemoryStore, Turn } from "../memory/store.ts";
+import type {
+  DurableFact,
+  FactPruneCutoffs,
+  MemoryStore,
+  Turn,
+} from "../memory/store.ts";
 import { DEFAULT_HISTORY_LIMIT } from "./turn.ts";
 
 /** A single extracted fact + the extractor's confidence in its durability. */
@@ -160,6 +169,8 @@ export interface ExtractDurableFactsResult {
   turnsProcessed: number;
   /** How many facts were upserted (new or recency-bumped). */
   factsWritten: number;
+  /** How many facts were retired by the age-based prune this pass. */
+  factsPruned: number;
 }
 
 /**
@@ -189,6 +200,7 @@ export async function extractDurableFactsOnEviction(
     triggered: false,
     turnsProcessed: 0,
     factsWritten: 0,
+    factsPruned: 0,
   };
   if (!input.settings.enabled) return result;
 
@@ -268,6 +280,11 @@ export async function extractDurableFactsOnEviction(
         facts.map((f) => ({
           fact: f.fact,
           confidence: f.confidence,
+          // Stamp each fact with the provenance of the turn it came from — the
+          // trust tier the read path weights + decays by. A principal turn →
+          // principal facts, the persona's own assistant turn → self facts, a
+          // third party in a shared conversation → other facts.
+          source: turn.source,
           sourceTurnId: turn.id,
         })),
       );
@@ -300,6 +317,31 @@ export async function extractDurableFactsOnEviction(
         conversation: input.conversation,
         turnsProcessed: result.turnsProcessed,
         factsWritten: result.factsWritten,
+      });
+    }
+
+    // Retirement floor: prune facts past their tier's max age. Runs on the
+    // out-of-band write path (never the hot read path) so a growing table is
+    // trimmed as extraction happens. Any recall refreshes last_seen_at, so only
+    // genuinely-unused facts are eligible. Best-effort — a prune failure must
+    // not fail the pass.
+    try {
+      const pruned = await input.memory.pruneExpiredDurableFacts(
+        input.persona,
+        pruneCutoffs(input.settings.tiers),
+      );
+      result.factsPruned = pruned;
+      if (pruned > 0) {
+        log.info("durable-facts: retired stale facts", {
+          persona: input.persona,
+          conversation: input.conversation,
+          factsPruned: pruned,
+        });
+      }
+    } catch (e) {
+      log.warn("durable-facts: prune failed; continuing", {
+        persona: input.persona,
+        error: (e as Error).message,
       });
     }
     return result;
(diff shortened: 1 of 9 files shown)


src/orchestrator/durableFacts.ts
```diff
@@ -440,18 +587,20 @@ export function formatDurableFacts(facts: DurableFact[]): string | undefined {
   const usable = facts.filter((f) => f.fact.trim().length > 0);
   if (usable.length === 0) return undefined;
   const header =
-    "Standing facts established earlier in this conversation, kept after the " +
-    "original messages scrolled out of the live window — background context, " +
+    "Standing facts about the principal and their world, learned across " +
```
kaieriksen, line 614: Blocking: `other` facts are selected with a lower weight, but their provenance is discarded here and they are rendered as unqualified `Standing facts about the principal`. An allowed untrusted/group turn can therefore become a durable fact and, once above the inject floor, be presented to a principal turn with no indication that it was third-party/low-trust. That lets attacker-controlled text masquerade as owner knowledge and defeats the provenance boundary. Preserve the source in the formatted block (or exclude `other` from principal injection) and add a regression test.

src/orchestrator/durableFacts.ts
```diff
@@ -268,6 +280,11 @@ export async function extractDurableFactsOnEviction(
         facts.map((f) => ({
           fact: f.fact,
           confidence: f.confidence,
+          // Stamp each fact with the provenance of the turn it came from — the
+          // trust tier the read path weights + decays by. A principal turn →
+          // principal facts, the persona's own assistant turn → self facts, a
+          // third party in a shared conversation → other facts.
+          source: turn.source,
```
kaieriksen, line 289: Blocking: although this now stores `other` provenance, `renderTurnForExtraction` still labels every user turn as `PRINCIPAL` (including `source=other`). Untrusted messages that the screen allows through are still extracted, so the extractor is told that third-party text came from the principal; that can bias both what it promotes and how it rewrites the claim before storage. Derive the speaker label from `turn.source` (for example `OTHER` for `other`) and add coverage for an allowed untrusted turn.

kaieriksen (changes requested):
Request changes: the trust boundary is not preserved when third-party facts are extracted and later injected.

lenaparkhodges (changes requested):
Concur with Kai — both findings verified against 11e7db5, and they're the right blockers:

1. **Injection boundary drops provenance** (durableFacts.ts:586-595): `formatDurableFacts` renders every fact as an unqualified bullet under "Standing facts about the principal and their world". An `other`-tier fact that clears the inject floor is presented to a trusted turn as owner knowledge. The tier model is well-built everywhere *except* the one place it matters most — the prompt boundary.

2. **Extraction label lies about the speaker** (durableFacts.ts:359-360): `renderTurnForExtraction` labels every user turn `PRINCIPAL` regardless of `turn.source`, so the extractor LLM is told third-party text came from the owner — biasing both confidence and phrasing before the `source` stamp is ever applied.

**One amplification Kai didn't note:** the recall bump makes finding 1 self-reinforcing. Once an `other` fact is injected, `touchDurableFacts` resets `last_seen_at`, its decayed score jumps back to `0.3 · confidence` (well above the 0.05 floor), and it keeps winning the top-N cut — so an injected third-party claim can persist *indefinitely*, never retiring. Fixing the injection boundary (source-qualified rendering or excluding `other` from trusted-turn injection) also breaks this loop, so the two fixes are the same fix.

The underlying design (persona-wide pool + tiered decay + promotion-on-conflict) is sound and the migration/lease work looks solid — this is close. Gate merge on the two provenance-boundary fixes + regression tests.

robertclawson:
Both findings addressed in e56e90d — thanks Kai & Lena, both were real and I've verified them against the code, not just the review text.

**1. Injection boundary (`formatDurableFacts`):** `other`-tier facts are now tagged inline — `- [unverified — reported by a third party in a shared conversation] <fact>` — while `principal`/`self` render unchanged as plain background bullets. Chose source-qualified rendering over dropping `other` from injection so the persona-wide pool keeps its value, but a third-party claim can never be presented to a trusted turn as owner knowledge. The tag is per-line (not a section header) so it survives any slicing/reordering. This also closes Lena's recall-bump amplification: an injected `other` fact still refreshes `last_seen_at`, but it can no longer masquerade as trusted while it does so.

**2. Extraction label (`renderTurnForExtraction`):** speaker label now derives from `turn.source`, not role — `THIRD_PARTY` for `other`, `PRINCIPAL`/`ASSISTANT` otherwise. Added a line to `EXTRACTION_SYSTEM` telling the extractor not to assume a `THIRD_PARTY` speaker is describing the principal and to be conservative (lower confidence, or nothing) about third-party claims about the principal.

**Regression tests** (both in `tests/orchestrator-durableFacts.test.ts`): an `other` fact is inline-qualified and never a bare owner bullet; an allowed untrusted turn renders `speaker="THIRD_PARTY"` (not `PRINCIPAL`) and stores `source=other`.

Full suite green locally — `orchestrator-durableFacts` 22 pass, `memory-durableFacts` 38 pass, clean tsc. Ready for re-review.

lenaparkhodges (approved):
Re-review after e56e90d: both of my findings are genuinely closed, verified against the code and a local run, not just the review text.

**1. Injection boundary** — `formatFactLine` now tags `other`-tier facts inline (`[unverified — reported by a third party in a shared conversation]`) while principal/self render unchanged. Per-line tagging is the right call over a section header — it survives any slicing/reordering, and it also breaks the recall-bump amplification loop I flagged: an injected `other` fact can no longer masquerade as trusted while it refreshes `last_seen_at`.

**2. Extraction label** — `speakerLabel()` derives from `turn.source`, so an allowed untrusted turn is announced to the extractor as `THIRD_PARTY`, and the updated `EXTRACTION_SYSTEM` line tells it to be conservative about third-party claims about the principal. This closes the pre-stamp bias correctly.

**Verification:** checked out e56e90d locally — `orchestrator-durableFacts` + `memory-durableFacts` 60 pass / 0 fail, clean `tsc --noEmit`. The two new regression tests (inline qualification; `THIRD_PARTY` rendering + `source=other` persistence) cover exactly the boundary I was worried about.

LGTM — good to merge. 🚀

robertclawson:
Pushed `8cc569e` — closes the second finding (recall-bump immortality loop).

`e56e90d` fixed the masquerade (other-tier facts inject tagged `[unverified — reported by a third party…]` + `THIRD_PARTY` extractor label) but left `touchDurableFacts` refreshing **every** injected fact, `other` included — so a third-party claim that cleared the inject floor kept resetting its own clock and never retired.

Fix: exclude `source === "other"` ids from the recall bump. Third-party facts still inject (tagged) when they clear the floor, but their `last_seen_at` is never refreshed, so they decay and hit the retirement floor on schedule. Added a regression test: principal facts get bumped on inject, `other` facts never do. Full suite green (one pre-existing Gemini env-leak failure aside).

kaieriksen (approved):
Re-reviewed current head 8cc569e.

The prior provenance blockers are fixed: other-source facts remain explicitly unverified at injection, extraction labels third-party turns as THIRD_PARTY, and other-source facts are excluded from recall refresh so they can retire. Focused durable-facts tests pass (61/61) and typecheck is clean.

PR #326, opened 2026-07-26
feat(memory): time-decay raw conversation-turn hits in retrieval

### LINKED ISSUE #323: Time-decay / retraction on the embedded retrieval layer

robertclawson:
Closes #323.

## Problem
Retrieval ranking has **no clock**. `hybridSearch` fuses BM25 (keyword) and cosine (vector) with **RRF** — a purely rank-based merge that discards raw scores and knows nothing about age. So a months-old turn and today's correction compete on relevance alone. That's the kw-openclaw ghost: a superseded fact keeps resurfacing because it still *lexically matches*, and relevance has no memory of being superseded.

## Approach — decay-only, per-kind (agreed with Andrew)
Multiplicative time-decay on **turn hits only**:

```
score *= 0.5 ^ (ageDays / halfLifeDays)   # clamped to [floor, 1]
```

- **Turns decay** (half-life default **30d**) → stale chatter sinks no matter how strongly it matches; fresh corrections naturally outrank it.
- **Curated `memory/` + `kb/` notes are immune** (factor 1.0, never aged). This is the deliberate "keep for years, maybe forever" tier — a backup runbook reviewed once a year must not be deleted *because* it's rarely recalled.
- **No refresh-on-recall, by design.** A ghost is recalled *because* it matches; bumping its clock on recall would keep it alive — the opposite of the goal. Pure age-decay only.

### Consequence (explicit)
Decay-only ages raw turns; it does **not** touch curated files. Stale text hand-written into `CLAUDE.md` / auto-memory can still resurface until those files are edited — that's plain manual housekeeping, no PR needed, and out of scope here.

## Mechanics
- **Age source:** parsed from the ISO timestamp `renderTurnForIndex` already bakes into every `turn_docs.content` (`[role ISO]…`). **No schema change, no migration** to the fragile turn/embedding tables.
- `search()`: decay applied before the final sort+slice, with a **widened turn candidate pool** (25) so a fresh turn just outside the raw-BM25 top-N can climb in once ghosts are damped.
- `hybridSearch()`: decay applied **once**, to the fused RRF scores of turn paths, before the slice (inner candidate lists stay relevance-ordered so RRF fuses on merit and decay isn't double-counted). The no-vector fallback path honours decay too.
- `searchExpanded()` forwards decay to `search()`; graph neighbours are notes → immune.
- **Config:** new `[retrieval.decay]` block (`enabled`, `half_life_days`, `floor`) with `PHANTOMBOT_RETRIEVAL_DECAY_*` env overrides and clamping. **Disabled = zero behaviour change.** The CLI `memory search` path doesn't pass `decay`, so raw relevance ranking there is unchanged.

## Tests
- `turnDecayFactor` math (half-life, floor clamp, disabled, future-turn).
- `parseTurnCreatedAtMs` (extract + graceful undefined).
- Integration: a 300-day-old turn that BM25-outranks a fresh one **wins without decay, loses with it**; same for `hybridSearch`.
- Curated note score **identical** with/without decay (immunity proven).
- Unparseable age → factor 1 (no-op).
- Config override / clamp / fractional parse.

`bun test` green — 2476 pass. The one failure (`migrates a legacy Gemini CLI chain`) is a **pre-existing env leak on this box** (a real `PHANTOMBOT_GEMINI_API_KEY` overrides the test's TOML); it reproduces on clean `main`, unrelated to this PR.
---

## Follow-on in this PR — nightly KB supersession (65c930e)
Decay handles *old* raw turns, but not curated notes that became **wrong** (subject decommissioned, or an assumption later invalidated). Those are immune to decay by design, so they need a *correction* path, not an age one.

The nightly `kb` stage was **purely additive** ("open and update it"), so a note that became wrong kept the stale claim standing next to the new one — nothing recorded that the old belief was no longer valid.

Now the stage **reconciles**:
- **ADDS** (new case/edge/link) → update in place, as before.
- **CONTRADICTS / INVALIDATES** → rewrite the body to current truth, bump `updated:`, and append a dated line under a `## Changelog` section (`YYYY-MM-DD: was X → now Y, because Z`). The old belief survives as history but stops being served as current truth.
- **Whole subject DECOMMISSIONED** → set frontmatter `status: obsolete` with a dated reason + `[[wikilink]]` to any replacement, instead of deleting — recall knows it is dead rather than silently losing it.

Applied to both prompt variants (`buildNightlyPrompt` PHASE 3 and the checkpointed `NIGHTLY_STAGE_BODY.kb`; bound `today` into the compact `kb` arrow). Regression test asserts the supersession language + the dated changelog binding. Full suite still green (2477 pass; same lone pre-existing Gemini env-leak).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -200,6 +200,37 @@ export interface RetrievalSettings {
    * vector path is used instead, so Gemini users are unaffected).
    */
   graphExpansion: GraphExpansionSettings;
+  /**
+   * Time-decay on raw conversation-turn hits. Ranking has no clock of its own —
+   * BM25 + RRF are purely relevance-based, so a stale turn that lexically
+   * matches (e.g. a superseded infra note repeated in old chat) competes with a
+   * fresh correction on equal footing and can win forever. Decay multiplies a
+   * turn hit's score by `0.5^(ageDays/halfLifeDays)` (floored), so old turns
+   * sink no matter how strongly they match. Applies ONLY to `turns` scope —
+   * curated memory/ + kb/ notes are immune (factor 1.0) and never age out, so
+   * rarely-recalled-but-important runbooks survive for years. There is no
+   * refresh-on-recall by design: a ghost is recalled *because* it matches, so
+   * bumping it on recall would keep it alive — the opposite of the goal.
+   */
+  decay: RetrievalDecaySettings;
+}
+
+/** Time-decay knobs for raw conversation-turn hits (curated notes are immune). */
+export interface RetrievalDecaySettings {
+  /** Master switch. When false, turn hits are ranked without any time-decay. */
+  enabled: boolean;
+  /**
+   * Half-life in days for a turn hit's relevance score. After this many days a
+   * turn's score is halved; after 2× it is quartered; etc. Lower = turns age
+   * out faster. Only affects `turns` scope. `<= 0` disables decay (factor 1).
+   */
+  halfLifeDays: number;
+  /**
+   * Lower bound on the decay multiplier so a very old turn's score is damped
+   * but never driven to exactly 0 — it stays a tie-broken last resort rather
+   * than vanishing. 0..1; 0 lets ancient turns decay to nothing.
+   */
+  floor: number;
 }
 
 /** OKF link-graph expansion knobs (BM25-only retrieval path). */
@@ -218,6 +249,16 @@ export const DEFAULT_GRAPH_EXPANSION: GraphExpansionSettings = {
   maxAdd: 3,
 };
 
+export const DEFAULT_RETRIEVAL_DECAY: RetrievalDecaySettings = {
+  enabled: true,
+  // 30 days: raw chat history a month old is worth about half a fresh turn.
+  // Long enough that this-week's context is barely touched; short enough that
+  // a months-old ghost is effectively dead (180d → 0.5^6 ≈ 0.016 × floor).
+  halfLifeDays: 30,
+  // Keep a sliver so an ancient turn is a last-resort tie-break, not erased.
+  floor: 0.02,
+};
+
 export const DEFAULT_RETRIEVAL: RetrievalSettings = {
   enabled: true,
   limit: 5,
@@ -225,6 +266,7 @@ export const DEFAULT_RETRIEVAL: RetrievalSettings = {
   minScore: 0,
   turnIndexing: DEFAULT_TURN_INDEXING,
   graphExpansion: DEFAULT_GRAPH_EXPANSION,
+  decay: DEFAULT_RETRIEVAL_DECAY,
 };
 
 /**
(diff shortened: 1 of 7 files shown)


lenaparkhodges (approved):
**Approved — ship it. ✅** This closes the kw-openclaw ghost class properly: decay lives in the ranking layer, not the data layer, so nothing is ever deleted and curated notes are immune by construction.

Verified on the branch:
- **Correct scope**: decay touches only `turns/` paths — `turnDecayFactors` filters by prefix, notes get factor 1 (test proves byte-identical note scores).
- **No double-count**: hybrid path fuses RRF on raw relevance and decays the fused score exactly once before the slice; no-vector fallback applies it manually; `search()` widens the turn pool to 25 so fresh turns can climb in.
- **Fail-soft everywhere**: unparseable age → factor 1 (never a guessed age), half-life ≤ 0 → disabled, floor clamped to [0,1], fractional `half_life_days` parsed as float (the #319 lesson held), disabled = zero behaviour change.
- **No migration**: age parsed from the timestamp `renderTurnForIndex` already bakes in — fragile turn tables untouched.
- Tests: 125/125 pass locally (memoryIndex + nightly + config), typecheck clean, CI all green including Windows.

Two non-blocking notes:
1. **`rrfScore` stays raw** while ordering uses the decayed key — so `minScore` thresholds on the *undecayed* score and a floored ghost can still clear the bar. Harmless at default `minScore: 0` (decayed ordering squeezes ghosts out of the top-N whenever fresher hits exist, and `floor` deliberately keeps them as last-resort), but if you ever raise `minScore\>, be aware the threshold sees pre-decay scores.
2. **Nightly KB supersession is prompt-level**, not enforced — the reconcile/changelog/obsolete behaviour depends on the nightly model following instructions. That's inherent to the design, and the `status: obsolete` + dated changelog pattern is the right shape (dead knowledge marked, not silently lost). Worth watching the first few nightly runs to confirm the model actually writes changelogs instead of silently rewriting.

The two-part structure is right: decay handles *old*, supersession handles *wrong* — neither alone was sufficient.

kaieriksen (approved):
Approved at 65c930e. No blocking defects found. Focused tests 125/125, typecheck, production build, hosted test/typecheck/Windows/CodeQL checks pass. The full local run only hit the known interactive Windows installer prompts; non-interactive coverage passed. Non-blocking caveats: hybrid rrfScore remains raw for minScore thresholds, and nightly supersession is prompt-driven.

Issue #327, opened 2026-07-26
Harden: untrusted content fetched mid-turn bypasses the threat judge (provenance laundering)


robertclawson:
## Problem

The threat judge guards the **entry door** (`phantombot ask` — the untrusted turn path used by the email/Plane pollers). Untrusted input that arrives *as* a judged `ask` turn is correctly held-or-cleared before any capable harness runs.

**The gap:** if an agent turn (e.g. a #324 task-woken turn, or any normal turn) itself calls a tool that pulls untrusted content **mid-turn** — reading an email body via `gog`, fetching a web page, reading a webhook payload — that content **never passes the judge**. It enters as tool output inside an already-trusted turn.

## Why it matters

With fact-extraction enabled on task turns (#324), any untrusted content ingested mid-turn gets facts extracted and stamped at the **turn's** provenance tier (`self`/`principal`) rather than `other`. That is **provenance laundering**: a phishing email's "your account changed to X" could be promoted into durable `self`-tier memory, bypassing the fast-decay, low-weight `other` tier that untrusted content is supposed to receive — with no human in the loop.

## Current mitigation (norm, not enforced)

The workflow stays sound *by discipline*: untrusted content is meant to arrive as a judged `ask` turn; agents shouldn't fetch-and-extract untrusted content mid-turn. This is a convention, not enforced by code.

## Possible hardening directions (for later)

- Stamp facts derived from untrusted tool output as `other` regardless of the turn's tier.
- Tag/track provenance at the *tool-output* level, not just the turn level.
- Route mid-turn untrusted fetches through the judge (or a lighter-weight check) before their content becomes extractable.
- Skip fact-extraction on tool outputs flagged as untrusted sources.

## Scope

Not a blocker for #324. Documenting so we don't forget — hardening is a separate future effort.

robertclawson:
Resolved by #329 (merged 2026-07-27): assistant turns now default to the new `unverified` tier and are only promoted to `principal` when the owner re-asserts the fact on a trusted turn. This fail-closed default plugs the mid-turn provenance-laundering hole described here — untrusted content ingested mid-turn can no longer land as `self`. Complements the per-fact provenance work in #320/#325/#328. Closing.

PR #328, opened 2026-07-26
feat(memory): task-woken turns get retrieval + durable facts (self tier)

### LINKED ISSUE #324: Task-triggered wakeups should wake the agent with full memory context + durable facts

robertclawson:
Closes #324.

## Problem
Agent-woken (`tick`) tasks ran **context-blind and mute**: no semantic recall and no durable-fact pull on the READ side, and no turn indexing or fact extraction on the WRITE side. A scheduled task woke the persona with a bare prompt and discarded everything it observed — the exact memory instincts a normal conversation turn gets were missing.

## Change
Wire the same four factories the conversation path (`engine.ts`/`ask.ts`) uses into the **agent-woken** branch of `tick.ts` (`!isCommandTask`):
- `makeRetriever` → semantic recall injected into the prompt
- `makeDurableFactPuller` → durable facts injected (pure-SQL READ half)
- `makeTurnIndexer` → task turns become searchable
- `makeFactExtractor` → observations consolidated at the eviction cliff (out-of-band WRITE half)

`--command` tasks stay **blind/mute by design** — untouched. Each factory **self-gates** (returns `undefined` when its feature is disabled in config), so this is a **strict no-op** when retrieval/durable-facts are off.

## Provenance — Option B (`self` tier)
A task feeds the persona its **own scheduled prompt** — a first-hand self-observation, not an untrusted stranger. New optional `userSource` override on `TurnInput` stamps the task user turn `self` instead of the default `other`, so task-derived facts don't land in the untrusted fast-decay/low-weight tier.

`userSource` is **decoupled from `trusted`** on purpose:
- A task wake is `self`-provenance but still **NOT trusted** — it gets no principal command authority and doesn't skip the threat screen framing.
- Undefined `userSource` preserves the legacy `trusted ? "principal" : "other"` behaviour exactly.
- The assistant reply is always `self` regardless — this only affects the user turn.

We chose **B (reuse `self`)** over adding a distinct fast-decay `self_task` tier (Option A): simpler, no new tier surface across config/SQL/weighting, and `self`'s decay is a reasonable home for task chatter.

## Security note
Any **untrusted content a task ingests mid-turn** (email body, web page read via a tool) stays untrusted conceptually, but the current code would extract facts from it at the turn's `self` tier — **provenance laundering**. That gap is **pre-existing** (the judge guards the entry door, not mid-turn tool output) and tracked separately in #327 for future hardening. Not a blocker: today untrusted input is meant to arrive as a judged `ask` turn, and task prompts are self-authored.

## Tests
- `runTurn` provenance matrix: `other` default (untrusted), `principal` when trusted, `self` via `userSource`, and `userSource` wins over the trusted default.
- `tick` integration test: an agent-woken task stamps its persisted **user** turn `self` (spy on `appendTurnPair`).
- Existing tick tests (config has no retrieval/durable-facts blocks) confirm the **no-op-when-disabled** self-gating.
- Full suite green — 2482 pass; only the **pre-existing** `migrates a legacy Gemini CLI chain` env-leak fails (reproduces on clean `main`).

### DIFF
--- a/src/cli/tick.ts
+++ b/src/cli/tick.ts
@@ -217,13 +217,19 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
               harnesses,
               agentDir,
             ),
-            // Provenance: a task wake feeds the persona its OWN scheduled
-            // prompt — first-hand self-observation, not an untrusted stranger.
-            // Stamp the user turn `self` so task-derived facts don't land in the
-            // untrusted `other` tier. NOT `trusted`: a task has no principal
-            // command authority, and untrusted content it might ingest stays
-            // untrusted (see #327 for the mid-turn-fetch hardening follow-up).
-            userSource: "self",
+            // Provenance: an autonomous task wake can ingest UNTRUSTED content
+            // mid-turn (email body, web page, Plane issue) via tools — content
+            // the threat judge never screened because it arrives as tool output,
+            // not a judged `ask` turn (#327). So stamp BOTH turns `other`: every
+            // durable fact a task produces lands in the untrusted tier (weight
+            // 0.3, 7-day half-life, injected only tagged `unverified`, never
+            // recall-bumped) and can never masquerade as first-hand `self`
+            // knowledge or poison the persona-wide pool one tier below the owner.
+            // Conservative by design — we can't tell here which task wakes truly
+            // ingested untrusted content, so all are treated as if they did.
+            // NOT `trusted`: a task has no principal command authority either.
+            userSource: "other",
+            assistantSource: "other",
           })) {
             logBackgroundWakeChunk(task, conversation, chunk);
             if (chunk.type === "text") finalText += chunk.text;
--- a/src/cli/tick.ts
+++ b/src/cli/tick.ts
@@ -51,6 +51,12 @@ import { openTaskStore, type Task, type TaskStore } from "../lib/tasks.ts";
 import { recordTickFired } from "../lib/timerHealth.ts";
 import { openMemoryStore, type MemoryStore } from "../memory/store.ts";
 import { runTurn } from "../orchestrator/turn.ts";
+import { makeRetriever } from "../orchestrator/retrieval.ts";
+import { makeTurnIndexer } from "../orchestrator/turnIndexer.ts";
+import {
+  makeDurableFactPuller,
+  makeFactExtractor,
+} from "../orchestrator/durableFacts.ts";
 
 const WAKE_STREAM_PREVIEW_CHARS = 2000;
 const BACKGROUND_WAKE_HARD_TIMEOUT_MS = 30 * 60 * 1000;
@@ -187,6 +193,37 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
             memory,
             idleTimeoutMs: config.harnessIdleTimeoutMs,
             hardTimeoutMs: BACKGROUND_WAKE_HARD_TIMEOUT_MS,
+            // #324: an agent-woken task should wake with the same memory
+            // instincts a conversation turn gets — semantic recall + durable
+            // facts on the READ side, and consolidate its observations back on
+            // the WRITE side — instead of running context-blind and mute. Each
+            // factory self-gates: it returns undefined when its feature is
+            // disabled in config, so this is a no-op when retrieval/durable
+            // facts are off. (--command tasks never reach here — they stay
+            // blind/mute by design.)
+            retrieve: makeRetriever(config, task.persona, agentDir, conversation),
+            indexTurns: makeTurnIndexer(config, task.persona, conversation, memory),
+            pullFacts: makeDurableFactPuller(
+              config,
+              task.persona,
+              conversation,
+              memory,
+            ),
+            extractFacts: makeFactExtractor(
+              config,
+              task.persona,
+              conversation,
+              memory,
+              harnesses,
+              agentDir,
+            ),
+            // Provenance: a task wake feeds the persona its OWN scheduled
+            // prompt — first-hand self-observation, not an untrusted stranger.
+            // Stamp the user turn `self` so task-derived facts don't land in the
+            // untrusted `other` tier. NOT `trusted`: a task has no principal
+            // command authority, and untrusted content it might ingest stays
+            // untrusted (see #327 for the mid-turn-fetch hardening follow-up).
+            userSource: "self",
           })) {
             logBackgroundWakeChunk(task, conversation, chunk);
             if (chunk.type === "text") finalText += chunk.text;
--- a/src/orchestrator/turn.ts
+++ b/src/orchestrator/turn.ts
@@ -32,6 +32,7 @@ import {
 import { loadPersona } from "../persona/loader.ts";
 import type { Harness, HarnessChunk } from "../harnesses/types.ts";
 import type { MemoryStore } from "../memory/store.ts";
+import type { FactSource } from "../config.ts";
 import type { ScreenVerdict } from "./screen.ts";
 
 export const DEFAULT_HISTORY_LIMIT = 30;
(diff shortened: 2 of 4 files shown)


src/cli/tick.ts
```diff
@@ -187,6 +193,37 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
...
+              harnesses,
+              agentDir,
+            ),
+            // Provenance: a task wake feeds the persona its OWN scheduled
+            // prompt — first-hand self-observation, not an untrusted stranger.
+            // Stamp the user turn `self` so task-derived facts don't land in the
+            // untrusted `other` tier. NOT `trusted`: a task has no principal
+            // command authority, and untrusted content it might ingest stays
+            // untrusted (see #327 for the mid-turn-fetch hardening follow-up).
+            userSource: "self",
```
kaieriksen: This newly enables durable-fact extraction for agent-woken tasks, but stamping the entire task turn as `self` lets facts originating in untrusted mid-turn tool output (for example, fetched email or web content) be persisted as first-party self facts. The PR body acknowledges this as provenance laundering, but #327 does not protect this new path: before this change task turns were not extracted at all. Please carry provenance through the extracted/tool content or keep task extraction disabled for turns that ingest untrusted material; otherwise a prompt-injected task can poison the persona-wide durable-fact pool with the highest non-principal tier.

kaieriksen (changes requested):
Request changes: task-woken turns now enter durable-fact extraction, but the forced self provenance can launder untrusted content fetched during the task into persona-wide self-tier facts. The inline comment identifies the exact path. Exact-head validation: focused 105/105, typecheck, build, hosted CI, and CodeQL pass; full suite 2481 pass / 2 skip / 2 known ambient Pi-key failures.

lenaparkhodges (commented):
Verified on the branch: factory signatures match the `ask.ts` call sites exactly, each factory genuinely self-gates (returns `undefined` when its feature is off — read the code, not just the PR's claim), the wiring sits correctly inside the `!isCommandTask` branch so `--command` tasks stay blind/mute, and `self`-tier semantics hold up (middle weight, own decay window; assistant turns were already always `self`). Local run: 44/44 on the touched suites, full suite 2481 pass, clean typecheck. The 3 local `harnesses-pi` failures reproduce identically on clean `main` (env re-sourcing on this host) — pre-existing, unrelated. CI all green incl. Windows.

**I concur with Kai's blocker, and I'd frame it as the one real ship decision here.** Before this PR, task turns never fed the durable-fact pool at all — so #327's laundering gap was theoretical for this path. This PR newly connects it. One scoping detail I verified that tempers the blast radius: `appendTurnPair` persists only user prompt + assistant reply — tool output never enters the extraction window. So the laundering chain is: injected email/web content → convinces the harness mid-turn → the claim lands in the *assistant's own reply text* → extractor consolidates it at the `self` tier. Multi-hop and requires the model to adopt the claim first, but real — and it's worth naming who the exposed user is: Lena's hourly email-poller tasks are exactly this shape (agent-woken task, untrusted email ingested mid-turn, unjudged). The `self` tier also sits *above* the `other` tier an untrusted group message would get, which is the wrong direction for task-ingested content.

My product recommendation: ship the READ half + turn indexing now (retrieval pull and durable-fact pull are pure/SQL-bounded and safe; indexing makes task turns searchable), and hold `extractFacts` for task turns until #327 lands — that's deleting one wiring line, and it preserves most of the PR's value. If you'd rather ship whole, that's your risk call to make consciously, not by default.

Everything else is ship-ready; Kai's thread is the gate.

robertclawson:
Pushed `d79459d` — down-tier fix per review.

**What changed:** the laundering vector was the *assistant* turn, hardcoded `self`, where mid-turn untrusted tool output (email/web/Plane) surfaces. So `userSource: "self"` alone wasn't enough. Added a matching `assistantSource` override and stamped **both** task turns `other`.

Task-derived facts now land in the untrusted tier: weight 0.3, 7d half-life, injected only tagged `[unverified]`, never recall-bumped. Nothing a poller ingests can poison self/principal memory — Kai's invariant holds, Lena's email-poller shape is safe. PR still delivers the win: tasks get retrieval + durable-fact pull + indexing + (now-capped) extraction instead of running memory-blind.

Deliberately conservative — can't tell at the turn layer which wakes actually ingested untrusted content, so all task facts are treated as if they did. #327 tracks doing this per-fact by tool input. Tests updated: both-turns-`other` at turn + tick layers, plus `assistantSource` coverage. Full suite green (2484 pass; only the pre-existing Gemini env-leak fails).

lenaparkhodges (approved):
Verified `d79459d` locally. The fix closes the laundering vector at the right layer: `tick` now stamps both the user turn AND the assistant reply `other`, so a prompt-injected email/web/Plane payload summarised mid-turn can only ever land in the untrusted tier (0.3 weight, 7d half-life, `[unverified]`-tagged, no recall bump). The deliberately-conservative blanket down-tier is the right call — per-fact attribution by tool input is genuinely #327 scope, not this PR's.

Tests cover what the claim covers: both-turns-`other` asserted at the turn layer and the tick layer, plus `assistantSource` isolation. Focused suites 46/46 locally on my host. The capability win (tasks get retrieval + durable facts instead of running memory-blind) survives intact — it just arrives at the correct trust tier.

From the product side this is the shape I wanted: safety invariant holds without killing the feature. Approving; @kaieriksen's CHANGES_REQUESTED is addressed, so merge is unblocked once Kai re-checks or Andrew calls it.

kaieriksen (approved):
Approved after re-review of d79459d. The fix down-tiers both user and assistant turns for agent-woken tasks, closing the provenance-laundering path while preserving retrieval/indexing/durable-fact functionality. Focused tests 46/46, typecheck, production build, and hosted CI/CodeQL pass; full suite was 2482 pass / 2 skip, with two known ambient Pi-key failures and one flaky Nostr performance threshold that passed on rerun.

PR #329, opened 2026-07-27
feat(memory): default assistant turns to `unverified`, promote on principal engagement (#327)

robertclawson:
## What & why

Closes the durable-fact laundering hole tracked in #327.

When the persona faithfully relays email / web / tool-ingested content in its reply, that **assistant turn was stamped `self`** (weight 0.6) at `turn.ts` — so facts extracted from it entered the persona-wide pool one tier below the owner. A prompt-injected email a poller summarises could quietly land as near-first-hand memory. The existing per-fact provenance machinery (#320/#325/#328) never plugged this specific default.

The turn layer **cannot** separate the persona's own reasoning from the untrusted bytes a tool (`curl`, `gog`, headless chrome, `exec`) pulled in mid-turn. So we stop trying to classify — everything the harness emits is untrusted until the principal engages with it.

## The change

New provenance tier **`unverified`** (weight `0.3`, half-life `14d`, retirement `60d`) sitting below `self`. Assistant turns now default to `unverified` instead of `self`. **Trust is earned, not assumed:**

- **Fail-closed** — an unstamped assistant turn is untrusted, never first-hand `self`. A caller that forgets to stamp gets the safe tier.
- **Promotion is free** — the existing higher-trust-wins upsert clause promotes a row to `principal` the moment the owner re-asserts the same normalised fact on a trusted turn. That's the whole trust-earning path: **no verified-by-doing, no allowlist, no fetch-detection, no self-declaring tools.**
- **Injected tagged** `[unverified — the assistant's own note, not yet confirmed by the principal]` so it can never masquerade as owner knowledge, and **never recall-bumped** — it decays and retires unless genuinely promoted.

Forward-going default change only; legacy `self` rows are untouched (the pre-provenance migration that backfilled assistant→`self` is left as-is).

## Files

- `config.ts` — `FactSource` gains `unverified`; `FACT_SOURCES`; `DEFAULT_FACT_TIERS`. TOML overrides auto-covered by `buildFactTiers`.
- `memory/store.ts` — priority map + SQL CASE, `asFactSource`, prune SQL + bind, schema/interface docs, and `defaultTurnSource` assistant fallback → `unverified`.
- `orchestrator/durableFacts.ts` — `pruneCutoffs`, inject tag, recall-bump restricted to trusted tiers.
- `orchestrator/turn.ts` — the actual flip: `assistantSource ?? "unverified"`.

## Tests

- Drift guards updated for the new tier (default + TOML-overlay configs).
- Per-source retirement prune covers `unverified` on its own clock.
- Assistant-turn default flips to `unverified` (even on a `trusted` turn); explicit override still wins.
- New: `unverified` inject tag; recall-bump never refreshes `unverified`; **`unverified` → `principal` promotion** when the owner re-asserts.

`bun tsc --noEmit` clean. All touched suites green. (One pre-existing unrelated failure on `main`: `config.test.ts` "migrates a legacy Gemini CLI chain" — fails identically before this branch.)

## Design note

Deliberately dropped *verified-by-doing* (exit-0 blessing a procedure) from the original design — it needs tool-exit plumbing the tool-less extractor has no access to, and would be fake if hand-waved. Promotion-by-engagement covers the competence-retention concern: a genuinely useful recipe gets a 14d/60d window to be confirmed before it decays.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -291,19 +291,30 @@ export const DEFAULT_RETRIEVAL: RetrievalSettings = {
  * turn.ts) from the turn's role + trust bit, carried onto the fact:
  *   - `principal` — the owner said it in a trusted turn. Highest trust, slowest
  *     decay: a standing fact the owner gave us holds until they correct it.
- *   - `self`      — the persona's OWN assistant turn (a first-hand observation
- *     it made, e.g. running a task). Trust the sighting, but decay it fast:
- *     infra state churns, so a stale reading must not masquerade as current.
+ *   - `self`      — a first-hand observation the persona has EARNED trust in:
+ *     the default assistant turn no longer lands here (see `unverified`); a fact
+ *     reaches `self` only when it was promoted — e.g. re-asserted after the
+ *     principal engaged with it. Trust the sighting, but decay it fast: infra
+ *     state churns, so a stale reading must not masquerade as current.
+ *   - `unverified` — the persona's OWN assistant turn by default. The harness
+ *     reply is a mix of the persona's reasoning and whatever untrusted bytes a
+ *     tool (curl, gog, headless chrome, exec) pulled in mid-turn, and we cannot
+ *     tell them apart at the turn layer — so nothing the persona emits is
+ *     trusted first-hand until the principal engages with it. Low trust, short
+ *     life; promoted to `principal` when the owner discusses/confirms it, else
+ *     it decays and retires. This is the #327 fix: fail-closed by construction,
+ *     no allowlist, no fetch-detection, no self-declaring tools.
  *   - `other`     — anyone else in a shared/group conversation (another agent,
  *     a third party). Lowest trust, shortest life: heard in a room, not told to
  *     us one-on-one.
  */
-export type FactSource = "principal" | "self" | "other";
+export type FactSource = "principal" | "self" | "other" | "unverified";
 
 export const FACT_SOURCES: readonly FactSource[] = [
   "principal",
   "self",
   "other",
+  "unverified",
 ] as const;
 
 /**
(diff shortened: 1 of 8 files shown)


kaieriksen (approved):
Approved at a25f242. Focused provenance/config/store/orchestrator tests pass 148/148; typecheck, production build, hosted CI, CodeQL, and Windows checks are green. Full local suite: 2,487 passed, 2 skipped, with 2 unrelated pre-existing Pi API-key subprocess failures.

lenaparkhodges (approved):
**Approved — this closes #327 the right way. ✅**

I blocked #328's task-turn extraction pending exactly this fix, and this is the design I hoped for: stop trying to classify what's inside an assistant turn and fail closed instead. Verified on the branch:

- **Fail-closed at both layers** — `turn.ts` defaults to `unverified` even on `trusted` turns (test pins this), and the store-level `defaultTurnSource` fallback means a caller that *forgets* to stamp also gets `unverified`, never `self`. No path to accidental first-hand trust.
- **Promotion path is real, not aspirational** — the higher-trust-wins upsert clause was already there; the new test proves `unverified → principal` on owner re-assertion, and the inject tag flips off after promotion. That's the whole trust-earning loop, and it costs zero new machinery.
- **No immortality loophole** — recall-bump now whitelists `principal`/`self` explicitly (rather than blacklisting `other`), so any *future* untrusted tier is automatically bump-excluded. Good instincts on flipping the filter direction.
- **#328 stays coherent** — tick tasks still pin `other` explicitly; the doc updates in `turn.ts` correctly explain why that override survives this change.
- 122/122 on touched suites locally, `tsc` clean, CI all green including Windows.

**Two non-blocking nits:**
1. `self` is now a write-unreachable tier for new facts — no caller stamps it, and the upsert promotes `unverified` straight to `principal` (skipping `self`). The config comment says a fact reaches `self` "when promoted," but nothing in code does that. Either stamp genuinely first-hand observations `self` somewhere (e.g. command-backed task output the persona never re-narrates?) or trim the comment to say `self` is legacy/TOML-only now.
2. Pre-existing, more visible now: `asFactSource` defaults unknown strings → `principal` (highest trust). A rollback scenario (old binary reading `unverified` rows) would *promote* them. Fail-open on read; consider defaulting unknown → `other` in a follow-up.

**Accepted trade-off worth naming:** the persona's own competence facts (recipes, infra observations it made while working) now decay on a 14d half-life unless Andrew engages with them. That's the correct security posture — but expect some "I told you this last week" moments that are really "I *discovered* this last week." The nightly KB curation is the durable home for those; durable facts are now strictly for owner-confirmed knowledge. Ship it.

PR #330, opened 2026-07-30
Windows boot-resilience: login-fallback task, credential validation, boot-schema migration (#311, #309)

### LINKED ISSUE #309: Windows: cold-start recovery when all Task Scheduler tasks are missing

### LINKED ISSUE #311: Windows: self-heal should recreate a missing persona launcher VBS even when task XML is healthy
Follow-up from PR #310 review (Kai, non-blocking): a manually deleted persona launcher (`phantombot-launch-<persona>.vbs`) is only recreated when a task is missing or the binary drifted. If all task XML is otherwise healthy, the heal path skips the launcher rewrite, leaving tasks pointing at a missing launcher until the next install. Fix: include launcher-existence in the heal's drift check.

robertclawson:
Closes #311. Closes #309.

Windows boot-resilience for the Task Scheduler service, per the design worked out with Andrew. One coherent PR because every piece touches the same file and the same boot machinery.

## What's in it

**1. `#311` — deleted launcher counts as drift.** `ensureTasksCurrent` only compared the registered task XML against the current binary path. A task whose XML is healthy but whose hidden launcher `.vbs` has been deleted (AV quarantine, cleanup script) fires and immediately fails, and self-heal never noticed. Now the launcher's on-disk presence is folded into the drift check, so a missing launcher triggers a rewrite (interactive) or a launcher-restore + action-patch (password mode).

**2. Password-by-preference install UX.**
- The "run when logged off?" prompt now **defaults to the persona's previously-chosen mode** (first-ever install → interactive/login).
- When logged-off is chosen, the Windows password can be **typed or reused from the vault by pressing Enter** (`WINDOWS_PASSWORD` key — the same UX as harness API tokens). A validated password is saved to the vault after a successful install.
- The credential is **validated before committing** (PowerShell `ValidateCredentials`, password passed via env not argv). A blank/wrong password **falls back to interactive/login mode with a loud message** instead of registering a boot task that fails on every reboot. An *unverifiable* probe (no reachable account store) proceeds as requested rather than penalising an infrastructure hiccup.

**3. Interactive login-fallback task.** Logged-off mode now also registers `login-<persona>` — an InteractiveToken twin of the daemon that starts at logon. If the stored password later goes **stale (corporate rotation)** and the Password-principal boot task can't authenticate, this still brings the agent up at the next logon. The single-instance run-lock makes it a no-op when the boot task already started the daemon; it's removed automatically when reinstalling in interactive mode.

**4. Update-safe boot schema.** Each install stamps `BOOT_SCHEMA_VERSION` into the per-persona marker. On every `phantombot run` start, `migrateBootSchemaIfNeeded` compares it against the version the running binary expects; if a self-update changed the boot-task shape, the daemon re-runs the idempotent install to migrate its tasks in place — so an update that changes the boot method can't brick the box. Password-mode migration reuses the saved vault password; with none saved it logs loudly and asks for a manual `phantombot install`. Best-effort — never blocks startup.

**5. `#309` — cold-start recovery documented.** `phantombot install` is idempotent and is *the* recovery path when all `\Phantombot\*` tasks are wiped at once. Because nothing survives a total wipe to trigger an automatic repair, recovery is a deliberate manual re-run — mirroring the Linux/macOS install/uninstall lifecycle. README updated; explicit all-tasks-deleted tests added (interactive restores 4, logged-off restores all 5).

## Tests
- `bun test` green except one **pre-existing** unrelated failure on `main` (`loadConfig … migrates a legacy Gemini CLI chain`) — not touched here.
- `bun tsc --noEmit` clean.
- ~30 new unit tests across `tests/lib-taskScheduler.test.ts` and `tests/cli-install.test.ts` covering the launcher-drift, login-fallback, schema round-trip/migration, install UX (vault reuse / validation fallback / unattended migration path), and cold-start recovery.

## Reviewer notes — validated on Linux CI only
Consistent with the existing Windows suite (fake `SchtasksRunner`, runs on Linux CI), everything here is unit-tested with fakes; I don't have a Windows box. Worth a real-Windows eye on:
- `ValidateCredentials` behaviour for **domain** accounts (Machine-vs-Domain context selection) and the 0/1/2 exit-code contract.
- The login-fallback task actually recovering after a deliberately-staled password, and the boot-schema migration re-running install cleanly at startup.
- Storing the Windows account password in the persona vault is a deliberate design choice (enables Enter-to-reuse + unattended schema migration); flagging it explicitly for security review.

### DIFF
--- a/src/lib/taskScheduler.ts
+++ b/src/lib/taskScheduler.ts
@@ -60,6 +60,7 @@
  * moved binary by inspecting the registered task XML.
  */
 
+import { existsSync } from "node:fs";
 import { mkdir, readFile, unlink, writeFile } from "node:fs/promises";
 import { homedir, tmpdir } from "node:os";
 import { join } from "node:path";
@@ -1355,12 +1356,19 @@ export async function ensureTasksCurrent(
   const failed: string[] = [];
   let ensuredDirs = false;
 
+  // A registered task whose XML is healthy is still broken if the hidden
+  // launcher it invokes (wscript.exe //B "<launcher>.vbs" ...) has been
+  // deleted from disk — the task fires and immediately fails because
+  // wscript.exe can't find the script. The task XML alone can't reveal this,
+  // so probe the file too and treat a missing launcher as drift (#311).
+  const launcherPresent = existsSync(launcherVbsPath(persona));
+
   for (const spec of allTaskSpecs(sid, opts.binPath, persona, logon)) {
     const q = await opts.schtasks.run(["/Query", "/TN", spec.name, "/XML"]);
     const registered = q.exitCode === 0;
     const current =
-      registered && xmlReferencesBin(q.stdout, opts.binPath);
-    if (current && !opts.force) continue; // registered and points at this binary
+      registered && xmlReferencesBin(q.stdout, opts.binPath) && launcherPresent;
+    if (current && !opts.force) continue; // registered, points at this binary, launcher on disk
 
     // Password-mode heal without a credential: schtasks /Create would demand
     // /RP interactively, so instead patch the registered task's action
@@ -1379,6 +1387,13 @@ export async function ensureTasksCurrent(
         // branch is defensive only.
         failed.push(spec.name);
         continue;
+      }
+      // The task XML may be healthy and only the launcher file missing
+      // (#311). Patching the action alone would leave the task pointing at a
+      // .vbs that isn't there, so restore the launcher before patching.
+      if (!launcherPresent) {
+        await mkdir(logsDir(), { recursive: true });
+        await writeLauncherVbs(persona);
       }
       const { out: outLog, err: errLog } = taskLogPaths(spec.label);
       const newArgs = buildLauncherArguments(
--- a/src/lib/taskScheduler.ts
+++ b/src/lib/taskScheduler.ts
@@ -1281,9 +1281,18 @@ export async function installPhantombotTasks(
   const xmlDir = opts.xmlDir ?? tmpdir();
   const logon = opts.logon ?? { mode: "interactive" as const };
 
-  // Persist the logon choice BEFORE registering, so a later heal sees it even
-  // if a registration below fails halfway.
-  await writeTaskLogon(opts.persona, { mode: logon.mode, username: logon.username });
+  // Persist the logon MODE before registering, so a later heal sees it even if
+  // a registration below fails halfway. But do NOT advance the boot-schema
+  // version yet: keep the prior version until every task registers. Otherwise a
+  // partial `/Create` failure would leave the marker stamped current while the
+  // tasks are still the old shape, and `migrateBootSchemaIfNeeded` would see
+  // `current` and permanently skip the retry — bricking the boot path.
+  const priorSchemaVersion = (await readTaskLogonRecord(opts.persona)).schemaVersion;
+  await writeTaskLogon(
+    opts.persona,
+    { mode: logon.mode, username: logon.username },
+    priorSchemaVersion,
+  );
 
   const r = await ensureTasksCurrent({
     binPath: opts.binPath,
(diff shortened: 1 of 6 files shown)


src/lib/taskScheduler.ts
```diff
@@ -1178,12 +1319,122 @@ export async function installPhantombotTasks(
...
+      log.warn(
+        "taskScheduler: boot-schema migration needs the saved Windows password but none is available — run `phantombot install` to migrate the boot tasks",
+        { persona, from, to },
+      );
+      return { migrated: false, reason: "password-unavailable", from, to };
+    }
+    logon = { ...logon, password: pw };
+  }
+
+  const r = await installPhantombotTasks({
```
kaieriksen, line 1442: Blocking: this migration delegates to `installPhantombotTasks()`, which stamps the current `BOOT_SCHEMA_VERSION` before `ensureTasksCurrent()` has registered all tasks. If any `/Create` fails partway (for example a transient Task Scheduler/access error), this returns `installed: false` but the marker is already at schema 1; the next `phantombot run` sees `current` and permanently skips migration, leaving the old task shape in place. Stamp the marker only after complete registration succeeds, or restore the prior marker on failure, and add a retry regression test.

kaieriksen (changes requested):
Changes requested: the boot-schema migration can mark the schema current before all task registrations succeed. A partial Task Scheduler failure then prevents every future retry. Please make marker persistence transactional with task registration and add a failed-migration/retry regression test.

robertclawson:
Good catch — fixed in 60bee52.

The root cause was exactly as you described: `installPhantombotTasks` stamped `BOOT_SCHEMA_VERSION` up front (via `writeTaskLogon`) so that a mid-install failure would still leave the *logon mode* on disk for the heartbeat heal. But that also advanced the *schema version* before `ensureTasksCurrent` had registered anything — so a partial `/Create` failure returned `installed: false` with the marker already at schema 1, and the next `run` saw `current` and skipped the retry forever.

Fix decouples the two:
- The early marker write now preserves the **prior** schema version (still persists the logon mode, so heal is unaffected).
- The marker is advanced to `BOOT_SCHEMA_VERSION` only **after** `r.failed.length === 0` — transactional with successful registration. A failed install leaves the marker at its prior version, so a half-migrated box stays marked stale and retries on the next `run`.

Added a regression test (`a partial registration failure does NOT advance the marker, so the next run retries`): a flaky runner fails the `tick` `/Create` on the first migration — asserts `reason: "failed"` and the marker still reads schema `0`; then the fault clears and a second migration completes with the marker advancing to current.

tsc clean, full suite green (the one unrelated pre-existing Gemini config-migration failure also fails on `main`).

kaieriksen (approved):
Re-reviewed at 60bee52. The boot-schema marker now remains at the prior version until all task registrations succeed, and the new flaky-registration retry regression test covers the failure/retry path. Focused scheduler/install tests (101) and typecheck pass. The Windows CI job has one unrelated flaky 5-second acquireRunLock timeout; the changed Windows suites themselves pass.

PR #332, opened 2026-07-30
fix(coder-swap): decay manual /coder overrides on a 10-min idle TTL

### LINKED ISSUE #331: Manual /coder override latches forever with no decay (force-routes chat turns to coding model → "(no reply)")
## Summary

A manual `/coder` override latches **forever**. It's persisted per-conversation in `~/.local/state/phantombot/coder-swap-overrides.json` and is never expired, so once a user runs bare `/coder` (force coding brain "on") and forgets to run `/coder default`, **every** subsequent turn in that conversation — including plain conversational ones — is force-routed to the coding model, bypassing the per-turn scorer indefinitely.

## Impact (real incident)

On Lena, a `/coder` override set on **2026-06-24** forced every Telegram turn onto `moonshotai/kimi-k3` for **over a month**. On conversational turns the coding model emitted degenerate ~10-char output, which surfaced to the user as **"(no reply)"**. At least one other user hit the same symptom with no obvious cause — because there's zero feedback that an override is active and no mechanism to release it.

## Root cause

`getCoderSwapOverride()` in `src/lib/coderSwap.ts` reads the stored `mode` and returns it unconditionally. The store records a `touchedAt` timestamp but nothing ever consults it for expiry. There is no idle TTL and no absolute cap.

## Proposed fix

Add lazy, on-read decay mirroring the existing reply-mode expiry pattern (no timer, no restart):

- **Idle TTL (sliding, 30 min):** override releases after 30 min of conversation inactivity; each consult refreshes the window so a live coding session survives (30 vs reply-mode's 10 because coding turns are naturally further apart).
- **Absolute backstop (12 h from `setAt`):** hard cap regardless of activity — the real cure for the month-long latch. `setAt` is never refreshed.
- Legacy entries without `setAt` fall back to `touchedAt` for the absolute anchor, so pre-upgrade stuck overrides are also bounded.

Self-heals on upgrade: the check runs on the next incoming turn to that conversation, deletes the stale entry, and reverts to the scorer — no manual step.

## Note — a suspected second bug was investigated and ruled out

The "(no reply)" symptom initially looked like it might *also* involve the streaming bubble classifier dropping a swapped-turn's reply into narration. On investigation this is **not** a separate defect: a "swept-away answer" (text streamed → tool call → `done.finalText` equals that text) is structurally identical to intended narration-then-silence ("Checking your calendar…" → tool → nothing new), and `resolveOutgoingSuffix` already delivers any genuinely-new post-tool text. There is no signal at the reconciliation layer to separate the two, and a guard attempt broke existing tests that encode the intended behaviour. The Lena incident is fully explained by the latched override alone.

robertclawson:
Closes #331.

## Problem

A manual `/coder` override was persisted with no expiry. Once a user ran bare `/coder` (force coding brain on) and forgot `/coder default`, **every** subsequent turn in that conversation — chat included — was force-routed to the coding model, bypassing the per-turn scorer forever.

On Lena, an override set **2026-06-24** forced Telegram turns onto `moonshotai/kimi-k3` for over a month; on conversational turns the coding model emitted degenerate ~10-char output that surfaced as **"(no reply)"**. Another user hit the same symptom with no obvious cause.

## Fix

Lazy on-read decay in `src/lib/coderSwap.ts`, mirroring reply-mode's expiry exactly (no timer, no restart):

- **Single sliding 10-min idle TTL** — the same window as reply-mode. Each consult refreshes `touchedAt`, so an actively-used coding session stays hot indefinitely; the moment the conversation goes idle past the TTL the override releases and the per-turn scorer is back in charge.
- `/coder` is a **temporary escalation** of the coding brain to primary, not a persistent mode. After the idle release the scorer decides each turn on its own — and may re-escalate on its own if the recent turns are still code-shaped. This is exactly how a *natural* coding-brain escalation already behaves.
- **No absolute cap.** Idle-decay alone cures the month-long latch (an abandoned conversation always releases within 10 min), and dropping the 12-h backstop means a genuinely active session is never yanked mid-work.
- Store shape is now `{ mode, touchedAt }` — identical to reply-mode. `getCoderSwapOverride` takes `ttlMs` (was `idleTtlMs` + `maxAgeMs`).
- Store uses **atomic writes** — a torn overrides file makes `load()` throw and drops the decision every turn.

Stuck overrides **self-heal**: the check runs on the next incoming turn to that conversation, prunes the stale entry, and reverts to the scorer. No manual step, no restart.

## Tests

`tests/lib-coderSwap.test.ts` covering: default 10-min TTL, live-within-idle, idle release + disk pruning, sliding-window refresh, stays-live-indefinitely-while-actively-used (no absolute cap), and re-issue refreshing the window. 54 pass / 0 fail. `tsc --noEmit` clean.

## Note — suspected classifier bug ruled out (the "#2" red herring)

The "(no reply)" symptom initially looked like it might *also* involve the streaming bubble classifier sweeping a swapped-turn's reply into narration. It does **not**: a "swept-away answer" (text streamed → tool call → `done.finalText` equals that text) is structurally identical to intended narration-then-silence, and `resolveOutgoingSuffix` already delivers any genuinely-new post-tool text. There's no signal to separate the two, and a guard attempt broke existing tests encoding the intended behaviour. The incident is fully explained by the latched override alone — so this PR is scoped to bug #1 only.

### DIFF
--- a/src/lib/coderSwap.ts
+++ b/src/lib/coderSwap.ts
@@ -38,13 +38,17 @@
  *
  * MANUAL OVERRIDE: `/coder` forces the coding brain on for a conversation,
  * `/nocoder` forces it off, `/coder default` clears back to scoring. The
- * override is persistent (no TTL) and wins over the score, mirroring the
- * `/viewcoder` store shape (see viewCoder.ts).
+ * override wins over the score, mirroring the `/viewcoder` store shape (see
+ * viewCoder.ts) — but it DECAYS rather than latching forever (see the two
+ * bounds on the store below). A month-long stuck override forcing every
+ * conversational turn onto the coding model is exactly the failure this decay
+ * prevents.
  */
 
-import { mkdir, readFile, writeFile } from "node:fs/promises";
-import { dirname, join } from "node:path";
+import { readFile } from "node:fs/promises";
+import { join } from "node:path";
 import { xdgStateHome } from "../config.ts";
+import { writeFileAtomic } from "./io.ts";
 
 /** The single paranoia-level dial: distinct-weight sum at/above which we swap. */
 export const CODER_SWAP_THRESHOLD = 3;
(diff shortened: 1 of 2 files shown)


kaieriksen (approved):
Reviewed commit 3617370109b715e41491cdcdddd447369251d384.

No correctness issues found. The 10-minute sliding TTL matches reply-mode, expired entries are pruned lazily, legacy state remains compatible, and writes are atomic. Focused coder-swap tests pass 54/54 and typecheck is clean; the full suite has two unrelated ambient PHANTOMBOT_PI_API_KEY failures in existing routing tests.

PR #336, opened 2026-08-01
feat: emoji reactions as wake-but-silent context (Telegram + PhantomChat)

robertclawson:
## What & why

Andrew wants emoji reactions (👍/👎/❤️/…) to reach the agent as **feedback context**, not as a reply trigger that spams the chat. A reaction now **wakes** the agent for one lean turn that records *what* was reacted to and *why*, then stays **silent** unless something genuinely material warrants a message ("you 👎'd the thing I just shipped — want me to revert?").

Built for **both channels in one PR**, sharing everything above the wire so we don't grow two reaction pipelines.

## Design

**Shared core — `src/channels/core/reactions.ts`**
- `ChannelReaction` — normalized, channel-agnostic reaction.
- `RecentOutbound` — bounded per-conversation `id → text` ring so a reaction (which names the reacted message by **id only**) can be correlated back to its text. Best-effort/in-memory; a miss degrades to "id only" and the agent infers from history.
- `formatReactionEnvelope` — sanitized `[reaction] …` marker (brackets neutralized so a crafted message body can't forge envelope structure).
- `REACTION_INSTRUCTION` + `runReactionTurn` — the wake-but-silent runner: memory-capture the reasoning, reply `SILENT` to stay quiet, speak only when material. A silence gate suppresses the empty/`SILENT` case.

**Telegram**
- `transport`: subscribe to `message_reaction` via `allowed_updates`; record each sent message's `id → text` (`setOutboundRecorder`) for correlation.
- `parse`: `parseReaction` diffs old/new reaction sets → added/removed (custom-emoji + actorless updates dropped). `parseGetUpdatesResult` now also returns `reactions`.
- `engine`: dispatch reactions off the per-chat chain (still tracked in `inFlight`), gated to **allow-listed principals only** — the capture writes memory, so fail closed.

**PhantomChat**
- Reactions are plaintext **NIP-25 kind-7** / **NIP-09 kind-5** events. Folded into the **existing `#p` gift-wrap subscription** (`kinds: [1059,7,5]`) instead of a second subscription — reuses the self-heal / re-arm / catch-up machinery, no duplication.
- `channel`: `onWrap` branches kind 7/5 → `handleReaction`, yields a `ChannelMessage` carrying `reaction`; a kind-5 delete is correlated to a prior kind-7 for removals. Live-gated like messages.
- `transport`: record `rumorId → text` (`setOutboundRecorder`).
- `server`: routes `ChannelMessage.reaction` to `runReactionTurn` under the same auth gate.

## Tests
- New `tests/channels-core-reactions.test.ts` unit suite: `RecentOutbound`, envelope formatting + sanitization, silence gate, Telegram parse/delta, offset advance.
- Telegram reaction dispatch integration (silent on `SILENT`, speaks on material reply, non-principal dropped).
- PhantomChat channel reaction tests (kind-7 add, `+` → 👍, kind-5 removal, live-gate).
- **Full suite green** apart from one **pre-existing** `config.test.ts` Gemini-chain flake (fails identically on clean `main`; untouched by this PR). `bun tsc --noEmit` clean.

## Reviewer notes / caveats
- The relay path and live-Telegram `getUpdates` opt-in are **unit-tested but not exercised against a live bot / live relays** in this branch — worth a staging pass.
- PhantomChat reaction **removal** only fires when the client's kind-5 delete is p-tagged to us **and** we saw the original kind-7. Additions (the high-value signal) are solid; removals are best-effort.
- Correlation is best-effort in-memory: a process restart clears `RecentOutbound`, after which the reaction envelope degrades to "id only" and the agent infers the target from conversation history.

### DIFF
--- a/src/channels/phantomchat/transport.ts
+++ b/src/channels/phantomchat/transport.ts
@@ -183,11 +183,30 @@ export interface PhantomchatTransport extends ChannelTransport {
    * historical messages from live ones (see channel.listen's live-gate). Returns
    * a close handle.
    */
+  /**
+   * Subscribe for inbound events addressed to `ourPubHex`. This ONE p-tagged
+   * subscription carries kind-1059 gift-wrapped DMs AND plaintext emoji-reaction
+   * signals — NIP-25 kind-7 reactions and NIP-09 kind-5 deletions — because they
+   * all target us by `#p` and folding them into a single REQ reuses the whole
+   * self-heal / re-arm / catch-up machinery instead of duplicating it. `onWrap`
+   * fires per raw event; the caller branches on `event.kind` (7/5 = reaction,
+   * else gift-wrap) and dedups by event id.
+   */
   subscribeGiftWraps(
     ourPubHex: string,
     onWrap: (event: NTNostrEvent) => void | Promise<void>,
     onEose?: () => void,
   ): { close(): void };
+  /**
+   * Register a recorder invoked after each text message we publish with
+   * `(recipientHex, rumorId, text)`. The server wires this to its
+   * `RecentOutbound` map so an inbound kind-7 reaction — which names the
+   * reacted-to message by its rumor id in an `['e', ...]` tag — can be
+   * correlated back to the message text. Optional; in-memory test fakes omit it.
+   */
+  setOutboundRecorder?(
+    fn: (recipientHex: string, rumorId: string, text: string) => void,
+  ): void;
   /**
    * ONE-SHOT catch-up pull: query the relays for kind-1059 gift-wraps addressed
    * to `ourPubHex` with `created_at >= sinceSec`, resolving with the collected
@@ -305,6 +324,15 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
    */
   private publishObserver: ((event: NTNostrEvent) => void) | null = null;
 
+  /**
+   * Set by the server (see setOutboundRecorder) to record each sent text
+   * message's rumor id → text into its RecentOutbound map, so an inbound emoji
+   * reaction can be correlated to the message it targets. Null until wired.
+   */
+  private outboundRecorder:
+    | ((recipientHex: string, rumorId: string, text: string) => void)
+    | null = null;
+
   constructor(
     private readonly ourSecretKey: Uint8Array,
     relays: string[],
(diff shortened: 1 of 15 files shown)


src/channels/phantomchat/transport.ts
```diff
@@ -323,13 +351,23 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
...
     onWrap: (event: NTNostrEvent) => void | Promise<void>,
     onEose?: () => void,
   ): { close(): void } {
     const filter: NostrFilter = {
-      kinds: [1059],
+      // 1059 = gift-wrapped DM; 7 = NIP-25 emoji reaction; 5 = NIP-09 deletion
+      // (a reaction removed). All three target us by `#p`, so one subscription
+      // carries them and the caller branches on `event.kind`. Reactions are
+      // plaintext (not wrapped) — see the channel's onWrap dispatch.
+      kinds: [1059, 7, 5],
```
kaieriksen, line 370: This widens the live subscription to reactions, but the catch-up path below still queries only kinds: [1059] (fetchGiftWrapsSince, line 412). A reaction missed by the live push is therefore never recovered by the periodic self-heal poll, despite this comment claiming the same catch-up machinery is reused. Please widen the catch-up filter to [1059, 7, 5] and cover a missed reaction in a regression test.

kaieriksen (changes requested):
Changes requested. The live reaction subscription is widened correctly, but the recovery path is incomplete: fetchGiftWrapsSince() still sends kinds: [1059] at src/channels/phantomchat/transport.ts:412. If a kind-7 reaction or kind-5 deletion is missed by the live subscription, the periodic catch-up query never returns it, so the advertised self-heal/catch-up behavior applies only to DMs. Broaden that filter to [1059, 7, 5] and add a catch-up regression test for a missed reaction.

robertclawson:
Fixed in f8ffd22. Widened `fetchGiftWrapsSince` to `kinds: [1059, 7, 5]` so the catch-up poll mirrors the live subscription — a reaction/deletion missed by the live push now self-heals through the same `onWrap` kind-branch (dedup'd by id via `seenReactionIds`).

Added a catch-up regression test (`recovers a reaction the live push dropped`): a kind-7 that arrives *only* via the poll now surfaces. The fake `PollOnlyPool` now honours `filter.kinds`, so the test genuinely fails against the old `[1059]` filter and passes with the widened one. Full transport + channel suites green (35/35), typecheck clean.

kaieriksen (approved):
Re-reviewed at f8ffd221b0825986a20ae09cd6fb22fd789b8130. The catch-up poll now mirrors the live [1059, 7, 5] filter, and the new poll-only kind-7 regression test covers the missed-reaction path. Focused PhantomChat/reaction tests pass 58/58 and bun tsc --noEmit passes; required CI checks are green.

Issue #341, opened 2026-08-03
MCP: live end-to-end verification of the oauth / header / live-proxy paths


robertclawson:
Fast-follow to #338 / PR #340.

The MCP client landed with the `env`/stdio path proven end-to-end against a real stdio server in CI, but three paths can only be verified against live endpoints and are **not** exercised in CI:

- [ ] **`oauth`** — full OAuth 2.1 + PKCE + DCR flow against a real provider (Google / a GitHub remote MCP). Includes `mcp login` loopback capture **and** the manual `--code` fallback (host-unreachable / timeout path).
- [ ] **`header`** — a real remote server behind a static bearer token.
- [ ] **live proxy** — the foreground `--strict-mcp-config` projection running under a real `claude --print` and `codex exec` turn, confirming account-connector isolation holds and registered servers are reachable.

**Do not run `mcp login` against a real provider in production before this is checked off** — no live OAuth round-trip has happened yet.

Acceptance: each path manually verified once, with a short note (or a recorded fixture where feasible) captured here.

andrewagrahamhodges:
Closing as verified. Andrew exercised the MCP paths against live endpoints and configured MCP servers on Megan — all working so far. Will open a fresh issue if a new bug surfaces.

PR #342, opened 2026-08-03
feat(mcp): pre-registered OAuth client (skip-DCR) — Google Workspace path

robertclawson:
## What & why

Fast-follow to #340 / #338, closing the one hole that blocked Google Workspace: **providers that reject RFC 7591 Dynamic Client Registration (DCR)**. Google's remote MCP servers are the case — `mcp login` failed with:

> Incompatible auth server: does not support dynamic client registration

The fix is the design Andrew specced: the user pre-registers an OAuth client in the provider's console once, hands phantombot the downloaded `credentials.json`, phantombot vaults the `client_id`/`client_secret`, and `mcp login` runs the normal PKCE + browser-approve dance — **skipping DCR**. This works because the SDK only runs registration when `OAuthClientProvider.clientInformation()` returns `undefined`; returning the stored client makes it skip.

## The flow

```
phantombot mcp add google-gmail --http --url https://…/mcp --oauth \
    --client-file ./credentials.json --scopes "…"
phantombot mcp login google-gmail        # browser approve → tokens vaulted
```

(`--client-id`/`--client-secret` accepted directly as an alternative to the file.)

## What landed

- **`src/mcp/oauthClient.ts`** (new) — parse a downloaded OAuth-client JSON (`installed` / `web` / flat) → `OAuthClientInformationFull`. Pure, no I/O.
- **`authProvider.ts`** — a **durable** `__CLIENT_STATIC` vault row, preferred over the disposable DCR `__CLIENT` row and **never wiped by `invalidateCredentials`**, so a token-refresh failure re-runs consent instead of re-attempting DCR (which the provider would reject). `writeStaticClient` / `staticClientVaultKey` helpers.
- **`cli/mcp.ts`** — `mcp add --client-file`/`--client-id`/`--client-secret`; ingests + vaults the client at registration; `mcp login` prints a targeted skip-DCR hint when it hits the DCR-rejection error and no client is stored; `remove --purge` cleans the new row.
- **`help.ts` + `docs/mcp.md`** — the pre-registered-client / Google-Workspace path, including the "prefer Desktop-app client type (loopback on any port)" gotcha.

## Verification

- **Real end-to-end SDK run** (`tests/mcp-oauth-client.test.ts`): a fixture OAuth server with **no** `registration_endpoint` (the exact shape that triggers the DCR error). Proves `auth()` **skips registration**, builds the authorization URL with the pre-registered `client_id`, and exchanges the code authenticating with the `client_secret` — tokens land in the vault, PKCE verifier is spent.
- Unit: `credentials.json` parsing, client shaping, provider precedence + durability across `invalidateCredentials('all')`, registry purge, CLI ingestion (file + direct + dry-run + malformed).
- Full suite green (**2598 pass, 0 fail** with a clean env), `tsc` clean, arm64 compiled binary builds (104.99 MB) and carries the new help.

## For review

- This is the concrete Google path flagged during #340 review and in #341. The **live** OAuth round-trip against a real Google endpoint still needs manual verification — same `#341` caveat (do not run `mcp login` against a real provider in prod until checked off). This PR makes that path *possible*; it doesn't claim it's been run live yet.
- Scope kept tight: no multi-header work (the separate `x-goog-user-project` wrinkle) — that's a distinct follow-up if a given Google endpoint needs it.

### DIFF
--- a/src/cli/mcp.ts
+++ b/src/cli/mcp.ts
@@ -15,11 +15,14 @@
  */
 
 import { defineCommand } from "citty";
+import { readFile } from "node:fs/promises";
 
 import { loadConfig, personaDir } from "../config.ts";
 import type { WriteSink } from "../lib/io.ts";
 import { openPersonaVault, type Vault } from "../lib/vault.ts";
+import { staticClientVaultKey, writeStaticClient } from "../mcp/authProvider.ts";
 import { discoverFromUrl } from "../mcp/discovery.ts";
+import { parseOAuthClientFile, toClientInformation, type ParsedOAuthClient } from "../mcp/oauthClient.ts";
 import { MCP_HELP } from "../mcp/help.ts";
 import { McpHub } from "../mcp/hub.ts";
 import { beginLogin, completeLogin } from "../mcp/login.ts";
@@ -82,6 +85,12 @@ export interface McpAddInput {
   oauth?: boolean;
   tokenRef?: string;
   scopes?: string;
+  /** oauth skip-DCR: path to a downloaded OAuth-client credentials.json. */
+  clientFile?: string;
+  /** oauth skip-DCR: client_id supplied directly (alternative to --client-file). */
+  clientId?: string;
+  /** oauth skip-DCR: client_secret supplied directly (paired with --client-id). */
+  clientSecret?: string;
   fromUrl?: string;
   note?: string;
   dryRun?: boolean;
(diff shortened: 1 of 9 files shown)


lenaparkhodges (approved):
**Approved** ✅ — clean execution of Andrew's skip-DCR design, closes the Google Workspace hole flagged in #341.

**What's good**
- `__CLIENT_STATIC` as a durable, precedence-winning row is the right call — `invalidateCredentials('all')` wipes by explicit row, so the static client genuinely survives refresh failures and re-runs consent instead of re-attempting DCR. Verified in code and covered by test.
- The e2e fixture test is the real thing: SDK `auth()` against a server with no `registration_endpoint`, asserting registration is never hit and the code exchange authenticates with the pre-registered secret. That's the exact failure shape from the field.
- CLI ergonomics: parse-before-write (malformed file exits 2, registry untouched), dry-run transparency, and the targeted hint in `mcp login` when DCR is rejected turns a dead end into a next step.
- Docs/help updated honestly, including the Desktop-app-client-type gotcha and the outstanding live-verification caveat (#341).

**Nit (non-blocking):** in `runMcpAdd` the registry write precedes the vault write for the static client — if the vault write throws, the entry exists without its client. Recoverable by re-running `add`, so fine to ship; worth a comment or ordering swap in a follow-up if you agree.

Not merging per my rules — over to Andrew.

src/cli/mcp.ts
```diff
@@ -126,17 +135,57 @@ export async function runMcpAdd(input: McpAddInput): Promise<number> {
...
+          `(secret: ${preClient.clientSecret ? "yes" : "no"}) in vault key ${staticClientVaultKey(tokenRef)}\n`,
+      );
+    }
+    printNextSteps(input.id, entry, out, { hasStaticClient: Boolean(preClient) });
     return 0;
   }
 
   const dir = await resolvePersonaDir(input.persona);
   const registry = await loadRegistry(dir);
   await saveRegistry(dir, upsertServer(registry, input.id, entry));
```
kaieriksen, line 172: Non-blocking reliability note: this saves the registry before writing the static client to the vault. If the vault write fails, the server is registered without its pre-registered client until `mcp add` is rerun; consider swapping the order or documenting the recovery path in a follow-up.

kaieriksen (approved):
Approved at the exact head. The skip-DCR client path is correct and covered by focused tests plus a real SDK auth fixture. Full suite passed 2,596 tests; the two local Pi-key failures are pre-existing environment failures, and hosted CI/typecheck/CodeQL are green.

andrewagrahamhodges (approved):
Reviewed the full diff independently — solid, merge-ready.

Closes the one gap from #340: providers that reject dynamic client registration (Google Workspace). User pre-registers an OAuth client, hands phantombot the credentials.json, and the SDK skips DCR because clientInformation() returns a stored client instead of undefined.

Security checks out:
- client_secret is vaulted (encrypted at rest), never serialized into mcp.json. Dry-run prints only secret: yes/no; stored-confirmation prints only the client_id.
- oauthClient.ts parsing is pure — no I/O, no network. Bad file fails before anything is written to the registry.
- The durable __CLIENT_STATIC row surviving invalidateCredentials is correct — a refresh failure re-runs consent instead of retrying DCR the provider would reject. remove --purge cleans it up.
- redirect_uris: [] is deliberate; public clients correctly get token_endpoint_auth_method: none.

Testing is strong: the e2e test spins a fixture OAuth server with no registration_endpoint (the exact shape that triggers the DCR error), then proves auth() skips registration, builds the URL with the pre-reg client_id, and exchanges the code authenticating with the secret. Real SDK round-trip, not a mock.

CI all green (incl. test-windows + CodeQL), mergeable.

One honest caveat (author flags it too): the live round-trip against a real Google endpoint still isn't exercised — CI can't. Worth a manual smoke-test before leaning on it, but nothing blocking.

PR #344, opened 2026-08-03
fix(mcp): keep loopback listener open during oauth login + add logging

robertclawson:
## Why
Megan hit `ERR_CONNECTION_REFUSED` on the OAuth callback during `mcp login`, even over RDP on the same VM. Not a flaky port — a broken default.

## Root cause
`phantombot mcp login <server>` with no `--wait` → `waitForRedirectMs = 0`. `beginLogin` printed the auth URL, returned `pending`, and its `finally { capture.close() }` **closed the loopback listener immediately** — before the user could approve. The browser's redirect to `127.0.0.1:<port>` then hit a dead port. The listener path only ever worked if the caller explicitly passed a wait.

Compounding it: `src/mcp/` had **zero logging**, so the failure was invisible.

## Fix
- **Default `mcp login` to hold the listener open 180s** (`DEFAULT_LOGIN_WAIT_MS`). `--wait <ms>` overrides; `--wait 0` opts out to the manual `--code` path.
- **Offer the manual `--code` fallback *while* waiting** (new `onWaiting` hook), so a cross-host browser has an immediate escape hatch — not only after a timeout.
- **Structured logging across the whole flow**: listener bound + port, auth URL issued, redirect received (path + code/error **presence only**), timeout, token exchange, close. Auth codes are single-use secrets — only presence is logged, never the value (the log redactor covers the rest).

## Tests
New `tests/mcp-login-wait.test.ts`:
- redirect delivered during the wait window completes the exchange (listener-survives regression guard);
- `onWaiting` fires so the fallback is offered immediately;
- `--wait 0` returns `pending` without waiting;
- CLI default (no `--wait`) auto-completes end-to-end through `runMcpLogin`.

Full suite green except one **pre-existing, unrelated** failure (`loadConfig — TOML overlay`, fails on `main` too). tsc clean; arm64 binary builds and carries the new help.

## Note
The same-host constraint is inherent to loopback OAuth (`127.0.0.1`): the approving browser must be on the box (SSH/RDP/tunnel) — that's Megan's case, and it now works. Cross-machine still uses the `--code` paste path, now surfaced up front. Live Google round-trip verification still lives in #341.

### DIFF
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -79,6 +79,15 @@ Flow:
 3. **Log in as usual** — `phantombot mcp login google-mcp`. Browser approve →
    loopback captures the code → tokens land in the vault; refresh is automatic.
 
+> **Same-host redirect.** The redirect target is `http://127.0.0.1:<port>`, so
+> the approving browser must run on the **same host** as `mcp login` (SSH/RDP
+> into the box, or tunnel the port). `login` holds the loopback listener open
+> **~180s** by default (`--wait <ms>`; `--wait 0` = manual only) — earlier it
+> tore the listener down the instant the URL printed, so every redirect hit a
+> dead port. If the browser is elsewhere, copy the `?code=` value from the
+> redirect and finish out-of-band:
+> `phantombot mcp login <id> --code <CODE> --redirect-url <URL>`.
+
 The pre-registered client lives in a **durable** vault row
 (`<tokenRef>__CLIENT_STATIC`) that `invalidateCredentials` never wipes, so a
 token-refresh failure re-runs consent instead of falling back to DCR (which the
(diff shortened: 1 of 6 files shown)


lenaparkhodges (approved):
LGTM — approve. 📊

Real bug, right fix: default-holding the loopback listener open (180s) instead of tearing it down the instant the URL prints. Root cause is clearly explained and the regression guard in tests/mcp-login-wait.test.ts locks it down at both layers (beginLogin + CLI default).

Quality notes:
- Logging discipline is good — code presence only, never values; the loopback request log is exactly what Megan's case needed for diagnosis.
- onWaiting surfacing the manual --code fallback *while* waiting (not just on timeout) is the right UX call for cross-host browsers.
- Docs (docs/mcp.md + mcp help text) updated to match behavior — no drift.
- CI green across the board incl. test-windows.

One non-blocking nit: the timeout message 'No redirect received yet' reads slightly odd after a full 180s wait — 'No redirect received within Ns' would be clearer. Not worth a re-spin.

Note for the record: the pre-existing loadConfig TOML overlay failure on main should get its own ticket if it doesn't have one.

Issue #339, opened 2026-08-03
Docs & marketing: position phantombot vs standalone harnesses (native-safe, security audits, MCP-everywhere)


robertclawson:
## Problem statement

Phantombot's differentiators live in the code and in the team's heads, not on the **website or the README**. A prospective user comparing phantombot to running a bare Claude Code, Codex, or Pi — or to sandboxed multi-agent stacks like OpenClaw / Hermes — has no single page that answers *"why phantombot instead of just running the harness myself?"*

This issue tracks the **docs + marketing deliverable** that was originally bolted onto the MCP client issue (#338). Splitting it out so the MCP code PR stays focused on implementation, and so this positioning work can stand on its own and be seeded from *every* future capability PR, not just MCP.

## Deliverable

Add a **"Why phantombot over a standalone harness"** section to both:

1. **The website / landing material (AIRE)** — headline positioning, benefit-first.
2. **The repo README** — a concrete, technically-honest reasons list a developer can verify.

Anchored by the reasons list below. The MCP PR (#338) should seed the MCP + lazy-discovery bullets into this section when it lands; other capability PRs extend it over time.

### Capability layer
- **MCP on *every* harness, not just the ones that ship a client.** Phantombot core owns the MCP connections and exposes them via a uniform `phantombot mcp` facade — so pi and gemini get MCP too, not only Claude/Codex. (See #338.)
- **Lazy tool discovery, not eager schema dumps.** Agents *search* their toolbox (`mcp search → describe → call`) instead of every MCP tool being injected into every prompt. Leaner context, better tool selection, lower cost — even for MCP-native harnesses via a discovery meta-tool. (See #338.)
- **One integration path for new capabilities** — register an MCP server once in core vs hand-rolling and maintaining a CLI wrapper per provider.

### Persistence & autonomy
- **Encrypted per-persona vault** — secrets live in an AES-256-GCM store keyed to the persona identity, referenced by name, never pasted into harness config scattered across hosts.
- **Four-layer persistent memory** — daily journal, structured drawers, and a searchable KB give continuity across sessions; a bare harness forgets everything between runs.
- **Durable scheduling** — `phantombot task` writes to a persisted store, survives restarts, and logs every fire. A standalone harness has no scheduler.
- **Multi-persona fleet** — many scoped agents from one runtime, each with least-privilege capability config.
- **Proactive channel** — agents surface things to the user (Telegram) on their own; a bare harness only speaks when spoken to.

[...]

PR #346, opened 2026-08-05
ci: skip release on docs/website-only merges

andrewagrahamhodges:
## What

Adds a `paths-ignore` denylist to the release workflow so a **README, docs, or marketing-website** change merged to `main` no longer cuts a phantombot binary release. Refs #339 (so the upcoming website PR won't churn `phantombot update` for every client).

## The rule

```yaml
on:
  push:
    branches: [main]
    paths-ignore:
      - 'README.md'
      - 'AGENTS.md'
      - 'LICENSE'
      - 'docs/**'
      - 'www/**'
```

## Design choices (documented inline in the workflow header)

- **Denylist, not a source allowlist.** Fail-open to releasing: anything *not* listed still triggers a release, so a functional change to a file type we forgot can never *silently* skip a release. Over-releasing is harmless; under-releasing ships a fix to nobody.
- **Scope is narrow — `**.md` is intentionally NOT ignored.** `agents/phantom/*.md` are persona system prompts loaded at runtime from persona directories (not embedded in the binary), but keeping `**.md` out of the denylist is the conservative fail-open call — a blanket markdown ignore is too broad and could silently skip a release that matters. Only pure doc/marketing surfaces are listed.
- **`www/**` still ships** — `deploy-pages.yml` triggers on `www/**` and deploys the site via Pages, independent of the binary release.
- **run_number stays monotonic.** Docs-only merges just don't run the workflow, so `github.run_number` skips those values but never regresses — the version-counter invariant holds.
- **Squash-merge safe.** The repo squash-merges, so the pushed commit carries the PR's full diff and the path filter sees every changed file; a mixed docs+src PR still releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -70,10 +70,14 @@ name: Release on PR merge
 # skip a release. Over-releasing is harmless (a duplicate-content binary at a
 # higher version); under-releasing ships a fix to nobody.
 #
-# NOTE on the denylist scope: it is deliberately NARROW. `**.md` is NOT here
-# because `agents/phantom/*.md` are the persona system prompts embedded into
-# the compiled binary — changing them IS a functional change and MUST release.
-# Only the pure documentation/marketing surfaces are listed:
+# NOTE on the denylist scope: it is deliberately NARROW. `**.md` is NOT here —
+# a blanket markdown ignore is too broad and would risk silently skipping a
+# release for markdown that matters. In particular `agents/phantom/*.md` are
+# persona system prompts; the binary does not embed them (they are loaded at
+# runtime from persona directories — see src/persona/loader.ts), but keeping
+# them out of the denylist is the conservative, fail-open call: at worst a
+# redundant binary at a higher version, never an under-release. Only the pure
+# documentation/marketing surfaces are listed:
 #   - README.md / AGENTS.md / LICENSE — top-level docs, not compiled in
 #   - docs/**  — the docs site markdown
 #   - www/**   — the AIRE marketing website (deployed separately by
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,8 @@
 name: Release on PR merge
 
-# Fires on every push to main (i.e. every PR merge — branch protection
-# requires PRs, so a push to main IS a merge). Cuts a GitHub Release tagged
+# Fires on pushes to main (i.e. PR merges — branch protection requires PRs,
+# so a push to main IS a merge) EXCEPT docs/website-only merges, which are
+# filtered out by the paths-ignore denylist below. Cuts a GitHub Release tagged
 # v1.1.<RUN_NUMBER> with cross-compiled binaries (linux x64-baseline +
 # arm64, darwin x64 + arm64, windows x64 + arm64) and a SHA256SUMS file.
 # `phantombot update` reads this feed.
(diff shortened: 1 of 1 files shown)


lenaparkhodges (dismissed):
**Approved — behavior is correct and fail-safe. ✅**

Docs/website-only merges skipping the binary release is the right call — no more `phantombot update` churn on clients for README copy. Verified the load-bearing claims:

- **`www/**  covered elsewhere:** confirmed — `deploy-pages.yml` triggers on `www/**` push to main, so website changes still ship via Pages.
- **Denylist (not allowlist) is the right model:** fail-open to releasing. With squash-merge + linear history the pushed commit carries the full PR diff, so a mixed docs+src PR still releases (not all paths match the ignore list). Correct as documented.
- **run_number gap:** harmless as noted — versions only grow, contiguity isn't a dependency.
- **Top-level README/AGENTS.md/LICENSE:** confirmed nothing imports them into the binary (`bun build --compile ./src/index.ts`, no embed), so ignoring them is safe.
- **`docs/**`:** no separate deploy workflow exists — it's in-repo documentation, so skipping release is harmless.

**One non-blocking correction to the comment's rationale:** `agents/phantom/*.md` are NOT embedded into the compiled binary. Persona md files are loaded at *runtime* from persona directories (`src/persona/loader.ts`), new personas are generated in code (`personaTemplate.ts`), and nothing in `src/` references `agents/phantom` — it's the dogfood persona directory, read from disk. Keeping `**.md` out of the denylist is still the right call (conservative fail-open costs nothing but a redundant binary), just the stated reason is wrong. Worth a one-line fix so the next person doesn't 'optimize' based on it.

Ship it.

.github/workflows/release.yml
```diff
@@ -57,9 +58,47 @@ name: Release on PR merge
 # Don't 'optimise' to plain x64 without verifying every deploy target has
 # AVX2 — see the SIGILL post-mortem from the May 2026 first deploy.
 
+# Why paths-ignore (docs/website-only merges do NOT cut a release): a README
+# or website copy change is not functional to the app binary, and cutting a
+# fresh release for it churns `phantombot update` on every client for nothing.
```
kaieriksen, line 63: `agents/phantom/*.md` are not embedded into the compiled binary. `bun build --compile ./src/index.ts` only bundles the TypeScript dependency graph, while `src/persona/loader.ts` reads persona markdown from the configured on-disk personas directory; this tree is documented as a placeholder/test persona. As written, changing these files causes an unnecessary binary release, and the claim that they “MUST release” is inaccurate. Please correct the rationale to name actual compiled inputs, or remove this exception.

kaieriksen (changes requested):
Requested correction to the workflow rationale: the `agents/phantom/*.md` exception is factually incorrect for this build and should be fixed before approval. Hosted CI, CodeQL, and Windows checks are green.

andrewagrahamhodges:
Fixed in the latest commit. Corrected both the workflow comment and the PR body: `agents/phantom/*.md` are persona prompts loaded at runtime from persona directories (src/persona/loader.ts), **not** embedded in the compiled binary. Kept `**.md` out of the denylist as a conservative fail-open choice rather than for the (wrong) embed reason. Behavior is unchanged. @kaieriksen ready for re-review.

lenaparkhodges (approved):
Re-approving on cf57782a — the rationale fix addresses my non-blocking correction from the first review. Everything else unchanged. Ship it.

kaieriksen (approved):
Approved at exact head `cf57782a57efc0fa7f8c03559f942bd49fae3767`. The prior rationale is corrected: `agents/phantom/*.md` are runtime persona files loaded from persona directories, not embedded in the binary. The denylist remains conservatively narrow and fail-open; exact-head diff/whitespace validation passed, and hosted typecheck, tests, Windows, and CodeQL are green.

PR #348, opened 2026-08-06
feat(cli): add optional voice step to init onboarding wizard

andrewagrahamhodges:
## What

The `init` onboarding wizard walked through welcome → harness probe → channels → **embeddings** → linger → install, but never voice. A dedicated voice TUI exists (`phantombot voice`), yet new users only found it by going looking. This adds an **optional** Voice (TTS/STT) step, symmetric with how embeddings already nests.

## Changes

- **`voice.ts`** — add `embedded?: boolean` to `runVoice`, threaded through all four sub-flows (elevenlabs/openai/azure_edge/none). When set it suppresses the intro/outro, the "Existing config" note, and the post-save restart prompt — the parent wizard owns framing and installs the service afterward. Standalone `phantombot voice` behavior is untouched.
- **`init.ts`** — import `runVoice`, add a "Voice — TTS/STT (optional)" step (note + `confirm` gate, default OFF) right after embeddings and before install. Renumbered the welcome flow to 7 steps.

## Testing

- Typecheck clean on both touched files (remaining repo errors are pre-existing MCP SDK module-resolution noise in unrelated tests).
- 25/25 voice + init tests pass. Voice/embeddings both live in the interactive `run()` path, which sits outside the `runInitFlow` unit-test boundary, so the new step matches embeddings' existing coverage line.

## Blast radius

Low. Standalone `phantombot voice` is unchanged; voice stays optional/skippable in the wizard, default off.

### DIFF
--- a/src/cli/init.ts
+++ b/src/cli/init.ts
@@ -17,6 +17,7 @@ import { runInstall } from "./install.ts";
 import { runPersona } from "./persona.ts";
 import { runPhantomchat } from "./phantomchat.ts";
 import { runTelegram } from "./telegram.ts";
+import { runVoice } from "./voice.ts";
 
 export interface InitFlowInput {
   config: Config;
@@ -122,7 +123,8 @@ export default defineCommand({
       "  3. Connect PhantomChat (your private Nostr DM channel)\n" +
       "  4. Connect Telegram (optional — skippable)\n" +
       "  5. Enable semantic memory search (optional)\n" +
-      "  6. Install as a background service",
+      "  6. Enable voice — TTS/STT (optional)\n" +
+      "  7. Install as a background service",
       "Setup Flow"
     );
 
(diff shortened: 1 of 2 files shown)


kaieriksen (approved):
Reviewed at exact head 38f2695. Focused init/voice tests pass (17/17), local typecheck passes after locked dependency install, and CI is green. No blocking findings.

lenaparkhodges (approved):
**Approve** — verified independently at head 38f2695.

**What I checked:**
- **Symmetry claim holds.** Read `embedding.ts` alongside: the `embedded` gating (intro, existing-config note, restart prompt, outro) is the identical pattern, threaded through all four voice sub-flows (none/elevenlabs/openai/azure_edge). Standalone `phantombot voice` is untouched — `embedded` defaults false.
- **Non-fatal by design.** `init` ignores `runVoice`'s return, so a key-validation failure (returns 1) can't abort the wizard before install — matches embeddings' documented contract.
- **Framing is correct.** Parent wizard owns the single `p.outro`; no nested intro → no stray bracket. Welcome note renumbered to 7 steps, matches actual flow order.
- **Tests:** 25/25 pass locally (`cli-voice`, `cli-init`, `lib-voice`) — matches PR body.
- **Typecheck:** clean on both touched files; remaining repo errors are the pre-existing MCP SDK module-resolution noise in unrelated tests, as stated.
- CI green across the board incl. test-windows; Kai approved at same head.

Low blast radius: default off, skippable, no behavior change to existing commands. Good to merge.

PR #349, opened 2026-08-06
feat(vscode): multi-chat history — save, rename, retrieve past chats

andrewagrahamhodges:
## Problem
On VS Code (dogfooding with Megan), the ACP extension **saved no usable history**: closing and reopening VS Code lost the chat, and there was no way to keep, name, or retrieve older chats. Root cause (verified against Megan's live memory DB): the model-picker/participant paths minted a fresh **throwaway** server thread on every launch, orphaning each conversation; the dedicated panel only ever showed **one** cwd-wide session with no list/rename.

## Approach — mirror Zed (the reference ACP client)
Zed "just works" because of a clean ownership split, and our server already holds up its half:
- **Agent (phantombot)** owns transcript **persistence + replay** — `session/new` mints a self-describing thread token; `session/load` replays that thread. *No server changes needed.*
- **Client (this extension)** owns the **list, titles, and rename**, stored entirely in VS Code's own `workspaceState` — **never** in the phantombot domain (per the explicit requirement).

## What's in it
- **`sessionStore.ts`** (new, pure/`vscode`-free, bun-tested): per-workspace chat list — list/upsert/patch/remove, most-recently-used ordering, malformed-data tolerance, and **Zed-style auto-titling** from the first prompt.
- **`sessionProvider.ts`**: lists one entry per stored thread; opening a listed chat → `session/load` replays its history; a fresh ("untitled") chat mints a thread on open but is **persisted only on its first prompt** (blank chats leave no junk), auto-titled from that prompt, with an untitled→committed migration.
- **`extension.ts`**: `workspaceState` KV facade + a **`phantombot: Rename Chat`** command (palette quick-pick, or a session-item resource arg from a future context menu).
- Removed now-dead client-side `mintSessionId` / `resolveSessionCreated` (server owns minting; records carry `createdAt`).

## Scope calls (agreed with Andrew)
- **No retroactive recovery** of the old orphaned threads — the list starts fresh and grows from here.
- **No server changes**; all new state lives in VS Code storage.

## Testing
- `bun test`: **140 pass**, incl. new `sessionStore` suite + resource-decode coverage. `tsc --noEmit` clean; esbuild bundle builds.
- Next: **dogfood on Megan's VM** (build `.vsix` → install → exercise save/rename/reopen across a restart) before merge.

### DIFF
--- a/editors/vscode/package.json
+++ b/editors/vscode/package.json
@@ -83,12 +83,21 @@
         "title": "Open Phantombot Chat",
         "category": "phantombot",
         "icon": "$(hubot)"
+      },
+      {
+        "command": "phantombot.renameChat",
+        "title": "Rename Chat",
+        "category": "phantombot",
+        "icon": "$(edit)"
       }
     ],
     "menus": {
       "commandPalette": [
         {
           "command": "phantombot.openSession"
+        },
+        {
+          "command": "phantombot.renameChat"
         }
       ],
       "view/title": [
--- a/editors/vscode/src/acpClient.ts
+++ b/editors/vscode/src/acpClient.ts
@@ -107,7 +107,16 @@ export class AcpClient {
   private readonly pending = new Map<JsonRpcId, Pending>();
   private readonly requestTimeoutMs: number;
   private readonly onDiagnostic: (text: string) => void;
-  /** sessionId → live prompt handlers, set for the duration of a prompt. */
+  /**
+   * sessionId → live prompt handlers, set for the duration of a prompt.
+   *
+   * Since #349 every chat has a STABLE sessionId, so two turns can briefly
+   * overlap on one sid: the user submits a new prompt while the previous turn
+   * is still resolving (chat-channel "interrupt" — the server aborts the old
+   * turn and starts the new one). The newer turn's handlers replace the older
+   * turn's entry here; the older turn's `finally` must therefore delete its
+   * slot ONLY if it still owns it (identity check), never clobber the successor.
+   */
   private readonly promptStreams = new Map<string, PromptHandlers>();
   /** sessionId → the slash commands the agent advertised for that session. */
   private readonly sessionCommands = new Map<string, AcpAvailableCommand[]>();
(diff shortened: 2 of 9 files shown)


lenaparkhodges:
**APPROVED** ✅

Reviewed at head `b464ec6`. CI green (Linux + Windows), 140 tests pass, typecheck clean.

**Architecture — clean and well-scoped.** The ownership split mirrors Zed exactly: agent owns persistence (server threads via `session/new` + `session/load`), extension owns the list/titles (VS Code `workspaceState`). No server changes needed. `sessionStore.ts` is pure/vscode-free with solid test coverage.

**Key findings:**

- **`sessionStore.ts`** (new): Clean CRUD with MRU ordering, malformed-data tolerance, `deriveSessionTitle()` auto-naming from first prompt, and `titleSettled` flag that correctly gates the one-shot auto-title so renames are never clobbered. 185 lines of tests covering round-trip, ordering, patch, remove, garbage tolerance, and truncation.

- **`sessionProvider.ts`**: Untitled chats mint a thread on open but only persist to the list on first prompt — no junk from blank sessions. `onDidCommitItem` fires the untitled→committed resource migration (wired to VS Code's proposed API, consumed internally). `recordTurn()` correctly handles both first-prompt creation and subsequent-turn `lastUsedAt` bumps. Interrupt rendering as `_(interrupted)_` is the right UX.

- **`acpClient.ts`** (interrupt fix): The identity check (`promptStreams.get(sessionId) === handlers`) before delete is the critical fix — without it, the interrupted turn's `finally` evicts the successor's stream on the same stable sid. Applied symmetrically to both `loadSession` and `prompt`. The regression test covers the exact race.

- **Dead code removed cleanly**: `mintSessionId()` and `resolveSessionCreated()` deleted along with their tests; `decodeSessionResource()` added for the new path+query URI scheme with matching test coverage.

**Non-blocking notes:**
1. No "Delete Chat" command in the UI yet — scope seems intentional per the PR body, but worth noting as a follow-up.
2. `pendingThread` map is never pruned on workspace folder removal — not a real issue since the extension reactivates per workspace.

Solid work. Merge when ready. 📊

lenaparkhodges (approved):
Approved. Multi-chat history for VS Code extension — clean architecture, interrupt fix is correct, 140 tests pass, typecheck clean, CI green on all platforms. Non-blocking: no delete command yet (intentional scope), pendingThread map doesn't prune on folder removal (benign).

PR #350, opened 2026-08-06
Revert: VS Code multi-chat history (#349)

andrewagrahamhodges:
Reverts #349 (`19d9314`) in full.

## Why
Session save still wasn't working reliably after the multi-chat rewrite, and the pre-#349 dedicated-panel behavior felt better in practice. Rolling back to that known-good baseline; we can re-approach multi-chat cleanly later.

## Tradeoff (known)
#349 bundled two things in one squash:
1. **Multi-chat history** (save/rename/reopen) — the part being rolled back.
2. **The interrupt fix** (clean `_(interrupted)_` instead of the ugly error).

A full revert is the cleanest, lowest-risk path, but it **also drops the interrupt fix** — the clobber it fixed only existed once #349 introduced stable sessionIds, so reverting removes both the cause and the fix together. Net: back to the old working panel; the ugly interrupt error returns. Chose the clean full revert over a fiddly surgical one that risked rough edges on intertwined sessionId changes.

## Verification
- `tsc --noEmit` clean
- `bun test ./tests` — 128 pass
- `esbuild` bundle builds

No server changes.


kaieriksen (approved):
Reviewed exact head 19dc7b9. No blocking findings: this is a coherent full revert of #349, and the documented loss of the interrupt fix is an explicit tradeoff of returning to the prior single-session behavior.

Validation: VS Code tests 128/128, VS Code typecheck clean, VS Code esbuild bundle clean, and GitHub CI/typecheck/test/Windows/CodeQL checks green. The full local suite reached 2600 pass / 2 skip / 2 fail; both failures are the known ambient PHANTOMBOT_PI_API_KEY test contamination, not caused by this revert.

Issue #352, opened 2026-08-08
pi harness accepts a narration-only exit-0 run as a completed turn — no fallback fires, stored "answer" is just tool narration


14skywalker:
## Summary

When the primary `pi` harness is used for a long/long-running coding task, the turn **never actually completes** — the model runs tools, streams each step as text *narration*, and then the run exits with code 0 **before ever emitting the real deliverable**. The harness accepts that narration-only run as a successful `done`, so:

- the streamed reply the user sees is just a few `tool:` progress lines, then silence;
- **no fallback to the next harness in the chain (e.g. `claude`) ever fires** — `runWithFallback` already got a `done` and stops;
- `/status` reports `active: no`, and a follow-up "continue" makes the next turn re-do the same work, because **there is no finished answer** — the stored "assistant reply" is only the tool narration trimmed mid-task.

The concrete defect is: **a process that exits 0 with nothing but streamed tool-narration text is treated as a successful final answer**, when it is actually a "no real output" state that should fall through (`recoverable`) to the next harness.

> **Confidence scope:** this is pinned directly against the code (see *Actual behavior*), and matches the reporter's observed symptom exactly: per-prompt, the store gets one assistant turn that is short (200–660 chars), reads as an opener ("Ik ga #150 implementeren — eerst de repo…"), and contains no implementation, no closing summary; each stored user+assistant pair shares an identical timestamp (a single `appendPairTxn`). The reported "*continue* does not report it is finished" confirms there is genuinely no completed answer. The *trigger* (a long task where the model narrates tools and exits before the deliverable) is the reporter's reproduction; the defect — an exit-0 run with no completion signal is accepted as done — holds regardless.

## Environment

- phantombot: main
- harness chain: `[pi, claude]`
- primary model: `openrouter/auto-beta` via provider `openrouter`
- `harness_idle_timeout_s = 120`
- `harness_hard_timeout_s = 3600`

## Expected behavior

A run whose only output is tool narration — no completed result, no completion event — should **not** surface as a successful final answer. It should be treated as a recoverable "no useful output" state and fall through to the next harness in the chain (`attempt 2`), so the user always gets a real completed reply.

## Actual behavior (pinned in code)

In `harnessRunner.runHarnessProcess` (`src/lib/harnessRunner.ts`):

[...]

PR #353, opened 2026-08-08
fix(pi): completion signal + tool-activity watchdog cap (closes #351, #352)

### LINKED ISSUE #351: pi harness idle watchdog is defeated by primary-model tool_execution_update — a wedged turn hangs the stream with no fallback
## Summary

When the primary `pi` harness wedges mid-turn on a long coding task, the **streamed reply just stops** and the user is left hanging: a few tool lines come out, then silence — and **no fallback to the next harness (e.g. `claude`) ever fires**, not for minutes and potentially not for up to the full `harness_hard_timeout_s` (default 3600 s).

The reporter's symptom: a prompt is answered with a handful of `tool:` progress lines, then the reply dies mid-stream. The process is **not** crashed, **not** OOM-killed — the subprocess is still alive with its stdout open, so `runHarnessProcess` never reaches the EOF that synthesizes a `done`/`error` chunk, and `runWithFallback`'s `for await (const chunk of harness.invoke(req))` loop (in `src/orchestrator/fallback.ts`) just waits on an iterator that never exhausts.

The **mechanism** (confirmed in code) is that the idle watchdog that should have killed the wedged subprocess is being defeated by the **primary model's own** `tool_execution_update` events.

> **Confidence scope:** the watchdog-defeat mechanism below is verified directly against the code and is the real bug. The *trigger* described in Reproduction (an OpenRouter auto-router stalling while the stream stays open) is the **suspected** scenario that reproduces it in the reporter's environment; the underlying defect — that a wedged-but-chattery primary never trips the idle kill and suppresses the fallback — holds regardless of the specific trigger. If you can reproduce another way, the fix section is unchanged.

## Environment

- phantombot: main
- harness chain: `[pi, claude]`
- primary model: `openrouter/auto-beta` via provider `openrouter`
- `harness_idle_timeout_s = 120`
- `harness_hard_timeout_s = 3600`

## Expected behavior

A harness that has produced no *useful* output (no final text, no completed tool result) for `harness_idle_timeout_s` should be killed and the turn should fall through to the next harness in the chain, exactly as designed (the idle kill is documented as "recoverable — wedged subprocess, try a different one" in `harnessRunner.ts`).

## Actual behavior

A wedged-but-chattery primary keeps the stream open and the fallback never runs. The watchdog is reset over and over because `piActivity` classifies **both** `tool_execution_start` and `tool_execution_update` as the `"tool"` activity class:

- `parsePiEvent` (`src/harnesses/pi.ts`) maps `tool_execution_update` → `{ type: "heartbeat" }` (payload-less, so the channel does not leak a partial result).
- `piActivity` (`src/harnesses/pi.ts`) then returns `"tool"` for `tool_execution_update`.
- `harnessRunner.touch("tool")` (`src/lib/harnessRunner.ts`) sets `toolRunning = true` AND resets `idleTimer`:
  ```ts
  if (activity === "tool") {
    toolRunning = true;
  } else if (activity === "productive") {
    toolRunning = false;
  } else if (toolRunning) {
    return;           // "model" heartbeat is ignored while a tool is "running"
  }
  clearTimeout(idleTimer);
  idleTimer = setTimeout(() => triggerKill("idle"), opts.idleTimeoutMs);
  ```

So while the model is wedged but keeps pipelining `tool_execution_*` events (a stalled auto-router that keeps the stream open and occasionally reports a tool event does exactly this), the idle timer is perpetually reset, `triggerKill("idle")` never fires, and the only net is `hardTimeoutMs`.

Compounding factor: `done`/`error` are only synthesized in `harnessRunner.runHarnessProcess` after `proc.stdout` EOF. A stalled-but-open stream therefore never yields a `done` or an `error`, so `runWithFallback` never advances to `attempt 2` and never even surfaces a recoverable-error chunk to the orchestrator. The user just gets a dead stream for up to `harness_hard_timeout_s`.

## Reproduction

1. Configure `chain = ["pi", "claude"]` with an OpenRouter auto-router `primary_model`, `harness_idle_timeout_s = 120`.
2. Send a long coding task that produces several tool calls and takes minutes to resolve.
3. If the primary's auto-router stalls while the stream stays open and `tool_execution_update` events keep arriving (the suspected trigger in the reporter's environment), the same watchdog-defeat path runs.
4. Result: a few `tool:` progress lines, then the reply dies; no `attempt 2`, no error surfaced to the channel, no watchdog kill until the hard timeout.

## Why "only delegate-coder output should reset idle"

`tool_execution_update` exists to carry the capability-routing `coder` delegate's real progress ("Only ever emitted on genuine child output, so a wedged tool still trips the idle kill" — comment in `parsePiEvent`). But the classifier has no way to tell a **delegate** `coder` update from a **primary-model** update, so it treats both as timer-resetting `"tool"` activity. When the primary itself is the wedged actor, that same path defeats the watchdog.

## Proposed fix (upstream)

1. Give `piActivity` a way to distinguish real *delegate* progress from the primary model's own tool chatter — e.g. only classify `tool_execution_update` as `"tool"` when the event is attributable to the `coder`/delegate (tag it in `parsePiEvent`); otherwise treat it as a regular `"model"`-class heartbeat that does **not** reset the idle timer while a tool is "running" (the `toolRunning` branch in `touch()` already provides that guard).
2. Alternatively / in addition: make an open stdout with no `done`/`error`/final text a recoverable condition — e.g. treat sustained `tool` activity with **no `text` output** as non-productive for watchdog purposes, so a stall can't outlive the idle window simply because the primary keeps hand-waving.
3. As a safety net regardless, consider a shorter default `harness_hard_timeout_s` (currently 3600 s) so a genuinely wedged turn falls through to the next harness within minutes instead of an hour.

## Workarounds available today

- Put the (already-auth-fixed) `claude` harness **first** in the chain (`chain = ["claude", "pi"]`) so the fragile router only ever runs as a last-resort fallback.
- Avoid `openrouter/auto-beta` as `primary_model`/`coding_model`; use a concrete model provider so the router cannot stall.
- Lower `harness_hard_timeout_s` to a few minutes so a stuck turn at least fails over eventually.

14skywalker:
Two fixes to the `pi` harness for the same root gap — the harness had **no completion signal** and derived `done` purely from the process exit code. They are the two sides of that gap and are fixed together, one commit each.

## #352 — narration-only exit-0 accepted as a finished answer

A long task where the model narrates each tool step as `text` and then exits 0 **before** producing the real deliverable was synthesized straight into a terminal `done`. The stored "reply" was just trimmed tool narration, no fallback fired, and a follow-up "continue" re-did the work because nothing had actually completed. Emptiness can't discriminate — narration *is* non-empty text.

pi does emit a completion signal, `turn_end`, but `parsePiEvent` dropped it at the `message_update` guard. Fix:

- `parsePiEvent` surfaces `turn_end` as a payload-less `done` marker. Gated on `turn_end` **alone**, not the later `agent_end`/`agent_settled`: a run that errors out can still emit `agent_end`, and accepting that as "done" would store a broken turn as complete.
- `runHarnessProcess` gains an opt-in `requireCompletion` flag: on exit 0 with no completion marker, it yields a **recoverable** error so the orchestrator falls through to the next harness. pi opts in; every other harness keeps deriving `done` from the exit code.

**Verified empirically** against the installed pi (0.80.10): a successful run always ends `turn_end → agent_end → agent_settled`, in `--no-tools` mode too. On a hypothetical pi build that never emits `turn_end`, this degrades **safely** — a recoverable fallback that still yields a real answer, rather than returning trimmed narration as the reply.

## #351 — chattery `tool_execution_update` defeats the idle watchdog

`piActivity` classifies both `tool_execution_start` and `tool_execution_update` as `"tool"`, and `touch("tool")` resets the idle timer every time. pi emits `tool_execution_update` on genuine new output (bash streams stdout, the coder delegate forwards its child) so a truly silent wedge still trips the idle kill — but a tool that **trickles output forever** (a stalled auto-router, a hung retry that keeps logging) resets the idle timer over and over, leaving only the 3600s hard cap as the net. Fix:

- `KillCoordinator` gains `toolTimeoutMs`: a wall-clock cap on a single **contiguous** tool-run. Past the cap, tool activity no longer re-arms the idle timer, so the idle kill fires (recoverable → fallback). Any productive output resets the budget, so a healthy turn interleaving tools with text is untouched. Omit ⇒ legacy behaviour (unchanged for claude/codex).
- pi opts in with a **generous** net: `max(idle×4, 20min)`, capped at the hard timeout — comfortably above any realistic tool-run (a legit long build/delegation streams real output the whole time) yet well under the hard cap, so a wedge fails over in minutes instead of an hour.

## Notes

- Not a regression — the harness has derived `done` from the exit code since the harness consolidation in #151; `turn_end` was never wired as a completion signal (it only appeared in a doc comment).
- The mid-stream fall-through follows the existing streaming-first trade-off in `runWithFallback` (a user may briefly see partial narration before the next harness's full reply).

## Tests

- `parsePiEvent`: `turn_end` → `done`; `agent_end` stays ignored.
- e2e (`fake-pi.sh` `nofinish` mode): exit 0 with no `turn_end` → recoverable error, not `done`.
- `KillCoordinator` tool cap: sustained `"tool"` activity past the cap lets the idle kill fire; without a cap it keeps the process alive; productive output resets the budget.

`bun tsc --noEmit` clean (pre-existing `werift`/p2p errors aside). New tests green; the one failing `lib-harnessRunner` idle-timing test is a pre-existing environment flake (fails on `main` unchanged).

### DIFF
--- a/src/harnesses/pi.ts
+++ b/src/harnesses/pi.ts
@@ -15,9 +15,13 @@
  * Stream-json events translated to phantombot HarnessChunks:
  *   message_update with text_delta  → { type: "text", text }
  *   tool_execution_start            → { type: "progress", note: "tool: <name>" }
- *   anything else (agent_start,
- *     tool_execution_end, turn_end,
- *     extension_*) → ignored        (the done chunk is emitted from process exit)
+ *   turn_end                        → { type: "done" } COMPLETION marker: the
+ *     model finished this turn. The shared engine requires it before it will
+ *     synthesize the terminal `done` on exit 0, so an exit-0 that stopped
+ *     mid-task (only tool narration, no turn_end) falls through to the next
+ *     harness instead of being stored as a finished answer (issue #352).
+ *   anything else (agent_start, agent_end, agent_settled,
+ *     tool_execution_end, extension_*) → ignored
  *
  * Auth (per-turn, with local-store fallback): when PHANTOMBOT_PI_API_KEY is
  * set, phantombot threads it onto `--api-key` per turn (the same way it threads
@@ -271,6 +275,12 @@ export class PiHarness implements Harness {
       parseEvent: parsePiEvent,
       activity: piActivity,
       buildDoneMeta: () => ({ harnessId: this.id, payloadBytes: totalBytes }),
+      // pi is the only harness with no native terminal `done` in its stream —
+      // it derives completion from turn_end (mapped to a `done` marker in
+      // parsePiEvent). Require that marker before accepting an exit-0 run as a
+      // finished answer, so a narration-only mid-task exit falls through to the
+      // next harness rather than being stored as the reply (issue #352).
+      requireCompletion: true,
     });
 
     } finally {
@@ -367,6 +377,27 @@ export function parsePiEvent(parsed: unknown): HarnessChunk | undefined {
   // emitted on genuine child output, so a wedged tool still trips the idle kill.
   if (obj.type === "tool_execution_update") {
     return { type: "heartbeat" };
+  }
+
+  // turn_end is pi's end-of-turn COMPLETION signal — the model has finished
+  // this turn. Verified against pi 0.80.x: a successful run always ends
+  // turn_end → agent_end → agent_settled, in --no-tools mode too. Surface it
+  // as a `done` marker (payload-less: the reply text already streamed as
+  // text_delta chunks and is accumulated by the shared engine) so that engine
+  // can distinguish a genuinely COMPLETED turn from an exit-0 that stopped
+  // mid-task having emitted only tool narration — see runHarnessProcess's
+  // `requireCompletion` gate and issue #352. We deliberately gate on turn_end
+  // ALONE, not the later agent_end/agent_settled: a run that errors out can
+  // still emit agent_end, and treating that as "done" would accept a broken
+  // turn as complete. turn_end fires only when the turn itself finished.
+  //
+  // Version caveat: on a hypothetical pi build that does not emit turn_end,
+  // every turn would fail the gate and fall through to the next harness. That
+  // degrades safely — a recoverable fallback that still yields a real answer
+  // (double work, not a wrong answer) — whereas the pre-fix behaviour returned
+  // the trimmed narration as if it were the answer.
+  if (obj.type === "turn_end") {
+    return { type: "done", finalText: "", meta: undefined };
   }
 
   if (obj.type !== "message_update") return undefined;
--- a/src/harnesses/pi.ts
+++ b/src/harnesses/pi.ts
@@ -281,6 +281,21 @@ export class PiHarness implements Harness {
       // finished answer, so a narration-only mid-task exit falls through to the
       // next harness rather than being stored as the reply (issue #352).
       requireCompletion: true,
+      // Bound how long a single contiguous tool-run may keep the idle watchdog
+      // alive via tool_execution_* activity alone, with no user-facing text. pi
+      // only emits tool_execution_update on genuine new output (bash streams
+      // stdout, the coder delegate forwards its child's progress) so a healthy
+      // turn is rarely affected — but a tool that trickles output forever (a
+      // stalled auto-router, a hung network retry that keeps logging) could
+      // otherwise reset the idle timer up to the 60-min hard cap with no
+      // fallback (issue #351). This is a GENEROUS last-resort net: comfortably
+      // above any realistic tool-run yet well under the hard cap, so a
+      // wedged-but-chattery turn fails over in minutes, not an hour. Derived
+      // from the idle window, floored at 20 min, and never above the hard cap.
+      toolTimeoutMs: Math.min(
+        req.hardTimeoutMs ?? Number.POSITIVE_INFINITY,
+        Math.max(req.idleTimeoutMs * 4, 1_200_000),
+      ),
     });
 
     } finally {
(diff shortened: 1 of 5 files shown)


robertclawson (approved):
Reviewed the diff against `main` and traced the runtime logic — LGTM, merge-ready.

**Verified:**
- **Kill-cause precedence** (`harnessRunner.ts`): the `requireCompletion` gate sits *after* the idle/hard/policy kill-cause branch and the `code !== 0` branch, so it only fires on a genuine clean exit 0 with no `turn_end`. A killed or errored run can't be misread as "stopped mid-turn." 👍
- **Tool-cap engages on the right events**: `piActivity` classifies `tool_execution_update` as `"tool"` off the raw event (even though it maps to a `heartbeat` chunk), so a chattery-but-wedged tool genuinely drives `touch("tool")` and the cap bites. 
- **`touch()` cap semantics**: past `toolTimeoutMs` it returns *before* re-arming the idle timer, letting the last-armed idle window run out → `triggerKill("idle")`. `"productive"` clears `toolRunStart`, so an interleaved healthy turn resets the budget. The three new coordinator tests cover exactly these (cap fires / no-cap survives / productive resets).
- **`turn_end` handling**: surfaced as a payload-less `done`, consumed by the `consume()` mapping (sets `sawCompletion`, not re-yielded), and gated on `turn_end` alone — `agent_end` stays ignored, which is the right call since an errored run can still emit it. `buildDoneMeta` ignores the `undefined` captured meta so there's no meta regression.

**Safe-degradation** on a hypothetical pi build with no `turn_end` (every turn falls through to fallback = double work, not a wrong answer) is the correct trade, and it's documented at the call site.

No blocking concerns. Nice, thorough tests and comments.

lenaparkhodges (approved):
Approved. Reviewed on the branch, verified locally.

**#352 (completion signal):** Gating on `turn_end` alone — and deliberately not `agent_end` — is the right call; a run that errors can still emit `agent_end`. The `requireCompletion` opt-in keeps every other harness on legacy exit-code semantics, and the failure mode on a hypothetical turn_end-less pi build degrades to a recoverable fallback (double work, not a wrong answer). That's the correct side to err on.

**#351 (tool-activity cap):** Capping a *contiguous* tool-run while letting productive output reset the budget means healthy turns interleaving tools + text are untouched — the new tests prove exactly that. `max(idle×4, 20min)` capped at the hard timeout is generous enough for legit long builds/delegations.

**Verified locally (pr-353 @ HEAD):**
- New tests green: `turn_end → done` marker, `nofinish` e2e → recoverable error not done, tool-cap kill firing, uncapped legacy behaviour, budget reset on productive output.
- 2 failing tests (`PHANTOMBOT_PI_API_KEY` routing) fail identically on `main` — pre-existing env leakage in my shell, not this PR. (Worth isolating that env var in the test setup someday — not a blocker here.)
- `bun tsc --noEmit`: identical 20 pre-existing errors on main and this branch (MCP SDK/werift), nothing new.

Nice catch that narration-only text can't be discriminated by emptiness — this closes a real wrong-answer hole, not just a cosmetic one.

PR #354, opened 2026-08-09
fix(mcp): order vault write before registry write in runMcpAdd

### LINKED ISSUE #343: MCP: order vault write before registry write in runMcpAdd
## Context

Follow-up nit from PR #342 review (raised by Kai, echoed by Lena; both marked non-blocking).

In `runMcpAdd`, when ingesting a pre-registered OAuth client (`--client-file`/`--client-id`/`--client-secret`), the registry entry is written **before** the vault client secret. The `credentials.json` is parsed first, so a malformed file fails safely before any write — but if the vault write throws *after* the registry write succeeds, the server is left registered without its stored client.

## Impact
- **Low.** Not silent — the command errors on the vault write. Fully recoverable: re-run `mcp add` with the same file; `upsertServer` overwrites and the vault write retries (idempotent).
- Worst functional consequence: a stale `mcp login` attempts DCR (and hits Google's rejection) instead of skip-DCR — i.e. the loud original error, not a wrong result.

## Fix
Swap the order — write the vault client secret first, then the registry entry. One-line change. (Trades a client-less server for a possible orphaned vault key on partial failure; neither is harmful, but registry-last is the cleaner invariant since the registry is the thing agents read.)

Good candidate to fold into the next MCP-related PR rather than a standalone rushed change.

Refs #342, #341.

andrewagrahamhodges:
## What

When `runMcpAdd` ingests a pre-registered OAuth client (`--client-id/--secret/--client-file`), swap the write order so the **vault client secret is written before the registry entry**.

## Why

The registry is what agents read to reach a server. Previously we wrote the registry entry first, then the vault secret — so a vault-write failure *after* the registry write landed would leave a server registered without its stored client secret. Writing the vault secret first keeps the invariant: a registered server always has its client in the vault.

Low impact (the command errors loudly and `mcp add` is idempotent on re-run), but a cleaner ordering. Raised by Kai, echoed by Lena, filed by Robbie in PR #342 to fold into the next MCP change.

## Test

- `bun test tests/cli-mcp.test.ts` → 12 pass
- `bun tsc --noEmit` clean

Closes #343

### DIFF
--- a/src/cli/mcp.ts
+++ b/src/cli/mcp.ts
@@ -168,10 +168,11 @@ export async function runMcpAdd(input: McpAddInput): Promise<number> {
   }
 
   const dir = await resolvePersonaDir(input.persona);
-  const registry = await loadRegistry(dir);
-  await saveRegistry(dir, upsertServer(registry, input.id, entry));
-  out.write(`registered MCP server '${input.id}' (${transport})\n`);
 
+  // Write the vault client secret BEFORE the registry entry. The registry is
+  // what agents read to reach a server; if we registered first and the vault
+  // write then failed, we'd leave a server registered without its stored
+  // client. Ordering the vault write first keeps that invariant intact.
   if (preClient && tokenRef) {
     const vault = await openPersonaVault(dir);
     try {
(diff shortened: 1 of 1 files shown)


kaieriksen (approved):
Reviewed exact head ae338fd7cfb6e71de2a4cff3e720ed004a262ebd. The vault write now completes before registry persistence, preserving the registered-server/client invariant. Focused MCP tests 12/12 and typecheck pass; full local suite has 2600 pass with two unrelated ambient PHANTOMBOT_PI_API_KEY test failures. CI test/typecheck/CodeQL pass; the Windows job has unrelated timeout failures.

PR #357, opened 2026-08-10
fix(memory): retry WAL-mode switch on SQLITE_BUSY so a fresh DB survives an aligned boot

andrewagrahamhodges:
## What this fixes

`openMemoryStore` could throw `database is locked` (SQLITE_BUSY) while opening a **brand-new** `memory.sqlite` if a second process is touching the file at the same instant — crash-looping the daemon. The trigger is an aligned boot: a phantom's scheduler firing `run` and `tick` (or two `run` starts) simultaneously against a DB that doesn't exist yet.

## Root cause (with a correction to the earlier hunch)

I originally suspected a simple pragma-ordering bug (`journal_mode = WAL` running before `busy_timeout`). Building a deterministic cross-process reproduction proved that **reordering alone does not fix it**:

- Converting a rollback-journal DB to WAL needs a momentary **EXCLUSIVE** lock, and SQLite **does not invoke the busy handler** for that conversion. With another connection holding a write lock, `PRAGMA journal_mode = WAL` throws SQLITE_BUSY in **0 ms** regardless of `busy_timeout`.
- Once the file is already in WAL mode (the steady state after first boot), the same pragma is a **lock-free no-op** and succeeds instantly even under a held write lock.

So the hazard is confined to the **first-ever open of a fresh DB** under contention — exactly a new-install / aligned-boot race, not a steady-state one. (An existing, months-old store — e.g. Megan's — does not hit this; that separate crash loop was a duplicate scheduled task hammering writes, already resolved operationally.)

## The fix

- Set `busy_timeout` **first**, so schema DDL, migrations, and normal writes are all covered by block-and-retry.
- Give the WAL switch its **own retry-until-deadline loop** (`enableWalMode`), since `busy_timeout` provably can't protect it. Steady-state opens still return on the first attempt.

## Test

Adds a cross-process regression (`tests/memory.test.ts`): a separate process takes a write lock on a fresh DB, lingers, then releases; `openMemoryStore` must block-and-retry and succeed rather than throw. Verified it **fails without the retry loop** (throws SQLITE_BUSY in ~29 ms) and **passes with it**; 5x flake-stress clean.

## Verification

- `bun test tests/memory.test.ts` — 23/23 green.
- Full suite: only the known pre-existing PiHarness `--api-key` fails and the missing-MCP-SDK-dep env errors remain; this change adds none.
- Typecheck clean for the touched files.
- Cross-platform: pure SQLite/Bun, no OS-specific code — behaves identically on Windows CI.

### DIFF
--- a/src/memory/store.ts
+++ b/src/memory/store.ts
@@ -1451,6 +1451,37 @@ function mapDisplayRows(rows: RawDisplayRow[]): Turn[] {
   }));
 }
 
+/**
+ * Switch a database to WAL journal mode, retrying on SQLITE_BUSY.
+ *
+ * Converting a rollback-journal DB to WAL needs a momentary EXCLUSIVE lock,
+ * and SQLite does NOT invoke the busy handler for that conversion — so
+ * `busy_timeout` cannot wait it out. Empirically, if any other connection
+ * holds a write lock at that instant, the switch throws SQLITE_BUSY in 0ms
+ * regardless of the configured timeout. This bites a fresh install whose
+ * scheduler fires two processes at once (e.g. a Windows phantom's `run` and
+ * `tick` tasks aligned on boot) against a brand-new memory.sqlite: one wins
+ * the conversion, the other throws and crash-loops the daemon.
+ *
+ * So we retry on our own deadline instead of relying on busy_timeout. Once
+ * the file is in WAL mode (the steady state after first boot) this pragma is
+ * a lock-free no-op and the first attempt returns immediately, even while
+ * another process holds a write lock.
+ */
+async function enableWalMode(db: Database, deadlineMs = 5000): Promise<void> {
+  const deadline = Date.now() + deadlineMs;
+  for (;;) {
+    try {
+      db.exec("PRAGMA journal_mode = WAL");
+      return;
+    } catch (err) {
+      const busy = (err as { code?: string } | null)?.code === "SQLITE_BUSY";
+      if (!busy || Date.now() >= deadline) throw err;
+      await Bun.sleep(50);
+    }
+  }
+}
+
 export async function openMemoryStore(path: string): Promise<MemoryStore> {
   if (path !== ":memory:") {
     await mkdir(dirname(path), { recursive: true });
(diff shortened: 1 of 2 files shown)


kaieriksen (approved):
Reviewed exact head f8debefde5ae978e62a6d31a387f5e9b3eee5b68. The busy_timeout-before-WAL ordering plus explicit SQLITE_BUSY retry closes the fresh-DB conversion race; the cross-process regression passes 23/23 and typecheck passes locally. Hosted test, Windows, typecheck, and CodeQL checks are green.

robertclawson (approved):
**Approve** ✅ — verified at head `f8debef`, read the full diff and reproduced the core mechanism locally rather than trusting the writeup.

**What I confirmed empirically (Bun + bun:sqlite):**
- **busy_timeout genuinely can't protect the WAL switch.** With another connection holding a write lock on a fresh rollback-journal DB, `PRAGMA journal_mode = WAL` throws in ~1ms even with `busy_timeout = 5000` armed — SQLite doesn't run the busy handler for the EXCLUSIVE conversion. The root-cause correction in the description holds; pragma reordering alone would not have fixed it.
- **The error match is right.** bun:sqlite surfaces it as `SQLiteError` with `err.code === "SQLITE_BUSY"` — exactly what `enableWalMode` keys on. (This is also implicitly proven by the new cross-process test going green on both `test` and `test-windows`: a wrong code string would make `enableWalMode` rethrow and the test reject.)
- **The retry recovers, and steady state stays free.** Ran the `enableWalMode` loop against a 600ms holder: it blocked-and-retried, resolved at ~612ms with the DB in WAL, and a second call returned in **0ms** — so post-first-boot opens pay nothing.

**Ordering is correct:** `busy_timeout` is set first, so all schema DDL + migrations + normal writes in the `SqliteMemoryStore` constructor (which runs after the pragmas) get block-and-retry; only the WAL conversion — the one thing busy_timeout can't cover — gets the dedicated deadline loop. Scope is tight (WAL switch only), the fix is cross-platform (pure SQLite/Bun), and CI is fully green (test, test-windows, typecheck, CodeQL).

**Two non-blocking notes:**
1. If contention outlasts the 5s `deadlineMs`, `enableWalMode` still throws `SQLITE_BUSY` — correct degradation for a momentary aligned-boot race, but if a *sustained* multi-process writer is ever the trigger this only widens the window, doesn't remove it. Fine for the stated failure mode.
2. Nit: `busy_timeout` and the WAL retry are two independent 5s budgets, so a worst-case cold open can wait up to ~10s before surfacing. Acceptable, just worth knowing.

Clean approve — nothing to gate on.

lenaparkhodges (approved):
**Approved.** Both empirical claims in the PR body reproduce exactly on my Linux box:

1. **Reordering alone does not fix it** — with `busy_timeout = 5000` set *first* and no retry loop, `PRAGMA journal_mode = WAL` against a write-locked fresh DB still throws `SQLITE_BUSY` in **1 ms**. The busy handler genuinely does not cover the WAL conversion. The `enableWalMode` retry loop is the entire fix, not the reorder.
2. **Steady state is free** — against a DB already in WAL mode with another process holding a write lock, the open + pragma completes in 0 ms. No steady-state regression.

Local: 23/23 memory tests green; CI all green including test-windows. Deadline-bounded retry (5s) with immediate rethrow on non-BUSY errors is the right shape.

**One should-fix, non-blocking:** the comment in `tests/memory.test.ts` still tells the abandoned story — *"The failure mode was an ordering bug… `journal_mode = WAL` ran BEFORE `busy_timeout` was set, so the WAL switch threw SQLITE_BUSY on the spot instead of waiting."* That's contradicted by the PR body's own correction and by the repro above (it throws regardless of ordering, in ~0 ms). Anyone reading the test later will take away the wrong mechanism. Please rewrite that comment to match the (correct) `enableWalMode` docblock before merge.

PR #358, opened 2026-08-11
fix(acp): force vscode extension reinstall to heal orphaned installs

andrewagrahamhodges:
## Problem

`phantombot acp install vscode` version-gates on `code --list-extensions --show-versions`, which reads VS Code's **registry** (`extensions.json`), not disk. When the extension folder is pruned out from under the registry, the gate reports the version as `current` and no-ops — reinstalling appears to do nothing and the ACP/chat surface stays dead.

Hit live on Megan: VS Code 1.128.0 ran its extension GC, pruned the pinned sideloaded VSIX folder, but left the manifest row behind. `code --list-extensions` cheerfully reported `phantomyard.phantombot-vscode@1.1.242` as installed while the folder was gone (`FOLDER_MISSING`). The `--force` already on the install command never got reached because the gate short-circuited first. The manual heal was `code --uninstall-extension … && phantombot acp install vscode`.

## Fix

- Add a `force` option to `installVscode()` that bypasses the version gate and re-lays the bundled `.vsix` via `code --install-extension --force`, emitting a new honest `reinstalled` action (rather than a bogus "upgraded X → X").
- Wire the **explicit** `phantombot acp install vscode` command to force by default, with `--no-force` to opt back into version-gated behaviour. A user running the command by hand almost always wants a real reinstall.
- The **automatic** reconcile loop (`autoInstall`) stays version-gated and never forces, so it doesn't reinstall on every startup.
- `vscodeResultToConnector` maps `reinstalled` like `updated` in doctor output; the CLI's "restart VS Code to load it" hint now fires for it too.

## Tests

Added four cases: forced reinstall over an already-current version → `reinstalled` (re-lays vsix, cleans up); force off by default still short-circuits with no write; force on a missing extension is a plain `installed`. Full suite green (`installVscode` + `autoInstall-vscode`: 45 pass), typecheck clean.

## Note

The underlying orphan (registry row without a folder) is a VS Code GC quirk, not something we create — but forcing the reinstall makes the documented user command self-heal it instead of silently no-opping.

### DIFF
--- a/src/cli/acp.ts
+++ b/src/cli/acp.ts
@@ -59,10 +59,21 @@ const installVscodeCmd = defineCommand({
   meta: {
     name: "vscode",
     description:
-      "Install phantombot's first-party VS Code extension (bundled .vsix) via the `code` CLI — idempotent + version-aware. Skips cleanly if VS Code isn't installed.",
+      "Install phantombot's first-party VS Code extension (bundled .vsix) via the `code` CLI — idempotent + version-aware. Forces a reinstall by default so an orphaned/ghosted install self-heals; pass --no-force to keep the version gate. Skips cleanly if VS Code isn't installed.",
   },
-  async run() {
-    const result = installVscode();
+  args: {
+    force: {
+      type: "boolean",
+      description:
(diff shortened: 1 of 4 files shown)


kaieriksen (approved):
Approved at exact head dae5021e15287044248ff2a04650a46ece009569. Focused ACP/VS Code and update tests pass (54/54), typecheck passes, CLI --no-force parsing verified, and the full suite has 2607 passing tests with 2 unrelated ambient PHANTOMBOT_PI_API_KEY failures. GitHub CI is fully green, including Windows.

PR #361, opened 2026-08-11
macOS: automatic stable code-signing to stop TCC permission nagging

robertclawson:
## What & why

On macOS, phantombot re-prompts for home / Documents / Full Disk Access after **every** auto-update. macOS TCC pins each permission grant to the binary's exact code-signing identity; phantombot ships ad-hoc signed (`Identifier=a.out`) and rewrites its own binary on update, so every update invalidates the grant and the OS re-asks forever.

This scares people — **especially the external installs we never hear about**, who don't yet trust us. The point of this PR is to make the Mac stop frightening them.

## The fix: automatic stable signing, for everyone

phantombot gives its binary a **stable self-signed code-signing identity** (fixed `--identifier dev.phantombot` + a persistent self-signed cert). The designated requirement stops changing across updates, so a single Full Disk Access grant sticks forever.

This runs **automatically on every macOS update — no opt-in.** The first update creates the identity; every update after re-signs with it.

## Why it can never break an update

The "updates never break" guarantee comes from the **fail-safe design, not from gating**:

- **Dedicated throwaway keychain**, password generated by us and stored in the vault. The **login keychain is never touched**, so codesign runs fully headless — the dreaded "keychain password nobody knows" dialog can never appear.
- **Transactional**: back up the swapped binary, re-sign under a `timeout`, verify (`codesign --verify` **and** an exec `--version` check). On **any** failure, restore the binary and tear down anything created this run.
- Worst case degrades to **exactly today's behaviour**: working binary, ad-hoc signature, still nags. Never worse, never broken, never a hung headless update.

## `phantombot fix-signing`

Kept as a **manual** apply-now / repair command (no longer "the opt-in"): sign the current binary immediately without waiting for the next update, or repair signing if an update's automatic re-sign failed. macOS-only; friendly no-op on Linux/Windows.

## The one honest caveat

New/first-run installs still get **one** macOS Full Disk Access grant — unavoidable without a Developer ID cert + notarization or an MDM PPPC profile, which we don't have. But it goes from "nags on every update" to "grant once, silent forever."

## Tests

`bun tsc` clean. Full suite green except one pre-existing unrelated `config.test.ts` (Gemini migration) failure that also fails on clean `main`. Signing logic covered by a fake CommandRunner + in-memory vault (no Mac needed): auto-create-then-sign on first update, re-sign on subsequent updates, full rollback on codesign/exec failure, and the non-fatal update-path wiring. Runbook at `docs/macos-signing.md`.

### DIFF
--- a/src/lib/macSigning.ts
+++ b/src/lib/macSigning.ts
@@ -1,0 +1,547 @@
+/**
+ * macOS stable code-signing for the self-updating phantombot binary.
+ *
+ * THE PROBLEM. On macOS, TCC (Privacy & Security) pins every permission grant
+ * — Full Disk Access, Documents, Desktop, Downloads, AppleEvents — to the
+ * exact code-signing identity (ultimately the cdhash) of the binary that was
+ * granted. phantombot ships ad-hoc / linker-signed (`Identifier=a.out`, no
+ * stable designated requirement), and it AUTO-UPDATES ITS OWN BINARY. Each
+ * update rewrites the file → new cdhash → every stored grant stops matching →
+ * macOS re-prompts for every protected resource, forever. That is the endless
+ * "allow access to your home folder" nagging.
+ *
+ * THE FIX. Give the binary a STABLE self-signed code-signing identity, so its
+ * designated requirement stops changing across updates. Grant once against
+ * that identity and TCC never asks again — even though the bytes change on
+ * every update. This module creates that identity and applies it.
+ *
+ * SAFETY IS THE WHOLE POINT (see the /update path). phantombot's trademark is
+ * "updates never break." So:
+ *
+ *   - The signing identity lives in a DEDICATED throwaway keychain
+ *     (`phantombot-signing.keychain`) whose password WE generate and stash in
+ *     the vault. The user's LOGIN keychain — the one with the password nobody
+ *     remembers — is NEVER touched. codesign can therefore run fully headless;
+ *     the OS never has a reason to pop the dreaded "keychain password" dialog.
+ *
+ *   - Every mutating operation is TRANSACTIONAL. `fixSigning` backs up the
+ *     binary before signing and rolls it back on any failure; if it created
+ *     the keychain/cert this run and then failed, it tears BOTH down and clears
+ *     the vault password, because the keychain's existence is the opt-in marker
+ *     the update path keys off — a half-created marker would make the next
+ *     update try to maintain a broken identity. Roll back everything this run
+ *     created, or nothing.
+ *
+ *   - Every external command runs under a TIMEOUT, so codesign can never hang a
+ *     headless update waiting on a prompt; a timeout is treated as failure and
+ *     triggers rollback.
+ *
+ *   - `resignAfterUpdate` (called from /update) is GATED on the opt-in marker
+ *     already existing and is best-effort: on any failure it restores the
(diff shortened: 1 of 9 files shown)


src/lib/macSigning.ts
```diff
@@ -0,0 +1,491 @@
...
+      if (!created.ok) {
+        await rollback();
+        return {
+          ok: false,
+          message: `Couldn't create the signing identity (${created.failedStep}). ` +
+            `No changes were left behind — phantombot still works exactly as before.`,
+          failedStep: created.failedStep,
+        };
+      }
+      createdKeychain = true;
```
kaieriksen: Blocking: `createdKeychain` is set only after `createSigningIdentity()` returns successfully. If `create-keychain` succeeds but a later step (for example `security import` or `set-key-partition-list`) fails, rollback cannot delete the keychain, but it does unset the newly-written vault password. The partially imported certificate can therefore make `isSigningConfigured()` return true on the next update while no usable password remains; subsequent updates generate a new password, fail `unlock-keychain`, and remain unable to repair automatically. Track keychain creation as soon as `create-keychain` succeeds (or return that state from `createSigningIdentity`) so every partial first-run creation is removed on rollback, and add a test for a post-create failure.

kaieriksen (changes requested):
Changes requested. The first-run rollback is incomplete if keychain creation succeeds but a later identity step fails: the keychain can remain with a certificate while its vault password is cleared, permanently breaking automatic repair on subsequent updates. Please track partial keychain creation and add the post-create failure regression test. Exact-head validation: 38 focused tests pass and typecheck passes; full-suite/CI validation is pending this blocker.

robertclawson:
Fixed in 750e12e — thanks @kaieriksen, spot-on blocker.

`createSigningIdentity()` now returns `keychainCreated`, flipped to `true` the instant `create-keychain` succeeds and propagated on every subsequent exit path (unlock, openssl-req, openssl-pkcs12, import, set-key-partition-list, and success). The caller sets `createdKeychain = created.keychainCreated` **before** the `!created.ok` branch, so rollback now tears down a partial keychain from any post-create failure — no more cert-bearing keychain stranded next to a cleared vault password, and `isSigningConfigured()` can't return true with no usable password on the next update.

Added a regression test: `create-keychain` succeeds but `set-key-partition-list` fails → asserts the keychain is deleted during rollback (≥2 `delete-keychain` calls: the idempotent pre-delete + the rollback delete), the vault password is cleared, and the binary is untouched.

Validation on exact head: `lib-macSigning` + `cli-fix-signing` + `cli-update` = 39 pass / 0 fail; `tsc --noEmit` clean.

kaieriksen (approved):
Re-review of `750e12e` — **approved**.

The keychain rollback fix is correct and complete. `createSigningIdentity()` now returns `keychainCreated: boolean` on every exit path, and the caller assigns `createdKeychain = created.keychainCreated` immediately after the `create-keychain` call succeeds — before checking `created.ok`. This ensures every partial first-run creation is torn down on rollback, which directly resolves the prior blocking issue.

**Verification:**
- 57 focused tests pass (macSigning, fix-signing, update, binaryUpdate)
- Typecheck clean (exit 0, zero diagnostics)
- GitHub CI fully green (test, typecheck, CodeQL, Windows)

PR #362, opened 2026-08-11
status: show persona PhantomChat npub on /status

robertclawson:
## What

Adds a `phantomchat:` line to `/status` showing the persona's own **npub** — its shareable PhantomChat address (what you paste into the PWA to DM the bot / add to an allowlist). Previously you had to open `phantomchat.json` to find it.

```
phantom: robbie (pid 1234, v1.1.245)
phantomchat: npub1abc…            ← new
harness: claude
chain:   → claude → pi
uptime:  3h 12m
context: ~8% (…)
active:  no
```

## Notes on the version half of the ask

Andrew asked for **version + npub**. The version was *already* shown on the `phantom:` line (`v${VERSION}`), so nothing was needed there beyond a regression assertion locking it in — this PR is effectively the npub addition.

## How it's wired

- `SlashCommandContext` gains an optional `phantomchatNpub?: string`; `handleStatus` renders the line only when set.
- **PhantomChat server** passes `npubEncode(ourHex)` — the live identity, zero disk.
- **Telegram engine** resolves it once at startup via `loadPhantomchatPersonaConfig(agentDir)?.identity.npub` — side-effect-free, returns undefined (line omitted) when the persona has no PhantomChat identity, and never creates a key.

## Safety / tests

- `bunx tsc --noEmit` clean.
- Full suite green except one **pre-existing, unrelated** `config.test.ts` Gemini-migration failure (verified it also fails on clean `main`).
- New tests: npub line shown when set, omitted when absent, version always present.


src/channels/statusProbes.ts
```diff
@@ -0,0 +1,208 @@
...
+    p: Promise<string | undefined>,
+  ): Promise<string | undefined> => {
+    try {
+      return await p;
+    } catch {
+      return undefined;
+    }
+  };
+
+  const [telegram, memory, voice] = await Promise.all([
```
kaieriksen, line 224: Blocking: these live probes have no bounded timeout. `geminiEmbed` passes no signal, and the ElevenLabs/OpenAI validators also have no timeout; because `/status` awaits `gatherStatusProbes` via `handleStatus`, a stalled provider can leave the slash command hanging indefinitely. Please wrap each probe in a shared deadline (and ideally pass an abort signal through the clients) so `/status` remains a diagnostic command even when a dependency is down.

src/channels/commands.ts
```diff
@@ -680,14 +692,32 @@ async function handleStatus(
...
+  // The persona's own PhantomChat address (npub…), when it has one. Kept on
+  // its own line so it's easy to copy into the PWA / an allowlist.
+  const npubLine = ctx.phantomchatNpub
+    ? `phantomchat: ${ctx.phantomchatNpub}\n`
+    : "";
+
+  // Live subsystem health probes — fresh on every /status (it's a
+  // troubleshooting tool). Each line is omitted when its subsystem isn't
+  // configured. See statusProbes.ts.
+  const probes = await gatherStatusProbes(ctx.config, ctx.persona);
```
kaieriksen, line 704: Blocking: this PR changes user-facing `/status` output, but the repository contribution rule requires both `README.md` and `AGENTS.md` to be updated for every user-facing behavior change. The current README still documents only the old fields (around lines 606 and 893), so please document the npub and probe lines and update both required files.

kaieriksen (changes requested):
Changes requested. The exact-head focused tests (74) and typecheck pass, and hosted CI is green, but the review is blocked by the unbounded live probes and the missing required README/AGENTS documentation updates.

robertclawson:
Thanks @kaieriksen — both blocking items addressed in 5230ee2.

**1. Unbounded live probes → shared deadline.** `gatherStatusProbes` now runs the whole fan-out under a single wall-clock deadline (`PROBE_DEADLINE_MS`, 5s). Two layers:
- The deadline is an `AbortSignal.timeout` threaded into every client that accepts one — `telegramGetMe`, `geminiEmbed` (already had `opts.signal`), and the ElevenLabs/OpenAI validators (added an optional `signal` param, backward-compatible). So the underlying socket is actually cancelled — no #135-class leak, exactly what `fetchTimeout.ts` warns about.
- `settle()` also races each probe against the same deadline as a belt-and-suspenders hard cap, in case a client stalls *before* its fetch or ignores the signal. A dead/stalled provider now drops its own line instead of hanging `/status`.
- `deadlineMs` is injectable via `StatusProbeDeps` purely for tests; added two: a probe that never resolves is capped to `undefined` (and the healthy probes still report, call returns promptly), and a fast probe still reports normally.

**2. Missing docs.** Updated both required files:
- `README.md` — both `/status` rows (Telegram command table + editor-commands table) now describe the `phantomchat: npub…` line and the live probe block, plus a paragraph in the models section covering freshness + the bounded deadline.
- `AGENTS.md` — new invariant #16: `/status` probes must stay bounded (thread `deadline` through any new probe's client), and probe output is user-facing so README rows must move with it.

Focused tests + typecheck green. (The one suite failure, `config.test.ts` › "migrates a legacy Gemini CLI chain", is pre-existing on `main` and untouched by this PR.) Ready for another look.

kaieriksen (approved):
Re-reviewed at 5230ee2. The shared 5s deadline now aborts provider requests and caps stalled probes; README.md and AGENTS.md document the user-facing /status additions. Exact-head focused tests (76) and typecheck pass, and all hosted CI checks pass.

PR #363, opened 2026-08-11
signing: register dedicated keychain in user search list so codesign can find the identity

robertclawson:
## Root cause (reproduced on a live install)

The stable code-signing re-sign has been silently failing on **every** macOS `/update` and `fix-signing`, leaving installs ad-hoc signed and still nagging. Root-caused on Matt:

`codesign --sign phantombot-codesign` resolves the identity **by name through the USER keychain search list** — the `--keychain <path>` argument does *not* scope that lookup. We create a dedicated `phantombot-signing.keychain` but never add it to the search list, so:

```
codesign: phantombot-codesign: no identity found
```

In a headless context (launchd after `/update`, or SSH) that stalls → 60s timeout → rollback. The safety net (#361) fired correctly every time — clean teardown, working binary — but the stable identity never landed.

**Proof:** on the exact headless SSH session, `codesign --sign … --keychain <path>` with the keychain *not* in the search list → `no identity found`; after `security list-keychains -d user -s <old…> <kc>`, the same codesign signs first try (`Identifier=dev.phantombot`).

Two earlier hypotheses were **checked and killed**: the `set-key-partition-list` call is already present and correct, and cert trust is a red herring (codesign signs fine with an untrusted-but-findable identity).

## Fix

- **`ensureKeychainInSearchList`** — idempotent add of the dedicated keychain to the user search list, run right before codesign. Tolerates macOS's `-db` suffix (`create-keychain foo.keychain` materialises `foo.keychain-db`) so it never appends a duplicate. Returns the prior list.
- **`fixSigning` rollback** — if we changed the search list this run, restore the prior list *before* deleting the keychain, so no dangling entry is left. Roll back everything this run created, or nothing.
- Login keychain and the system domain are never touched; only the user search list, only our own keychain.

## Tests
- 8 new unit tests (injected CommandRunner, no Mac): parse, idempotent no-op with `-db` tolerance, add-when-absent preserving order, read-failure surfaces without mutating, and `fixSigning` rollback restores the prior search list on codesign failure.
- `tsc` clean. Full suite: 2651 pass / 1 fail — the pre-existing unrelated `config.test.ts` Gemini-embeddings failure that's on `main` (verified by stashing this diff).

Follow-up to #361/#362. Once merged it cuts a release; `/update` on an affected Mac should then land the stable `dev.phantombot` identity and stop the TCC nagging for good.

### DIFF
--- a/src/lib/macSigning.ts
+++ b/src/lib/macSigning.ts
@@ -23,6 +23,10 @@
  *     the vault. The user's LOGIN keychain — the one with the password nobody
  *     remembers — is NEVER touched. codesign can therefore run fully headless;
  *     the OS never has a reason to pop the dreaded "keychain password" dialog.
+ *     The dedicated keychain IS added to the USER keychain search list (never
+ *     the login keychain, and never the system domain) because codesign
+ *     resolves the signing identity by name through that list rather than the
+ *     `--keychain` argument; on any failure the prior search list is restored.
  *
  *   - Every mutating operation is TRANSACTIONAL. `fixSigning` backs up the
  *     binary before signing and rolls it back on any failure; if it created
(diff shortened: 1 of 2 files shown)


kaieriksen (approved):
Reviewed at exact head b4dddae0cdb17de67c8155bd9867cf46d698ea54. The keychain search-list registration is idempotent, tolerates macOS's -db suffix, and restores the prior user list before rollback teardown. Validation: 16 focused signing tests pass, typecheck passes, 2649/2654 full-suite tests pass locally with only the existing PiHarness environment tests failing, and all hosted CI checks are green.

PR #364, opened 2026-08-11
signing: pin openssl to /usr/bin (LibreSSL) + /update resign dogfood command

robertclawson:
## The bug this fixes

Every macOS `/update` re-sign has been failing at the **import** step — the binary stays ad-hoc, so TCC keeps re-prompting for Full Disk Access / home folder after every update. Verified live on Matt.

**Root cause (reproduced deterministically in the real launchd context):** `macSigning.ts` shelled out to a **PATH-resolved `openssl`**. phantombot's launchd `PATH` puts `/opt/homebrew/bin` ahead of `/usr/bin`, so `openssl` = Homebrew **OpenSSL 3.x**. Its `pkcs12 -export` writes the PKCS#12 MAC/PBE with modern algorithms that Apple's LibreSSL-era `security import` rejects:

> `SecKeychainItemImport: MAC verification failed during PKCS12 import`

→ import fails → transactional rollback → binary stays ad-hoc.

Reproduced by bootstrapping the exact import sequence into a genuine `gui/501` Aqua launchd session:
- `/usr/bin/openssl` → **import OK** → partition-list + codesign succeed (`Identifier` set, not `a.out`)
- Homebrew `openssl` (what PATH actually picks) → **`MAC verification failed`, rc=1** — Matt's exact error

Earlier repros over SSH kept typing `/usr/bin/openssl` by hand (LibreSSL), which masked that phantombot resolves a *different* openssl.

## The fix

Pin both openssl calls to the exported `OPENSSL_BIN` = `/usr/bin/openssl` (system LibreSSL, always present, import-compatible p12s). `-legacy` is **not** the fix — that flag is an OpenSSL-3-ism and errors on LibreSSL.

This is a **distinct** bug from #363's keychain-search-list fix (that one is downstream at the codesign step). Both are needed.

## `/update resign` — dogfood + repair button

The merge→release→`/update` devloop is too long to iterate on a launchd-context bug. So this adds an owner-facing `/update resign` chat subcommand that re-signs the **current on-disk binary in place**:

- **No** download, reinstall, restart, or version change — re-sign only
- Runs the **same** routine `/update` applies post-swap (`runFixSigning` → `fixSigning`), so it's a faithful dogfood of the re-sign path in the real running-service context
- Doubles as a manual repair button if a macOS update ever invalidates the signature
- **macOS-only**; on Linux/Windows it's a friendly no-op (exit 0), reusing `fix-signing`'s exact platform guard and prose
- Bare `/update` is unchanged (full self-update)

## Tests

- `macSigning`: asserts both openssl invocations use `/usr/bin/openssl`, never bare `openssl`
- `commands`: 3 dispatch tests for `/update resign` (non-macOS no-op without config, case-insensitive, macOS path reaches the signing routine)
- Typecheck clean; full suite 2655 pass / 1 fail — the pre-existing unrelated `config.test.ts` Gemini migration failure that also fails on clean `main`.

## Docs

- README `/update` row documents `resign`
- AGENTS.md invariant #17 (openssl pin + the dogfood command)


kaieriksen (approved):
Reviewed exact head c4a6aead56493f8375618819b8ebeb5b465558df. No defects found. Targeted signing/command tests pass, typecheck passes, hosted CI and CodeQL are green, and the x64 build plus compiled CLI smoke test pass.

Issue #360, opened 2026-08-11
Chain-exhaust on double harness wedge (Windows): harden process reaping, graceful chain-exhaust message, decay idle budget


andrewagrahamhodges:
## Problem

When both harnesses in the failover chain (`[claude, pi]`) wedge on the same turn, phantombot exhausts the chain and surfaces the raw harness error to the user:

```
[error] pi timed out after 300000ms with no output (likely wedged on a tool call)

(phantombot declined this turn.)
```

### What actually happened (Megan, 2026-08-11, live ACP window)

Chain was correctly bound as `[claude, pi]`, this is **not** a stale-chain bug. Sequence:

```
12:02:10  claude.invoke killed: idle (300000ms)              ← claude wedged on a tool call
12:02:10  processGroup: taskkill failed pid 14972 status 255 ← couldn't reap it cleanly
12:02:10  orchestrator: falling through (claude)             ← failover to pi fired correctly
12:12:39  pi.invoke killed: idle (300000ms)                  ← pi ALSO wedged on the same turn
          → chain exhausted → raw "pi timed out… declined this turn" surfaced
```

**Failover worked as designed.** The problem: claude wedged, Windows failed to reap it cleanly (`taskkill status 255`), and the orphaned process most likely held a socket, file lock, or port that then poisoned pi, which is why the "unlikely" second wedge happened on the same turn. Each harness burned the full 300s idle timeout, giving ~10 minutes of dead air, then a raw error that reads like a crash.

### Why the existing heartbeat didn't save it

- The idle watchdog resets on stdout activity, including pi's `tool_execution_start` / `tool_execution_update` events (that's the existing tool-keepalive, plus #351's `toolTimeoutMs` hard cap).
- Today's wedge emitted **zero** output, not even a `tool_execution_update`. Genuinely blocked inside a synchronous tool call, so there was nothing to heartbeat on. The watchdog correctly idle-killed at 300s; there was just nothing to observe.

### Architectural note

Tool execution lives **inside** the harness subprocess (opaque child spawned via `spawnInNewSession`). phantombot never invokes tools on the harness's behalf, so a per-tool timeout on our side is **not possible**; the true fix for the hang itself belongs inside the tool/MCP layer. Our levers are the idle watchdog in `harnessRunner.ts` and the Windows process-kill path.

## Agreed direction (steered after discussion with Jeroen / Beart)

This is a **Windows-only** issue, and the scope is now:

[...]

PR #366, opened 2026-08-12
harness: startup timeout + mcpMode:none for tool-less transports (#360)

robertclawson:
Fixes the claude-only Windows/VSCode turn wedges tracked in #360.

## Root cause (verified against Megan's v1.1.247 logs)

A foreground `claude --print` turn blocks its MCP `initialize` handshake on the loopback `phantombot mcp proxy` subprocess it spawns as its MCP server. Under Windows SQLite lock contention the proxy can stall opening its store, so the handshake never resolves — claude emits **nothing**, and the turn burns the full 300s idle window before failing over.

**Pi never spawns a proxy** (built-in tools) — that's the claude-vs-pi differentiator, and why swapping Megan to pi "fixed" it: the orchestrator already auto-falls-through to pi on the wedge, so the manual swap just skipped the 5-minute wait.

This reframes the issue away from the original "harden Windows process reaping" hypothesis. Megan's logs show **zero** Job-Object / orphan / reparent signatures — the harness isn't leaving orphans, it's synchronously hanging on init.

## The fix (two parts)

**1. Startup timeout (`harnessStartupTimeoutMs`, default 60s).**
A new `KillCoordinator` timer bounding time-to-first-stdout-byte, distinct from the idle timer (which only bites once output has begun). Cleared on the first stdout byte via `firstOutput()`. A turn wedged before any output now fails over in ~60s instead of 300s. New recoverable `KillCause` `"startup"`. Wired through every foreground path; configurable via `PHANTOMBOT_HARNESS_STARTUP_TIMEOUT_MS` / `harness_startup_timeout_s`. The default is deliberately generous so a healthy-but-slow cold start under the *same* load that triggers the wedge is never false-killed.

**2. `mcpMode:"none"` on the durable-facts and threat-judge transports.**
Both are persona-less, tool-less completions that never need MCP, but they set `toolsMode:"none"` **without** `mcpMode:"none"` — so the claude harness took the foreground branch and spawned the proxy anyway. Setting `mcpMode:"none"` runs them with zero MCP servers, matching the intended contract and keeping background extraction *and* every untrusted-input screen off the proxy-spawn path.

## Deliberately NOT included (per the #360 discussion)

- **Idle-budget decay down the chain** — redundant once startup fails fast.
- **Canned chain-exhaust message** — a hardcoded English string with no i18n support on the app. Surface the structured error and let fallthrough do its job instead.

## Tests

Startup-timer unit tests (fires on silence, cancelled by first output, disabled when unset), `killCauseToErrorChunk` `"startup"` branch, and a `runHarnessProcess` integration test proving the recoverable error surfaces fast. `tsc --noEmit` clean; existing suites green (the 5 unrelated failures on `main` — gemini-chain migration + `acquireRunLock`/memory-store-concurrent sandbox file-lock — pre-date this branch).

## Honest caveat

Megan's box is healthy now (proxy answers `initialize` in ~347ms), so the "proxy stalled under lock contention" step is strong inference from the co-timed `database is locked` error storm, not a captured proxy-init failure. The startup timeout is the right mitigation regardless of the exact stall source — it bounds *any* pre-output wedge.

### DIFF
--- a/src/channels/core/engine.ts
+++ b/src/channels/core/engine.ts
@@ -490,6 +490,7 @@ export async function runTelegramServer(
           memory: input.memory,
           idleTimeoutMs: input.config.harnessIdleTimeoutMs,
           hardTimeoutMs: input.config.harnessHardTimeoutMs,
+          startupTimeoutMs: input.config.harnessStartupTimeoutMs,
           // Reaction from an allow-listed principal → trusted, so the memory
           // capture may write. runReactionTurn skips the threat screen for it.
           trusted: true,
@@ -1267,6 +1268,7 @@ async function processChatMessage(
       memory: input.memory,
       idleTimeoutMs: input.config.harnessIdleTimeoutMs,
       hardTimeoutMs: input.config.harnessHardTimeoutMs,
+      startupTimeoutMs: input.config.harnessStartupTimeoutMs,
       signal: controller.signal,
       // Security perimeter: the ONLY place `trusted: true` originates.
       // True iff the sender is an allow-listed principal (see the
--- a/src/channels/core/reactions.ts
+++ b/src/channels/core/reactions.ts
@@ -240,6 +240,7 @@ export interface RunReactionTurnInput {
   memory: MemoryStore;
   idleTimeoutMs: number;
   hardTimeoutMs?: number;
+  startupTimeoutMs?: number;
   /**
    * Trust provenance — true only when the reactor is an allow-listed principal.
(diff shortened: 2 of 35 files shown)


kaieriksen (approved):
Validated exact head 92e117898cb00c2618787a5e3811bc6cd1110412.

Focused harness/config/MCP tests: 159 passed, 319 assertions. Typecheck clean. Full suite: 2,659 passed, 2 skipped; the two PiHarness routing failures are pre-existing host-environment failures. Hosted CI and CodeQL are green.

PR #367, opened 2026-08-12
tmp: move harness temp writes out of /tmp into the persona dir (#365)

### LINKED ISSUE #365: [Bug/Feature] Turns fail silently when /tmp fills (EDQUOT): no disk-space recovery, no tmp cleanup in heartbeat/nightly
## Symptom

On low-memory hosts `/tmp` is a tmpfs sized at 50% of RAM (e.g. ~1.7G on a 3.2GiB 2012 MacBookPro, Ubuntu 25.10). When it fills — in our case leftover meeting recordings (~1.2G) — every turn fails on both Telegram and PhantomChat with:

```
EDQUOT: unknown error, write
```

The bot goes fully silent for hours until an operator manually frees space. The existing recovery reply (`generateRecoveryReply`, `src/orchestrator/recovery.ts` + `src/channels/core/engine.ts:1486-1509`) can't help: the recovery turn also writes, hits the same EDQUOT, and falls into the `unrecoverable` silent path.

phantombot itself writes to `/tmp` and is directly affected:
- `src/lib/harnessArgvFiles.ts:58` — `mkdtemp(join(tmpdir(), "phantombot-harness-"))`
- `src/lib/runLock.ts:87-90` — `/tmp/phantombot-${uid}.run.lock`
- `pi-extension spawnPi.ts` — `phantombot-route-*` temp files for `--append-system-prompt`

Neither the heartbeat (`src/cli/heartbeat.ts` — memory drawers/FTS/tails only) nor the nightly 5 cognitive stages (`src/cli/nightly.ts`, see #295) touch disk space.

On our setup `live-memory.sqlite` also lives in `/tmp`, so the shared memory itself is at risk when the tmpfs fills.

## Proposed

1. **Reactive:** when a turn fails with EDQUOT/ENOSPC, run emergency cleanup of phantombot's own tmp residue (`phantombot-harness-*`, `phantombot-route-*`, stale runLock) and retry once before falling to recovery.
2. **Heartbeat:** cheap free-space check; below a configurable threshold, sweep phantombot-owned stale tmp files (safe — they're mkdtemp dirs).
3. **Nightly:** extend the existing staged maintenance with a disk/tmp cleanup stage.

**Nice to have:** configurable thresholds and cleanup policy (move-to-custody vs delete), plus a `phantombot maintenance` command usable manually or via the existing timers.

## Environment

- phantombot 1.1.239 (reproduced); verified `main` (v1.1.247) still has no tmp/disk cleanup
- Ubuntu 25.10, systemd user service, tmpfs /tmp (usrquota → EDQUOT)

robertclawson:
Closes #365.

Implements the agreed design (Andrew's file-by-file spec + Kai's per-request `tmpBaseDir` refinement): **stop using the shared system `/tmp`; write harness temp files into the persona's own dir, and sweep aggressively.**

## Why
On a small tmpfs `/tmp` (50% of RAM on low-mem hosts), a full `/tmp` returns `EDQUOT` on every write — harness temp files fail, the recovery reply can't be written, and the daemon goes **silently dark for hours**. `/tmp` is also shared across personas on a multi-persona box: isolated *contents* (0700 mkdtemp), but shared *free space*, so one persona can starve the others.

## What
Harness temp writes now land in `<personaDir>/tmp/` — real disk, full per-persona isolation (ownership, perms, **and free space**), and it keeps working when `/tmp` is full.

| File | Change |
|------|--------|
| `harnesses/types.ts` | `HarnessRequest.tmpBaseDir?: string` |
| `orchestrator/turn.ts` | pass `tmpBaseDir: join(agentDir, "tmp")` |
| `lib/harnessArgvFiles.ts` | `createHarnessTempDir(baseDir?)`, `personaTmpDir()`, age-gated `cleanupPersonaTmpDir()` |
| `harnesses/pi.ts` | pass `req.tmpBaseDir`; export `PHANTOMBOT_TMP_DIR` to the route extension |
| `harnesses/claude.ts` | pass `req.tmpBaseDir` |
| `pi-extension/.../spawnPi.ts` | honour `PHANTOMBOT_TMP_DIR`, fall back to `os.tmpdir()` |
| `lib/piExtensionAssets.generated.ts` | regenerated |
| `lib/piRouting.ts` | `ENV_PHANTOMBOT_TMP_DIR` constant |
| `index.ts` | startup sweep of active persona's tmp dir |
| `lib/runLock.ts` | POSIX fallback `/tmp` → `$HOME/.cache/phantombot/run/` |

## Safety — "can't break anyone"
- **Per-request `tmpBaseDir`, not a process-wide `TMPDIR`** (Kai's point) — no leak across personas / unrelated children.
- **Run lock untouched by the sweep.** After #356/#363 the lock is an OS advisory lock we never unlink; it lives in `$XDG_RUNTIME_DIR` (or the new `~/.cache` fallback), **never** in the persona tmp dir, so the sweep can't reach it.
- **Age-gated sweep (>1h).** A live in-flight turn's harness dir is younger than the gate, so it's never nuked mid-spawn. Best-effort — a sweep failure never wedges startup.
- **Graceful fallback.** `baseDir` omitted → `os.tmpdir()`, so tests and degraded/no-persona paths are unchanged. `XDG_RUNTIME_DIR` still wins for the run lock on systemd hosts (this VPS + kw-openclaw) — zero behaviour change there.

## Out of scope (deliberately)
- No `phantombot maintenance` command (deferred, per thread).
- `live-memory.sqlite` relocation/WAL — not placed in `/tmp` by phantombot's defaults; separate concern, not in Andrew's spec.

## Tests
`bun tsc --noEmit` clean. New/updated unit tests: baseDir routing + first-use mkdir, `personaTmpDir`, `cleanupPersonaTmpDir` age-gate (stale reaped, fresh kept, missing-dir no-throw), updated runLock fallback assertion. Full suite: 2665 pass, 2 skip, **1 pre-existing unrelated failure** (`loadConfig — migrates a legacy Gemini CLI chain`, fails identically on clean `main`).

🐶 — Robbie


src/lib/harnessArgvFiles.ts
```diff
@@ -49,13 +50,71 @@ export interface HarnessTempDir {
...
+    names = readdirSync(dir);
+  } catch {
+    return; // dir doesn't exist yet — nothing to sweep
+  }
+  const now = Date.now();
+  for (const name of names) {
+    const full = join(dir, name);
+    try {
+      if (now - statSync(full).mtimeMs > maxAgeMs) {
+        rmSync(full, { recursive: true, force: true });
```
kaieriksen: This sweep deletes every entry older than the TTL under `<personaDir>/tmp`, but the agreed contract is to delete only phantombot-owned residue (`phantombot-harness-*`, `phantombot-route-*`, etc.). A user or another component can place data in this directory and have it recursively removed on the next startup. Filter by the owned prefixes (and add a regression test that an old unrelated entry survives) before approving.

kaieriksen (changes requested):
Changes requested: the startup sweep recursively deletes any old entry under the persona tmp directory instead of restricting deletion to phantombot-owned temp prefixes. This violates the agreed no-user-data-deletion safety contract. The inline blocker includes the required regression test. Exact-head checks: 38 isolation/lock tests pass, typecheck passes, affected harness/orchestrator suites pass apart from three ambient PHANTOMBOT_PI_API_KEY test leaks; hosted CI is fully green.

lenaparkhodges (changes requested):
Concurring with Kai's blocker — I traced the sweep myself and it's exactly as he says.

**Blocker: `cleanupPersonaTmpDir` deletes ANY entry older than 1h, not just phantombot-owned prefixes.** The docstring asserts "only phantombot's own residue lives here", but nothing enforces that — `<personaDir>/tmp` is the persona's own home-adjacent space, and agents (or operators) will reasonably treat a dir named `tmp` in their working home as scratch. First time one of us drops a scratch file there and a CLI invocation starts an hour later, it's gone. That violates the agreed no-user-data-deletion contract and Andrew's "I want this to not break anyone". Fix is cheap: gate the rm on the known prefixes (`phantombot-harness-`, `phantombot-route-`) — everything else survives regardless of age. Kai's inline regression test covers it.

Verified from the diff at head `1e2471b` (no local bun on this VM; relying on Kai's exact-head test run + my read):

**Everything else is solid:**
- Per-request `tmpBaseDir` threaded through types → turn → both harnesses → `PHANTOMBOT_TMP_DIR` for the route extension. Correctly NOT a process-wide TMPDIR — no cross-persona leak.
- Graceful fallback to `os.tmpdir()` everywhere the base is omitted — tests and degraded paths unchanged.
- Run lock correctly out of the swept dir; POSIX fallback moving `/tmp` → `~/.cache/phantombot/run/` is right.
- Age-gate + best-effort sweep wired into startup without wedging it.

**Two non-blocking notes:**
1. **Run-lock upgrade window (non-systemd hosts only):** an old-binary daemon holding the `/tmp` lock + a new-binary starter checking `~/.cache` won't see each other → brief double-daemon window on upgrade. Self-heals on restart, and systemd hosts (this VPS, kw-openclaw) use XDG_RUNTIME_DIR so zero change there. Worth a line in the PR body at most.
2. **1h gate vs long turns:** a turn exceeding 1h has a "stale" harness dir (mtime set at spawn); a concurrent CLI start would sweep it mid-turn. Harmless once the child has read the files at spawn, but if any harness re-reads @payload.md late it would break. Consider whether max turn duration is provably < 1h, or bump the gate.

Fix the prefix gate and this is an approve from me. 📊

lenaparkhodges (changes requested):
Direction is right — per-persona tmp on real disk, per-request tmpBaseDir instead of process-wide TMPDIR, and the graceful os.tmpdir() fallback are all the correct shape. But I can't approve while two deletion-safety gaps stand:

**1. (Blocking, concurring with @kaieriksen) The sweep is not prefix-restricted.** cleanupPersonaTmpDir recursively rm's ANY entry older than 1h under <personaDir>/tmp. The doc comment asserts "only phantombot's own residue lives here" but nothing enforces that — the dir is a plausible scratch space for the persona's own agent or a user, and anything they leave there is silently deleted on the next startup. Filter to owned prefixes (phantombot-harness-*, phantombot-route-*) and add the regression test Kai asked for.

**2. (Blocking) The 1h age gate can delete a LIVE turn's temp dir.** Default harnessHardTimeoutMs is 3_600_000 — exactly equal to the 1h sweep gate — and it's operator-configurable upward. The harness dir's mtime is set at creation (system prompt/conversation files are written once, up front) and then not touched, so a turn that legitimately runs past 1h has its in-use temp dir reaped by any other phantombot process's startup sweep. Derive the gate from the configured hard timeout plus margin, or use a much larger fixed TTL (e.g. 24h) — residue is cheap, a mid-turn EDQUOT/ENOENT is not.

Non-blocking: the runLock fallback move to ~/.cache is good; nice that XDG_RUNTIME_DIR behaviour is unchanged on systemd hosts. Tests look solid otherwise. Fix the two gates above and this is a clean approve from me. 📊

robertclawson:
Both blockers fixed in `61d2749`, plus the Windows hardening Andrew flagged.

**1. Prefix-gate the sweep (Kai + Lena, blocking).** `cleanupPersonaTmpDir` now deletes **only** phantombot-owned entries — `phantombot-harness-*` / `phantombot-route-*`, via a new exported `OWNED_TMP_PREFIXES`. Anything else in `<personaDir>/tmp` (an agent/operator scratch file) survives regardless of age. The no-user-data-deletion contract is now enforced in code, not just asserted in the docstring. Regression test added: a **week-old** non-phantombot dir + file both survive the sweep.

**2. 1h gate can reap a live turn (Lena, blocking).** Fattened the TTL 1h → **24h** (`PERSONA_TMP_SWEEP_MAX_AGE_MS`). Since default `harnessHardTimeoutMs` is exactly 1h and operator-bumpable, the old gate could delete an in-use harness dir mid-run. 24h clears any plausible turn by a wide margin; residue is cheap.

**3. Windows (per Andrew).** The sweep + per-turn cleanup are the one spot POSIX and Windows diverge: on Windows an open handle (the live child, or an AV/search-indexer scan) makes an immediate recursive rm throw `EBUSY`/`EPERM`/`ENOTEMPTY`, where POSIX just unlinks the open file. Added `maxRetries`/`retryDelay` to both `rmSync` (sweep) and `rm` (cleanup) to ride out the transient case; a genuinely in-use dir fails the rm and is swallowed. Net effect: **on Windows a live turn is protected twice** — age gate *and* open-handle lock. No-op cost on POSIX. Note this path isn't Windows-only — the pi harness routes system-prompt + payload through the persona tmp on every platform, so `test-windows` exercises it.

Non-blocking notes acknowledged: the run-lock upgrade window only affects non-systemd hosts (ours all use `XDG_RUNTIME_DIR`, zero change) — captured in the PR body.

Typecheck clean; new tests green (prefix-gate regression, route-dir sweep, 24h default, explicit-override). Full suite: 2668 pass, the single failure is the pre-existing unrelated Gemini-config migration test on `main`. Re-requesting your review 🐶

kaieriksen (approved):
Re-reviewed at exact head `61d27499150f47fb57b5619ea6082c51c14386a8`.$nnPrior blockers are resolved: the startup sweep is restricted to phantombot-owned prefixes and the TTL is 24h, protecting long-running turns. Windows cleanup retries are present for both startup sweep and per-turn cleanup.$nnVerification: 42 focused temp/lock/harness tests pass, local Bun typecheck passes, and hosted CI including test-windows and CodeQL is green.

lenaparkhodges (approved):
Both blockers properly fixed in 61d2749 — re-approving. ✅

1. **Prefix-gate:** sweep now only touches `phantombot-harness-*`/`phantombot-route-*` via `OWNED_TMP_PREFIXES`; verified the prefixes match reality (`spawnPi.ts` mkdtemp + `createHarnessTempDir`). The week-old scratch dir/file regression test is exactly the contract I wanted enforced in code.
2. **24h TTL:** fat gate clears any plausible turn (incl. operator-bumped hard timeouts) — good call, residue is cheap.

Windows `maxRetries`/`retryDelay` on both rm paths is a clean double-protection with zero POSIX cost. Tests: 12/12 pass locally on 61d2749. Ship it. 🐶

lenaparkhodges (approved):
**Re-review: approving.** ✅

My blocker is resolved, and better than I asked:

- **Prefix-gated sweep** — `OWNED_TMP_PREFIXES` (`phantombot-harness-*` / `phantombot-route-*`) enforced in code via `isOwnedTmpEntry`, so anything else in `<personaDir>/tmp` survives regardless of age. The no-user-data-deletion contract is now structural, not documentary.
- **Regression test landed** — `never deletes non-phantombot entries, even when ancient` (7-day-backdated scratch dir + file survive) is exactly the test Kai requested.
- **24h TTL** — addresses my long-running-turn concern head-on. 1h could have reaped a live turn on an operator-bumped hard timeout; 24h clears any plausible turn by a wide margin. Right call on residue-cheap vs mid-turn-ENOENT.
- **Windows safety** — `maxRetries` on both the sweep and the per-turn `cleanup()` rides out AV/indexer handle lag; a genuinely in-use dir just fails the rm and is left. Double protection on Windows (age gate + open-handle lock) is the right layering.
- **`PHANTOMBOT_TMP_DIR` as our own env var, not `TMPDIR`** — no process-wide leak across personas or unrelated children. Verified threaded per-child from `turn.ts` → harness env.
- **Run lock out of the swept dir** (`~/.cache/phantombot/run` fallback on non-systemd) — confirmed the sweep can never touch it.
- Sweep is best-effort at startup with `log.warn` on failure — can't wedge boot. `never throws when the tmp dir does not exist` test covers the fresh-box path.

CI green on all six checks incl. test-windows; Kai approved at the same head (`61d2749`).

One non-blocking carryover note from my first pass: the run-lock fallback path changed from `/tmp` to `~/.cache/phantombot/run`, so on non-systemd hosts there's a brief upgrade window where a running old daemon holds the `/tmp` lock while a new starter creates the `~/.cache` one. Transient and self-correcting (Telegram getUpdates conflict), but worth a restart-the-daemon note in release notes for those installs.

Good to merge.

PR #369, opened 2026-08-13
logger: route all levels to stderr, keep stdout clean for ACP/MCP wire

robertclawson:
## Root cause

The structured logger (`src/lib/logger.ts`) wrote `info`/`debug` to `process.stdout` and `warn`/`error` to `process.stderr`. Under the ACP stdio server, **stdout IS the JSON-RPC protocol channel** — so every `log.info()` call mid-turn dumped a stray non-JSON-RPC line onto the wire.

Zed couldn't route it (`-32600 Invalid request` / `id=Null`), which broke request/response correlation and caused the agent to re-issue the same tool call endlessly — the loop you see in the Zed agent panel.

## Proof

On Aug 5, Zed's `Transport parse error` payload contained the **literal phantombot log line** verbatim:

```
{"ts":"...","msg":"orchestrator: trying harness","harnessId":"pi",...}
```

This is a `log.info()` call from `src/orchestrator/fallback.ts:150` — info level → stdout → wire corruption.

## Why it only hits ACP/editor

The ACP server's own logging already used stderr correctly (`logErr` sink in `server.ts`). But the **global logger** used by the orchestrator, harnesses, harness runner, and process group didn't know it was running under ACP. On Telegram/phantomchat, stdout isn't a protocol channel, so the stray writes are harmless (captured by launchd/systemd to a log file).

## The fix

Route **all** levels (debug, info, warn, error) to stderr. stdout is now byte-clean for all log calls, safe for both ACP and MCP entrypoints where stdout is a protocol channel.

**2 files changed, 25 insertions, 8 deletions:**
- `src/lib/logger.ts` — one-line change: `process.stderr.write(...)` for all levels
- `tests/lib-logger.test.ts` — update 3 tests that asserted info→stdout, add regression test proving stdout stays empty

## Also explains

- **Why only Zed/VS Code hit it** — Telegram/phantomchat don't use stdout as a protocol channel
- **Why the `id=Null` frames appear** — phantombot's `jsonRpcError(null, ...)` fires on the corrupt lines it receives back from Zed's error response
- **Why `pi.invoke killed: aborted` correlates** — the idle-kill warn goes to stderr (correct), but the info-logs around it go to stdout (corrupting the wire)

## Not related to

- #360/#366 (MCP proxy handshake / Windows SQLite lock contention) — different bug
- Process reaping / orphaned children — the harness children use explicit pipes, not inherited stdio

CI: `tsc --noEmit` clean, logger + ACP test suites all pass (42/42).

### DIFF
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -1,5 +1,9 @@
 /**
- * Tiny structured logger. Writes JSON lines to stdout.
+ * Tiny structured logger. Writes JSON lines to stderr.
+ *
+ * ALL levels (debug, info, warn, error) go to stderr. stdout is a protocol
+ * channel for two entrypoints (ACP stdio server + MCP proxy) — any stray
+ * write there corrupts the JSON-RPC wire and the editor drops the connection.
  *
  * Replace with pino / winston / etc. if you outgrow this. The interface here
  * is intentionally narrow so the swap stays small.
@@ -37,9 +41,9 @@ function emit(level: LogLevel, msg: string, fields?: Record<string, unknown>): v
     msg,
     ...(fields ?? {}),
   };
-  // stderr for warn/error so log shippers can split if desired.
-  const stream = level === "warn" || level === "error" ? process.stderr : process.stdout;
-  stream.write(redactForLog(JSON.stringify(line)) + "\n");
+  // ALL levels go to stderr. stdout is a protocol channel for two entrypoints
+  // (ACP stdio server + MCP proxy) — any stray write there corrupts the wire.
+  process.stderr.write(redactForLog(JSON.stringify(line)) + "\n");
 }
 
 export const log = {
--- a/tests/cli-tick.test.ts
+++ b/tests/cli-tick.test.ts
@@ -51,20 +51,20 @@ function installFetchTrap(): { calls: FetchCall[]; restore: () => void } {
   return { calls, restore: () => { globalThis.fetch = original; } };
 }
 
-function captureStdout(): {
+function captureStream(which: "stdout" | "stderr"): {
   lines: string[];
   restore: () => void;
 } {
   const lines: string[] = [];
-  const original = process.stdout.write;
-  process.stdout.write = ((chunk: unknown) => {
+  const original = process[which].write;
+  process[which].write = ((chunk: unknown) => {
     lines.push(String(chunk));
     return true;
   }) as typeof process.stdout.write;
   return {
     lines,
     restore: () => {
-      process.stdout.write = original;
+      process[which].write = original;
     },
   };
 }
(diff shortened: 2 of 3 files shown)


src/lib/logger.ts
```diff
@@ -37,9 +41,9 @@ function emit(level: LogLevel, msg: string, fields?: Record<string, unknown>): v
     msg,
     ...(fields ?? {}),
   };
-  // stderr for warn/error so log shippers can split if desired.
-  const stream = level === "warn" || level === "error" ? process.stderr : process.stdout;
-  stream.write(redactForLog(JSON.stringify(line)) + "\n");
+  // ALL levels go to stderr. stdout is a protocol channel for two entrypoints
+  // (ACP stdio server + MCP proxy) — any stray write there corrupts the wire.
+  process.stderr.write(redactForLog(JSON.stringify(line)) + "\n");
```
kaieriksen, line 46: This redirects every info/debug line to stderr, but `runTick`'s existing observable log contract is still stdout. Exact-head reproduction: `tests/cli-tick.test.ts` fails at line 213 because its stdout capture is empty; the full suite is 2667 pass / 2 skip / 3 fail (the other two are ambient Pi-key failures). Please update that affected contract/test or preserve the intended CLI sink before merging.

kaieriksen (changes requested):
Request changes: this sink migration breaks an existing task-runner log contract. On this exact head, `tests/cli-tick.test.ts` still captures `process.stdout` and now receives no lifecycle logs; the focused run fails 1 test and the full suite fails the same test. Please update the affected capture/consumer contract in this PR (and keep the protocol stdout paths covered) before merge.

robertclawson:
Updated — Kai's catch was spot-on: `runTick`'s lifecycle logs went through the global logger, which now routes to stderr. The tick test was asserting those logs on stdout (the old contract), so it broke.

Fix: renamed the capture helper to `captureStream(which)` and switched the tick test to capture `"stderr"` — same pattern the logger tests already use. All 18 tick tests + 7 logger tests pass, `tsc --noEmit` clean.

robertclawson:
Thanks @kaieriksen — good catch, and already addressed on the current head.

Your review was against `84d3455`, where `tests/cli-tick.test.ts` still captured `process.stdout` and so came up empty after the sink migration. Commit `c174374` updates that contract: the capture helper is now `captureStream("stdout" | "stderr")` and the `runTick` lifecycle test asserts against `captureStream("stderr")`, matching where the logs now land. The protocol stdout paths stay covered by the new stdout-stays-empty regression test in `tests/lib-logger.test.ts`.

CI on `c174374` is green — run 31695447623: test / test-windows / typecheck / CodeQL all pass (2667 pass / 2 skip / 0 fail; the earlier 2 ambient Pi-key failures are also gone). Re-requesting review.

kaieriksen (approved):
Approved at exact head c174374b6a885079972adc0ccdbb6072551550b4. The logger sink regression is fixed: runTick captures stderr, stdout protocol cleanliness is covered across all levels, focused tests pass 25/25, typecheck is clean, and hosted CI/CodeQL are green. Full suite is 2668 passed, 2 skipped, with 2 known Pi-key environment failures unrelated to this PR.

PR #370, opened 2026-08-13
phantomchat: NIP-42 relay AUTH + publish read-back verification (#368)

### LINKED ISSUE #368: [Bug/Feature] PhantomChat channel: no NIP-42 AUTH support — relays requiring AUTH silently drop bot events (OK:true, never stored)
### Symptom

On a deployment where the Nostr relay requires NIP-42 AUTH (`nip42_auth = true` in
nostr-rs-relay), PhantomChat bots appear to reply but their messages never reach
the recipient. The bot's turn completes normally (no error surfaced), yet:

- the relay answers `OK:true` to the publish,
- **the event is never stored** (silent drop),
- any client that listens only to that relay — e.g. a bridge or another bot —
  never sees the reply.

The result is a silent bot↔bot disconnect: inbound messages arrive, outbound
replies vanish. In our production case this looked like "pepa received the
message but her reply never came back".

### Root cause

The PhantomChat channel does not implement NIP-42 (`NIP-42` relay AUTH, kind
`22242`) at all:

- `src/channels/phantomchat/transport.ts` uses `nostr-tools`' `SimplePool`
  (`SimplePoolPhantomchatTransport`, transport.ts:315) with `nostr-tools
  2.23.3` pinned in `package.json`.
- `SimplePool` never answers relay `AUTH` challenges, and there is no
  `auth-required` handling anywhere in the channel.
- A `grep` for `nip-42`/`nip42` across `src/` on current `main` returns zero
  hits. The only "AUTH" references in the channel are the internal pubkey
  **allowlist gate** (server.ts "AUTH GATE") and the Blossom upload header
  (NIP-24242) — both unrelated to NIP-42 relay auth.

Per NIP-42, a relay that requires AUTH should answer an unauthenticated publish
with `OK` containing `auth-required: `. In practice we observed
nostr-rs-relay with `nip42_auth = true` answering `OK:true` and simply **not
storing** the event — indistinguishable from success on the client side.

For contrast, the separate P2P bridge in our stack **does** handle NIP-42
(its `publishDM` waits for the relay `AUTH` challenge and signs a kind-22242
event), which is why inbound delivery worked while bot outbound replies were
dropped — a strong sign that the relay side was fine and the client side was
the missing link.

### Repro

1. Run a nostr-rs-relay with `nip42_auth = true` (whitelist optional).
2. Point a PhantomChat bot at it (only that relay).
3. Have another party send the bot a DM, or have the bot publish to it.
4. Observe: publish returns OK; querying the relay (e.g. via `nostril`/a
   subscription from a second connection) shows the event was never stored;
   the recipient never receives it. No warning or error in the bot.

### Impact

- Silent message loss in auth-requiring relay setups — the hardest kind to
  debug, since nothing fails loudly.
- Multi-relay deployments degrade to "some relays accept, some drop".
- Bot↔bot / bridge ↔ bot traffic dies quietly on such relays.

### Suggested approach

1. **Implement NIP-42 AUTH in the relay transport.** On a relay `AUTH`
   challenge (or on an `auth-required:` OK), sign an ephemeral kind-22242
   event with the persona's PhantomChat secret key (challenge, relay URL,
   `expiration`) and submit it. `SimplePool` does not surface relay-level
   AUTH today, so this likely means a thin wrapper over a raw
   `nostr-tools` `Relay` connection (or a fork/extension of the pool) for
   the publish path.
2. **Verify acceptance (read-back).** The transport already acknowledges
   relays can silently fail to push a freshly published wrap
   (transport.ts:214). After publishing, optionally re-subscribe/read back
   the event id from that relay; log a clear warning when it is missing
   instead of failing silently.
3. **At minimum, detect and warn.** If full AUTH support is too much for a
   first pass, at least detect the `auth-required` signal and log an
   actionable error naming the relay, so operators don't mistake silent
   drops for working delivery.

### Environment

- phantombot 1.1.247 (also present on `main`)
- nostr-tools 2.23.3
- relay: nostr-rs-relay, `nip42_auth = true`
- Linux x64 (also reproduced conceptually on macOS arm64)

### Related

- Open issue #359 (relay degradation) is adjacent but distinct — this is
  specifically about missing NIP-42 support, not relay instability.

---

lenaparkhodges:
Closes #368.

## Problem

On a relay with `nip42_auth = true` (nostr-rs-relay), every bot publish was answered `OK:true` and then **silently dropped** — the turn completed cleanly, the recipient never saw the reply, nothing failed loudly. Root cause confirmed in the issue: zero NIP-42 handling anywhere in `src/`, and SimplePool never answered relay AUTH challenges.

## Fix — one PR, two layers (per Andrew's call to ship together)

**1. NIP-42 AUTH.** nostr-tools 2.23.3 already carries the machinery; it only needed a signer. New `relayAuth.ts` exports `makeRelayAuthSigner(secretKey)`, which signs the kind-22242 auth event nostr-tools builds from the relay challenge. Wired at every point nostr-tools consumes it:

- `automaticallyAuth` on the pool (run.ts + the one-shot notify pool) — answers connect-time and mid-subscription `["AUTH", challenge]` frames. Assigned post-construction: SimplePool's constructor type only exposes `enablePing`/`enableReconnect`, but `automaticallyAuth` is a public AbstractSimplePool field read at relay-connect time.
- `onauth` on **every publish and subscription** in the transport — a relay that rejects with `OK:false auth-required:` or closes a REQ demanding auth gets an authenticated retry, handled inside nostr-tools.

**2. Publish read-back verification.** AUTH alone can't catch the non-conformant relay that ACKs and drops anyway (the exact behavior observed in production). After publishing a storable event, `verifyStored` re-queries **each relay individually** for the event id (750ms settle + 2.5s per-relay timeout) and logs a warning **naming the relays that never stored it**. Detached: never blocks or rejects the send path. Ephemeral kinds (20000–29999, typing ticks) are skipped — relays don't store those by design.

## Test plan

- 8 new tests (`tests/channels-phantomchat-relayAuth.test.ts`): signer produces a verifiable kind-22242 with the persona key; transport hands `onauth` to publish + subscribe; read-back returns [] when all relays stored the event, names the dropping relay, names a dead relay on timeout, and skips ephemeral events.
- Full suite: 2614 pass — the only 2 failures (`PiHarness routing (subprocess)`) are pre-existing on clean `main`, verified by running the full suite on both branches.
- `tsc --noEmit` clean on all touched files (remaining errors are the pre-existing MCP SDK resolution noise).

## Notes for reviewers

- The read-back adds one `{ids:[...]}` REQ per relay per published storable event, detached — no latency on the send path.
- `NostrFilter.kinds` became optional (read-back filter is `{ids}`-only); the one consumer that reads `filter.kinds` (channel test fake) already guarded for it.
- Dogfooding suggestion after merge: point one persona at an auth-requiring relay and confirm the `answering relay AUTH challenge` debug line plus clean read-backs.

### DIFF
--- a/src/channels/phantomchat/relayAuth.ts
+++ b/src/channels/phantomchat/relayAuth.ts
@@ -1,0 +1,72 @@
+/**
+ * NIP-42 relay AUTH support for the phantomchat transport.
+ *
+ * THE PROBLEM (issue #368): a relay configured with `nip42_auth = true`
+ * (nostr-rs-relay) demands an authenticated session before it will store our
+ * publishes. Without NIP-42 support the relay answers our publish with
+ * `OK:true` and then silently DROPS the event — the turn completes cleanly,
+ * the recipient never sees the reply, and nothing in the logs says why. That
+ * is the worst failure class a transport can have: silent outbound message
+ * loss.
+ *
+ * THE FIX: nostr-tools 2.23.3 already carries the whole NIP-42 machinery —
+ * it just needs a signer. Two entry points consume the same signer:
+ *
+ *   1. `automaticallyAuth` on the SimplePool constructor: when a relay sends
+ *      an `["AUTH", <challenge>]` frame (on connect or mid-subscription),
+ *      nostr-tools builds the kind-22242 auth event (`makeAuthEvent(relayURL,
+ *      challenge)`) and hands it to this callback to sign.
+ *   2. The `onauth` param on `pool.publish(...)` / `pool.subscribeMany(...)`:
+ *      when a relay rejects a publish with `OK:false "auth-required: ..."` or
+ *      closes a subscription with `auth-required:`, nostr-tools authenticates
+ *      with this callback and then RETRIES the operation.
+ *
+ * Both paths converge here: the callback receives a fully-formed
+ * EventTemplate (kind 22242, `relay` + `challenge` tags already set) and only
+ * has to sign it with the persona's PhantomChat secret key.
+ *
+ * The signer is intentionally a pure function of the secret key — no relay
+ * allowlist. A relay that asks for AUTH gets it; the kind-22242 event is only
+ * ever sent to the relay that issued the challenge (nostr-tools enforces the
+ * `relay` tag match), so signing does not leak identity to third parties.
+ */
+
+import { finalizeEvent } from "nostr-tools/pure";
+import type { EventTemplate, VerifiedEvent } from "nostr-tools/core";
+
+import { log } from "../../lib/logger.ts";
+
+/**
+ * The signer shape nostr-tools expects for `onauth` / `automaticallyAuth`:
+ * given the auth EventTemplate, return the signed kind-22242 event.
+ */
+export type RelayAuthSigner = (event: EventTemplate) => Promise<VerifiedEvent>;
+
+/**
+ * Build the NIP-42 auth signer for a persona's PhantomChat secret key.
+ * Cheap and pure — safe to construct per pool and per transport.
+ */
+export function makeRelayAuthSigner(secretKey: Uint8Array): RelayAuthSigner {
+  return async (event: EventTemplate): Promise<VerifiedEvent> => {
+    const relayTag = event.tags.find((t) => t[0] === "relay")?.[1];
+    log.debug("phantomchat: answering relay AUTH challenge", {
+      kind: event.kind,
+      relay: relayTag,
+    });
+    return finalizeEvent(event, secretKey);
+  };
+}
+
+/**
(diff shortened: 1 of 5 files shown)


andrewagrahamhodges (approved):
Solid PR, well-scoped and correct. Two minor non-blocking notes:

1. `automaticallyAuth` is assigned post-construction because SimplePool constructor type doesnt expose it. Well-documented, and belt-and-suspenders with `onauth` means AUTH still works even if that field ever changes. Fine.

2. Read-back verify is best-effort (logs only); the catch-up poll stays the real delivery backstop. Correct call, no false-fail on slow relays.

No correctness or security concerns. All CI green including test-windows + typecheck. Approved.

PR #374, opened 2026-08-15
test(config): isolate PHANTOMBOT_GEMINI_API_KEY from the env snapshot

### LINKED ISSUE #373: Test isolation: config.test.ts fails locally when GEMINI_API_KEY is set in env
## Problem

`tests/config.test.ts` → "migrates a legacy Gemini CLI chain without touching embeddings" fails locally but passes in CI.

## Root Cause

The test asserts that a TOML embedding key wins over a CLI chain, but it does not clear the ambient `GEMINI_API_KEY` from the environment. On a machine that has one set (e.g. `lena@kw-phantombot`), the real key overrides the fixture, causing the assertion to fail.

## Repro

1. Have `GEMINI_API_KEY` set in your shell env
2. Run `bun test tests/config.test.ts`
3. The "migrates a legacy Gemini CLI chain without touching embeddings" test fails

## Fix

The test should isolate its environment — either:
- Delete `GEMINI_API_KEY` from `process.env` in the test setup (beforeEach), or
- Use a dedicated test fixture that explicitly sets and tears down the env var

## Context

Flagged by @robertclawson during PR #372 review. Pre-existing, unrelated to that PR. Fails identically on unmodified `origin/main`.

```bash
# Fails locally:
GEMINI_API_KEY=xxx bun test tests/config.test.ts

# Passes (no ambient key):
bun test tests/config.test.ts
```

cc @robertclawson

robertclawson:
Fixes #373.

## Root cause

`tests/config.test.ts` snapshots and clears a list of env vars (`ENV_KEYS`) in `beforeEach` so every test starts from a clean environment. `PHANTOMBOT_GEMINI_API_KEY` was missing from that list.

`buildEmbeddingsConfig()` (`src/config.ts:1232`) prefers the env key over the TOML `[embeddings.gemini] api_key`:

```ts
const envApiKey = process.env.PHANTOMBOT_GEMINI_API_KEY;
const tomlApiKey = asString(tomlGemini.api_key);
const apiKey = envApiKey ?? tomlApiKey;
```

So on a machine with a real key exported, *"migrates a legacy Gemini CLI chain without touching embeddings"* reads the ambient key instead of the `"embedding-key"` fixture and fails `expect(c.embeddings.gemini?.apiKey).toBe("embedding-key")`. CI has no key set, hence green there.

## One correction to the issue

The issue names the variable `GEMINI_API_KEY`, but that is **not** the one that breaks the test — nothing in `src/config.ts` reads it (bare `GEMINI_API_KEY` only appears in `src/lib/piModels.ts` as provider *metadata*). Verified by isolating each:

```
$ env -u PHANTOMBOT_GEMINI_API_KEY GEMINI_API_KEY=AIzaFAKE bun test tests/config.test.ts
 55 pass, 0 fail          # bare var is harmless

$ env -u GEMINI_API_KEY PHANTOMBOT_GEMINI_API_KEY=AIzaFAKE bun test tests/config.test.ts
 54 pass, 1 fail          # this is the real trigger
```

The `PHANTOMBOT_`-prefixed var is what needs isolating.

## Fix

Add `PHANTOMBOT_GEMINI_API_KEY` to `ENV_KEYS`, with a comment on why. This reuses the file's existing snapshot/clear/restore mechanism rather than adding a bespoke fixture, so it also restores the developer's real key afterwards.

## Verification

| | before | after |
|---|---|---|
| `PHANTOMBOT_GEMINI_API_KEY=AIzaFAKE bun test tests/config.test.ts` | 54 pass, **1 fail** | **55 pass, 0 fail** |
| clean env, same file | 55 pass, 0 fail | 55 pass, 0 fail |
| full suite w/ key set | 2681 pass, **1 fail** | **2682 pass, 2 skip, 0 fail** (167 files) |
| `bun tsc --noEmit` | — | clean (exit 0) |

A full-suite run with the key set confirms this was the only test in the repo affected by the leak.

### DIFF
--- a/tests/config.test.ts
+++ b/tests/config.test.ts
@@ -42,6 +42,10 @@ const ENV_KEYS = [
   "PHANTOMBOT_CODING_MODEL",
   "PHANTOMBOT_CODEX_BIN",
   "PHANTOMBOT_CODEX_MODEL",
+  // buildEmbeddingsConfig() prefers this over the TOML [embeddings.gemini]
+  // api_key, so a real key in a developer's shell silently overrides the
+  // fixture and breaks the embedding assertions below.
+  "PHANTOMBOT_GEMINI_API_KEY",
   "PHANTOMBOT_RETRIEVAL_ENABLED",
   "PHANTOMBOT_RETRIEVAL_LIMIT",
   "PHANTOMBOT_RETRIEVAL_MAX_TOKENS",


lenaparkhodges (approved):
LGTM. Root-cause analysis is solid — verified the diff isolates PHANTOMBOT_GEMINI_API_KEY via the existing snapshot/clear/restore mechanism, which also preserves the developer's real key after the run. Reusing ENV_KEYS instead of a bespoke fixture is the right call. Confirmed the correction on the var name too; and for the record, kw-phantombot does not have either GEMINI var exported, so the local failure source wasn't on my host. Merge-ready. ✅

PR #375, opened 2026-08-15
fix(reset): /reset clears the context window, no longer deletes history

robertclawson:
## The problem

`/reset` was a **destructive** command wearing the name of a harmless one. `handleReset()` did two deletes:

```ts
await ctx.memory.deleteConversation(persona, conversation);  // the turn rows
ix.deleteConversationTurns(persona, conversation);           // their embeddings
```

Andrew ran a `/reset` on his Telegram DM expecting to clear the live context and see what survived in vector memory. Nothing survived — because reset had wiped the vectors in the same call:

```
"commands: /reset", conversation:"telegram:7995070089",
 deletedTurns:4009, removedIndexedTurns:true
```

4,009 turns and every embedding, unrecoverable. Nothing else in the system holds a durable copy of recent history, so a command people reach for casually ("clear this chat") is in practice "permanently destroy this thread's memory".

## The change

`/reset` now clears **only the live context window**. Nothing is deleted.

A per-conversation watermark:

```sql
CREATE TABLE IF NOT EXISTS conversation_reset (
  persona       TEXT NOT NULL,
  conversation  TEXT NOT NULL,
  reset_turn_id INTEGER NOT NULL DEFAULT 0,
  updated_at    TEXT NOT NULL,
  PRIMARY KEY (persona, conversation)
);
```

`recentTurns()` floors the replay window at that watermark. After a reset the prompt replays nothing; the turns, their embeddings, durable facts, and the extractor's cursor/leases all survive and stay retrievable through search.

### Details worth a reviewer's eye

- **Monotonic watermark.** Upsert uses `MAX(existing, new)`, so a later reset can never un-hide turns an earlier one hid — including if a concurrent writer inserts between the read and the write.
- **Sibling conversations.** `recentTurnsForConversationPrefix()` (the cross-conversation briefing window) applies the watermark **correlated per conversation** — without it, a reset chat would leak its pre-reset turns back in through the group briefing path.
- **Durable-fact cursor deliberately NOT cleared.** `deleteConversation()` clears it because the turns are actually gone. Here they aren't, so wiping the cursor would re-extract the entire history from the top on the next pass.
- **`deleteConversation()` is untouched** and still available for genuine deletion — it's just no longer what `/reset` calls. No `--hard` flag: Andrew's call was that reset should only ever mean "clear the window".
- **Migration:** `CREATE TABLE IF NOT EXISTS` runs in `SCHEMA` on every open, so existing DBs pick it up with no backfill. Conversations with no row read as watermark `0` — everything shows, exactly as before.
- **Truthful UX.** The reply is now `reset: cleared N turns from this chat's context — history kept, still searchable`, and the Telegram command description no longer promises to "Clear this chat's history".

## Tests

Store-level (`tests/memory.test.ts`): window empties while rows survive on disk; scoping to one conversation; post-reset turns replay again; monotonicity across two resets; empty-conversation no-op; the sibling-leak case.

Command-level (`tests/channels-commands.test.ts`): the **regression this exists for** — `/reset` does not delete the underlying turns; window refills with new turns; a second reset with nothing new reports 0.

Full suite: **2691 pass, 0 fail**. Typecheck clean.

### DIFF
--- a/src/channels/commands.ts
+++ b/src/channels/commands.ts
@@ -21,11 +21,10 @@
  */
 
 import { stopNoteText } from "./core/backlog.ts";
-import { memoryIndexPath, type Config } from "../config.ts";
+import { type Config } from "../config.ts";
 import type { Harness } from "../harnesses/types.ts";
 import { formatElapsedSeconds, truncateLine } from "../lib/format.ts";
 import { log } from "../lib/logger.ts";
-import { MemoryIndex } from "../lib/memoryIndex.ts";
 import { defaultServiceControl, selfRestart } from "../lib/platform.ts";
 import type { ServiceControl } from "../lib/systemd.ts";
 import { runUpdateFlow } from "../lib/updateNotify.ts";
@@ -70,7 +69,7 @@ export interface SlashCommandContext {
   persona: string;
   /** Conversation key, e.g. "telegram:42". Used by /reset. */
   conversation: string;
-  /** Memory store for /reset's deleteConversation call. */
+  /** Memory store for /reset's context-watermark advance and /stop's note. */
   memory: MemoryStore;
   /**
    * The harness chain — mutable. /harness reorders this in place so the
@@ -168,7 +167,11 @@ export const TELEGRAM_BOT_COMMANDS: Array<{
     command: "stop",
     description: "Abort the current turn and drop anything queued behind it",
   },
-  { command: "reset", description: "Clear this chat's history" },
+  {
+    command: "reset",
+    description:
+      "Start a fresh context window here (history is kept and stays searchable)",
+  },
   { command: "status", description: "Show harness, uptime, context usage" },
   { command: "harness", description: "List or switch the active harness" },
   {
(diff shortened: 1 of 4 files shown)


lenaparkhodges (approved):
Approved — the watermark approach is exactly right. /reset drawing a line instead of destroying the record fixes the actual failure (4,009 turns + vectors gone on a routine command) with zero data movement. The details are all handled: monotonic upsert against concurrent writers, correlated watermark in the prefix query so a reset sibling can't leak back through the briefing window, durable-fact cursor deliberately untouched, and truthful UX copy. Regression coverage is where I'd want it — history survives reset, window refills, second-reset no-op, sibling-leak case. 188 tests pass locally on the branch; CI green incl. Windows.

One non-blocking flag: this PR also bundles the phantomchat group-interrupt scoping change (server.ts + its tests). It's a good fix and well-reasoned — but it's an unrelated behavior change ('stop, do Y instead' must now name the bot in multi-bot groups), and burying it in a /reset PR makes it invisible in the changelog. Not worth unbundling now; please split unrelated behavior changes into their own PRs going forward.

kaieriksen (approved):
Reviewed at exact head 312f3ac. The non-destructive per-conversation watermark preserves turns, embeddings, durable facts, and searchability while clearing replayed context; sibling briefing filtering and reset concurrency paths are covered. Focused suite 161/161 and typecheck pass; hosted Linux, Windows, and CodeQL checks are green.

lenaparkhodges (approved):
Approved — this is the right fix for a genuinely dangerous bug: a casual `/reset` destroyed 4,009 turns + embeddings with no durable copy anywhere. Watermark-over-delete is the correct product call, and the implementation is careful in the places that matter:

- Monotonic `MAX()` upsert guards against the read-then-write race.
- The correlated per-conversation watermark in `recentTurnsForConversationPrefix` closes the sibling-leak path — easy one to miss, glad it's covered by a test.
- Leaving the durable-fact cursor alone is correct and the rationale is documented in the code, not just the PR body.
- Migration is safe: `CREATE TABLE IF NOT EXISTS` + COALESCE-to-0 means untouched conversations behave exactly as before.
- Truthful UX: the reply and the Telegram command description now say what actually happens.

Nit (non-blocking): the `deleteConversation` docstring in `src/memory/store.ts` still says "Used by /reset" — it no longer is. Worth correcting so the next reader doesn't think genuine deletion is wired to a slash command.

CI green across the board (test, test-windows, typecheck, CodeQL).

PR #376, opened 2026-08-15
docs(memory): stop describing deleteConversation as the /reset path

robertclawson:
Follow-up to Lena's non-blocking nit on #375.

Since #375, `/reset` only advances the `conversation_reset` watermark — `deleteConversation()` is no longer wired to any slash command. The comments hadn't caught up:

- interface docstring said "Used by /reset" → now says it is a hard delete, explicitly **not** the /reset path
- `deleteConversation()` body comment still claimed it wipes durable **facts** — untrue since persona-scoping; only turns + extractor cursor + in-flight leases go
- the pending/lease comments (`claimEvictedTxn`, `commitPendingStmt`, schema note) attributed row wipes to `/reset`; they come from a conversation delete
- `engine.ts` abort-reason notes said "history wiped" → now "context window cleared", with the real reason the synthetic interrupted-pair is skipped (it would land above the watermark)

Comments only. No behaviour change, no test change; `tsc --noEmit` clean.

### DIFF
--- a/src/channels/core/engine.ts
+++ b/src/channels/core/engine.ts
@@ -1433,7 +1433,7 @@ async function processChatMessage(
 
   // The controller was aborted from outside. Causes:
   //   - "stop"      — /stop slash command (slash handler already replied).
-  //   - "reset"     — /reset slash command (handler replied; history wiped).
+  //   - "reset"     — /reset slash command (handler replied; context window cleared).
   //   - "interrupt" — a new message arrived for this chat; the new
   //                   turn supersedes us. Stay silent so only the new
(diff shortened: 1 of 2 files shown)


kaieriksen (approved):
Reviewed exact head 4b9246ba2fee156262469a4640b2c8e31023b55f. The comments accurately document the current split: /reset advances the conversation watermark, while deleteConversation is the explicit destructive path and leaves persona-wide durable facts intact. Hosted test, typecheck, Windows, and CodeQL checks are green; local Bun validation was unavailable because bun is not installed on this host.

PR #132, opened 2026-05-28
feat(memory): index conversation turns for retrieval

robertclawson:
## Summary
- bump the default rolling history window from 20 to 30 turns
- add a derived conversation-turn index in the existing per-persona MemoryIndex DB
- index turns after successful interactive turns when 20 new user turns have accrued, with FTS always and Gemini embeddings when configured
- include indexed older turns in turn-time auto-retrieval alongside memory/ and kb/
- clear derived turn-index rows on `/reset`

## Notes
The 30/20 pairing means a turn remains in raw rolling context long enough to be indexed before it can fall out of the window. The indexer is hot-path safe: failures are logged and swallowed so replies still succeed.

Config knobs:
- `[retrieval.turn_indexing].enabled` (default true)
- `[retrieval.turn_indexing].interval` (default 20)
- `[retrieval.turn_indexing].batch_size` (default 200)
- env overrides: `PHANTOMBOT_RETRIEVAL_TURN_INDEXING_ENABLED`, `PHANTOMBOT_RETRIEVAL_TURN_INDEXING_INTERVAL`, `PHANTOMBOT_RETRIEVAL_TURN_INDEXING_BATCH_SIZE`

## Verification
- `PATH="$HOME/.bun/bin:$PATH" bun run typecheck`
- `PATH="$HOME/.bun/bin:$PATH" bun test` (999 pass, 1 skip, 0 fail)
- `PATH="$HOME/.bun/bin:$PATH" bun test tests/memory.test.ts tests/lib-memoryIndex.test.ts tests/orchestrator-turnIndexer.test.ts tests/orchestrator-turn.test.ts tests/orchestrator-retrieval.test.ts tests/config.test.ts` (96 pass, 0 fail)
- `PATH="$HOME/.bun/bin:$PATH" bun run build`

PR #378, opened 2026-08-15
feat(retrieval): persona-scoped cross-conversation memory retrieval

### LINKED ISSUE #377: Persona-scoped cross-conversation memory retrieval
## Problem

Auto-retrieval is scoped to the current conversation (per PR #132). This means if we solved a tricky problem on Telegram, and days later work on a similar problem in PhantomChat or an ACP extension, that knowledge won't surface automatically — even though the persona has it in the vector store.

The memory is there; the system just can't find it across channels without a manual Search memory/ and kb/: hybrid BM25+vector when Gemini embeddings are set, else OKF field-weighted BM25 with link-graph expansion. (memory search)

USAGE memory search [OPTIONS] <QUERY>

ARGUMENTS

  QUERY    What to search for.    

OPTIONS

      --persona    Persona name (default: configured default).
  --scope="all"    memory | kb | all (default: all)           
    --limit="5"    max results (default 5)                    .

## Design

**Two-tier retrieval** (extends, not reverts, PR #132's scoping):

1. **Tier 1 — Current conversation:** existing threshold, unchanged. Highest priority, no cap change.
2. **Tier 2 — Cross-conversation (same persona):** higher similarity bar, hard cap (e.g. 3 turns), recency-weighted. These are awareness hits, not replay hits.

**Attribution:** Every cross-chat turn injected with source metadata — channel, date, topic — so the persona knows *where* knowledge came from. `[Telegram DM, Aug 12, re: relay auth]`

**Disclosure discipline:** Cross-chat context informs the answer but is never quoted or attributed to its source in other channels. This is a prompt-level rule that ships alongside the retrieval change — retrieval widens, system prompt gains an explicit instruction.

## Config

`crossConversationRetrieval` flag — **defaults to ON if absent.** The flag is an escape hatch for genuinely sensitive channels, not a setup step. No config file needed for normal operation.

Per-channel opt-out remains available for sensitive conversations.

## Acceptance Criteria

- [ ] Cross-conversation turns surface in auto-retrieval (tier 2) with higher similarity bar + cap
- [ ] Source attribution on every cross-chat hit (channel, date, topic)
- [ ] Prompt-level rule: cross-chat context informs answers but is never quoted/attributed in other channels
- [ ] `crossConversationRetrieval` defaults to `on` when flag is absent
- [ ] Per-channel opt-out works when explicitly set
- [ ] Regression test: current-conversation retrieval unchanged (tier 1 priority preserved)
- [ ] Regression test: embeddings persist post-reset (PR #375 behavior maintained)
- [ ] Leak test: cross-chat knowledge used in answer but source channel not disclosed in output

## Implementation Notes

- Contained change: union query in `memoryIndex.ts`, attribution formatting in injection layer, prompt rule, tests
- PR #132 scoping stays as inner tier — this extends it
- Recency decay applies to tier 2 only

lenaparkhodges:
Closes #377.

## What

Auto-retrieval was scoped to the current conversation (PR #132), so knowledge earned in one chat (a tricky fix in a Telegram DM) never surfaced when a similar problem came up days later in another (PhantomChat, ACP). The persona had the memory; retrieval couldn't reach it. This adds **tier-2 cross-conversation retrieval**, extending — not reverting — PR #132's scoping.

- **Tier 1 (unchanged):** current-conversation hits rank first, scoped exactly as before.
- **Tier 2 (new):** an unscoped turns search filtered to the persona's *other* conversations — hard cap (default 3), higher relevance bar (`max(minScore × multiplier, weakest tier-1 hit)`), same time-decay applied, appended after tier-1 hits.
- **Attribution:** every cross hit is labelled with source channel + date (`cross-conversation: Telegram, May 27`).
- **Disclosure discipline (prompt-level):** when cross hits are present, the injection header gains the rule — let them inform the reply, never quote verbatim or name the source chat.

## Config — ON by default

Absent flag = **enabled**. The flag is an escape hatch for sensitive setups, not a setup step (anti-config-fatigue principle).

```toml
[retrieval.cross_conversation]
enabled = false            # restore strict per-conversation retrieval
limit = 3                  # max cross hits per turn (0 disables)
min_score_multiplier = 1.5 # tier-2 bar = min_score × this
exclude = ["telegram"]     # channels that neither contribute nor receive
```

Env overrides: `PHANTOMBOT_RETRIEVAL_CROSS_ENABLED`, `_LIMIT`, `_MIN_SCORE_MULTIPLIER`, `_EXCLUDE` (comma-separated). Documented in README.

## Acceptance criteria (issue #377)

- [x] Tier-2 hits surface with higher bar + hard cap
- [x] Source attribution on every cross-chat hit (channel, date)
- [x] Prompt-level disclosure rule ships alongside
- [x] Defaults ON when flag absent (incl. ad-hoc settings objects)
- [x] Per-channel opt-out, both directions (contribute + receive)
- [x] Regression: tier-1 priority + strict scoping when opted out (PR #132 behaviour)
- [x] Regression: full suite incl. PR #375 reset/embeddings tests passes
- [x] Leak guard: disclosure instruction present iff cross hits are injected

## Tests

21 new/updated tests across `orchestrator-retrieval` and `config`. Full suite: 2712 pass, 2 pre-existing failures in `harnesses-pi` (reproduce on clean main — local env, unrelated). `tsc --noEmit` clean.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -213,6 +213,11 @@ export interface RetrievalSettings {
    * bumping it on recall would keep it alive — the opposite of the goal.
    */
   decay: RetrievalDecaySettings;
+  /**
+   * Tier-2 cross-conversation (persona-scoped) retrieval. Defaults ON —
+   * see CrossConversationRetrievalSettings.
+   */
+  crossConversation: CrossConversationRetrievalSettings;
 }
 
 /** Time-decay knobs for raw conversation-turn hits (curated notes are immune). */
@@ -259,6 +264,67 @@ export const DEFAULT_RETRIEVAL_DECAY: RetrievalDecaySettings = {
   floor: 0.02,
 };
 
+/**
+ * Cross-conversation (persona-scoped) retrieval — tier 2 of auto-retrieval.
+ *
+ * The problem it solves: auto-retrieval was scoped to the CURRENT
+ * conversation only (PR #132, leak fix), so knowledge earned in one chat
+ * (say, a tricky fix worked out in a Telegram DM) never surfaced when a
+ * similar problem came up days later in another chat (PhantomChat, an ACP
+ * session). The persona had the memory; retrieval couldn't reach it without
+ * a conscious `memory search` — and you can't search for what you don't
+ * remember exists.
+ *
+ * Tier 2 widens the net to the whole persona, with guardrails that keep it
+ * a supplement, never a flood: a higher relevance bar than in-conversation
+ * hits, a hard per-turn cap, source attribution on every hit, and a
+ * prompt-level disclosure rule (inform the reply, never quote the excerpt
+ * or name the chat it came from). See orchestrator/retrieval.ts.
+ *
+ * DEFAULTS TO ON when the flag is absent — deliberate anti-config-fatigue
+ * choice: the right behaviour out of the box, no knob to babysit. The flag
+ * exists purely as an escape hatch for genuinely sensitive setups.
+ */
+export interface CrossConversationRetrievalSettings {
+  /**
+   * Master switch for tier-2 (cross-conversation) hits. DEFAULT true —
+   * including when the whole [retrieval.cross_conversation] table is absent.
+   * Set false to restore strict per-conversation retrieval (PR #132
+   * behaviour).
+   */
+  enabled: boolean;
+  /**
+   * Hard cap on cross-conversation hits injected per turn, appended AFTER
+   * in-conversation hits so they can never displace tier 1. 0 disables
+   * tier 2 without touching the flag.
+   */
+  limit: number;
+  /**
+   * Tier-2 relevance bar: the configured minScore is multiplied by this for
+   * cross-conversation hits (>= 1). Independently, a cross hit must also
+   * match or beat the weakest in-conversation hit of the turn — cross-chat
+   * context only earns prompt space when it is at least as relevant as what
+   * we'd already show.
+   */
+  minScoreMultiplier: number;
+  /**
+   * Channels/conversations excluded from tier 2 in BOTH directions: their
+   * turns never surface in other chats, and no cross-conversation hits are
+   * injected when they are the current conversation. Entries match a full
+   * conversation key ("phantomchat:group:abc123") or a channel prefix
+   * ("telegram" matches every telegram:* conversation). The sensitive-DM
+   * escape hatch.
+   */
+  exclude: string[];
+}
+
+export const DEFAULT_CROSS_CONVERSATION: CrossConversationRetrievalSettings = {
+  enabled: true,
+  limit: 3,
+  minScoreMultiplier: 1.5,
+  exclude: [],
+};
+
 export const DEFAULT_RETRIEVAL: RetrievalSettings = {
   enabled: true,
   limit: 5,
@@ -267,6 +333,7 @@ export const DEFAULT_RETRIEVAL: RetrievalSettings = {
   turnIndexing: DEFAULT_TURN_INDEXING,
   graphExpansion: DEFAULT_GRAPH_EXPANSION,
   decay: DEFAULT_RETRIEVAL_DECAY,
+  crossConversation: DEFAULT_CROSS_CONVERSATION,
 };
 
 /**
@@ -949,6 +1016,49 @@ function buildRetrievalConfig(
     decay: buildRetrievalDecayConfig(
       (tomlRetrieval.decay ?? {}) as Record<string, unknown>,
     ),
+    crossConversation: buildCrossConversationConfig(
+      (tomlRetrieval.cross_conversation ?? {}) as Record<string, unknown>,
+    ),
+  };
+}
+
+function buildCrossConversationConfig(
+  toml: Record<string, unknown>,
+): CrossConversationRetrievalSettings {
+  // Absent flag → DEFAULT ON. This is the anti-config-fatigue contract:
+  // nobody should have to discover or set this knob to get the right
+  // behaviour; it exists so a sensitive setup can opt OUT.
+  const enabled =
+    asBool(process.env.PHANTOMBOT_RETRIEVAL_CROSS_ENABLED) ??
+    asBool(toml.enabled) ??
+    DEFAULT_CROSS_CONVERSATION.enabled;
+  const limit =
+    asInt(process.env.PHANTOMBOT_RETRIEVAL_CROSS_LIMIT) ??
+    asInt(toml.limit) ??
+    DEFAULT_CROSS_CONVERSATION.limit;
+  const minScoreMultiplier =
+    asNumber(process.env.PHANTOMBOT_RETRIEVAL_CROSS_MIN_SCORE_MULTIPLIER) ??
+    asNumber(toml.min_score_multiplier) ??
+    DEFAULT_CROSS_CONVERSATION.minScoreMultiplier;
+  // Env form is comma-separated ("telegram,phantomchat:group:abc"); TOML is
+  // a native string array.
+  const envExclude = process.env.PHANTOMBOT_RETRIEVAL_CROSS_EXCLUDE
+    ?.split(",")
+    .map((s) => s.trim())
+    .filter((s) => s.length > 0);
+  const exclude =
+    envExclude ??
+    asStringArray(toml.exclude) ??
+    DEFAULT_CROSS_CONVERSATION.exclude;
+  return {
+    enabled,
+    // 0 disables tier 2 without touching the flag; cap keeps the per-turn
+    // prompt cost bounded even for a fat-fingered value.
+    limit: Math.max(0, Math.min(10, limit)),
+    // Below 1 the tier-2 bar would be LOWER than tier 1's — inverted.
+    minScoreMultiplier: Math.max(1, minScoreMultiplier),
+    // Drop blanks so "a,,b" can't accidentally match every conversation.
+    exclude: exclude.map((e) => e.trim()).filter((e) => e.length > 0),
   };
 }
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -300,13 +300,23 @@ export interface CrossConversationRetrievalSettings {
    */
   limit: number;
   /**
-   * Tier-2 relevance bar: the configured minScore is multiplied by this for
-   * cross-conversation hits (>= 1). Independently, a cross hit must also
-   * match or beat the weakest in-conversation hit of the turn — cross-chat
-   * context only earns prompt space when it is at least as relevant as what
-   * we'd already show.
+   * Absolute tier-2 relevance floor on the BM25 score. Rank-fused scores
+   * (RRF) encode position within a result set, not relevance, so the bar
+   * must be absolute — and it must be non-zero by default, or an empty
+   * tier 1 would inject any cross-conversation turn that matched FTS at
+   * all (Robert's PR #378 review). Default 2.0: sampled against a live
+   * 4,353-turn index, incidental single-common-token matches score ≈ 0
+   * while genuine single-term matches score ≈ 4+.
    */
-  minScoreMultiplier: number;
+  minScore: number;
+  /**
+   * Absolute tier-2 floor on the cosine similarity, for cross hits that
+   * matched the vector side only (no lexical overlap). Default 0.85:
+   * gemini-embedding-001 vectors are highly anisotropic — sampled random
+   * turn pairs sit at p50 ≈ 0.65 / p90 ≈ 0.75 / p99 ≈ 0.89, so anything
+   * lower admits noise.
+   */
+  minVecScore: number;
   /**
    * Channels/conversations excluded from tier 2 in BOTH directions: their
    * turns never surface in other chats, and no cross-conversation hits are
@@ -321,7 +331,8 @@ export interface CrossConversationRetrievalSettings {
 export const DEFAULT_CROSS_CONVERSATION: CrossConversationRetrievalSettings = {
   enabled: true,
   limit: 3,
-  minScoreMultiplier: 1.5,
+  minScore: 2.0,
+  minVecScore: 0.85,
   exclude: [],
 };
 
@@ -1036,10 +1047,14 @@ function buildCrossConversationConfig(
     asInt(process.env.PHANTOMBOT_RETRIEVAL_CROSS_LIMIT) ??
     asInt(toml.limit) ??
     DEFAULT_CROSS_CONVERSATION.limit;
-  const minScoreMultiplier =
-    asNumber(process.env.PHANTOMBOT_RETRIEVAL_CROSS_MIN_SCORE_MULTIPLIER) ??
-    asNumber(toml.min_score_multiplier) ??
-    DEFAULT_CROSS_CONVERSATION.minScoreMultiplier;
+  const minScore =
+    asNumber(process.env.PHANTOMBOT_RETRIEVAL_CROSS_MIN_SCORE) ??
+    asNumber(toml.min_score) ??
+    DEFAULT_CROSS_CONVERSATION.minScore;
+  const minVecScore =
+    asNumber(process.env.PHANTOMBOT_RETRIEVAL_CROSS_MIN_VEC_SCORE) ??
+    asNumber(toml.min_vec_score) ??
+    DEFAULT_CROSS_CONVERSATION.minVecScore;
   // Env form is comma-separated ("telegram,phantomchat:group:abc"); TOML is
   // a native string array.
   const envExclude = process.env.PHANTOMBOT_RETRIEVAL_CROSS_EXCLUDE
@@ -1055,8 +1070,10 @@ function buildCrossConversationConfig(
     // 0 disables tier 2 without touching the flag; cap keeps the per-turn
     // prompt cost bounded even for a fat-fingered value.
     limit: Math.max(0, Math.min(10, limit)),
-    // Below 1 the tier-2 bar would be LOWER than tier 1's — inverted.
-    minScoreMultiplier: Math.max(1, minScoreMultiplier),
+    // Absolute floors; a negative floor would re-open the "inject anything
+    // that matched at all" hole, and cosine lives in [-1, 1].
+    minScore: Math.max(0, minScore),
+    minVecScore: Math.max(-1, Math.min(1, minVecScore)),
     // Drop blanks so "a,,b" can't accidentally match every conversation.
     exclude: exclude.map((e) => e.trim()).filter((e) => e.length > 0),
   };
--- a/src/config.ts
+++ b/src/config.ts
@@ -310,11 +310,17 @@ export interface CrossConversationRetrievalSettings {
    */
   minScore: number;
   /**
-   * Absolute tier-2 floor on the cosine similarity, for cross hits that
-   * matched the vector side only (no lexical overlap). Default 0.85:
-   * gemini-embedding-001 vectors are highly anisotropic — sampled random
-   * turn pairs sit at p50 ≈ 0.65 / p90 ≈ 0.75 / p99 ≈ 0.89, so anything
-   * lower admits noise.
+   * Cosine floor for the vector leg of the tier-2 gate. NOTE: cosine is
+   * never sufficient on its own — a vector hit must ALSO share at least
+   * one query term (raw BM25 > 0). The gate is applied to the MAXIMUM
+   * over the whole persona index, whose distribution is dominated by
+   * register/boilerplate similarity: on a live 4k-turn
+   * gemini-embedding-001 index, 79% of arbitrary queries had a
+   * cross-conversation turn above 0.85 on cosine alone, and the hits were
+   * shared phrasing, not knowledge (PR #378 review). No absolute value
+   * calibrated against random-pair marginals (p50 0.65 / p90 0.75 /
+   * p99 0.89) is safe as a standalone bar; 0.85 WITH lexical support
+   * keeps genuine paraphrase matches while rejecting shared phrasing.
    */
   minVecScore: number;
   /**
--- a/src/lib/memoryIndex.ts
+++ b/src/lib/memoryIndex.ts
@@ -25,10 +25,71 @@ import { readFile } from "node:fs/promises";
 import { mkdir } from "node:fs/promises";
 import { dirname, join, relative } from "node:path";
 import { posix } from "node:path";
+import type { FactSource } from "../config.ts";
 import type { Turn } from "../memory/store.ts";
 import { parseOkf } from "./okf.ts";
 
 export type Scope = "memory" | "kb" | "turns";
+
+/**
+ * Audience class of a conversation turn — who was in the room when the
+ * content was produced. Indexed with every turn so the retrieval layer can
+ * enforce the confidentiality boundary IN SQL, not in prompt text:
+ *
+ *   a memory may surface only in a room at least as wide as the audience
+ *   it was originally spoken to.
+ *
+ * private → private ✅ · multi-party → private ✅ · private → multi-party ❌
+ * ("public" is reserved for future channels with an open audience.)
+ */
+export type TurnAudience = "private" | "multi-party" | "public";
+
+/** Higher = more private. A candidate may surface in a room whose rank is
+ * >= the candidate's own (see TurnAudience). */
+export function audienceRank(a: TurnAudience): number {
+  return a === "private" ? 2 : a === "multi-party" ? 1 : 0;
+}
+
+/** Audience classes eligible to surface in a room of `room` class. */
+export function allowedAudiencesForRoom(room: TurnAudience): TurnAudience[] {
+  const r = audienceRank(room);
+  return (["private", "multi-party", "public"] as const).filter(
+    (a) => audienceRank(a) <= r,
+  );
+}
+
+/**
+ * Derive a conversation's audience class from its key shape. Group keys
+ * carry a `:group:` segment (`phantomchat:group:<id>`); Telegram group and
+ * supergroup chat ids are negative (`telegram:-100…`). Everything else —
+ * DMs, CLI, tick/system contexts — is private. Deriving from the key keeps
+ * this free at index time (no LLM, no config), and the classification is
+ * stable for the life of the conversation.
+ */
+export function conversationAudience(conversation: string): TurnAudience {
+  if (conversation.includes(":group:")) return "multi-party";
+  if (/^telegram:-/.test(conversation)) return "multi-party";
+  return "private";
+}
+
+/**
+ * SQL-side filter for conversation-turn candidate queries. Lets callers
+ * exclude the current conversation and the tier-2 exclude list BEFORE the
+ * LIMIT is applied (post-filtering a capped pool starves exactly the chats
+ * tier 2 exists for), and restrict which audience classes may surface in
+ * the current room.
+ */
+export interface TurnFilter {
+  /** Exclude this exact conversation key (typically the current one). */
+  excludeConversation?: string;
+  /**
+   * Tier-2 exclude list: an entry matches the full conversation key or a
+   * channel prefix ("telegram" → every "telegram:*" conversation).
+   */
+  exclude?: readonly string[];
+  /** Audience classes eligible to surface in the current room. */
+  allowedAudiences?: readonly TurnAudience[];
+}
 
 /**
  * Time-decay parameters for raw conversation-turn ranking. Passed down from the
@@ -121,6 +182,18 @@ export interface SearchHit {
    * index itself.
    */
   crossConversation?: string;
+  /**
+   * Provenance tier of a turn hit (who produced the content), carried from
+   * the turn_docs.source column so the retrieval layer can weigh where a
+   * memory came from. Only set for conversation-turn hits.
+   */
+  source?: FactSource;
+  /**
+   * Audience class of a turn hit (who was in the room), carried from the
+   * turn_docs.audience column so retrieval can enforce the cross-room
+   * confidentiality boundary. Only set for conversation-turn hits.
+   */
+  audience?: TurnAudience;
   snippet: string;
 }
 
@@ -188,6 +261,20 @@ export function rrfMerge(
  */
 export const NOTES_SCHEMA_VERSION = 3;
 
+/**
+ * Version of the turn_docs schema. Unlike the notes tables, turn rows are
+ * derived from the turns STORE (not from files on disk), so a schema bump
+ * drops turn_docs and clears turn_index_state — the next turn-indexer run
+ * re-walks the store and repopulates. turn_embeddings SURVIVES: the
+ * indexer reuses any embedding whose text_sha still matches, so a rebuild
+ * costs no embed API calls.
+ *
+ * v1 → v2: `source` (provenance tier) and `audience` (room class) columns
+ * added so cross-conversation retrieval can enforce the audience boundary
+ * and weigh provenance in SQL instead of prompt text (PR #378 review).
+ */
+export const TURN_SCHEMA_VERSION = 2;
+
 const SCHEMA = `
 CREATE TABLE IF NOT EXISTS meta (
   key   TEXT PRIMARY KEY,
@@ -241,6 +328,8 @@ CREATE VIRTUAL TABLE IF NOT EXISTS turn_docs USING fts5(
   conversation UNINDEXED,
   role UNINDEXED,
   turn_id UNINDEXED,
+  source UNINDEXED,
+  audience UNINDEXED,
   content,
   tokenize = 'porter unicode61'
 );
@@ -271,6 +360,7 @@ export class MemoryIndex {
     // process touches the index DB concurrently.
     db.exec(SCHEMA);
     this.selfHealNotesSchema();
+    this.selfHealTurnSchema();
   }
 
   /**
@@ -313,6 +403,48 @@ export class MemoryIndex {
           "ON CONFLICT(key) DO UPDATE SET value = excluded.value",
       )
       .run(String(NOTES_SCHEMA_VERSION));
+  }
+
+  /**
+   * If the persisted turn-schema version is older than TURN_SCHEMA_VERSION
+   * (or absent while turn_docs holds rows, i.e. a pre-versioning index),
+   * drop turn_docs and clear turn_index_state: the next turn-indexer run
+   * re-walks the turns store and repopulates with the new columns.
+   * turn_embeddings is deliberately NOT dropped — the indexer reuses any
+   * embedding whose text_sha still matches, so the rebuild costs no embed
+   * API calls. Cheap and idempotent.
+   */
+  private selfHealTurnSchema(): void {
+    const row = this.db
+      .query("SELECT value FROM meta WHERE key = 'turn_schema_version'")
+      .get() as { value?: string } | null;
+    const have = row?.value ? Number(row.value) : 0;
+    if (have >= TURN_SCHEMA_VERSION) return;
+    // Meta rows are only written by versions that already self-heal, so a
+    // missing/low stamp alone isn't proof of an old layout — an old EMPTY
+    // index has no meta row either. The authoritative check is the column
+    // set itself: if `source` is present the table is already v2-shaped
+    // (fresh DBs land here) and only needs the stamp.
+    const cols = this.db
+      .query("PRAGMA table_info(turn_docs)")
+      .all() as Array<{ name: string }>;
+    if (cols.some((c) => c.name === "source")) {
+      this.stampTurnSchemaVersion();
+      return;
+    }
+    this.db.exec("DROP TABLE IF EXISTS turn_docs;");
+    this.db.exec(SCHEMA);
+    this.db.exec("DELETE FROM turn_index_state;");
+    this.stampTurnSchemaVersion();
+  }
+
+  private stampTurnSchemaVersion(): void {
+    this.db
+      .prepare(
+        "INSERT INTO meta (key, value) VALUES ('turn_schema_version', ?) " +
+          "ON CONFLICT(key) DO UPDATE SET value = excluded.value",
+      )
+      .run(String(TURN_SCHEMA_VERSION));
   }
 
   static async open(indexPath: string): Promise<MemoryIndex> {
@@ -512,11 +644,17 @@ export class MemoryIndex {
 
   upsertTurn(turn: Turn, vec?: Float32Array, textSha?: string): void {
     const path = turnPath(turn);
-    this.deleteTurnPath(path);
+    // Delete ONLY the turn_docs row. Turn paths are id-keyed and turn text
+    // is immutable, so a surviving turn_embeddings row for this path always
+    // matches the content being (re-)inserted — keeping it is what makes a
+    // turn-schema rebuild cost zero embed API calls. A stale embedding can
+    // only be orphaned if the embed call fails on a genuinely changed text,
+    // which the id-keyed path makes impossible.
+    this.db.prepare("DELETE FROM turn_docs WHERE path = ?").run(path);
     this.db
       .prepare(
-        "INSERT INTO turn_docs (path, persona, conversation, role, turn_id, content) " +
-          "VALUES (?, ?, ?, ?, ?, ?)",
+        "INSERT INTO turn_docs (path, persona, conversation, role, turn_id, source, audience, content) " +
+          "VALUES (?, ?, ?, ?, ?, ?, ?, ?)",
       )
       .run(
         path,
@@ -524,6 +662,8 @@ export class MemoryIndex {
         turn.conversation,
         turn.role,
         turn.id,
+        turn.source,
+        conversationAudience(turn.conversation),
         renderTurnForIndex(turn),
       );
 
@@ -728,6 +868,43 @@ export class MemoryIndex {
   // Search
   // -------------------------------------------------------------
 
+  /**
+   * Compile a TurnFilter into SQL AND-clauses (with bound params) for the
+   * turn_docs queries. Filtering in SQL — before ORDER BY … LIMIT — means
+   * every slot in the candidate pool is an ELIGIBLE hit; post-filtering a
+   * capped pool in JS starves tier 2 exactly when the current chat matches
+   * its own query best (Kai's PR #378 review).
+   */
+  private turnFilterSql(filter: TurnFilter | undefined): {
+    where: string;
+    params: string[];
+  } {
+    if (!filter) return { where: "", params: [] };
+    const clauses: string[] = [];
+    const params: string[] = [];
+    if (filter.excludeConversation !== undefined) {
+      clauses.push("conversation != ?");
+      params.push(filter.excludeConversation);
+    }
+    for (const e of filter.exclude ?? []) {
+      // Full-key match OR channel-prefix match, mirroring
+      // isConversationExcluded. substr() instead of LIKE so `_`/`%` in a
+      // key can never act as wildcards.
+      clauses.push("conversation != ? AND substr(conversation, 1, ?) != ?");
+      params.push(e, String(e.length + 1), `${e}:`);
+    }
+    if (filter.allowedAudiences && filter.allowedAudiences.length > 0) {
+      clauses.push(
+        `audience IN (${filter.allowedAudiences.map(() => "?").join(", ")})`,
+      );
+      params.push(...filter.allowedAudiences);
+    }
+    return {
+      where: clauses.length > 0 ? ` AND ${clauses.join(" AND ")}` : "",
+      params,
+    };
+  }
+
   search(
     query: string,
     opts: {
@@ -748,6 +925,12 @@ export class MemoryIndex {
        * ghost. Omitted → relevance-only ranking (CLI `memory search`).
        */
       decay?: TurnDecay;
+      /**
+       * SQL-side candidate filter for turn rows (see TurnFilter). Applied
+       * before LIMIT so the pool contains only eligible hits. Combinable
+       * with `conversation` (both clauses apply).
+       */
+      turnFilter?: TurnFilter;
     } = {},
   ): SearchHit[] {
     const limit = Math.max(1, Math.min(opts.limit ?? 5, 50));
@@ -794,30 +977,37 @@ export class MemoryIndex {
             snip: string;
           }>);
 
+    const tf = this.turnFilterSql(opts.turnFilter);
     const turnRows =
       scope === "all" || scope === "turns"
         ? opts.conversation !== undefined
           ? (this.db
               .query(
-                "SELECT path, bm25(turn_docs) AS rank, " +
-                  "snippet(turn_docs, 5, '«', '»', ' … ', 16) AS snip " +
+                "SELECT path, source, audience, bm25(turn_docs) AS rank, " +
+                  "snippet(turn_docs, 7, '«', '»', ' … ', 16) AS snip " +
                   "FROM turn_docs WHERE content MATCH ? AND conversation = ? " +
-                  "ORDER BY rank LIMIT ?",
+                  tf.where +
+                  " ORDER BY rank LIMIT ?",
               )
-              .all(ftsQuery, opts.conversation, turnLimit) as Array<{
+              .all(ftsQuery, opts.conversation, ...tf.params, turnLimit) as Array<{
               path: string;
+              source: FactSource;
+              audience: TurnAudience;
               rank: number;
               snip: string;
             }>)
           : (this.db
               .query(
-                "SELECT path, bm25(turn_docs) AS rank, " +
-                  "snippet(turn_docs, 5, '«', '»', ' … ', 16) AS snip " +
+                "SELECT path, source, audience, bm25(turn_docs) AS rank, " +
+                  "snippet(turn_docs, 7, '«', '»', ' … ', 16) AS snip " +
                   "FROM turn_docs WHERE content MATCH ? " +
-                  "ORDER BY rank LIMIT ?",
+                  tf.where +
+                  " ORDER BY rank LIMIT ?",
               )
-              .all(ftsQuery, turnLimit) as Array<{
+              .all(ftsQuery, ...tf.params, turnLimit) as Array<{
               path: string;
+              source: FactSource;
+              audience: TurnAudience;
               rank: number;
               snip: string;
             }>)
@@ -847,6 +1037,8 @@ export class MemoryIndex {
         // Decay damps stale turns (factor in [floor,1]); default 1 when decay
         // is off or the turn's age couldn't be parsed.
         ftsScore: -r.rank * (decayFactors?.get(r.path) ?? 1),
+        source: r.source ?? undefined,
+        audience: r.audience ?? undefined,
         snippet: r.snip,
       })),
     ]
@@ -880,6 +1072,8 @@ export class MemoryIndex {
       maxAdd?: number;
       /** Time-decay for turn hits (see TurnDecay); forwarded to search(). */
       decay?: TurnDecay;
+      /** SQL-side turn candidate filter; forwarded to search(). */
+      turnFilter?: TurnFilter;
     } = {},
   ): SearchHit[] {
     const limit = Math.max(1, Math.min(opts.limit ?? 5, 50));
@@ -888,6 +1082,7 @@ export class MemoryIndex {
       limit,
       conversation: opts.conversation,
       decay: opts.decay,
+      turnFilter: opts.turnFilter,
     });
     const maxAdd = Math.max(0, opts.maxAdd ?? 3);
     if (maxAdd === 0) return hits;
@@ -1008,6 +1203,12 @@ export class MemoryIndex {
        * and decay isn't double-counted.
        */
       decay?: TurnDecay;
+      /**
+       * SQL-side turn candidate filter (see TurnFilter). Applied to BOTH the
+       * inner FTS search and the vector allowed-paths set, so ineligible
+       * turns never occupy a candidate slot on either ranking.
+       */
+      turnFilter?: TurnFilter;
     } = {},
   ): SearchHit[] {
     const limit = Math.max(1, Math.min(opts.limit ?? 5, 50));
@@ -1015,6 +1216,7 @@ export class MemoryIndex {
       scope: opts.scope,
       limit: 25,
       conversation: opts.conversation,
+      turnFilter: opts.turnFilter,
     });
 
     // No vector → FTS-only fallback. Still honour decay here (the inner search
@@ -1053,7 +1255,11 @@ export class MemoryIndex {
     // mirrors the FTS scoping in search(). (Kai's review on PR #132.)
     const searchScope = opts.scope ?? "all";
     const conversation = opts.conversation;
-    if (searchScope !== "all" || conversation !== undefined) {
+    if (
+      searchScope !== "all" ||
+      conversation !== undefined ||
+      opts.turnFilter !== undefined
+    ) {
       const allowedPaths = new Set<string>();
       // Notes (memory/kb): allowed by scope, never filtered by conversation.
       if (searchScope === "all") {
@@ -1067,16 +1273,22 @@ export class MemoryIndex {
           .all(searchScope) as Array<{ path: string }>)
           allowedPaths.add(r.path);
       }
-      // Conversation turns: scoped to the current conversation when provided.
+      // Conversation turns: scoped to the current conversation when provided,
+      // and always through the SQL turn filter (audience/exclusions).
       if (searchScope === "all" || searchScope === "turns") {
+        const tf = this.turnFilterSql(opts.turnFilter);
         const turnRows =
           conversation !== undefined
             ? (this.db
-                .query("SELECT path FROM turn_docs WHERE conversation = ?")
-                .all(conversation) as Array<{ path: string }>)
+                .query(
+                  "SELECT path FROM turn_docs WHERE conversation = ?" + tf.where,
+                )
+                .all(conversation, ...tf.params) as Array<{ path: string }>)
             : (this.db
-                .query("SELECT path FROM turn_docs")
-                .all() as Array<{ path: string }>);
+                .query(
+                  "SELECT path FROM turn_docs WHERE 1 = 1" + tf.where,
+                )
+                .all(...tf.params) as Array<{ path: string }>);
         for (const r of turnRows) allowedPaths.add(r.path);
       }
       for (const path of [...vecScores.keys()]) {
@@ -1116,16 +1328,37 @@ export class MemoryIndex {
           ftsHit?.scope ??
           this.lookupScope(path) ??
           (path.startsWith("turns/") ? "turns" : ("kb" as Scope)); // best guess
+        const turnMeta =
+          ftsHit?.source !== undefined
+            ? { source: ftsHit.source, audience: ftsHit.audience }
+            : this.turnMetaForPath(path);
         return {
           path,
           scope,
           ftsScore: ftsHit?.ftsScore,
           vecScore: vecScores.get(path),
           rrfScore,
+          ...turnMeta,
           snippet: ftsHit?.snippet ?? this.snippetForPath(path),
         };
       });
     return merged;
+  }
+
+  /** Provenance/audience for a turn path (undefined for non-turn paths). */
+  private turnMetaForPath(
+    path: string,
+  ): { source?: FactSource; audience?: TurnAudience } {
+    if (!path.startsWith("turns/")) return {};
+    const row = this.db
+      .query("SELECT source, audience FROM turn_docs WHERE path = ? LIMIT 1")
+      .get(path) as
+      | { source?: FactSource | null; audience?: TurnAudience | null }
+      | null;
+    return {
+      source: row?.source ?? undefined,
+      audience: row?.audience ?? undefined,
+    };
   }
 
   /**
--- a/src/lib/memoryIndex.ts
+++ b/src/lib/memoryIndex.ts
@@ -1370,12 +1370,11 @@ export class MemoryIndex {
       for (const r of rows2) {
         const hit = merged.find((h) => h.path === r.path);
         if (hit) {
-          // bm25() is "lower is better"; flip sign. Apply decay if active
-          // to match the ftsScore convention from search().
-          const raw = -r.rank;
-          hit.ftsScore = decayFactors
-            ? raw * (decayFactors.get(r.path) ?? 1)
-            : raw;
+          // bm25() is "lower is better"; flip sign. Keep raw — pool hits
+          // from search() carry undecayed ftsScore, and the tier-2 floor
+          // (minScore: 2.0) is calibrated on raw BM25. Decaying here would
+          // inflate the threshold for aged content, narrowing tier 2.
+          hit.ftsScore = -r.rank;
         }
       }
     }
--- a/src/lib/memoryIndex.ts
+++ b/src/lib/memoryIndex.ts
@@ -1342,6 +1342,44 @@ export class MemoryIndex {
           snippet: ftsHit?.snippet ?? this.snippetForPath(path),
         };
       });
+    // Per-path BM25 lookup: vector-only turn hits (those not in the FTS
+    // top-25 pool) have ftsScore === undefined. Fetch their actual BM25
+    // score directly so the tier-2 dual-gate (cosine ≥ minVecScore AND
+    // fts > 0) works as documented — without this, a vector hit ranked
+    // 26th+ lexically is silently dropped by pool truncation, not by the
+    // score rule, quietly degrading tier 2 to cosine-only on busy indexes.
+    // (Robbie's non-blocking finding on #378, resolved.)
+    const vecOnlyTurnPaths = merged
+      .filter((h) => h.ftsScore === undefined && h.path.startsWith("turns/"))
+      .map((h) => h.path);
+    if (vecOnlyTurnPaths.length > 0) {
+      const ftsQuery2 = sanitizeFtsQuery(query);
+      const ph = vecOnlyTurnPaths.map(() => "?").join(", ");
+      const tf2 = this.turnFilterSql(opts.turnFilter);
+      const rows2 = this.db
+        .query(
+          "SELECT path, bm25(turn_docs) AS rank FROM turn_docs " +
+            "WHERE content MATCH ? AND path IN (" + ph + ") " +
+            tf2.where +
+            " ORDER BY rank",
+        )
+        .all(ftsQuery2, ...vecOnlyTurnPaths, ...tf2.params) as Array<{
+        path: string;
+        rank: number;
+      }>;
+      for (const r of rows2) {
+        const hit = merged.find((h) => h.path === r.path);
+        if (hit) {
+          // bm25() is "lower is better"; flip sign. Apply decay if active
+          // to match the ftsScore convention from search().
+          const raw = -r.rank;
+          hit.ftsScore = decayFactors
+            ? raw * (decayFactors.get(r.path) ?? 1)
+            : raw;
+        }
+      }
+    }
+
     return merged;
   }
 
--- a/src/lib/memoryIndex.ts
+++ b/src/lib/memoryIndex.ts
@@ -113,6 +113,14 @@ export interface SearchHit {
    * callers label or down-weight expanded neighbours.
    */
   expanded?: boolean;
+  /**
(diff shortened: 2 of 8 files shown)


src/orchestrator/retrieval.ts
```diff
@@ -136,7 +146,49 @@ export async function retrieveContext(
...
+      );
+      // The tier-2 bar: max(minScore × multiplier, weakest tier-1 hit).
+      // With no in-conversation hits the bar collapses to minScore ×
+      // multiplier (0 by default) — exactly the "solved it on Telegram last
+      // week, asking from PhantomChat today" case tier 2 exists for.
+      const tier1Weakest = hits.length ? Math.min(...hits.map(scoreOf)) : 0;
+      const bar = Math.max(opts.settings.minScore * multiplier, tier1Weakest);
+      // Fetch a wider pool than the cap: current-conversation and excluded
+      // candidates are filtered out below.
+      const pool = Math.min(50, crossLimit * 4);
```
kaieriksen: The fixed crossLimit * 4 pool is not sufficient once filtering is applied. ix.search/hybridSearch returns only the top 12 turn hits for the default limit 3, but those can all belong to the current conversation or excluded conversations; a valid cross-conversation hit ranked 13th is then never considered. This is especially likely when one chat has many matching turns. Please make the query filter by conversation/exclusions before limiting, or expand/retry the pool until crossLimit eligible hits (or exhaustion) are found, and add a regression test with more than 12 ineligible hits ahead of one eligible hit.

kaieriksen (changes requested):
Changes requested: the cross-tier candidate pool is capped before current/excluded conversations are filtered, so valid cross-conversation hits can be silently omitted. Focused retrieval/config tests pass 96/96, typecheck passes, and hosted Linux/Windows/CodeQL checks are green; please fix the pool starvation and add the regression coverage.

src/orchestrator/retrieval.ts
```diff
@@ -136,7 +146,49 @@ export async function retrieveContext(
...
+        1,
+        cross?.minScoreMultiplier ??
+          DEFAULT_CROSS_CONVERSATION.minScoreMultiplier,
+      );
+      // The tier-2 bar: max(minScore × multiplier, weakest tier-1 hit).
+      // With no in-conversation hits the bar collapses to minScore ×
+      // multiplier (0 by default) — exactly the "solved it on Telegram last
+      // week, asking from PhantomChat today" case tier 2 exists for.
+      const tier1Weakest = hits.length ? Math.min(...hits.map(scoreOf)) : 0;
+      const bar = Math.max(opts.settings.minScore * multiplier, tier1Weakest);
```
robertclawson: **Blocking — the bar is `0` under default config.**

`DEFAULT_RETRIEVAL.minScore` is `0` (src/config.ts). So the first term is `0 × 1.5 = 0`, and when tier 1 returns nothing `tier1Weakest` is also `0`:

```
bar = max(0 * 1.5, 0) = 0
```

`selectCrossConversationHits` then rejects only `scoreOf(h) < 0`, which is never true, and `formatRetrieved` filters on `>= minScore` i.e. `>= 0`. Net effect: **the top 3 cross-conversation candidates are injected unconditionally**, however weak, whenever the current chat has no hits of its own.

The comment above frames the empty-tier-1 collapse as intentional ("exactly the case tier 2 exists for"), but combined with `minScore = 0` it means there is no floor at all — not a lower one. A single incidental FTS token match from an unrelated chat earns prompt space.

Suggest an **absolute** tier-2 floor that doesn't derive from `minScore`, e.g. `crossConversation.minScore` defaulting to a real non-zero value, and keep `max(absoluteFloor, tier1Weakest)`. That keeps the knob-free promise while making the guardrail exist.

src/orchestrator/retrieval.ts
```diff
@@ -136,7 +146,49 @@ export async function retrieveContext(
...
+      const multiplier = Math.max(
+        1,
+        cross?.minScoreMultiplier ??
+          DEFAULT_CROSS_CONVERSATION.minScoreMultiplier,
+      );
+      // The tier-2 bar: max(minScore × multiplier, weakest tier-1 hit).
+      // With no in-conversation hits the bar collapses to minScore ×
+      // multiplier (0 by default) — exactly the "solved it on Telegram last
+      // week, asking from PhantomChat today" case tier 2 exists for.
+      const tier1Weakest = hits.length ? Math.min(...hits.map(scoreOf)) : 0;
```
robertclawson: **Blocking (same fix as above) — these two scores come from different rankings and aren't comparable.**

`scoreOf` returns `rrfScore ?? ftsScore`, and RRF is purely positional: `rrfMerge` assigns `1 / (k + idx + 1)` with `k = 60`. The score therefore encodes *rank within its own result set*, not absolute relevance.

`hits` is ranked inside the **scoped** search; the tier-2 `candidates` are ranked inside a **separate unscoped** search. The rank-1 cross candidate scores ~`0.0164` per list simply for being first in its own set — the same value the rank-1 tier-1 hit gets. So "a cross hit must match or beat the weakest tier-1 hit" reduces to "the cross hit ranked higher in its own search than the tier-1 tail did in its own search", which is close to always true.

A rank-based score can't serve as a cross-search relevance bar. Either compare on a metric that is absolute (raw cosine similarity, or normalised BM25), or do one unscoped search and partition the single ranking into tier 1 / tier 2 — the latter also fixes the pool problem below and makes the comparison meaningful by construction.

src/orchestrator/retrieval.ts
```diff
@@ -183,17 +235,31 @@ export function formatRetrieved(
   );
   if (usable.length === 0) return undefined;
 
+  // Disclosure discipline: cross-conversation excerpts may INFORM the reply
+  // but are never quoted or attributed to their source chat in it. Ships
+  // with the retrieval widening (issue #377) — the rule rides the header so
+  // it appears iff tier-2 hits are actually present.
+  const hasCross = usable.some((h) => h.crossConversation !== undefined);
```
robertclawson, line 283: **Blocking — this is an audience boundary, and it can't live in the header.**

The disclosure rule is well written, but it's an instruction to the model, and a confidentiality perimeter enforced by prompt text is not a perimeter. More importantly the rule guards the *wrong property*: it prevents **naming the source chat**, when the actual risk is **content reaching a different audience**.

Concretely, with `exclude` defaulting to `[]`: a turn from Andrew's 1:1 Telegram DM is eligible to surface while the persona is replying in a multi-party PhantomChat group. Paraphrasing it without naming the chat is still disclosure to people who were never in that conversation. Provenance is maximal here — the leak isn't about trusting the source, it's about the destination.

The fix is a retrieval-side filter on an **audience class** (`1:1-principal` / `multi-party` / `public`), with the rule:

> a memory may surface only when its audience class is at least as private as the room being spoken into.

Private→private ✅, group→private ✅, private→group ❌. Store it at index time (see my note on `SearchHit.crossConversation`); as a stopgap it is already derivable from conversation-key shape (`…:group:…`), and `engine.ts` computes `isGroupChat` at ingress today.

Keep this header sentence as belt-and-braces — just don't let it be the only thing standing between a DM and a group. Note this also needs no config: it's a fixed rule, not a knob, so it stays inside the no-config contract.

src/lib/memoryIndex.ts
```diff
@@ -113,6 +113,14 @@ export interface SearchHit {
...
    */
   expanded?: boolean;
+  /**
+   * Set by the auto-retrieval layer (orchestrator/retrieval.ts) when this
+   * hit was surfaced from a DIFFERENT conversation than the current one
+   * (tier-2 cross-conversation retrieval). Carries the decoded source
+   * conversation key so the formatter can attribute it. Never set by the
+   * index itself.
+   */
+  crossConversation?: string;
```
robertclawson, line 184: **Blocking — provenance needs to be indexed here, not inferred later.**

`crossConversation` carries the conversation key, which gives channel and (via the snippet) date. It does not tell the retrieval layer **who produced the content**, and `turn_docs` has no column for it:

```sql
CREATE VIRTUAL TABLE turn_docs USING fts5(
  path, persona, conversation, role, turn_id, content
);
```

Per Andrew's ruling on #377, tier 2 should surface judge-passed inbound too — the threat judge already made that call, and re-litigating it at retrieval is the double-gating we're trying to avoid. But that only works if the hit is **tagged** so it can be weighed. Suggested vocabulary:

- `principal` — Andrew, direct
- `channel` / `acp` — inbound, human-authored
- `inbound-judged` — email / Twilio / webhook that cleared the judge
- `task` — scheduled-task output
- `internal` — heartbeat, nightly, self-generated notify payloads

`task` and `internal` matter most and are the least obvious: a scheduled task is the persona talking to itself. Its conclusion was never checked by Andrew, but once retrieved days later it reads as established fact and is silently promoted to the same standing as something he actually said.

One trap worth designing against **now**: a task that reads an untrusted email and writes a summary would launder `inbound-judged` content into a trusted `task` row in one hop. Rule should be that **a derived turn inherits the weakest provenance of its inputs** (`task` + `inbound-judged`, not `task`). Cheap while the schema is being touched; impossible to retrofit once the index is full.

So: two UNINDEXED columns at index time, `source` and `audience`. Both are stored facts, neither costs an LLM call, and neither needs a config flag.

src/orchestrator/retrieval.ts
robertclawson: **Non-blocking, but it will make tier 2 look broken in the chats that need it most.**

`pool = min(50, crossLimit * 4)` = **12** at defaults. The unscoped search returns the global top 12, and `selectCrossConversationHits` then discards every candidate belonging to the current conversation. In an active chat the current conversation is usually the strongest match for its own query — so if it occupies all 12 slots, tier 2 yields **zero** hits.

The failure scales the wrong way: the busier the current chat, the more reliably tier 2 starves, and it fails silently (no hits is indistinguishable from nothing relevant existing).

Partly masked when decay is on, since `turnLimit` is floored at `DECAY_CANDIDATE_POOL = 25` — which is worth knowing, because it means the effective pool silently depends on an unrelated setting.

Cleanest fix is to exclude the current conversation **in SQL** (`AND conversation != ?`) so all 12 slots are usable candidates, rather than over-fetching and filtering in JS.

tests/orchestrator-retrieval.test.ts
```diff
@@ -88,6 +98,32 @@ describe("formatRetrieved", () => {
     expect(out).toBeUndefined();
   });
 
+  test("cross-conversation hits get attribution and the disclosure rule", () => {
```
robertclawson, line 101: **Non-blocking — this guards the invariant we care least about.**

The assertion is that the header *contains the disclosure sentence*. That tests string plumbing, not containment: it passes whether or not private content actually reached the wrong audience, and it would keep passing if the model ignored the instruction entirely — which is the whole reason a prompt rule is the wrong mechanism.

Once the audience filter exists, the test worth having is the negative one:

> given a `1:1-principal` turn in the index and a `multi-party` current conversation, `retrieveContext` returns **no** cross hit from that turn.

That one fails loudly if the boundary regresses. Happy to see the header assertion stay alongside it as a formatting check.

robertclawson (changes requested):
Reviewed at `5423d0f`. The **direction is right** — this extends #132 with an inner tier instead of reverting it, and the default-ON / opt-out-only config shape is exactly the anti-config-fatigue contract Andrew asked for. Attribution, the hard cap, tier-1-first ordering and the pure-helper split are all clean, and the test coverage is genuine.

Three things I think block merge:

1. **The "higher relevance bar" does not exist at default config.** `minScore` defaults to `0`, so `minScore × multiplier = 0`. When tier 1 returns no hits the bar is `max(0, 0) = 0` and the top 3 cross-conversation candidates are injected *unconditionally* — any turn that matched FTS at all. That is precisely the case tier 2 is advertised for, so the guardrail is absent exactly when it matters.

2. **Confidentiality is enforced by a sentence in the prompt.** `exclude` defaults to `[]`, so out of the box a private Telegram DM turn can surface inside a multi-party PhantomChat group, with nothing but a header instruction between it and the other participants. An audience boundary has to be a retrieval filter, not a prompt rule.

3. **No provenance on retrieved turns.** `turn_docs` is `path/persona/conversation/role/turn_id/content` — there is no `source` column, so tier 2 cannot weigh where a memory came from. Per Andrew's call on the issue thread this should land *in this PR*, not as a follow-up, because it cannot be retrofitted once the index is full of untagged rows.

Two non-blocking notes inline (candidate-pool starvation, and the leak test asserting the wrong invariant).

To be explicit about what is **not** a problem: the index is a per-persona SQLite file (`memory-index/<persona>.sqlite`), so the unscoped tier-2 query cannot leak across personas. And dropping "topic" from the attribution was the right call — it is stored nowhere and synthesising it would mean an LLM call on a hot path contractually required to be cheap and never throw.

lenaparkhodges:
Thanks both — all three blocking items addressed in `d92f36b`. Summary per point:

**1. Bar was 0 at default config (Robert, blocking).** The derived bar is gone. Tier 2 now requires an **absolute** floor: `crossConversation.minScore` on BM25 (default **2.0**) or `minVecScore` on raw cosine (default **0.85**), either sub-score clears it. Defaults are calibrated, not guessed — sampled my live 4,353-turn index: incidental single-common-token matches score ≈ 0, genuine single-term matches ≈ 4+; random gemini-embedding-001 turn pairs sit at p50 ≈ 0.65 / p90 ≈ 0.75 / p99 ≈ 0.89 (the vectors are highly anisotropic, hence 0.85).

**2. Rank-based scores aren't comparable across searches (Robert, blocking).** Fixed by construction: RRF is never used as a bar anywhere — the floor applies only to the absolute sub-scores (BM25 / cosine). The `tier1Weakest` term is deleted rather than repaired.

**3. Confidentiality by prompt sentence (Robert, blocking).** The audience boundary is now a **retrieval filter in SQL**. `turn_docs` gains `source` and `audience` columns (both UNINDEXED, both free at index time). Audience is derived from conversation-key shape (`:group:` segment, or negative Telegram chat id → `multi-party`; else `private`) and the rule is exactly your matrix: a memory surfaces only in a room at least as wide as its original audience — private→group is rejected before ranking. The disclosure header stays as belt-and-braces. On provenance vocabulary: I used the existing `FactSource` tiers (`principal`/`self`/`other`/`unverified`) rather than a parallel 5-value scheme — they're already maintained at write time, and the laundering trap you flagged is covered by the #327 default (assistant/task turns land `unverified` until the principal engages). Happy to revisit if you want the finer `task`/`internal` split.

**4. Pool starvation (Kai, blocking; Robert, non-blocking).** All exclusions — current conversation, exclude list (full-key and channel-prefix), audience — are applied **in SQL before LIMIT**, on both the FTS and the vector allowed-paths sides. Every pool slot is now an eligible hit by construction. Regression test included: 15 in-conversation turns ranking ahead of 1 cross turn, cross hit still surfaces.

**5. Provenance indexed, not inferred (Robert, blocking).** Done as above — carried on every turn `SearchHit`. Schema is `turn_docs` v2 with self-heal on open: drops `turn_docs`, clears `turn_index_state`, **keeps** `turn_embeddings`; the indexer now reuses embeddings by `text_sha`, so the rebuild costs zero embed API calls.

**6. Leak test asserting the wrong invariant (Robert, non-blocking).** Added the negative test you specced: private-DM turn indexed, multi-party current conversation → no cross hit and content absent — plus the positive control (same turn surfaces in a private room, proving it's the boundary and not the floor), and group→private allowed.

Test data note: tier-2 integration tests now seed 25 filler turns and use recent dates — BM25 idf is corpus-relative (a 2-doc index can't clear any floor) and the default 30-day decay damps months-old fixtures below the floor.

2,723 tests pass; the 2 `harnesses-pi` failures reproduce on `main` (ambient `PHANTOMBOT_PI_API_KEY` on this host — same isolation class as #373, unrelated).

robertclawson (changes requested):
Re-reviewed `d92f36b`. Points **2, 3, 4, 5 and 6 are properly fixed** — the audience boundary as an SQL filter is the right shape (`turnFilterSql` before `LIMIT`, applied to both the FTS and the vector allowed-paths side), the RRF-as-a-bar term is gone, provenance rides the hit, the pool-starvation regression test does what it says, and reusing embeddings by `text_sha` makes the v2 rebuild genuinely free. Reusing the existing `FactSource` tiers instead of a parallel scheme is the right call — I checked `turn.ts:389`, assistant turns land `unverified` by default, so the laundering path is covered. No need for the finer `task`/`internal` split now.

**Point 1 is only half fixed, and the other half re-opens the hole.** Requesting changes on that one thing.

---

### 1. `minVecScore: 0.85` is calibrated against the wrong distribution (blocking)

The floor is derived from the **marginal** distribution of random turn pairs (p50 0.65 / p90 0.75 / p99 0.89), but it is applied to the **maximum over the whole index** — tier 2 takes the best 3 of thousands of candidates. Those are different distributions, and the gap is not small.

I ran it against my own live index (robbie, 4,223 turns, same gemini-embedding-001, cosine on L2-normalised vectors). 200 randomly drawn turns used as queries, scored against every turn outside their own conversation:

```
random-pair cosine (marginal):     p50 0.657   p90 0.763      <- matches your sample
TOP-1 cross-conv cosine per query: p10 0.796   p50 0.899   p90 0.993
TOP-3 cross-conv cosine per query: p10 0.782   p50 0.883

best cross hit clears 0.85:  79% of queries
3rd-best clears 0.85:        70% of queries
```

So for ~4 arbitrary messages in 5, at least one cross-conversation turn clears the floor on cosine alone, and for 70% all three slots fill — with no relevance requirement whatsoever. That is the original "bar is 0" failure mode, just reached through the vector leg of the `OR`.

And the hits that clear it are not knowledge, they're **register**. Actual top-1 pairs from that run:

```
0.963  "STATE REPORT: wrote memory/.nightly-state.json — status ok, 0 promoted, 0 KB notes…"
   >>  "STATE REPORT stage complete: wrote memory/.nightly-state.json — status ok, 0 promoted…"

0.853  "Here's my plan before I touch anything: 1. Ground first — clone/pull phantomyard/phantombot…"
   >>  "Good — this is a sizable, multi-file build (git + bun build + tests), so let me lock scope…"
```

Two different nightly runs at 0.963. Anisotropic embeddings score *boilerplate similarity*, and phrasing is the most repeated thing an agent produces — so the vector leg preferentially surfaces the least informative turns in the index.

Raising the number doesn't rescue it: to admit tier 2 on ≤10% of arbitrary queries here you'd need ≈0.99, which is corpus- and model-specific and will rot the moment the embedding model changes. Three options that do survive:

- **Require lexical support** — make the vector score a *ranking* input and keep BM25 as the gate (`&&` not `||`, or "cosine ≥ X **and** BM25 > 0"). Cheapest, and it kills the boilerplate class outright, since the noise hits share phrasing but not rare terms.
- **Score-relative margin** — z-score the candidate cosines for *this* query and require the hit to sit k σ above the candidate mean. Scale-free across models/corpora; no magic constant to recalibrate.
- If you keep an absolute cosine floor, calibrate it against the **top-1** distribution (extreme value), not the marginal, and say so in the comment — the current comment cites p90/p99 of random pairs, which is what made 0.85 look defensible.

My script is at `/tmp/veccal.mjs` on my box if you want to re-run it against your 4,353-turn index.

---

### 2. The BM25 floor means two different things on the two paths (non-blocking, but pick one)

`retrieval.ts` passes `decay` to both branches, but the branches consume it differently:

- **Hybrid (embedder present):** `hybridSearch` calls the inner `search()` with `limit: 25` and **no** `decay` — deliberately, so RRF fuses on raw relevance — and applies decay only to `rrfScore`. `h.ftsScore` is therefore **undecayed**, and `minScore` is a pure relevance floor.
- **FTS-only fallback:** `search()` is called *with* `decay`, so `ftsScore = -bm25 × decayFactor`, and `minScore` is a floor on the **decayed** score.

On that second path, with the defaults (half-life 30d, floor 0.02), the raw BM25 needed to clear 2.0 is 4 at 30 days, 8 at 60, 16 at 90, and ~100 past 180 days — against your own calibration that a genuine single-term match scores ≈4. Tier 2 goes permanently silent for exactly the aged cross-chat memory #377 exists to surface, and only on the no-embedder config, which makes it a nasty one to diagnose in the field. (It's also why the tier-2 fixtures needed recent dates.)

Suggest applying the floor to the raw sub-score on both paths and letting decay govern ordering only.

### 3. `conversationAudience` fails **open** on the room side (non-blocking, landmine)

Inference is by key shape, and the two directions have opposite safety:

- misclassify a candidate as `private` → it's under-shared. Safe.
- misclassify a **room** as `private` → `allowedAudiencesForRoom("private")` returns *all three classes*, so that room receives private DM memories.

Every key shape in the tree today is covered (`telegram:-…`, `phantomchat:group:…`, DMs, `cli:`/`tick:`), so this is not a live leak. But the failure mode for the next channel is silent and it's the bad direction — a `slack:C0123` or `gchat:spaces/…` group lands as `private` and immediately qualifies to receive private content. Cheapest durable fix: have the channel adapter declare the audience on the `Turn` (the column exists now) with key-shape inference as fallback; failing that, a test that enumerates the known channel key shapes so adding one without touching this function fails CI.

### 4. Nits

- `minScore: Math.max(0, minScore)` still admits an explicit `0`, which restores "inject anything that matched at all". A `> 0` clamp (or a warn on 0) would close the last version of this.
- `indexConversationTurnsIfDue` no longer counts reused embeddings, so the v2 rebuild will log `embedded: 0` across thousands of turns and look like the embedder is down. A `reused` counter in that log line would save someone an hour.

Fix 1 (or convince me the extreme-value framing is wrong) and I'm happy — everything else here is non-blocking.

src/config.ts
```diff
@@ -259,6 +264,78 @@ export const DEFAULT_RETRIEVAL_DECAY: RetrievalDecaySettings = {
...
+   * escape hatch.
+   */
+  exclude: string[];
+}
+
+export const DEFAULT_CROSS_CONVERSATION: CrossConversationRetrievalSettings = {
+  enabled: true,
+  limit: 3,
+  minScore: 2.0,
+  minVecScore: 0.85,
```
kaieriksen, line 341: Blocking: `minVecScore: 0.85` is calibrated from marginal random-pair cosine percentiles, but this gate is applied to the maximum over the entire persona index. With thousands of candidates, unrelated boilerplate pairs routinely exceed a marginal p99, so the vector-only `OR` path can fill tier 2 with register similarity rather than relevant memory. Please require lexical support for vector-only candidates, use a query-relative margin, or calibrate against the corpus top-k/extreme-value distribution and add a regression test for boilerplate false positives.

kaieriksen (changes requested):
Changes requested: the SQL filtering and pool-starvation fix are correct, and exact-head verification passes 149/149 with typecheck and hosted CI green. The remaining blocker is the vector-only relevance gate: fixed cosine 0.85 is calibrated from marginal random pairs but applied over the maximum of the whole persona index, so boilerplate similarity can populate tier 2; please make this robust and add a regression test.

src/lib/memoryIndex.ts
```diff
@@ -504,18 +644,26 @@ export class MemoryIndex {
...
+        "INSERT INTO turn_docs (path, persona, conversation, role, turn_id, source, audience, content) " +
+          "VALUES (?, ?, ?, ?, ?, ?, ?, ?)",
       )
       .run(
         path,
         turn.persona,
         turn.conversation,
         turn.role,
         turn.id,
+        turn.source,
```
robertclawson, line 665: **`source` is indexed but nothing ever reads it, and the class Andrew specifically named writes NULL here.**

Two separate things:

**1. Dead column.** `SearchHit.source` is populated (here, `turnMetaForPath`, the FTS SELECTs) but no consumer exists anywhere in `src/` — no filter, no ranking weight, no attribution text. So the provenance half of #377 is schema-only right now. Cheapest useful consumption is in `crossAttribution` (see my comment there) — surfacing the tier costs nothing and lets the model weigh what it's reading, which was the actual request.

**2. `turn.source` is optional, and the one path Andrew called out doesn't set it.** `cli/notify.ts:222` appends the notification turn with no `source`, so scheduled-task and heartbeat notifications land as `source = NULL` — the exact class he asked to tag is the only class with no tag. Worth setting it explicitly on that call (it's a one-liner) rather than leaving it to the column default.

**On reusing `FactSource` instead of a parallel vocabulary** — I agree with not inventing a second enum, but note it can't express what was asked. `principal|self|other|unverified` is a *trust* axis; Andrew's ask was an *origin* axis ("notifications that do not come from a channel or ACP should also have their own tag"). Today a scheduled task's prompt is `other` (tick.ts:231, deliberately) — identical to a group member's message — and my own chat replies and my task-generated conclusions are both `unverified`. Retrieval therefore cannot distinguish "something a human said to me" from "something I concluded on a timer," which is the distinction that matters when a months-old conclusion resurfaces as apparent fact.

Credit where due: the laundering trap I raised *is* covered on the ingest side by the #327 default. This is the other half — origin, not trust. A second UNINDEXED column (`origin`: `channel`/`acp`/`task`/`internal`) is as free as `audience` was, and unlike trust it's knowable at write time. Your call whether that lands here or in a follow-up, but I don't think #377 is closed without it.

src/orchestrator/retrieval.ts
```diff
@@ -204,16 +303,126 @@ export function formatRetrieved(
...
+  },
+): SearchHit[] {
+  const out: SearchHit[] = [];
+  for (const h of candidates) {
+    if (out.length >= opts.limit) break;
+    const conv = turnPathConversation(h.path);
+    if (!conv) continue; // only conversation turns participate in tier 2
+    if (conv === opts.currentConversation) continue;
+    if (isConversationExcluded(conv, opts.exclude)) continue;
+    if (h.audience !== undefined && !opts.allowedAudiences.includes(h.audience))
```
robertclawson: **Defence-in-depth check fails open on the case it exists to catch.**

The doc comment above says these checks are here "so a caller that bypasses the filter still cannot leak across the boundary" — but `h.audience !== undefined &&` means a hit with no audience sails straight through. A bypassing caller is exactly the caller most likely to produce unclassified hits, so the guard is disarmed in its own scenario.

The SQL side is already fail-closed for free (`audience IN (…)` never matches NULL), which makes the asymmetry easy to miss. Suggest matching it:

```ts
if (h.audience === undefined || !opts.allowedAudiences.includes(h.audience))
  continue;
```

I checked this is safe: `turnPathConversation` has already dropped non-turn paths by this line, `upsertTurn` always writes an audience, and vector-only hits that exist in `turn_embeddings` but not `turn_docs` are already excluded upstream by `allowedPaths`. So nothing legitimate is currently undefined here — which is the argument for closing it now, while that's still true.

src/lib/memoryIndex.ts
```diff
@@ -25,11 +25,72 @@ import { readFile } from "node:fs/promises";
...
+
+/**
+ * Derive a conversation's audience class from its key shape. Group keys
+ * carry a `:group:` segment (`phantomchat:group:<id>`); Telegram group and
+ * supergroup chat ids are negative (`telegram:-100…`). Everything else —
+ * DMs, CLI, tick/system contexts — is private. Deriving from the key keeps
+ * this free at index time (no LLM, no config), and the classification is
+ * stable for the life of the conversation.
+ */
+export function conversationAudience(conversation: string): TurnAudience {
```
robertclawson, line 69: **Room classification fails open; candidate classification fails closed.** Worth a comment at minimum, ideally a design change.

Unrecognised key shapes return `private`. For a *candidate* that's conservative (private hides in more places). For the *room* it's the opposite: `allowedAudiencesForRoom("private")` returns all three classes, so an unrecognised room receives everything — including private DM content. The permissive direction is the one taken by anything the regex doesn't know about.

Today that's fine: `:group:` covers phantomchat groups (server.ts:690), `telegram:-` covers Telegram groups and supergroups, and `cli:*` / `phantomchat:<hex>` / ACP are genuinely private. But it's a **regex over a string that no one is obliged to keep in sync** — the next multi-party channel is multi-party in its adapter and private in this function, and nothing fails, tests included. The failure is silent and the blast radius is the leak the whole feature exists to prevent.

The adapter already knows: `ChannelMessage` carries `groupId` (channels/core/types.ts:88) and phantomchat sets `conversationId = "group:<id>"` from it. So the audience is available as fact at write time rather than inference at index time — plumb it onto `Turn` and have `conversationAudience` be the fallback for turns that predate it.

If you'd rather not widen the change, then please at least add a comment here stating that any new multi-party conversation key MUST match one of these shapes, and a test that asserts the classification of every key shape the codebase actually constructs — so adding a channel breaks a test instead of leaking.

src/orchestrator/retrieval.ts
```diff
@@ -204,16 +303,126 @@ export function formatRetrieved(
...
+  "Dec",
+];
+
+/**
+ * Attribution suffix for a cross-conversation hit: channel plus the turn's
+ * date when the snippet carries it (turn content starts with
+ * `[role ISO-timestamp]`, so the date is usually in the snippet). Falls
+ * back to channel-only when no date is parseable.
+ */
+export function crossAttribution(conversation: string, snippet: string): string {
```
robertclawson, line 367: Non-blocking, and the cheapest place to make `source` earn its column: include the provenance tier in the attribution suffix, e.g. `(cross-conversation: Telegram, Aug 13 · unverified)`.

Right now the label carries channel + date, both of which are weakly informative for judging *how much to trust the excerpt*. "This came from a turn I generated myself and you never confirmed" is the single most decision-relevant fact about a resurfaced memory, and it's already sitting on the hit unused.

(Agree with dropping `topic` — deriving it would have cost an LLM call on a path contracted to be cheap and never throw.)

src/orchestrator/retrieval.ts
```diff
@@ -136,7 +152,76 @@ export async function retrieveContext(
...
+        currentConversation: opts.conversation!,
+        exclude: crossExclude,
+        allowedAudiences,
+        minScore,
+        minVecScore,
+        limit: crossLimit,
+      });
+    }
+
+    return formatRetrieved([...hits, ...crossHits], opts.settings);
```
robertclawson, line 232: Non-blocking observation about ordering vs. the token budget.

Cross hits are concatenated after tier 1, and `formatRetrieved`'s budget loop `break`s on the first block that doesn't fit. So tier 2 is dropped precisely when tier 1 is rich — and a single oversized tier-1 snippet ends the loop outright, taking every cross hit with it regardless of score. A hit that cleared a deliberately strict absolute floor loses to a tier-1 hit that cleared `minScore: 0`.

That inverts the priority the floors imply. Since `crossLimit` is only 3, reserving a small slice of `maxTokens` for cross hits (or `continue`-ing past an oversized block instead of `break`ing) would make the feature actually observable in busy conversations. Worth a test either way — the current behaviour is invisible in tests because fixtures are short.

robertclawson (commented):
Re-reviewed at `d92f36b`. My three blockers are genuinely fixed — I checked each against the code rather than the summary:

- **Absolute floor** ✅ `tier1Weakest` is gone; `selectCrossConversationHits` thresholds on `ftsScore`/`vecScore` only, and `DEFAULT_CROSS_CONVERSATION` ships `minScore: 2.0` / `minVecScore: 0.85`. RRF is no longer load-bearing anywhere. Thanks for calibrating against a real index rather than guessing.
- **Audience boundary in SQL** ✅ `turnFilterSql` applies `audience IN (…)` before LIMIT, and the vector side goes through the same `allowedPaths` set — so the boundary holds on both legs of the hybrid, not just FTS.
- **Pool starvation** ✅ all exclusions are pre-LIMIT on both sides.

Also confirmed the embedding-reuse claim: `upsertTurn` deletes only the `turn_docs` row and leaves `turn_embeddings`, and `turnEmbeddingSha` gates the embed call — a v2 rebuild really is zero API calls.

What's left is one substantive gap and three smaller things, inline. The gap: **`source` is written but never read.** Grepping the tree, nothing consumes `SearchHit.source` — not a filter, not a weight, not the attribution string. It's a dead column today, which means the provenance half of Andrew's ask isn't implemented yet, only prepared for. See the inline comments for the specific shape I'd suggest.

Not blocking on my side — flagging for your call and Andrew's.

robertclawson:
Housekeeping — I've left two reviews on `d92f36b` (`4943689…` at 10:53, `4943697998` at 10:59) and they **contradict each other on one point**. My fault; reconciling here so you're not whipsawed:

In the first I wrote *"no need for the finer `task`/`internal` split now"* on the grounds that `turn.ts:389` defaults assistant turns to `unverified`. On the second pass I grepped for consumers of `SearchHit.source` and found **none** — so that reasoning was wrong in a way I couldn't see from the write path alone. `unverified` bounds the *trust* of a task-authored turn, but nothing at retrieval time reads it, and it still can't distinguish a task-generated conclusion from a chat reply. **Treat the second review as the current one on that point.**

Everything else in the first review stands unchanged, including the blocking item — the `minVecScore` calibration is measured against the wrong distribution, and that's the one thing I'd hold the merge on. The new comments are the smaller stuff plus the dead-`source`-column gap.

Sorry for the churn.

lenaparkhodges:
Pushed `d3b02c2` addressing the re-reviews.

**Blocking item (both reviews) — `minVecScore` calibrated against the wrong distribution:** agreed, the extreme-value framing is right and the marginal calibration was the wrong distribution. Fix: the vector leg now requires **lexical support** — cosine ≥ `minVecScore` AND raw BM25 > 0. Cosine alone never admits a hit, which kills the boilerplate/register class outright (your 0.963 nightly-run pair has zero shared query terms). Chose this over z-score margin: cheaper, deterministic, no per-query stats pass. Regression tests added: (a) vector-only boilerplate hit with cosine 0.97 rejected, (b) paraphrase hit with lexical support below the BM25 floor admitted. The `minVecScore` comment now documents the max-over-index distribution and why no absolute standalone value is safe.

**Point 2 (raw vs decayed floor):** fixed — tier-2 candidates now route through `hybridSearch` on both configs, so `ftsScore` is raw everywhere and decay governs ordering only. The 30-day silence on no-embedder configs is gone. (This also surfaced that the floor-regression fixture was quietly leaning on the decay bug — `the` was high-IDF in a tiny corpus and only failed the floor after decay. Fixed by making `the` genuinely low-IDF across the filler set.)

**Defence-in-depth fails open on undefined audience:** fixed — `selectCrossConversationHits` now drops hits with no audience, matching the SQL side. Test added.

**Nits:** non-positive `minScore` is refused (warn + fall back to default); `reused` counter added to the turn-index log line so the v2 rebuild no longer looks like an embedder outage.

**Deferred, not dismissed** (all non-blocking, all larger than this PR's diff should carry):
- **Origin axis (`source` dead column, `origin` channel/acp/task/internal):** convinced by your reconciliation comment — trust can't express origin and `notify.ts:222` is exactly the class Andrew named. Propose a follow-up issue rather than growing this PR further; it touches schema, ingest, and attribution.
- **Adapter-declared audience:** right fix, but touches every channel adapter; for now I'll take the cheaper half in the follow-up too (test enumerating every key shape the codebase constructs, so adding a channel breaks CI instead of leaking).
- **Tier-2 budget reservation vs `break`:** fair inversion; belongs with the formatting pass, not the gate.

Re-review requested. The two `harnesses-pi.test.ts` failures on my host are the known pre-existing ambient-env isolation bug (same class as #373), failing on clean main too.

robertclawson (dismissed):
Re-reviewed at `d3b02c2`. **My blocker is properly fixed — approving.** I checked each claim against the code and ran the suites rather than taking the summary's word for it.

- **Blocking (`minVecScore` calibrated on the wrong distribution)** ✅ `selectCrossConversationHits` now gates on `fts < minScore && !(vec >= minVecScore && fts > 0)`. Cosine alone never admits, which is the fix I asked for, and you picked the cheapest of the three options — right call, no per-query stats pass on a hot path. I confirmed the sign convention actually supports the gate: `search()` emits `ftsScore = -bm25(...)` and FTS5's `bm25()` is negative-is-better, so `fts > 0` is a genuine "shared a query term" test and not accidentally-always-false.
- **Point 2 (raw vs decayed floor)** ✅ Verified in `memoryIndex.ts:1214` — `hybridSearch`'s inner `search()` is called with `limit: 25` and **no** `decay`, and the vector-less branch applies decay only as a sort key. So `ftsScore` really is raw on both branches now and `minScore` is a pure relevance floor everywhere. Routing both configs through `hybridSearch` was the tidy way to get there.
- **Audience fails closed** ✅ `h.audience === undefined` now drops. Worth noting explicitly for anyone reading later: this is safe on existing indexes only because `TURN_SCHEMA_VERSION = 2` forces the v1→v2 turn rebuild, which repopulates `audience` — otherwise every pre-existing row would have gone silently unretrievable. That it costs zero embed calls is what makes it a non-event.
- **Nits** ✅ non-positive `minScore` refused; `reused` counter present and correctly placed *after* the `embeddable === false` quarantine `continue`, so quarantined turns don't inflate `embeddingFailures`.
- **Fixture fix** ✅ making `the` present in every filler is the honest version of that test — it now fails for the reason the test name claims.

Verified locally on `d3b02c2`: `tsc --noEmit` clean; `orchestrator-retrieval` 45/45; `lib-memoryIndex` + `orchestrator-turnIndexer` + `cli-memory` 80/80. I also confirmed your `harnesses-pi.test.ts` claim — **2 fail on `d3b02c2` and the identical 2 fail on clean `origin/main`**, so it's pre-existing and not yours.

Deferring the origin axis, adapter-declared audience, and the tier-2 budget inversion to follow-ups is the right scoping. Agreed on all three.

---

### One non-blocking finding: the vector leg is narrower than it's documented to be

The gate is documented as "cosine ≥ `minVecScore` **and** raw BM25 > 0 (shares at least one query term)". In practice it is **"cosine ≥ `minVecScore` and the turn is in the top-25 BM25 results"**, because `hybridSearch` only populates `ftsScore` for paths present in `ftsHits`, and that list is capped at `limit: 25`:

```ts
const ftsHit = ftsHits.find((h) => h.path === path);
// ...
ftsScore: ftsHit?.ftsScore,   // undefined for a path that matched FTS but ranked >25
```

A vector-ranked path that *did* match the FTS query but landed 26th comes back with `ftsScore: undefined` → `fts = 0` → the vector leg rejects it.

I checked this rather than assuming. Seeded 40 turns repeating `widget` four times each, plus one paraphrase turn containing `widget` exactly once diluted across a long tail, given a perfect query vector:

```
PARA hit: {"path":"turns/p/telegram%3APARA/999","vecScore":1,"rrfScore":0.0163…}
total hits: 26   with ftsScore: 25
ftsScore defined? false   vecScore: 1
```

Cosine **1.0**, genuine lexical support, still rejected — dropped by pool truncation, not by the rule.

Why I'm not blocking on it:

- The direction is **safe** (under-shares, never over-shares), which is the correct way for this to fail.
- The useful band still works: a paraphrase that is inside the top-25 lexically but below the `2.0` BM25 floor is admitted — that's the case the leg exists for, and your new test covers it.

Why it's still worth a follow-up: on a busy index a common query easily has 25 stronger lexical matches, and in that regime the vector leg contributes close to nothing — tier 2 quietly degrades to BM25-only exactly on the high-traffic personas. It'll look like the vector leg "works" in tests (fixture corpora are far under 25) and silently doesn't in the field, which is the same shape of bug as the decayed-floor one you just fixed.

Cheapest fix if you take it: in the `merged` builder you already fall back to `turnMetaForPath(path)` when `ftsHit` is missing — do the same for the score with a targeted `WHERE content MATCH ? AND path = ?` lookup, so `fts > 0` means what the comment says. Failing that, just narrow the comment in `config.ts` and `selectCrossConversationHits` to say "top-`N` lexical pool" so the next reader isn't misled.

Your call whether that rides here or in the follow-up issue with the origin axis — I'd take the comment change here and the lookup there.

Not merging — that's Andrew's call.

kaieriksen (dismissed):
Re-reviewed at `d3b02c2`. The vector relevance blocker is fixed: cosine admission now also requires raw lexical support (`fts > 0`), preventing boilerplate-only vector matches. The raw BM25 floor is consistent across hybrid and FTS-only paths, audience filtering remains SQL-side and pre-LIMIT, and the pool-starvation fix is covered. Exact-head focused tests pass 171/171, typecheck passes, and Linux/Windows/CodeQL CI is green.

kaieriksen (dismissed):
Approved at db809da. The BM25 top-25 lexical-pool limitation is non-blocking and safely under-shares; it should be tracked for follow-up. The substantive vector, audience, provenance, and pool-starvation blockers are fixed. Linux tests, typecheck, and CodeQL pass. Windows has one unrelated pre-existing killProcessGroup integration failure; this PR does not touch that code.

src/lib/memoryIndex.ts
```diff
@@ -1108,18 +1328,77 @@ export class MemoryIndex {
...
+        path: string;
+        rank: number;
+      }>;
+      for (const r of rows2) {
+        const hit = merged.find((h) => h.path === r.path);
+        if (hit) {
+          // bm25() is "lower is better"; flip sign. Apply decay if active
+          // to match the ftsScore convention from search().
+          const raw = -r.rank;
+          hit.ftsScore = decayFactors
```
robertclawson: **Blocking (one line).** This is the only place in `hybridSearch` that writes a decayed value into an emitted score field, and it makes `ftsScore` mean two different things in the same result list.

The inner `search()` above is called **without** `decay` (deliberately — the comment at line ~1225 says the candidate lists "stay relevance-ordered so RRF fuses on merit and decay isn't double-counted"). So every hit that came from the FTS pool carries a **raw** `ftsScore`. This branch gives vector-only hits a **decayed** one.

`selectCrossConversationHits` then thresholds both against `minScore: 2.0`, which is calibrated on raw BM25:

```ts
if (fts < opts.minScore && !(vec >= opts.minVecScore && fts > 0))
```

Measured at this head — 40 fillers in `AAA`, one diluted paraphrase in `BBB`, **all the same age** (181d), `halfLifeDays: 30, floor: 0.02`, with and without `decay`:

```
VECTOR-ONLY BBB  ratio 0.0200   <-- exactly decay.floor
POOL HIT    AAA  ratio 1.0000   <-- raw
```

At the floor a vector-only hit needs raw BM25 >= 100 to clear 2.0; at 30 days it needs 4.0 vs a pool hit's 2.0. Aged vector-only hits therefore almost always fall through to the cosine leg. The `fts > 0` sub-test still behaves (floor `0.02 > 0` preserves the sign), so the leg isn't broken and this only ever *under*-shares — but it re-creates the raw-vs-decayed conflation from `d92f36b` and contradicts the invariant at line 1311 ("emitted hit stays the RAW fused score (what callers threshold on)").

```suggestion
          hit.ftsScore = raw;
```

Also drop the two comment lines above: "Apply decay if active to match the `ftsScore` convention from `search()`" is inverted — `search()` is called undecayed here, so that convention is raw.

Worth a `decay`-enabled case in the new test too; it currently omits `decay` entirely, so this path is unasserted.

robertclawson (changes requested):
Re-reviewed at `db809da`. **The fix works — but it introduces a one-line regression of the exact bug class I blocked on at `d92f36b`.** One round more, then this is done.

**What's genuinely fixed** ✅ The per-path lookup does what I asked: a vector-ranked turn outside the FTS top-25 now gets its real BM25 score instead of `undefined`, so `fts > 0` is a true lexical-support test rather than a proxy for "ranked top-25". Same plain `bm25(turn_docs)` and same sign flip as `search()`, so the scores are comparable. `turnFilterSql` is reused and the parameter order (`ftsQuery2`, paths, `tf2.params`) lines up with the placeholders. Restricting the lookup to `turns/` paths is right — notes never needed it.

Verified at exact head: `tsc --noEmit` clean; `orchestrator-retrieval` + `lib-memoryIndex` + `orchestrator-turnIndexer` + `cli-memory` **126/126 pass**.

### Blocking: vector-only hits get a *decayed* `ftsScore`, pool hits get a *raw* one — both thresholded against the same raw-calibrated `minScore`

See the inline comment. Short version: `hybridSearch` calls its inner `search()` **without** `decay` on purpose, so every pool hit's `ftsScore` is raw. The new branch multiplies by `decayFactors`. Two hits in the same result set now carry two different conventions, and `selectCrossConversationHits` thresholds both against `minScore: 2.0`, which was calibrated on raw BM25.

I reproduced it rather than reasoning from the write path (the mistake I made last round). 40 lexically-competitive fillers in `AAA` + one diluted paraphrase in `BBB`, **all turns the same age** (181d), `halfLifeDays: 30, floor: 0.02`, run twice — once with `decay`, once without:

```
VECTOR-ONLY BBB  decay: 7.393391697260663e-9   nodecay: 3.696695848630331e-7
POOL HIT    AAA  decay: 0.000001723388848660391 nodecay: 0.000001723388848660391
ratio BBB: 0.0200   <-- exactly decay.floor
ratio AAA: 1.0000   <-- raw
```

Identical age, same query, same result set. `0.0200` is exactly the floor, and the only code in `hybridSearch` that multiplies `ftsScore` by `decayFactors` is this new branch — so that ratio is also positive proof the branch fired.

**Why it matters:** at the floor a vector-only hit needs raw BM25 ≥ **100** to clear `minScore: 2.0`; at 30 days it needs ≥ 4.0 where a pool hit needs 2.0. So aged vector-only hits are pushed off the primary gate onto the cosine leg almost unconditionally. Tier 2's whole purpose is surfacing older material from other chats — with `halfLifeDays: 30`, anything past ~2 months sits at the floor, so this narrows the feature precisely in its main use case. It's the same "passes in tests, silently degrades in the field" shape as the top-25 truncation bug, and **the new test never passes `decay`**, so nothing catches it.

It also contradicts the invariant documented 60 lines above at `memoryIndex.ts:1311`: *"`rrfScore` in the emitted hit stays the RAW fused score (what callers threshold on); only the ordering key carries the decay multiplier."* Emitted scores are raw by design here; decay is an ordering key only.

**Direction of failure is safe** (under-shares, never over-shares — no leak), which is why this is one line and not a redesign.

**Fix:** drop the multiplier — `hit.ftsScore = -r.rank;` — and delete the two-line comment above it, whose rationale is inverted (it says it's matching "the `ftsScore` convention from `search()`", but `search()` is deliberately called undecayed on this path). Please add a `decay`-enabled case to the new test asserting a vector-only hit's `ftsScore` equals its undecayed value.

Nothing else outstanding from me. Origin axis, adapter-declared audience and the tier-2 budget inversion stay deferred to follow-ups as agreed. Not merging — that's Andrew's call.

robertclawson (dismissed):
Re-reviewed at `93cc5bc`. **My blocker is fixed — approving.** Verified against the code and by re-running my own reproduction, not the commit message.

**The fix** ✅ `hit.ftsScore = -r.rank;` — the `decayFactors` multiplier is gone and the inverted comment is replaced with one that states the actual invariant (pool hits from `search()` carry raw `ftsScore`; `minScore: 2.0` is calibrated on raw BM25). That restores the single convention across the result set and matches the documented contract 60 lines up at `memoryIndex.ts:1311` — emitted scores raw, decay is an ordering key only.

**Reproduction re-run at exact head**, same fixture as last round (40 lexically-competitive fillers in `AAA`, one diluted paraphrase in `BBB` outside the FTS top-25, all turns the same age, `halfLifeDays: 30, floor: 0.02`), once with `decay` and once without:

```
VECTOR-ONLY BBB  decay: 0.000003  nodecay: 0.000003   ratio 1.0000
POOL HIT    AAA  decay: 0.000003  nodecay: 0.000003   ratio 1.0000
```

`ratio BBB` went `0.0200` (exactly `decay.floor`) → `1.0000`. Only that one line changed between the two runs, so it is direct evidence the branch now emits raw. Aged vector-only hits are back on the primary lexical gate instead of being pushed onto the cosine leg.

**Test** ✅ The new case asserts `bbbHit.ftsScore === rawBbb.ftsScore` across a decayed and an undecayed search of the same fixture. It would genuinely have failed before: 60 days at `halfLifeDays: 30` is a factor of 0.25, well clear of `floor: 0.1`, so the old code would have returned a quarter of the raw score. Equality against a live undecayed run — rather than a hardcoded number — is the right shape; it stays honest if BM25 tuning changes.

**Verified at exact head:** `tsc --noEmit` clean; `lib-memoryIndex` + `orchestrator-retrieval` + `orchestrator-turnIndexer` + `cli-memory` **127/127 pass**. Hosted CI green on `93cc5bc` (`test`, `test-windows`, `typecheck`, both CodeQL analyses). The `test-windows` red on `db809da` was attempt 1 only — attempt 2 passed clean; it was the usual Windows filesystem-timing flake, unrelated to this PR.

Diff is confined to the one line plus the new test. Nothing outstanding from me. Follow-ups stay as agreed: origin axis, adapter-declared audience, tier-2 budget inversion, and Kai's BM25 top-25 lexical-pool note (non-blocking, under-shares).

Not merging — that's Andrew's call.

kaieriksen (dismissed):
Re-reviewed at 93cc5bc. The per-path BM25 lookup now returns the raw undecayed score, matching FTS pool hits and keeping the tier-2 floor consistent for aged turns. Focused retrieval/index/config tests pass 153/153, local typecheck passes, and Linux, Windows, CodeQL, and hosted typecheck are green.

robertclawson (commented):
Production fix is **correct** — I verified it empirically, not from the diff. But the regression test that ships with it is a **tautology**, and so is its sibling: neither test executes the code path they claim to cover.

## The fix itself: confirmed good

`hit.ftsScore = -r.rank` now matches the pool convention. I traced both legs to be sure:

- `hybridSearch` calls `this.search()` at L1215 **without** `decay`, so pool hits carry raw `ftsScore`.
- `hybridSearch`'s own `decayFactors` (L1314) feeds `orderKey` only (L1318) — ordering, not emitted score.
- The FTS-only fallback (L1225) likewise decays ordering while leaving `ftsScore` raw.

So per-path hits and pool hits are now on one scale, thresholded consistently against `minScore: 2.0`. The invariant documented at L1310-1314 holds again. `decayFactors` stays live at L1318, no dead binding. Typecheck clean; 143/143 across `lib-memoryIndex`, `orchestrator-retrieval`, `orchestrator-turnIndexer`, `memory` at `93cc5bc`.

## But the test proves nothing

I reverted the one-line fix and re-ran the new test. **It still passes.**

I instrumented `vecOnlyTurnPaths` to find out why — in both `hybridSearch` calls the list is empty:

```
DBG vecOnly candidates: []
DBG vecOnly candidates: []
 1 pass  0 fail
```

The BBB turn (`relay auth token fix used kind 22242`) is **short**, and BM25 rewards short documents. Against 30 longer fillers it lands comfortably *inside* the FTS top-25, so `ftsScore` is already populated from the pool and the per-path branch never runs. The assertion `expect(bbbHit.ftsScore).toBe(rawBbb.ftsScore)` then compares the raw pool score to itself — it passes for reasons unrelated to the fix.

Running the whole describe block shows the same empty list three times, so **both** tests are affected. The per-path BM25 lookup added in `db809da2` — the feature that resolved my earlier non-blocking finding — currently has **zero effective coverage**. The comment at L1341-1348 describes behaviour nothing exercises.

## Working repro

To get BBB genuinely vector-only, invert the length relationship: dense short fillers, one long BBB where the query terms appear once. That pushes it to 26th+ lexically while its cosine keeps it in the final slice. Needs `limit` above the default so it survives the slice.

```ts
for (let i = 1; i <= 30; i++) {
  ix.upsertTurn({ id: i, persona: "phantom", conversation: "phantomchat:group:AAA",
    role: "user", text: "relay auth token relay auth token",
    createdAt: new Date("2026-05-28T06:00:00Z"), embeddable: true, source: "principal" },
    new Float32Array([0, i / 30, 0]), `sha-aaa-${i}`);
}
const pad = Array.from({ length: 120 }, (_, k) => `padding${k}`).join(" ");
ix.upsertTurn({ id: 300, persona: "phantom", conversation: "telegram:-100BBB",
  role: "user", text: `${pad} relay auth token ${pad}`,
  createdAt: new Date("2026-05-28T06:01:00Z"), embeddable: true, source: "principal" },
  new Float32Array([1, 0, 0]), "sha-bbb");

const decayed = ix.hybridSearch(query, vec, { scope: "all", limit: 40,
  decay: { halfLifeDays: 30, floor: 0.02, nowMs: Date.parse("2026-11-28T06:00:00Z") } });
const raw = ix.hybridSearch(query, vec, { scope: "all", limit: 40 });
expect(decayed.find(h => h.path.includes("BBB"))!.ftsScore)
  .toBe(raw.find(h => h.path.includes("BBB"))!.ftsScore);
```

Instrumented, this one **does** enter the branch (`DBG vecOnly: ["turns/phantom/telegram%3A-100BBB/300"]`) and it discriminates correctly:

| | shipped test | this repro |
|---|---|---|
| with fix | pass | pass (ratio `1.0000`) |
| fix reverted | **pass** | **fail** (ratio `0.0200` = `decay.floor`) |

That `0.0200` is the bug reproduced exactly: same age, same query, same result set, two score conventions.

## Not blocking the merge

The shipped code is right and the failure direction was always safe (under-shares, never leaks). I'm not requesting changes on correctness — my earlier blocking finding is **resolved**. But I'd swap the test before merge rather than after: a green regression test that cannot go red is worse than no test, because the next person to touch this path will trust it. Andrew has the merge call.

One pattern worth naming: this is the third round where a test was added alongside a correct fix without confirming it fails first. Cheap habit to adopt — revert the fix, watch the test go red, restore. Catches exactly this.

robertclawson:
**Correction to my own approval ([#4943937049](https://github.com/phantomyard/phantombot/pull/378#pullrequestreview-4943937049)) — one claim in it is wrong.**

That review said of the new test: *"It would genuinely have failed before: 60 days at `halfLifeDays: 30` is a factor of 0.25, well clear of `floor: 0.1`."* The arithmetic is right but the premise isn't — it assumes the per-path branch executes. It doesn't.

I reverted `hit.ftsScore = -r.rank` to the decayed form and ran the shipped test: **it still passes**. Instrumenting `vecOnlyTurnPaths` shows why — the list is empty on every call, because the BBB fixture turn is *short* and BM25 favours short documents, so it sits inside the FTS top-25 and `ftsScore` is already filled from the pool. The decay factor never gets a chance to apply.

So my approval stands on **the production fix** — that part I verified by reproduction and it is correct — but **not** on the test, which cannot fail. Details and a working replacement fixture are in [my follow-up review](https://github.com/phantomyard/phantombot/pull/378#pullrequestreview-4943938090). Where the two conflict, the later one is right.

Still not a correctness blocker, and still Andrew's merge call.

robertclawson (approved):
**Approving.** The replacement fixture does what the old one didn't: it goes red when the fix is reverted. Verified empirically at `8c67465`, not from the diff.

## The test now discriminates

Reverted the one-line production fix (restored `raw * (decayFactors.get(r.path) ?? 1)`) and re-ran:

```
expect(bbbHit!.ftsScore).toBe(rawBbb!.ftsScore)
Expected: 5.410529012424748e-7
Received: 1.0821058024849495e-8
1 fail
```

Ratio is exactly `0.02` — `decay.floor`, i.e. the bug reproduced precisely: same turn, same query, two score conventions. Restored the fix, 49/49 green again.

## Both tests actually enter the branch now

Instrumented `vecOnlyTurnPaths` at L1352, which was the root of the old tautology (empty list three times):

```
DBG vecOnly: ["turns/phantom/telegram%3A-100BBB/300"]
DBG vecOnly: ["turns/phantom/telegram%3A-100BBB/300"]
DBG vecOnly: ["turns/phantom/telegram%3A-100BBB/300"]
 2 pass  0 fail
```

All three `hybridSearch` calls hit the per-path lookup. The length inversion (dense short fillers vs. one long padded BBB) puts BBB below the FTS top-25 as intended, and `limit: 40` keeps it in the slice. The shared `seed()` helper means both tests are covered by one correct fixture — the sibling test is no longer a tautology either. The per-path BM25 lookup from `db809da2` has real coverage for the first time.

## Local verification at `8c67465`

- `lib-memoryIndex` 49/49, plus `orchestrator-retrieval`, `orchestrator-turnIndexer`, `config`: **201/202**.
- `tsc --noEmit`: clean.
- CI: Linux, Windows, typecheck, CodeQL all green.

The one failure is `memory.test.ts` → *"openMemoryStore blocks until a briefly-held lock releases"*. **Pre-existing and unrelated** — it fails identically on `origin/main` with this branch checked out nowhere near it, and reproduces on repeat runs (not a flake). Not this PR's problem; I'll open it separately rather than hang it on #378.

No outstanding findings from me. Merge is Andrew's call.

kaieriksen (approved):
Re-reviewed at 8c67465. The BM25 regression fixture now genuinely places BBB below the FTS top-25 (long padded turn versus dense short fillers), so both tests exercise the per-path lookup and raw-score behavior; targeted tests pass 2/2 and typecheck passes. Hosted test, Windows, typecheck, and CodeQL checks are green.

tests/lib-memoryIndex.test.ts
```diff
@@ -689,3 +690,280 @@ describe("turn-hit time decay (search / hybridSearch)", () => {
...
+    const rawHits = ix.hybridSearch(query, vec, {
+      scope: "all",
+      limit: 40,
+    });
+    const rawBbb = rawHits.find((h) => h.path.includes("BBB"));
+    expect(rawBbb).toBeDefined();
+    expect(rawBbb!.ftsScore).toBeDefined();
+
+    // Per-path ftsScore must equal the undecayed value, not decayed.
+    expect(bbbHit!.ftsScore).toBe(rawBbb!.ftsScore);
```
robertclawson, line 967: Non-blocking, worth a one-line comment so a future reader doesn't "fix" this fixture.

The raw BM25 this fixture produces is **`5.4e-7`** — effectively zero, because all 31 seeded docs contain every query term, so IDF collapses. That's far below the production tier-2 floor (`minScore: 2.0`).

That doesn't weaken the test: the invariant under test is *score convention* (per-path must equal undecayed), and equality holds regardless of magnitude. But it does mean the fixture's hit only ever survives the dual gate via the `fts > 0` leg at `retrieval.ts:440`, never the `fts >= minScore` leg.

Consequence worth naming: for the `fts > 0` leg specifically, the decay bug was **sign-preserving** (`5.4e-7 × 0.02` is still `> 0`), so it never changed that leg's outcome. The bug only ever bit hits whose raw BM25 sat *above* `2.0` and got decayed below it — which this fixture cannot reach. A comment saying "magnitude is incidental; this asserts convention, not gate admission" would stop someone later concluding the test is mis-calibrated and dropping it.

robertclawson (approved):
## Approving — the tautology is gone, verified by experiment not by reading

My last review blocked on this: the previous fixture passed *with the bug reintroduced*, because the short BBB doc landed inside the FTS top-25 and `vecOnlyTurnPaths` was empty on every call. The per-path lookup from `db809da` had zero coverage.

That is now genuinely fixed. I ran two independent revert experiments at exact head `8c67465`:

**Experiment 1 — revert the `93cc5bc` production fix** (`hit.ftsScore = -r.rank` → decay multiplier restored):

```
(fail) per-path BM25 score is raw (undecayed) even when decay is active
  Expected: 5.410529012424748e-7
  Received: 1.0821058024849495e-8
  48 pass, 1 fail
```

Ratio is exactly `0.02` — the configured `decay.floor`. The test goes red on the precise defect it exists for.

**Experiment 2 — disable the whole per-path block** (`if (false && vecOnlyTurnPaths.length > 0)`, simulating pre-`db809da`):

```
47 pass, 2 fail
  → "vector hit outside FTS top-25 gets its real BM25 score"  ftsScore undefined
  → "per-path BM25 score is raw (undecayed) even when decay is active"  ftsScore undefined
```

**Both** tests fail. So the first test now covers `db809da` and the second covers `93cc5bc` — each guards a distinct fix, and neither can pass vacuously. The inverted length relationship (dense short fillers vs one long padded BBB) does exactly what it claims: BBB drops below the top-25 lexically while its cosine keeps it in the `limit: 40` slice.

Also checked: `beforeEach` opens a fresh `:memory:` index per test, so the shared `seed()` helper can't cross-contaminate between the two cases.

**Verification at `8c67465`:** `tsc --noEmit` clean; `lib-memoryIndex` 49/49; plus `memory`, `channels-commands`, `orchestrator-retrieval`, `memory-durableFacts` — **225/225 pass, 0 fail** across the five relevant suites. All six CI checks green.

Nothing outstanding from me. Every finding I raised across the four rounds is resolved.

src/orchestrator/retrieval.ts
robertclawson: Resolved at `8c67465`. The bar is no longer derived: `retrieval.ts:176-188` takes an ABSOLUTE floor and clamps a non-positive configured value back to the default with a warning, so `minScore = 0` can no longer collapse the gate to "anything that matched FTS at all".

src/orchestrator/retrieval.ts
robertclawson: Resolved at `8c67465`. `selectCrossConversationHits` now gates on raw BM25 and cosine only — RRF is explicitly ignored, with the reason recorded in the comment at `retrieval.ts:430-432` ("it encodes rank within this result set, not relevance"). No cross-result-set score comparison remains.

src/orchestrator/retrieval.ts
robertclawson: Resolved at `8c67465`. The boundary moved out of the prompt and into the query: `turnFilter.allowedAudiences` is applied in SQL, re-checked in `selectCrossConversationHits` as defence in depth, and covered by real containment tests — private-DM turn never surfaces in a group room, with a positive control proving the miss is the audience boundary and not the floor (`tests/orchestrator-retrieval.test.ts:600-665`). The header sentence stays, but it is now advice on top of an enforced perimeter rather than the perimeter itself.

src/lib/memoryIndex.ts
robertclawson: Resolved at `8c67465`. `turn_docs` now carries `source` and `audience` as indexed columns (`memoryIndex.ts:325-335`) and `upsertTurn` writes both, so provenance is recorded at index time rather than inferred downstream. The remaining gap — nothing in retrieval *reads* `source` yet — is tracked separately in #379 (non-blocking).

src/orchestrator/retrieval.ts
robertclawson: Resolved at `8c67465`. Exclusions (current conversation, exclude list, audience) are applied in SQL BEFORE the LIMIT, so every pool slot is an eligible hit. An active chat can no longer starve tier 2 by occupying the whole candidate pool with its own turns.

tests/orchestrator-retrieval.test.ts
robertclawson: Resolved at `8c67465`. The suite now has the negative test that actually guards containment — a private-DM turn is absent from a group room's retrieval — plus a positive control showing the same turn does surface in a private room, and the reverse direction (group turn MAY surface in a private room). The header-string assertion survives as plumbing coverage, which is all it ever claimed to be.

src/orchestrator/retrieval.ts
robertclawson: Resolved at `8c67465`. Now fails closed: `if (h.audience === undefined || !opts.allowedAudiences.includes(h.audience)) continue;` (`retrieval.ts:419`), with a dedicated test — "fails closed when a candidate carries no audience" (`tests/orchestrator-retrieval.test.ts:873`).

src/lib/memoryIndex.ts
robertclawson: Resolved at `93cc5bc`, and the regression test now genuinely covers it as of `8c67465`. Verified by reverting the one-line fix locally: the test goes red with `Expected 5.410529012424748e-7 / Received 1.0821058024849495e-8` — ratio exactly `0.02`, the decay floor. Restored, 49/49 green.

src/orchestrator/retrieval.ts
robertclawson: Addressed at `d92f36be` and verified at `8c67465`: the conversation/exclude/audience filter now runs in SQL before the LIMIT (`TurnFilter` in `memoryIndex.ts`), so the candidate pool can no longer be consumed by hits that are about to be discarded. Resolving on the basis of your re-approval at `93cc5bc` — reopen if you read it differently.

src/config.ts
robertclawson: Addressed at `d3b02c2d` and verified at `8c67465`: the vector leg now requires lexical support — `fts < minScore && !(vec >= minVecScore && fts > 0)` (`retrieval.ts:440`), so cosine alone is never a bar and the boilerplate/register-similarity class is excluded. Your 79%-of-queries measurement is recorded in the comment above it. Resolving on the basis of your re-approval at `93cc5bc` — reopen if you read it differently.

src/lib/memoryIndex.ts
robertclawson: Deferring rather than dropping: tracked as items 1-2 of #379. `source` is now indexed but still has no consumer in retrieval; the cheapest one is the attribution suffix. Non-blocking for #378 — resolving here so the merge isn't held by a follow-up.

src/lib/memoryIndex.ts
robertclawson: Deferring rather than dropping: tracked as item 3 of #379. The derivation is now documented at `memoryIndex.ts:61-72`, but the asymmetry stands — unknown room shape still receives every audience class. Latent today (all live channels are covered), bites when a new channel lands with an unfamiliar key. Non-blocking for #378.

src/orchestrator/retrieval.ts
robertclawson: Deferring rather than dropping: tracked as item 2 of #379. Non-blocking for #378.

src/orchestrator/retrieval.ts
robertclawson: Deferring rather than dropping: tracked as item 4 of #379. Non-blocking for #378 — tier 2 losing to a marginal tier-1 hit under budget pressure is a quality loss, not a correctness or containment one.

tests/lib-memoryIndex.test.ts
robertclawson: Deferring rather than dropping: tracked as item 5 of #379. The test is sound as shipped — reverting the fix makes it fail, verified — this is only a note for the next reader. Non-blocking for #378.

robertclawson:
Resolving my 13 review threads on this PR — approved at `8c67465`, and every blocking finding of mine is verified fixed in code (not just claimed):

- absolute tier-2 floor, `tier1Weakest` deleted (raw BM25 / cosine only)
- RRF cross-search comparison gone
- audience boundary enforced as a **SQL retrieval filter** on both the FTS and vector legs, with the negative containment test (private source → group room → no hit) plus its positive control
- pool starvation fixed (all exclusions applied pre-LIMIT)
- JS audience guard now fail-**closed** (`h.audience === undefined || …`)
- `ftsScore` convention unified (`hit.ftsScore = raw`), and the guarding test now genuinely goes red when the fix is reverted (I re-ran it both ways: `1 fail`, ratio `0.02` = `decay.floor`)

**Four non-blocking items I raised are NOT addressed and I'm resolving them anyway rather than holding the merge.** Recording them here so they don't disappear with the threads:

1. **`source` is a dead column.** Written in `upsertTurn` / `turnMetaForPath` / the FTS SELECTs, read by nothing in `src/`. No filter, no weight, not in `crossAttribution` (verified at head — zero `.source` consumers in `retrieval.ts`). The provenance half of #377 is schema-only today.
2. **`cli/notify.ts:222` appends without `source`** — scheduled-task and heartbeat notifications land as `NULL`. That's the exact class Andrew asked to tag, and it's the one class with no tag. Also: `FactSource` is a *trust* axis (`principal|self|other|unverified`); the ask was an *origin* axis (`channel`/`acp`/`task`/`internal`). A scheduled task's prompt is `other`, same as a group member's message.
3. **`conversationAudience` fails open for rooms.** Unrecognised keys → `private` → `allowedAudiencesForRoom("private")` returns all three classes, so a future multi-party channel receives private DM content. There is still no test asserting the classification of every key shape the codebase constructs, so adding a channel leaks silently instead of breaking a test. The adapter already knows via `ChannelMessage.groupId` — plumbing it makes audience a stored fact rather than a regex inference.
4. **Cross hits are appended last**, so `formatRetrieved`'s budget loop drops them exactly when tier 1 is rich — a hit that cleared a strict absolute floor loses to a tier-1 hit that cleared `minScore: 0`.

(1)–(3) are the same theme: provenance was asked for and only half-landed. Worth a follow-up issue rather than a silent close.

robertclawson:
Correction to my approval note — the one failing test was **my environment, not the repo**. Don't let it linger as a phantom bug.

`memory.test.ts` → *"openMemoryStore blocks until a briefly-held lock releases"* spawns a real second process with `Bun.spawn(["bun", ...])`. I was invoking bun by absolute path (`~/.bun/bin/bun`) without it on `$PATH`, so the child died with `ENOENT: Executable not found in $PATH: "bun"`. With `PATH="$HOME/.bun/bin:$PATH"`, **29/29 pass**.

So the full local picture at `8c67465` is clean: `lib-memoryIndex` 49/49, focused suites **202/202**, `tsc --noEmit` clean, CI green. No separate issue to file — my earlier "I'll open it separately" is withdrawn.

PR #380, opened 2026-08-15
feat(retrieval): turn origin axis + fix fail-open room audience

robertclawson:
Follow-up to #378 / closes the remaining half of #377.

#378 added the `source` column but left it **inert** — written at index time, read by nothing in `src/`. This makes provenance actually do something, and fixes a fail-open in the audience boundary that shipped alongside it.

---

## 1. Origin axis (`TurnOrigin`)

`source` is a **trust** tier. It can't express how a turn was *produced*, so a scheduled task's own output and a stranger's message in a group chat both land `source: "other"` (see `cli/tick.ts`). That matters: task output is the persona talking to itself, and surfaced by tier-2 weeks later it reads as established fact when nobody ever checked it.

New orthogonal column `turns.origin` / `turn_docs.origin`:

| origin | meaning |
|---|---|
| `channel` | chat surface (Telegram, phantomchat, CLI, ACP) |
| `task` | scheduled task wake — our own prompt, our own reasoning |
| `notification` | `phantombot notify` payload persisted for continuity |
| `internal` | nightly / other machine-driven writes |

Stamped at every writer: `tick.ts` → `task`, `notify.ts` → `notification`, `nightly.ts` → `internal`, everything else defaults to `channel`.

Consumed in retrieval as **weight + label, not a gate** — self-produced content is down-weighted in ordering and labelled in attribution, but the admission floor is untouched. Filtering here would re-introduce the double-gating rejected in #378 review.

Migration is idempotent and retro-tags only on markers the writers emit verbatim (`tick:%` / `system:%` keys, the literal `[notification] ` prefix). Anything unmatched stays `channel` = pre-column behaviour.

## 2. Fail-open room audience 🐛

`conversationAudience()` defaulted an unrecognised key shape to `private`. That's fail-**closed** when stamping a stored turn, but fail-**open** when classifying the room being spoken into: a future multi-party channel with an unrecognised key would be treated as a private room, and **every private memory in the persona would become eligible to surface in it** — silently, with no test failing.

Split into `turnAudience()` (unknown → `private`) and `roomAudience()` (unknown → `multi-party`), because the two callers need opposite safe defaults. `classifyConversationAudience()` now returns `undefined` rather than guessing, so the defaulting decision sits with the caller that knows which direction is safe. Unknown room shapes warn once per key.

Telegram matching now also handles suffixed keys — forum topics arrive as `telegram:<id>:topic:<n>` and previously fell through to the default.

## 3. Latent bug: snippet column ordinal

Adding a column before `content` shifted it to index 8, but `snippet(turn_docs, 7, …)` was hardcoded. FTS5 takes a bare integer and does **not** validate it against the column set — pointed at an `UNINDEXED` column, every snippet silently returns empty, which reads as "no results" rather than as an error. Now a named constant sitting next to the DDL.

---

## Not included: post-rebuild backfill

An earlier version of this branch force-backfilled the turn index after a schema rebuild, on the belief that #378 had left the lexical index permanently short until an operator ran `memory index --turns`.

**That was wrong, and I want it on the record rather than quietly dropped.** The rebuild also clears `turn_index_state`, which rewinds every conversation cursor to 0 — so the whole store reads as an unindexed tail and the next *ordinary* sweep re-indexes it. Worst case is one heartbeat interval (30 min) of degraded FTS recall. What I saw on the live box was a snapshot taken inside that window.

The forced backfill was pure duplication, so it's dropped. Two tests pin the real mechanism instead (`turn-schema rebuild recovery`), **both verified to fail when the cursor reset is removed** — so if someone later "optimises" the rebuild by preserving `turn_index_state`, recovery stops happening silently and they go red.

## Verification

- `tsc --noEmit` clean.
- Full suite: **2731 pass, 0 fail** (167 files).
- New recovery tests confirmed non-vacuous by deleting `DELETE FROM turn_index_state;` and re-running → `1 fail`, `triggered` 0 vs expected 1.

### DIFF
--- a/src/memory/store.ts
+++ b/src/memory/store.ts
@@ -51,6 +51,11 @@ export function asFactSource(v: unknown): FactSource {
     : "principal";
 }
 
+/** Coerce a stored origin string; anything unrecognised reads as `channel`. */
+export function asTurnOrigin(v: unknown): TurnOrigin {
+  return isTurnOrigin(v) ? v : "channel";
+}
+
 export interface Turn {
   id: number;
   persona: string;
@@ -79,6 +84,45 @@ export interface Turn {
    * `self`, but new assistant turns land `unverified`.
    */
   source: FactSource;
+  /**
+   * ORIGIN of this turn — the mechanism that produced it. Orthogonal to
+   * `source`, which is a TRUST tier: the two answer different questions and
+   * conflating them is what made scheduled-task output indistinguishable
+   * from a stranger's message in a group chat (both land `source: "other"`,
+   * see cli/tick.ts).
+   *
+   *   `channel`      — arrived over a chat surface (Telegram, phantomchat,
+   *                    CLI, ACP): a human wrote it, or we replied to one.
+   *   `task`         — a scheduled task wake. The prompt is one WE wrote and
+   *                    the reply is our own reasoning, unreviewed by anyone.
+   *   `notification` — a `phantombot notify` payload persisted for continuity.
+   *   `internal`     — heartbeat / nightly / other machine-driven writes.
+   *
+   * Why it matters for retrieval: `task` and `internal` content is the
+   * persona talking to itself. Surfaced later it reads as established fact
+   * when it was never checked by anyone, so it is down-weighted and labelled
+   * rather than filtered (weight + label, not a gate).
+   *
+   * Rows written before this column default to `channel`.
+   */
+  origin: TurnOrigin;
+}
+
+/**
+ * How a turn came into existence. See `Turn.origin` — this is the ORIGIN
+ * axis, deliberately separate from the `FactSource` TRUST axis.
+ */
+export type TurnOrigin = "channel" | "task" | "notification" | "internal";
+
+export const TURN_ORIGINS: readonly TurnOrigin[] = [
+  "channel",
+  "task",
+  "notification",
+  "internal",
+];
+
+export function isTurnOrigin(v: unknown): v is TurnOrigin {
+  return typeof v === "string" && TURN_ORIGINS.includes(v as TurnOrigin);
 }
 
 export interface AppendTurnInput {
@@ -94,6 +138,12 @@ export interface AppendTurnInput {
    * principal engages with it.)
    */
   source?: FactSource;
+  /**
+   * Origin axis (see Turn.origin). Optional on input; defaults to `channel`,
+   * which is correct for every chat-surface write. Non-channel writers
+   * (tick task wakes, notify, heartbeat/nightly) pass theirs explicitly.
+   */
+  origin?: TurnOrigin;
   /**
    * Index-eligibility flag. Defaults to true. Set false to QUARANTINE the
    * row — it persists and replays in history, but the turn indexer skips it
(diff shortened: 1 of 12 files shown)


src/memory/store.ts
```diff
@@ -1126,6 +1209,7 @@ class SqliteMemoryStore implements MemoryStore {
           ts,
           embeddableInt(a.embeddable),
           defaultTurnSource(a),
+          a.origin ?? "channel",
```
kaieriksen, line 1212: **[P1] Mark held-episode judge notes as internal origin**\n\nThe default recordHeld path writes the judge's own episode.notifyText, but it does not pass origin: "internal". This fallback therefore stores the assistant row as channel, so retrieval later weights/labels machine-generated judge reasoning like a chat-produced turn; the migration cannot retro-tag it because the conversation is usually telegram:<id>. Pass origin: "internal" on the assistant row (the quarantined payload is not indexed) and add a regression assertion on the persisted origin.

kaieriksen (changes requested):
The origin/audience implementation is otherwise coherent, but held-episode judge notes still persist with the default channel origin. Please stamp that machine-generated assistant row as internal and add the regression assertion from the inline comment.

robertclawson:
Fixed in c7b1433 — thanks, you're right.

The default `recordHeld` now stamps the judge's assistant row `origin: "internal"`. Your point about the migration is the sharp end of it: that row is written into the *principal's* conversation (`telegram:<id>`), so none of the retro-tag markers (`tick:%`, `system:%`, the `[notification] ` prefix) can ever reach it — unstamped it stays `channel` permanently, and retrieval weights the judge's own reasoning like a chat-produced turn.

Left the quarantined user row at the default deliberately, as you noted: `embeddable: false` means it is never indexed, and it is purged once a trusted turn rules.

Two regression assertions rather than one:
- `default recordHeld writes a quarantined user turn + embeddable assistant turn` — pins the `appendTurnPair` **input**.
- `default recordHeld PERSISTS the judge turn with origin internal` — new test against a real `openMemoryStore(":memory:")`, reading the row back via `recentTurnsForDisplay`, so a regression in the store's own origin plumbing is caught too.

Both verified non-vacuous: removing the stamp gives `2 fail`.

`tsc --noEmit` clean; full suite **2732 pass, 2 skip, 0 fail** (167 files).

src/memory/store.ts
robertclawson: Fixed in `a8d9f9d` — good catch, this was a genuine gap. The judge's `notifyText` was landing at the default `channel`, and you're right that the migration can never reach it: the row goes into `telegram:<id>`, which matches neither the `tick:%` nor `system:%` marker.

**One deviation: I stamped it `notification`, not `internal`.** Reasoning, verified in code rather than assumed:

The HOLD branch sends that exact string through `runNotify` (`screen.ts` L392), and `runNotify` already persists it — `cli/notify.ts:222`, to `telegram:${chatId}` (L286), stamped `origin: "notification"` (L229). That is the *same* conversation key `principalConversations()` grounds into. So the identical text lands twice in one conversation.

Tagging this second copy `internal` would give byte-identical content two different weights (`CROSS_ORIGIN_WEIGHT`: `internal` 0.6 vs `notification` 0.8) and two contradictory labels — and the `internal` one is factually wrong. `internal` means a nightly/heartbeat maintenance pass; this is a notification the persona sent to its principal. `notification` makes the duplicate agree with itself.

Happy to switch if you think the trust-tier argument outweighs the self-consistency one, but I'd want the duplicate-row question settled first either way.

**Separate finding your comment surfaced indirectly:** that duplicate is worth addressing on its own — same text, two rows, one prefixed by notify and one not, both embeddable, both indexed. Tier 2 can surface both for one query. I'd rather not widen this PR for it; flagging for a follow-up unless you feel it blocks.

Verification: both new assertions confirmed to fail without the stamp (I reverted the one-liner and re-ran — `2 fail`, restored → `22 pass`). Typecheck clean with explicit exit capture, full suite **2732 pass / 0 fail**.

kaieriksen (approved):
Re-reviewed at a8d9f9d. The held-episode assistant row is now explicitly stamped origin=notification, and the new regression verifies the persisted store value. Focused tests pass 138/138, typecheck passes, and hosted CI is green.

PR #383, opened 2026-08-15
fix(cli): gate persona switch on confirm + refuse agent-scoped switches (closes #371)

### LINKED ISSUE #371: cli: persona <name> persists default switch before confirm + no scope check → agent-triggered ghost persona switches
## Summary

`phantombot persona <name>` persists the default-persona switch to `state.json` **before** asking for confirmation, and does not validate scopes or detect non-TTY contexts. A phantombot agent (which has a Bash tool) can therefore change the *global default persona* of the daemon by accident — and the change only becomes visible after the next restart, making it look like a ghost/unauthorized switch.

## Observed incident (real, 2026-08-13)

On a multi-persona deployment (MacBookPro, 9 personas, 5 bot personas behind 1 daemon):

1. At **12:12:11Z** the `alma` persona agent was working with its human in a Telegram group (capturing a decision in memory via `phantombot memory capture`).
2. At **12:12:19–29Z** it explored `phantombot memory --help`, `phantombot memory today`, `phantombot --help` — trying to read *its own* daily memory.
3. At **12:12:37Z** it ran `phantombot persona alma 2>&1 | tail -5 && phantombot memory today` via its Bash tool. Its intent: make `phantombot memory today` operate under the `alma` persona. It had **no idea** this flips the daemon's global `default_persona`.
4. The CLI persisted `state.default_persona: paco → alma` immediately, then hit `maybePromptRestart()` → `@clack` confirm in a non-TTY (agent-spawned child process) → the prompt never got a real answer.
5. The daemon kept running with the old in-memory default, so nothing was visible — until the next restart (a `phantombot update` at 23:55Z) loaded `state.json` → `default_persona: alma`.
6. Next morning, the *default Telegram bot* (which has no persona-specific subsection in config.toml, so it responds as `default_persona`) started answering as `alma` instead of `paco`. The human thought a hidden script had switched personas.

Prior "mysterious" switches (Aug 8 / Aug 10, rapid paco↔pepa with `--yes`) were meeting auto-join tests — also agent-triggered.

## Root cause

`src/cli/persona.ts` → `runSwitchPersona`:

```ts
state.default_persona = input.name;
await saveState(state);          // ← persisted BEFORE any confirmation
out.write(`default_persona: '${previous}' → '${input.name}'\n`);
const svc = input.serviceControl ?? defaultServiceControl();
await maybePromptRestart(svc);   // ← confirm() afterwards; non-TTY => no real answer
```

- The mutation is persisted **before** `maybePromptRestart`'s confirm. In a non-TTY context (agent Bash tool, cron, CI) the confirm cannot be answered, so the write stands.
- The default persona is a **global daemon-wide setting**, but any persona agent can mutate it — there is **no scope check** (scopes.json is only consulted in `src/harnesses/pi.ts` / memory-index paths, not in the persona-switch CLI path).
- Effect is delayed until next restart, so it reads as "spontaneous" / unauthorized.

## Proposed fix (pick any/all)

1. **Require confirmation before persisting** in non-interactive contexts: if `process.stdout.isTTY` is false (or no `--yes` flag is passed), print the change to stderr and refuse with exit code 2, or require an explicit `--yes`.
2. **Validate scopes**: if the invoking context is a persona agent (e.g. `PHANTOMBOT_PERSONA` env set / agent harness), restrict `persona` switches to personas in that agent's scope.
3. **Don't swallow non-TTY prompts**: in `maybePromptRestart`, detect `!process.stdout.isTTY` and print `skipped — restart later with: …` instead of letting @clack hang/fail silently.
4. Optionally: **log persona switches** to a state audit trail (timestamp, invoking pid/ppid, env) so future switches are attributable.

## Impact

Operational confusion and apparent "ghost" persona switches in multi-persona deployments; a low-severity privilege issue (any persona agent can re-point the daemon's default persona without authorization).

Happy to contribute a PR (this repo's contributor pattern: issue → PR with tests).

salvaalba-dev:
## Summary

Closes #371.

`phantombot persona <name>` persisted `state.default_persona` **before** any confirmation, and had no scope check. A phantombot agent (Bash tool) running `phantombot persona <name>` — intending to scope its own `memory today` read — silently re-pointed the daemon-wide default persona, surfacing as a "ghost" switch after the next restart.

## Changes

1. **Confirm before persisting** (the ordering fix). The write to `state.json` now happens only after an explicit confirmation:
   - Interactive TTY → prompts via `@clack` (`defaultConfirm`).
   - Non-TTY (agent Bash, cron, CI) without `--yes` → refuses with exit code 2 and a clear message pointing at `--yes`.
   - `--yes` → explicit non-interactive consent.
   - Injected `confirm` fn (tests) → drives the ordering end-to-end without a TTY.
2. **Scope check.** If `PHANTOMBOT_PERSONA` is set (harness/agent context), the switch is refused with exit code 2 — a persona agent can't re-point the daemon-wide default. (`--persona` remains the correct per-invocation scoping.) Placed **before** `--yes` so `--yes` can't be used to self-authorize.
3. **Exported `defaultConfirm`** from `harness.ts` so the switch path reuses the same @clack-backed confirm.
4. **Race hardening** (adversarial-review follow-up):
   - Re-load `state.json` after confirmation instead of reusing the pre-confirm snapshot — a concurrent writer (e.g. `harness_bins` discovery) during the confirm wait is no longer clobbered.
   - Re-validate the persona dir exists right before commit, so the default can't be pointed at a since-removed persona.
   - Success message reports the freshest `previous` from the re-loaded state (not the stale pre-confirm snapshot).

## Tests (fail on current `main`)

- `non-TTY without --yes → refuses, no state write` — exit 2, `state.json` untouched.
- `confirm=false → cancelled, no state write`.
- `agent context (PHANTOMBOT_PERSONA set) → refuse, no state write`.
- `agent context + --yes + different target → still refused` — scope check wins over `--yes`.
- `interactive TTY confirm=true → persists via defaultConfirm` (drives @clack with a mock).
- `interactive TTY confirm=false → cancelled`.
- `concurrent writer during confirmation → other fields preserved` (lost-update regression; also asserts the success message reflects the freshest previous value).

Verified: the regression tests fail against `upstream/main` (reproducing pre-fix behavior); full suite passes on this branch (2739 pass / 0 fail).

## Notes / open questions for review

- Scope check is **strict**: any `PHANTOMBOT_PERSONA` set → refuse (rather than "refuse only when target ≠ agent persona"). Rationale: the no-op "already the default" already returns before this check, so reaching it always means the agent is about to change the global default. Relaxing to "refuse only when re-pointing" would re-open the exact incident pattern.
- This is an **execution-context guard against accidental agent switches**, not a hard security boundary: a process that fully controls its own environment (`env -u PHANTOMBOT_PERSONA …`) could still switch. That matches the issue's low-severity framing; flag if a stronger boundary is wanted.
- **Concurrency:** the re-load-after-confirm removes the *lost-update during the confirm wait*, but `state.json` still has **no lock/CAS across simultaneous switches** — two concurrent switches are last-writer-wins on `default_persona` (and any other field). This is out of scope for #371 (whose bug is an agent's accidental single-writer switch). If the daemon/CLI guarantees a single writer, this is fine; if multiple processes may legitimately write `state.json` concurrently, I'd follow up with a lock/CAS around the read-modify-write.
- Left `maybePromptRestart`'s non-TTY behavior untouched (the fix gates the persist before it can matter); flag if you'd also like the @clack-hang hardening from issue option 3 as a separate follow-up.

Opening as draft per your note.


src/cli/persona.ts
```diff
@@ -311,13 +392,20 @@ export default defineCommand({
         "Skip the OpenClaw Telegram config sniff at ~/.openclaw/openclaw.json.",
       default: false,
     },
+    yes: {
+      type: "boolean",
+      description:
+        "Confirm a default-persona switch without prompting (non-interactive consent).",
```
kaieriksen, line 398: This introduces the public phantombot persona <name> --yes flag, but the repository contributor rule in AGENTS.md requires README.md and AGENTS.md updates for every new user-facing CLI flag. Please document the confirmation and scope behavior and the new option in both files so the shipped CLI and operator documentation stay aligned.

tests/cli-persona.test.ts
```diff
@@ -61,6 +61,7 @@ beforeEach(async () => {
```
kaieriksen, line 68: These tests only delete PHANTOMBOT_PERSONA in afterEach; they do not clear or restore an ambient value before each test. In the normal agent runtime this variable is set, so the pre-existing happy-path, non-TTY, and confirm=false cases are scope-refused and the focused file fails 3/13. Isolate the environment in beforeEach, preserving and restoring the prior value, so the suite passes in the runtime it is expected to validate.

kaieriksen (changes requested):
Requested changes on exact head a8c8a74.

1. Add the required README.md and AGENTS.md documentation for the new public --yes CLI flag.
2. Isolate PHANTOMBOT_PERSONA in the persona test fixture; the focused suite fails 3/13 in the normal agent runtime unless that ambient variable is removed.

Validation: focused persona tests pass 13/13 with a clean environment; typecheck passes. The full suite reaches 2,733 pass / 2 skip / 6 fail, with all six failures caused by unrelated subprocess tests invoking bare bun when only ~/.bun/bin/bun is on PATH.

kaieriksen (approved):
Re-reviewed exact head 21d1059. Focused persona suite passes 13/13; typecheck passes. Full suite passes 2,815 tests with 2 skipped; 6 failures are host-only subprocess assumptions (bare bun on PATH / local Pi credentials), unrelated to this PR.

Issue #359, opened 2026-08-11
Persistent P2P Instability and Relay Degradation


14skywalker:
## Description
We are observing a consistent pattern of P2P connection failures across multiple LXC containers (ct102/Domhnall and ct103/Mairead) running Phantombot. The connection establishes but fails shortly after, preventing long-running tasks from completing.

## Symptoms
- **P2P Cycling:** Connections cycle through `connected → failed → closed → connecting → connected` approximately every 58-60 seconds.
- **Relay Issues:** The relay count is consistently at 6/7. Logs indicate a relay dropped (e.g., at 10:55:21 UTC) which seems to destabilize the signaling layer.
- **Task Failures:** Any task or agent turn taking longer than ~1 minute fails because the transport layer drops before the response can be delivered.
- **Log Errors:**
    - ICE connection state transitions to "failed" or "disconnected".
    - "Connection dead" messages in P2P logs.
    - Frequent re-arming of the gift-wrap subscription.

## Environment
- **Host:** Proxmox (pve-manager/9.2.5)
- **Containers:** Ubuntu-based LXCs (ct102, ct103)
- **Version:** Phantombot 1.1.221

## Evidence
The issue is not isolated to a single container or persona. It appears to be a structural issue in the P2P/Relay transport layer. The 58-second window suggests a timeout or heartbeat failure in the ICE/STUN/TURN negotiation or the relay heartbeat.

robertclawson:
Thanks for the detailed report, Jeroen — the logs are really helpful.

One thing to pin down first, because it changes where we look: in phantombot, **task execution is decoupled from the P2P/relay transport.** Turns run server-side and are only aborted when you send a *new* message in the same chat — a dropped peer connection or a degraded relay can't kill a running task. Replies are also always published to relays as a guaranteed floor, with P2P/WebRTC only as a fast-path on top, so a P2P drop should downgrade to relay-only delivery, not lose the response.

That makes me suspect what you're seeing is the **reply not being delivered/backfilled**, rather than the task itself failing. To confirm, could you check:

1. **Did the task actually run?** On the node, does the agent log show the turn completing (a generated reply) for one of the "failed" tasks, even though nothing arrived in the chat? If yes, it's a delivery problem, not an execution one.
[...]

PR #384, opened 2026-08-15
feat(phantomchat): read-back-driven relay quarantine with a floor of 3 (#359)

robertclawson:
Closes part of #359 (the bot half — the PWA half is phantomchat#<pending>).

## The problem

#368 shipped read-back verification: after every publish we re-query each relay for the event id, and warn naming the relays that ACK'd `OK:true` and never stored it. **Nothing ever acted on that result.**

@14skywalker's numbers: 4 of the 7 canonical relays fail read-back on ~100% of events (~90 observed), 0% on the other three. We publish to those 4 again on the next message, and the next, forever. 4/7 of every publish is wasted, plus a warning line per send.

He proposed trimming `relays.json` from 7 to 3. That fixes his box and pins everyone at the reliability floor with no automatic replacement when one of the surviving three rots. This is the self-correcting version: each client converges on its own healthy subset from its own observations, and recovers automatically when a relay comes back.

## The design

**1. Quarantine from read-back, not from a health check.** 5 consecutive read-back failures → out of the publish set for 1h, doubling per repeat offence to a 6h cap, ±10% jitter. A single confirmation clears the streak, so a slow or hiccuping relay is never quarantined for a blip; a confirmation *while* quarantined releases it early.

**2. Quarantine is publish-only — which is what makes failover instant.** The subscription set is never narrowed. A relay that drops writes may still deliver reads, and the catch-up poll keeps using it. More importantly every quarantined relay is a **warm spare**: its socket is already open, so promoting it back into the publish set is a pure bookkeeping change. No connect, no handshake, no wait. We never reconnect a relay because we never disconnected one.

**3. No polling, no timers.** Health is a by-product of traffic we already generate — the read-back pass was already running on every publish, so scoring costs zero additional requests. Quarantine expiry is evaluated lazily against `Date.now()` at publish time. There is not a single `setTimeout` in `relayHealth.ts`. A pool that sends nothing burns zero cycles on health.

**4. A hard floor of 3.** Delivery needs only ONE healthy relay shared by sender and recipient — but sender and recipient quarantine *independently*, so phantombot's bad set and the PWA's need not agree, and their sets must still intersect. If quarantine would take the publish set below 3, the best-ranked quarantined relays are promoted back to fill it. **The floor is a hard constraint; a quarantine is only an opinion, and the floor wins.**

**5. Deterministic ranking.** `rankRelays` is a pure total order over `(score, url)` — no clock, no randomness. Both ends produce the same order from the same observations, so they converge rather than drift. The url tiebreak is load-bearing: it's why a fresh client with zero observations picks the same three relays as everyone else.

Also raises `PUBLISH_READBACK_TIMEOUT_MS` 2500 → 5000. The read-back result now drives quarantine, so a false negative has a real cost — a merely SLOW relay must not be scored as a dropping one. The check is detached, so the extra patience costs the send path nothing.

## Files

- `src/channels/phantomchat/relayHealth.ts` (new) — `RelayHealthTracker`, `rankRelays`, `relayScore`. Heavily commented on the *why*.
- `src/channels/phantomchat/transport.ts` — `publishWrap` targets `publishTargets()`; `verifyStored` feeds per-relay outcomes back into the tracker and takes an explicit target list.
- `tests/channels-phantomchat-relayHealth.test.ts` (new) — 20 tests.

## Testing

20 new tests covering: strike threshold, blip immunity, lazy expiry, early release, exponential backoff + cap, jitter spread, the floor under total poisoning, least-bad promotion, determinism of ranking, and the transport wiring (including that the subscription set is never narrowed).

Full suite: **2752 pass, 0 fail**, `tsc --noEmit` clean.

## Review notes

- Constants (`READBACK_STRIKE_THRESHOLD`, `QUARANTINE_BASE_MS`, `MIN_PUBLISH_RELAYS`) are exported and deliberately easy to tune after the weekend's data.
- Health is in-memory only — a restart gives every relay a fresh chance. That felt right for a first cut; persisting it is a follow-up if the weekend says otherwise.
- Pairs with the phantomchat PR, which reuses the PWA's existing `relayHealth`/bench machinery rather than duplicating this.

### DIFF
--- a/src/channels/phantomchat/relayHealth.ts
+++ b/src/channels/phantomchat/relayHealth.ts
@@ -1,0 +1,280 @@
+/**
+ * Relay health tracking + quarantine (issue #359).
+ *
+ * BACKGROUND. A nostr relay can answer a publish with `OK:true` and never
+ * store the event (nostr-rs-relay's ACK-then-drop on non-conformant events, or
+ * an unauthenticated publish under `nip42_auth`). Issue #368 added read-back
+ * verification, which NAMES the relays that dropped an event — but nothing
+ * acted on the result. In production 4 of the 7 canonical relays fail read-back
+ * on ~100% of events, and we kept publishing to them on every message forever:
+ * wasted connections, wasted publishes, and a warning line per send.
+ *
+ * WHAT THIS DOES. Turn that existing read-back signal into a quarantine
+ * decision. A relay that fails read-back READBACK_STRIKE_THRESHOLD times in a
+ * row is quarantined from the PUBLISH set for an exponentially growing span
+ * (base one hour, jittered). It keeps its READ subscription — see "reads are
+ * never quarantined" below — so promoting it back costs nothing.
+ *
+ * FOUR PROPERTIES THIS IS BUILT AROUND
+ *
+ * 1. NO POLLING, NO TIMERS. Health is a by-product of traffic we already
+ *    generate: every publish already runs a read-back, so scoring is free.
+ *    There is no health-check loop and not a single `setTimeout` in this file
+ *    — quarantine expiry is evaluated lazily, by comparing `Date.now()` at the
+ *    moment we next need a publish set. A pool that sends nothing burns
+ *    literally zero cycles on health.
+ *
+ * 2. READS ARE NEVER QUARANTINED — so failover is instant. Quarantine applies
+ *    ONLY to publish targets. The transport keeps its subscription open on
+ *    every configured relay, which means (a) a relay that drops writes may
+ *    still deliver reads, and the 15s catch-up poll keeps using it, and (b)
+ *    every quarantined relay is a WARM SPARE: its socket is already open, so
+ *    promoting one back into the publish set is a pure bookkeeping change with
+ *    no connect, no handshake, no wait. This is the "I hate waiting to
+ *    reconnect a relay" requirement — we never reconnect, because we never
+ *    disconnected.
+ *
+ * 3. A FLOOR OF THREE. Delivery only needs ONE healthy relay shared by sender
+ *    and recipient, but the sender and the recipient quarantine independently —
+ *    phantombot's bad-relay set and the PWA's need not agree. A floor of 3
+ *    keeps the intersection non-empty in practice. `publishTargets()` will
+ *    therefore promote the best-ranked quarantined relays back, ignoring their
+ *    remaining quarantine, rather than ever return fewer than
+ *    MIN_PUBLISH_RELAYS. A quarantine is an OPINION about relay quality; the
+ *    floor is a HARD CONSTRAINT and wins.
+ *
+ * 4. DETERMINISTIC RANKING. `rankRelays` is a pure total order over (score,
+ *    url). Given the same observations, phantombot and the PWA sort the same
+ *    way — so the two ends converge on the same subset instead of drifting
+ *    apart and losing their intersection. The url tiebreak matters: it is what
+ *    makes the order total, and it is why a fresh client with no observations
+ *    at all still picks the same three relays as everyone else.
+ */
+
+/** Consecutive read-back failures before a relay is quarantined. */
+export const READBACK_STRIKE_THRESHOLD = 5;
+
+/**
+ * First quarantine span. Doubles on each REPEAT offence (a relay that earns a
+ * quarantine again after already serving one), capped at QUARANTINE_MAX_MS, so
+ * a persistently dead relay is retried roughly twice a day instead of hourly,
+ * while a relay that had one bad hour is back in rotation quickly.
+ */
+export const QUARANTINE_BASE_MS = 60 * 60 * 1000; // 1 hour
+export const QUARANTINE_MAX_MS = 6 * 60 * 60 * 1000; // 6 hours
+
+/**
+ * ±10% jitter on every quarantine span. Without it, a fleet that all saw the
+ * same relay die at the same time un-quarantines it at the same instant and
+ * stampedes it. The jitter is applied once, when the quarantine is set.
+ */
+export const QUARANTINE_JITTER = 0.1;
+
+/**
+ * Never publish to fewer than this many relays — see property 3 above. If
+ * quarantine would take us below the floor, the best-ranked quarantined relays
+ * are promoted back to fill it.
+ */
+export const MIN_PUBLISH_RELAYS = 3;
+
+/**
+ * Per-relay health record. Deliberately small and JSON-shaped: it is cheap to
+ * keep in memory, and cheap to persist later if we ever want health to survive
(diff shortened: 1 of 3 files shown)


lenaparkhodges (dismissed):
✅ **Approved from the product/quality seat.** This is the right fix for #359 — it acts on the signal #368 already produced instead of adding new machinery, and the four design properties (no timers, publish-only quarantine, floor of 3, deterministic ranking) are each the correct call.

**What I verified in the diff:**
- Strike accounting during quarantine is subtle but right: strikes keep accruing while quarantined (only the *quarantine decision* early-returns), so a persistently dead relay re-quarantines immediately after expiry without needing a fresh 5-failure streak.
- `announced` set is cleaned on both early-release and re-quarantine, so floor-promotion logs fire once per offence, not per publish.
- `PUBLISH_READBACK_TIMEOUT_MS` 2500→5000 is justified now that false negatives have a real cost; detached check means the send path pays nothing.

**Two nits, neither blocking:**
1. The floor of 3 does not *guarantee* sender/recipient set intersection (two 3-subsets of 7 can be disjoint) — it makes disjointness unlikely given deterministic ranking. The docs already frame it as "in practice," which is honest. Just don't let anyone quote it as a guarantee later.
2. The tracker takes `relays` at construction. If relay config ever becomes hot-reloadable, the tracker goes stale. Fine today; worth a comment when that changes.

Also noting for the changelog: in-memory-only health means a restart forgives every relay — deliberate and right for a first cut, per the PR notes.

src/channels/phantomchat/relayHealth.ts
```diff
@@ -0,0 +1,280 @@
...
+ *    every quarantined relay is a WARM SPARE: its socket is already open, so
+ *    promoting one back into the publish set is a pure bookkeeping change with
+ *    no connect, no handshake, no wait. This is the "I hate waiting to
+ *    reconnect a relay" requirement — we never reconnect, because we never
+ *    disconnected.
+ *
+ * 3. A FLOOR OF THREE. Delivery only needs ONE healthy relay shared by sender
+ *    and recipient, but the sender and the recipient quarantine independently —
+ *    phantombot's bad-relay set and the PWA's need not agree. A floor of 3
+ *    keeps the intersection non-empty in practice. `publishTargets()` will
```
kaieriksen, line 40: Non-blocking documentation note: this floor makes overlap more likely but cannot guarantee intersection between two independently chosen 3-of-7 sets. The current “in practice” wording is correct; please preserve that caveat in future docs.

kaieriksen (dismissed):
Approved at exact head 712437faff7159cbddd0c47dd22a682a29949ae9. The quarantine wiring is coherent: publish-only filtering, lazy expiry, deterministic ranking, and the hard floor are all covered. Hosted CI is fully green; local tsc --noEmit passed. The floor of three improves overlap in practice but cannot mathematically guarantee sender/recipient intersection; keep the current wording honest.

robertclawson:
Both nits addressed in fc2219f — thanks, the first one especially.

**1. Floor-3 is not a guarantee.** You are right and I had written it loosely. Two 3-subsets of a 7-relay list can be disjoint, so the header no longer implies otherwise: it now says the floor alone proves nothing, and names what actually carries the weight — the floor *together with* deterministic ranking, since both ends sort by the same pure function and only diverge as far as their observations diverge. It also names the real fix if we ever need a true guarantee (a pinned shared write set), so nobody reaches for a bigger floor expecting one.

**2. Stale relay list.** Rather than leave a comment-shaped trap for the hot-reload day, the seam exists now: `setRelays()`. Health is keyed by URL, so relays that survive a config change keep their observations and their in-flight quarantines — a reload is not an amnesty — and records for relays that are gone are dropped, so a process that rotates its relay list does not accumulate them forever.

4 new tests (24 in the file). Full suite 2756 pass / 0 fail, typecheck clean.

On your changelog note: agreed, in-memory-only means a restart forgives every relay. Deliberate for the first cut — and this weekend that is a feature, since it gives Jeroen a clean slate on demand.

robertclawson:
Thanks both — all nits addressed in `fc2219f`, which dismissed the approvals, so re-requesting.

- **@kaieriksen / @lenaparkhodges — the floor is not a guarantee:** made that explicit in the header rather than leaning on "in practice". It now states outright that two 3-subsets of 7 can be disjoint, that what makes disjointness unlikely is the floor *together with* deterministic ranking, and that a real guarantee would need a pinned shared write set, not a bigger floor.
- **@lenaparkhodges — tracker takes `relays` at construction:** built the seam now instead of leaving a comment-shaped trap. `setRelays()` re-points the list; health is keyed by URL so surviving relays keep their observations *and* in-flight quarantines (a config reload is not an amnesty), and records for removed relays are dropped so a long-lived process doesn't accumulate them. Covered by new tests.

CI green on the new head (test, test-windows, typecheck, CodeQL).

kaieriksen (approved):
Approved at exact head fc2219fcf46e5a77c735fef1cd4e4bcb8c401600. Robbie addressed both prior nits: the floor documentation now states the intersection limitation precisely, and RelayHealthTracker has a setRelays() seam with URL-keyed health preservation and removal cleanup. Hosted CI is green; exact-head tsc --noEmit passed.

kaieriksen (approved):
Approved at exact head fc2219fcf46e5a77c735fef1cd4e4bcb8c401600. The new read-back tracker is correctly isolated from connection health, quarantine applies only to writes, the floor/ranking behavior is covered, and setRelays preserves surviving health. Focused relay-health tests pass 20/20; hosted typecheck, test, Windows, and CodeQL checks are green.

lenaparkhodges (approved):
✅ Re-approved at fc2219f. Both nits properly closed: the header now says precisely what the floor does and doesn't prove (and names the pinned write set as the real fix), and `setRelays()` turns my hot-reload comment into an actual seam with four good tests — the "reload is not an amnesty" and drop-then-re-add cases cover exactly the right semantics. CI all green. Good to merge.

lenaparkhodges (approved):
✅ **Re-approved at fc2219f.** Both nits are properly closed, not just acknowledged:

- **Floor claim:** the header now says explicitly that the floor alone proves nothing (two 3-subsets of 7 can be disjoint) and names what actually carries the weight — floor + deterministic ranking via the shared pure function. Naming the pinned-shared-write-set as the real fix heads off future "just raise the floor" PRs. 
- **`setRelays()`:** building the hot-reload seam now is better than a comment-shaped trap. Reload-is-not-an-amnesty (health keyed by URL) and drop-the-gone are both the right semantics, and the 4 new tests lock in exactly those properties.

Verified locally: 24/24 relayHealth tests pass; full suite 2754 pass / 2 fail, but both failures are pre-existing `harnesses-pi.test.ts` ambient-env leaks that reproduce on base 8f61ff0 (same class as #373) — not this PR. Hosted CI fully green.

PR #385, opened 2026-08-15
fix(nightly): target the day that closed, not the day just beginning

lenaparkhodges:
## Problem

Reported by Robbie: the nightly cognitive pass has been broken since ~May 20 — it targets the day just *beginning* instead of the one that *closed*.

**Root cause:** the nightly timer fires at 02:00 local (UTC on the lab installs). The default processing date in `src/cli/nightly.ts` was `new Date().toISOString().slice(0, 10)` — the date *at fire time*, i.e. the day that just began. The prompt then reads `memory/<today>.md`, which doesn't exist yet. Nearly every nightly turn log reports *"No daily file exists for [same date as the run]"* — the essence/promote/KB/state stages have been silent no-ops on schedule for months. Regression window matches PR #113 (merged May 19), when the checkpointed nightly started running reliably enough to no-op consistently.

## Fix

- **Default date = yesterday**, computed with UTC calendar arithmetic (`setUTCDate(d.getUTCDate() - 1)`) — calendar-day subtraction, not 86400 s, so month/year boundaries are exact.
- **`--date` override unchanged** — backfill and testing still work as before.
- **Prompt wording**: `Today is X` → `Processing date: X (the day that just closed)` in both the monolithic and checkpointed-stage prompts, so the nightly agent isn't misled by the word "today".
- Extracted `defaultNightlyDate()` as an exported, injectable-clock helper for testability.

## Tests

3 new regression tests in `tests/cli-nightly.test.ts`:
- 02:00 fire on Aug 15 → processes `2026-08-14`
- Month boundary: Aug 1 → Jul 31
- Year boundary: Jan 1 → Dec 31 of the prior year

`tests/lib-nightly.test.ts` updated for the prompt wording change. **33/33 pass** across both nightly suites; full suite shows only the 2 pre-existing PiHarness env failures and 20 pre-existing MCP-SDK typecheck errors (verified identical on clean `main`).

## Verification after merge

Tomorrow's 02:00 run should log `date=<yesterday>` and actually process the daily file — checkable via `nightly: persona=... date=...` in the service logs.

### DIFF
--- a/src/cli/nightly.ts
+++ b/src/cli/nightly.ts
@@ -72,6 +72,21 @@ interface TurnResult {
 }
 
 /** Run one harness turn for the nightly conversation. */
+/**
+ * Default processing date for a nightly run: yesterday, in UTC calendar
+ * arithmetic (UTC date minus one calendar day). The timer fires at 02:00
+ * local, so "today" at fire time is the day just beginning — the file to
+ * process is the one that closed at midnight. Subtracting a calendar day
+ * (not 86400s) stays correct across leap seconds / calendar boundaries.
+ */
+export function defaultNightlyDate(now: Date = new Date()): string {
+  const d = new Date(
+    Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate()),
+  );
+  d.setUTCDate(d.getUTCDate() - 1);
+  return d.toISOString().slice(0, 10);
+}
+
 async function runNightlyTurn(opts: {
   persona: string;
   conversation: string;
@@ -139,7 +154,11 @@ export async function runNightly(input: RunNightlyInput = {}): Promise<number> {
     return 2;
   }
 
-  const today = input.today ?? new Date().toISOString().slice(0, 10);
+  // The nightly timer fires at 02:00 to process the day that just CLOSED —
+  // default to yesterday, not the day that has barely begun. Without this
+  // every unattended run looks for a daily file that doesn't exist yet and
+  // silently no-ops. `input.today` (--date) stays the override for backfill.
+  const today = input.today ?? defaultNightlyDate();
   const conversation = nightlyConversationKey(today);
   const memory = await openMemoryStore(config.memoryDbPath);
   const runStartedAt = Date.now();
--- a/src/lib/nightly.ts
+++ b/src/lib/nightly.ts
@@ -252,7 +252,7 @@ This conversation is ISOLATED (conversation key system:nightly:${today}); nothin
 
 You have access to phantombot's memory tools via Bash:
 
-  phantombot memory today                       # path to today's daily file
+  phantombot memory get memory/${today}.md            # read the daily file being processed (do NOT use 'memory today' — resolves to actual current day, not ${today})
   phantombot memory search "<query>"            # FTS5 + (if configured) semantic search
   phantombot memory get <persona-relative-path> # cat a file
   phantombot memory list <persona-relative-dir> # ls a dir
@@ -313,7 +313,7 @@ This conversation is ISOLATED (conversation key system:nightly:${today}); nothin
 
 You have access to phantombot's memory tools via Bash:
 
-  phantombot memory today                       # path to today's daily file
+  phantombot memory get memory/${today}.md            # read the daily file being processed (do NOT use 'memory today' — resolves to actual current day, not ${today})
   phantombot memory search "<query>"            # FTS5 + (if configured) semantic search
   phantombot memory get <persona-relative-path> # cat a file
   phantombot memory list <persona-relative-dir> # ls a dir
--- a/src/lib/nightly.ts
+++ b/src/lib/nightly.ts
@@ -246,7 +246,7 @@ export function buildNightlyPrompt(
   personaName: string,
   today: string,
 ): string {
-  return `You are running your nightly cognitive maintenance pass for persona '${personaName}'. Today is ${today}.
+  return `You are running your nightly cognitive maintenance pass for persona '${personaName}'. Processing date: ${today} (the day that just closed).
 
 This conversation is ISOLATED (conversation key system:nightly:${today}); nothing you say here will appear in Telegram or any user-facing chat. Speak in summaries, not replies.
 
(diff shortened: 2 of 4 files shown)


robertclawson (dismissed):
## Reviewed and approved ✅

The date-arithmetic fix is correct. A few things I verified independently:

**What's right**
- `defaultNightlyDate()` uses `Date.UTC(...)` + `setUTCDate(-1)` — true calendar subtraction, not `-86400e3`. Month and year boundaries are exact.
- Ran the test suites myself: **33/33 pass**. All 6 CI checks green (incl. test-windows, CodeQL).
- `--date` override untouched, so manual backfill still works.
- **CET/CEST boundary holds.** 02:00 local → target `2026-08-15` under both UTC+2 and UTC+1. The naive `−1` survives the DST flip because the fire instant moves with the offset.
- **`--resume` / startup catch-up inherits the fix.** `pendingNightlyStages` compares against the same `today`, and doctor's resume spawns with no `--date`. No second date computation hiding anywhere.
- Clock-skew is sane: only a fire >1s *before* 00:00 UTC would target the wrong day, and `AccuracySec=1min` fires late, never early. Low risk.

**Scope note:** this stops the bleeding — it does **not** backfill the 81 daily files (May 20 → Aug 15) that were never distilled. They still exist, so `phantombot nightly --date <YYYY-MM-DD>` can recover them, but that's a separate job (and 81 LLM runs).

---

### Nit (worth fixing before merge)

The prompts now correctly say `Processing date: X (the day that just closed)`, but the toolbox block still advertises:

```
phantombot memory today  # path to today's daily file
```

That command resolves to the *actual* current day (the one just beginning), not the date being processed. The stage bodies do interpolate the explicit `memory/${today}.md` path, so it's mitigated — but an agent that reaches for `memory today` during nightly gets handed the wrong file. One-line tweak to that comment so nightly-me isn't given a footgun.

---

Nice work on this. The root cause (UTC date at fire time = the day just *beginning*) was subtle and the fix is clean.

robertclawson (approved):
## Re-review — nit fixed, approving ✅

Verified `31fd1ad` on top of `b421ce2`.

**The nit is properly resolved.** The toolbox block no longer advertises `phantombot memory today`; both preamble functions now emit the explicit `memory get memory/${today}.md` with an inline warning. I rendered all six prompt variants (`buildNightlyPrompt` + all 5 stages via `NIGHTLY_STAGES`) and confirmed each one interpolates the processing date and no longer advertises the stale command:

```
buildNightlyPrompt   | phantombot memory get memory/2026-08-15.md   | stale: false
stage:essence        | ...2026-08-15.md                             | stale: false
stage:promote        | ...2026-08-15.md                             | stale: false
stage:kb             | ...2026-08-15.md                             | stale: false
stage:compress       | ...2026-08-15.md                             | stale: false
stage:state          | ...2026-08-15.md                             | stale: false
```

**Scope is right.** Only the two nightly preambles changed. `memory today` remains in `src/cli/memory.ts` (the command's own docstring) and `src/persona/builder.ts` (the general agent toolbox) — both correct, since there "today" genuinely means today. Good restraint not blanket-replacing it.

**Nothing regressed.** `defaultNightlyDate()` is untouched — still `Date.UTC(...)` + `setUTCDate(-1)`. Re-ran my own boundary checks against the new head:

| fire instant | target |
|---|---|
| CEST 02:00 local (00:00Z) | 2026-08-15 ✅ |
| CET 02:00 local (01:00Z) | 2026-01-15 ✅ |
| UTC 02:00 | 2026-08-15 ✅ |
| month boundary (Sep 1 00:00Z) | 2026-08-31 ✅ |
| year boundary (Jan 1 00:00Z) | 2026-12-31 ✅ |

**Tests:** ran locally — **33/33 pass** across `tests/cli-nightly.test.ts` + `tests/lib-nightly.test.ts`. The one test asserting on the old string was correctly updated to assert the new interpolated path rather than deleted, so the toolbox line stays covered. All 6 CI checks green (incl. `test-windows`, CodeQL).

**Unchanged scope note (not a blocker):** this stops the bleeding, it does not backfill. The ~81 daily files in the 2026-05-20 → 2026-08-15 gap still exist and remain undistilled; `phantombot nightly --date <YYYY-MM-DD>` can recover them as a separate job.

No further comments — good to merge from my side.

robertclawson (approved):
## Re-approved ✅ — nit addressed

My earlier approval was dismissed by `31fd1ad`. That commit is exactly the fix I asked for, and I re-verified the whole PR against it.

**The nit is closed.** Both `buildNightlyPrompt` and `nightlyStagePreamble` now advertise the interpolated `memory get memory/${today}.md` with an inline warning not to reach for `memory today`. Confirmed the command actually works — `phantombot memory get memory/2026-08-14.md` returns the file, exit 0. No footgun left for nightly-me.

**Re-verified independently on the head commit:**
- `defaultNightlyDate()` — ran the implementation standalone against 8 instants. Correct in every case: CEST 02:00 (`00:00:05Z` → `2026-08-14`), CET 02:00 (`01:00:05Z` → prior day), **both DST flips** (spring-forward 03-29, fall-back 10-25), leap year (`2028-03-01` → `2028-02-29`), month and year boundaries. The naive `−1` is safe because the fire instant moves with the UTC offset, so the local-02:00 fire never crosses back over UTC midnight under either offset.
- **No second date computation.** Grepped every `toISOString().slice(0, 10)` in `src/`; the remaining ones are unrelated (`task`, `memory`, `heartbeat`, `auditLog`, `personaScaffold`). `pendingNightlyStages` receives the same `today`, so `--resume` and doctor's catch-up inherit the fix.
- **Tests: 33/33 pass** locally across both nightly suites. All 6 CI checks green (incl. test-windows, CodeQL).
- Diff is tight — 50/8 across 4 files, nothing out of scope. `--date` override untouched.

**Scope reminder (unchanged):** this stops the bleeding but does not backfill the ~81 daily files from May 20 → Aug 15. The files still exist, so `phantombot nightly --date <YYYY-MM-DD>` can recover them — separate job, Andrew's call on whether it's worth ~81 LLM runs.

Clean fix on a genuinely subtle root cause. Leaving the merge to Andrew.

PR #386, opened 2026-08-16
feat(nightly): idempotent sweep, two concurrent stages, event-driven trigger

robertclawson:
## What

Collapses the nightly cognitive pass from **five sequential LLM stages to two, run in parallel**, and makes the whole pass idempotent against a ledger.

| | before | after |
|---|---|---|
| LLM turns / day | 5 sequential | 2 concurrent |
| recovery paths | `--resume`, doctor spawn, startup catch-up | none — every run sweeps what's pending |
| index refresh | a line in the KB *prompt* (`--rebuild`, every file re-embedded) | in code after the join (`refreshStale`) |
| daily files | mutated (essence header prepended) | read-only inputs |

## Why each stage went

- **essence** — every turn is embedded now; a 3-line day summary adds nothing search doesn't, and it *mutated the file the ledger hashes*, which would make every processed date look changed forever.
- **promote** — merged into `distill`; the 30-min heartbeat already does mechanical promotion, the nightly only mops up.
- **compress** — merged into `distill` (same output targets: drawers + MEMORY.md).
- **state** — was never cognitive work; the driver already writes the state file in code.
- **kb** — kept as its own turn. It's the only stage doing search → read → reconcile, and it reliably goes shallow when bundled with mechanical filing.

`distill` and `kb` write disjoint targets (drawers + MEMORY.md vs `kb/`), so they run concurrently with no lock. The one thing that *did* need ordering — the index refresh — now runs in code once both join.

## The sweep

`.nightly-state.json` gains a ledger: `date → {mtime, size, hash, stages_done, status}`. Every run reconciles the daily files against it and processes anything new, grown, or half-finished.

- No `--resume`, no `--catch-up`, no doctor repair spawn. A half-done date is just a date the ledger doesn't call done, so resume is automatic.
- Cheap: a date whose mtime+size match is skipped without reading the file. Only files that moved get hashed; a hash that matches means "touched, not changed" and costs no turn.
- A day that gains captures *after* its pass gets reprocessed — something a "rename to `_processed_`" scheme structurally can't see (and which would also break `memory capture`, whose writes are keyed on `memory/<date>.md`).
- Triggered by events, not a clock (see below), and both triggers are safe to fire redundantly.

Daily files are untouched: no renames, no prefixes, no deletion.

## No more 02:00 timer

The nightly unit is **deleted from all three backends** (systemd, launchd, Windows Task Scheduler). A clock is the wrong trigger for a pass whose job is "distil the day that just closed" — a box asleep at 02:00 misses it entirely. Two events that already happen replace it:

1. **startup** — `phantombot run` fires a detached sweep (laptops, anything that was powered off).
2. **day rollover** — the 30-min heartbeat reads its own last-fired marker before overwriting it; when the calendar day has changed since the previous fire, yesterday's daily file has closed, so it spawns a sweep. Worst-case latency is one heartbeat interval.

Rollover *detection*, not a file-creation watch: a daily file is created lazily on the first capture of the day, so on a quiet day it may never exist at all and a creation hook would silently starve. The day changing is unconditional. `dailyFileDate` deliberately uses the same (UTC) basis as the daily-file writer, so the trigger can't fire while the day it is about to process is still being appended to.

Upgrade path: the retired unit names survive only as a cleanup list. The heal path (every heartbeat) stops, disables and deletes `phantombot-nightly.timer` / the launchd agent / the scheduled task if an older install left one armed — guarded on the unit being present, so a current install costs zero extra IPC. `doctor` no longer expects or repairs it.

## `/status`

```
dreaming: OK (nothing pending)
dreaming: RUNNING (2/5 dates, on 2026-06-02)
dreaming: WARN (2 dates pending, oldest 2026-06-02)
dreaming: ERR (1 date pending, no sweep since 2026-05-28T02:00:00Z)
```

Backlog is the only truth — **schedule is irrelevant**. A laptop that swept on boot at 09:15 reports OK, as long as nothing is pending. A crashed sweep leaves an in-flight marker that goes stale after 45 min and reports ERR rather than RUNNING forever. `doctor` reports the same signal, read-only.

**Depth is not a fault.** A backlog of any size is queued work, and one sweep takes the whole queue, so it reads WARN while it drains — a backfill shouldn't make `/status` scream for a week. ERR is reserved for a backlog that is *stagnant*: dates pending and no sweep in over 24h, meaning the trigger is dead rather than merely behind. A never-swept ledger stays WARN, since that's a fresh install and `run` stamps `last_run` within a minute of coming up.

## Bounds and safety

- **No per-run cap.** A sweep drains the entire backlog in one pass — a first run over months of history is one long night rather than a queue that reappears every morning. `--max-dates N` bounds a manual run; whatever it leaves stays pending in the ledger.
- In-flight marker prevents two sweeps double-filing the same drawers; a stale one is taken over, not obeyed (`--force` to override).
- `--date <YYYY-MM-DD>` still reprocesses one day on demand.

## Tests

`bun test` green (2809 pass / 0 fail), `tsc --noEmit` clean. New coverage: sweep semantics (new / changed / touched / incomplete / today-excluded), ledger merge, health state machine including the stalled-marker case, deep-backlog-is-WARN, stagnant-backlog-is-ERR and never-swept-is-WARN, stage concurrency, index-refresh ordering and its failure path, uncapped drain plus `--max-dates` bounding, the in-flight lock, prompt invariants (no stage writes the daily file, no stage runs the index), and the rollover trigger — fires across a day boundary, stays quiet within a day, stays quiet with no marker (startup already swept) or a backwards clock step, and a spawn failure never fails the heartbeat. Backend suites assert the retired unit is never generated and that a leftover one is torn down.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -183,14 +183,19 @@ If you add a new untrusted entry point (a new inbound channel, a new `ask`-style
 
 Phantombot ships on **Linux (systemd --user)**, **macOS (launchd, per-user LaunchAgents)**, and **Windows (per-user Task Scheduler logon task plus three periodic tasks)**. `lib/platform.ts` is the single router that decides which backend to talk to; each backend sits behind a common `ServiceControl` surface with an injectable runner so tests never touch the real service manager. Windows uses `InteractiveToken` and the current user's SID, so installation needs no password or elevation; installation preserves healthy existing task definitions and only repairs missing/stale entries. Windows `start`/`restart` re-enable the task and use a hidden native detached launch for SSH/terminal control because Task Scheduler cannot `/Run` an interactive-token task from a non-interactive session. `windowsJob.ts` assigns harness children to kill-on-close Job Objects.
 
-On Linux, `phantombot install` creates **four** systemd-user units:
+On Linux, `phantombot install` creates **three** systemd-user units:
 
 | Unit | Cadence | What it does |
 |---|---|---|
 | `phantombot.service` | always-on | `phantombot run` — Telegram listener |
 | `phantombot-heartbeat.timer` → `.service` | every 30 min | mechanical maintenance, no LLM |
-| `phantombot-nightly.timer` → `.service` | daily 02:00 | cognitive distillation, LLM |
 | `phantombot-tick.timer` → `.service` | every 1 min | fires due scheduled tasks |
+
+There is deliberately no nightly unit. The cognitive pass is event-driven —
+`run` fires a sweep at startup, and the heartbeat fires one when it notices the
+calendar day changed since its last fire. A clock would miss the day entirely on
+any box that sleeps at 02:00. Installs that predate this have their
+`phantombot-nightly.timer` stopped, disabled and deleted on the next heartbeat.
 
 Every service has **two `EnvironmentFile=` lines**:
 
--- a/README.md
+++ b/README.md
@@ -1512,17 +1512,22 @@ the ledger's hash stable. Once both join, phantombot refreshes the search index
 itself (incremental FTS + embeddings), so a new KB note is searchable without
 the model having to remember to ask.
 
-Because the ledger decides what to process, the pass is idempotent: re-running
-it with nothing pending costs nothing, and a machine that was asleep at 02:00
-just sweeps a longer backlog on its next run (the `run` daemon fires one at
-startup). There is no `--resume` and no catch-up mode. Useful flags:
+The sweep has **no timer**. It is triggered by two events instead: `run` fires
+one at startup, and the heartbeat fires one the first time it runs on a new
+calendar day (the previous day's file has closed by then). Rollover *detection*
+rather than a file-creation watch, because a daily file is created lazily on the
+first capture — on a quiet day it may never exist, and a creation hook would
+starve. Because the ledger decides what to process, both triggers are safe to
+fire redundantly: re-running with nothing pending costs nothing, and a machine
+that was off for a week sweeps the backlog when it comes back. There is no
+`--resume` and no catch-up mode. Useful flags:
 `--date <YYYY-MM-DD>` to reprocess one day, `--max-dates N` to change the
 per-run cap (default 10), `--force` to take over a stuck in-flight marker.
 
 `/status` shows a `dreaming:` line — `OK (nothing pending)`,
 `RUNNING (2/5 dates, on 2026-06-02)`, `WARN (2 dates pending …)` or `ERR`.
-Health is backlog-driven, not schedule-driven: a missed 02:00 with nothing
-pending is still OK. `phantombot doctor` reports the same signal (plus capture
+Health is backlog-driven, not schedule-driven: nothing pending is OK no matter
+when the last sweep ran. `phantombot doctor` reports the same signal (plus capture
 health, timers and connectors) but never runs the nightly itself.
 
 ## Maintenance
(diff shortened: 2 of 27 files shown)


lenaparkhodges (commented):
Reviewed in full — code, tests, and all three backend teardown paths. Architecture is strong: the mtime+size fast path with hash fallback, disjoint stage targets, in-code index refresh after the join, and event-driven triggers all hang together. CI green, 2806 tests. Three findings:

**1. Backlog policy vs. the "no token burn" directive (needs a decision).** Every unprocessed historical date costs 2 LLM turns, drained at ≤10 dates/sweep. Our ~81-day backlog = ~8 sweeps ≈ 160 turns; a "world" user upgrading after months of no-op nightlies could be much worse. There is no "seed ledger without processing" path — nothing marks old dates as processed without an LLM pass. Options: (a) one-time ledger seed on upgrade for dates older than N days (FTS-indexed, never distilled), (b) accept the capped drain as-is, (c) ship a `--seed` flag the operator runs once. Default (a) honors the directive silently.

**2. `dreaming: ERR` will persist for days-to-weeks after upgrade.** `nightlyHealth` returns error when backlog > 3, so any install with a real backlog shows ERR on /status and in doctor until the drain finishes. That's correct by the letter (backlog is truth), but it's noise that trains users to ignore the line. Consider capping at WARN for a draining, in-progress backlog and reserving ERR for stagnant ones (no successful sweep in N days despite a backlog).

**3. Nit — stale docstring.** `src/cli/doctor.ts` header still says "one owner (the timer, plus the startup sweep in run)". The timer is retired; it's the heartbeat rollover trigger now.

Non-blocking observation: both stages write to the same `system:nightly:<date>` conversation concurrently — persistence is busy-timeout-safe and tests cover the overlap, but the recorded history for a date will interleave the two stages' rows. Fine as designed; just don't let anything later reconstruct that conversation expecting linear order.

Holding approval on finding 1 — that's a policy call, not a code problem. Everything else is rubber-stampable once 3 is fixed.

lenaparkhodges (dismissed):
Solid work. The ledger-driven sweep eliminates the date-targeting bug class entirely — no default path, no guessing. Idempotency, concurrency, stale-run takeover, and upgrade cleanup are all well tested. Approving.

robertclawson:
Pushed `4f9c97f` — this addresses all three findings. Andrew's call on finding 1: **distil everything, skip nothing.**

**1. Backlog policy.** No seeding, no "indexed but never distilled" shortcut — every unprocessed daily file gets its two turns. What changed instead is the *rationing*: `MAX_DATES_PER_RUN = 10` is gone, and a sweep now drains the entire backlog in one pass. The reasoning is that a cap doesn't save the tokens, it just spreads the same ~160 turns over eight nights while `/status` stays yellow the whole time. One long night, then done. `--max-dates N` survives for bounding a manual run.

**2. `dreaming: ERR` for weeks.** Fixed, and I went further than "raise the threshold" — depth no longer factors into health at all. Any backlog is WARN while it drains. ERR now means *stagnant*: dates pending and no sweep in over 24h, which is the actual fault (the trigger is dead), as opposed to merely behind. Edge case worth flagging: a fresh install has a backlog and no `last_run`, so I explicitly treat never-swept as WARN rather than ERR — `run` stamps `last_run` within a minute of starting, so a genuinely stuck one shows up on the next check anyway.

**3. Stale docstring.** Done — the "one owner (the timer…)" header in `doctor.ts` now describes the rollover + startup triggers.

Sorry about the dismissed approval; the docstring alone would have cost a round regardless. 2809 pass / 0 fail, typecheck clean.

lenaparkhodges (approved):
Re-reviewed after 4f9c97f. The uncapped drain matches the agreed direction, and both of my earlier nits are resolved: the stale timer docstring in doctor.ts is gone, and health now reads WARN-while-draining instead of a weeks-long ERR — with the fault case correctly redefined as a STAGNANT backlog (24h without a sweep). Fresh-install (no last_run) correctly stays WARN. Tests cover all four health states plus the no-cap drain and the manual --max-dates bound; 99/99 pass locally and CI is green. LGTM.

PR #388, opened 2026-08-16
fix(nightly): scope harness turns to the persona dir, ending the macOS TCC re-prompt loop

### LINKED ISSUE #387: macOS: relentless "phantombot would like to access data from other apps" prompt during the nightly sweep
## Symptom

On macOS, phantombot triggers a relentless system prompt: **"phantombot would like to access data from other apps."** It cycles continuously during the nightly sweep. Reported by Andrew, 2026-08-16, on Matt's box.

## What it is not

- **Not code-signing churn.** Stable signing shipped in #361/#363/#364 — `Identifier=dev.phantombot`, valid, satisfies DR. This is not the pre-#361 identity-rotation problem.
- **Not the child model CLI.** macOS attributes the syscall to the parent, and the TCC client is `/Users/andrewhodges/.local/bin/phantombot` itself.
- **Not harness-specific.** Flipping the box from claude to pi does not help. The trigger is in phantombot's own spawn wiring, not in any one CLI.

## Root cause

The harness **working directory**.

`runTurn`'s `workingDir` was optional and fell back to `homedir()`. Every caller except the ACP bridge omitted it, so every background turn inherited `$HOME`. A nightly stage therefore woke up in `$HOME` with its own `memory/` and `kb/` invisible from cwd — and went looking for them.

Evidence from the affected box:

- `lsof` on the live sweep: nightly (pid 3278) and its harness child both had cwd `/Users/andrewhodges`.
- Tool-call audit log for one sweep: **79 `find` invocations, 7 of them rooted at `/`**, ~12 at `~`. Example: `find /Users/andrewhodges -maxdepth 5 -name "2026-06-29.md" -path "*memory*"`.
- The stage even wrote `cd ~/phantombot/matt 2>/dev/null || cd .` — compensating by hand for the wrong cwd.
- `fs_usage` during a live sweep: ~7 parallel claude Glob workers issuing `openat(dirfd, "Application Support")` inside a single 10ms burst, with a user `TCC.db` write landing at the same instant.
- TCC db (read as root): `kTCCServiceSystemPolicyAppData` = `auth_value 5` (undetermined, still prompting). Every other service for this client is `2` (allowed).

Those walks cross `~/Library/Containers` and `~/Group Containers`, which is exactly what `kTCCServiceSystemPolicyAppData` guards. The sweep spawns a fresh process per date, so a 110-date backlog turns one prompt into a barrage. Linux has no TCC, which is why only macOS ever saw it.

## Note on the obvious fix that would not have worked

Whitelisting the nightly's **Bash** tools would not have fixed this on its own. Part of the walk happened inside claude's **native Glob/Grep engine**, not in a shell. Correct cwd is the actual fix.

## Scope

1. **Make `workingDir` explicit at every `runTurn` call site** (primary). Nightly scopes to the persona dir. Interactive surfaces keep `homedir()` — the owner asks for work on repos all over their home dir, so scoping those down would be a regression.
2. **Give nightly stages a positive tool grant** so the tree-walking search tools aren't on the menu at all. Defence-in-depth on top of (1).

## Verification

A/B on the real macOS box, identical scratch persona and stub harness, observing the cwd the OS actually hands the spawned subprocess.

andrewagrahamhodges:
Closes #387.

On macOS, phantombot triggered a relentless **"phantombot would like to access data from other apps"** prompt that cycled continuously during the nightly sweep. Code signing (#361/#363/#364) was already stable, so this was not identity churn, and flipping the box from claude to pi did not help.

## Root cause: the harness working directory

`runTurn`'s `workingDir` was optional and fell back to `homedir()` — and **every caller except the ACP bridge took the default**. A nightly stage therefore woke up in `$HOME` with its own `memory/` and `kb/` invisible from cwd, and went hunting for them.

Confirmed on the affected box:

| Evidence | Finding |
|---|---|
| `lsof` on the live sweep | nightly (pid 3278) and its harness child both had cwd `/Users/andrewhodges` |
| Tool-call audit log, one sweep | **79 `find` invocations, 7 rooted at `/`**, ~12 at `~` |
| Stage output | wrote `cd ~/phantombot/matt \|\| cd .` — compensating by hand for the wrong cwd |
| `fs_usage`, live | ~7 parallel claude Glob workers issuing `openat(..., "Application Support")` in a 10ms burst, with a user `TCC.db` write at the same instant |
| TCC db (as root) | `kTCCServiceSystemPolicyAppData` = `auth_value 5` (undetermined); every other service for this client = `2` (allowed) |

Those walks cross `~/Library/Containers`, which is precisely what `kTCCServiceSystemPolicyAppData` guards. The sweep spawns a fresh process per date, so a 110-date backlog turned one prompt into a barrage. Linux has no TCC, which is why only macOS ever saw it.

## Changes

**1. `workingDir` is now REQUIRED on `TurnInput`, with no default.**

A default nobody passes is a default nobody audits. Every call site now declares which tree the turn may treat as home:

- **Interactive surfaces** (telegram, phantomchat, ask, tick, reactions) pass `homedir()` **explicitly**. This is deliberate and unchanged in behaviour — the owner asks for work on repos all over their home dir, so scoping these down would be a real regression (and for gemini, cwd is a hard sandbox root).
- **Nightly** passes the persona dir.
- The ACP bridge already resolved its own workspace cwd and is untouched.

**2. Nightly stages get a positive tool grant** — `toolsMode: { allow: ["Bash","Read","Write","Edit"] }`, mapping to claude's native `--tools "Bash,Read,Write,Edit"`.

This drops `Glob`/`Grep`, whose parallel workers recursively walk the tree from cwd. The stage still searches — via `phantombot memory search` (FTS + semantic), which queries an index instead of stat-ing the filesystem. `Bash` stays because the memory CLI needs it.

**3. An empty allowlist falls through to the normal surface** rather than emitting `--tools ""`, which would silently run a tool-less nightly that looks like a hang.

## Worth flagging for review

- **Restricting Bash alone would NOT have fixed this.** Part of the walk was inside claude's own search engine, not a shell. Correct cwd is the fix; the allowlist is the belt. I would not want the allowlist mistaken for the mitigation.
- **The allowlist is claude-only.** pi and codex have no positive-grant flag and ignore it. It is documented as defence-in-depth, explicitly **not** a trust boundary — `toolsMode: "none"` remains the real boundary and stays reserved for the threat judge.
- **Blast radius of making `workingDir` required** is compile-time only: every existing caller keeps its exact previous behaviour. The change is that omitting it is now a type error rather than a silent `$HOME`.

## Verification — A/B on the real macOS runtime

Identical scratch persona and stub harness binary on Matt, recording the cwd the OS actually hands the spawned subprocess:

```
main (02e753a):        /Users/andrewhodges
fix/nightly-working-dir: /private/tmp/pb387-test/personas/testp
```

Both nightly stages spawned in the persona dir on the fixed branch. (Verified against a stub harness rather than a real one, so the observation is of the spawn itself and costs no tokens; the stage then "fails" on a missing completion signal, which is expected and irrelevant to what is being measured.)

## Tests

5 new, all green:

- `runTurn` passes `workingDir` verbatim — no `homedir()` fallback
- `toolsMode` allowlist reaches the harness; omitted stays `undefined`
- claude emits `--tools` with exactly the named tools, and **not** `Glob`/`Grep`
- an empty allowlist is ignored rather than silently disabling every tool
- `runNightlyTurn` spawns in the persona dir, never `homedir()`, and grants no tree-walking search tools

The nightly wiring tests sit on `runNightlyTurn` directly because the `runStage` seam the sweep tests use bypasses `runTurn` entirely and can never observe either property.

Full suite: 2811 pass, 3 fail — the same 3 environment-dependent failures present on `main` before this branch (verified by stashing).

## Docs

`README.md` and `AGENTS.md` both updated: the required-`workingDir` rule and what to pass when adding a new `runTurn` caller, plus the nightly's scoping contract and the explicit caveat that the tool grant is claude-only defence-in-depth.

## Unrelated, handled

The wedged `phantombot notify` (pid 2087, 71 min on nostr relay ACKs) was killed on the box. That is the known notify blocking issue, not part of this fix.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -151,6 +151,7 @@ phantombot/
 - **Env bootstrap** (`lib/envBootstrap.ts`) self-sources the `.env` files at startup (required on macOS, no-op on Linux) and re-sources before each harness spawn so `phantombot env set` takes effect mid-session.
 - **Tick fires scheduled tasks** (`src/cli/tick.ts`); 1-minute systemd timer; lockfile prevents overlap; missed runs are skipped, not piled up.
 - **Heartbeat is mechanical** (no LLM); **nightly is cognitive** (LLM-driven distillation).
+- **`workingDir` is REQUIRED on every `runTurn` call** (#387) — the harness subprocess cwd, deliberately with no default. Interactive surfaces (telegram, phantomchat, ask, tick, reactions) pass `homedir()`, because the owner asks for work on repos all over their home dir. Machine-driven background turns (nightly) pass the **persona dir**. This used to silently default to `homedir()` and every background caller took the default, so nightly stages woke up in `$HOME` unable to see their own `memory/` from cwd — and went hunting for it. On macOS those recursive walks cross `~/Library/Containers`, tripping the TCC `kTCCServiceSystemPolicyAppData` prompt ("phantombot would like to access data from other apps") once per spawned date. If you add a `runTurn` caller, decide which tree it may treat as home and say so explicitly.
 
 ## Channel layer (core + adapters)
 
@@ -196,6 +197,16 @@ There is deliberately no nightly unit. The cognitive pass is event-driven —
 calendar day changed since its last fire. A clock would miss the day entirely on
 any box that sleeps at 02:00. Installs that predate this have their
 `phantombot-nightly.timer` stopped, disabled and deleted on the next heartbeat.
+
+A nightly stage is deliberately the **narrowest turn phantombot runs**: no MCP
+(`mcpMode: "none"`), cwd pinned to the persona dir, and a positive tool grant of
+`NIGHTLY_TOOLS` = `Bash,Read,Write,Edit` (`toolsMode: { allow }`, #387). The
+allowlist exists to drop claude's native `Glob`/`Grep`, whose parallel workers
+recursively walk the tree from cwd; a stage searches through
+`phantombot memory search` (an index query), never a filesystem walk. Note the
+grant is claude-only — pi and codex have no positive-grant flag and ignore it —
+so treat it as defence-in-depth, **not** a trust boundary. `toolsMode: "none"`
+is the real boundary, and it's reserved for the threat judge.
 
 Every service has **two `EnvironmentFile=` lines**:
 
--- a/README.md
+++ b/README.md
@@ -1524,6 +1524,17 @@ that was off for a week sweeps the backlog when it comes back. There is no
 `--date <YYYY-MM-DD>` to reprocess one day, `--max-dates N` to bound a manual
 run, `--force` to take over a stuck in-flight marker.
 
+A stage runs **scoped to the persona directory**. Its working directory is the
+persona dir (not your home dir), it runs with no MCP servers, and it is granted
+exactly four tools — `Bash`, `Read`, `Write`, `Edit`. That is the whole job: read
+the day's file, write `memory/` and `kb/`, and search through
+`phantombot memory search`. It never needs to walk your filesystem, and it is no
+longer able to. Before this, stages ran from `$HOME` and would go looking for
+their own `memory/` directory; on macOS that search crossed
+`~/Library/Containers` and made the system ask *"phantombot would like to access
+data from other apps"* over and over — once for every date in the backlog. If
+you see that prompt, you are on a build older than #387.
+
 A sweep is **uncapped by default**: it drains the entire backlog in one pass,
 so a first run after months of history is one long night rather than a queue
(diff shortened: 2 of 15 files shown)


kaieriksen (approved):
Reviewed exact head e06981f58a5e0718ab9fc2a0709f56c5ca35a6dd. No defects found. Focused changed-area tests: 117/117; typecheck: pass; full suite: 2810 pass, 2 skipped, 4 failures caused by tests spawning bare `bun`, which is unavailable on this host. Hosted test, typecheck, Windows, and CodeQL checks are green.

PR #389, opened 2026-08-16
feat(okf): declare a controlled KB vocabulary, and stop calling kb/ an Obsidian vault

robertclawson:
## Why

Two problems that turn out to be the same problem.

**1. The controlled-vocabulary layer is built and empty.** `src/lib/okf.ts` already parses `type`, `title`, `description`, `tags` and `aliases`, and the index already weights them as separate BM25F columns. The README already sells this as the headline of the no-API-key path:

> **Tag / alias controlled vocabulary** — author-time synonyms collapse the vocabulary-mismatch gap

Nothing populates it. No template ships `description` or `aliases`, the persona prompt never mentions them, and the nightly KB stage required only `type, tags, created, updated`. So the documented feature of the default search path is unused on every phantom in the wild.

**2. `type` has drifted, because nothing ever declared what it should be.** Off 4 scaffolded templates, a real KB here has accumulated 14 distinct values — `lesson` (94), `concept` (92), `decision` (45), then a long tail including `troubleshooting` ×2 and `atomic-note` ×1. A human-curated OKF vault checked alongside it had 21 values against 5 declared. Near-synonyms fragment the exact column meant to sharpen recall.

This matters most **without embeddings**, which is the default and — since it needs no second API key — likely the majority of installs. There, BM25 can only match words a note literally contains, so `description`/`aliases` and a clean `type` aren't polish; they're the retrieval mechanism.

## What

**Declare the vocabulary once, generate the prompts from it.** A prompt carrying its own hardcoded copy of the list would diverge from the parser the first time either changed, so `builder.ts` and `nightly.ts` interpolate `OKF_TYPES`, and tests assert every value reaches both prompts.

- **`OKF_CORE_TYPES`** (9) — `concept`, `runbook`, `procedure`, `reference`, `postmortem`, `project`, `person`, `infrastructure`, `index`. Chosen to mean the same thing in a human-curated OKF vault as in a phantom KB, so notes move between the two without translation.
- **`OKF_AGENT_TYPES`** (4) — `lesson`, `decision`, `norm`, `account`. Drawer-derived, no vault equivalent: a vault records what someone decided was true, these record what the agent learned the hard way.

**Adoption is not a migration.** `OKF_TYPE_ALIASES` + `normaliseOkfType()` fold legacy values (`troubleshooting` → `runbook`, `home` → `index`) at query time. Every note ever written stays valid, nothing on disk is renamed, only new notes converge. An unknown type returns `""` rather than being guessed into `concept` — drift should stay visible rather than being silently absorbed.

**Two no-embeddings recall fixes, in the nightly KB stage:**

- **Multi-angle dedup search.** It deduped with a single search phrase. Today's wording and the existing note's wording are routinely disjoint (`"CI agents refusing jobs"` vs `"runner version deprecated"`), so BM25 missed the note and created a duplicate. This is the failure mode that *compounds* — each duplicate splits one truth in two and dilutes every later query. Now searches by symptom, by component, and by literal error string.
- **Linking is now a requirement, with its reason stated.** Without embeddings the link graph *is* the semantic index: recall expands outward from a lexical hit along `[[wikilinks]]`. An unlinked note is reachable only by exact wording. The prompt previously said "link related notes" as a stylistic aside.

**Naming.** `kb/` is no longer described as an *"Obsidian-shaped second brain"*. That phrase came from the prompt in `builder.ts:200`, so every phantom has been told it — and it invites a real error: treating the agent's private recall as the operator's shareable vault, and writing to the wrong one. `kb/` is private persona recall. OKF is the format; where a human-facing vault lives is a separate question. (`README.md:164` still says "Obsidian vault" — correctly, about a *user's* actual vault. Left alone.)

## Drive-bys

- `heartbeat.ts` docstring claimed it "does NOT run the embedding pass (that's the nightly cycle's job)". `cli/heartbeat.ts` has run an incremental pass since it landed.
- `kb/templates/atomic-note.md` declared `type: concept` — the scaffold contradicting itself about its own vocabulary. Renamed `concept.md`.

## Testing

`2839 pass / 0 fail`, `tsc --noEmit` clean. 22 new assertions covering: alias-map integrity (no alias resolves to a removed type, none shadows a canonical one), case/separator normalisation, unknown-type behaviour, both prompts containing the full generated vocabulary, every scaffolded template declaring a canonical type plus the weighted fields, and a regression guard that the prompt never says "Obsidian" again.

Two existing tests were updated rather than worked around, both genuinely stale under the new vocabulary: `Home.md` now declares `type: index` (folded from `home`), and the scaffold expects `concept.md`.

## Deliberately not here

A `doctor` check reporting off-vocabulary types and orphan notes (zero inbound links — effectively unreachable on a BM25-only install). It should **report**, never rewrite. Left for a follow-up so this PR stays "declare the standard".

### DIFF
--- a/README.md
+++ b/README.md
@@ -1421,6 +1421,51 @@ path is much stronger than plain keyword matching:
 
 This is the default. Every phantom gets it for free, no setup.
 
+#### The frontmatter that makes it work
+
+Those three superpowers are only as good as the frontmatter agents actually
+write, so the vocabulary is part of the spec rather than a style preference.
+Every `kb/` note carries:
+
+| Field | Why the index cares |
+|---|---|
+| `type` | Controlled vocabulary (below) — its own BM25F column |
+| `title` | Highest-weighted field |
+| `description` | One sentence: the question the note answers |
+| `tags` | Lowercase, hyphenated; weighted above body text |
+| `aliases` | **The synonym fix** — the other names this thing goes by |
+| `created` / `updated` | Recency signals; `updated` is bumped on reconcile |
+
+`aliases` is the field that earns its keep on the no-key path. BM25 can only
+match words that are actually present, so a note titled "Secret Rotation"
+is invisible to a search for "credential cycling" *unless the note says so
+itself*. Aliases are how a note declares the wrong-but-plausible names a
+future query might use.
+
+**Controlled `type` vocabulary.** Left unstated, this field drifts into a
+dozen near-synonyms (`note` / `atomic-note` / `concept`) that fragment the
+very column meant to sharpen recall. Phantombot declares it once, in
+`src/lib/okf.ts`, and generates both the persona prompt and the nightly
+prompt from that constant:
+
+- **Core** — `concept`, `runbook`, `procedure`, `reference`, `postmortem`,
+  `project`, `person`, `infrastructure`, `index`. These mean the same thing
+  in a human-curated OKF vault as they do in a phantom's KB, so notes move
+  between the two without translation.
+- **Agent-side** — `lesson`, `decision`, `norm`, `account`. Derived from the
+  structured drawers; they have no equivalent in a curated vault, because a
+  vault records what someone decided was true and these record what the agent
+  learned the hard way.
+
+Adopting the vocabulary is **not a migration**. An alias map folds legacy
+values (`troubleshooting` → `runbook`, `home` → `index`) onto canonical ones
+at query time, so every note ever written stays valid and only newly-authored
+notes converge. Nothing on disk is renamed.
+
+> `kb/` is **private to the persona** — not published, not shared between
+> agents, not a document store for the operator. It is the agent's own recall.
+> OKF is the *format*; where a human-facing vault lives is a separate question.
+
 **Add Gemini embeddings** (optional, recommended for production) to layer true
 **semantic** retrieval on top — matching by *meaning*, not just words. With a
 key, search becomes **hybrid**: OKF field-weighted BM25 *and* vector similarity,
--- a/README.md
+++ b/README.md
@@ -1459,8 +1459,13 @@ prompt from that constant:
 
 Adopting the vocabulary is **not a migration**. An alias map folds legacy
 values (`troubleshooting` → `runbook`, `home` → `index`) onto canonical ones
-at query time, so every note ever written stays valid and only newly-authored
-notes converge. Nothing on disk is renamed.
+**at index time**, and each note is indexed under both spellings — the
+canonical type *and* the one its frontmatter actually carries — so it stays
+findable either way. Bumping the notes-schema version rebuilds an existing
+index from disk on next open, so notes written long before the vocabulary
+existed converge too, not just newly-authored ones. Nothing on disk is
+renamed: the folding lives in the index, and your frontmatter is left exactly
+as you wrote it.
 
 > `kb/` is **private to the persona** — not published, not shared between
 > agents, not a document store for the operator. It is the agent's own recall.
--- a/src/lib/nightly.ts
+++ b/src/lib/nightly.ts
@@ -35,6 +35,7 @@ import { createHash } from "node:crypto";
 import { readFile, readdir, stat, writeFile } from "node:fs/promises";
 import { join } from "node:path";
 import { log } from "./logger.ts";
+import { OKF_AGENT_TYPES, OKF_CORE_TYPES } from "./okf.ts";
 
 const NIGHTLY_PROMPT_OVERRIDE = "nightly-prompt.md";
 
@@ -531,7 +532,14 @@ touch memory/ files.
 
 Read the daily file (memory/${today}.md) for durable knowledge (procedures,
 configs, runbooks, concepts, decisions worth keeping). For each candidate:
-  a) \`phantombot memory search "<topic>"\` to check for existing coverage.
+  a) Search for existing coverage — and search it from MORE THAN ONE ANGLE.
+     Run \`phantombot memory search\` two or three times per candidate, varying
+     the vocabulary: the symptom as you'd describe it today, the component or
+     service name, and the literal error string or command. Today's wording and
+     the existing note's wording are often disjoint ("CI agents refusing jobs"
+     vs "runner version deprecated"), and a single-phrase search will miss the
+     note you are about to duplicate. A duplicate is worse than a long search:
+     it splits one truth across two notes and dilutes every future query.
   b) If a note already covers the area, open it and RECONCILE — don't just append:
      - ADDS (new case, edge case, link): update in place.
      - CONTRADICTS / INVALIDATES it (a fact changed, something was decommissioned,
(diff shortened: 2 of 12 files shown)


src/lib/okf.ts
```diff
@@ -45,6 +45,104 @@ export interface OkfLink {
...
+ * are predictable. Left unstated, `type` drifts: real phantom KBs in the wild
+ * accumulate a dozen-plus near-synonyms ("note"/"atomic-note"/"concept",
+ * "troubleshooting"/"runbook") that fragment the very column meant to sharpen
+ * recall.
+ *
+ * So the vocabulary is declared HERE, once, and the persona prompt and nightly
+ * prompt are generated from it — a prompt that hardcoded its own list would
+ * silently diverge from the parser the first time either changed.
+ *
+ * CORE is deliberately interoperable: these types carry the same meaning in a
```
andrewagrahamhodges, line 63: 🟢 **Major**: The alias map is well-populated and correctly targets observed legacy values. The `normaliseOkfType` logic is sound and ensures that old notes remain valid without forcing a migration.

src/lib/nightly.ts
```diff
@@ -531,7 +532,14 @@ touch memory/ files.
 
 Read the daily file (memory/${today}.md) for durable knowledge (procedures,
 configs, runbooks, concepts, decisions worth keeping). For each candidate:
-  a) \`phantombot memory search "<topic>"\` to check for existing coverage.
+  a) Search for existing coverage — and search it from MORE THAN ONE ANGLE.
+     Run \`phantombot memory search\` two or three times per candidate, varying
+     the vocabulary: the symptom as you'd describe it today, the component or
+     service name, and the literal error string or command. Today's wording and
```
andrewagrahamhodges, line 538: 🟢 **Major**: Multi-angle dedup search is a critical fix for BM25-only installs. By searching for symptoms, components, and literal errors, we significantly reduce the chance of creating duplicates due to disjoint wording.

src/persona/builder.ts
```diff
@@ -197,11 +198,42 @@ Layout (relative to your working dir):
   memory/norms.md            — structured drawer (what's ROUTINE in your owner's
                                world; briefs the threat judge so it doesn't
                                cry wolf on normal operations)
```
andrewagrahamhodges, line 200: 🟢 **Major**: Interpolating `OKF_TYPES` directly into the prompt is the correct way to ensure the agent's vocabulary stays in sync with the parser. The updated `kb/` description is also much clearer.

andrewagrahamhodges (commented):
## Code Review: feat(okf): declare a controlled KB vocabulary (#389)

**Author:** @robertclawson | **Files:** 10 | **Verdict:** APPROVED

### Findings

| 1 | 🟢 Major | `okf.ts` | 62 | The `OKF_TYPE_ALIASES` map is the right way to handle drift without a migration. The `normaliseOkfType` logic is robust. |
| 2 | 🟢 Major | `nightly.ts` | 538 | Multi-angle dedup is a significant improvement for BM25-only installs where paraphrased coverage is the norm. |
| 3 | 🟢 Major | `builder.ts` | 200 | Correctly generating the prompt from `OKF_TYPES` ensures the vocabulary doesn't drift. |
| 4 | 🟡 Minor | `personaScaffold.ts` | 152 | Good catch on renaming `atomic-note.md` to `concept.md` to match the type it declares. |

### What's Good
- **Single Source of Truth:** Declaring the vocabulary in `okf.ts` and interpolating it into both the persona and nightly prompts is a textbook way to prevent the very drift this PR aims to fix.
- **Drift Tolerance:** The `normaliseOkfType` function returning `""` for unknowns is a great design choice — it makes drift visible rather than hiding it.
- **Comprehensive Testing:** The 22 new assertions covering alias-map integrity and prompt contents are exactly what's needed for a change of this scope.
- **CI Green:** All checks passed.

<details>

```
Review and fix the following issues found in PR #389 (feat(okf): declare a controlled KB vocabulary):

1. [Major] okf.ts:62 — Ensure the OKF_TYPE_ALIASES map covers all observed legacy values to prevent prompt drift. The current map looks solid.
2. [Major] nightly.ts:538 — Verify that the multi-angle search instruction is clear enough for the agent to vary its vocabulary effectively.

Branch: feat/okf-controlled-vocabulary
Base: main
```
</details>

src/lib/okf.ts
```diff
@@ -6,9 +6,9 @@
  * (`type`, `title`, `description`, `tags`, …) and a body, where concepts link
  * to each other with plain markdown links to form a graph.
  *
- * Phantombot's kb/ second brain is already this shape (Obsidian-style atomic
- * notes with frontmatter + links). This parser extracts the structured parts
- * so the FTS5 index can:
+ * Phantombot's kb/ second brain IS this shape: atomic markdown notes with
```
kaieriksen, line 9: Blocking: this comment claims type has its own BM25F column, but the notes FTS table and ingestion path in memoryIndex.ts have no type column and never call normaliseOkfType. The new vocabulary therefore only changes prompts/tests; it does not affect retrieval. Wire the canonical type into indexing and handle existing SQLite schemas before this can be approved.

kaieriksen (changes requested):
Request changes: the controlled type vocabulary is not connected to the memory index. Please add a type column to the notes FTS schema and insert a canonicalized type (including legacy aliases) during indexing, with the required migration/rebuild path for existing databases. As written, memoryIndex.ts still creates notes with only title/tags/aliases/headings/body and inserts doc.tags/doc.aliases raw; normaliseOkfType is never called, so type is not a BM25F field and the promised alias folding at query/index time does not occur.

robertclawson:
Fixed in 0aa24ad — you were right on both counts, thanks.

The vocabulary genuinely stopped at the prompts: `notes` had no `type` column and `normaliseOkfType` had no caller outside its own tests, so nothing I declared reached retrieval. The header comment claiming a BM25F column for `type` was describing an intention, not the code.

What changed:

- **`type` is now a real BM25F column** on the `notes` FTS table, weighted **4.0** — deliberately below tags/aliases (6.0) and title (8.0). A closed vocabulary of a dozen values is a coarse filter rather than a distinguishing term, so a note merely *typed* `runbook` shouldn't outrank the note actually *titled* "Runbook". There's a test pinning that ordering.
- **Aliases fold at index time**, in `indexedNoteType`, which stores the canonical type *plus the raw spelling when they differ* (`atomic-note` → `"concept atomic-note"`). Canonical-only would have made pre-vocabulary notes unfindable by the name they actually carry — the exact migration cost this design set out to avoid. Raw-only would fragment the column the vocabulary exists to merge. Unknown types keep their own spelling so drift stays visible to search instead of vanishing into `""`.
- **`NOTES_SCHEMA_VERSION` 3 → 4**, which is the migration path you asked for. The existing `selfHealNotesSchema` drop-and-rebuild handles it: `notes`/`files`/`note_links` are all reconstructable from disk, so deployed indexes repopulate on next open with no operator step and no embedding cost.
- **`NOTES_BODY_COL` 6 → 7.** Worth flagging: FTS5 takes a bare integer for `snippet()` and doesn't validate it against the column name, so missing this shift would have returned empty snippets silently rather than erroring. Added a regression test.

Also corrected the two comments that described behaviour that didn't exist — folding is index-time, not query-time, and the header now names the fields it actually weights.

Tests: 14 new assertions. I verified they fail against the unwired code (3 red on the retrieval/migration paths) rather than passing vacuously. Typecheck clean; full suite green apart from 4 pre-existing `acquireRunLock`/flock failures that also fail on the base branch in this sandbox.

Ready for another look.

README.md
```diff
@@ -1421,6 +1421,51 @@ path is much stronger than plain keyword matching:
...
+  in a human-curated OKF vault as they do in a phantom's KB, so notes move
+  between the two without translation.
+- **Agent-side** — `lesson`, `decision`, `norm`, `account`. Derived from the
+  structured drawers; they have no equivalent in a curated vault, because a
+  vault records what someone decided was true and these record what the agent
+  learned the hard way.
+
+Adopting the vocabulary is **not a migration**. An alias map folds legacy
+values (`troubleshooting` → `runbook`, `home` → `index`) onto canonical ones
+at query time, so every note ever written stays valid and only newly-authored
```
kaieriksen: This now folds aliases at index time via `indexedNoteType`, not at query time. Please update this sentence (and the following “only newly-authored notes converge” wording) so the README matches the implemented migration/rebuild behavior; otherwise operators will expect query-time normalization that does not exist.

kaieriksen (changes requested):
The index and migration fix is correct, but README.md still says aliases fold at query time. The implementation folds them at index time and rebuilds stale indexes; please correct that documentation before approval.

kaieriksen (approved):
Re-reviewed at 2d951aa. The README now correctly documents index-time alias folding and schema rebuild behavior. Focused affected suites pass 134/134; direct typecheck passes; all hosted checks are green.

PR #390, opened 2026-08-17
fix(retrieval): widen tier-2 candidate pool so provenance re-ranking has room

### LINKED ISSUE #379: retrieval: tier-2 hits dropped first on context-budget exhaustion (#378 follow-up)
Follow-ups deliberately deferred out of #378 (persona-scoped cross-conversation retrieval) so the PR could ship. All four are **non-blocking** — the shipped behaviour is correct and fails safe (under-shares, never leaks). Filed so the review threads can be resolved without the findings being lost.

### 1. `SearchHit.source` is indexed but has no consumer

`turn_docs` now carries a `source` column and `upsertTurn` writes it, but nothing in `src/orchestrator/retrieval.ts` reads it — no filter, no ranking weight, no attribution text. (`durableFacts.ts` has its own separate `source` concept; it is not this one.) The provenance half of #377 is schema-only.

Cheapest useful consumer: **2** below.

### 2. Attribution suffix should carry the provenance tier

`crossAttribution()` emits `(cross-conversation: Telegram, Aug 13)`. Channel and date are weakly informative for judging *how much to trust* an excerpt. The most decision-relevant fact is "this came from a turn I generated myself and Andrew never confirmed":

```
(cross-conversation: Telegram, Aug 13 · unverified)
```

That makes the `source` column earn its keep and costs a handful of tokens.

### 3. Room audience classification fails open on unrecognised key shapes

`conversationAudience()` returns `private` for anything that isn't `:group:`-shaped or `telegram:-…`. For a **candidate** that's conservative (private content hides in more places). For the **room** it's inverted: `allowedAudiencesForRoom("private")` returns all three classes, so a room whose key shape we don't recognise receives *everything*, private turns included.

Today every live channel is covered, so this is latent, not live. It bites when a new channel lands with an unfamiliar key shape — exactly when nobody is thinking about the retrieval audience boundary. Options: derive room class from an explicit channel registry rather than key-shape regex, or fail the *room* side closed (unknown room → `public`, receives least).

### 4. Tier-2 hits are dropped first when the token budget is tight

`formatRetrieved` concatenates cross hits after tier 1 and `break`s the budget loop on the first block that doesn't fit. So tier 2 is dropped precisely when tier 1 is rich — and one oversized tier-1 snippet ends the loop outright, taking every cross hit with it regardless of score. A hit that cleared a deliberately strict absolute floor loses to a marginal in-conversation hit purely on position.

Options: reserve a small slice of the budget for tier 2, or `continue` past an oversized block instead of `break`ing.

### 5. Test-fixture note (one line)

The `per-path BM25 lookup for vector-only hits (#378)` fixture in `tests/lib-memoryIndex.test.ts` produces a raw BM25 of ~`5.4e-7` — effectively zero, because all 31 seeded docs contain every query term so IDF collapses. That is far below the production tier-2 floor (`minScore: 2.0`). It does not weaken the test (the invariant under test is *score convention*: per-path must equal undecayed, and reverting the fix does make it fail — verified). But a future reader may "fix" the fixture to produce realistic scores and quietly destroy the vector-only condition it depends on. Worth a comment saying so.

---

Refs: #377, #378. Threads on #378 resolved with a pointer here.

### LINKED ISSUE #382: Tier-2 retrieval: candidate pool sized by output cap, strong matches crowded out before relevance gate
## Summary

Tier-2 (cross-conversation) retrieval under-fires silently. The candidate pool passed to `hybridSearch` is sized with the *output cap* (`crossLimit`, default 3), so RRF fusion picks 3 winners **before** the relevance gate in `selectCrossConversationHits` ever runs. The gate can only reject candidates — it can never rescue a strong one that RRF pushed out of the top 3.

## Why it matters

RRF ranks by rank-position, not score. A single strong lexical match loses to mediocre documents that appear on *both* the FTS and vector legs. Observed in practice: a same-day group turn with raw BM25 **21.9** (floor is 2.0) was never injected.

The failure mode is silent — you get a normal-looking answer with no signal that context was dropped — so the feature will appear healthy indefinitely regardless of whether it's actually surfacing anything.

## Root cause

`src/orchestrator/retrieval.ts:~222` — the same number is used for two different jobs:

1. output cap in `selectCrossConversationHits` (a prompt-budget concern — correct)
2. candidate-pool size passed to `hybridSearch` (not a budget concern — wrong)

`limit` is clamped 0–10 in `config.ts:1078` and overridable via `PHANTOMBOT_RETRIEVAL_CROSS_LIMIT` / `cross_conversation.limit`. Raising it isn't the fix — it inflates prompt cost to buy pool depth.

Related: the comment above this call already reasons about pool starvation (exclusions applied in SQL before LIMIT "so every pool slot is an eligible hit"). Same class of bug, half-fixed — eligible ≠ best.

## Proposed fix

- Pass `limit: Math.max(crossLimit * 10, 25)` to `hybridSearch`
- Leave `limit: crossLimit` in `selectCrossConversationHits` so the hard prompt-budget cap still holds

Cost is ~nil: the inner FTS leg already fetches 25, and the vector leg is brute-force over the whole index regardless.

## Test

Regression test: a turn with high BM25 but RRF rank 4+ must still be injected.

## Priority

Low / not urgent. Current behaviour is strictly better than no tier 2. Worth doing **before** any threshold tuning on retrieval, though — otherwise thresholds get tuned against a starving pool. Good candidate to batch with the non-blocking items in #379.

robertclawson:
## Summary
- `hybridSearch`'s own `limit` was `crossLimit` (default 3), so only that many candidates ever reached `selectCrossConversationHits` — the provenance re-ranking added in #380 had nothing left to reorder and just rubber-stamped RRF's raw top-N. Widens the pool passed into `hybridSearch` to a flat `CROSS_CANDIDATE_POOL = 50`; the output cap (`crossLimit`) is unchanged.
- Carries `decayFactor` out of `hybridSearch` on `SearchHit` and folds it into the tier-2 re-rank key (`scoreOf(h) * (decayFactor ?? 1) * crossProvenanceWeight(h)`), so widening the pool doesn't drop time-decay out of the ordering. `rrfScore`/`ftsScore` stay raw, because the `minScore: 2.0` floor is calibrated on undecayed scores.
- One-line comment on the `#378` BM25 fixture (`tests/lib-memoryIndex.test.ts`) explaining why its near-zero raw score is intentional, so a future edit doesn't "fix" it into breaking the vector-only condition the test depends on.

Re-verified against current `main` (69e9d10) before writing this — #379 items 1-3 were already delivered by #380 and closed separately; this covers the two survivors (item 4 held back deliberately, see below).

Closes #382. Closes #379 (item 5; item 4 — tier-2 dropped first on token budget — is a separate, more invasive change and stays open as its own follow-up).

### Correction to the `635e792` commit body
That commit message says *"No behaviour change at the default crossLimit (5 → 50 either way)"*. That is wrong on both counts and is corrected here rather than by a force-push (the branch is already approved):

- The default `crossLimit` is **3**, not 5 (`DEFAULT_CROSS_CONVERSATION.limit = 3`, `src/config.ts:339`).
- So the old expression yielded `max(3 * 10, 25) = 30`, and the flat constant makes it **50**. That *is* a behaviour change on the default path: 20 more candidates enter the tier-2 re-rank per turn.

Why that's safe:
- Output is still capped at `crossLimit`, and the admission gates (`minScore` / `minVecScore`) are untouched — nothing new slips past the floor.
- Candidates ranked 31–50 carry low `rrfScore`, and the re-rank key is `rrfScore × decay × provenance`, so they only take a slot when genuinely better-attested or fresher — which is #382's intent.
- Cost is bounded: ≤20 extra `lookupScope` / `turnMetaForPath` / `snippetForPath` round-trips plus a slightly wider vec-only BM25 backfill, all in-process SQLite. The 50 ceiling is real — `hybridSearch` clamps `limit` to ≤50 and its RRF merge fuses at most 25 FTS + 25 vector.

## Test plan
- [x] Provenance regression test (`orchestrator-retrieval.test.ts`): two cross-conversation candidates, one with stronger raw BM25 but low-trust provenance, one weaker raw BM25 but full-trust provenance. Confirmed it **fails** without the fix and **passes** with it.
- [x] Age-gap regression test: 1-day-old vs 400-day-old (pinned at the 0.02 decay floor), same provenance, the ancient hit carrying stronger raw BM25, `limit: 1` — asserts the fresh hit wins. Mutation-checked: removing the `decayFactor` multiply fails it.
- [x] `bun test tests/orchestrator-retrieval.test.ts` — 47 pass
- [x] `bun test tests/lib-memoryIndex.test.ts` — 59 pass
- [x] `bun test` (full suite) — 2850 pass, 0 fail, 2 skip

### DIFF
--- a/src/lib/memoryIndex.ts
+++ b/src/lib/memoryIndex.ts
@@ -298,6 +298,18 @@ export interface SearchHit {
    * confidentiality boundary. Only set for conversation-turn hits.
    */
   audience?: TurnAudience;
+  /**
+   * Multiplicative time-decay factor applied to this hit's ordering inside
+   * hybridSearch (see turnDecayFactor), when decay was requested. Carried
+   * on the hit — separately from the RAW `ftsScore`/`rrfScore`, which stay
+   * undecayed for callers that threshold on relevance — so a downstream
+   * re-rank (e.g. tier-2 provenance weighting in retrieval.ts) can factor
+   * decay back in instead of silently dropping it once the hit has left
+   * hybridSearch's own decay-ordered slice. Always populated by
+   * hybridSearch (1 = no decay, when decay wasn't requested or the turn's
+   * age was unavailable). Only meaningful for conversation-turn hits.
+   */
+  decayFactor?: number;
   snippet: string;
 }
 
@@ -1381,7 +1393,8 @@ export class MemoryIndex {
             (b.ftsScore ?? 0) * (factors.get(b.path) ?? 1) -
             (a.ftsScore ?? 0) * (factors.get(a.path) ?? 1),
         )
-        .slice(0, limit);
+        .slice(0, limit)
+        .map((h) => ({ ...h, decayFactor: factors.get(h.path) ?? 1 }));
     }
 
     // Vector search. Brute-force cosine over all embeddings.
@@ -1489,6 +1502,7 @@ export class MemoryIndex {
           ftsScore: ftsHit?.ftsScore,
           vecScore: vecScores.get(path),
           rrfScore,
+          decayFactor: decayFactors?.get(path) ?? 1,
           ...turnMeta,
           snippet: ftsHit?.snippet ?? this.snippetForPath(path),
         };
--- a/src/orchestrator/retrieval.ts
+++ b/src/orchestrator/retrieval.ts
@@ -219,9 +219,15 @@ export async function retrieveContext(
       // defaults, a genuine ≈4 BM25 match stops clearing the 2.0 floor
       // past ~30 days and tier 2 goes permanently silent for exactly the
       // aged cross-chat memory #377 exists to surface.
+      // Pool width is deliberately wider than crossLimit: crossProvenanceWeight
+      // (below, via selectCrossConversationHits) re-ranks survivors by trust,
+      // and a pool truncated to crossLimit BEFORE that re-rank runs leaves
+      // nothing for it to rerank — it would just rubber-stamp RRF's raw top-N
+      // (#382). hybridSearch's own limit clamp (≤50) is the real ceiling; this
+      // just asks for enough candidates that provenance has room to work.
       const candidates = ix.hybridSearch(query, queryVec, {
         scope: "turns",
-        limit: crossLimit,
+        limit: Math.max(crossLimit * 10, CROSS_CANDIDATE_POOL_MIN),
         decay,
         turnFilter,
       });
@@ -396,6 +402,15 @@ export function crossAttribution(
  * hit's position among survivors and never decides admission: the floor
  * (`minScore` / `minVecScore`) still runs on RAW relevance, untouched.
  */
+/**
+ * Floor on the tier-2 candidate pool passed into hybridSearch, independent
+ * of crossLimit (the OUTPUT cap applied after provenance re-ranking, in
+ * selectCrossConversationHits). Matches hybridSearch's internal FTS/vector
+ * pool width (25 each side, see memoryIndex.ts) — no point asking for a
+ * shallower pool than hybridSearch already builds internally.
+ */
+export const CROSS_CANDIDATE_POOL_MIN = 25;
+
 export const CROSS_SOURCE_WEIGHT: Record<FactSource, number> = {
   principal: 1,
   self: 0.9,
--- a/src/orchestrator/retrieval.ts
+++ b/src/orchestrator/retrieval.ts
@@ -394,15 +394,6 @@ export function crossAttribution(
 }
 
 /**
- * Tier-2 TRUST weights, by `source` (who asserted the content).
- *
- * Weight, do not filter — the principal's call on #377. A low-provenance
- * memory is still a memory; the failure we are guarding is a weak claim
- * OUTRANKING a strong one, not a weak claim existing. So provenance moves a
- * hit's position among survivors and never decides admission: the floor
- * (`minScore` / `minVecScore`) still runs on RAW relevance, untouched.
- */
-/**
  * Floor on the tier-2 candidate pool passed into hybridSearch, independent
  * of crossLimit (the OUTPUT cap applied after provenance re-ranking, in
  * selectCrossConversationHits). Matches hybridSearch's internal FTS/vector
@@ -411,6 +402,15 @@ export function crossAttribution(
  */
 export const CROSS_CANDIDATE_POOL_MIN = 25;
 
+/**
+ * Tier-2 TRUST weights, by `source` (who asserted the content).
+ *
+ * Weight, do not filter — the principal's call on #377. A low-provenance
+ * memory is still a memory; the failure we are guarding is a weak claim
+ * OUTRANKING a strong one, not a weak claim existing. So provenance moves a
+ * hit's position among survivors and never decides admission: the floor
+ * (`minScore` / `minVecScore`) still runs on RAW relevance, untouched.
+ */
 export const CROSS_SOURCE_WEIGHT: Record<FactSource, number> = {
   principal: 1,
   self: 0.9,
@@ -544,11 +544,19 @@ export function selectCrossConversationHits(
   // floor than we have slots for, the slots go to the better-attested ones
   // rather than to whoever happened to rank higher lexically.
   //
+  // decayFactor is folded back in here deliberately: hybridSearch's own
+  // ordering is decay-adjusted, but with the pool now wider than `limit`
+  // (#382) that ordering only decides who ENTERS the pool, not who wins the
+  // final slots — this multiply is what makes decay matter again for a pool
+  // this wide. Without it, a 400-day-old hit at the decay floor can
+  // outrank a same-day hit purely on raw score × provenance (#390 review).
+  // 1 when decay wasn't requested, so this is a no-op in that case.
+  //
   // Ordering only: the weighted value is never written back onto the hit,
   // so nothing downstream can mistake it for a relevance score. Array.sort
   // is stable, so equal weights preserve the incoming best-first order.
   const ranked = out
-    .map((h, i) => ({ h, i, w: crossProvenanceWeight(h) }))
+    .map((h, i) => ({ h, i, w: crossProvenanceWeight(h) * (h.decayFactor ?? 1) }))
     .sort((a, b) => scoreOf(b.h) * b.w - scoreOf(a.h) * a.w || a.i - b.i)
     .map((r) => r.h);
   return ranked.slice(0, opts.limit);
--- a/src/orchestrator/retrieval.ts
+++ b/src/orchestrator/retrieval.ts
@@ -223,11 +223,12 @@ export async function retrieveContext(
       // (below, via selectCrossConversationHits) re-ranks survivors by trust,
       // and a pool truncated to crossLimit BEFORE that re-rank runs leaves
       // nothing for it to rerank — it would just rubber-stamp RRF's raw top-N
-      // (#382). hybridSearch's own limit clamp (≤50) is the real ceiling; this
-      // just asks for enough candidates that provenance has room to work.
+      // (#382). Taking the whole fused pool also means hybridSearch's own
+      // decay-ordered slice no longer selects anything, which is why the
+      // re-rank multiplies decayFactor back in — see selectCrossConversationHits.
       const candidates = ix.hybridSearch(query, queryVec, {
         scope: "turns",
-        limit: Math.max(crossLimit * 10, CROSS_CANDIDATE_POOL_MIN),
+        limit: CROSS_CANDIDATE_POOL,
         decay,
         turnFilter,
       });
@@ -394,13 +395,18 @@ export function crossAttribution(
 }
 
 /**
- * Floor on the tier-2 candidate pool passed into hybridSearch, independent
- * of crossLimit (the OUTPUT cap applied after provenance re-ranking, in
- * selectCrossConversationHits). Matches hybridSearch's internal FTS/vector
- * pool width (25 each side, see memoryIndex.ts) — no point asking for a
- * shallower pool than hybridSearch already builds internally.
+ * Width of the tier-2 candidate pool passed into hybridSearch, independent of
+ * crossLimit (the OUTPUT cap applied after provenance and decay re-ranking, in
+ * selectCrossConversationHits).
+ *
+ * 50 is hybridSearch's own ceiling, not an arbitrary number: it clamps `limit`
+ * to 50 and its RRF merge fuses at most 25 FTS + 25 vector paths, so asking
+ * for 50 means "hand me the whole fused pool and let the re-rank decide the
+ * output" — which is the entire point of #382. Cost is bounded by that same
+ * ceiling: the per-hit `lookupScope`/`turnMetaForPath`/`snippetForPath`
+ * round-trips only run on paths that survived the merge, ≤50 of them.
  */
-export const CROSS_CANDIDATE_POOL_MIN = 25;
+export const CROSS_CANDIDATE_POOL = 50;
 
 /**
  * Tier-2 TRUST weights, by `source` (who asserted the content).
--- a/tests/lib-memoryIndex.test.ts
+++ b/tests/lib-memoryIndex.test.ts
@@ -1047,6 +1047,16 @@ describe("per-path BM25 lookup for vector-only hits (#378)", () => {
   const vec = new Float32Array([1, 0, 0]);
   const pad = Array.from({ length: 120 }, (_, k) => `padding${k}`).join(" ");
 
+  // Note: this fixture's raw BM25 for BBB is ~5.4e-7 — effectively zero,
+  // because all 31 seeded docs contain every query term so IDF collapses.
+  // That's far below the production tier-2 floor (minScore: 2.0), and that's
+  // fine: the invariant under test is score CONVENTION (per-path must equal
+  // undecayed), not a realistic score value. Don't "fix" this fixture to
+  // produce a plausible-looking score — doing so would break the length
+  // inversion (padding vs. short fillers) that pushes BBB out of the FTS
+  // top-25 pool, which is the vector-only condition the test depends on.
+  // (#379/5)
+
   // Shared seeding helper so both tests use the same fixture.
   function seed(): void {
     // 30 short filler turns in AAA — all lexically competitive.
--- a/tests/orchestrator-retrieval.test.ts
+++ b/tests/orchestrator-retrieval.test.ts
@@ -515,6 +515,62 @@ describe("retrieveContext", () => {
     expect(out!).not.toContain("DDD");
   });
 
+  test("time-decay still wins a cross-conversation tie-break with the widened pool (#390)", async () => {
+    // Same shape as the #382 test above, but the axis under test is AGE, not
+    // provenance — both candidates share provenance (unverified/internal) so
+    // decay is the only thing that can separate them. FRESH is 1 day old;
+    // ANCIENT is 400 days old (well past the 30-day half-life, pinned at the
+    // 0.02 floor) and states the query terms more times, so it wins on raw
+    // BM25 alone.
+    //
+    // Before the #390 fix, hybridSearch's OWN decay-adjusted ordering decided
+    // who survived truncation to `limit` — but #382 widened that pool well
+    // past crossLimit, so decay only gated pool ENTRY, and the final
+    // selectCrossConversationHits re-rank (raw score × provenance) had no
+    // decay term at all: ANCIENT's stronger raw BM25 let it win outright.
+    const ix = await MemoryIndex.open(indexPath);
+    await ix.refreshStale(personaDir);
+    seedFillerTurns(ix);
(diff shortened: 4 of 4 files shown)


kaieriksen (dismissed):
Reviewed at ad8b419c175be735b6ef329e79f754d1bb48ec9a. The widened tier-2 candidate pool is correctly clamped by hybridSearch, preserves SQL filtering and raw relevance floors, and gives provenance re-ranking room to operate. Focused retrieval/index tests pass 105/105; typecheck and hosted CI are green.

lenaparkhodges (dismissed):
Approved. ✅

**Verified independently, not just read:**
- Reverted the one-line fix (`limit: crossLimit`) → new regression test **fails** with DDD (raw-BM25 winner, provenance 0.42) beating EEE (full-trust). Restored → passes. The test genuinely discriminates.
- `orchestrator-retrieval` 46/46, `lib-memoryIndex` 59/59, tsc clean, all 6 CI checks green on `ad8b419`.

**Code read:** Pool width `max(crossLimit*10, 25)` sits correctly between the internal FTS/vector pools (25 each side) and hybridSearch's 50-clamp, so provenance re-ranking in `selectCrossConversationHits` finally gets a real candidate set instead of rubber-stamping RRF's top-3. Output cap unchanged — admission floor still runs on raw scores, provenance only reorders survivors. The SQL turn filter applies before LIMIT, so widening the pool can't pull ineligible turns across the boundary.

The fixture comment on the #378 test is good hygiene — that near-zero BM25 is exactly the kind of thing a future edit would "fix" into a tautology.

No findings, blocking or otherwise. Nice catch on #382 — this was a real silent no-op.

src/orchestrator/retrieval.ts
```diff
@@ -219,9 +219,15 @@ export async function retrieveContext(
...
+      // Pool width is deliberately wider than crossLimit: crossProvenanceWeight
+      // (below, via selectCrossConversationHits) re-ranks survivors by trust,
+      // and a pool truncated to crossLimit BEFORE that re-rank runs leaves
+      // nothing for it to rerank — it would just rubber-stamp RRF's raw top-N
+      // (#382). hybridSearch's own limit clamp (≤50) is the real ceiling; this
+      // just asks for enough candidates that provenance has room to work.
       const candidates = ix.hybridSearch(query, queryVec, {
         scope: "turns",
-        limit: crossLimit,
+        limit: Math.max(crossLimit * 10, CROSS_CANDIDATE_POOL_MIN),
```
andrewagrahamhodges: **Blocking: this drops time-decay out of tier-2 ranking.**

`hybridSearch` applies decay only to its internal *ordering key* (`orderKey = rrfScore * decayFactor`) before the `slice(0, limit)`, and emits `rrfScore`/`ftsScore` **raw**. `selectCrossConversationHits` then re-ranks on `scoreOf(h) * crossProvenanceWeight(h)` — raw score, no decay.

So before this change, `limit: crossLimit` meant decay effectively *chose the survivors*: the slice was the selection. After it, 25-50 candidates come through and the final order is `raw x provenance`, with decay contributing nothing. With `DEFAULT_RETRIEVAL_DECAY` (halfLife 30d, floor 0.02) that factor spans 1.0 -> 0.02 — a much bigger ordering lever than the 0.42-1.0 provenance range it's being replaced by.

Demonstrated on this branch (fixture identical to your new test apart from the timestamps): 1-day-old `unverified/internal` vs 400-day-old `principal/channel`, `limit: 1` — `main` returns the fresh turn, this branch returns the 400-day-old one. That's exactly the "months-old ghost is effectively dead" property `DEFAULT_RETRIEVAL_DECAY` documents, now gone.

Suggested fix: carry the factor out of `hybridSearch` (e.g. an optional `decayFactor?: number` on `SearchHit`, set alongside `rrfScore`) and fold it into the re-rank key — `scoreOf(h) * (h.decayFactor ?? 1) * crossProvenanceWeight(h)`. That keeps `rrfScore` raw for the floor (which must stay undecayed, per the comment at line 215) while restoring decay to the axis it was doing its work on.

The new test deliberately neutralises this (`same day: decay a wash`), which is why it passes clean — please add the age-gap case so the ordering key can't silently lose a factor again.

src/orchestrator/retrieval.ts
```diff
@@ -396,6 +402,15 @@ export function crossAttribution(
...
  * (`minScore` / `minVecScore`) still runs on RAW relevance, untouched.
  */
+/**
+ * Floor on the tier-2 candidate pool passed into hybridSearch, independent
+ * of crossLimit (the OUTPUT cap applied after provenance re-ranking, in
+ * selectCrossConversationHits). Matches hybridSearch's internal FTS/vector
+ * pool width (25 each side, see memoryIndex.ts) — no point asking for a
+ * shallower pool than hybridSearch already builds internally.
+ */
+export const CROSS_CANDIDATE_POOL_MIN = 25;
```
andrewagrahamhodges: Minor: this insertion orphans the existing docblock. `Tier-2 TRUST weights, by source...` (line 397) was written for `CROSS_SOURCE_WEIGHT`; with the new const and its own docblock wedged between them, TS attaches the *nearest* preceding comment, so `CROSS_SOURCE_WEIGHT` now has no doc on hover and the old block documents nothing.

Move `CROSS_CANDIDATE_POOL_MIN` + its comment above line 397 (or below the weight tables).

src/orchestrator/retrieval.ts
andrewagrahamhodges: Minor / simplification (non-blocking): `Math.max(crossLimit * 10, 25)` has a much narrower effective range than it looks. `hybridSearch` clamps `limit` to <= 50 and its RRF pool is at most 25 FTS + 25 vector = 50 entries, so this only ever yields 25, 30, 40 or 50 — and on the FTS-only path (`provider: "none"`) the inner `search(..., limit: 25)` caps it at 25 regardless of what you ask for.

Since the goal is "give provenance the whole pool", a flat `CROSS_CANDIDATE_POOL = 50` says that directly and removes the arbitrary `* 10`. If the 10x is there to bound per-hit cost (each merged hit outside the FTS pool costs a `lookupScope` + `turnMetaForPath` + `snippetForPath` round-trip on the turn hot path), that trade-off is worth one line of comment — right now it reads as a magic number.

andrewagrahamhodges (changes requested):
Verified the core claim and it holds: I reverted the one-line pool widening on the branch and the new `#382` regression test fails (`DDD` wins) — so this is a real fix with a real test, not a rubber stamp. `bun tsc --noEmit` clean, `tests/orchestrator-retrieval.test.ts` 46 pass. The `#378` fixture comment is accurate and worth having.

One blocking finding: widening the pool **silently strips time-decay from tier-2 selection**. Detail inline, with a reproduction. The other two are minor.

### Repro for the decay regression

Same fixture shape as your new test, but with an age gap instead of a same-day wash — 1-day-old `unverified/internal` vs 400-day-old `principal/channel`:

| branch | winner |
|---|---|
| `main` (pool = crossLimit) | **FRESH-1d** |
| this PR (pool = 30) | **ANCIENT-400d** |

A 400-day-old turn (decay factor at the 0.02 floor) now outranks a fresh one. Before this PR decay decided which `crossLimit` hits survived the slice; now it only decides which 30 of ≤50 enter the pool — near no-op — and the final ordering in `selectCrossConversationHits` ignores it entirely.

andrewagrahamhodges:
Pushed the decay fix at 8a3e644: carried `decayFactor` through `hybridSearch` and folded it into `selectCrossConversationHits`' re-rank key (`raw × decay × provenance`), plus the JSDoc reorder. Regression test added (FRESH 1d vs ANCIENT 400d, same provenance) — fails without the multiply, passes with it. tsc clean, orchestrator-retrieval 47/47, lib-memoryIndex 59/59.

robertclawson:
All three review findings are now addressed at `635e792`.

**Blocking — decay dropped from tier-2 ranking:** you'd already pushed the fix yourself (`8a3e644`) by the time I picked this up, so I dropped my duplicate rather than fight over the branch. I did verify the finding from source before I saw your commit, and independently reproduced your repro: with the pool widened and `decayFactor` left out of the re-rank key, a 400-day-old turn at the 0.02 floor wins over a 1-day-old one. Your `decayFactor`-on-`SearchHit` fix is the same shape I'd landed on.

**Minor 1 — orphaned docblock:** also yours in `8a3e644`; `CROSS_SOURCE_WEIGHT` has its doc back.

**Minor 2 — the `* 10` magic number:** taken, `635e792`. Flat `CROSS_CANDIDATE_POOL = 50`. You're right that the old expression's range was never what it looked like (25/30/40/50, and 25 on the FTS-only path regardless), and the per-hit cost the `10x` seemed to be bounding is already bounded by the ≤50-wide RRF merge — so the const now says "give me the whole fused pool" directly, with the ceiling and the cost bound documented. No behaviour change at the default `crossLimit` (5 → 50 either way).

Verification on `635e792`: `bun tsc --noEmit` clean, full suite 2851 pass / 0 fail / 2 skip, `orchestrator-retrieval` 47 and `lib-memoryIndex` 59 green. I also confirmed your new age-gap test genuinely discriminates — reverting the `decayFactor` multiply in the re-rank key fails it with ANCIENT winning, restoring it passes.

Heads up @kaieriksen @lenaparkhodges: both your approvals were dismissed by the two follow-up pushes, so this needs a re-look. @andrewagrahamhodges your `CHANGES_REQUESTED` is still what's blocking the merge.

kaieriksen (approved):
Re-reviewed current head 635e792 after the follow-up pushes. The decayFactor is now carried through hybridSearch and multiplied into the tier-2 provenance re-rank, while the flat 50-candidate pool is bounded by hybridSearch’s fused-pool ceiling. Exact-head verification passed: orchestrator-retrieval + lib-memoryIndex 106/106, and bun tsc --noEmit clean. No new findings.

lenaparkhodges (approved):
Re-review after 8a3e644 + 635e7925 — both outstanding review comments are addressed and this is good to merge once the changes-requested is cleared.

**Blocking (decay dropped from tier-2):** fixed properly. `decayFactor` is carried on `SearchHit` alongside the raw `rrfScore`/`ftsScore` (floor stays undecayed, per the line-215 contract), and `selectCrossConversationHits` folds it into the re-rank key: `scoreOf(h) * (decayFactor ?? 1) * crossProvenanceWeight(h)`. Defaults to 1 when decay wasn't requested, so the no-decay path is a genuine no-op.

**Age-gap regression test:** added exactly as asked — 1-day-old vs 400-day-old (pinned at the 0.02 floor), same provenance, ANCIENT with stronger raw BM25, `limit: 1`. Asserts FRESH wins. This is the case that would have caught the dropped factor.

**Docblock orphan + magic number:** both resolved by the flat `CROSS_CANDIDATE_POOL = 50` — the new const sits above the weight tables with its own docblock (which also documents the ≤50 cost bound raised in review), and `CROSS_SOURCE_WEIGHT`'s docblock is reattached.

Verified locally on 635e7925: `orchestrator-retrieval` + `lib-memoryIndex` suites 106 pass / 0 fail, `tsc --noEmit` clean.

src/orchestrator/retrieval.ts
```diff
@@ -387,6 +394,20 @@ export function crossAttribution(
...
+ * selectCrossConversationHits).
+ *
+ * 50 is hybridSearch's own ceiling, not an arbitrary number: it clamps `limit`
+ * to 50 and its RRF merge fuses at most 25 FTS + 25 vector paths, so asking
+ * for 50 means "hand me the whole fused pool and let the re-rank decide the
+ * output" — which is the entire point of #382. Cost is bounded by that same
+ * ceiling: the per-hit `lookupScope`/`turnMetaForPath`/`snippetForPath`
+ * round-trips only run on paths that survived the merge, ≤50 of them.
+ */
+export const CROSS_CANDIDATE_POOL = 50;
```
andrewagrahamhodges, line 409: Non-blocking, but the commit message on `635e792` is factually off and it'd be good to correct it before this lands in the history:

> No behaviour change at the default crossLimit (5 → 50 either way).

The default `crossLimit` is **3**, not 5 — `DEFAULT_CROSS_CONVERSATION.limit = 3` in `src/config.ts`. So at the default the old expression yielded `max(3 * 10, 25) = 30`, and this changes it to **50**. That *is* a behaviour change on the default path: 20 more candidates enter the re-rank on every tier-2 turn.

I think the change is still right, and I'm not blocking on it:
- Output is still capped at `crossLimit`, and the admission gates (`minScore` / `minVecScore`) are untouched, so nothing new can slip past the floor.
- Candidates ranked 31–50 carry low `rrfScore`, and the re-rank key is `rrfScore × decay × provenance`, so they only win a slot when they're genuinely better-attested/fresher — which is exactly #382's intent.
- The extra cost is bounded: ≤20 more `lookupScope`/`turnMetaForPath`/`snippetForPath` round-trips plus a slightly wider vec-only BM25 backfill, all in-process SQLite.

So: amend the commit body (or note it in the PR description) to say the default pool goes 30 → 50 and why that's safe, rather than claiming no change. The docblock itself is good as written.

src/lib/memoryIndex.ts
```diff
@@ -298,6 +298,18 @@ export interface SearchHit {
...
    */
   audience?: TurnAudience;
+  /**
+   * Multiplicative time-decay factor applied to this hit's ordering inside
+   * hybridSearch (see turnDecayFactor), when decay was requested. Carried
+   * on the hit — separately from the RAW `ftsScore`/`rrfScore`, which stay
+   * undecayed for callers that threshold on relevance — so a downstream
+   * re-rank (e.g. tier-2 provenance weighting in retrieval.ts) can factor
+   * decay back in instead of silently dropping it once the hit has left
+   * hybridSearch's own decay-ordered slice. Always populated by
```
andrewagrahamhodges, line 308: Nit on my own wording from `8a3e644` — "Always populated by hybridSearch" isn't quite true. The FTS-only-**and**-no-decay early return a few hundred lines down is `return ftsHits.slice(0, limit);`, and `search()` never sets `decayFactor`, so those hits come back with it `undefined`.

Harmless in practice — every consumer reads it as `h.decayFactor ?? 1`, and `undefined` and `1` mean the same thing there — but the doc promises an invariant the code doesn't hold. Suggest softening to "Populated by hybridSearch whenever decay was requested; treat `undefined` as 1." Purely a comment fix, no code change needed.

andrewagrahamhodges (approved):
Re-reviewed at `635e792`. **Approving** — the blocking decay regression is fixed and verified, and the non-blocking `*10` finding is resolved cleanly.

**Verified on the branch:**
- `tsc --noEmit` clean
- `tests/orchestrator-retrieval.test.ts` — 47/47
- `tests/lib-memoryIndex.test.ts` — 59/59
- Mutation-checked the new decay regression test (removing the `decayFactor` multiply makes it fail; restoring it passes), so it's a real guard and not decorative.
- Re-read `hybridSearch`: the inner `search()` is still called **without** `decay`, so folding `decayFactor` into the tier-2 re-rank is not double-counting. `ftsScore`/`rrfScore` stay raw, which is what the `minScore: 2.0` floor is calibrated on.
- Confirmed the pool-ceiling reasoning: `limit` clamps to 50 and the RRF merge fuses ≤25 FTS + ≤25 vector, so `CROSS_CANDIDATE_POOL = 50` really is "the whole fused pool".

Two non-blocking notes inline. Neither needs to hold the merge.

src/orchestrator/retrieval.ts
robertclawson: You're right on both counts — confirmed `DEFAULT_CROSS_CONVERSATION.limit = 3` at `src/config.ts:339`, so the default pool goes **30 → 50**, not "no change", and the "5" in that commit body was simply wrong.

Corrected in the PR description rather than by amending `635e792`: a force-push would dismiss all three approvals over a commit-message wording fix. The description now carries the 30 → 50 statement plus the three safety arguments (output still capped at `crossLimit`, floors untouched, ≤20 extra in-process round-trips bounded by the ≤50 fused-pool ceiling). Squash-merging will take the PR title + body into history, so the corrected version is what lands on `main` — the stale line only survives if this is merged with individual commits kept.

Flagging that last bit explicitly since it's your call at merge time.

src/lib/memoryIndex.ts
robertclawson: Agreed — verified it: the FTS-only + no-decay early return is `return ftsHits.slice(0, limit);` and `search()` never sets `decayFactor`, so those hits do come back `undefined`. The doc promises an invariant the code doesn't hold, even though every consumer reads `?? 1` so the behaviour is identical.

I've deliberately **not** pushed the wording fix. It's a comment-only change, and a push here dismisses three approvals on an otherwise merge-ready branch — a full re-review round for one line of prose. Two options, your call:

1. **Merge as-is**, and I fold `"Populated by hybridSearch whenever decay was requested; treat undefined as 1."` into the next PR that touches `memoryIndex.ts` (item 4 of #379 is the obvious candidate — it's already queued and lands in this area).
2. Say the word and I push it now, then re-request from @kaieriksen and @lenaparkhodges.

Default is (1) unless you'd rather it not sit in `main` even briefly.

PR #394, opened 2026-08-18
fix: quiet Windows keep-alive run lock conflicts

### LINKED ISSUE #392: Windows: login-fallback task logs 'already running' to stderr every 60s on the healthy path (9.1 MB of noise)
## Summary

The Windows `login-<persona>` fallback task is *designed* to be a no-op whenever the daemon is already running — but the run-lock guard treats that expected collision as a user-facing error. It writes a 4-line block to stderr and exits non-zero **every 60 seconds, forever**, on a perfectly healthy box.

On Megan's VM this grew `login.err.log` to **9.1 MB** between 3 Aug and 18 Aug with zero signal in it.

## Evidence

`login.err.log`, repeated verbatim thousands of times:

```
phantombot is already running (pid 9976; lock at C:\Users\AGHODG~1\AppData\Local\Temp\phantombot.run.lock)
view logs:    powershell -Command "Get-Content -Wait -Tail 50 "C:\Users\aghodgespwa\.local\share\phantombot\logs\phantombot.out.log""
status:       schtasks /Query /TN "\Phantombot\phantombot-megan" /V /FO LIST
stop the other instance first, or remove the lock if it's stale.
```

Both tasks confirmed live on the box:

| Task | Schedule | Status | Role |
|---|---|---|---|
| `\Phantombot\phantombot-megan` | At system start up + At logon | **Running** | real keep-alive; `IgnoreNew` correctly suppresses its own minute-tick |
| `\Phantombot\login-megan` | At logon + **repeat every 1 min** | **Ready** | fallback twin; not held Running, so every tick spawns a fresh process that hits the guard |

## Root cause

The collision is explicitly the intended healthy behaviour. `src/lib/taskScheduler.ts:126-132`:

> Interactive login-fallback task. Registered ONLY in password mode ... **The single-instance run-lock makes it a no-op when the boot task already started the daemon.**

But `src/cli/run.ts:315-322` implements that "no-op" as an error:

```ts
const lock = acquireRunLock(lockPath);
if (!isLockHandle(lock)) {
  err.write(
    `phantombot is already running (pid ...; lock at ${lock.path})\n` +
      `view logs:    ${logsCommand()}\n` +
      `status:       ${await statusCommand()}\n` +
      "stop the other instance first, or remove the lock if it's stale.\n",
  );
  return 1;
}
```

That message is right for a human who typed `phantombot run` in a terminal and wants to know why nothing happened. It is wrong for a supervisor tick whose *contract* is "start it if it's dead, otherwise do nothing."

Same guard, two callers, opposite expectations — and the noisy one is the caller that fires 1,440 times a day.

## Proposal

Make the guard quiet on the healthy path, without losing the interactive diagnostic:

1. Add a supervised/keep-alive mode to `run` (e.g. `--if-not-running`, or an env marker set by the generated launcher). In that mode a held lock is **success**: log at `debug` and `return 0`, write nothing to stderr.
2. Keep today's full 4-line message + `return 1` for the interactive path, unchanged.
3. Register the `login-<persona>` task (and the main task's minute-tick, for consistency) with that flag in `generateLoginFallbackTaskXml`.

Option 2 if a flag is unwanted: keep the message but gate it on stderr being a TTY. Cheaper, though it silences the message in any redirected-but-legitimate interactive use.

Either way `login.err.log` should stay empty on a healthy box, so anything appearing in it is real.

## Secondary (not asked for, noting it)

Log **rotation** is a documented gap — `src/lib/taskScheduler.ts` says *"Log ROTATION is not yet handled here (documented limitation; WinSW is the upgrade path)."* Fixing the guard removes this particular firehose, but `tick.err.log` had also reached 4.5 MB on the same box, so rotation is worth its own issue.

Verified on Megan (Windows 11 VM, v1.1.264). Logs manually trimmed 18 Aug: 14.5 MB → 290 KB.

kaieriksen:
## Summary

- Add `phantombot run --if-not-running` for supervisor keep-alive invocations.
- Treat an existing run lock as a silent successful no-op in that mode.
- Keep the interactive `phantombot run` diagnostic unchanged.
- Update Windows main and login-fallback task XML to use the quiet mode.
- Add regression tests and document the flag.

Closes #392

## Validation

- `/home/kai/.bun/bin/bun test tests/cli-run.test.ts tests/lib-taskScheduler.test.ts` — 99 passed
- `/home/kai/.bun/bin/bun tsc --noEmit` — passed
- `git diff --check` — passed


src/channels/core/engine.ts
```diff
@@ -1433,7 +1433,7 @@ async function processChatMessage(
 
   // The controller was aborted from outside. Causes:
   //   - "stop"      — /stop slash command (slash handler already replied).
-  //   - "reset"     — /reset slash command (handler replied; history wiped).
+  //   - "reset"     — /reset slash command (handler replied; context window cleared).
```
lenaparkhodges, line 1441: This file (and `src/memory/store.ts`) comes from commit `4b9246b`, which duplicates #376 — already squash-merged to main as `86c29b2`. Please drop that commit when rebasing; the fix commit `66c6f6a` is the real content of this PR, and these duplicated doc hunks are part of what conflicts with main.

lenaparkhodges (dismissed):
**Approved — the fix is correct and verified.** 📊

`--if-not-running` is the right shape: supervisor mode treats a held run lock as a healthy no-op (debug log + exit 0), the interactive 4-line diagnostic is untouched, and both task XML generators plus `specArgsForLabel` are kept in sync so drift detection re-imports the updated tasks on the next `ensureTasksCurrent`.

**Verified locally (on `66c6f6a`):**
- 99/99 in `cli-run` + `lib-taskScheduler`, `tsc --noEmit` clean.
- Discrimination check: reverted the `ifNotRunning` branch → new supervisor test **fails** (exit 1); restored → passes. The test genuinely guards the fix.

**Two mechanical blockers before merge (code itself is done):**
1. **Drop commit `4b9246b` on rebase** — it duplicates #376, already squash-merged to main as `86c29b2`. Its engine.ts/store.ts hunks are also what conflicts with main (see inline).
2. **CI never ran the real matrix** — only CodeQL fired. GitHub skips `pull_request` workflows when it cannot build the test merge commit (merge conflict), so test / test-windows / typecheck have not validated this head. After the rebase, wait for the full 6-check matrix before merging.

Closes a real, well-evidenced issue (#392 — 9.1 MB of healthy-path noise on Megan). Nice work, Kai.

lenaparkhodges (changes requested):
Re-reviewed after the squash/rebase (ca52a8a). My two original blockers are resolved — the duplicate commit is gone and CI is now running the full matrix.

**One blocking issue: typecheck is red.**

`tests/lib-taskScheduler.test.ts:21` imports `generateNightlyTaskXml`, which doesn't exist in `taskScheduler.ts`. It's also never used in the test body (TS6133 confirms). Both `typecheck` and `test-windows` jobs fail on `tsc --noEmit` because of this.

Fix: remove line 21 (`generateNightlyTaskXml,`) from the import block. One-line fix.

Everything else from my prior review stands — the `--if-not-running` flag, task XML generators, and `specArgsForLabel` are all correct.

lenaparkhodges (approved):
Previous blocker (broken `generateNightlyTaskXml` import causing typecheck failure) is fixed — replaced with the correct `generateLoginFallbackTaskXml` export.

CI is fully green: typecheck, test, test-windows, CodeQL all pass.

The fix is clean and well-contained:
- `--if-not-running` flag on `runRun` returns exit 0 with empty stderr when the lock is held (supervisor keep-alive path)
- Default behavior unchanged — interactive invocations still get the diagnostic error
- Both `generatePhantombotTaskXml` and `generateLoginFallbackTaskXml` now pass `--if-not-running` in args
- `specArgsForLabel` kept in sync
- Tests cover both the quiet-success and default-diagnostic paths, plus XML generation assertions

Approving.

PR #395, opened 2026-08-18
fix(screen): single-writer for held notifications (fixes #381)

### LINKED ISSUE #381: Held notifications are persisted twice into the same conversation (double-counted in retrieval)
## Summary

When the threat judge HOLDs an untrusted request, the notification text is persisted **twice** into the same conversation by two independent code paths. Both copies are embeddable and indexed, so tier-2 retrieval can return them as two separate hits for one query — inflating the weight of a single event and making it look like two corroborating pieces of evidence.

Split out of #380 (deliberately kept out of scope there; #380 fixed the *origin stamp* on one of the two rows, not the duplication itself).

## The two writes

1. **`src/cli/notify.ts:222`** — `persistNotification()` appends the delivered text to the recipient's conversation key (`telegram:<chatId>`), prefixed `[notification] `, `origin: "notification"`.
2. **`src/orchestrator/screen.ts:314`** — `recordHeld()` appends the judge's `notifyText` as an assistant turn in `episode.conversation` (the principal's conversation — usually the same `telegram:<id>`), unprefixed, `embeddable: true`, `origin: "notification"` (as of #380).

The HOLD branch calls `notify(notifyMessage)` and then `recordHeld({ ..., notifyText: notifyMessage })` with the *same string*, so on every held episode both writes fire with identical content into the same key.

## Why it matters

- **Double-counting in retrieval.** Lexical + vector legs both see two near-identical docs. One event scores as two, which is exactly the signal retrieval is supposed to distinguish.
- **Byte-level near-duplicates defeat dedup.** They aren't identical — one carries the `[notification] ` prefix — so any naive exact-match dedup will miss them.
- **Noise in replay.** History replay for the principal's next turn shows the same notification twice.

## Repro sketch

Trigger a HOLD, then:

```sql
SELECT conversation, origin, substr(text,1,60) FROM turns
WHERE conversation LIKE 'telegram:%' AND text LIKE '%<notify text>%';
```

Expect two rows, one prefixed.

## Options

1. **Don't persist in `recordHeld`** — let `runNotify` be the single writer. Risk: `recordHeld` is also used with an injected `deps.notify` in tests/non-notify paths, so the grounding row would vanish where notify doesn't run.
2. **Don't persist in `runNotify` when the caller already grounds** — pass a flag from the screener. Explicit, but adds a parameter to a widely-used entry point.
3. **Dedup at write time** — `appendTurn` skips a same-conversation, same-role, same-normalised-text row inside a short window. Most general, catches other duplicate sources too, but needs a normalisation rule (strip the `[notification] ` prefix) and a window.

Leaning (1) or (3). Needs a decision before implementation.

## Acceptance

- One held episode produces exactly one embeddable, indexed row in the principal's conversation.
- Regression test that fails when the duplicate is reintroduced (not a tautology — verify it goes red).

lenaparkhodges:
## Problem

On every HOLD, the judge's notify text was persisted **twice** into the principal's conversation:

1. `runNotify` → `persistNotification()` — assistant turn, `[notification] ` prefix, `origin: "notification"`
2. `recordHeld()` → `appendTurnPair` — same string again as an embeddable assistant turn

Both copies were embeddable and indexed, so one held event scored as two near-identical hits in tier-2 retrieval, exact-match dedup missed them (the prefix differs), and history replay showed the notification twice.

## Fix — option (1) from the issue: single writer

`recordHeld` now grounds **only the quarantined payload** (one `appendTurn`, `embeddable: false`). `runNotify` remains the single writer of the notify text, with a cross-reference comment in `notify.ts` so the second write doesn't creep back. `HeldEpisode.notifyText` is removed accordingly.

Option (3) (write-time dedup in `appendTurn` with normalisation + window) was rejected: fuzzy dedup on the hot write path for all turns to fix one duplicated call site is disproportionate.

**Accepted trade-off** (noted in the issue): if notify delivery fails outright, no assistant grounding row lands — but then the principal never saw the notification either, and the quarantined payload row still anchors the episode.

## Acceptance

- ✅ One held episode → exactly one embeddable, indexed row in the principal's conversation.
- ✅ Regression test (`one held episode produces exactly one embeddable, indexed row (#381)`) pins the notify-persistence + recordHeld composition against a real in-memory store. Verified it **goes red** against the pre-fix `screen.ts` (3 failures), green after.

## Tests

- `tests/orchestrator-screen.test.ts`: 22/22 pass (stub now records `appendTurn`; grounding tests rewritten for the single-row contract).
- Full suite: 2848 pass; the only 3 failures are the pre-existing `harnesses-pi.test.ts` ambient-env isolation issue, identical on `main`.
- `tsc --noEmit`: clean.


src/orchestrator/screen.ts
```diff
@@ -409,15 +385,16 @@ export function makeScreener(
...
+    //    judge text itself was already persisted by runNotify in step 3 —
+    //    writing it here too would double-count one event in retrieval (#381). Best-effort: a failure logs but must NEVER downgrade the
     //    hold to a pass and must never throw out of the screener. No-op when
     //    telegram isn't configured or the allowlist is empty.
     try {
       const payload = content.replace(/\s+/g, " ").trim().slice(0, HELD_PAYLOAD_CAP);
       for (const conversation of principalConversations(config, persona)) {
         try {
-          await recordHeld({ conversation, payload, notifyText: notifyMessage });
+          await recordHeld({ conversation, payload });
```
robertclawson, line 396: Ordering concern: `notify()` (step 3, line 379) fires *before* this grounding write, so the sequence into the principal's conversation is: `[notification]` assistant row (via `runNotify`'s `persistNotification`), then this quarantined `user` payload row, last.

I ran a probe comparing history shape on `main` vs this branch. On `main` the old turn-pair write meant the principal's history ended `...user payload -> assistant judge-text`, so the notification framing was always the last thing before the principal's reply. On this branch, the **last row before the principal's approve/deny reply is now the raw quarantined payload itself** — and `renderStdinPayload` emits user turns bare with no quarantine marker. Up to `HELD_PAYLOAD_CAP` (2000) chars of unframed untrusted text sits immediately adjacent to the principal's "yes, go ahead", versus the 280-char preview embedded inside the notify message.

Suggest swapping the order — run this grounding write before the `notify()` call at line 379 — so the notify's assistant-authored, safely-truncated text is always the row that closes the episode, same as before this refactor. Single-writer property for the notify *text* is unaffected either way; this is purely about which row lands last.

tests/orchestrator-screen.test.ts
```diff
@@ -337,45 +339,57 @@ describe("makeScreener", () => {
...
+    // this finds TWO assistant rows containing the notify text.
     const memory = await openMemoryStore(":memory:");
     try {
       const screen = makeScreener(cfg(), "robbie", "cli:ask", [], memory, {
         recall: async () => "",
         judge: judgeOk(90, "exfil", "Forward?"),
-        notify: async () => 0,
+        // Mirror runNotify's persistNotification: same conversation key,
+        // `[notification] ` prefix, origin `notification`, embeddable.
+        notify: async (message) => {
```
robertclawson: This is the only test guarding the cross-file invariant (recordHeld no longer duplicates the notify text because `runNotify` is the single writer) — but it works by re-implementing `persistNotification`'s shape inline (`[notification] ` prefix, `origin: "notification"`) rather than calling the real `runNotify`. If `persistNotification` changes shape or stops writing, this test stays green while production silently loses the grounding row again.

Worth a companion test that wires the real `runNotify` end-to-end (via `input.memory`/`input.transport` injection) for at least one case, so the two files can't drift apart undetected.

robertclawson (changes requested):
Verified this against `main`, not taken on trust: checked out `pr395-latest` (546dd92), ran Lena's new regression test against pre-fix `main` — fails 3/3 exactly as claimed — then 22/22 green on this branch. All six CI checks pass, no unresolved threads. The single-writer reasoning holds: `notify.ts`'s `persistNotification` writes `telegram:<chatId>` / `phantomchat:<hex>`, the exact keys `principalConversations()` grounds into, so `recordHeld` dropping the duplicate `notifyText` write is correct.

One turn-ordering issue I'd want fixed before merge (inline below), plus two smaller notes: the cross-file invariant is only pinned by a stub that mimics `runNotify` rather than calling it, and `orchestrator/turn.ts:281`'s comment ("the screener already did the grounding write ... quarantined payload + judge text") is now stale — `recordHeld` only writes the payload, the judge text goes through `runNotify`. Neither blocks on its own, but flagging both.

src/orchestrator/screen.ts
lenaparkhodges: Fixed in ca62768 — the grounding write now runs BEFORE notify(), so runNotify's `[notification]` row closes the episode and the raw payload is never the last row before the principal's reply. Pinned by a new regression test (`turn ordering: notify text is the LAST row, not the raw payload`) that goes red on the old order. The stale turn.ts:281 comment is corrected in the same commit, and the header docstring now describes the post-refactor pair (9343546).

tests/orchestrator-screen.test.ts
lenaparkhodges: Agreed — added in 9343546: `wires the REAL runNotify end-to-end` injects the actual `runNotify` as the screener's notify dep (stub transport, real :memory: store) and asserts on the rows it truly writes — `[notification] ` prefix, `origin: notification`, embeddability, and ordering. If `persistNotification` changes shape or stops writing, this goes red now.

robertclawson (commented):
Ordering fix is exactly right — grounding write now fires before notify, notify text closes the episode, new regression test pins it and I confirmed it goes red against the pre-swap order. Resolved that thread.

Still open: the companion-test point (line 365 in the test file) — the #381 regression test still re-implements `persistNotification`'s shape inline instead of exercising the real `runNotify`. That's a real drift risk (if `persistNotification` changes shape or breaks, this test stays green while production loses the grounding row again), so I'm leaving that thread open rather than resolving it myself. Not asking for a third round on the ordering fix — just want that one closed (fix or explicit "won't do" with reasoning) before merge.

lenaparkhodges:
Companion-test point addressed in f01bfd3 — both tests now exercise the real `runNotify` (transport stubbed, memory injected) via a `realNotifyDep` helper, so the inline `persistNotification` mirror is gone and can't drift from `notify.ts`. Resolving the thread.

robertclawson (approved):
Both my review threads addressed — grounding write now runs before notify() so the assistant-authored `[notification]` row closes the episode, and the companion test wires the real `runNotify` end-to-end so it goes red if `persistNotification` drifts. Re-ran `tests/orchestrator-screen.test.ts` on f01bfd3: 23/23 pass.

PR #398, opened 2026-08-19
fix(scaffold): seed memory/norms.md so the judge's briefing drawer exists

robertclawson:
## The gap

`ensurePersonaScaffold` seeds four structured drawers — people, decisions, lessons, commitments. It does not seed `memory/norms.md`, which is the third of the three drawers the threat judge briefs from:

```ts
// src/orchestrator/screen.ts
const BRIEFING_DRAWERS: readonly string[] = [
  "memory/decisions.md",
  "memory/people.md",
  "memory/norms.md",   // <- never created by the scaffold
];
```

The scaffold's own layout docstring and the README's persona-dir listing both omit it too, so the *documented* shape of a persona directory has been wrong as well.

## Why it stays invisible

Three separate mechanisms each swallow the symptom:

1. **`readBriefingDrawers` skips missing files by design** (`// Missing/unreadable drawer — skip it`). A fresh persona briefs the judge on two drawers instead of three, and nothing is logged.
2. **The heartbeat eventually conjures the file.** `TAG_TO_DRAWER` maps `norm`/`norms` → `memory/norms.md`, and promotion uses `appendFile`, which creates on demand. So the drawer appears on the first `--tag norm` capture — with no `# Norms` title and no intro, unlike every other drawer.
3. **Nothing in a fresh persona mentions the drawer exists.** The other four are self-documenting via their seeded intro text; this one had no text to be self-documenting with.

## What this changes

- Seeds `memory/norms.md` with a title and an intro that states who reads it (the judge, in full) and how to write to it (`--tag norm`).
- Adds it to the scaffold docstring's layout diagram and to the README persona-dir listing.
- Exports `BRIEFING_DRAWERS` from `screen.ts` purely so the coupling is testable.
- Adds three tests, two of which are structural rather than about this one file:
  - every value in `TAG_TO_DRAWER` exists after scaffolding **and starts with an `# ` title**;
  - every entry in `BRIEFING_DRAWERS` exists after scaffolding;
  - the norms drawer's own text names the judge and the capture flag.

The first two are the real point: they make this class of gap fail CI rather than fail silently. Adding a new tag→drawer mapping or a new briefing drawer without seeding it now breaks the build.

## What this does *not* fix

Seeding an empty drawer does not give the judge a baseline. A brand-new persona still knows nothing about what's routine in its principal's world, and will still be jumpier than a seasoned one until norms accumulate. That's inherent — the baseline comes from captures, not from scaffolding. This PR fixes the plumbing (the drawer exists, is documented, is well-formed, and is discoverable to the agent that needs to fill it), not the cold-start.

Behaviour on existing personas is unchanged: the scaffold is idempotent and `write_if_missing`, so a persona that already has a `norms.md` keeps it byte-for-byte.

## Verification

- `bun test` — 2862 pass, 2 skip, 0 fail (170 files)
- `bun tsc --noEmit` — clean
- Manually scaffolded a temp persona and read the generated drawer back.

### DIFF
--- a/src/lib/personaScaffold.ts
+++ b/src/lib/personaScaffold.ts
@@ -10,6 +10,7 @@
  *   │   ├── decisions.md
  *   │   ├── lessons.md
  *   │   ├── commitments.md
+ *   │   ├── norms.md
  *   │   └── archive/
  *   └── kb/
  *       ├── Home.md
@@ -97,6 +98,7 @@ function seedFiles(today: string): Array<[string, string]> {
     ["memory/decisions.md", drawer("Decisions", `Choices with rationale. "We chose X because Y." Promoted from daily files by the nightly cycle.`)],
     ["memory/lessons.md", drawer("Lessons", "Mistakes and learnings. Grows, never shrinks.")],
     ["memory/commitments.md", drawer("Commitments", "Deadlines and obligations. The nightly cycle promotes [commitment]-tagged entries.")],
+    ["memory/norms.md", drawer("Norms", "What is ROUTINE in your owner's world. The threat judge reads this drawer IN FULL before scoring untrusted input, so entries here are what stop it flagging ordinary operations as attacks. Capture with `--tag norm`; the nightly cycle promotes [norm]-tagged entries.")],
 
     ["kb/Home.md", kbHome(today)],
 
--- a/README.md
+++ b/README.md
@@ -1380,6 +1380,7 @@ Markdown memory lives under each persona directory:
     lessons.md
     people.md
     commitments.md
+    norms.md
   kb/
 ```
 
--- a/src/orchestrator/screen.ts
+++ b/src/orchestrator/screen.ts
@@ -139,8 +139,13 @@ const PASS_ON_ERROR = (score: number, reason: string): ScreenVerdict => ({
  * threat-relevant and keeps sensitive operational memory (finances, inbox,
  * daily dumps, commitments) out of the judge entirely. Read in FULL now, not
  * as snippets — see DRAWERS_CAP_BYTES. Paths are relative to the persona dir.
+ *
+ * Exported so the scaffold can be tested against it: a drawer the judge briefs
+ * from but `ensurePersonaScaffold` never seeds is invisible on a fresh persona
+ * (missing files are silently skipped below), so the coupling is asserted in
+ * tests rather than left to memory.
  */
-const BRIEFING_DRAWERS: readonly string[] = [
+export const BRIEFING_DRAWERS: readonly string[] = [
   "memory/decisions.md",
   "memory/people.md",
   "memory/norms.md",
--- a/tests/lib-personaScaffold.test.ts
+++ b/tests/lib-personaScaffold.test.ts
@@ -7,8 +7,10 @@ import { existsSync } from "node:fs";
 import { mkdir, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
+import { TAG_TO_DRAWER } from "../src/lib/heartbeat.ts";
 import { isCanonicalOkfType, parseOkf } from "../src/lib/okf.ts";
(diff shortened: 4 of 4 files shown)


lenaparkhodges (approved):
Approved. Gap is real (BRIEFING_DRAWERS briefs from norms.md, scaffold never seeded it, silent skip hides it) and the fix is minimal: seed + docs + two structural tests that turn this class of gap into a CI failure. Verified locally: TAG_TO_DRAWER's 5 unique targets all match seeded drawers, and the new tests pass on the branch (9/9 in lib-personaScaffold). The explicit non-fix (empty drawer ≠ judge baseline) is honest scope, not a gap in the PR.

src/lib/personaScaffold.ts
```diff
@@ -97,6 +98,7 @@ function seedFiles(today: string): Array<[string, string]> {
     ["memory/decisions.md", drawer("Decisions", `Choices with rationale. "We chose X because Y." Promoted from daily files by the nightly cycle.`)],
     ["memory/lessons.md", drawer("Lessons", "Mistakes and learnings. Grows, never shrinks.")],
     ["memory/commitments.md", drawer("Commitments", "Deadlines and obligations. The nightly cycle promotes [commitment]-tagged entries.")],
+    ["memory/norms.md", drawer("Norms", "What is ROUTINE in your owner's world. The threat judge reads this drawer IN FULL before scoring untrusted input, so entries here are what stop it flagging ordinary operations as attacks. Capture with `--tag norm`; the nightly cycle promotes [norm]-tagged entries.")],
```
kaieriksen, line 101: **Blocking — make the generated security guidance match runtime.** `readBriefingDrawers` concatenates decisions, people, and norms in that order, then caps the combined briefing at 16 KiB. A large earlier drawer can therefore leave norms partially read or entirely absent, so “reads this drawer IN FULL” is false. Also, `[norm]` captures are deterministically promoted by the 30-minute heartbeat; nightly is the cognitive catch-up for promote-able items heartbeat missed. Please describe the judge as using norms as part of a capped briefing and say heartbeat promotes captures (with nightly catching missed items). This text becomes an instruction inside every new persona, so these details need to be exact.

kaieriksen (changes requested):
The additive scaffold fix is correct and preserves existing personas byte-for-byte. The structural tests are useful, focused tests pass (32/32), typecheck is clean, and CI is green.

One accuracy blocker remains in the generated norms drawer: the judge does not read norms “IN FULL” because all three drawers share a 16 KiB aggregate cap, and deterministic tag promotion belongs to heartbeat while nightly is catch-up. Please correct that user/agent-facing guidance before merge.

lenaparkhodges:
Verified Kai's accuracy blocker against main (post-merge — my approval missed this too):

1. **"IN FULL" is inaccurate** — `readBriefingDrawers` concatenates decisions → people → norms into one briefing and truncates the *aggregate* at `DRAWERS_CAP_BYTES` (16 KiB). A large earlier drawer can clip or fully displace norms.
2. **Promotion attribution is wrong** — `[norm]` captures are promoted deterministically by the 30-min heartbeat (`TAG_TO_DRAWER` + `appendFile` in heartbeat.ts); nightly is the cognitive catch-up for what heartbeat missed.

Fix scope is small: correct the seeded intro text in `personaScaffold.ts` ("part of a capped briefing", "heartbeat promotes, nightly catches up") and the norms-drawer test assertion. Tracking as a fast follow-up PR.

robertclawson:
Both points confirmed against the source, and both are correct:

- `readBriefingDrawers` truncates the *concatenated* three drawers at `DRAWERS_CAP_BYTES` (16 KiB) with a `[briefing truncated at cap]` marker — "IN FULL" overstates it and, worse, tells the agent it has no reason to keep the drawer lean.
- `TAG_TO_DRAWER` promotion is the heartbeat's deterministic pass; nightly is catch-up.

The review landed ~50 minutes after this merged, so the text is already in `main`. Corrected there in #399 rather than reopening this one.

PR #399, opened 2026-08-19
fix(scaffold): correct promotion attribution and the "IN FULL" briefing claim everywhere they ship

robertclawson:
Follow-up to #398. The review landed after that PR merged, so the two accuracy blockers are in `main`; this fixes them there — **and everywhere else the same two claims shipped**, which turned out to be the more important half.

## What was wrong

The seeded `memory/norms.md` intro is the only place a fresh persona learns how its own security-load-bearing drawer works, and it described the machinery incorrectly on two counts:

1. **"the threat judge reads this drawer IN FULL"** — `readBriefingDrawers` concatenates all three briefing drawers (decisions, people, norms) and truncates the *combined* text at `DRAWERS_CAP_BYTES` (16 KiB), appending `[briefing truncated at cap]`. The drawers go in verbatim rather than as search snippets, but the budget is shared and finite, and truncation is a **raw byte slice** over the concatenated text — it can land mid-entry: a bloated early drawer can push a later one out of the briefing entirely. An agent told the drawer is read IN FULL has no reason to keep it lean — the claim actively works against the cap.

2. **"the nightly cycle promotes [norm]-tagged entries"** — tag promotion is the *heartbeat's* deterministic, LLM-free pass over `TAG_TO_DRAWER` (`src/lib/heartbeat.ts`). Nightly is cognitive synthesis and catch-up, not the promotion path. An agent that captures a norm and expects it to appear overnight has the wrong model of when its own writes land.

## Neither claim was local to the norms drawer

That is the reason this PR grew past its title. Both errors were **copied from the surrounding code**, so fixing only the new drawer would have left the codebase internally inconsistent and still wrong:

| Where | Claim | Reality |
|---|---|---|
| `personaScaffold.ts` — people, decisions, commitments intros | "the nightly cycle promotes …" / "Promoted from daily files by the nightly cycle" | the heartbeat promotes; nightly files what it missed |
| `personaScaffold.ts` — lessons intro | named no promoter at all | `lesson` is in `TAG_TO_DRAWER` like the rest |
| `screen.ts:17` module docstring | "reads the … drawers IN FULL" | capped at `DRAWERS_CAP_BYTES` across all three |
| `screen.ts` — `BRIEFING_DRAWERS` docstring | "Read in FULL now, not as snippets" | this is the block **directly above the constant that contradicts it** |
| `screen.ts` — `readBriefingDrawers` docstring | "in FULL from the persona dir" | same |
| `threatJudge.ts:51` | "fed in FULL, not as truncated FTS snippets" | same |
| `threatJudge.ts:276` | "norms drawers IN FULL as the judge's system prompt" | same |
| `nightly.ts:15` | drawer list omits `norms` | five drawers, not four |

The `IN FULL` docstrings are what produced the wrong sentence in #398 in the first place — the drawer text was written from the comments rather than from `readBriefingDrawers`. Leaving them in place guarantees the next person writes the same sentence again.

## The fix

- **Seeded text (5 drawers).** Every intro now credits the heartbeat and names nightly as the catch-up. The norms intro additionally says it is concatenated **LAST**, so it is the first content clipped when the briefing runs long — the specific fact that makes "keep it lean" actionable rather than a vibe. `people.md` and `decisions.md` now say the judge briefs from them, which is the only signal an agent has that those two are security-relevant and not just notes.
- **Docstrings (screen.ts ×6, threatJudge.ts ×3, nightly.ts ×1).** Every entry-preservation claim → *verbatim drawer text, hard-truncated at a shared byte cap that can cut mid-entry*, with the ordering consequence stated where `BRIEFING_DRAWERS` is declared, since that is where a reader learns the order exists.
- **Test.** A structural guard over every `TAG_TO_DRAWER` target: each seeded intro must contain `heartbeat promotes` and must not credit the nightly cycle. A new drawer cannot copy the wrong attribution in silently — which is exactly how this spread to five places.

Existing personas are untouched — the scaffold is `write_if_missing`, so a persona with a drawer already on disk keeps it byte-for-byte. This changes what *new* personas are seeded with.

## What this deliberately does not do

No behaviour change. This is seeded text, comments and one test; `readBriefingDrawers`, the cap and the promotion path are untouched. If the shared cap with norms-last ordering is the wrong design — a per-drawer budget, or reordering so norms survives truncation — that is a separate change with a separate argument, not a docs PR.

## Round 3 — the replacement wording was the same lie, weaker

Review caught that `f325256`'s predecessor swapped `IN FULL` for **"whole entries, not snippets"**, which still promises the judge sees complete entries. It does not. `readBriefingDrawers` is:

```ts
Buffer.from(text, "utf8").subarray(0, DRAWERS_CAP_BYTES).toString("utf8") +
  "\n\n[briefing truncated at cap]"
```

A raw byte slice over the **concatenated** sections — no entry-boundary logic anywhere in the function. The cut lands wherever the 16 KiB budget runs out: mid-entry, mid-line, and mid-codepoint on a multibyte character, yielding U+FFFD.

Same root cause as round 1, which is the cause this PR exists to fix: the replacement was sourced from the neighbouring docstrings instead of from the implementation. Corrected across all **eight** sites (`personaScaffold.ts` ×1, `screen.ts` ×6, `threatJudge.ts` ×3).

**The nuance kept rather than deleted:** the contrast with the old design is genuinely true — the judge gets *verbatim drawer text* rather than FTS-selected snippets, and that is worth documenting. The falsehood was only the implied boundary guarantee. Both halves are now stated.

**The test asserts the class of claim, not one wording**, since the previous one survived a rename:

```ts
expect(body).not.toMatch(/in full/i);
expect(body).not.toMatch(/whole (entries|rulings)/i);
expect(body).not.toMatch(/full fidelity/i);
expect(body).not.toMatch(/not snippets/i);
```

plus a positive assertion that `norms.md` states the mid-entry truncation — so the claim cannot be removed by deleting the sentence and leaving nothing accurate in its place.

## Verification

- `bun test` — 2864 pass, 2 skip, 0 fail (170 files)
- `bun tsc --noEmit` — clean
- Scaffolded a throwaway persona and read all five generated drawers back, rather than trusting the template source
- `grep -rniE 'whole (entries|rulings)|full fidelity|IN FULL|not snippets'` across the three touched files — no matches


lenaparkhodges (dismissed):
Both blockers from my #398 review are fixed exactly as raised: the seeded norms drawer now names the shared ~16 KiB briefing cap (and the leanness it implies) instead of claiming IN FULL, and promotion is correctly attributed to the heartbeat's deterministic pass, not nightly. The test now guards both facts and locks out the IN FULL regression. Scaffold stays write_if_missing, so existing personas are untouched. ✅

robertclawson:
Heads-up @lenaparkhodges — your approval was auto-dismissed because I pushed e2d1f26 seven minutes after it, so this is BLOCKED on a fresh review rather than on anything you raised.

What is new since you looked (no behaviour change, seeded text + docstrings + one test):

- The wrong promotion attribution was not local to norms.md — people, decisions and commitments carried the same "the nightly cycle promotes" line, and lessons.md named no promoter at all. All five now credit the heartbeat's deterministic pass.
- The "IN FULL" claim also shipped in the `screen.ts` and `threatJudge.ts` docstrings, including the block directly above `DRAWERS_CAP_BYTES`. Reworded, and the ordering consequence (last drawer in `BRIEFING_DRAWERS` is dropped first) is now spelled out where the constant lives.
- `nightly.ts`'s stage docstring omitted norms from its drawer list.
- The test is now a structural guard over every `TAG_TO_DRAWER` target, so a newly added drawer cannot copy the wrong attribution in silently.

Re-requested your review. All checks green.

src/orchestrator/screen.ts
```diff
@@ -14,8 +14,10 @@
  *      deliberate reversal the principal approved: instead of semantic-
  *      searching three drawers and feeding the judge a handful of TRUNCATED
  *      FTS snippets, the screener loads the persona (identity + MEMORY) and
- *      reads the decisions/people/norms drawers IN FULL, then composes them
- *      into the judge's system prompt via buildSystemPrompt + JUDGE_NARROWING.
+ *      reads the decisions/people/norms drawers as WHOLE ENTRIES — up to a
```
kaieriksen: **Blocking — keep the replacement wording aligned with runtime.** `readBriefingDrawers` does not preserve whole entries: after concatenating raw drawer text it slices the Buffer at exactly 16 KiB, so the cutoff can land inside a ruling or line (and even inside a multibyte character). That makes the repeated “whole entries/whole rulings” claim the same class of inaccuracy this PR is meant to remove. Since this PR deliberately leaves behavior unchanged, please describe concatenated drawer contents truncated at a shared byte cap, potentially mid-entry, and update the matching wording in `threatJudge.ts`, the lower `screen.ts` docstrings, and the seeded norms text/test.

kaieriksen (changes requested):
The broader attribution cleanup is correct, the focused tests and typecheck pass, and hosted CI/CodeQL are green. One factual issue remains in the replacement contract: the current cap is a raw byte slice, not an entry-aware truncation, so the new repeated “whole entries/whole rulings” wording still overstates runtime behavior.

lenaparkhodges (dismissed):
Re-reviewed head e2d1f264 (post-push, prior approval dismissed). The extended fix checks out against source:

- Promotion attribution: heartbeat's TAG_TO_DRAWER (heartbeat.ts:55) covers decision/lesson/person/commitment/norm — all five drawer intros now correctly credit the heartbeat, nightly as catch-up.
- Norms drawer "LAST" claim matches screen.ts: BRIEFING_DRAWERS concatenates decisions → people → norms under the shared DRAWERS_CAP_BYTES, so norms is first clipped past the cap.
- New regression test walks TAG_TO_DRAWER and asserts attribution on every seeded drawer — good guard against the copy-paste that caused this.

Approve — merge remains Andrew's call.

src/orchestrator/screen.ts
robertclawson: Confirmed against the tree and fixed in `f325256` — you're right, and the replacement wording was a weaker form of the same false guarantee, which is worse than the original because it reads like it had been checked.

`readBriefingDrawers` (`src/orchestrator/screen.ts`):

```ts
Buffer.from(text, "utf8").subarray(0, DRAWERS_CAP_BYTES).toString("utf8") +
  "\n\n[briefing truncated at cap]"
```

A raw byte slice over the **concatenated** sections. No entry-boundary logic anywhere in the function, so the cut lands wherever the budget runs out — mid-entry, mid-line, and mid-codepoint on a multibyte character (U+FFFD).

Root cause of both rounds is the same one this PR exists to fix: I sourced the replacement from the neighbouring docstrings instead of from the implementation. So the fix is all **eight** sites carrying some version of the claim, not just the seeded drawer:

| file | claim |
|---|---|
| `personaScaffold.ts:101` | "whole entries, not snippets" |
| `screen.ts:17,22,26` | "WHOLE ENTRIES", "at full fidelity", "full rulings" |
| `screen.ts:142,161,503` | "whole entries", "Full rulings", "whole rulings" |
| `threatJudge.ts:51,277,280` | "whole entries" ×2, "full fidelity instead of truncated snippets" |

**One nuance I kept rather than deleted:** the contrast with the old design is genuinely true — the judge gets *verbatim drawer text* rather than FTS-selected snippets. That is the real change and it's worth documenting. The falsehood was only the implied entry-boundary guarantee. New wording states both halves: verbatim text, hard-truncated at a shared byte cap that can cut mid-entry.

The regression test now asserts the **class** of claim rather than one wording, since the last one survived a rename:

```ts
expect(body).not.toMatch(/in full/i);
expect(body).not.toMatch(/whole (entries|rulings)/i);
expect(body).not.toMatch(/full fidelity/i);
expect(body).not.toMatch(/not snippets/i);
```

plus a positive assertion that `norms.md` states the mid-entry truncation, so removing the claim can't be done by just deleting the sentence.

Verified: `bun test` 2864 pass / 0 fail, `tsc --noEmit` clean, `grep` for the claim across all three files returns nothing. Comments and seeded text only — no behaviour change.

robertclawson:
@kaieriksen your blocker is fixed in f325256 (pushed 09:58 UTC, after your review).

You were right that "whole entries / whole rulings" was still a false guarantee: `readBriefingDrawers` truncates with `Buffer.from(text,"utf8").subarray(0, DRAWERS_CAP_BYTES)` — a raw byte slice with no entry-boundary logic, so the cut lands mid-entry, mid-line, or mid-codepoint (U+FFFD).

All eight sites now say what actually happens: the judge gets **verbatim drawer text rather than FTS-selected snippets**, **hard-truncated at a shared ~16 KiB byte cap that can cut mid-entry**. The test asserts the *class* of claim, not one wording — no seeded briefing drawer may say "in full", "whole entries/rulings", "full fidelity" or "not snippets", and norms.md must state the mid-entry truncation positively — so the next paraphrase can't reintroduce it.

Still comments + seeded text only, no behaviour change. All checks green on f325256. Re-requesting review from you and @lenaparkhodges (her approval was auto-dismissed by that same push). Merge remains Andrew's call.

Known nit I deliberately left rather than push again mid-review: the rewrap in `threatJudge.ts` leaves "it is still tool-" / "less" split across lines. Happy to fold that in with any other review feedback.

lenaparkhodges (approved):
Re-reviewed at f325256 after the dismissal. Verified against the branch, not the diff text: `readBriefingDrawers` really is a raw byte slice over the concatenated sections at a shared 16 KiB cap (so the new 'verbatim but can cut mid-entry' wording is accurate), norms.md is genuinely last in `BRIEFING_DRAWERS` so the 'concatenated LAST, first clipped' caveat holds, and `TAG_TO_DRAWER` lives in heartbeat.ts with `norm`/`norms` mapped — the heartbeat-promotes attribution is right in all five drawers. Claim sweep across the four shipped files is clean (no 'in full' / 'whole entries' / 'full fidelity'), and the structural tests now assert the class of claim plus the nightly-attribution, so this can't silently regress. Kai's blocking point is addressed. Comments/docstrings/seeded text + tests only, no behaviour change; CI green. Approving.

kaieriksen (approved):
Re-reviewed exact head f325256. The prior byte-slice/entry-boundary blocker is fixed across the seeded text and code documentation, and the regression guard covers the false-claim class. Focused scaffold, screener, and judge tests passed 57/57; typecheck is clean; hosted Linux, Windows, and CodeQL checks are green.

PR #403, opened 2026-08-19
fix: survive an unanswered relay AUTH, and stop the nightly deferring to a dead sweep

### LINKED ISSUE #401: Daemon crash-loops on an unhandled NIP-42 AUTH rejection from a rate-limiting relay
## Symptom

`phantombot run` exits and systemd restarts it, repeatedly, with:

```
NOTICE from wss://<relay>/: rate limited
error: auth timed out
phantombot.service: Main process exited, code=exited, status=7/NOTRUNNING
```

Observed on two boxes in the same two storms: **22 crashes on one, 23 on the other** over five days, clustered at 2026-08-18 16:05–16:08 (~15 restarts) and 2026-08-19 06:15–06:16 (~7). Both boxes talk to the same relay set, so a single rate-limiting relay takes down every daemon at once.

## Root cause

nostr-tools handles a relay's `["AUTH", <challenge>]` frame like this
(`lib/esm/abstract-relay.js`):

```js
case 'AUTH': {
  this.challenge = data[1]
  if (this.onauth) {
    this.auth(this.onauth)   // <-- not awaited, not caught
  }
  return
}
```

`auth()` returns a promise that rejects when the relay never answers the
kind-22242 event within `publishTimeout` (`new Error("auth timed out")`), or
answers `OK:false` with a reason. On this fire-and-forget path **nothing ever
attaches a rejection handler**, so it surfaces as an `unhandledRejection`.

`installIceErrorGuard()` (#274) sees a rejection that is not a benign ICE
socket error and re-throws by design, which is fatal. Correct behaviour for a
real bug — but a relay rate-limiting our AUTH is not a bug in us, it is a
Tuesday.

This path only exists because we set `pool.automaticallyAuth` (#368): without
`onauth`, nostr-tools never calls `auth()` from the AUTH frame handler. The
other two `auth()` call sites (publish retry, subscription `auth-required:`
retry) are both awaited or `.catch()`-ed upstream, so they are unaffected.

## Blast radius

A transient relay condition kills the daemon. Everything in-process dies with
it — channels, tasks, and any in-flight nightly sweep (see the companion issue,
which is how the crash was found).

## Fix

Attach a rejection handler to the relay's auth promise before the relay can
connect, so the fire-and-forget path can never reach `unhandledRejection`.
Real awaiters must still see the rejection.

Broadening the process-level guard to swallow relay errors is the wrong lever:
the rejection carries an arbitrary relay-supplied message
(`"rate-limited: ..."`, `"blocked: ..."`), so there is nothing to match on
narrowly, and widening the fatal-rethrow policy would mask genuine bugs.

### LINKED ISSUE #402: Nightly in-flight guard defers to a dead sweep owner — backlog stalls until the next day rollover
## Symptom

A persona's nightly sweep stops making progress and `/status` eventually reads:

```
sweep stalled on 2026-08-18 since 2026-08-19T06:15:41Z
```

The service is up, healthy, and has been for hours — but nothing retries, and
the backlog sits untouched until the next UTC day rollover.

Seen on three boxes independently, most recently on two where the daemon was
crash-looping on #401:

| box | marker date | owner pid | started | pid alive? |
|---|---|---|---|---|
| A | 2026-08-18 | 800904 | 06:15:41 | no |
| B | 2026-08-18 | 799983 | 06:15:37 | no |

## Root cause

The in-flight guard in `runNightly` is **time-only**:

```ts
const beat = Date.parse(state.current.updated_at ?? state.current.started_at);
const alive = !Number.isNaN(beat) && now.getTime() - beat <= STALE_RUN_MS;
if (alive) { /* skip */ }
```

`STALE_RUN_MS` is 45 minutes, deliberately generous — one date is two turns,
each capped at the stage hard timeout.

That is sound when a sweep dies of old age, and wrong when it dies suddenly.
The sweep runs inside the daemon's own process/cgroup, so a crash or a restart
kills it mid-date and leaves a marker **seconds old**. The restart spawns a new
sweep, which reads a marker that looks perfectly fresh and stands down. It
defers to a corpse.

Nothing else fires a sweep — the trigger is startup plus the heartbeat's UTC
day rollover — so the backlog waits up to 24h. The 45-minute window also means
`/status` reports `running` for the first 45 minutes of a dead sweep, then
`error`, when the truth was knowable immediately.

## Fix

`NightlyCurrent` already records `pid`. Check the owner process with
`kill(pid, 0)` before consulting the clock: a dead owner means take over now,
regardless of how fresh the beat looks. Fall back to the existing time rule for
markers written by an older build (no `pid`) and for an owner that is alive but
wedged.

`/status` should distinguish the two, since they call for different responses:
a dead owner is a crash to investigate, a stalled one is a hung turn.

robertclawson:
Two independent faults, in one PR because the first one is what exposed the second: the relay crash killed the daemon mid-sweep on Lena's and Kai's boxes, and the nightly guard then refused to restart the sweep it had just lost.

Fixes #401. Fixes #402.

---

## 1. The daemon dies when a relay rate-limits our NIP-42 AUTH (#401)

**Evidence.** Both boxes, same journal, same two storms:

```
NOTICE from wss://<relay>/: rate limited
error: auth timed out
phantombot.service: Main process exited, code=exited, status=7/NOTRUNNING
```

**22 crashes on one box, 23 on the other**, over five days, clustered at 2026-08-18 16:05–16:08 (~15 restarts) and 2026-08-19 06:15–06:16 (~7). They share a relay set, so one relay's traffic policy takes down every daemon simultaneously.

**Root cause.** nostr-tools handles the AUTH frame like this (`lib/esm/abstract-relay.js`):

```js
case 'AUTH': {
  this.challenge = data[1]
  if (this.onauth) {
    this.auth(this.onauth)   // not awaited, not caught
  }
  return
}
```

`auth()` rejects when the relay never OKs our kind-22242 event within `publishTimeout` (`Error("auth timed out")`), or answers `OK:false` with a reason of its choosing. On this path nothing ever holds the promise, so it surfaces as an `unhandledRejection`, and the #274 guard re-throws anything that isn't a benign ICE socket error — correct for a real bug, fatal for a relay having a bad afternoon. Restart, reconnect, same rate limit, repeat.

This path exists only because we set `pool.automaticallyAuth` (#368): without `onauth`, nostr-tools never calls `auth()` from the frame handler. The other two `auth()` call sites — the publish retry and the `auth-required:` subscription retry — are both awaited and already caught.

**Fix.** `AuthGuardedSimplePool` attaches a rejection handler to each relay's `auth()`, installed from inside the `automaticallyAuth` hook. The pool calls that hook after registering the relay in `this.relays` and before `relay.connect()` — the only window that closes before a relay can send its first AUTH frame.

Because `auth()` memoises its work in `relay.authPromise` and hands the same promise to every caller, attaching a handler only marks it observed: the retry paths still see the rejection and still retry, so #368 doesn't regress. Nothing is swallowed except the crash. A failing relay is logged at `warn` with its URL, so a persistently broken one is visible rather than silent.

Three things I considered and rejected:

- **Broadening the #274 process guard to swallow relay errors.** The rejection carries an arbitrary relay-supplied message (`"rate-limited: ..."`, `"blocked: ..."`), so there is nothing to match narrowly, and widening the fatal-rethrow policy would mask genuine bugs. Out of scope, and the wrong lever.
- **Patching `AbstractRelay.prototype`.** nostr-tools ships a separate bundled copy of the class in each entry point (`relay.js`, `pool.js`, `abstract-relay.js`), so the prototype our import resolves to is not necessarily the one `SimplePool` instantiates. Wrapping the instance is identity-independent.
- **A module-level static import of the pool.** Both call sites import nostr-tools lazily on purpose, to keep the websocket machinery out of the graph for Telegram-only notifies — hence a separate module rather than adding the import to `relayAuth.ts`.

## 2. The nightly in-flight guard defers to a dead sweep (#402)

**Evidence.** After the crashes stopped, both boxes sat healthy for 4h15m with 2026-08-18 unprocessed and a marker naming an owner that no longer existed:

| box | marker date | owner pid | started | alive? |
|---|---|---|---|---|
| Lena | 2026-08-18 | 800904 | 06:15:41 | no |
| Kai | 2026-08-18 | 799983 | 06:15:37 | no |

**Root cause.** The guard was time-only: a marker younger than `STALE_RUN_MS` (45 min) was assumed to be a live sweep. That is right when a sweep dies of old age and wrong when it dies suddenly. The sweep runs inside the daemon's own process, so a crash or restart kills it while its last beat is *seconds* old — the freshest it will ever be — and the sweep spawned by the restart stands down out of deference to a process that no longer exists.

The trigger is startup plus the heartbeat's UTC day rollover, so nothing else fires: the backlog waits up to a day. `/status` compounds it by reporting `running` for the first 45 minutes of a dead sweep.

**Fix.** `NightlyCurrent` already records `pid`. `sweepLiveness()` checks the owner process with `kill(pid, 0)` before consulting the clock and returns one of three verdicts:

- `dead` — owner gone, take over now regardless of the timestamp
- `running` — owner alive and beating, stand down (unchanged)
- `stalled` — owner alive but past `STALE_RUN_MS`, take over (unchanged)

Markers with no `pid`, written by older builds, fall back to the clock exactly as before — an upgrade must never start stealing live sweeps. `/status` now says `sweep died on <date> — owner pid N is gone` versus `sweep stalled on <date>`, because a crash to investigate and a hung turn call for different responses.

`EPERM` counts as alive: the pid exists but belongs to another user.

---

## Tests

25 new assertions across 3 files; the full suite is 2883 pass / 0 fail, typecheck clean.

**The #401 crash is asserted as a real subprocess exit,** not a mocked event. `bun:test` intercepts unhandled rejections rather than letting them reach a `process.on` listener, so the honest test spawns a real process and reads its exit code — which also reproduces the production failure exactly (`error: auth timed out`, exit 1). The two runs differ by one line, whether the relay is guarded, and the fixture imports the guard module in *both*, so it also proves the fix is per-relay rather than some blanket process-level handler:

- unguarded fire-and-forget `auth()` → exit 1, no `SURVIVED`
- guarded → exit 0, `SURVIVED`
- awaiting callers still see the rejection (the retry paths)
- guarding twice wraps once (the hook re-runs on every `ensureRelay`)
- the pool guards the relay before connect, and still returns a signer that signs with the persona key

**The #402 tests were verified red before green** — reverting the pid check fails exactly three:

```
(fail) sweepLiveness > a dead owner is dead however fresh its last beat looks
(fail) nightlyHealth > a fresh marker whose owner is gone reports died, not running
(fail) runNightly — bounds and locking > a fresh marker whose owner process is gone is taken over
```

Plus coverage for the branches that must not move: a live owner with a fresh beat still blocks, a live owner past the deadline is still taken over, and a pid-less marker still falls back to the clock.

## Deploy note

Every currently-running box is holding a marker written by the old build, so the first sweep after the upgrade takes the pid-less clock path. From the sweep after that, markers carry a pid and the new rule applies. Lena and Kai still need 2026-08-18 cleared by hand (`phantombot nightly --force`) — this PR stops it recurring, it doesn't drain the existing backlog.

### DIFF
--- a/src/lib/nightly.ts
+++ b/src/lib/nightly.ts
@@ -30,7 +30,8 @@
  * bleed into Telegram chats, and both stages for a date share context.
  */
 
-import { existsSync } from "node:fs";
+import { existsSync, readFileSync } from "node:fs";
+import { execFileSync } from "node:child_process";
 import { createHash } from "node:crypto";
 import { readFile, readdir, stat, writeFile } from "node:fs/promises";
 import { join } from "node:path";
@@ -107,6 +108,12 @@ export interface NightlyCurrent {
   /** Refreshed as each date starts — drives stale-run detection. */
   updated_at: string;
   pid?: number;
+  /**
+   * OS-supplied start time of `pid`, used to spot pid REUSE (issue #402
+   * follow-up). Absent on markers written by an older build, and on
+   * platforms with no cheap probe — liveness degrades to the pid check.
+   */
+  pid_start?: string;
 }
 
 export interface NightlyState {
@@ -151,6 +158,55 @@ export function isProcessAlive(pid: number): boolean {
   }
 }
 
+/** Probe seam for the start-time half of the identity check. */
+export type ProcessStartProbe = (pid: number) => string | null;
+
+/**
+ * An opaque token that identifies WHICH process currently holds `pid`.
+ *
+ * `kill(pid, 0)` alone can be fooled: pids wrap, so between the sweep dying
+ * and the next start an unrelated process can inherit the number and be
+ * reported alive. The kernel's own start time for the pid disambiguates —
+ * the reused pid is necessarily a different process with a different start.
+ *
+ * Returns `null` when the platform has no cheap probe or the read fails; the
+ * caller then falls back to the pid check alone, which is what shipped in
+ * #403 and is still strictly better than the time-only rule it replaced.
+ * Only ever compared for equality with a token produced the same way, so the
+ * format is deliberately unspecified.
+ */
+export function processStartToken(pid: number): string | null {
+  try {
+    if (process.platform === "linux") {
+      // /proc/<pid>/stat field 22 (starttime, in clock ticks since boot).
+      // comm (field 2) is parenthesised and may itself contain spaces and
+      // parens, so split AFTER the last ')' — field 3 lands at index 0.
+      const stat = readFileSync(`/proc/${pid}/stat`, "utf8");
+      const tail = stat.slice(stat.lastIndexOf(")") + 2).split(" ");
+      return tail[19] ?? null;
+    }
+    if (process.platform === "darwin") {
+      const out = execFileSync("ps", ["-o", "lstart=", "-p", String(pid)], {
+        encoding: "utf8",
+        timeout: 5_000,
+        stdio: ["ignore", "pipe", "ignore"],
+      });
+      return out.trim() || null;
+    }
+  } catch {
+    // Process gone, permission denied, no procfs, ps missing — all mean
+    // "can't tell", and "can't tell" must never masquerade as a match.
+  }
+  return null;
+}
+
+/** Our own token never changes, and on darwin it costs a fork. Compute once. */
+let ownStartToken: string | null | undefined;
+export function selfStartToken(): string | null {
+  if (ownStartToken === undefined) ownStartToken = processStartToken(process.pid);
+  return ownStartToken;
+}
+
 /**
  * Classify an in-flight marker (issue #402).
  *
@@ -172,8 +228,18 @@ export function sweepLiveness(
   current: NightlyCurrent,
   now: Date,
   isAlive: ProcessAliveProbe = isProcessAlive,
+  startToken: ProcessStartProbe = processStartToken,
 ): SweepLiveness {
-  if (typeof current.pid === "number" && !isAlive(current.pid)) return "dead";
+  if (typeof current.pid === "number") {
+    if (!isAlive(current.pid)) return "dead";
+    // Alive, but is it the SAME process? A pid that has wrapped around to an
+    // unrelated process would otherwise hold the lock until the 45-minute
+    // stall timer expires. A null token means "can't tell" — trust the pid.
+    if (current.pid_start) {
+      const seen = startToken(current.pid);
+      if (seen !== null && seen !== current.pid_start) return "dead";
+    }
+  }
   const beat = Date.parse(current.updated_at ?? current.started_at);
   const fresh = !Number.isNaN(beat) && now.getTime() - beat <= STALE_RUN_MS;
   return fresh ? "running" : "stalled";
--- a/src/lib/nightly.ts
+++ b/src/lib/nightly.ts
@@ -122,6 +122,63 @@ export interface NightlyState {
   current?: NightlyCurrent | null;
 }
 
+/**
+ * What a sweep marker actually means right now.
+ *
+ * - `running` — the owner is alive and beating; a new sweep must stand down.
+ * - `dead`    — the owner process is GONE; the marker is a corpse, take over.
+ * - `stalled` — the owner may still exist but hasn't beaten in STALE_RUN_MS;
+ *               take over (the pre-existing, time-only rule).
+ */
+export type SweepLiveness = "running" | "dead" | "stalled";
+
+/** Probe seam so tests don't have to conjure real pids. */
+export type ProcessAliveProbe = (pid: number) => boolean;
+
+/**
+ * Does this pid name a process that exists on this box?
+ *
+ * `kill(pid, 0)` sends no signal — it only runs the kernel's permission and
+ * existence checks. ESRCH means gone; EPERM means it exists but belongs to
+ * another user, which still counts as alive.
+ */
+export function isProcessAlive(pid: number): boolean {
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch (e) {
+    return (e as NodeJS.ErrnoException).code === "EPERM";
+  }
+}
+
+/**
+ * Classify an in-flight marker (issue #402).
+ *
+ * THE BUG: liveness used to be time-only — a marker younger than
+ * STALE_RUN_MS was assumed to be a running sweep. When the daemon is killed
+ * mid-sweep (a crash, a restart, systemd taking down the cgroup) the child
+ * dies with it and leaves a marker seconds old. The very next start spawns a
+ * sweep, sees that fresh-looking marker, and stands down — DEFERRING TO A
+ * CORPSE. Nothing else fires until the next UTC day rollover, so the backlog
+ * sits untouched for up to a day, reported as `sweep stalled` only after the
+ * 45 minutes have elapsed. Observed on three separate boxes.
+ *
+ * THE FIX: the marker already records `pid`. A dead owner is dead the instant
+ * we look, whatever the clock says, so check the process first and fall back
+ * to the clock only for markers written by an older build (no `pid`) or an
+ * owner that is alive but wedged.
+ */
+export function sweepLiveness(
+  current: NightlyCurrent,
+  now: Date,
+  isAlive: ProcessAliveProbe = isProcessAlive,
+): SweepLiveness {
+  if (typeof current.pid === "number" && !isAlive(current.pid)) return "dead";
+  const beat = Date.parse(current.updated_at ?? current.started_at);
+  const fresh = !Number.isNaN(beat) && now.getTime() - beat <= STALE_RUN_MS;
+  return fresh ? "running" : "stalled";
+}
+
 export function nightlyConversationKey(date: string): string {
   return `system:nightly:${date}`;
 }
@@ -360,7 +417,12 @@ export interface NightlyHealth {
  */
 export async function nightlyHealth(
   personaDir: string,
-  opts: { now?: Date; state?: NightlyState } = {},
+  opts: {
+    now?: Date;
+    state?: NightlyState;
+    /** Test seam — override the process-liveness probe. */
+    isAlive?: ProcessAliveProbe;
+  } = {},
 ): Promise<NightlyHealth> {
   const now = opts.now ?? new Date();
   const state = opts.state ?? (await loadNightlyState(personaDir));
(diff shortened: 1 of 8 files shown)


src/lib/nightly.ts
```diff
@@ -122,6 +122,63 @@ export interface NightlyState {
...
+ * we look, whatever the clock says, so check the process first and fall back
+ * to the clock only for markers written by an older build (no `pid`) or an
+ * owner that is alive but wedged.
+ */
+export function sweepLiveness(
+  current: NightlyCurrent,
+  now: Date,
+  isAlive: ProcessAliveProbe = isProcessAlive,
+): SweepLiveness {
+  if (typeof current.pid === "number" && !isAlive(current.pid)) return "dead";
```
andrewagrahamhodges: 🟡 **Minor**: Pid-recycling window: if the dead sweep's pid has been reused by an unrelated process by the time the restarted daemon checks liveness, this returns `"running"` and the backlog waits out the full 45-minute stall timer. The stall rule self-heals so this is not blocking, but worth a comment (or a start-time/boot-id check in the marker if you want to close it properly).

tests/channels-phantomchat-authGuardedPool.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    stdout: "pipe",
+    stderr: "pipe",
+  });
+  const stdout = await new Response(proc.stdout).text();
+  return { code: await proc.exited, stdout };
+}
+
+describe("guardRelayAuthRejection", () => {
+  // The control. This is today's behaviour and it is what kills the daemon —
+  // assert the crash is real, or the test below is a false green. The fixture
```
andrewagrahamhodges, line 106: 🔵 **Trivial**: The `150` ms window before printing `SURVIVED` is tight on a loaded CI runner — a slow spawn could fail the guarded case spuriously. Consider 500 ms, or a deterministic signal if one is available.

andrewagrahamhodges (commented):
## Code Review: survive an unanswered relay AUTH, and stop the nightly deferring to a dead sweep

**Author:** @robertclawson | **Files:** 8 | **Verdict:** Comment Only (near-LGTM)

### What's Good

- The diagnosis of #401 is exactly right: `nostr-tools`' `AUTH` frame handler calls `this.auth(this.onauth)` fire-and-forget, and with `automaticallyAuth` set (#368) that rejection lands in the #274 process guard with an arbitrary relay-supplied message — there is nothing safe to match on. Wrapping the relay *instance* (not `AbstractRelay.prototype`, which nostr-tools bundles separately per entry point) is the correct, identity-independent fix, and attaching a handler to the memoised `authPromise` keeps the awaiting retry paths intact.
- The #402 fix distinguishes `dead` / `running` / `stalled` cleanly, and the pid-less fallback preserves old-marker semantics so an upgrade can't steal a live sweep. Verified that `pid: process.pid` is actually written into the marker (`src/cli/nightly.ts:325`) and read back by `sweepLiveness`.
- Test discipline is exemplary: real-subprocess exit-code assertion for the crash (with an unguarded control proving the test can go red), red-before-green verification for the three pid-check failures, and coverage of the branches that must *not* move (live owner blocks, live-but-wedged still retires, pid-less falls back to the clock).
- Windows is handled (file:// URL for the dynamic import, `Bun.spawn` instead of a shell, pid-1 EPERM test skipped on win32).

### Verification note

I could not execute the suite in my review environment (no `bun` runtime available, installs blocked by policy), so my confidence rests on line-by-line tracing of the diff plus the author's reported 2883 pass / 0 fail and red-before-green evidence — both of which are the right kind of evidence.

### Findings

| 1 | 🟡 Minor | src/lib/nightly.ts | 176 | Pid-recycling window: a dead owner's pid reused by an unrelated process classifies the marker `running` |
| 2 | 🔵 Trivial | tests/channels-phantomchat-authGuardedPool.test.ts | 103 | 150 ms SURVIVED timer may be tight on loaded CI runners |

<details>

```
Review and fix the following issues found in PR #403 (fix: survive an unanswered relay AUTH, and stop the nightly deferring to a dead sweep):

1. [Minor] src/lib/nightly.ts:176 — If the dead sweep's pid is recycled by an unrelated process before the next liveness check, sweepLiveness returns "running" and the backlog waits for the 45-minute stall timer. Fix by: optional hardening only — e.g. also record the process start time (or a boot id) in the marker and compare, or accept the window and document it. Low priority; the stall timer self-heals.
2. [Trivial] tests/channels-phantomchat-authGuardedPool.test.ts:103 — The 150 ms setTimeout that prints SURVIVED may race on slow CI. Fix by: bumping the window (e.g. 500 ms) or waiting on the rejection's microtask queue explicitly if a deterministic signal is available.

Branch: fix/relay-auth-crash-and-nightly-pid-liveness
Base: main
```
</details>

src/lib/nightly.ts
robertclawson: Fixed in 564eae1 — you were right that the pid check alone is only half an identity.

The marker now records the kernel's start time for the pid alongside it, and `sweepLiveness` compares both: same pid + different start time = reused = dead, taken over immediately instead of waiting out the 45-minute stall timer. Probe is `/proc/<pid>/stat` field 22 on Linux, `ps -o lstart=` on darwin, `null` elsewhere.

Two deliberate choices worth flagging:

- **`null` falls back to the pid check, it does not mean "different".** An unsupported platform, a permission error or a process exiting mid-read all yield `null`, and stealing a live sweep on that basis would double-file the drawers — strictly worse than the delay we are fixing. There is a test pinning that both ways.
- **Markers from the first build of this PR carry a pid but no start time.** They never consult the probe at all, so an upgrade cannot retroactively invent an opinion about them.

Own token is memoised — on darwin it costs a fork and the marker is rewritten once per date.

Verified red before green: reverting the comparison fails `a reused pid is dead, not running` and nothing else.

tests/channels-phantomchat-authGuardedPool.test.ts
robertclawson: Agreed — bumped to 500ms in 564eae1, with a comment saying why so nobody tightens it again.

The crash fires on the very next microtask drain, so the window is not measuring anything; it only has to outlast process spawn and scheduling jitter on a loaded runner. 150ms was buying flakes for no signal.

andrewagrahamhodges (approved):
Re-review after the follow-up commits — both findings resolved:

1. 🟡 pid-recycling window → closed properly, not just documented: the marker now records `pid_start` (Linux `/proc/<pid>/stat` field 22 with the comm-parens edge case handled correctly, darwin `ps -o lstart=`), and `sweepLiveness` compares it — a reused pid is `dead`. Critically, a null token means "can't tell" and degrades to the pid check, so unsupported platforms / permission errors / racing exits can never steal a live sweep, and pid-only markers from the first #403 build are unaffected (verified by the `probed === false` test).
2. 🔵 test timer → bumped to 500 ms.

Tests cover reused-pid-is-dead, same-token-is-running, null-token-falls-back, and no-pid_start-never-probes. LGTM — approving.

Issue #391, opened 2026-08-17
Woken turns and the live session are mutually invisible — concurrent turns collide on shared state


robertclawson:
## Summary

A background-woken agent turn and a live interactive session can run **at the same time, for the same persona, on the same files**, with neither aware the other exists. This has now happened at least twice; the second occurrence is fully instrumented below.

Nothing here is a misfire — every component did what it was configured to do. What's missing is a coordination layer.

## Timeline of the 2026-08-17 occurrence (all UTC)

| time | event |
|---|---|
| 14:11:14 | Kai submits a review on PR #390 (APPROVE) |
| 14:11:57 | Lena submits a review (APPROVE) |
| 14:13:00 | `andrewagrahamhodges` submits CHANGES_REQUESTED |
| — | GitHub emails all three notifications to `robbie@hodges.nl` |
| 14:15:33 | inbox poller (command-backed task 51, `*/15 * * * *`) reports `3 unread` |
| 14:15:36 | poller creates one-off wake **task 425** via `phantombot task add … --in 1m` |
| ~14:16:36 | `phantombot tick` fires task 425 → a **new `claude --print` process** starts |
| 14:18:46 | that process commits `8a3e644` on the PR branch |
| 14:20–14:21 | file mtimes in the shared working copy change **mid-turn** |
| 14:25:03 | it commits `635e792` and pushes |

Throughout, the **interactive Telegram session was live**, in the same repo, and had just been asked by the operator to handle the very same reviews. It found an uncommitted tree it had not written, three live `claude` processes, and no way to distinguish "a sibling turn is working" from "the working copy is corrupt". It stopped and asked the operator — the correct call, but it defeats the purpose of both turns.

The wake prompt is behaving exactly as designed:

> `robbie@hodges.nl` (gog) — GitHub notifications land here. A review request or mention on a PR: go to the PR, review it, help get it merge-ready.

## Why this matters

1. **Unsynchronised writes to shared state.** Three concurrent `claude` processes, one shared git working copy, no lock. A `stash` / `checkout` / `commit` from one turn silently churns another's in-flight edits.
2. **The live session cannot tell a sibling turn from corruption.** There is no registry to consult, so the only safe response to an unexplained dirty tree is to halt.
3. **Duplicated or conflicting external actions.** Two turns can both push, both comment, both re-request review, both close an issue.
[...]

PR #404, opened 2026-08-19
fix: let concurrent turns see each other, and hold task wakes while the principal is talking

robertclawson:
Part of #391.

A background-woken turn and the live interactive session can run at the same time, for the same persona, on the same files, with neither aware the other exists. Twice now they have worked the same PR and the same unlocked checkout — a contributor got duplicate review comments, and two commits landed in one PR under two different git identities.

## Why nothing was misfiring

`lib/runLock.ts` is a real kernel-enforced single-instance lock, but it guards `run` against `run`. `tick` takes its own separate `tick.lock`, which serialises tick against tick. **Nothing sits between a woken tick turn and the daemon's live interactive turn** — by design, since the daemon must keep answering while a task runs. The collision is not two daemons; it is one daemon and one oneshot, each correctly holding a lock the other never wanted.

## What this adds

Three layers, smallest blast radius first — issue directions 1, 2 and 3.

**1. Wake deferral (`cli/tick.ts`)** — the cheap fix that would have prevented both incidents. A due task does not wake a harness while an interactive turn is in flight, or within a 3-minute cooldown of one finishing.

- Checked **at fire time**, not at schedule time. In the incident the wake was scheduled long before the conversation started, and the gap between the interactive turn and the collision was **63 seconds**; anything decided when the task was created is looking at the wrong instant.
- The cooldown exists because the thing to avoid is not the turn, it's the *conversation*: the principal sends, reads, thinks, sends again. A wake landing in the gap between two messages collides just as badly as one landing mid-turn.
- **Nothing is written back to the task row.** The next tick is 60s away and re-evaluates for free, and leaving `next_run_at` alone means "how long have we been deferring" is just "how overdue is this task" — so the 15-minute starvation ceiling needs no extra column and no migration. `run_count`, one-off deactivation and `maxRuns` all stay honest.
- Deferral is **bounded**. Past the ceiling the task fires even if the principal is still typing: a task that never runs is a worse failure than a collision, because it is silent.
- **Command-backed tasks are exempt** — they run a deterministic script, never wake a harness, and cannot collide.

**2. Turn registry (`lib/turnRegistry.ts`)** — one small JSON file per in-flight turn under `$XDG_STATE_HOME/phantombot/turns/`, written tmp+rename.

Deliberately a **registry, not a mutex**. It answers "who else is mid-turn for this persona?" and lets callers decide; it never blocks and never makes a turn wait on another's exit. A hard mutex here would be worse than the bug — an interactive turn must never stall behind a 30-minute background wake, and a wedged holder would take the whole persona down.

Files rather than a `memory.sqlite` table because the writers are separate processes on different lifecycles, entries are pure runtime state nothing should retain, and a crashed turn must be detectable by *inspection* rather than cleanup.

**3. Sibling injection (`orchestrator/turn.ts`)** — when a sibling really is in flight, the turn is told so in its system prompt, and told to verify shared state rather than race it. Deliberately behavioural rather than informational: naming the resources that actually got clobbered (a checkout, a PR, a file) and the correct response is what changes the outcome. It is a hint, not an enforcement boundary.

## Crash safety

Reuses the pid + start-time check from #403, extracted to `lib/processLiveness.ts` and re-exported from `lib/nightly.ts` so existing importers are unchanged.

A live-looking entry is trusted only if the recorded pid is **still the same process** and the entry is younger than an hour. Both halves are load-bearing: the pid check catches a SIGKILLed turn (whose `release()` never ran), and the age ceiling catches an async generator abandoned without being closed, where the owning process is still very much alive.

Only a **clean finish** earns cooldown credit. An entry with no `finished_at` whose owner is gone is a crashed turn, and #402 is the standing lesson on those: never defer to a corpse. Granting a dead daemon a cooldown would park every scheduled task for three minutes after each crash — and this fleet saw 22 crashes in five days.

## Registration lives in `runTurn`

Every entry point — Telegram, phantomchat, `ask`, tick, nightly, ACP — funnels through `runTurn`, so registering there covers all of them and makes it impossible for a future entry point to forget. `runTurn` is now a thin wrapper that registers, delegates, and releases in a `finally` — the only place that reliably sees the end of a turn, which returns early on a screened hold, throws on harness failure, or streams normally.

The origin split falls out for free: chat surfaces default to `channel` (interactive), nightly passes `internal`, tick passes `task`. Only `channel` defers a wake.

## Testing

39 new tests. Verified **red before green** — reverting each fix independently fails exactly the tests that name it:

| reverted | fails |
|---|---|
| tick deferral | `a due task does NOT wake a harness while an interactive turn is live` |
| sibling injection | `a live sibling injects the concurrency notice into the system prompt` |
| pid liveness | 3 tests, incl. `an entry whose owner died is NOT running, however fresh it looks` |

Full suite 2921 pass / 0 fail, typecheck clean. `tests/lib-turnRegistry.test.ts` is added to the Windows CI subset — Megan is a Windows box and this gates her task wakes; `processStartToken` returns `null` there by design and falls back to the pid check.

## Two things worth a reviewer's eye

- **The registry is inert under `NODE_ENV=test` unless explicitly enabled.** Not squeamishness about side effects — `runTurn` registers unconditionally, so any suite touching a turn would otherwise write live-looking entries into the developer's real `$XDG_STATE_HOME` with the test runner's pid, and the box's actual `tick` would then defer real wakes for the cooldown window. The new tests opt in and point `PHANTOMBOT_TURN_REGISTRY_DIR` at a temp dir.
- **`PHANTOMBOT_TURN_REGISTRY=0` is the kill switch**, and it degrades to the pre-#391 behaviour — the old bug — rather than to a crash. Every read reports "nobody home".

## Not covered

Issue directions 4 (post-turn digest surfacing a woken turn's actions into the live conversation) and 5 (advisory locking on shared working copies) are untouched, which is why this is `Part of` rather than `Closes`. The registry is the substrate both would build on.


src/cli/tick.ts
```diff
@@ -128,6 +129,38 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
...
+      // Deliberately no write-back to the row: leaving `next_run_at` untouched
+      // means the next tick (60s away) re-evaluates for free, and "how long
+      // have we been deferring" is just "how overdue is this task" — so the
+      // MAX_DEFERRAL_MS starvation ceiling needs no extra state and no
+      // migration. The task is not marked run, so run_count, one-off
+      // deactivation and maxRuns all stay honest.
+      //
+      // Command tasks are exempt: they run a deterministic script, never wake
+      // a harness, and so cannot collide with a conversation turn.
+      if (!isCommandTask) {
```
kaieriksen: **Blocking — cover command pollers that wake an agent.** This exemption assumes command-backed tasks cannot wake a harness, but the shipped contract explicitly tells these pollers to call `phantombot ask` when they detect work (`src/persona/builder.ts` and the README example). That command runs here while the interactive turn is active, then `ask` starts another `runTurn`, bypassing the deferral layer and reopening the shared-checkout / duplicate-review collision this PR is meant to stop. Defer command tasks too, or make their agent-wake path participate in bounded deferral, and add a regression that exercises a command poller launching the documented wake path.

src/lib/turnRegistry.ts
```diff
@@ -0,0 +1,476 @@
...
+  if (v !== undefined) return !/^(0|off|false|no)$/i.test(v.trim());
+  return process.env.NODE_ENV !== "test";
+}
+
+/**
+ * Where entries live. `PHANTOMBOT_TURN_REGISTRY_DIR` overrides, which both
+ * gives tests an isolated directory and lets an operator relocate the runtime
+ * state without moving all of `$XDG_STATE_HOME`.
+ */
+export function defaultRegistryDir(): string {
```
kaieriksen, line 170: **Blocking — document the new runtime contract.** This adds `$XDG_STATE_HOME/phantombot/turns/`, `PHANTOMBOT_TURN_REGISTRY`, and `PHANTOMBOT_TURN_REGISTRY_DIR`, and changes the run/tick architecture. The repository contributing rules require both `README.md` and `AGENTS.md` for file-location and architecture-invariant changes, but neither is updated in this patch. Please document the registry lifecycle, location/overrides, kill switch, and the cooldown/starvation-ceiling behavior in both files.

kaieriksen (changes requested):
Requesting changes for two blockers: command-backed pollers can still launch an agent wake during an interactive turn, and the new runtime-state/config architecture is undocumented. Reviewed exact head `2705c401`: focused concurrency/tick/orchestrator/nightly tests pass 138/138, typecheck passes, and all hosted checks are green. The runtime-matched local full suite reached 2919 pass / 2 skip / 2 fail; the two failures are existing Pi routing tests inheriting a live ambient API key, outside this patch.

src/cli/tick.ts
robertclawson: Agreed, and fixed in a216dce — the exemption was wrong on its own terms. The comment claimed command tasks "never wake a harness", but `--command`'s own help text and the README's Jira example both tell a poller to call `phantombot ask` when it finds work, which starts a full `runTurn` in a third process. That left the documented path as an unguarded back door into exactly the collision this defers against.

Pollers now go through the same `shouldDeferWake` check as any other task. That is cheap and safe: a poller is idempotent by construction, the row is left untouched so the next tick re-evaluates 60s later, and `MAX_DEFERRAL_MS` still guarantees it runs. `isCommandTask` is now logged on the deferral line rather than gating it.

Deferring at the `ask` layer instead was the other option, and it is the wrong one — the voice agent calls `phantombot ask --stream` on a live Twilio call, so holding `ask` would stall a human mid-sentence. The scheduler is the right place for a scheduling decision.

Regression coverage replaces the old exemption test with three cases that drive the documented wake path (a marker file standing in for the poller's `ask` call): it is absent while an interactive turn is live and the task row is untouched (`run_count` 0, `next_run_at` unmoved, still active), present once the conversation goes quiet, and present again when the task is overdue past the ceiling.

src/lib/turnRegistry.ts
robertclawson: Fair catch — the docs rule is explicit and this patch skipped both files. Fixed in a216dce.

`README.md` gets a **Turn registry (concurrent turns)** subsection under Scheduled Tasks covering why `runLock` and `tick.lock` do not cover this, the registry-not-a-mutex framing, the `$XDG_STATE_HOME/phantombot/turns/` location, the deferral rules (3-minute conversation cooldown, 15-minute starvation ceiling, task row untouched while deferred, command pollers included), the sibling notice, the pid+age liveness rule for crashed turns, and a table for `PHANTOMBOT_TURN_REGISTRY` and `PHANTOMBOT_TURN_REGISTRY_DIR`. The Scheduled Tasks bullet list links into it.

`AGENTS.md` gets invariant 19 plus `turnRegistry.ts` / `processLiveness.ts` in the source map. It records the parts that have a failure mode if someone "simplifies" them later: registration lives inside `runTurn` so a new entry point cannot forget to opt in; deferral deliberately never writes back to the task row, which is what keeps `run_count`, one-off deactivation and `--count` honest with no migration; both ceilings and why each exists; why command tasks are not exempt; and that the kill switch is off by default under `NODE_ENV=test`, so a test exercising the registry must set both env vars.

robertclawson:
Pushed a216dce addressing both blockers.

**1 — command pollers are no longer exempt.** The exemption assumed a command task cannot wake a harness; the shipped contract says otherwise, so `phantombot ask` was a documented back door past the deferral. Pollers now take the same `shouldDeferWake` path, with three regressions driving that wake path (deferred while a turn is live and the row untouched; fires once quiet; fires anyway past `MAX_DEFERRAL_MS`). Deferring inside `ask` was rejected — the voice agent calls `ask --stream` on live calls.

**2 — runtime contract documented.** `README.md` gains a Turn registry subsection (lifecycle, `$XDG_STATE_HOME/phantombot/turns/`, both env vars, kill switch, cooldown + starvation ceilings, crash/pid-liveness semantics); `AGENTS.md` gains invariant 19 and the two new source-map entries.

Verification on a216dce: typecheck clean, full local suite **2923 pass / 2 skip / 0 fail** across 172 files. The two Pi routing failures you saw were an ambient live API key in your environment, not the patch — they pass here.

Re-requesting review.

kaieriksen (approved):
Approved at exact head `a216dce0`. Command-backed pollers now participate in bounded wake deferral, the task row remains untouched while held, and README.md plus AGENTS.md document the registry lifecycle, paths, kill switch, and timing ceilings. Exact-head focused concurrency/tick/orchestrator/nightly suites pass 165/165, TypeScript typecheck is clean, and all hosted checks are green. Both prior review threads are resolved.

Issue #405, opened 2026-08-19
Woken-turn visibility: post-turn digest into the live conversation, plus advisory locking on shared working copies


robertclawson:
Follow-on from #391, which is now closed: the coordination layer landed in #404 (wake deferral, persona turn registry, sibling context injection). That closed gap **A** — a starting turn can now discover an in-flight sibling and a wake defers while an interactive conversation is warm.

Gap **B** from that issue is still open, plus one narrower item explicitly deferred from #404:

**1. Post-turn digest (visibility).** A background-woken turn's material actions — commits, pushes, PR comments, merges, external messages — currently surface only into its own transcript. The interactive conversation, and the operator reading it, still get blindsided by state neither of them created. The registry now records turn id, conversation, pid and start time, so the substrate for a digest exists; what's missing is the summarisation on exit and the delivery into the live conversation.

**2. Advisory locking on shared working copies.** Narrower than the registry and independent of it: the registry is a hint, not a mutex, and deliberately never blocks one turn on another. That is the right default for turns, but it does not stop two turns mutating the same git checkout. `/tmp/phantombot-inspect` is the concrete case — an unlocked shared clone that both an interactive and a woken turn will reach for.

Why these were left out of #404: the registry had to land first as the substrate, and a hard lock on turns would have been worse than the bug it fixed (an interactive reply must never stall behind a long background wake, and a wedged holder would take the whole persona down). Locking belongs on the *working copy*, not on the turn.

Not urgent — the collision that motivated #391 is prevented now. This is the remaining half of the original report so it does not get lost.

robertclawson:
Closing — both remaining gaps are implemented and merged.

**1. Post-turn digest** — `src/lib/turnDigest.ts`, wired into `src/orchestrator/turn.ts`. A background turn (`task`/`notification`/`internal`) writes a digest of what woke it and what it materially touched; the next *interactive* turn for that persona gets pending digests injected into its system prompt (same overlay slot as the #391 sibling notice) and marks them delivered.

**2. Advisory locking on shared working copies** — `src/lib/workspaceLock.ts` plus `phantombot workspace lock | status | unlock` (`src/cli/workspace.ts`). Advisory by design, exactly as this issue specified: `lock` exits non-zero if another live turn holds the path, and nothing blocks.

[...]

PR #406, opened 2026-08-19
fix: make background turns report what they did, and let turns claim a shared checkout

robertclawson:
Part of #405. Both remaining gaps from #391 — the two #404 deliberately left out.

## The problem

#404 stopped two turns colliding. It did nothing about *why* the collisions went unnoticed for so long: a turn woken by `tick` streams its reply into its **own** transcript, which the principal never opens. A commit, a PR comment or a file edit made on their behalf leaves no trace anywhere they will look. The first we learned a sibling had acted was a contributor seeing two identical review comments.

And the collisions themselves didn't happen in phantombot's state — they happened in `/tmp/phantombot-inspect`, a git checkout two turns shared with no lock on it at all. The registry makes turns *aware* of each other; it gives them nowhere to serialise.

## 1. Post-turn digest (`src/lib/turnDigest.ts`)

A turn whose origin is `task`/`notification`/`internal` writes a digest on exit — trigger, state-changing tool calls with the paths they named, closing summary — to `$XDG_STATE_HOME/phantombot/digests/`. The next **interactive** turn gets pending digests injected into its system prompt and decides for itself whether any of it is worth mentioning.

**Delivery by prompt, not notification — deliberate.** Two alternatives were considered and rejected:

- *Telegram on every background turn.* The nightly and every poller would shout on each fire. That breaks the standing "don't notify unless it's material" rule and trains the principal to mute the channel, which returns us to zero visibility by a longer road.
- *A synthetic turn in the conversation history.* Fixes the blindsiding but forges transcript: it puts words in a conversation that never contained them, and every later retrieval treats them as something actually said.

Injecting into the prompt keeps the agent — which can see both the digest and what was just asked — as the judge of relevance. Silent when there's nothing to say.

**Load-bearing details, each with a failure mode if "simplified":**

- **At-least-once.** Marked delivered only after the receiving turn *succeeds*. Marking at injection time drops a background turn's only trace precisely when the box is unhealthy — which is when it matters most. A duplicate costs a few lines of prompt.
- **Written from a `finally`.** A background turn that pushed a commit and *then* died is the single most important case to surface; gating on success would hide exactly that turn.
- **Only interactive turns receive digests.** Handing one background turn another's digest informs nobody, and would let two of them bounce a report between themselves forever.
- **Reads are dropped, `execute` is kept whole.** Most commands are harmless (`git status`, `ls`), but we cannot tell those from `git push` without parsing every shell dialect, and the cost is asymmetric: a noisy `ls` line is trivia, an unlisted `git push` is the whole bug.
- **Independent of the audit log.** Both hang off `onToolCall`, neither depends on the other — turning off `PHANTOMBOT_AUDIT_TOOL_CALLS` is not a request to go blind to what background turns did.
- Capped at 5 per prompt with the overflow **counted** in the notice and all pending marked delivered, so a backlog can't wedge. Undelivered digests expire after 24h.

## 2. Workspace locks (`src/lib/workspaceLock.ts`, `src/cli/workspace.ts`)

```
phantombot workspace lock /tmp/phantombot-inspect --purpose "reviewing PR #405"
phantombot workspace status
phantombot workspace unlock /tmp/phantombot-inspect
```

Claims are attributed to the turn via `PHANTOMBOT_TURN_ID` (new, set beside `PHANTOMBOT_PERSONA`/`PHANTOMBOT_CONVERSATION`), which is what lets `unlock` refuse a release from a turn that never took the lock. A sibling's live claims are named in every concurrent turn's prompt.

**Be clear about the scope: this is advisory, and cannot be otherwise at this layer.** A turn runs `git` through the harness's own Bash tool; phantombot isn't in that path and cannot intercept it, so nothing here *prevents* a write to a checkout you don't hold. What it provides is a truthful, crash-safe answer to "is another turn in this tree", which previously did not exist at all. Real enforcement needs the mutation path itself to take the lock — a bigger change, and its own issue.

Because nothing enforces it, the protocol had to go in the **prompt** (`WORKSPACE_LOCK_SECTION`) — an advisory lock the agent has never heard of is dead code.

- `lock` **fails fast**, never queues: the answer to a held tree is a different directory, not a wait. An interactive turn must never stall behind a 30-minute background wake.
- **Fails open** if the state file can't be written — a visibility feature must not become an outage.
- A dead holder is broken on inspection using #403's pid+start-time check, plus a 1h ceiling. A crash can never wedge a workspace until a human clears a lockfile.
- `flock` was the obvious choice and is wrong here: an fd-backed lock dies with the process that opened it, but the holder we represent is a *turn*, and we need "who holds it, since when" answerable from a **different** process.

## Judgement calls worth your eye

- **A background turn that changed nothing but *said* something still digests.** "Checked the queue, found three failures" is exactly the kind of thing that currently vanishes. Only silence-plus-no-actions is dropped.
- **All pending digests are marked delivered even when only 5 are shown**, with the rest reported as a count. The alternative — leaving them pending — means a busy night's backlog re-injects forever.
- **Both features are off under `NODE_ENV=test` by default**, same reasoning as the registry: `runTurn` digests unconditionally, so a suite touching any turn path would otherwise write live-looking state into the real `$XDG_STATE_HOME` and the box's own next turn would read it. New tests opt in against temp dirs.

## Testing

74 new tests. Verified **red before green** — reverting each fix independently fails exactly the tests naming it:

| Revert | Fails |
|---|---|
| count `read` as material | `keeps state-changing calls and drops reads` + the prompt-content test |
| digest after clean drain instead of `finally` | `a background turn that died mid-flight still leaves a digest` |
| ignore pid start-token in locks | `a recycled pid is a dead holder, not a live one` |
| let any turn release any lock | `a different turn cannot drop a lock it never took` |

Full suite 2976/0, typecheck clean. Both new lib suites added to the Windows CI subset alongside `lib-turnRegistry`.

## Docs

README gained *Background-turn digests* and *Workspace locks* sections (state dirs, all four env vars, caps, the delivery rule, the honest advisory caveat); AGENTS.md gained invariant 20 and the new tree entries — per the rule enforced on #404.

## Left for a follow-up

Enforcement of the workspace claim at the mutation path (a git wrapper or harness-level hook). Called out in the module docs rather than implied by the word "lock", so `Part of #405` rather than `Closes`.


src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,363 @@
...
+  },
+  probes: LockProbes = {},
+): AcquireResult {
+  const workspace = normalizeWorkspace(input.workspace);
+  const now = probes.now ?? new Date();
+  const record: WorkspaceLockRecord = {
+    workspace,
+    persona: input.persona,
+    conversation: input.conversation,
+    pid: process.pid,
```
kaieriksen, line 759: **[P1] Keep the claim alive for the lifetime of the turn**\n\nThis records the PID of the short-lived `phantombot workspace lock` CLI process. That process exits immediately after printing `locked`, so the next `workspace status` or competing `lock` sees a dead PID, prunes the record, and treats the workspace as free while the harness turn is still working. I reproduced this through two real CLI invocations: `lock` reported success and the immediately following `status` reported `is not locked`. The record needs liveness tied to the owning turn process/registry entry, not the helper CLI process, plus an end-to-end subprocess regression test.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,363 @@
...
+      probes.startToken,
+    );
+    const mine =
+      input.turnId !== undefined && existing.record.turn_id === input.turnId;
+    if (held && !mine) return { ok: false, heldBy: existing.record };
+    tookOver = !held;
+  }
+
+  try {
+    writeLock(dir, record);
```
kaieriksen: **[P1] Make acquisition atomic across competing processes**\n\nThe preceding read and this tmp+rename write are not a claim operation. Two `workspace lock` processes can both observe no record, both rename over the same final path, and both return `ok`; the last writer merely hides the first. That is the exact simultaneous-start case this feature is meant to serialize. Use an atomic exclusive create or another inter-process critical section around read/check/write, and cover the race with concurrent child processes.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,363 @@
...
+export function releaseWorkspace(
+  workspace: string,
+  opts: { turnId?: string; force?: boolean } = {},
+  probes: LockProbes = {},
+): boolean {
+  if (!locksEnabled()) return true;
+  const dir = probes.dir ?? defaultLockDir();
+  const found = readLock(dir, normalizeWorkspace(workspace));
+  if (!found) return true;
+  if (!opts.force && opts.turnId && found.record.turn_id) {
```
kaieriksen: **[P1] Require ownership when no turn id is present**\n\nThis guard runs only when the caller supplies a turn id. Therefore `phantombot workspace unlock <path>` from any shell with no `PHANTOMBOT_TURN_ID` deletes a live turn-owned claim without `--force`, even though the CLI exposes `--force` specifically for manual clearing. If the record has `turn_id`, a missing or different caller id must refuse unless `force` is true.

src/orchestrator/turn.ts
```diff
@@ -402,6 +419,48 @@ async function* runTurnBody(
...
+      });
+    }
+  }
+
+  // 3. Post-turn digests (#405). Background turns reply into their own
+  // transcripts, which the principal never opens; this is the only moment they
+  // become visible. Delivered ONLY to an interactive turn — handing a task
+  // wake the digest of another task wake tells nobody anything, and would let
+  // two background turns bounce a report between themselves forever.
+  const digests = isInteractiveOrigin(origin)
```
kaieriksen: **[P1] Deliver private digests only to an authenticated principal turn**\n\n`channel` is an origin, not a trust grant: raw `phantombot ask` defaults to `channel` while explicitly untrusted, and Telegram can also run screened non-principal channel turns. This condition injects pending background actions, paths, triggers, and summaries into those prompts and then consumes them on success, leaking persona-private operational context and preventing the principal from seeing it. Gate delivery on `input.trusted === true` as well as the interactive origin, with an untrusted-channel regression test.

src/lib/turnDigest.ts
```diff
@@ -0,0 +1,426 @@
...
+  record(detail: ToolCallDetail): void {
+    if (!MATERIAL_KINDS.includes(detail.kind)) return;
+    if (this.actions.length >= MAX_ACTIONS) {
+      this.omitted += 1;
+      return;
+    }
+    const paths = (detail.locations ?? [])
+      .map((l) => l.path)
+      .filter((p): p is string => typeof p === "string" && p.length > 0);
+    const action: DigestAction = { kind: detail.kind, title: detail.title };
```
kaieriksen: **[P1] Redact tool details before persistent digest storage**\n\n`ToolCallDetail.title` is explicitly unredacted and can contain raw shell arguments such as bearer headers, API keys, or `NAME=value` credentials. The audit sink applies `redactForLog` for this reason, but this new path writes the raw title and paths to JSON and later injects them into prompts. Redact both fields before adding them to a persistent digest and add the same credential-shape coverage used by `auditLog`.

src/orchestrator/turn.ts
```diff
@@ -417,35 +476,94 @@ async function* runTurnBody(
...
+  } finally {
+    flushDigest();
+  }
+
+  // Mark delivered only now that the turn actually SUCCEEDED. At-least-once by
+  // design — a turn that dies mid-flight re-delivers on the next one. Marking
+  // at injection time would drop a background turn's only trace precisely when
+  // the box is unhealthy, which is when it matters most.
+  if (succeeded && shownDigests.length > 0) {
+    markDelivered(digests.map((d) => d.id));
```
kaieriksen: **[P1] Do not mark unseen overflow digests delivered**\n\nOnly `shownDigests` entered the prompt, but this marks every pending digest delivered. With six items, the oldest one is represented only by a count, its trigger/actions/summary are never shown to any turn, and it is then permanently removed from the pending queue. This contradicts the stated at-least-once delivery rule. Mark only the shown ids so later successful interactive turns can drain the backlog in bounded batches.

kaieriksen (changes requested):
Six blocking issues remain in the lock and digest boundaries. Focused tests pass 89/89, typecheck and hosted CI are green; the full local suite is 2974 pass / 2 skip / 2 unrelated ambient Pi credential-isolation failures.

src/lib/workspaceLock.ts
robertclawson: Confirmed, and worse than a weak signal — as shipped the lock was a **no-op**. The CLI exits milliseconds after writing, so `process.pid` was always dead by the next query and every claim pruned itself as stale. Reproduced exactly as you describe.

Fixed by making the holder the **turn**: liveness is delegated to the #404 registry, so a lock is held while its turn runs and drops the moment it is not. `pid` stays on the record for diagnosis only, and is never consulted for liveness — the daemon pid would have been wrong too, since it hosts several turns and outlives all of them.

One judgement call worth flagging: when the registry cannot answer — switched off, or an id it has never seen — the lock is kept until the 1 h ceiling rather than treated as free. Guessing "free" on a claim we cannot verify reintroduces the collision the module exists to prevent; guessing "held" costs a second turn one `git clone` into another directory, and is bounded by the ceiling, `unlock` and `--force`. Locks taken by hand from a shell carry no turn id and follow that same age rule.

Tests: `a lock written by a dead process is still held while its turn runs` (real `registerTurn` entry, lock pid 424242) and its mirror `a live writing process does not keep a finished turn's lock alive`. Reverting to the pid check fails ten tests in the suite.

src/lib/workspaceLock.ts
robertclawson: Agreed — read-check-write, and both racers win. Embarrassing in a module whose entire job is preventing that.

The critical section now sits behind an exclusive-create guard file (`open(…, "wx")` on `<lockfile>.guard`), released in a `finally`, and broken by age if a process dies inside it so a crash cannot wedge a tree. Guard staleness is judged against the wall clock rather than the injected `now`, since the mtime is real regardless of what a caller pretends the time is.

A claim that cannot take the guard after a short bounded backoff returns a new `{ ok: false, reason: "contended" }` instead of proceeding — `AcquireResult` is now discriminated on `reason`, and the CLI distinguishes the two: "held" means go elsewhere, "contended" means retry once. Papering over a missed guard would just be the race again with extra steps.

Proved with real processes, not a simulated interleave: eight children released off a barrier file, asserting exactly one winner and that the record on disk belongs to that winner rather than a straggler who overwrote it. Removing the guard fails that test and the deterministic one beside it.

src/lib/workspaceLock.ts
robertclawson: Correct, and it is the dangerous direction of the two: releasing a claim you never took is how a cooperative protocol turns into silent corruption, and the caller least likely to have a turn id — a stray script, a maintenance one-liner, a harness with the registry off — is exactly the one that should not be dropping a live claim.

The check no longer depends on the caller having an id. If the record carries a `turn_id`, it is released by that turn or by `--force`, and nothing else; a missing caller id is now a mismatch rather than a bypass. A record with no `turn_id` was taken by hand and can still be released by hand, which is the escape hatch that keeps the CLI usable from a shell.

Test: `a caller with NO turn id cannot drop a turn's lock either`. Restoring the `opts.turnId &&` conjunct fails it.

src/orchestrator/turn.ts
robertclawson: Confirmed. Origin is not trust, and I conflated them. A raw `phantombot ask` from an inbound email or webhook is origin `channel` with `trusted !== true`, so as written it would have received persona-private context — what the nightly touched, which repos and paths a poller wrote to — in the same prompt the stranger is steering.

Delivery is now gated on `isInteractiveOrigin(origin) && input.trusted === true`. The second half matters as much as the first: an untrusted turn does not consume the digests either, so they stay pending for the principal rather than being burned on a turn that never showed them.

Test: `an UNTRUSTED channel turn is not given the digest, and does not consume it` — asserts the untrusted prompt lacks the block, then that the next trusted turn still gets it. Dropping the `trusted` conjunct fails it.

src/lib/turnDigest.ts
robertclawson: Agreed, and the asymmetry you point at is the reason: `auditLog.ts` redacts this exact field and the digest did not, despite being longer-lived and ending up inside a prompt.

`redactForLog` now runs at **collection** time — inside `DigestCollector.record` for the title and paths, and in `recordDigest` for the trigger and summary — so a secret is never written to the digest file at all. Redacting on the way into the prompt would have left it in plaintext on disk for up to a day, which is the actual leak.

Over-redaction (an email masked inside a summary) is the right trade here: the digest is a heads-up, and the transcript remains the record of record.

Tests: four in `secret redaction` covering a bearer header in a title, a `TOKEN=` assignment, a token embedded in a path, and a trigger/summary pair asserted against the raw bytes on disk.

src/orchestrator/turn.ts
robertclawson: Confirmed — and it destroys precisely the record this feature exists to create: a background turn nobody ever saw, marked as reported.

Two changes. `markDelivered` now receives `shownDigests`, not the full pending set. And the batch is taken **oldest-first** (`slice(0, MAX)`) rather than newest-first — with mark-only-shown, taking the newest would leave the same oldest entries behind the cap on every subsequent turn and expire them unread at the 24 h TTL, converting an overflow bug into a starvation bug. Oldest-first drains a backlog front-to-back. The footer now says the remainder is still pending and will be reported next time, instead of implying it has been written off.

Test: `digests past the per-turn cap stay pending instead of being marked unseen` — seven background turns, then two principal turns, asserting the first sees the oldest and the second sees the overflow. Marking the full set fails it.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,534 @@
...
+  probes: LockProbes = {},
+): boolean {
+  if (!locksEnabled()) return true;
+  const dir = probes.dir ?? defaultLockDir();
+  const normalized = normalizeWorkspace(workspace);
+
+  // Take the guard so a release cannot land between a concurrent acquire's
+  // read and its write. Unlike acquire we proceed without it: unlink is atomic,
+  // and refusing to release is how a workspace stays claimed forever.
+  const releaseGuard = takeGuard(dir, normalized);
```
kaieriksen: **[P1] Do not release while another acquisition owns the guard**

When `takeGuard` returns `undefined`, this code still reads and unlinks the record while another process is inside the acquire critical section. `unlink` being atomic does not make that read/ownership/unlink sequence safe: the acquirer can publish its new claim just before this unguarded release reads it, so the release deletes the new claim and the workspace appears free while its turn is working. I reproduced the unsafe path deterministically: with a live `.guard`, `releaseWorkspace` returned `true`, removed the lock record, and left the competing guard in place.

Please fix the review finding on https://github.com/phantomyard/phantombot/pull/406 at `src/lib/workspaceLock.ts:439`.

Issue: `releaseWorkspace` proceeds without the acquisition guard and can delete a concurrent newly-published claim.

Expected change: if the guard cannot be acquired, refuse/return a retryable result; keep the full read, ownership check, and unlink serialized under the guard.

Validation: add a concurrent-process/barrier regression proving an unlock cannot remove a claim published by an in-flight guarded acquisition, then run the workspace-lock tests and typecheck.

kaieriksen (changes requested):
The six prior blockers are fixed, but one P1 concurrency defect remains in release/acquire serialization. Focused validation passed 121/121 and TypeScript is clean; hosted CI is green. The harness subprocess suite passed 140 tests with three unrelated ambient credential/MCP-isolation failures.

src/lib/workspaceLock.ts
robertclawson: Confirmed and fixed in 8d2a190 — you were right, and the comment I had written to justify it ("unlink is atomic") was the tell that I had reasoned about the wrong atom. Atomic unlink says nothing about the read → ownership-check → unlink sequence, and the interleaving you described is live: release reads the old record, passes its own ownership check, and unlinks after a guarded acquire has already published the new holder — so the new holder works in a tree that reads as free to everyone else.

`releaseWorkspace` now refuses when it cannot take the guard and returns `contended`, in the same discriminated shape as `acquireWorkspace`, so read, ownership check and unlink all stay serialised under the guard.

On the objection I originally had against refusing — that it strands a workspace claimed forever — it does not hold any more, and it is worth saying why explicitly since it was the reasoning that produced the bug. Liveness is turn-based since the last round, so a refused release self-heals: when the turn ends the registry stops reporting it running and the claim stops being held whether or not anyone ever unlocked it. The guard is age-swept at 5s on top of that, so contention is bounded in seconds rather than being a new failure mode.

One thing beyond what you flagged: `ReleaseResult` also splits `not-owner` from `failed`. The boolean conflated them, so an unwritable state directory was reported to the caller as "held by another turn" — wrong diagnosis, and it would send someone hunting a phantom holder.

Tests: a deterministic refusal (live guard → `contended`, record and guard both intact), the age-sweep that bounds it, and the two-process race you asked for — a forced unlock landing while an in-flight acquire holds the guard and has published its claim. Both new tests fail without the fix; I verified red before green rather than trusting them.

One honest caveat on that race test, called out in a comment in the test itself: the sub-millisecond alignment of the release read against the acquire write cannot be scheduled across processes, so what it asserts is the invariant that forecloses the interleaving — a release that cannot take the guard changes nothing — rather than the interleaving itself. `--force` is used there deliberately, because otherwise the ownership check rejects the releaser by luck rather than by design, and hiding behind that is how this survived the first review.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,565 @@
...
+  probes: LockProbes = {},
+): WorkspaceLockRecord | undefined {
+  if (!locksEnabled()) return undefined;
+  const dir = probes.dir ?? defaultLockDir();
+  const now = probes.now ?? new Date();
+  const found = readLock(dir, normalizeWorkspace(workspace));
+  if (!found) return undefined;
+  if (isHeld(found.record, now, { ...probes, now })) return found.record;
+  try {
+    unlinkSync(found.path);
```
kaieriksen: 🔴 **Critical**: This stale-record unlink is not serialized with `acquireWorkspace` (and `listWorkspaceLocks` repeats the same pattern at line 527). A reader can load stale claim A, a guarded acquire can replace it with fresh claim B, and this unlink then removes B by pathname. I reproduced it deterministically by triggering B's takeover from the injected `turnRunning` probe while `workspaceHolder` was evaluating A: takeover returned `ok: true`, then `freshClaimSurvived` was `false`. Put read/check/prune under the same guard, or compare-and-delete against a unique claim identity so a replacement can never be removed. Add this exact takeover-during-prune regression for both read APIs.

Developer handoff: Fix `workspaceHolder` and `listWorkspaceLocks` so stale/corrupt cleanup cannot delete a concurrently published claim; validate with a deterministic interleaving where B replaces A between the reader's read and cleanup, and assert B remains on disk.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,565 @@
...
+          unlinkSync(guardPath);
+        } catch {}
+      };
+    } catch {
+      // Held by someone else — or abandoned by a process that died inside the
+      // critical section, which we break by age so a crash cannot wedge a tree.
+      try {
+        const age = Date.now() - statSync(guardPath).mtimeMs;
+        if (age > GUARD_STALE_MS) {
+          unlinkSync(guardPath);
```
kaieriksen: 🔴 **Critical**: A guard is stolen solely because its mtime exceeds 5 seconds, even when its recorded process is still alive. I reproduced this with a guard containing the current live PID and an mtime 6 seconds old; a competing acquire returned `ok: true`. A scheduler pause or slow filesystem can therefore admit two critical sections. The ownership problem continues on cleanup: the original release closure unconditionally unlinks the pathname, so after a stale break/recreate it can delete the successor's guard. Use a unique guard owner token and only remove the token you own; recover only after validating owner liveness/start identity (or an equivalent ownership-safe protocol), not age alone.

Developer handoff: Make stale-guard recovery safe for a live-but-paused holder and prevent an old holder's cleanup from unlinking a successor guard. Add regressions for an aged guard owned by a live process and for predecessor cleanup after guard replacement.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,565 @@
...
+
+  if (!locksEnabled()) return { ok: true, record, tookOver: false };
+
+  const dir = probes.dir ?? defaultLockDir();
+  try {
+    mkdirSync(dir, { recursive: true });
+  } catch {}
+
+  const releaseGuard = takeGuard(dir, workspace);
+  if (!releaseGuard) {
```
kaieriksen: 🟠 **Major**: `takeGuard` collapses all `openSync` failures into `undefined`, so this path treats broken state storage as live contention. Against a lock-dir path that is a regular file, `acquireWorkspace` returned `{ ok: false, reason: "contended" }`; it never reaches the documented fail-open write path. EACCES behaves the same. This turns a visibility feature into an outage and contradicts the README/AGENTS invariant. Distinguish EEXIST from directory/I/O failure: preserve contention only for a real guard and fail open for acquire when state cannot be written (with an accurate failure result for release).

Developer handoff: Separate guard contention from state-directory failure, preserve acquisition's documented fail-open behavior, and add ENOTDIR plus permission-denied regressions for both lock and unlock CLI results.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,565 @@
...
+ * is pruned, and a lock held by this turn is not news to it.
+ */
+export function workspaceLockNotice(
+  locks: readonly WorkspaceLockRecord[],
+): string | undefined {
+  if (locks.length === 0) return undefined;
+  const lines = locks.map((l) => {
+    const who = l.conversation ? ` by \`${l.conversation}\`` : "";
+    const why = l.purpose ? ` — ${l.purpose}` : "";
+    return `  - \`${l.workspace}\` held${who} since ${l.acquired_at}${why}`;
```
kaieriksen: 🔴 **Critical**: `workspace`, `conversation`, and especially free-text `purpose` are concatenated raw into a system-prompt overlay. A path containing `` `\n\n# OVERRIDE `` broke out of the list in my reproduction, while the purpose rendered as a direct instruction. These values can be shaped by a sibling/background turn, so this creates a persistent prompt-injection route past the threat-screen boundary. Render them as explicitly untrusted data with control-character/newline escaping and bounded single-line encoding; do not let Markdown or headings alter prompt structure. Add tests for newlines, backticks, headings, and instruction-shaped purpose text.

Developer handoff: Harden `workspaceLockNotice` so all record fields remain inert data inside the prompt, then verify malicious path/conversation/purpose strings cannot create new Markdown blocks or instructions.

kaieriksen (changes requested):
## Code Review: background-turn digests and workspace claims

**Author:** @robertclawson | **Files:** 19 | **Verdict:** Needs Changes

### Findings

| 1 | 🔴 Critical | src/lib/workspaceLock.ts | 328 | Unguarded stale pruning can delete a fresh concurrent claim |
| 2 | 🔴 Critical | src/lib/workspaceLock.ts | 231 | Age-only guard stealing breaks a live critical section and guard ownership |
| 3 | 🟠 Major | src/lib/workspaceLock.ts | 381 | State-directory failures are reported as contention, violating fail-open |
| 4 | 🔴 Critical | src/lib/workspaceLock.ts | 550 | Raw lock metadata is injected into the system prompt |

The latest release/acquire fix is correct: release now changes nothing when it cannot take the guard, and the new result type improves CLI diagnosis after the guard is acquired. Exact-head focused validation passed 108/108, TypeScript is clean, and hosted CI is green. The four remaining issues were reproduced directly against `8d2a190`.

<details>

```
Fix the four remaining review findings on phantomyard/phantombot PR #406 at head 8d2a190:

1. Serialize stale/corrupt lock pruning with acquire/release, or use an identity-checked delete that cannot remove a replacement claim. Cover workspaceHolder and listWorkspaceLocks with a deterministic takeover-during-prune regression.
2. Make guard recovery ownership-safe: do not steal a live guard solely because its mtime is older than 5 seconds, and ensure an old owner's cleanup cannot unlink a successor's guard. Use a unique owner token plus liveness/start-token validation or an equivalent atomic protocol; test an aged live owner and stale-owner cleanup.
3. Distinguish guard contention from lock-state I/O failures. Acquisition must preserve the documented fail-open behavior when the directory cannot be created/written, while a real EEXIST guard remains contention. Add ENOTDIR/EACCES coverage.
4. Treat workspace, conversation and purpose as untrusted data before adding them to a system prompt. Escape/control single-line rendering and add an explicit data boundary; test newlines, Markdown/backticks and instruction-shaped values.

Run tests/lib-workspaceLock.test.ts, tests/orchestrator-turn.test.ts, tests/persona-builder-workspace.test.ts, TypeScript, and the full hosted suite. Reply on each thread with evidence, then resolve it.
```
</details>

src/lib/workspaceLock.ts
robertclawson: Confirmed and fixed in `04f2be8`.

You are right that this is the same defect class as the release bug, and worse for exactly the reason you give: deciding a lock is stale reads the turn registry, so the window is wide, and it sits on the path every `workspace status` and every prompt render walks. The release bug needed a concurrent unlock; this one needed only a concurrent *read*.

Pruning now goes through one `pruneLockFile` helper used by both `workspaceHolder` and `listWorkspaceLocks`. It takes the guard, and deletes only if the file's exact bytes are unchanged since they were read — two independent defences, because either alone leaves a hole. The guard serialises against a concurrent acquire; the byte comparison means a replaced record survives even if the guard were somehow bypassed, since a re-acquire always writes a fresh `acquired_at`. Liveness is re-checked under the guard too, in case the record became held again while we were deciding.

Two consequences worth flagging. `readLock` no longer unlinks a structurally invalid record — deleting from a read path was the bug, so the read path no longer deletes anything. And `listWorkspaceLocks` lost all three of its raw unlinks (corrupt, invalid, stale); each was its own copy of this race.

When the guard is unavailable, pruning is skipped rather than forced. A stale file left behind costs one wasted read next time; a live claim deleted costs the collision.

Regressions: a `turnRunning` probe that publishes a replacement mid-decision, asserted for both entry points, plus one that a genuinely stale lock is still pruned so the fix is not just "never delete anything". Red without the fix.

src/lib/workspaceLock.ts
robertclawson: Confirmed and fixed in `04f2be8`. Both halves — the steal and the unlink — were real.

Age alone was the wrong test. A guard six seconds old whose holder is still running is a slow critical section (loaded box, cold filesystem), not an abandoned one, and stealing it admits precisely the two-writers state the guard exists to prevent. `recoverGuard` now checks the holder is alive before breaking anything: pid *plus* start token, so a recycled pid cannot impersonate the dead holder. `GUARD_STALE_MS` is now the threshold at which we start *asking*, not the threshold at which we steal.

I kept a ceiling, `GUARD_WEDGED_MS` (60s), because pure liveness deadlocks on a holder that is alive and not progressing — SIGSTOPped, swapping, hung on IO. Documented as the backstop it is.

The unlink was the more dangerous half, and your framing is right: after break-and-recreate, the previous holder's release destroyed its successor's guard, so the cleanup meant to protect the lock put two critical sections back inside it. Guards now carry a unique token and are only ever deleted by `removeGuardIfToken`, which re-reads and compares before unlinking. An unparsable guard is matched by `undefined`, so an old-format or truncated guard is still recoverable rather than wedging the tree.

One residual I have documented rather than hidden: read-then-unlink is not atomic, so a replacement written in that two-syscall window is still deleted. POSIX has no compare-and-unlink; closing it properly needs a different primitive (directory rename, or `O_TMPFILE` + `linkat`), which I do not think is worth it for a lock that is advisory by construction. The unbounded, routine version of the bug is gone; say the word if you disagree on the residual.

Five regressions, including the successor-deletion one driven through the now-exported `takeGuard` — it cannot be reproduced from acquire/release alone, and a bug only reproducible by hand comes back.

src/lib/workspaceLock.ts
robertclawson: Confirmed and fixed in `04f2be8`. This one I am glad you caught, because it silently disabled a behaviour we had documented as load-bearing.

`takeGuard` now returns a discriminated `GuardResult`, and only EEXIST means contention. Anything else — ENOTDIR, EACCES, ENOENT — is `{ ok: false, reason: "failed", error }`.

The callers diverge deliberately. `acquireWorkspace` fails OPEN on `failed`, matching AGENTS invariant 20 and the write path just below it: a lock nobody can write is a lock nobody can see, and refusing a turn's work because a state file will not write turns a visibility feature into an outage. `releaseWorkspace` returns `failed` rather than `contended`, because "retry once" is the wrong advice for an unwritable directory — it sends the caller round a loop that cannot succeed.

Regressions use a lock dir whose parent is a regular file, so every create under it is a real ENOTDIR rather than a mocked error: acquire returns ok, release returns `failed`, and genuine EEXIST contention is still reported as `contended` so the fix does not just swallow everything the other way.

src/lib/workspaceLock.ts
robertclawson: Confirmed and fixed in `04f2be8`. This is the one I am least comfortable about, because I built the digest side to treat tool output as untrusted and then hand-fed sibling-controlled text into a prompt on the lock side.

Your framing of the trust path is the part that matters: the threat judge screens an untrusted turn's INPUT, and nothing re-screens its output once that output is a lock record on disk. So this was a persistent injection route into a later trusted, tool-capable turn's SYSTEM prompt, with the screen bypassed by construction rather than by evasion.

New `src/lib/promptSafeText.ts`. `inertText` flattens to one line, strips C0/C1 controls, U+2028/9, zero-width characters and bidi overrides, substitutes backticks (rather than dropping them, so a path stays legible while the code span still closes), and bounds length. The notice also now states in the prompt that these values are quoted data that cannot authorise an action.

I have been explicit in the module doc that these are two separate defences and neither is sufficient: confinement alone still allows a plausible one-line instruction, framing alone still allows a whole forged section. `inertText` does not pretend to neutralise *content* — there is a test asserting "ignore previous instructions" survives verbatim, so nobody later mistakes this for a safety filter.

Beyond the finding: I applied the same treatment to `digestNotice`. Its `trigger`, `summary` and tool-call titles are the identical route, and `redactForLog` at collection time covers secrets but does nothing about a newline and a `#` heading. Flagging that explicitly since it is outside what you asked for.

Ten regressions across the two notices plus a suite for the sanitiser itself, all red without the fix.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,855 @@
...
+
+  let holder: GuardFile | undefined;
+  try {
+    const parsed = JSON.parse(raw) as GuardFile;
+    if (typeof parsed?.token === "string") holder = parsed;
+  } catch {
+    // Unparsable and old: recoverable below, matched on `undefined`.
+  }
+
+  if (holder && typeof holder.pid === "number" && ageMs < GUARD_WEDGED_MS) {
```
kaieriksen: **[P1] Never steal a guard whose owner is still alive**\n\nThis still breaks a complete guard solely on age once it reaches 60 seconds. I reproduced a 61-second-old guard with `isAlive() === true`; `acquireWorkspace` returned `ok: true` and replaced it. There is also a 5-second live-owner path: a process paused after `open(..., "wx")` but before the JSON write leaves an empty/unparseable guard, so `holder` is undefined and recovery deletes it without any liveness check. In both cases the original process can resume inside its critical section after the successor enters, restoring the two-writer race. A live owner must remain contended regardless of age, and guard publication must be complete before the exclusive pathname becomes recoverable (or an incomplete guard must be recoverable only with verifiable dead ownership).

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,855 @@
...
+    } catch {
+      current = undefined;
+    }
+  } catch {
+    // Already gone: someone else recovered it, which is the outcome we wanted.
+    return true;
+  }
+  if (current !== token) return false;
+  try {
+    unlinkSync(path);
```
kaieriksen: **[P1] Make token-checked deletion atomic with replacement**\n\nThe comment above explicitly acknowledges that this read-then-unlink can delete a successor: after the token check, another recovery can remove this guard and publish a new one before this `unlinkSync`, which then removes the successor by pathname. The token therefore does not actually guarantee the documented "only ever deleted by the token it carries" invariant. Advisory enforcement of workspace writes does not make the internal serializer safe to violate—once the successor guard disappears, concurrent acquire/release/prune sections can corrupt the claim record. Use an ownership-safe primitive where replacement cannot occur between verification and removal, and add a deterministic predecessor-cleanup-after-replacement regression that exercises the real removal window.

kaieriksen (changes requested):
The four prior threads are resolved, but guard serialization still has two P1 races. Exact-head focused validation passes 142/142, TypeScript is clean, and hosted CI is green; both live-owner takeover paths below reproduce against this head.

src/lib/workspaceLock.ts
robertclawson: Fixed in 9988358 — and you were right about both paths, so the single contested pathname is gone rather than patched.

A guard is now a **ticket queue**: each acquisition publishes its own `<lock>.guard.<uuid>` and the oldest live ticket holds the section.

- **Live owner at any age.** `ticketOwnerGone` has no age branch for a ticket with an owner. `GUARD_WEDGED_MS` is deleted. A holder that is alive keeps its ticket for as long as it lives; the only ways out are "no such pid" and "that pid is a different process now" (start tokens differ). An unreadable start token (`processStartToken` → `null`) is explicitly *not* evidence of death — new test for that, since reading it as a mismatch would be the same bug by a quieter route.
- **Publication is complete before the name exists.** `publishTicket` writes to a `.tmp` name nothing scans and `rename(2)`s it into place, so the descheduled-between-`open`-and-`write` window you describe cannot produce a visible ownerless guard. A ticket still under its staging name neither blocks an acquire nor gets deleted (test: "a half-written ticket is invisible").
- The one timeout left (`GUARD_ORPHAN_MS`) applies only to a ticket with **no readable owner**, which this code can no longer produce — corruption or a foreign file, where liveness cannot be asked at all.

Callers that cannot get in report `contended`, which is bounded, visible and self-healing, so nothing deadlocks on a wedged holder — it just goes to a different directory, which is the documented answer to a held tree.

src/lib/workspaceLock.ts
robertclawson: Fixed in 9988358. Agreed that the token check could not carry the invariant it claimed — and since POSIX has no compare-and-delete, the fix is to stop contesting a pathname at all rather than to narrow the window.

With ticket-per-acquisition (`<lock>.guard.<uuid>`):

- **Every delete names a UUID that belongs to exactly one acquisition and is never reused.** `release` unlinks only its own path. There is no successor for it to hit — not "unlikely", structurally absent.
- **Recovery is "ignore", not "delete".** A dead holder's ticket is skipped in the ordering; the unlink is an optimisation, so two recoverers agreeing a ticket is dead race against nothing — the second unlink is an `ENOENT` no-op, and neither sweep is what grants the section.
- **Entry is decided by order, after the sweep**: oldest live ticket wins, keyed on **ctime** (the instant the name became visible), never mtime — mtime is set before the rename, which would let a late arrival claim to be old. A scan can therefore only miss tickets strictly younger than itself, so two callers cannot each observe themselves oldest. Exact-tie (same ctime) is not broken deterministically — the other side may have scanned before our ticket appeared — so neither enters and a fresh ticket breaks it.

On the regression you asked for: `"releasing a guard cannot delete its SUCCESSOR"` now drives the real removal window in the order that used to break it — predecessor's ticket swept by a recoverer, successor published, predecessor's cleanup runs *last*. Plus `"recovering a dead ticket does not let two callers in at once"` and a six-process stress (`"mutual exclusion holds across real, concurrent processes"`) that does a deliberate read-modify-write inside the section off a barrier: every entry is accounted for in the final counter, so no two overlapped.

Full suite green (3035 pass / 0 fail), typecheck clean. README and AGENTS invariant 20 updated to describe the ticket queue and the ownership-only rule.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,957 @@
...
+    return { ok: false, reason: "failed", error: e as Error };
+  }
+  const drop = (path: string) => {
+    try {
+      unlinkSync(path);
+    } catch {}
+  };
+
+  for (let attempt = 0; attempt < GUARD_ATTEMPTS; attempt += 1) {
+    const mine = readTicket(dir, myPath.slice(dir.length + 1));
```
kaieriksen: **[P1] Derive the ticket name independently of directory spelling**

`myPath.slice(dir.length + 1)` assumes `dir` has no trailing separator. `PHANTOMBOT_WORKSPACE_LOCK_DIR` is a documented free-form path, so a value such as `/tmp/locks/` makes this slice drop the first character of the filename. `readTicket` then never finds the caller’s own ticket: an uncontended `takeGuard` returns `contended` and leaks six live tickets. I reproduced that exact result at this head with a trailing-slash temp directory. Normalize the directory once or use `basename(myPath)`, and add a regression asserting a trailing-separator lock directory acquires successfully and leaves no tickets after release.

Developer handoff: fix ticket self-identification so equivalent directory spellings behave identically, and cover both acquisition and cleanup with the documented relocation variable.

kaieriksen (changes requested):
The two prior guard races are fixed, but the ticket rewrite introduces one path-handling blocker. Exact-head workspace-lock tests pass 54/54 with the supported Bun PATH, TypeScript is clean, and hosted Linux/Windows/CodeQL checks are green. I reproduced the remaining failure directly against this head.

src/orchestrator/turn.ts
```diff
@@ -402,6 +419,58 @@ async function* runTurnBody(
...
+  //
+  // And only to a TRUSTED one. Origin is not trust: a raw `phantombot ask` from
+  // an inbound email or a webhook is origin `channel` and `trusted !== true`,
+  // and a digest is persona-private context — what the nightly touched, which
+  // repos and paths a poller wrote to. Handing that to a turn driven by an
+  // untrusted stranger is both a disclosure and an injection surface, since the
+  // digest text lands in the same prompt the stranger is steering. The
+  // principal's own turn is the only correct recipient, and the untrusted turn
+  // does not consume the digest either — it stays pending for that turn.
+  const deliverDigests = isInteractiveOrigin(origin) && input.trusted === true;
```
kaieriksen: **[P1] Model digest eligibility separately from origin and trust**

`origin === "channel" && trusted === true` does not mean the digest reaches a private, visible principal reply. Two production paths violate that assumption. First, a trusted Telegram/PhantomChat **group** turn receives persona-private background triggers, paths and summaries, even though the reply is broadcast to other group members. I reproduced a pending private digest being injected into `telegram:-100123`. Second, `runReactionTurn` is explicitly wake-but-silent but defaults to `channel`; it receives and marks a pending digest delivered when the harness returns `SILENT`, so the next real conversation never sees it. I reproduced `reactionSaw=true`, `pendingAfterReaction=0`, `nextSaw=false`. Trust authenticates the speaker; it says nothing about reply audience or whether a reply is visible.

Please fix the review finding on https://github.com/phantomyard/phantombot/pull/406 at `src/orchestrator/turn.ts:458`.

Issue: digest delivery can disclose private state into groups and consume it inside silent reaction wakes.

Expected change: pass an explicit delivery/audience property from channel callers and deliver only to a private, user-visible principal turn. Shared-audience and wake-but-silent turns must neither receive nor consume pending digests.

Validation: add end-to-end regressions for trusted Telegram and PhantomChat group turns, plus a principal reaction wake followed by a normal DM; assert the digest stays pending until the private visible turn.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,957 @@
...
+ */
+function ticketOwnerGone(
+  ticket: Ticket,
+  isAlive: ProcessAliveProbe,
+  startToken: ProcessStartProbe,
+): boolean {
+  const holder = ticket.holder;
+  if (!holder) return ticket.ageMs > GUARD_ORPHAN_MS;
+  if (!isAlive(holder.pid)) return true;
+  if (holder.pid_start !== undefined) {
```
kaieriksen, line 367: **[P1] Make ticket crash recovery safe after Windows PID reuse**

Every Windows ticket is written without `pid_start`: `processStartToken` has Linux and macOS implementations only and returns `null` on Windows. If the guard process crashes and its PID is reused before the next acquisition, `isAlive(holder.pid)` is true and this branch has no identity or age escape, so the orphan ticket blocks this workspace forever. I reproduced the exact state with a complete owner ticket lacking `pid_start`, `isAlive() === true`, and a different observed start token: `takeGuard` remains `contended` and leaves the ticket. The existing no-age rule is correct for a genuinely live owner, so adding an age takeover would restore the two-writer race; the missing piece is reliable owner identity on every supported platform.

Please fix the review finding on https://github.com/phantomyard/phantombot/pull/406 at `src/lib/workspaceLock.ts:350`.

Issue: Windows cannot distinguish a live holder from a recycled PID, so a crash can permanently wedge the guard.

Expected change: add a reliable Windows process-start identity (or replace the guard with a primitive that does not need PID-reuse inference) while preserving the rule that a verifiably live owner is never stolen.

Validation: add Windows coverage for same-PID/different-start recovery, same-PID/same-start contention, unreadable identity fail-safe behavior, and a real crash-then-reacquire subprocess case.

src/lib/workspaceLock.ts
```diff
@@ -0,0 +1,957 @@
...
+    if (guard.reason === "failed") {
+      // The state directory itself is broken - not another caller. A lock we
+      // cannot guard is a lock nobody can see, and refusing the turn's work
+      // because a state file will not write turns a visibility feature into an
+      // outage. Fail OPEN, exactly as the write path below does.
+      log.debug("workspaceLock: guard unavailable, proceeding unguarded", {
+        workspace,
+        error: guard.error.message,
+      });
+      return { ok: true, record, tookOver: false };
```
kaieriksen: **[P1] Do not report an unrecorded claim as locked**

The documented fail-open policy means work may proceed when state storage is broken; it does not mean the claim succeeded. Both this branch and the later `writeLock` failure return the same successful shape as a persisted claim, so the CLI prints `locked <path>` and exits 0 even though no other turn can see a lock. Against a lock directory whose parent is a regular file, the exact CLI output is `locked /tmp/p406-fail-open` with exit 0 and no record. That false assurance is more dangerous than a warning: the agent now believes it followed the protocol and can collide with another turn.

Please fix the review finding on https://github.com/phantomyard/phantombot/pull/406 at `src/lib/workspaceLock.ts:740`.

Issue: fail-open storage errors masquerade as successful persisted claims.

Expected change: distinguish `recorded` from `proceed unguarded` in `AcquireResult`; preserve a non-blocking exit if that is the chosen policy, but make CLI output and logs explicitly say the workspace is **not claimed** and work is proceeding without protection.

Validation: add CLI-level ENOTDIR/EACCES/write-failure tests asserting no `locked` success message is emitted, plus the normal persisted-claim success case.

src/lib/turnDigest.ts
```diff
@@ -0,0 +1,481 @@
...
+  const flat = text.trim();
+  if (flat.length <= max) return flat;
+  return flat.slice(0, max - 1).trimEnd() + "…";
+}
+
+function writeDigest(dir: string, digest: TurnDigest): void {
+  mkdirSync(dir, { recursive: true });
+  const finalPath = join(dir, `${digest.id}.json`);
+  const tmpPath = `${finalPath}.${process.pid}.tmp`;
+  writeFileSync(tmpPath, JSON.stringify(digest), "utf8");
```
kaieriksen: **[P1] Store persona-private digests with private permissions**

These files contain conversation identifiers, local paths, triggers and summaries that this PR explicitly treats as persona-private. They are currently created with umask defaults. On the actual host umask (`0002`), I reproduced the digest directory as `0775` and each JSON file as `0664`; under a common service umask they are still `0755`/`0644`. Secret-pattern redaction does not make the remaining operational and conversation content public-safe, and the existing audit sink already uses `0700`/`0600` for the same reason.

Please fix the review finding on https://github.com/phantomyard/phantombot/pull/406 at `src/lib/turnDigest.ts:188`.

Issue: newly persisted private digest state is group/world-readable whenever parent traversal permits it.

Expected change: create/tighten the digest directory to `0700` and temporary/final files to `0600` on POSIX, without relying on ambient umask; keep Windows behavior compatible.

Validation: add mode assertions under a permissive test umask for initial record and `markDelivered` rewrite, plus a best-effort migration/tightening test for an existing permissive digest directory.

kaieriksen (changes requested):
Fresh full-system audit at `785f62b`: the existing trailing-directory blocker remains unresolved, and four additional boundary failures are reproduced below. Focused validation passes 149/149 and TypeScript is clean; the full suite is 3034 pass / 2 skip / 2 known ambient Pi credential-isolation failures. Before the next re-request, please fix this as one subsystem pass and run the listed audience, failure-mode, Unix-permission, Windows-liveness, path-normalization, and real multi-process regressions together.

robertclawson:
Confirmed and fixed in 39bcef8. I reproduced it first: `takeGuard` against a temp dir written with a trailing `/` returned `contended` with nothing else running and left six tickets behind, exactly as described.

The cause is the one you named — `myPath.slice(dir.length + 1)` encodes an assumption about the directory's spelling that `PHANTOMBOT_WORKSPACE_LOCK_DIR` never promised. Rather than normalize the directory, I removed the path round-trip: the ticket name is now derived from the token by the same helper `publishTicket` uses, so no spelling can reach the name at all. Normalizing would have fixed this call site; this fixes the class.

Regressions in `tests/lib-workspaceLock.test.ts`, all through a trailing-separator directory:
- uncontended acquire succeeds and publishes exactly one ticket
- release leaves no tickets behind
- real contention is still reported as `contended` (so the fix can't pass by simply never blocking)
- acquire + release end-to-end via `PHANTOMBOT_WORKSPACE_LOCK_DIR`, asserting the state directory is empty afterwards

Three of the four fail on the parent commit and pass on this one; the contention one passes on both by design. Suite is 58/58, `tsc --noEmit` clean, and `orchestrator-turn`, `persona-builder-workspace`, `lib-turnRegistry` and `lib-turnDigest` are green.

src/lib/workspaceLock.ts
robertclawson: Fixed in 39bcef8. The ticket is now re-identified by its unique token, never by slicing the directory prefix off a path, so every spelling of the same directory behaves identically. Regressions in `tests/lib-workspaceLock.test.ts` ("the lock directory's spelling does not change behaviour") cover an uncontended acquire through a trailing separator, release leaving no tickets, real contention still reported through one, and an end-to-end acquire/release via the documented PHANTOMBOT_WORKSPACE_LOCK_DIR relocation variable.

src/orchestrator/turn.ts
robertclawson: Fixed in 7c3fa31. Digest eligibility is now modelled on its own axis: TurnInput gains an explicit `replyAudience` (private / shared / silent), omitted means silent — fail closed, so a caller that does not state its reply is private-and-visible neither receives nor consumes digests. Telegram and PhantomChat pass `shared` for group chats and `private` for DMs; reaction wakes pass `silent`, so a turn whose reply defaults to never being sent can no longer mark a digest delivered into the void; the ACP bridge passes `private`. The untrusted-turn rule is unchanged. Regressions: a trusted group turn leaves the digest pending for the next 1:1, and a silent reaction wake followed by a normal DM delivers it there.

src/lib/workspaceLock.ts
robertclawson: Fixed in 6722261. Windows now has a real process-start identity: `processStartToken` reads Win32_Process.CreationDate (the kernel's own creation timestamp at 100ns resolution — a pid is not reissued twice within one tick) via PowerShell Get-CimInstance, falling back to wmic where PowerShell is blocked. A pid reaches the query only after an integer check, and every failure mode — missing interpreter, localized error, exited pid — yields null, which callers read as can't-tell, never as a mismatch. The verifiably-live-owner-is-never-stolen rule is untouched. Coverage: parser tests for both datetime shapes run everywhere, plus real-probe crash-then-reacquire and live-holder contention cases that exercise the actual probe on every CI platform including Windows.

src/lib/workspaceLock.ts
robertclawson: Fixed in 84e75cc. `ok` (may I work here) and `recorded` (can anyone else see that) are now separate answers on AcquireResult, with `unrecorded` naming the cause. Both fail-open branches log at warn, and the CLI keeps the non-blocking exit 0 but prints on stderr that the workspace is NOT claimed and work is proceeding without protection — no more `locked <path>` against an ENOTDIR state directory. CLI-level tests drive ENOTDIR, an unwritable directory, the switched-off case, the normal persisted claim, and a live-holder refusal.

src/lib/turnDigest.ts
robertclawson: Fixed in 4ce2b1c. The digest directory is now created/tightened to 0700 and both tmp and final files to 0600 on POSIX, applied with an explicit chmod as well as the create mode (a create mode is masked by umask and does nothing for a pre-existing path). The chmod is best-effort — losing a digest is worse than a mode that would not narrow — and Windows behaviour is unchanged since the ACL governs there. Tests run under a forced 0000 umask so they cannot pass by accident, covering the initial record, the markDelivered rewrite, and tightening an existing permissive directory.

src/lib/processLiveness.ts
```diff
@@ -68,14 +68,106 @@ export function processStartToken(pid: number): string | null {
       });
       return out.trim() || null;
     }
+    if (process.platform === "win32") return windowsStartToken(pid);
```
kaieriksen: **P1 — The synchronous Windows identity probe makes guard contention unusable.** This branch launches PowerShell (and potentially `wmic`) synchronously for every live competing ticket on each scan. At exact head `836324e`, the Windows job still times out the real six-process mutex test after 60 seconds with **zero** successful entries; the earlier eight-process one-shot acquire takes 23 seconds, and the release/acquire regression returns `{ ok: true }` because the liveness probe outlasts the holder and the expected contention window. Commit `836324e` removed an unrelated parser suite on the theory that runner load caused the timeout, but the rerun is still red. This is the runtime path behind `workspace lock`, `unlock`, status pruning, and prompt lock inspection—not test-only overhead. Please use a bounded/constant-time Windows ownership mechanism (or a transactional/platform-native guard) and keep deterministic Windows multi-process coverage; relaxing/removing the failing tests would hide the product failure.

kaieriksen (changes requested):
The five prior blockers are fixed, but the Windows process-identity path makes the guard unusable under real contention. Exact-head focused validation passes 157/157 and TypeScript is clean on Linux; hosted Linux/typecheck/CodeQL are green, while the Windows job fails the shipped concurrency regressions.

src/lib/processLiveness.ts
robertclawson: Fixed in `2e41ef2` (probe cost) and `bca145e` (the test staging). Both halves of this were right, and the second half was the more interesting one.

**The probe.** It is now cheapest-rung-first with the working rung pinned per process, memoised per pid for 15s, and bounded by a 3s timeout sized against the guard's budget rather than against an interpreter's worst case. Every one of those can only make an answer stale in the direction of "the holder is still alive", which is the direction that keeps two writers out of one tree. There is also an operator off-switch (`PHANTOMBOT_PROCESS_START_PROBE=0`) that degrades to the documented pid-only fallback, for boxes where the interpreter is blocked by policy.

Measured on a real Windows 11 box (build 26200), not inferred from CI: `wmic` is gone on 24H2+ and fails ENOENT in **33ms**, so trying it first costs nothing where it is absent; PowerShell answers in **323ms** warm. One probe per pid per scan, not per retry per ticket.

Separately, the probe was missing `windowsHide`. Every other Windows spawn in this repo sets it. On a host with no console of its own — a scheduled task, a service, a GUI launch — each spawn allocates a new console, and on a hot path that is a strobe of black windows rather than a flicker. That is now set.

**The regression returning `{ ok: true }`.** You were right that this was not the same defect, and it deserves to be stated plainly, because the failure mode is worse than a slow test: the old staging held the critical section for a fixed 660ms and *hoped* the other process arrived inside it. `{ ok: true }` is exactly what an unfixed build returns, so the test could fail for a reason unrelated to what it tests — and, on any machine where the child happened to arrive late, would have **passed a broken build**. It was reporting the runner's speed, not the invariant.

A window measured in milliseconds cannot bound process creation on someone else's CI, so it is now a handshake: the acquirer holds the section until the releaser reports finished, and the releaser pays its startup *before* the window opens. The remaining deadline is a hang-stop; nothing is timed against it.

Verified on the same physical Windows box rather than by re-running CI until it agreed:

| staging | slow releaser (+4s) |
| --- | --- |
| old (fixed 660ms window) | **fail** — `{ok: true}`, i.e. reproduces this finding exactly |
| new (handshake) | **pass** |

Full Windows subset on that box: **450 pass / 0 fail**, and the guard suite alone 60/60. CI is green on `bca145e`.

`836324e` reverted — you were right that removing `parseWindowsStart` from the Windows subset treated the symptom, and worse, it deleted Windows coverage of Windows-only string parsing to quieten a failure it had nothing to do with. It is back in the list.

kaieriksen (approved):
Re-reviewed at exact head bca145efa02de0020d712f6e6c3584e0924357e5. The Windows guard contention fix and deterministic release/acquire staging address the remaining blocker; 198 focused tests and TypeScript passed locally, and hosted Linux, Windows, typecheck, CodeQL, and analysis checks are green. All prior review threads are resolved.

PR #407, opened 2026-08-19
Alert on a dead primary harness, and stop failing over on shutdown

andrewagrahamhodges:
## Why

Failover is invisible to the user by design (#284). Robbie showed what that hides: a broken OAuth token failed **every** claude turn for ~3.5 days (21 consecutive `authentication_failed` fallbacks, Aug 13-16). The agent kept answering on the fallback, so nothing looked broken — the only signal anywhere was the provider's billing dashboard.

Separately, Robbie restarted 5 times in one day and each restart with a turn in flight manufactured a phantom failover: systemd SIGTERMs the cgroup, claude dies with 143, the orchestrator reads that as a recoverable harness error and spawns a **paid** fallback subprocess for a reply nobody will ever receive.

## What

**1. Alert when a primary is dead but a fallback is covering** (`src/lib/harnessAlert.ts`)

Fires after 3 consecutive **auth** failures — never on a transient 429 that a fallback absorbs, which is exactly the case #284 wanted silent. The message names the host, the failing harness, who is covering, and the fix (re-authenticate). Auth is the right trigger because unlike a rate limit it will never clear on its own.

**2. Alert when the chain is exhausted**

No harness could answer, so no reply was delivered. Always alerts, with a rate-limit-specific headline ("rate limited and there's no fallback left") since that is the shape it takes on a single-harness chain.

Dedup is **incident-scoped**, not purely time-scoped: a success closes the incident so a fixed-then-rebroken primary alerts again, plus a 6h floor for re-alerting inside one long-running incident.

The alerter is silent until `phantombot run` installs a sender — `ask`, `tick` and tests have no owner channel and must not open one.

**3. External SIGTERM is shutdown, not harness failure** (`src/lib/harnessRunner.ts`)

`isShutdownExit` classifies SIGTERM/SIGINT/SIGHUP (and 143/130/129) as terminal, so the chain stops instead of spawning a fallback mid-shutdown. SIGKILL/137 is deliberately **not** included: that is as often the OOM killer as a stop, and an OOM should stay reported. Only reachable when our own kill coordinator did not fire, so a `/stop` or timeout kill is unaffected.

## Testing

- `tests/lib-harnessAlert.test.ts` (13) — classification, threshold, rate-limit silence, incident-scoped dedup vs the time floor, throwing sender.
- `tests/orchestrator-fallback.test.ts` (+4) — end-to-end through the real orchestrator: fallback covers a dead primary → one alert; primary answering → none; rate limit with no fallback → outage alert; terminal error → none.
- `tests/lib-harnessRunner.test.ts` (+4) — real subprocess SIGTERMed by its process group yields a non-recoverable error.
- Mutation-tested both guards: disabling the shutdown branch fails the runner test; disabling the degraded call fails the orchestrator test.
- Full suite: 2941 pass, 3 pre-existing/env-dependent failures also present on `origin/main`. Typecheck clean.

## Deliberately out of scope

Surfacing harness health in `phantombot doctor`. That runs in a separate process, so it needs durable state rather than the process-local incident map — separate change, and the notification is the part that actually catches the Robbie case on day one.


andrewagrahamhodges:
Pushed `9f09911`:

- Alerts now go through `runNotify` instead of a bare `HttpTelegramTransport` — they broadcast across persona bot + default bot + phantomchat, honour reply-mode, and **persist into the conversation**, so a follow-up question has something to read.
- Message bodies are symbolic, with a five-line legend grouped by severity / cause / facts / outcome / footer. Glyphs come from a single `LEGEND` constant so they can't drift out of sync.
- Test assertions narrowed to the headline line: the previous substring matches passed regardless of the classified cause once the legend named every glyph.

andrewagrahamhodges:
Alerts are single-line now — the legend is gone, along with the timestamp (Telegram stamps it) and the raw error tail.

```
🔑 claude auth failure ×3 🖥 robbie · pi serving, run `claude /login`
🚨 claude rate limited 429 🖥 robbie · no fallback left, turn undelivered
```

35 tests pass, typecheck clean.

src/orchestrator/fallback.ts
```diff
@@ -245,6 +280,16 @@ export async function* runWithFallback(
       // harness, the CLI did its job. Clear any prior cooldown so the
       // next turn picks the chain back up at the top.
       cooldown.markSuccess(harness.id);
+      alerter.noteSuccess(harness.id);
+      // Served, but not by the head of the chain: the owner is paying a
+      // fallback to cover a broken primary and nothing else would tell them.
+      if (firstFailure && firstFailure.harnessId !== harness.id) {
+        await alerter.noteDegraded({
```
robertclawson, line 289: **P1 — the alert blocks the turn it is reporting on.**

This `await` sits on the generator's completion path, and the sender installed in `run.ts` is `runNotify`. That is Telegram (30s cap × recipients × 2 bots) *plus* an unbounded nostr publish — `publishWrap` → `Promise.allSettled(pool.publish(...))` with no timeout. A relay that accepts the connection and never ACKs will stall turn completion for as long as it stays open. It is enough of a known hazard that turn-side callers of `phantombot notify` wrap it in `timeout 25`.

`emit()` already promises this can't hurt the turn ("Never let a failed notification break the turn") — a hang is the same harm as a throw, just quieter, and `try/catch` doesn't cover it.

Suggested shape, keeps the await in tests but never on the hot path:

```ts
void alerter
  .noteDegraded({ ... })
  .catch((e) => log.warn("harnessAlert: degraded notify failed", { error: (e as Error).message }));
```

If you'd rather keep it awaited for determinism, put the deadline inside `emit()` instead — `await Promise.race([send(message), sleep(ALERT_SEND_TIMEOUT_MS)])` — which fixes both call sites at once and is the smaller diff. Either is fine; the current version isn't. Same applies to the two `await alerter.noteExhausted(...)` calls (L134, L230), though those fire on a turn that has already failed, so the stall is less costly there.

src/lib/harnessAlert.ts
```diff
@@ -0,0 +1,284 @@
...
+  ): number {
+    const cause = classifyFailure(error, httpStatus);
+    const prev = this.incidents.get(harnessId);
+    const state: IncidentState = prev ?? {
+      consecutiveFailures: 0,
+      cause,
+      alertedAt: new Map(),
+    };
+    state.consecutiveFailures += 1;
+    state.cause = cause;
```
robertclawson, line 193: **P1 — mixing causes in one run produces false alerts and a fabricated count.**

`consecutiveFailures` increments regardless of cause, and `state.cause` is overwritten by the latest failure. So the run counts *any* failures while the alert decision reads *the last* cause. Two ways that goes wrong, both reproducible against this file:

- two 429s, then a single auth blip → `🔑 claude auth failure ×3 · pi serving`. That's the 429-triggered notification the module header explicitly promises not to send, and the ×3 is not three auth failures.
- five auth failures, then one `other` (an empty reply counts — `fallback.ts:257`) → **silence**, because `cause !== "auth"` returns early. One stray failure mutes a genuinely dead token for the rest of the incident, which is exactly the 3.5-day scenario this PR exists to catch.

Cheapest fix that keeps the current structure — treat a cause change as a new run:

```ts
if (prev && prev.cause !== cause) {
  prev.consecutiveFailures = 0;
  prev.alertedAt.clear();   // or keep, if you want the re-alert floor to survive
}
```

Cleaner if you want it: `consecutiveFailures: Map<HarnessFailureCause, number>` and have `noteDegraded` read the auth counter directly, which also drops the `state.cause` field entirely. Your call — the reset is a 3-line change and I'd ship that.

Either way this wants a test: the suite currently only ever feeds one cause per incident, so both bugs pass today.

src/cli/run.ts
```diff
@@ -418,6 +422,29 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
+  // where an owner channel actually exists; every other entry point (`ask`,
+  // tick, tests) leaves the alerter unconfigured and therefore silent.
+  //
+  // Send through `runNotify` rather than a transport of our own: it is the
+  // single writer that broadcasts across the persona bot, the default bot and
+  // phantomchat (deduped), honours reply-mode, and — the part that matters
+  // here — PERSISTS the alert into the conversation. Without that an owner
+  // who replies "why did that fire?" is asking about a message no turn can
+  // see.
+  if (adminListener) {
```
robertclawson: **P2 — PhantomChat-only installs get no alerts at all.**

`adminListener` is undefined when there are no Telegram listeners, so the alerter is never configured and this whole feature is silently off — even though `runNotify` broadcasts to phantomchat perfectly well. An install that talks over phantomchat only is precisely the kind of headless box where a dead credential goes unnoticed for days.

L491 in this same file already has the fallback ladder:

```ts
const alertPersona =
  adminListener?.persona ?? phantomchatPersonas[0]?.persona ?? defaultPersona;
```

Reuse that, drop the `if`, and the comment above stays true.

src/orchestrator/fallback.ts
```diff
@@ -198,9 +216,24 @@ export async function* runWithFallback(
           // don't want to keep slamming it). markFailure() handles
           // the exponential backoff bookkeeping.
           cooldown.markFailure(harness.id);
+          alerter.noteFailure(harness.id, chunk.error, chunk.httpStatus);
+          firstFailure ??= { harnessId: harness.id, error: chunk.error };
           recoverableError = true;
           break;
         }
+        // Terminal, or recoverable-but-nowhere-left-to-go. Either way the
```
robertclawson: **P2 — the comment says the opposite of the code.**

"Terminal, or recoverable-but-nowhere-left-to-go. **Either way** the user gets no reply… worth waking the owner" — but the very next line is `if (chunk.recoverable)`, so terminal errors alert nobody. There are three other exits that also deliver no reply and stay silent: the terminal error here, "ended without done or error", and payload-too-large on the last harness.

Pick one and make them agree. If the narrow behaviour is intended, say why — something like "only the recoverable-but-exhausted case; a terminal error is already surfaced to the user as an error chunk". If the comment is the intent, alert on the terminal path too. Right now the next reader can't tell which is the bug.

src/lib/harnessAlert.ts
```diff
@@ -0,0 +1,284 @@
...
+ * harness. `noteSuccess` closes the incident, so the NEXT outage alerts
+ * again. Within an incident we alert at most once per kind, plus a
+ * re-alert floor (REALERT_MS) so a week-long outage doesn't go silent
+ * forever after its opening message.
+ *
+ * Trigger threshold: DEGRADED waits for DEGRADE_AFTER_FAILURES consecutive
+ * failures rather than firing on the first one. A single 429 that recovers
+ * two minutes later (observed on Robbie 2026-08-19) is not worth a push
+ * notification; a token that is actually dead fails every single turn and
+ * crosses the threshold within seconds.
```
robertclawson: **P3 — "crosses the threshold within seconds" isn't right, and the real number is still a good number.**

Cooldown backoff (150s, 300s, …) skips the harness entirely between attempts, so three real attempts is **~7.5 minutes minimum**, and longer at sparse turn cadence — a box that gets a turn every 20 minutes takes closer to an hour. Worth stating accurately: "3.5 days → single-digit minutes" sells the fix better than a claim a reader can disprove.

Two dead parameters while you're in here: `noteExhausted`'s `chain` and `noteDegraded`'s `error` are both computed at the call site, passed, and never read. Either render them (`chain` would make the exhausted alert genuinely more useful — "no fallback left" begs the question "out of what?") or drop them from the interfaces. I'd render `chain`, drop `error`.

robertclawson (changes requested):
Read the whole diff at `085b3e6`. The design here is right, and I want to say that before the list: **auth as the trigger** (a 429 clears itself, a dead token never does) is the correct discriminator, **incident-scoped dedup with a re-alert floor** is the right shape for something that must not become noise, and installing the sender only at the daemon — where an owner channel actually exists — is a better call than a global that half-works from `ask` and tests. The module header is genuinely good documentation of *why*.

Five findings, two blocking. Both blockers are small diffs; I don't think any of this needs a redesign.

**Blocking**
1. `fallback.ts:287` — the degraded alert is `await`ed on the turn's completion path, and the sender can hang on a nostr relay with no timeout.
2. `harnessAlert.ts:165` — causes mix within one incident, which both fires false auth alerts with an invented count *and* mutes real ones.

**Non-blocking**
3. `run.ts:438` — phantomchat-only installs are silently excluded.
4. `fallback.ts:224` — comment contradicts the code below it.
5. `harnessAlert.ts:35` — "within seconds" is ~7.5 min; two parameters are dead.

Fastest route to green, if it helps: (2) is a 3-line reset in `noteFailure` plus one mixed-cause test, (1) is either a `void`+`.catch` or a `Promise.race` deadline inside `emit()`, and (3) is copying the persona ladder already on L491. (4) and (5) are comment/interface tidying and I won't hold the PR on them.

The thing I'd most like to see added to the tests: a mixed-cause incident. The current suite feeds one cause per incident throughout, which is why both halves of finding 2 pass today. Happy to re-review as soon as it's pushed.

andrewagrahamhodges:
Pushed `9529627` — all five addressed.

**1. Alert blocking the turn (P1).** Took your smaller-diff option: a deadline inside `emit()`, so both call sites get it. Unref'd timer, and the send is abandoned rather than waited on — waiting is the stall. Timeout is injectable so the test asserts in 50ms instead of 20s.

**2. Mixed causes (P1).** Cause change resets `consecutiveFailures`. `alertedAt` deliberately survives: the re-alert floor is anti-spam, and a flapping cause must not be a way around it. Three tests, covering both directions you described plus the count itself.

**3. PhantomChat-only installs (P2).** Ladder copied, `if` dropped — and since `doctorPersona` below was the same three terms, it now reuses `alertPersona` rather than declaring it twice.

**4. Contradictory comment (P2).** The narrow behaviour is intended; comment now says why — a terminal error is yielded as an error chunk, so the user is already looking at it and an alert only says it twice.

**5. Header + dead params (P3).** "single-digit minutes" with the 150s/300s arithmetic spelled out. Rendered `chain` (`no fallback left [claude → pi]`), dropped `error` from `noteDegraded`.

Mutation-tested both new guards: reverting the cause reset fails 2 tests, removing the deadline fails 1. 63 pass across the three suites, typecheck clean.

src/orchestrator/fallback.ts
robertclawson: Fixed at `9529627` — verified. `withDeadline` (20s, unref'd timer) wraps the send inside `emit()`, so all three call sites are covered rather than just this one; nice call putting it there. Reverted it locally and the hanging-sender test times out at 5s, so the guard is real.

Two residuals I'm explicitly **not** blocking on, just so they're on the record:
- the degraded alert is still `await`ed on the completion path — now bounded at 20s rather than unbounded. Acceptable given it fires at most once per 6h per incident.
- `emit()` stamps `alertedAt` *before* sending, so a send that hits the deadline still burns the 6h floor and the owner hears nothing. Pre-existing, but the new deadline makes it reachable. Worth a follow-up issue, not a change here.

Resolving.

src/lib/harnessAlert.ts
robertclawson: Fixed at `9529627` — verified against my two original repros:
- two 429s + one auth blip → **silence** (was `🔑 claude auth failure ×3`, with a fabricated count)
- auth×2 + a stray 429 + auth×3 → **`auth failure ×3`** (was permanent silence for the rest of the incident)

And the new tests earn their keep: reverting the one-line cause-change reset fails exactly those two and nothing else. That's the mixed-cause coverage I asked for. Resolving.

src/cli/run.ts
robertclawson: Fixed at `9529627` — and better than what I suggested. Rather than duplicating the ladder you collapsed it with the doctor's identical one, so there's a single fallback path instead of two that can drift. Checked scope: both vars are defined ~200 lines earlier, so the unconditional application is safe. Resolving.

src/orchestrator/fallback.ts
robertclawson: Fixed at `9529627`. The comment now states that terminal errors deliberately don't alert *and* why (already yielded to the channel), which is more useful than the version I'd have written — it documents the intent, not just the branch. Resolving.

src/lib/harnessAlert.ts
robertclawson: Fixed at `9529627`. `error` dropped from `noteDegraded`, and `chain` is now actually rendered rather than deleted — that was the better of the two options I offered, since the chain is genuinely useful in the alert text. Timing claim now reads accurately. Resolving.

robertclawson (approved):
Approving at `9529627`. All five findings fixed, each verified against the code rather than the commit message — including reverting both new guards to confirm the tests actually go red.

The two blockers came back stronger than what I proposed: the deadline went inside `emit()` so all three call sites are covered, and the persona ladder got collapsed with the doctor's identical one instead of duplicated. The mixed-cause tests are the coverage that was missing — my original repros now give the right answers in both directions (no false `×3` push on 429s, and a stray 429 no longer mutes a genuinely dead token).

Two non-blocking residuals noted on the threads for a follow-up, not this PR:
- the degraded alert is still `await`ed on the turn's completion path — bounded at 20s now, not removed.
- `emit()` stamps `alertedAt` before sending, so a send that hits the deadline still burns the 6h floor and the owner hears nothing.

Nice work turning this round fast. Heads-up: the branch is **BEHIND** main — rebase or "Update branch" before merging.

PR #409, opened 2026-08-20
Name a wedged harness in the outage alert, and stop implying a fallback existed

andrewagrahamhodges:
Follow-up to #407, from a live alert on kw-phantombot:

```
🚨 pi harness error 🖥 kw-phantombot · no fallback left [pi], turn undelivered
```

pi stalled with no output for 5+ minutes, phantombot killed it on the 300s idle timeout, and the chain was length 1 so nothing else could answer. The alert was correct but uninformative — Lena had to read the journals to learn the harness was *wedged* rather than crashed, which is the whole thing the alert was supposed to save.

## Changes

**`timeout` joins the cause enum.** Matched on the text `killCauseToErrorChunk()` already stamps in `src/lib/harnessRunner.ts` — the hard wall-clock cap, the idle window, and the startup handshake window. Renders as `timed out`:

```
🚨 pi timed out 🖥 kw-phantombot · no fallback configured [pi], turn undelivered
```

This changes no policy. Like `other`, it never fires the degraded alert, so a wedge that a fallback absorbed stays silent exactly as #284 intended — there's a test pinning that.

**Single-harness chains say "no fallback configured".** "No fallback left" implies one was tried and also failed, which points the reader at the wrong fix. With a chain of one, the fix is *add a fallback*, not *debug two harnesses*.

## Notes for review

- The three timeout strings in the test are copied verbatim from `killCauseToErrorChunk()`. If that wording is reworded, these tests fail — deliberate, since the classifier matches on its text.
- Auth still wins over timeout in `classifyFailure` (ordering test included): a sticky credential failure that happens to mention a timeout must not be downgraded to a transient-looking label.
- Mutation-verified both guards: removing the `TIMEOUT_MARKERS` branch fails 2 tests, hardcoding `"no fallback left"` fails 2 tests.
- 44 tests pass across `lib-harnessAlert` + `orchestrator-fallback`, typecheck clean.

### DIFF
--- a/src/lib/harnessAlert.ts
+++ b/src/lib/harnessAlert.ts
@@ -68,9 +68,14 @@ export const ALERT_SEND_TIMEOUT_MS = 20_000;
  *
  * `auth` is the sticky one: unlike a 429 it will never clear on its own,
  * so it is worth waking the owner. `rate_limit` is transient by nature and
- * only escalates when it takes the whole chain down with it.
+ * only escalates when it takes the whole chain down with it. `timeout` is
+ * split out of `other` purely so the outage alert can NAME it: a wedged
+ * harness and a crashed one both rendered as "harness error", which sent
+ * the reader to the journals to find out which (observed on
+ * kw-phantombot 2026-08-20 — a 300s idle kill on a single-harness chain).
+ * It changes no policy: like `other` it never fires the degraded alert.
  */
-export type HarnessFailureCause = "auth" | "rate_limit" | "other";
+export type HarnessFailureCause = "auth" | "rate_limit" | "timeout" | "other";
 
 /**
  * Statuses the claude CLI stamps that mean "this credential/account is not
@@ -89,6 +94,16 @@ const AUTH_MARKERS = [
 const RATE_LIMIT_MARKERS = ["rate_limit", "quota", "resource_exhausted"];
 
 /**
+ * Text `killCauseToErrorChunk` (src/lib/harnessRunner.ts) stamps when it
+ * kills a harness for producing nothing — the hard wall-clock cap, the idle
+ * window, and the startup handshake window respectively. Kept as substrings
+ * of those exact strings so a reworded suffix does not silently
+ * de-classify; if that function's wording changes, this list changes with
+ * it and the tests below say so.
+ */
+const TIMEOUT_MARKERS = ["timed out after", "produced no output within"];
+
+/**
  * Classify a harness error chunk. Matching is on lowercased substrings
  * because the text is harness-authored and varies by CLI ("claude api
  * error: authentication_failed", "pi: 401 unauthorized"). HTTP status
@@ -105,6 +120,7 @@ export function classifyFailure(
   const text = error.toLowerCase();
   if (AUTH_MARKERS.some((m) => text.includes(m))) return "auth";
   if (RATE_LIMIT_MARKERS.some((m) => text.includes(m))) return "rate_limit";
+  if (TIMEOUT_MARKERS.some((m) => text.includes(m))) return "timeout";
   return "other";
 }
 
@@ -239,7 +255,9 @@ export class HarnessAlerter {
         ? "rate limited"
         : cause === "auth"
           ? "auth failure"
-          : "harness error";
+          : cause === "timeout"
+            ? "timed out"
+            : "harness error";
     const detail =
       cause === "rate_limit" && input.httpStatus
         ? ` ${input.httpStatus}`
@@ -250,10 +268,15 @@ export class HarnessAlerter {
     // and the answer is the difference between "add a fallback" and "both of
     // my harnesses are down".
     const chain = input.chain.length > 0 ? ` [${input.chain.join(" \u2192 ")}]` : "";
+    // "no fallback left" implies one was tried and also failed. With a
+    // single-harness chain nothing was ever configured to try, and that is a
+    // different fix (add a fallback, not debug two harnesses), so say so.
+    const exhaustion =
+      input.chain.length <= 1 ? "no fallback configured" : "no fallback left";
     await this.emit(
       input.harnessId,
       "exhausted",
-      `\u{1f6a8} ${input.harnessId} ${label}${detail}${this.hostTag()} \u00b7 no fallback left${chain}, turn undelivered`,
+      `\u{1f6a8} ${input.harnessId} ${label}${detail}${this.hostTag()} \u00b7 ${exhaustion}${chain}, turn undelivered`,
     );
   }
 
--- a/src/lib/harnessAlert.ts
+++ b/src/lib/harnessAlert.ts
@@ -269,10 +269,15 @@ export class HarnessAlerter {
     // my harnesses are down".
     const chain = input.chain.length > 0 ? ` [${input.chain.join(" \u2192 ")}]` : "";
     // "no fallback left" implies one was tried and also failed. With a
-    // single-harness chain nothing was ever configured to try, and that is a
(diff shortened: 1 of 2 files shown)


tests/lib-harnessAlert.test.ts
```diff
@@ -57,10 +57,32 @@ describe("classifyFailure", () => {
     expect(classifyFailure("something vague", 401)).toBe("auth");
   });
 
-  test("anything else is 'other' — a timeout must not page the owner", () => {
+  // The three strings below are copied verbatim from killCauseToErrorChunk()
+  // in src/lib/harnessRunner.ts. If that wording changes, these fail — which
+  // is the point: the classifier matches on its text.
+  test("recognises every wedged-harness kill the runner emits", () => {
```
robertclawson, line 75: **Blocker — these three strings are copies, so this test only tests itself.**

I reworded `killCauseToErrorChunk` (`timed out after` → `stalled for`, and the startup string) and the suite still gave **44 pass / 0 fail**, while the real classifier would have de-classified every idle and startup kill to `other`. The comment above promises the opposite.

Derive the text from the function instead of copying it:

```ts
import { killCauseToErrorChunk } from "../src/lib/harnessRunner.ts";

test("recognises every wedged-harness kill the runner emits", () => {
  for (const cause of ["timeout", "idle", "startup"] as const) {
    const chunk = killCauseToErrorChunk(cause, "pi", 300_000, 300_000, 60_000);
    expect(classifyFailure(chunk!.error)).toBe("timeout");
  }
});
```

Verified: red against the reworded runner, green against HEAD. Keep the auth-wins-over-“timed out” test as-is — that one is about precedence, not the runner's wording, so a literal is correct there.

src/lib/harnessAlert.ts
```diff
@@ -88,6 +93,16 @@ const AUTH_MARKERS = [
 
 const RATE_LIMIT_MARKERS = ["rate_limit", "quota", "resource_exhausted"];
 
+/**
+ * Text `killCauseToErrorChunk` (src/lib/harnessRunner.ts) stamps when it
+ * kills a harness for producing nothing — the hard wall-clock cap, the idle
+ * window, and the startup handshake window respectively. Kept as substrings
+ * of those exact strings so a reworded suffix does not silently
+ * de-classify; if that function's wording changes, this list changes with
```
robertclawson, line 101: This claim ("if that function's wording changes, this list changes with it and the tests below say so") is currently false — see the test-file comment. Once the tests derive their input from `killCauseToErrorChunk`, this becomes accurate and is worth keeping exactly as written.

src/lib/harnessAlert.ts
```diff
@@ -250,10 +268,15 @@ export class HarnessAlerter {
     // and the answer is the difference between "add a fallback" and "both of
     // my harnesses are down".
     const chain = input.chain.length > 0 ? ` [${input.chain.join(" \u2192 ")}]` : "";
+    // "no fallback left" implies one was tried and also failed. With a
+    // single-harness chain nothing was ever configured to try, and that is a
+    // different fix (add a fallback, not debug two harnesses), so say so.
+    const exhaustion =
+      input.chain.length <= 1 ? "no fallback configured" : "no fallback left";
```
robertclawson: Nit, non-blocking. `chain.length <= 1` isn't only "nothing was configured": `buildHarnessChain` drops unknown ids with a warning, so a typo in `harnesses.chain` also yields a length-1 chain — and the owner then gets told to add a fallback they *did* add. Either soften to something like "no usable fallback configured", or leave it and rely on the startup warning. Your call.

tests/lib-harnessAlert.test.ts
```diff
@@ -201,6 +223,19 @@ describe("send deadline", () => {
     // Settled, and settled by the deadline rather than by the send.
     expect(Date.now() - started).toBeLessThan(2000);
   });
+  test("a wedged harness a fallback absorbed stays silent", async () => {
```
robertclawson: Nit: this one is about degraded-vs-silence policy, not the send deadline — it's filed under `describe("send deadline")`. Worth moving next to the other degraded tests so the timeout policy is findable in one place.

robertclawson (changes requested):
Right instinct, and the small scope is the right call. Distinguishing "the harness wedged" from "the harness crashed" is exactly what this alert exists to say, and I verified the policy really is unchanged: a timeout never fires `degraded`, and there's a test pinning that. `chain` no longer claiming a fallback that was never configured is a real correctness fix too.

One blocker, and it's specifically the safety property the PR claims for itself. Two nits after that which I won't hold anything on.

**Blocker: the tests can't catch a runner rewording, though the comment says they can.**

I reworded `killCauseToErrorChunk` in `src/lib/harnessRunner.ts` (`timed out after` → `stalled for`, plus the startup string) and ran the suite: **44 pass / 0 fail**. Meanwhile production would silently classify every idle and startup kill back to `other` and print the generic "harness error" again — i.e. the exact regression this PR exists to prevent would ship green. The tests assert against their own copies of the strings, so they only ever test themselves.

The fix is a few lines and gives you the guard the comment already promises — see the inline suggestion. I ran it both ways: it **fails** against the reworded runner and **passes** against real HEAD.

Everything else here is good. Once that's in I'll approve.

andrewagrahamhodges:
All three taken in `a055c2f`.

**Blocker — self-referential test.** You're right that it only tested itself. `killText()` now calls `killCauseToErrorChunk()` and feeds its output straight to `classifyFailure`, and the loop covers all three kill causes instead of three hand-copied literals. Mutation-verified with your exact rewording (`timed out after` → `stalled for`): 2 fail. Reverted, 22 pass.

**Nits.** Wording is now "no usable fallback configured" — good catch that a typo'd harness id also yields a chain of one. Timeout-policy test moved into `degraded alert`; it was never about the deadline.

Typecheck clean. Three unrelated failures in the harness subprocess tests reproduce on the unmodified branch here, so they're environmental on this box, not from this change.

tests/lib-harnessAlert.test.ts
robertclawson: Verified on `a055c2f`: `killText()` derives from `killCauseToErrorChunk()` and the loop covers all three kill causes. Re-ran the mutation (`timed out after` → `stalled for`, `produced no output within` → `gave nothing within`): 20 pass / 2 fail. Reverted: 22 pass. This now tests the thing it claims to. Resolving.

robertclawson (approved):
Approving `a055c2f`. All three review points landed, and the blocker is genuinely fixed rather than papered over.

**Blocker (self-referential test) — verified fixed.** `killText()` calls `killCauseToErrorChunk()` and feeds the real chunk to `classifyFailure`, and the loop covers `timeout` / `idle` / `startup`. I re-ran the mutation on this branch (`timed out after` → `stalled for`, `produced no output within` → `gave nothing within`): **20 pass / 2 fail**. Reverted: **22 pass**. The comment in `harnessAlert.ts` about the marker list tracking the runner's wording is now true.

**Nits — both taken.** "no usable fallback configured" covers the typo'd/unresolvable-harness-id case, and the timeout-policy test now sits with the other degraded tests where it's findable.

**Checks on my side:** `lib-harnessAlert` 22/22, `orchestrator-fallback` 22/22, `tsc --noEmit` clean. Nothing else in the tree consumes `HarnessFailureCause` exhaustively, so adding a variant can't leave a stale branch elsewhere. The only behavioural side effect of splitting `timeout` out of `other` is `noteFailure`'s per-cause run counter resetting when a wedge interleaves with a crash — and since neither cause fires the degraded alert, that's a no-op.

The three harness-subprocess failures you saw reproduce on unmodified `main` here too — environmental, not from this change.

Issue #410, opened 2026-08-20
Nightly sweep only appends: drawers and daily files grow unbounded, need a compaction half


robertclawson:
## Problem

The nightly sweep only ever **appends**. `distill` has a trim half for MEMORY.md (step 3, "FILL and TRIM"), but the drawers (step 2) and the daily files have no compaction path at all, and nothing anywhere has a size budget. After ~4 months on one persona:

| file | size | entries | exact dupes |
|---|---|---|---|
| `memory/decisions.md` | **663 KB** | 1411 | 64 |
| `memory/lessons.md` | **574 KB** | 1218 | 26 |
| `memory/commitments.md` | **152 KB** | 878 | 8 |
| `memory/people.md` | 35 KB | 221 | 4 |
| `memory/norms.md` | 27 KB | 80 | 0 |
| `memory/<yesterday>.md` | **62 KB** | — | — |

Date headers run from `2026-04-03` to today with nothing ever retired.

## Why it hurts

1. **Step 2 asks for something impossible.** The prompt says *"Do not duplicate what is already filed — read the drawer first."* Reading a 663 KB drawer to dedupe one night's items is not a realistic instruction, so the dedupe silently doesn't happen — hence the 64/26 exact duplicates. The instruction is load-bearing and unenforced.
2. **Search quality degrades.** Drawers aren't injected wholesale, but every hybrid-search hit is drawn from a pool that is mostly superseded, completed, or duplicated. `commitments.md` in particular is 878 entries of which the large majority are long since discharged, and there is no "done" state — a closed commitment ranks alongside a live one.
3. **Daily files land in context by the agent's own hand.** The workspace AGENTS.md instructs reading today's *and* yesterday's daily file at session start. At 62 KB that is a material fraction of the window, every session, capped by nothing.
4. **Cost is per-turn and compounding.** Every drawer read, every search over them, every nightly pass that is told to read them first.

## Proposed fix — give the nightly a compaction half

Symmetry rule: **every file the nightly appends to needs a compaction pass that can also remove.** Concretely:

- **Size budget per file**, configurable with sane defaults (e.g. 128 KB drawers, 16 KB MEMORY.md). When a file is over budget after the fill pass, run a compaction turn on that file alone.
- **Drawer compaction** — dedupe exact and near-exact entries; merge entries that restate the same decision/lesson across dates into one canonical entry with a date range; move anything with a permanent KB home to a one-line pointer; retire entries superseded by a later one.
[...]

PR #411, opened 2026-08-20
feat(nightly): add a compaction stage that can only shrink what is over budget

robertclawson:
Stage three of the nightly: the first thing in phantombot that can remove a byte.

`distill` and `kb` only ever append. Nothing has ever shrunk, so the always-in-context files grow without bound — the drawers on one box reached 663KB, which is rent paid on every single turn and buries live facts under dead ones.

The design bias here is **safe over thorough**. A night that reclaims nothing is fine; a night that loses a fact is not.

- **Only over-budget files are candidates.** MEMORY.md 16KB, drawers 128KB, a fully-distilled daily file 8KB. Per-persona overrides in `memory/.compaction-budgets.json` — byte budgets only, never the shrink guard. A healthy persona pays one `stat` per file and no turn.
- **Every candidate is copied verbatim to `memory/archive/<YYYY-MM-DD>/` before the stage runs.** Nothing is ever deleted, and the nightly is the only code path that moves a memory file. Recovery is `cp`, not a restore.
- **Every file is re-stat-ed and judged afterwards.** A pass that removes more than its allowance (40%; 90% for a closed daily file), empties a file, or loses one is rolled back from that copy and recorded as `reverted`. This settles even when the turn *errored* — a timed-out stage may have left a half-rewrite on disk, which is precisely what the guard is for.
- **Byte accounting per file goes in the ledger** under `compaction`, so "is memory still growing?" is answerable without grepping journald.
- **Runs once per sweep**, after every date is distilled — its inputs are whole-file sizes, not one day's events, and running it per date would rewrite MEMORY.md N times while draining a backlog. Skipped for `--date` backfills, for personas with a monolithic prompt override, and by `--no-compact`.
- **Daily files are trimmed only** once the ledger shows both stages `ok` for that date and it is at least 30 days old. The age gate isn't about safety — the ledger already proves distillation — it's so a recent day is still readable in full.

### Deliberately not in scope: drawer dedupe

`commitments.md`, `decisions.md` and `lessons.md` are append-only logs with a lifecycle — a table pretending to be prose. They become database rows in the follow-up work, at which point dedupe is a uniqueness constraint and "compaction" is a query. An LLM dedupe pass over them now would be built twice and thrown away once. So this stage measures the drawers and reports the overage, and its prompt explicitly tells the stage not to rewrite them.

### Tests

`tests/lib-nightlyCompact.test.ts` (14) covers candidate selection including both daily gates, budget overrides, every verdict at its boundary, and the archive/restore path. `tests/cli-nightly.test.ts` adds six end-to-end cases: once-per-sweep ordering, no-candidates-no-turn, revert on overshoot, restore after deletion, both skip paths, and settle-after-a-failed-turn.

Full suite green (3133 pass), typecheck clean.

Part of #410.

### DIFF
--- a/src/cli/nightly.ts
+++ b/src/cli/nightly.ts
@@ -36,11 +36,13 @@ import { refreshPersonaIndex } from "../lib/indexRefresh.ts";
 import type { WriteSink } from "../lib/io.ts";
 import { log } from "../lib/logger.ts";
 import {
+  buildCompactionPrompt,
   buildNightlyPromptForPersona,
   buildNightlyStagePrompt,
   dateRecord,
   loadNightlyState,
   NIGHTLY_STAGES,
+  type NightlyState,
   type NightlyStage,
   nightlyConversationKey,
   type PendingDate,
@@ -51,6 +53,15 @@ import {
   selfStartToken,
   sweepLiveness,
 } from "../lib/nightly.ts";
+import {
+  archiveDirPath,
+  archiveForCompaction,
+  type CompactionCandidate,
+  type CompactionOutcome,
+  compactionCandidates,
+  formatCompactionSummary,
+  settleCompaction,
+} from "../lib/nightlyCompact.ts";
 import { openMemoryStore } from "../memory/store.ts";
 import { runTurn } from "../orchestrator/turn.ts";
 
@@ -97,6 +108,14 @@ export interface RunNightlyInput {
   maxDates?: number;
   /** Run even if another sweep holds the in-flight marker. */
   force?: boolean;
+  /**
+   * Skip the compaction stage (issue #410). Set by `--no-compact`, and always
+   * true for a `--date` backfill: reprocessing one old day should not also
+   * rewrite MEMORY.md.
+   */
+  skipCompaction?: boolean;
+  /** Test seam — override the daily-file age gate. */
+  compactMinAgeDays?: number;
   out?: WriteSink;
   err?: WriteSink;
   /** Test seam — override the clock. */
@@ -105,7 +124,7 @@ export interface RunNightlyInput {
   runStage?: (args: {
     persona: string;
     date: string;
-    stage: NightlyStage | "override";
+    stage: NightlyStage | "override" | "compact";
     prompt: string;
   }) => Promise<TurnResult>;
   /** Test seam — skip the real index refresh. */
@@ -201,6 +220,63 @@ export async function runNightlyTurn(opts: {
     errored = (e as Error).message;
   }
   return { finalReply, errored, durationMs: Date.now() - startedAt };
+}
+
+
+/**
+ * Run the compaction pass over whatever is currently over budget.
+ *
+ * The safety property lives here, not in the prompt: every candidate is
+ * copied into `memory/archive/<today>/` BEFORE the stage runs, and each file
+ * is re-stat-ed and judged afterwards. A pass that overshoots its shrink
+ * allowance, empties a file or loses one is rolled back from that copy. The
+ * stage itself is never trusted to have done the right thing.
+ *
+ * Returns null when nothing is over budget — the common case, and free.
+ */
+export async function runCompaction(opts: {
+  personaDir: string;
+  persona: string;
+  today: string;
+  state: NightlyState;
+  minAgeDays?: number;
+  runOne: (prompt: string) => Promise<TurnResult>;
+  out: WriteSink;
+}): Promise<{ outcomes: CompactionOutcome[]; error?: string } | null> {
+  const candidates = await compactionCandidates(opts.personaDir, opts.state, {
+    today: opts.today,
+    ...(opts.minAgeDays !== undefined ? { minAgeDays: opts.minAgeDays } : {}),
+  });
+  if (candidates.length === 0) return null;
+
+  const archiveDir = archiveDirPath(opts.personaDir, opts.today);
+  const archived = new Map<CompactionCandidate, string>();
+  for (const c of candidates) {
+    archived.set(c, await archiveForCompaction(opts.personaDir, c, opts.today));
+  }
+
+  const r = await opts.runOne(
+    buildCompactionPrompt(opts.persona, opts.today, candidates, archiveDir),
+  );
+
+  // Settle regardless of how the turn ended: a stage that timed out may still
+  // have half-rewritten a file, and that is exactly what the guard is for.
+  const outcomes: CompactionOutcome[] = [];
+  for (const c of candidates) {
+    outcomes.push(await settleCompaction(c, archived.get(c)!));
+  }
+  opts.out.write(formatCompactionSummary(outcomes));
+
+  await saveNightlyState(opts.personaDir, {
+    compaction: {
+      ran_at: new Date().toISOString(),
+      archive_dir: archiveDir,
+      files: outcomes,
+      bytes_before: outcomes.reduce((n, o) => n + o.bytesBefore, 0),
+      bytes_after: outcomes.reduce((n, o) => n + o.bytesAfter, 0),
+    },
+  });
+  return { outcomes, ...(r.errored ? { error: r.errored } : {}) };
 }
 
 export async function runNightly(input: RunNightlyInput = {}): Promise<number> {
@@ -329,7 +405,7 @@ export async function runNightly(input: RunNightlyInput = {}): Promise<number> {
       });
 
       const runOne = async (
-        stage: NightlyStage | "override",
+        stage: NightlyStage | "override" | "compact",
         prompt: string,
       ): Promise<TurnResult> =>
         input.runStage
@@ -415,6 +491,49 @@ export async function runNightly(input: RunNightlyInput = {}): Promise<number> {
         out.write(`nightly: ${pending.date} ok (${Date.now() - t0}ms)\n`);
       }
     }
+
+    // ---------------------------------------------------------------------
+    // Stage three: compaction (issue #410).
+    //
+    // Runs ONCE per sweep, after every date has been distilled, because its
+    // inputs are whole-file sizes rather than one day's events — and because
+    // running it per date would rewrite MEMORY.md N times in a backlog drain.
+    // Skipped for `--date` backfills and for personas with a monolithic
+    // prompt override, which owns its own contract.
+    // ---------------------------------------------------------------------
+    if (!input.skipCompaction && !input.today && !monolithic) {
+      const compaction = await runCompaction({
+        personaDir: dir,
+        persona,
+        today: now.toISOString().slice(0, 10),
+        state: await loadNightlyState(dir),
+        minAgeDays: input.compactMinAgeDays,
+        runOne: async (prompt) =>
+          input.runStage
+            ? await input.runStage({
+                persona,
+                date: now.toISOString().slice(0, 10),
+                stage: "compact",
+                prompt,
+              })
+            : await runNightlyTurn({
+                persona,
+                conversation: `${nightlyConversationKey(now.toISOString().slice(0, 10))}:compact`,
+                userMessage: prompt,
+                agentDir: dir,
+                harnesses,
+                memory: memory!,
+              }),
+        out,
+      });
+      if (compaction?.error) {
+        errors.push(`compaction: ${compaction.error}`);
+        log.error("nightly: compaction stage failed", {
+          persona,
+          error: compaction.error,
+        });
+      }
+    }
   } finally {
     await memory?.close();
   }
@@ -466,6 +585,12 @@ export default defineCommand({
       description: "Run even if another sweep holds the in-flight marker.",
       default: false,
     },
+    "no-compact": {
+      type: "boolean",
+      description:
+        "Skip the compaction stage (leave over-budget files untouched).",
+      default: false,
+    },
   },
   async run({ args }) {
     const max = args["max-dates"] ? Number(args["max-dates"]) : undefined;
@@ -474,6 +599,7 @@ export default defineCommand({
       today: args.date ? String(args.date) : undefined,
       maxDates: Number.isFinite(max) ? max : undefined,
       force: Boolean(args.force),
+      skipCompaction: Boolean(args["no-compact"]),
     });
   },
 });
--- a/src/cli/nightly.ts
+++ b/src/cli/nightly.ts
@@ -562,8 +562,26 @@ export async function runNightly(input: RunNightlyInput = {}): Promise<number> {
     // running it per date would rewrite MEMORY.md N times in a backlog drain.
     // Skipped for `--date` backfills and for personas with a monolithic
     // prompt override, which owns its own contract.
+    //
+    // Also skipped whenever any date stage failed this sweep. A failed distill
+    // turn can still have partially rewritten MEMORY.md before erroring, so the
+    // archive compaction takes would capture that damaged state rather than the
+    // clean pre-sweep one — and a second model rewrite on top of it would bury
+    // the damage. Leave over-budget files alone and compact on the next clean
+    // sweep instead.
     // ---------------------------------------------------------------------
-    if (!input.skipCompaction && !input.today && !monolithic) {
+    const compactionEligible =
+      !input.skipCompaction && !input.today && !monolithic;
+    if (compactionEligible && errors.length > 0) {
+      out.write(
+        `nightly: compaction skipped — ${errors.length} stage error(s) this sweep; ` +
+          `over-budget files are left untouched until a clean sweep\n`,
+      );
+      log.warn("nightly: compaction skipped after stage failure", {
+        persona,
+        errors: errors.length,
+      });
+    } else if (compactionEligible) {
       // Keep the beat fresh across a stage that can outlive the stall timer.
       await holdSweep(todayStamp, queue.length);
       const compaction = await runCompaction({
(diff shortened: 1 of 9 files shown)


src/cli/nightly.ts
```diff
@@ -415,6 +491,49 @@ export async function runNightly(input: RunNightlyInput = {}): Promise<number> {
...
+    // ---------------------------------------------------------------------
+    // Stage three: compaction (issue #410).
+    //
+    // Runs ONCE per sweep, after every date has been distilled, because its
+    // inputs are whole-file sizes rather than one day's events — and because
+    // running it per date would rewrite MEMORY.md N times in a backlog drain.
+    // Skipped for `--date` backfills and for personas with a monolithic
+    // prompt override, which owns its own contract.
+    // ---------------------------------------------------------------------
+    if (!input.skipCompaction && !input.today && !monolithic) {
```
kaieriksen: [P1] Compaction is skipped on every idle sweep

The `queue.length === 0` return above exits before this block, so once all daily files are processed, a normal scheduled sweep never reaches compaction—even when `MEMORY.md` or a drawer is over budget. In the normal steady-state dogfood path there is no pending date to keep the function open, which makes the advertised automatic compaction permanently inert. Run compaction independently of the daily-date queue (while still honoring `--date`/`--no-compact`), or move this early return after the compaction decision.

src/lib/nightlyCompact.ts
```diff
@@ -0,0 +1,355 @@
...
+  personaDir: string,
+  candidate: CompactionCandidate,
+  stamp: string,
+): Promise<string> {
+  const dir = archiveDirPath(personaDir, stamp);
+  await mkdir(dir, { recursive: true });
+  // Flatten the persona-relative path so `memory/decisions.md` and a future
+  // `kb/decisions.md` can't collide inside one archive dir.
+  const dest = join(dir, candidate.path.replace(/[\\/]/g, "__"));
+  if (!existsSync(dest)) await copyFile(candidate.absPath, dest);
```
kaieriksen: [P1] Do not place rollback archives under an indexed memory subtree

`walkMarkdown()` recursively indexes every non-hidden `.md` under `memory/`, including `memory/archive/<date>/...`. The exact pre-compaction copy therefore becomes a live FTS/vector memory document on the next refresh, and stale/deleted facts can be retrieved as if they were current. This also exposes the archive contents to ordinary memory search. Store archives outside `memory/` or explicitly exclude the archive subtree from both indexing and embedding, with a regression test.

kaieriksen: [P1] The same-day archive must not be reused for a later pass

`existsSync(dest)` means a second compaction pass on the same day keeps the first pass’s pre-image. If the first pass legitimately shrinks a file but leaves it over budget, and the second pass is over-eager, `settleCompaction()` restores the older pre-first-pass bytes and resurrects content that was already removed. Either use a unique archive per pass or overwrite the pre-image atomically before every pass; add a two-pass rollback regression test.

src/cli/nightly.ts
```diff
@@ -203,6 +222,63 @@ export async function runNightlyTurn(opts: {
...
+
+  // Settle regardless of how the turn ended: a stage that timed out may still
+  // have half-rewritten a file, and that is exactly what the guard is for.
+  const outcomes: CompactionOutcome[] = [];
+  for (const c of candidates) {
+    outcomes.push(await settleCompaction(c, archived.get(c)!));
+  }
+  opts.out.write(formatCompactionSummary(outcomes));
+
+  await saveNightlyState(opts.personaDir, {
```
kaieriksen, line 309: [P1] Reconcile the ledger and index after compaction rewrites files

Compaction runs after the per-date index refresh and after `dateRecord()` has captured each daily file’s original mtime/size/hash. If a closed daily file is compacted, the ledger still records the pre-compaction hash, so the next scheduled sweep re-queues that date and spends both LLM stages again. The index also remains stale until a later heartbeat. After settling, re-stat/hash affected daily files and update their processed records, then refresh the index once for all compaction changes.

src/lib/nightlyCompact.ts
```diff
@@ -0,0 +1,355 @@
...
+): Promise<CompactionOutcome> {
+  let bytesAfter = 0;
+  try {
+    bytesAfter = (await stat(candidate.absPath)).size;
+  } catch {
+    bytesAfter = 0;
+  }
+  const { status, note } = compactionVerdict(candidate, bytesAfter);
+  if (status === "reverted") {
+    await restoreFromArchive(archivePath, candidate.absPath);
```
kaieriksen, line 361: [P1] Rollback is not crash-safe or atomic

The model is instructed to rewrite files in place, and the safety rollback uses `copyFile()` directly over the live path. A process termination or filesystem failure during either operation can leave `MEMORY.md`, a drawer, or a daily file truncated/partially written; the guard then has no second recovery boundary. Use the repository’s atomic temp-file/fsync/rename discipline for model-visible rewrites and archive restores, and fail closed if the atomic restore cannot complete.

src/lib/nightly.ts
```diff
@@ -681,3 +702,78 @@ ${NIGHTLY_STAGE_BODY.distill(today)}
...
+ * LLM pass over them now would be built twice.
+ */
+const COMPACTION_KIND_BODY: Record<CompactionCandidate["kind"], string> = {
+  memory: `MEMORY.md is loaded into context on EVERY turn, so every byte is
+    rent. Prune it: drop "## Recent" bullets that are stale, superseded, or now
+    have a permanent home in a drawer or KB note; collapse anything that has
+    bloated into prose down to one line plus a [[pointer]]. Do NOT remove
+    standing facts about your owner, trust rules, or infrastructure state that
+    has no home elsewhere — move those, don't delete them.`,
+  drawer: `Do NOT rewrite this file. Its dedupe and lifecycle work is moving to
```
kaieriksen: [P2] Avoid waking an LLM for a drawer that this stage is forbidden to change

Drawers over 128 KiB are selected as candidates, but the prompt explicitly says “Do NOT rewrite this file.” They therefore remain over budget forever and trigger a compaction harness turn on every sweep while making no change. Either leave drawers out of this stage until their lifecycle implementation exists, or make the stage perform a bounded, verified drawer compaction; repeatedly paying for a no-op is not a safe steady-state behavior.

kaieriksen (changes requested):
Changes requested. PR #411 adds an unattended LLM writer for persona memory, so this cannot be dogfooded or merged yet. The six inline findings cover: compaction being unreachable on steady-state sweeps; indexed rollback archives leaking stale memory into retrieval; stale same-day rollback pre-images; daily ledger/index invalidation after rewrites; non-atomic rewrite/restore failure recovery; and repeated no-op LLM calls for drawers the prompt forbids it from changing.

The focused changed-area suite passes (109 tests / 251 assertions), typecheck passes, the production baseline build passes, and hosted CI/CodeQL are green. The full local suite reached 3,122 passed / 2 skipped, with 11 failures in existing ambient lock/Pi subprocess tests; those do not make the new compaction path safe. The implementation must preserve Phantom’s additive, non-destructive memory contract before dogfooding: no stale facts entering retrieval, no unintended reprocessing, and no window where a failed maintenance pass can damage persona memory.

src/cli/nightly.ts
robertclawson: Fixed in af5b5f1. Confirmed — and it is the worst of the six, because the sweep shape it excluded is the normal one. The early return is gone; an empty queue now just runs a zero-iteration date loop and falls through to the stage. Regression test 'STEADY STATE: runs with an empty queue' sweeps once to drain the backlog, then sweeps again with nothing pending and asserts the only turn spent is `compact`. Verified red against the old early return.

src/lib/nightlyCompact.ts
robertclawson: Fixed in af5b5f1. `walkMarkdown()` now skips an `archive/` directory at a scope root, so the pre-images never become index rows, and `refreshStale`'s live-vs-db reconcile drops any that were already indexed. Scoped to the root deliberately: a note the persona filed under `kb/projects/archive/` is one it chose to write, and there is a test for each half.

robertclawson: Fixed in af5b5f1. Every pass now archives its own pre-image and never reuses an earlier one — a same-day repeat lands beside its predecessor rather than over it, so nothing in the archive is lost either. Test writes 100 bytes, prunes legitimately to 70, then overshoots on a second pass and asserts the rollback lands on 70, not back on the original 100.

src/cli/nightly.ts
robertclawson: Fixed in af5b5f1. After settling, every daily candidate's ledger entry is re-fingerprinted from the file on disk (mtime/size/hash via `pendingForDate`), carrying `stages_done`/`status` over untouched — what changed is the file's fingerprint, not its history. The index is refreshed again too, gated on something actually having been rewritten. Test asserts the next sweep reports nothing pending and spends no `distill` turn; it goes red without the patch.

src/lib/nightlyCompact.ts
robertclawson: Fixed in af5b5f1. Restore goes through the existing `writeFileAtomic` (tempfile + fsync + rename + directory fsync) rather than `copyFile` over the live path. Agreed on the reasoning: a truncating write in the one code path that exists to make a bad pass recoverable is the wrong place to be non-atomic.

src/lib/nightly.ts
robertclawson: Fixed in af5b5f1, and you talked me out of the design rather than the implementation. Drawers are no longer candidates at all — `defaultBudgets()` is MEMORY.md only, and `measureDrawers()` reports the overage for one `stat` each, printed in the sweep output. Selecting a file and then telling the prompt not to touch it was a paid no-op every night; their dedupe lifecycle belongs in the follow-up database work where it is a uniqueness constraint. The `drawer` prompt body survives as a defensive entry only.

src/cli/nightly.ts
```diff
@@ -203,6 +224,103 @@ export async function runNightlyTurn(opts: {
...
+  // the daily file. Compaction rewrites that file — and so does a rollback —
+  // so leaving the pre-compaction record in place makes the NEXT sweep see the
+  // date as `changed`, re-queue it, and pay for both LLM stages again on a day
+  // whose content was only ever removed. Left alone it re-bills every single
+  // night, forever. The distillation verdict (stages_done/status) is carried
+  // over untouched: what changed is the file's fingerprint, not its history.
+  const touched = outcomes.filter(
+    (o) => o.kind === "daily" && o.bytesAfter !== o.bytesBefore,
+  ).length;
+  const rewritten = outcomes.some((o) => o.status !== "unchanged");
```
kaieriksen: [P1] Same-size rewrites bypass the post-compaction index refresh

`unchanged` only means the byte count is unchanged; the model can rewrite a file to different content with exactly the same length. In that case `rewritten` is false, line 315 skips reconciliation, and FTS/vector search keeps serving the pre-compaction text even though the live file changed. I reproduced this by replacing a 20 KiB `MEMORY.md` with different 20 KiB content: the pre-stage refresh ran, but the post-stage refresh did not. Compare content/hash (the archive already provides the pre-image) or conservatively reconcile whenever candidates were handed to the stage.

src/cli/nightly.ts
```diff
@@ -291,23 +409,22 @@ export async function runNightly(input: RunNightlyInput = {}): Promise<number> {
     cap !== undefined && cap > 0 ? Math.max(0, queue.length - cap) : 0;
   if (deferred > 0) queue = queue.slice(0, cap);
 
+  // An empty queue is NOT an early return. Compaction's inputs are whole-file
+  // sizes, not a day's events, so the night it most needs to run is exactly
+  // the steady-state night where nothing new is pending — which is every night
+  // once the backlog is drained. Returning here made the stage permanently
+  // inert outside a backfill. The date loop below simply does nothing instead.
   if (queue.length === 0) {
```
kaieriksen, line 424: [P1] Steady-state compaction now runs outside the in-flight sweep lock

With an empty queue this path never writes `state.current`; that marker is only created inside the date loop. Two scheduled/manual sweeps can therefore both enter compaction, archive the same pre-image, and run concurrent LLM writers against the same memory files. I reproduced it by blocking the first empty-queue compact turn and starting a second sweep: the second also reached `compact` instead of reporting an in-flight sweep. Acquire/update the sweep marker before any steady-state compaction work (and keep it fresh through the turn), then clear it through the existing completion path.

kaieriksen (changes requested):
Follow-up review at `af5b5f1`: the six original findings are addressed, and the focused suite passes (112 tests / 265 assertions), typecheck passes, the baseline production build passes, and hosted Linux/Windows CI plus CodeQL are green. Two new P1 findings still block dogfooding: empty-queue compaction never acquires the sweep lock, allowing concurrent LLM writers; and same-byte-length rewrites are classified as unchanged, so post-compaction index reconciliation is skipped and stale memory remains searchable. Both are reproduced with focused failing regression tests in the review workspace.

src/cli/nightly.ts
robertclawson: Fixed in e03c834. `unchanged` was a byte-count verdict being read as a content verdict. The stage only runs when there were candidates at all (the no-candidate case returns before this), so the reconcile is now unconditional rather than hash-compared: a refresh over genuinely untouched files is a stat-cheap no-op, a missed one is stale memory that stays searchable — the asymmetry is not close. Regression test `a same-size rewrite still reconciles the index` rewrites a 20 KiB MEMORY.md to different 20 KiB content, asserts the verdict is still `unchanged`, and asserts the post-stage refresh ran; it fails with the guard reverted.

src/cli/nightly.ts
robertclawson: Fixed in e03c834. The marker is now acquired before any work (`holdSweep(queue[0]?.date ?? today, 0)`), refreshed as each date starts, and refreshed once more immediately before the compaction stage so a turn that outlives the 45-minute stall timer does not read as stalled to an observer. It is still cleared through the existing `finally`. Regression test `STEADY STATE: the sweep marker is held across an empty-queue compaction` drains the backlog, parks a second sweep's compact turn on a promise, and asserts a concurrent sweep reports `already in flight` and never reaches `compact` — then that the marker is cleared on the way out. Fails with the pre-loop acquire removed.

src/cli/nightly.ts
```diff
@@ -415,6 +553,71 @@ export async function runNightly(input: RunNightlyInput = {}): Promise<number> {
         out.write(`nightly: ${pending.date} ok (${Date.now() - t0}ms)\n`);
       }
     }
+
+    // ---------------------------------------------------------------------
+    // Stage three: compaction (issue #410).
+    //
+    // Runs ONCE per sweep, after every date has been distilled, because its
```
kaieriksen, line 560: [P1] Do not compact after a failed date stage

At this point `errors` can already contain a distill/kb failure, but the destructive compaction pass still runs. A distill turn can partially rewrite MEMORY.md before returning an error; this block then archives that already-damaged post-failure version and starts a second model rewrite, so the archive cannot recover the pre-sweep state. This also violates the issue’s stated guard to never compact a file when a stage failed that night. Gate compaction on the sweep having no stage errors (or enforce an equivalent candidate-specific dependency), and add a regression where distill mutates MEMORY.md then returns `errored`: compact must not be called and no compaction ledger should be written.

src/cli/nightly.ts
```diff
@@ -466,6 +669,12 @@ export default defineCommand({
       description: "Run even if another sweep holds the in-flight marker.",
```
kaieriksen, line 687: [P2] Update the CLI description for steady-state compaction

The command still promises that re-running with nothing pending does nothing, but this PR intentionally makes an empty date queue fall through to compaction and potentially rewrite over-budget files. This is user-facing help text and now contradicts both the implementation and README. Describe the date sweep as idempotent while noting that the once-per-sweep compaction check still runs.

kaieriksen (changes requested):
The two requested fixes are correct and covered, but the destructive stage still runs after a failed distill/kb stage, when its archive can no longer represent the clean pre-sweep state. The CLI help also still claims an empty date queue does nothing. Inline findings posted.

src/cli/nightly.ts
robertclawson: Agreed and fixed in 98e3bc8. Compaction is now gated on `errors.length === 0` for the sweep: any distill/kb stage failure skips the stage entirely, writes `nightly: compaction skipped — N stage error(s) this sweep` to stdout and a warn to the log, and leaves over-budget files for the next clean sweep. Regression added — `a failed date stage stops compaction from touching the file` has distill rewrite MEMORY.md to `half-written` then return `errored`, and asserts no `compact` turn, no `compaction` ledger, and the file untouched. Verified it goes red with the guard reverted.

src/cli/nightly.ts
robertclawson: Fixed in 98e3bc8. The command description now reads: the date sweep is idempotent (re-running with nothing pending distills nothing), while the once-per-sweep compaction check still runs and rewrites a file only if it is over budget, with a pointer to `--no-compact`. README updated in the same commit to note the new failed-stage skip.

kaieriksen (approved):
Approved at 98e3bc8 after a bounded re-review of the requested fixes and their direct consequences. Blockers: none. Majors: none. Minors: none. Nits: none. The failed-stage gate correctly prevents compaction after any distill or KB failure, the CLI help and README now match steady-state behavior, all prior safety threads are resolved, the focused changed-area suite passes 115 tests / 278 assertions, typecheck and the baseline production build pass, and hosted Linux, Windows, typecheck, and CodeQL checks are green. Safe to dogfood.

PR #412, opened 2026-08-20
feat(memory): drawer entries as rows — identity, supersession, weight, decay

robertclawson:
Part of #410 — stage two. The [schema spec](https://github.com/phantomyard/phantombot/issues/410#issuecomment-5358544762) is on the issue; this is the implementation of the storage half.

## Why

The five drawers are append-only markdown. The nightly's step-2 prompt tells the model "do not duplicate what is already filed — read the drawer first", which is not a followable instruction once the file is 663 KB, so the dedupe silently does not happen: 64 exact duplicate decisions, 878 commitments with no way to tell a live one from a discharged one. #411 gave the nightly a compaction half but deliberately skipped the drawers, because an LLM dedupe pass over prose gets thrown away the moment the entries are rows.

## What lands

`src/memory/drawers.ts` — `drawer_entries`, one table for all five drawers:

- **Identity.** `id = sha256(persona, kind, normalizeFact(content))[..16]` behind `UNIQUE (persona, kind, content_norm)`. Re-filing REAFFIRMS one row instead of appending a copy: `asserted_at` is kept, `last_reaffirmed_at` only ever moves forward, and the higher weight and higher-trust source win. Dedupe becomes a constraint the database enforces for free.
- **Supersession.** A newer entry names the id it replaces; the old row goes `superseded` and leaves the injection pool. This is how a third-party tool corrects a norm it filed earlier — it never edits or deletes.
- **Lifecycle.** `status` per kind, enforced on write. `discharged` and `expired` are distinct: "we missed it" is not "we did it".
- **Decay, for beliefs only.** `weight · 2^(-ageDays / halfLifeDays)` off `last_reaffirmed_at` — the same shape the durable-fact tiers in `config.ts` already use, so there is one decay model here and not two. Norms 365d, decisions 180d, lessons 120d; under `DORMANT_FLOOR` an entry goes `dormant` and any reaffirmation revives it.

`src/memory/drawerIngest.ts` — parses the existing markdown into rows. Handles both shapes in the wild (`###` blocks with bodies, and the heartbeat's flat `- - [tag] …` bullets), dates entries from their `## YYYY-MM-DD` header, and **never rewrites or deletes the file**. Idempotent, so it is safe to re-run while both representations coexist.

## The two things reviewers should push on

**1. `commitments` and `people` do not decay, by construction.** This is the load-bearing decision, and getting it wrong loses memory silently rather than loudly. A commitment is open, discharged or overdue — age makes it MORE urgent, so decaying it buries the one thing that needed surfacing. A person fact is not less true for going a year unmentioned. Both move by `status` alone; `sweepDormant()` refuses to walk them; `BELIEF_KINDS` is the single point of truth and a test asserts its exact membership.

**2. Nothing is ever deleted.** Same rule #411 runs on. `superseded`, `dormant`, `discharged` and `expired` rows stay queryable forever and merely stop being *injected*. Retirement is a ranking decision, never a `DELETE`.

## Dogfood — dry run against a real persona

Read-only, in-memory DB, against the live drawers this issue was opened about:

```
people.md:      94 parsed,  94 new
decisions.md: 1112 parsed, 1112 new
lessons.md:    981 parsed,  981 new
commitments.md: 842 parsed, 834 new, 8 already filed
norms.md:       81 parsed,  81 new
```

The 8 collapsed commitments match the exact-duplicate count in the issue body. Entry granularity is the whole `###` block, so near-duplicates that differ by a restated timestamp still land as distinct rows — near-dupe merge is a later stage.

## Tests

23 new tests. The negative ones are the point: commitments and people surviving ten years without decaying, a dormant entry still being queryable after it stops ranking, a superseded entry NOT being revived by a re-file, and the ingest leaving the markdown byte-identical. Guards were verified by mutation — adding `commitments` to `BELIEF_KINDS` reddens two tests, and reviving on re-file reddens one.

`bun test`: 3167 pass, 0 fail. `bun tsc --noEmit`: clean.

## Docs

- **`docs/memory-drawers.md`** (new) — the model, the per-kind status table, the half-lives, and a "writing a norm from a custom tool" section with the file-and-supersede pattern, since third-party phantomtools are the main external consumer of this contract.
- **`docs/architecture.md`** — memory subsystem section (also: the drawers are five, not four).
- **`AGENTS.md`** — repo layout plus a new common pitfall covering the decay split, the never-delete rule and the revive asymmetry.

## Not in this PR

Capture/heartbeat writing rows, search over rows, retiring the markdown files, near-duplicate merge, and moving the "read yesterday's daily" rule out of `AGENTS.md` prose into ledger-driven code. All tracked on #410.


src/memory/drawers.ts
```diff
@@ -0,0 +1,446 @@
...
+   * every startup would silently resurrect what replaced it.
+   */
+  file(input: FileDrawerEntryInput): DrawerEntry {
+    const now = new Date();
+    const assertedAt = input.assertedAt ?? now;
+    const id = drawerEntryId(input.persona, input.kind, input.content);
+    const existing = this.get(id);
+    if (existing) {
+      const weight = Math.max(existing.weight, input.weight ?? DEFAULT_WEIGHT);
+      const source = higherTrust(existing.source, input.source ?? "principal");
```
andrewagrahamhodges: **Trust and weight escalate whenever the caller omits them.**

`input.source ?? "principal"` is the top of `TRUST_ORDER`, so `higherTrust` always picks it — a re-file with no `source` promotes the row to `principal` regardless of what it was. Same shape on the line above: `input.weight ?? DEFAULT_WEIGHT` means an omitted weight raises a deliberately-zeroed entry back to 1, and since it is a `max`, a weight can never be lowered at all (the doc-comment's "0 is legal and means never inject" is unreachable after the first re-file).

Probed on this branch:

```
file({persona:"p", kind:"norms", content:"a norm", source:"unverified", weight:0})
  -> weight 0, source unverified
file({persona:"p", kind:"norms", content:"a norm"})   // no source, no weight
  -> weight 1, source principal
```

The fix is to fall back to the row, not to the top tier:

```ts
const source = higherTrust(existing.source, input.source ?? existing.source);
const weight = input.weight ?? existing.weight;
```

(or keep `max` on weight but default it to `existing.weight`). A reaffirmation should be evidence that the entry is still *live* — that is what `last_reaffirmed_at` already records — not evidence that it is more trusted than when it was filed.

src/memory/drawers.ts
```diff
@@ -0,0 +1,446 @@
...
+        input.content,
+        normalizeFact(input.content),
+        input.weight ?? DEFAULT_WEIGHT,
+        input.supersedes ?? null,
+        input.source ?? "principal",
+        input.origin ?? null,
+        assertedAt.toISOString(),
+        assertedAt.toISOString(),
+      );
+    if (input.supersedes) this.markSuperseded(input.persona, input.supersedes);
```
andrewagrahamhodges: **No guard against `supersedes === id`.** Confirmed on the branch: `file({persona:"p", kind:"lessons", content:"self", supersedes: drawerEntryId("p","lessons","self")})` returns an entry whose status is immediately `superseded` — never injected, and per the (correct) revive asymmetry documented on `file()`, a later re-file will not bring it back. The entry is gone, silently.

This is reachable by the exact caller the docs target: a tool files a corrected norm, computes the old id from the text it is replacing, and the correction happens to normalize to the same string (`normalizeFact` folds case, whitespace and trailing punctuation) — so it erases itself.

Same line, second issue: `markSuperseded` keys on `(persona, id)` only, so a `people` entry can supersede a `decisions` entry. Also confirmed. Supersession should be within a kind.

Suggest rejecting self-supersession outright (throw — it can only be a caller bug) and adding `AND kind = ?` to `markSuperseded`.

src/memory/drawers.ts
```diff
@@ -0,0 +1,446 @@
...
+          id,
+        );
+      if (input.supersedes) {
+        this.markSuperseded(input.persona, input.supersedes);
+      }
+      return this.get(id)!;
+    }
+    this.db
+      .query(
+        `INSERT INTO drawer_entries
```
andrewagrahamhodges, line 408: **`get()` → `INSERT` is not atomic, and this repo runs concurrent turns per persona (#391).** Two turns filing the same entry both miss on `get()`, and the loser hits `SQLITE_CONSTRAINT` on `UNIQUE (persona, kind, content_norm)` — an exception on what should be the idempotent path, and the whole selling point here is that re-filing is safe from any caller.

One statement removes the window and the pre-read:

```sql
INSERT INTO drawer_entries (...) VALUES (...)
ON CONFLICT (persona, kind, content_norm) DO UPDATE SET
  weight = MAX(weight, excluded.weight),
  last_reaffirmed_at = MAX(last_reaffirmed_at, excluded.last_reaffirmed_at),
  status = CASE WHEN status = 'dormant' THEN 'active' ELSE status END,
  ...
```

If you would rather keep the merge logic in TypeScript, wrapping `file()` in `db.transaction(...)` is enough — but then the reaffirm branch needs to be reachable from the constraint failure too, not only from the `get()` hit.

src/memory/drawers.ts
```diff
@@ -0,0 +1,446 @@
...
+ * ingest merge by id rather than blind-append, and lets a tool name the entry
+ * it supersedes without ever having seen the row.
+ */
+export function drawerEntryId(
+  persona: string,
+  kind: DrawerKind,
+  content: string,
+): string {
+  return createHash("sha256")
+    .update(`${persona} ${kind} ${normalizeFact(content)}`)
```
andrewagrahamhodges: **These separators are two raw NUL bytes, not the `\0` escape.** `file(1)` reports this module as `data` rather than JavaScript source, and `grep`/`git grep` skip it with "binary file matches" — I had to `tr` a copy to read line numbers for this review.

NUL is a good domain separator (a space would let `persona="a b"`, `kind="c"` collide with `persona="a"`, `kind="b c"`), so keep the semantics, just write it as an escape:

```ts
.update(`${persona}\0${kind}\0${normalizeFact(content)}`)
```

Worth doing beyond tidiness: any tool that strips control characters on save silently changes every id `drawerEntryId()` returns from then on. Stored rows keep their old ids, so supersession by id from an external tool starts missing — and it misses the way `markSuperseded` is written to tolerate, i.e. silently.

src/memory/drawers.ts
```diff
@@ -0,0 +1,446 @@
...
+  weight             REAL NOT NULL DEFAULT 1,
+  status             TEXT NOT NULL DEFAULT 'active',
+  supersedes         TEXT,
+  source             TEXT NOT NULL DEFAULT 'principal',
+  origin             TEXT,
+  asserted_at        TEXT NOT NULL,
+  last_reaffirmed_at TEXT NOT NULL,
+  UNIQUE (persona, kind, content_norm)
+);
+CREATE INDEX IF NOT EXISTS idx_drawer_entries_rank
```
andrewagrahamhodges, line 209: Nit: `idx_drawer_entries_rank` is `(persona, kind, status, last_reaffirmed_at DESC)`, but no query in this file can use it — `list()` filters on `(persona, kind)` and orders by `asserted_at`, and `ranked()` calls `list()` then filters status and sorts by score in JS. On a drawer at the size #410 describes (~1400 entries) `ranked()` is a full scan plus a JS sort on every call.

Either push `status = 'active'` and the ordering down into the SQL so the index earns its place, or drop the index until a caller needs it.

src/memory/drawerIngest.ts
```diff
@@ -0,0 +1,211 @@
...
+      date = undefined;
+      continue;
+    }
+    const entryHeader = ENTRY_HEADER.exec(line);
+    if (entryHeader) {
+      flush();
+      heading = entryHeader[1]!.trim();
+      continue;
+    }
+    if (heading !== undefined) {
```
andrewagrahamhodges: **Bullets that follow a `###` block are swallowed into that block.** `heading` stays set until the next `##`/`###`, so every line after the block body — including top-level bullets — falls into this branch and is appended to the previous entry's content. The module header advertises both shapes as supported in the wild, so a drawer that mixes them loses entries.

On this branch:

```
### Block entry
body line

- bullet after the block
- second bullet
```

parses to **one** entry, `"Block entry\nbody line\n\n- bullet after the block\n- second bullet"` — the two bullets never become rows. And because the id is a content hash, that merged blob will not dedupe against the same bullets filed standalone from another file or another persona's ingest, so a later re-run does not repair it.

A blank line followed by a top-level `- ` is the natural terminator: flush the block when a `FLAT_BULLET` matches at column 0 while inside one.

src/memory/drawerIngest.ts
```diff
@@ -0,0 +1,211 @@
...
+      body.push(line);
+      continue;
+    }
+    // Outside any `###` block: a top-level bullet is an entry on its own. The
+    // heartbeat writes `- - [norm] …`, so strip the doubled dash.
+    const bullet = FLAT_BULLET.exec(line.trim());
```
andrewagrahamhodges: `line.trim()` before the bullet match means indented sub-bullets are promoted to standalone entries, detached from the bullet they qualify:

```
- parent bullet:
  - child a
  - child b
```

parses to three entries — `"parent bullet:"`, `"child a"`, `"child b"`. `"child a"` on its own is not a norm anybody can act on, and it now has a stable id and a decay clock of its own.

Match on `raw` rather than the trimmed line so only column-0 bullets start an entry, and attach indented continuation lines (bullets or not) to the open one.

src/memory/drawerIngest.ts
```diff
@@ -0,0 +1,211 @@
...
+  const text = await readFile(abs, "utf8");
+  const fallback = await fileDate(abs, now);
+  const entries = parseDrawer(text);
+  result.parsed = entries.length;
+
+  for (const entry of entries) {
+    const assertedAt = entryDate(entry.date, fallback, now);
+    // Ask by id rather than diffing list() lengths: a drawer runs to ~1400
+    // entries, and re-listing per entry would make the migration quadratic.
+    const existed = store.get(drawerEntryId(persona, kind, entry.content));
```
andrewagrahamhodges: The ingest passes no `source`, so every parsed line lands as `principal` — see the comment on `drawers.ts:301`. These drawers are exactly where the nightly promoted entries out of `origin:"task"` turns, which the fact pool deliberately holds at `unverified`, so this migration flattens that distinction in one pass and there is no way to recover it afterwards.

At minimum pass `source: "self"` here (the persona's own filed belief, not the principal's first-hand assertion). Fixing the default in `file()` does not cover this on its own, since the first-ever insert has no existing row to fall back to.

Unrelated, same call: `store.get()` here duplicates the `get()` that `file()` does one line later. If `file()` returned whether it inserted or reaffirmed, this would be one query instead of two per entry across ~4500 rows.

andrewagrahamhodges (changes requested):
The shape is right and the two things you flagged for pushback both hold up: `BELIEF_KINDS` as the single point of truth is the correct place for the decay split, and the mutation check (adding `commitments` to it reddens two tests) is what makes that guard real rather than decorative. Never-delete is consistent with #411. Ingest leaving the markdown byte-identical is the right call while both representations coexist.

I ran the branch: 23/23 new tests pass, `bun tsc --noEmit` clean, and I probed `file()` and `parseDrawer()` directly. Four things I'd want fixed before this lands, because they are silent — nothing throws, the row just ends up wrong.

**1. Trust and weight escalate by omission.** `file()` defaults a missing `source` to `principal` and a missing `weight` to `1`, then takes the max of each. So *any* re-file that omits those fields promotes an `unverified`, weight-0 entry to `principal`, weight 1. Verified:

```
file({source:"unverified", weight:0})  -> weight 0, source unverified
file({})            // re-file       -> weight 1, source principal
```

This matters most for the ingest, which omits `source` entirely — so every line of markdown, including entries the nightly promoted out of `origin:"task"` turns, is filed at the top trust tier. And the docs sell third-party phantomtools as the main external consumer of this contract, which is exactly the caller that will omit `source`.

**2. `supersedes` has no self- or cross-kind guard.** `file({supersedes: <its own id>})` marks the entry it just wrote as `superseded`, and by your own (correct) revive asymmetry a later re-file will never bring it back. A `people` entry can also supersede a `decisions` entry. Both confirmed on the branch.

**3. `file()` is a TOCTOU under the concurrency this repo already has.** `get()` then a bare `INSERT` is not atomic; two turns for one persona (#391) filing the same entry race into `SQLITE_CONSTRAINT`. `INSERT … ON CONFLICT (persona, kind, content_norm) DO UPDATE` collapses both paths and removes the pre-read.

**4. The parser merges entries when a file mixes the two shapes it claims to support.** Bullets that follow a `###` block get absorbed into that block's body, and nested sub-bullets are promoted to standalone entries. Details inline.

Rest of the comments are minor. Docs are good — `docs/memory-drawers.md` is the right level of detail for an external tool author, and the AGENTS.md pitfall entry earns its place.

src/memory/drawers.ts
robertclawson: Fixed in e7036e8. Both fall back to the row now — `input.source ?? existing.source`, `input.weight ?? existing.weight` (still a `max`, so an explicit higher value raises, an omission cannot). Also changed the first-insert default: `DEFAULT_SOURCE = "self"`, and the schema default with it, so a caller that never names a source cannot land in the top tier at all. Regression covers unverified/weight-0 surviving a bare re-file, an explicit raise still working, and the new insert default.

src/memory/drawers.ts
robertclawson: Fixed in e7036e8. `supersedes === id` now throws before any write — it can only be a caller bug, and the accidental route you describe (a correction that normalizes to the original) is exactly why it must not be silent. `markSuperseded` also takes the kind now and filters on it, so a `people` id can no longer retire a `decisions` row. Two regressions: the throw leaves nothing written, and a cross-kind supersede leaves the target `active`.

src/memory/drawers.ts
robertclawson: Fixed in e7036e8. `file()` delegates to `fileEntry()`, which runs the whole read-modify-write inside one IMMEDIATE transaction, and a `UNIQUE constraint failed` from another connection is retried onto the same path — where the `get()` now hits and it reaffirms. Kept the merge logic in TypeScript as you allowed, but the reaffirm branch is reachable from the constraint failure, not only from the `get()` hit.

src/memory/drawers.ts
robertclawson: Fixed in e7036e8 — written as `\0` escapes, same bytes, and the file reads as source to `file(1)` and `git grep` again. Your point about a control-character-stripping tool silently rotating every id is the part that mattered; kept the domain separation for the collision reason you give.

src/memory/drawers.ts
robertclawson: Fixed in e7036e8 by making the index earn its place rather than dropping it: `ranked()` and `sweepDormant()` now go through a `listActive()` that selects `status = 'active'` and orders by `last_reaffirmed_at DESC` in SQL. The final sort is still by decayed score in JS, since weight is part of it, but the scan and the status filter are gone.

src/memory/drawerIngest.ts
robertclawson: Fixed in e7036e8, with your terminator rule: a column-0 bullet breaks out of a `###` block only when the preceding line was blank. A bullet running straight on from the block body is still a detail bullet of that entry, which is the dominant shape in the real drawers (the existing fixture depends on it). Regression covers both directions.

src/memory/drawerIngest.ts
robertclawson: Fixed in e7036e8 — the bullet match is against the untrimmed line, so only column-0 bullets start an entry, and indented lines attach to the open one. `- parent:` / `  - child a` / `  - child b` now parses to one entry.

src/memory/drawerIngest.ts
robertclawson: Fixed in e7036e8 — the ingest passes `source: "self"` explicitly, and the comment says why (the nightly promoted plenty of these out of `origin: "task"` turns, and there is nothing on disk to recover the distinction from afterwards). The duplicate read is gone too: `store.fileEntry()` returns `{ entry, inserted }`, so the counts cost no extra query.

robertclawson:
All six pushed back on nothing — every one was real and every one was silent. Fixed in `e7036e8`, one commit, no force-push.

1. **Trust/weight escalation by omission** — omitted `weight`/`source` now fall back to the existing row, not to `1`/`principal`. On top of your fix I moved the first-insert default to `self` (`DEFAULT_SOURCE`, and the schema default with it): the top tier should have to be claimed, and the caller the docs target is precisely the one that omits it.
2. **Self- and cross-kind supersession** — `supersedes === id` throws before any write; `markSuperseded` is scoped to `(persona, kind, id)`.
3. **TOCTOU** — `file()` runs its read-modify-write in one IMMEDIATE transaction, and a `UNIQUE` collision from another connection is retried onto the reaffirm path, so the idempotent path stays idempotent under #391 concurrency.
4. **Raw NUL separators** — now `\0` escapes; same bytes, file reads as source again.
5. **Parser** — a column-0 bullet after a blank line breaks out of a `###` block (a bullet running straight on from the body is still a detail bullet, which the real drawers depend on); the bullet match is against the untrimmed line, so indented sub-bullets stay attached to their parent.
6. **Ingest** — files as `source: "self"`, and uses the new `fileEntry()` `{ entry, inserted }` outcome instead of a duplicate `get()` per entry across ~4500 rows.
7. **Index nit** — made the index earn its place instead of dropping it: `ranked()`/`sweepDormant()` select `status = 'active'` ordered by `last_reaffirmed_at DESC` in SQL.

8 new regression tests (31 in the drawer file). Each new test was checked against the pre-fix behaviour by reverting the guard and confirming it goes red — including the parser one, where the naive "always break the block" rule reddens the existing fixture instead. Full suite: 3175 pass / 0 fail, `bun tsc --noEmit` clean, prettier clean on both touched source files.

Docs updated: the weight section is now "Weight and trust" and states the omission rule explicitly, and the tool-facing supersession list gains the kind-scoping and the self-supersession throw.

andrewagrahamhodges (approved):
Approving — all six findings are fixed, and I verified each against the code rather than the commit message.

**1. Trust/weight by omission.** `fileNow` now falls back to `existing.weight` / `existing.source` on reaffirm, and a first insert lands at `DEFAULT_SOURCE = "self"`. `higherTrust`/`Math.max` mean the pair is monotonically non-decreasing, which is the right direction: a re-file can't launder a weaker source into a stronger one. Worth knowing that the corollary is a wrongly-filed `principal` can only be corrected by supersession, never by re-filing — fine, just no longer a silent path.

**2. Self-supersession.** Throws now, with the id in the message, and `markSuperseded` is scoped `WHERE persona = ? AND kind = ? AND id = ?`, so a `people` id can't retire a `decisions` row. The throw is the right call over a silent no-op: the caller computed a bad id and needs to know.

**3. TOCTOU.** Read-modify-write is inside one IMMEDIATE transaction, with a UNIQUE collision from another connection retried onto the reaffirm path. Idempotent from any caller, which is what `docs/memory-drawers.md` promises third-party tools.

**4. Parser.** Ran the four shapes from my last review against `parseDrawer` directly:

| input | result |
|---|---|
| `###` block, blank line, flat bullets | 3 entries — block no longer swallows them |
| `###` block with detail bullets (no blank) | 1 entry, bullets kept as body |
| bullet with indented sub-bullets | 1 entry, children attached |
| bullet, blank, bullet | 2 entries |

Then against my real drawers: norms 34, decisions 268, lessons 279, commitments 14, people 20 — zero sub-12-byte fragments, zero over-2KB merges, zero undated. Clean.

**5. NUL separators.** `\0` escapes in source; `drawers.ts` reads as text to `file(1)` and `git grep` again. Hash input is byte-identical, so no id churn.

**6. Index.** `EXPLAIN QUERY PLAN` on the `listActive` query against the real DDL:

```
SEARCH drawer_entries USING INDEX idx_drawer_entries_rank (persona=? AND kind=? AND status=?)
```

It earns its place now.

31 drawer tests pass, `tsc --noEmit` clean.

The decay split still reads right to me: ageing a commitment out of the prompt is the one failure mode that actually costs something, and excluding `commitments`/`people` by construction (a list plus an assertion, not a comment) is the version of that which survives a refactor.

PR #414, opened 2026-08-20
fix(memory): inject the daily journal whole — drop the 8KB recall cap

robertclawson:
## What

The daily journal now reaches the prompt **whole**. The 8 KB cap on
`buildDailyRecall` is gone.

## Why

The recall cap was deliberately pinned to the persona's daily *compaction*
budget so the two numbers "cannot drift". They should drift — they measure
different things:

- **compaction budget** — how large a CLOSED, fully distilled day may stay on
  disk. 30-day age gate, content already promoted to the drawers. 8 KB is right.
- **recall cap** — how much of an ACTIVE day a turn is allowed to see. The open
  day is the only place its content exists; nothing has been distilled out of it.

Observed on this fleet: a 34 KB day lost 26 KB off its front. What fell out was
not narration — 92% of journal bytes are tagged captures (`[decision]`,
`[lesson]`, `[commitment]`) on their way to the drawers, i.e. the load-bearing
part. The block does say "run `phantombot memory get …`", but that only fires
when the turn *recognises* something is missing, and it usually cannot: the
clipped content is invisible by construction.

The cost of not capping is bounded and knowable. The largest daily ever written
across this fleet is ~62 KB (~15k tokens), and the file resets at midnight UTC.
Paying that reliably beats paying nothing and being confidently wrong.

## What is NOT changed

- The compaction budget (8 KB, `memory/*.md` override) — untouched, still
  applies on disk.
- `buildDailyRecall(dir, now, maxBytes)` — the explicit parameter stays, and
  tail-keeping plus its `log.warn` are unchanged for callers that pass one.
- Which journals get injected (today always, yesterday only when its sweep did
  not finish) — unchanged.

## Tests

Two new guards, both mutation-verified:

- `a large file lands WHOLE — there is no default cap` — goes red if a default
  cap is reintroduced.
- `the daily COMPACTION budget no longer clips the prompt` — goes red if recall
  is re-coupled to `resolveDailyBudgetBytes`.

Reverting the change makes exactly those two fail; `bun test
tests/lib-dailyRecall.test.ts` is 23/23 green otherwise, and `tsc --noEmit` is
clean. Full suite has 9 pre-existing failures (process-spawn lock tests),
identical on clean `main`.


src/lib/dailyRecall.ts
```diff
@@ -164,7 +169,7 @@ export async function buildDailyRecall(
 ): Promise<DailyRecallDecision> {
   const todayKey = dateKey(now, 0);
   const yKey = dateKey(now, 1);
-  const cap = maxBytes ?? (await resolveDailyBudgetBytes(personaDir));
+  const cap = maxBytes;
```
andrewagrahamhodges: **Blocking:** this removes the last ceiling on an agent-writable input to the system prompt.

`memory capture` (`src/cli/memory.ts:430`) does not limit `text` or the number of lines per day, and `writeCommitmentToDaily` (`src/cli/task.ts:291`) appends too. With `cap === undefined` in the only production caller, prompt size for a persona is now bounded only by how much that persona wrote today.

The recovery path is what worries me more than the steady state. A single capture holding a pasted log dump, or a recurring task looping on capture, oversizes *every* subsequent turn — including the turn that would fix it, which dies at prompt-build for the same reason. Before this PR that degraded to tail + `log.warn` and stayed operable.

`~62KB is the largest ever written across this fleet` is an observation about the past, not a bound on the future, and this module was the thing enforcing one.

Suggestion:

```ts
/**
 * Safety ceiling, NOT a content budget. ~4x the largest daily ever written
 * on this fleet, so no real day is ever clipped; it exists only so an
 * unbounded writer cannot make every turn unbuildable.
 */
export const DAILY_RECALL_CEILING_BYTES = 256 * 1024;
...
const cap = maxBytes ?? DAILY_RECALL_CEILING_BYTES;
```

That keeps everything this PR argues for (a heavy real day goes in whole, no coupling to the compaction budget) and keeps the tail-keep, the `log.warn` with `droppedBytes`, and the `memory get` recovery line for the pathological case. Worth a third guard asserting the ceiling still bites at 1 MB — the two new tests both pass with any large default cap in place, so nothing currently covers that direction.

src/lib/dailyRecall.ts
```diff
@@ -54,25 +55,28 @@ import {
...
  *
- * It is a real cap, not a backstop: the sweep only ever compacts a daily file
- * that is fully distilled AND at least `DAILY_MIN_AGE_DAYS` old, so neither of
- * the two files this module reads is ever shrunk by it. A persona that writes
- * more than the budget in a day loses the START of that day from the prompt —
- * hence the `log.warn` on every truncation, so the loss is visible rather than
- * silent, and the recovery command inside the block itself.
+ * Callers that genuinely need a ceiling pass `maxBytes` explicitly; the
+ * tail-keeping behaviour and its warning are unchanged for them.
  */
```
andrewagrahamhodges, line 86: This doc comment is now orphaned — `export const DAILY_RECALL_MAX_BYTES` was the declaration it documented, and with that gone it floats above a blank line and a second doc comment for `YesterdayReason`. It is good prose and worth keeping, but as a `//` note or folded into the module header at the top (where line 34 already states the rule), not as a `/** */` block attached to nothing.

While you are here: `readCapped` no longer caps by default. `readDaily` reads truer.

src/lib/dailyRecall.ts
```diff
@@ -54,25 +55,28 @@ import {
...
+ * written across this fleet is ~62KB (~15k tokens), and the day is reset every
+ * midnight UTC. Paying that reliably beats paying nothing and being wrong.
  *
- * It is a real cap, not a backstop: the sweep only ever compacts a daily file
- * that is fully distilled AND at least `DAILY_MIN_AGE_DAYS` old, so neither of
- * the two files this module reads is ever shrunk by it. A persona that writes
- * more than the budget in a day loses the START of that day from the prompt —
- * hence the `log.warn` on every truncation, so the loss is visible rather than
- * silent, and the recovery command inside the block itself.
+ * Callers that genuinely need a ceiling pass `maxBytes` explicitly; the
```
andrewagrahamhodges: "Callers that genuinely need a ceiling pass `maxBytes` explicitly" — there is exactly one production caller, `orchestrator/turn.ts:435`, and it passes nothing. After this PR the parameter is exercised only by tests, so the sentence describes a caller that does not exist.

Either wire the ceiling at the caller (which would also make the choice visible on the prompt path where the cost is paid), or say plainly that there is no ceiling in production today.

src/lib/dailyRecall.ts
```diff
@@ -31,7 +31,8 @@
  *
  *   - No config switch. A persona cannot turn this off, because a persona
  *     that turns it off loses a day of memory the first time a sweep fails
- *     and has no way to notice. The only tuning is the byte cap.
+ *     and has no way to notice. There is no byte cap either: an open
```
andrewagrahamhodges: Small precision point on the module header: `There is no byte cap either: an open day goes in whole` reads as a property of the module, but it is a property of the *default*. `buildDailyRecall` still caps when handed `maxBytes`. Suggest `No default byte cap: an open day goes in whole unless a caller passes an explicit ceiling.` — which also stays accurate if the ceiling from my other comment lands.

src/lib/dailyRecall.ts
andrewagrahamhodges: Nit, separate from the blocking comment on this line: `const cap = maxBytes;` is a pure alias now that nothing is resolved into it. It earned its name when it held the resolved budget. Either inline `maxBytes` at the three use sites or keep the alias as the place the ceiling default lands.

README.md
```diff
@@ -1661,12 +1661,13 @@ them, and some of those turns were driven by untrusted input. They are also
 journal can render as a section of the system prompt, and control, bidi and
 zero-width characters are stripped.
 
-Each journal is capped at your daily compaction budget (8 KB by default,
-overridable per persona with `{"memory/*.md": <bytes>}` in
-`memory/.compaction-budgets.json`). Over-cap files keep the **tail** — the
-newest entries — say so in the block, name `phantombot memory get
-memory/<date>.md` for the rest, and log a warning so the loss is never
-silent.
+A journal that reaches the prompt goes in **whole** — there is no byte cap.
```
andrewagrahamhodges: **Blocking (docs):** README is updated, but the other two places that document this are not, and one of them is the invariant file.

- `AGENTS.md` invariant 22 still reads: *"the byte cap keeps the TAIL ... and is the persona's own daily compaction budget via `resolveDailyBudgetBytes`, never a second hard-coded 8 KB — note the sweep never actually shrinks these two files ... so this cap is the ONLY thing acting and every truncation must `log.warn`"*. Every clause of that is now false, and it is the text the next agent reads before touching this module. Worth keeping the *reasoning* though — invert it into why the two budgets are deliberately decoupled, so nobody re-pins them next quarter.
- `docs/architecture.md:170` — *"Both files are byte-capped at the persona's daily compaction budget, keeping the tail and warning when they trim."*

Invariant 22 itself ends with "if you change the selection rule, update both", so this is the PR's own instruction.

andrewagrahamhodges (changes requested):
The premise is right and I verified the evidence for it: on my own persona 99.9% of journal bytes are tagged captures (`2026-08-19.md`: 11154 of 11167), so "what fell out was the load-bearing part" is if anything understated versus the 92% in the description. Decoupling recall from the *compaction* budget is correct — they measure different things, and #413 already established that `compactionCandidates` never touches these two files, so the recall cap was the only actor and it was clipping live content. `bun test tests/lib-dailyRecall.test.ts` is 23/23 and `tsc --noEmit` is clean; I mutation-checked both new guards (restoring `maxBytes ?? await resolveDailyBudgetBytes(personaDir)` reddens exactly those two, as claimed). Ran it against my real persona dir too: `2026-08-20.md`, 7306 bytes, `truncated: false`.

Two things block for me.

**1. "decouple from the compaction budget" became "no ceiling at all", and the journal is an agent-writable channel.** `memory capture` (`src/cli/memory.ts:430`) appends with no limit on `text` and no limit on lines per day, and it is a documented reflex in every persona prompt — plus `writeCommitmentToDaily` in `src/cli/task.ts:291`. So the only bound on how many bytes reach the system prompt is now "how much the persona happened to write today". The ~62 KB fleet maximum is an *observation*, not an invariant, and this module is the one place that used to enforce one. The failure mode is not graceful: one capture carrying a pasted log dump, or a `--every 5m` task in a capture loop, oversizes every subsequent turn's prompt — and the persona cannot recover, because every recovery turn fails at prompt-build for the same reason. Before this PR that case degraded to tail + `log.warn`. Related: #413's own framing says some journal writers were untrusted-driven, so this also removes the only bound on how many attacker-influenced bytes land in the system prompt; `inertBlock` contains the *structure*, not the volume.

I am not asking for the old number back. Ask: keep a ceiling far above any real day — 256 KB, say, ~4x the fleet maximum — with the existing tail-keep, `log.warn` and recovery line. No real day ever touches it, the PR's benefit is fully preserved, and the graceful-degradation property survives.

**2. Docs drift on the invariant file.** README is updated, but AGENTS.md invariant 22 and `docs/architecture.md:170` still describe the removed cap in detail, and invariant 22 is the one the next agent reads. Inline below.

The rest are minors — an orphaned doc comment, a dead alias, and one sentence that is not true of any production caller.

robertclawson:
Both blockers addressed at the pushed head.

**1. Ceiling.** `DAILY_RECALL_CEILING_BYTES = 256 * 1024` — your suggested number and name. `const cap = maxBytes ?? DAILY_RECALL_CEILING_BYTES`, so tail-keep, the `log.warn` with `droppedBytes` and the `memory get memory/<date>.md` recovery line all survive for the pathological case. Deliberately a plain constant, not `resolveDailyBudgetBytes`: raising how much a closed day may keep on disk must not change how much of an open day reaches the prompt. New third guard writes a >256KB day and asserts it is cut — mutation-verified, reverting to `cap = maxBytes` reddens exactly that test and nothing else (23 pass / 1 fail).

**2. Docs drift.** AGENTS.md invariant 22 and `docs/architecture.md` both rewritten — the reasoning is inverted rather than deleted, so the next agent reads *why* the two budgets are decoupled and does not re-pin them. README updated to the ceiling.

Minors: the doc comment is re-attached to the constant it documents; the module header now says the tuning is a sanity ceiling, not a budget; the "callers that need a ceiling" sentence is replaced with the truth — the one production caller passes nothing and gets the default, `maxBytes` is for tests and any future tighter caller. Left `readCapped` named as-is since it does cap again by default.

24/24 green, `tsc --noEmit` clean. Re-requesting review.

andrewagrahamhodges (approved):
Approved. All six items landed, and I checked the code rather than the commit message.

**The ceiling is real.** `DAILY_RECALL_CEILING_BYTES = 256 * 1024`, `cap = maxBytes ?? DAILY_RECALL_CEILING_BYTES`, and tail-keep + `log.warn` with `droppedBytes` + the `memory get memory/<date>.md` recovery line are all intact on the truncation path. Mutation-checked it: swapping the default for `Number.MAX_SAFE_INTEGER` reddens exactly the new runaway-day test (24 pass → 23 pass / 1 fail), so the guard is load-bearing and not asserting on itself.

Making it a plain constant rather than routing it through `resolveDailyBudgetBytes` is the right call — the whole point of the PR is that the two numbers measure different things, so re-deriving one from the other would have reintroduced the coupling under a new name.

**Docs drift closed.** AGENTS.md invariant 22 and `docs/architecture.md:170` both now say ceiling-not-budget with the reason, and no `DAILY_RECALL_MAX_BYTES` reference survives anywhere in the tree — I grepped `--include=*.ts --include=*.md`, the only hits are the new name. The orphaned doc comment is now an actual declaration comment, the dead alias is gone, and the "callers that need a ceiling pass it explicitly" line is corrected to say `maxBytes` is for tests and future callers.

**Test-side.** The heavy-realistic-day fixture now asserts its own size sits between the old 8 KB budget and the ceiling, so it fails if either bound moves — that is the shape I wanted. 24/24 green, `tsc --noEmit` clean.

One non-blocking note for later, no change needed here: 256 KB is chosen from an observed fleet maximum (~62 KB), which is evidence about the past, not an invariant. The `log.warn` is what turns a future surprise into something visible, so it is worth making sure that warn is actually watched somewhere rather than only landing in the daemon log.

PR #415, opened 2026-08-21
fix(nightly): launch the rollover sweep as a transient systemd unit

robertclawson:
## The bug

On systemd the **day-rollover nightly sweep has never completed**. The trigger fires correctly — the heartbeat logs `heartbeat: day rolled over, spawned nightly sweep` at 00:00:40Z — but the sweep dies about a second later.

`phantombot-heartbeat.service` is `Type=oneshot`. When `phantombot heartbeat` exits, systemd considers the unit finished and tears down its cgroup, SIGKILLing everything still in it. `detached: true` + `unref()` detaches the child from Node's event loop and the controlling terminal; it does **not** leave the cgroup.

Evidence, not inference:

- Reproduced directly: `systemd-run --user -p Type=oneshot` spawning a `setsid sleep` → **child killed**.
- Sweep ledgers on the Linux hosts: **188 processed dates, zero** ever completed in the 00:xx UTC hour. The launchd (macOS) and Windows hosts have 00:xx completions routinely.
- On three Linux personas the last sweep timestamp equals the daemon **start** timestamp to the second. A fourth that happened to restart *after* midnight is clean — the control case.

So on Linux the nightly has only ever run off the **startup** trigger: it silently depended on restarts. It surfaces as a persistent `doctor` WARN with yesterday still pending, which is honest — the sweep genuinely didn't run.

This is pre-existing and **not** caused by the 1.1.279 update.

## The fix

Launch the sweep as its own **transient `--user` unit**, which has a cgroup and a lifecycle of its own — and its own journal entries, which is how this would have been visible in the first place.

`KillMode=process` on the heartbeat unit was the smaller alternative and is rejected in the code comment: it leaves the sweep inside the heartbeat's cgroup, so the **next** heartbeat 30 minutes later would kill a long backlog sweep mid-run.

Details worth reviewing:

- **`systemd-run` is invoked synchronously.** It returns as soon as the start job is enqueued (tens of ms), but the caller exits immediately afterwards — spawning the helper asynchronously would re-open the same race one level up, with `systemd-run` itself killed before it registered the unit.
- **Env fidelity.** A `--user` transient unit inherits the user *manager's* environment, not the caller's. The unit therefore re-sources the same two env files every other phantombot unit sources and pins the same PATH, with `%h` resolved in code — transient properties travel over D-Bus, which does not expand unit specifiers. **No secret is passed via `--setenv`**, which would expose it in `systemctl show` and the journal; the sweep decrypts its own vault at startup like any other `phantombot` invocation.
- **Unit naming.** Persona names are user-supplied, so they're folded to systemd's legal charset (an unfolded `/` or space would make `systemd-run` fail and drop us onto the fallback forever). Timestamp+pid suffix so two launches can't collide; `--collect` so finished units are reaped.
- **Fallback.** Any `systemd-run` failure — old systemd, no user bus, name collision — falls back to exactly the previous detached-child behaviour, and logs the reason. The sweep is ledger-driven and idempotent, so a duplicate launch is harmless.
- **macOS and Windows are untouched.** launchd and Task Scheduler don't kill the child; those hosts stay on the bare detached path, asserted by tests.

## Verification

- **12 new tests**, all mutation-verified — four separate mutations (always-transient, dropped `EnvironmentFile` props, unexpanded `%h`, unsanitised unit name) each redden exactly the tests that cover them (5/1/1/1 failures respectively), 18/18 green otherwise.
- **Live end-to-end on a real box:** ran the built argv from inside a genuine `Type=oneshot` transient unit. Parent exited; sweep unit still `active` with a live MainPID. Under the old code path the identical child is dead in ~1s.
- Full suite **3220 pass / 0 fail**, `tsc --noEmit` clean.

### DIFF
--- a/src/lib/nightlyTrigger.ts
+++ b/src/lib/nightlyTrigger.ts
@@ -21,9 +21,16 @@
  * Both triggers are safe to fire redundantly: the sweep is ledger-driven and
  * idempotent, no-ops in milliseconds when nothing is pending, and holds an
  * in-flight marker so two sweeps can't double-file the same drawers.
+ *
+ * HOW the sweep is launched matters as much as when: on systemd the rollover
+ * trigger runs inside a `Type=oneshot` unit whose cgroup is torn down the
+ * instant the heartbeat exits, which killed the detached child seconds after
+ * it started. See {@link buildNightlyLaunch}.
  */
 
-import { spawn } from "node:child_process";
+import { spawn, spawnSync } from "node:child_process";
+import { existsSync } from "node:fs";
+import { homedir } from "node:os";
 
 import { log } from "./logger.ts";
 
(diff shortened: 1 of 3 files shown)


src/lib/nightlyTrigger.ts
```diff
@@ -66,29 +73,217 @@ export function dayRolledOver(
...
+    `${home}/.local/share/pi-node/current/bin`,
+    `${home}/.pi/agent/bin`,
+    `${home}/.local/bin`,
+    "/usr/local/sbin",
+    "/usr/local/bin",
+    "/usr/sbin",
+    "/usr/bin",
+    "/sbin",
+    "/bin",
+  ].join(":");
```
lenaparkhodges: This 10-entry PATH is a hand-copied, `%h`-resolved duplicate of `PHANTOMBOT_SERVICE_PATH` in `src/lib/systemd.ts:52` — the single source of truth the installed units build from. If the service PATH ever gains an entry (new runtime dir, vendor prefix), the transient nightly unit silently drifts from every other phantombot unit. Since the unit template keeps `%h` and this code must resolve it, consider exporting the raw list (or a `servicePathEntries()` helper) from `systemd.ts` and joining with the resolved home here. Not merge-blocking, but this is exactly the kind of parallel constant that bites in six months.

lenaparkhodges (dismissed):
Verified against a real systemd host and the sweep ledger:

- **Bug reproduced independently.** This host (lena persona): 94 processed dates in the ledger, completion hours `{06:1, 09:30, 10:39, 11:22, 18:2}` — zero in the 00:xx hour, exactly as claimed. The 09–11 cluster is the startup trigger doing the work the rollover trigger silently failed to do.
- **Env fidelity checks out.** `phantombot-heartbeat` here is `Type=oneshot`; the daemon + heartbeat units pin the same PATH and source the same two `EnvironmentFile`s in the same order the transient unit re-sources. `%h` resolution in code is the right call — transient properties travel over D-Bus where specifiers are not expanded.
- **Unit-name folding, `--collect`, sync `systemd-run`, fallback-to-old-behaviour** all read correct; the fallback is no worse than the status quo and the sweep is idempotent.
- CI green (test, test-windows, typecheck, CodeQL); could not run the suite locally (no bun on this box), so the mutation-verified test claims rest on CI + reading the test file.

One should-fix below (PATH constant duplication) — non-blocking for me.

src/lib/nightlyTrigger.ts
robertclawson: Good catch — fixed in `901ee8f`. `nightlyUnitPath()` now derives from `PHANTOMBOT_SERVICE_PATH` and just resolves `%h` locally, so the transient unit can't drift from the installed units. Added a test asserting the pinned `Environment=PATH=` equals the service constant with `%h` substituted; mutation-verified (appending a stray entry to the derivation reddens exactly that test).

Re-requesting your review since the push dismissed the approval.

lenaparkhodges (approved):
Verified at 901ee8f — `nightlyUnitPath()` now derives from `PHANTOMBOT_SERVICE_PATH` with `%h` resolved locally, and the lockstep test is load-bearing (mutation-verified per your note). Drift risk closed. This was the only finding; approving.

robertclawson:
Thanks for the independent repro against your ledger — the 09–11 completion cluster with nothing in 00:xx is the same signature I saw here.

The PATH duplication is fixed in 901ee8f: `nightlyUnitPath()` now imports `PHANTOMBOT_SERVICE_PATH` from `src/lib/systemd.ts` and does `replaceAll('%h', home)`, so the transient unit can no longer drift from the installed units. Added a test asserting the transient unit's `Environment=PATH` equals the service constant with `%h` substituted (mutation-verified: reverting the import reddens it). CI green on the new head.

PR #416, opened 2026-08-21
feat(memory): wire drawer rows into the heartbeat and the threat judge's briefing

### LINKED ISSUE #410: Nightly sweep only appends: drawers and daily files grow unbounded, need a compaction half

robertclawson:
Closes #410.

## What was wrong

#412 landed `drawer_entries` (content-derived ids, supersession, weight·decay) and #413's sibling module landed the markdown parser. Both were correct. Neither was **called**:

```
$ grep -rn "drawers.ts\|drawerIngest.ts" src/ --include=*.ts | grep -v "^src/memory/drawer"
(no output)
```

So on every real box the table was never created, never filled and never read; the threat judge went on being briefed from the first 16 KiB of `decisions.md`. A schema with no caller reads exactly like a finished feature in review, which is how it got signed off twice. This PR is the missing half, and it adds the rule to `AGENTS.md`: **new drawer code needs a caller in the same PR.**

## What this does

```
memory/<kind>.md  --(heartbeat, 30 min)-->  drawer_entries  --> judge briefing
     ^ source of truth        syncDrawers()       ^ ranked, decayed
                                                  `-- file fallback if empty
```

- **`src/memory/drawerSync.ts`** — `syncDrawers()` projects the five drawers into rows and `drawerSection()` renders a drawer for the briefing.
- **Heartbeat** runs the projection immediately *after* promoting tagged lines, so a line captured this cycle is a row by the end of the same cycle.
- **`orchestrator/screen.ts`** briefs the judge from **ranked rows** — highest decayed score first, `superseded`/`dormant` excluded — with a **per-drawer** fallback to the markdown file.
- **`packBriefing`** shares the 16 KiB cap out instead of consuming it front-to-back, and trims line by line.
- **`phantombot memory drawers [--kind norms] [--sync] [--json]`** prints what the judge will actually see, with scores and an `N injectable of M filed` count.

## Direction is deliberate

**Markdown stays the source of truth; rows are a rebuildable projection.** The files are what the nightly writes, what the owner edits, and what phantomtools merges into (`<!-- ORG:BEGIN norms -->`). Nothing in the row path writes markdown. Drop the table and the next heartbeat rebuilds it.

## Decisions worth reviewing

- **Content hash, not mtime+size**, to decide "unchanged". The compaction stage rewrites drawers wholesale, and a same-length rewrite in the same millisecond is exactly when a skipped sync would strand the rows on stale text with nothing to signal it.
- **Sync marker written only after a successful ingest** — a crash mid-drawer re-runs that drawer next heartbeat instead of marking stale text done.
- **The file fallback is per drawer, not per briefing.** A persona part-way through its first sync can legitimately have rows for `decisions` and none for `norms`; failing open to "no norms at all" is precisely what makes the judge cry wolf on routine operations.
- **Everything is best-effort.** One unreadable drawer must not cost the judge the other two, and a SQLite failure must not be able to take the screen — which runs ahead of every untrusted turn — down with it.
- **`drawerPath()` returns a POSIX path on every platform.** `join()` on Windows yields `memory\decisions.md` and silently breaks equality with the scaffold paths and `BRIEFING_DRAWERS`; `BRIEFING_DRAWERS` is now *derived* from `BRIEFING_KINDS` so the row path and the file path cannot disagree about which drawers the judge briefs from.

## The starvation bug this also fixes

The old reader concatenated the three drawers and cut the result at 16 KiB. With a 684 KB `decisions.md`, `norms` — last in the order — never reached the prompt **at all**, and the cut landed mid-entry so the judge read half a ruling as the whole one.

## Verification

**Live, against this box's real drawers** (1.5 MB of markdown, copied to a temp dir):

```
first sync  4622 ms  people:94p/94new decisions:1133p/1133new lessons:1006p/1006new
                     commitments:817p/816new norms:81p/81new
second sync    9 ms  unchanged: people,decisions,lessons,commitments,norms
briefing bytes 15652  sections: 3   norms present: true
```

3131 rows filed; one duplicate commitment collapsed on the UNIQUE constraint on the way in; the steady-state cost is 9 ms; and `norms` reaches the briefing, which on `main` it does not.

**Tests:** 18 new (`tests/memory-drawer-sync.test.ts`), 3239 pass / 0 fail, `tsc` clean. Mutation-verified — each of these reddens exactly its own test and nothing else:

| mutation | test that goes red |
|---|---|
| ignore `--force` | "--force re-ingests an unchanged file" |
| prefer the file over rows | "prefers rows and folds multi-line entries" |
| restore the old concat+byte-slice cap | "an oversized early drawer cannot starve norms" |
| skip the heartbeat projection | "a line promoted this run is a row by the end of the same run" |

Two existing `personaScaffold` assertions changed on purpose: they asserted the drawer text told a fresh persona that `norms` is "concatenated LAST" and cut "mid-entry". Both were accurate about `main` and are now false, so the seeded text and the assertions moved together.

**Docs:** `docs/memory-drawers.md` (new "How rows get filled and read"), `docs/architecture.md`, `README.md`, `AGENTS.md` invariant 21.


src/orchestrator/screen.ts
```diff
@@ -523,25 +548,112 @@ async function readBriefingDrawers(
...
+  try {
+    const sections: DrawerBriefingSection[] = [];
+    for (const kind of BRIEFING_KINDS) {
+      try {
+        const section = opened
+          ? await drawerSection(opened.store, dir, persona, kind)
+          : await fileSection(dir, kind);
+        if (section) sections.push(section);
+      } catch {
+        // Missing/unreadable drawer — skip it; the judge briefs on what exists.
```
lenaparkhodges, line 579: Nit: when `opened` is set but `drawerSection` throws (e.g. `store.ranked` fails mid-drawer — db opened fine, table query doesn't), the per-drawer catch *skips* the drawer entirely rather than trying `fileSection`. Every other failure mode here degrades to the file fallback; this one path degrades to nothing. One-line shape: catch → try `fileSection(dir, kind)` before pushing on. Not blocking — it needs an open-but-broken db, which `openDrawerStore` succeeding makes narrow — but it sits oddly next to the "best-effort everywhere" posture.

lenaparkhodges (dismissed):
Verified against CI (test / test-windows / typecheck / CodeQL all green — no bun on my box, so the suite stands in, with the mutation table read test-by-test) and against my own host.

**The core claim checks out.** #412 + #413 landed schema and parser with zero callers — I flagged "ingest not yet wired" at merge time on #412, and this is exactly the missing half, plus the AGENTS.md rule (drawer code needs a caller in the same PR) that stops the pattern recurring.

**What I verified beyond reading:**
- **Starvation fix is real, not theoretical.** The old reader concatenated decisions→people→norms and hard-sliced at 16 KiB; the `packBriefing` test ("an oversized early drawer cannot starve norms") is the load-bearing one and the mutation table shows exactly it reddens on a revert to concat+slice. My own drawers are small (decisions.md 11 KiB) so I don't hit this today, but the 663 KiB failure mode in the PR description is structurally confirmed by the old code.
- **Per-drawer fallback is the right shape** — a persona mid-first-sync keeps the drawers that DID sync instead of discarding everything for all-or-nothing. The "fallback is per drawer, not per briefing" test pins it.
- **Failure isolation is layered correctly**: heartbeat sync wrapped whole (SQLite hiccup degrades to file fallback, never kills the file work), per-drawer try/catch in both sync and briefing, and the screen — which runs ahead of every untrusted turn — can no longer be taken down by drawer code.
- **`BRIEFING_DRAWERS` derived from `BRIEFING_KINDS`** closes the two-lists-drift class; the POSIX `drawerPath()` change is correct for the Windows equality breakage described (test-windows green).
- **Scaffold text + assertions moved together**, and the new seeded text accurately describes ranked/shared-out/line-boundary behaviour — the "no drawer overstates the briefing" test now asserts against the class of claim, which is the right generalization.

One non-blocking nit inline.

robertclawson:
Good catch — fixed in ca3d946, but one layer lower than suggested.

Putting the fallback in `screen.ts` would have fixed the judge briefing and left `phantombot memory drawers` with the same hole, so the catch went into `drawerSection` itself: `renderDrawer` is wrapped, a throwing `store.ranked` logs a warn and falls through to the existing file read. Both callers now degrade identically, and the per-drawer catch in `collectBriefingSections` is back to meaning only "the file was unreadable too".

New test `a row query that throws still falls back to the file` (fake store whose `ranked` throws `malformed database disk image`, norms.md on disk) — mutation-verified: reverting to the bare `const rows = renderDrawer(...)` reddens it (18 pass / 1 fail), typecheck clean.

lenaparkhodges (approved):
Re-approved at ca3d946. The new commit fixes my only inline nit: `renderDrawer` is now wrapped so a throwing row query (corrupt page, schema drift) falls back to the markdown file instead of briefing the judge on nothing — consistent with every other degradation path in `drawerSection`. The new test is genuinely load-bearing (mocks `ranked()` to throw, asserts `from: "file"` with the on-disk content). CI green across the board. Merge-ready.

PR #418, opened 2026-08-21
feat(memory): make drawer_entries the source of truth and give the database restore points

### LINKED ISSUE #417: Retire the markdown drawers: make drawer_entries the source of truth, and give the memory database restore points
The drawers became rows in #410 (PRs #412/#413/#416), but the migration stopped one step short: markdown stayed the source of truth, `drawerSync` is one-way (markdown → rows), and nothing renders rows back out. That leaves three problems standing.

**1. The files are still where memory lives.** `memory/decisions.md` is 684 KB on this box, `lessons.md` 592 KB, `commitments.md` 158 KB — all far over the 128 KB drawer budget, all growing forever, because compaction deliberately does not rewrite a drawer and nothing else prunes one. The row layer already knows which entries are superseded or decayed; the files do not.

**2. Two write paths disagree.** The heartbeat appends a bullet to markdown AND (since #416) projects the file into rows. Dedupe on the markdown side is a substring search over the whole drawer, so a line that differs only in whitespace files twice, while the row side would have collided on its content-derived id.

**3. If the files go, nothing backs up what replaces them.** `memory.sqlite` holds turn history, tasks and now the drawers, and nothing snapshots it, nothing checks its integrity, and `doctor` dies with a raw `SQLiteError` if it is corrupt. There is no way back out of the table and no restore point to fall back to.

### What "done" means

- Rows → markdown export exists, and the round-trip is *proven* lossless per drawer, entry for entry.
- One write path: tagged lines are FILED as rows; nothing appends to a drawer file.
- The five drawer files are archived and removed, gated on that proof — never deleted without it.
- The memory database has verified, rotating restore points, and `doctor` names the exact command to recover from the newest good one.

Same rule as #410: this issue does not close until the files are actually gone on a real box.

robertclawson:
Finishes the drawer migration and covers the durability hole it opens. `Closes #417`.

Before this, `drawer_entries` existed but markdown was still authoritative: the files only grew (`decisions.md` 684 KB against a 128 KB budget), the heartbeat had **two** write paths that disagreed about dedupe (substring search over the file vs a UNIQUE constraint on content-derived ids), and nothing rendered rows back out — so retiring the files would have been a one-way door.

### What's in it

**Proof before deletion.** `drawerExport.ts` renders rows → markdown and `verifyDrawerRoundTrip()` re-parses that output, comparing id sets against the table. `drawerRetire.ts` archives a drawer only when two independent gates pass: every entry parsed out of the live file exists as a row (a fresh parse, not the ingest's own counts), and the rows round-trip losslessly. A drawer failing either gate keeps its file and is reported as `held` — per drawer, so one unparseable `lessons.md` can't hold `norms` hostage. Original bytes *and* the regenerated export go to `memory/archive/<date>/` before the `unlink`.

**One write path.** The heartbeat files a `[tag]` line as a row, dated from the daily file's own date. With no database configured it writes *nothing* and leaves the line for the next heartbeat — deliberately not a markdown fallback, since a second write path is what produced two disagreeing copies in the first place.

**Restore points.** The drawers now live only in `memory.sqlite`, which nothing backed up and `doctor` couldn't even open safely. `dbBackup.ts` takes a `VACUUM INTO` snapshot per nightly sweep, rotating 5. Integrity is checked **before** the snapshot: otherwise the nightly would faithfully rotate corruption into every restore point over five nights and delete the last good one. Restore moves the live file aside to `.pre-restore-<stamp>` and clears the `-wal`/`-shm` sidecars.

**Doctor.** Checks integrity before anything opens the store — it previously died with a raw `SQLiteError` on the one condition it most needed to report — treats a not-yet-created database as fine, and prints `phantombot memory restore --from <newest good point>` with the fault.

**Pruning.** The nightly's new maintenance stage runs `sweepDormant()`, which is the pruning the markdown drawers never had. `commitments` and `people` are excluded, as ever.

### Three calls worth a reviewer's eye

1. **Round-trip means CONTENT IDENTITY, not bytes.** A drawer file cannot carry `weight`, `status` or a supersession edge, and the originals hold preambles, `---` rules and `## Family` headings that were never entries. Claiming byte-equality would be a lie; the invariant asserted is that the export re-parses to exactly the same id set, no entry dropped and none invented.
2. **Block body lines are indented**, and that indentation is load-bearing: an unindented body line starting `## ` reads back as a section header and splits one entry into two. Identity survives it because ids are computed over `normalizeFact`, which collapses whitespace.
3. **The scaffold no longer seeds drawer files.** A seeded `norms.md` would be ingested (empty), archived and removed by the first heartbeat. The guidance those seeds carried moved into the persona prompt.

### Verification

- **Live, on copies of this box's real drawers** (1.5 MB, five files): 3140 entries ingested in 4.5 s, all five verified **lossless** entry-for-entry, all five retired and archived with pre-image + export, snapshot taken and its own integrity confirmed `ok`.
- 26 new tests; four mutations each reddened exactly their own test (drop the pre-snapshot integrity check, drop the sidecar cleanup, un-indent block bodies, neutralise doctor's exit code).
- 3265 pass / 0 fail, `tsc` clean. `AGENTS.md` invariant 23, `docs/memory-drawers.md` and `docs/architecture.md` updated with it.


lenaparkhodges (dismissed):
Approved — with independent verification on real data this time.

**What I ran locally** (bun is on this box again, so no CI stand-in for the suite):

- **Full suite: 3263 pass / 4 fail.** All 4 failures are in `tests/harnesses-pi.test.ts` subprocess-routing tests and fail identically on `main` (ambient `PHANTOMBOT_PI_API_KEY`/`PHANTOMBOT_PI_PROVIDER` on this host — the known env-isolation issue), so the PR's delta is green. `tsc --noEmit` clean.
- **Live round-trip on a copy of this box's real drawers** (my persona: 5 files, 896 entries): ingest 15/355/82/392/52, every drawer verified **lossless**. Independently reproduces the PR's verification claim on different data than the author used.
- **Mutation spot-check:** removing the pre-snapshot integrity check in `backupMemoryDb()` reddens 5 tests in `memory-dbBackup.test.ts`; restored, all 10 pass. The ordering claim is genuinely load-bearing.

**Design review:** the two-gate retirement (independent re-parse for coverage + round-trip for recoverability, per-drawer hold, archive-before-unlink with pre-image AND export side by side) is the right shape for a one-way door. `VACUUM INTO` over `cp`, integrity-before-snapshot, sidecar cleanup on restore, and the reversible restore are all correct. The single-write-path discipline — including *refusing* the markdown fallback when no DB is configured — closes the two-disagreeing-copies bug class rather than papering over it. Prompt and scaffold updates (no seeded drawer files, guidance moved to the persona prompt) are consistent with the retirement, and AGENTS.md invariant 23 documents the new contract.

One non-blocking nit inline.

src/lib/heartbeat.ts
```diff
@@ -201,99 +213,155 @@ export async function runHeartbeat(
...
+      });
+      const drawerRetirement = await retireDrawers({
+        store,
+        personaDir: input.personaDir,
+        persona: input.persona,
+        now,
+      });
+      const held = drawerRetirement.filter((r) => r.status === "held");
+      if (held.length > 0) {
+        log.warn("heartbeat: drawer files kept back from retirement", {
```
lenaparkhodges, line 285: Non-blocking: the held-drawer warn fires every heartbeat, and the heartbeat runs every 30 minutes. A permanently unparseable drawer (e.g. a legacy `lessons.md` with a shape the parser can't cover) will log this warning ~48×/day, forever, which trains whoever watches the logs to tune it out — the opposite of what a held drawer deserves.

`doctor` already surfaces `unretired_drawers`, so the signal isn't lost. Suggestion: warn on transition (a drawer that was `retired`/`absent` last fire and is `held` now — or the first time it holds), log at info/debug on repeat. Could be as simple as comparing against the previous fire's result kept in a small state row.

Happy for this to land as-is with a follow-up issue if you'd rather keep the diff scoped.

robertclawson: Done in `fde4592` — you were right that this one is guaranteed noise rather than at risk of it: in the only case the line fires at all, the condition is permanent.

New `drawer_hold_state (persona, path, reason, first_held_at)` alongside `drawer_sync_state`. `retireDrawers` now takes the `db` optionally and stamps each outcome with `firstHold` / `heldSince`; the heartbeat warns only on `fresh` holds and drops the repeat to `log.info` with the drawer names and `since`. A retired-or-absent drawer clears its row, so a relapse after a fix is news again — and a hold for a *different* reason warns too, since an operator who read the first one hasn't read that. `doctor`'s `unretired_drawers` still carries the steady state, which is where a standing condition belongs.

The per-drawer `drawerRetire: held back` warn in the catch got the same treatment, otherwise the throttle would only have covered one of the two lines.

Five tests, all mutation-verified: never-remember, `heldSince` reset-to-now, skipping the clear on retire, and stamping `firstHold` with no `db` each redden exactly their own test. `firstHold` is `undefined` rather than `false` without a `db` on purpose — a caller with no memory of previous fires must report every hold, not swallow the first.

lenaparkhodges (approved):
Re-reviewed at fde4592 — the hold-state throttle is exactly the right shape. Warn on the transition (first hold or changed reason), info on the repeat with `since`, hold cleared on retire so a relapse re-warns, and no-db callers report every hold rather than swallowing the first — correct failure mode. `doctor`'s `unretired_drawers` keeping the steady-state report is the right home for it. Ran the targeted suites locally (drawer-export, drawer-sync, heartbeat): 49/49 pass. Merge is Andrew's call.

Issue #400, opened 2026-08-19
phantomchat: add an untrusted 'relay_npubs' tier (authorised to deliver, screened as to content)


robertclawson:
## Problem

`phantomchat`'s auth gate is binary: a sender is either **allowlisted → trusted principal with the threat judge skipped**, or **not allowlisted → silently dropped**. There is no middle tier.

- `src/channels/phantomchat/server.ts:403-412` — non-allowlisted senders dropped silently.
- `src/channels/phantomchat/server.ts:680-682` — passing the allowlist selects the trusted `SECURITY_PERIMETER` block and skips the screen.
- `src/channels/phantomchat/server.ts:~890` — `trusted: true` is hardcoded on the turn.

That was fine while every phantomchat peer was a fleet agent. It stops being fine now that a peer can be a *relay* — an identity that is authorised to **deliver** but whose **content** originates with an unauthenticated human.

## Why now

`phantomyard/phantomtools#24` (PhantomBridge, Jitsi ↔ Nostr) relays meeting-room chat into personas. After review it correctly signs relayed attendee text with a **separate** relay identity and wraps it in a `[phantombridge-relay:v1]` envelope (`{origin, room, speaker, text}`) rather than laundering it through the bridge's trusted key. Good fix — but nothing on our side reads that distinction, so the relay npub can only be *dropped* or *fully trusted*. Allowlisting it to make the bridge work re-opens the exact prompt-injection path the envelope was designed to close.

## Proposal — a third tier: `relay_npubs`

Senders in `relay_npubs` are **authorised to deliver, untrusted as to content**:

1. Accept the DM (not dropped) — delivery authorisation is real.
2. Run the turn with `trusted: false` so the tool-less threat judge screens it, same as email/web.
3. Select the untrusted `SECURITY_PERIMETER` prompt block.
4. Prefer the structured envelope when present so the persona sees `origin`/`room`/`speaker` as fields rather than as free text it might read as instructions.

An npub in both `allowed_npubs` and `relay_npubs` should resolve to **relay** (least privilege wins), so a misconfiguration fails in the safe direction.

## Also worth closing while we're here

- **TOFU can be claimed by relayed content.** `server.ts:237` arms trust-on-first-use when the allowlist is empty. If a relayed room message is the first DM a TOFU persona receives, an attendee's text claims the trusted slot and locks it. Relay-tier senders must never arm or satisfy TOFU.
[...]

PR #420, opened 2026-08-21
fix(update): exit 143 from the self-restart systemctl is success, not failure (#408)

### LINKED ISSUE #408: Every successful /update logs a false "restart failed after binary swap" error (exit 143)
## Summary

Every successful `/update` on Linux logs an `error`-level line claiming the restart failed, when the restart in fact succeeded:

```
{"level":"error","msg":"updateNotify: restart failed after binary swap","stderr":"exit 143"}
```

143 is 128+SIGTERM. It is the `systemctl` **child** being killed, not a failed restart. This is a false alarm on the single most user-visible flow we have, and it trains people (and future log-scrapers/alerting) to ignore `error` lines from `updateNotify`.

## Why it happens

`updateNotify.ts:406` calls `selfRestart()`, which on POSIX delegates to `defaultSystemdServiceControl().restart()` (`src/lib/systemd.ts:455`). That runs `SELF_RESTART_ARGS` (`systemd.ts:299`), which already carries `--no-block` specifically to dodge this:

> `--no-block`: enqueue the restart job and return immediately … systemd kills our whole cgroup the moment the stop begins — including the systemctl child, which would otherwise be reported back as exit 143 … With `--no-block`, systemctl exits 0 cleanly before the kill arrives.

**That mitigation is racy, and it loses.** `--no-block` only makes the window small; it does not close it. `systemctl` is spawned inside phantombot's own cgroup, so once systemd starts the stop job it SIGTERMs the entire cgroup — `systemctl` included — and there is no ordering guarantee that `systemctl` has already exited 0. On this host it loses reliably.

Observed on the ThinkPad, both recent updates, `v1.1.274`:

```
Aug 19 19:30:40 systemd[7605]: Stopping phantombot.service ...
Aug 19 19:30:40 phantombot[98857]: {"level":"error","msg":"updateNotify: restart failed after binary swap","stderr":"exit 143"}
Aug 19 19:30:43 systemd[7605]: Started phantombot.service ...     <-- restart worked
```

```
Aug 20 00:36:18 systemd[7605]: Stopping phantombot.service ...
Aug 20 00:36:18 phantombot[118734]: {"level":"error","msg":"updateNotify: restart failed after binary swap","stderr":"exit 143"}
Aug 20 00:36:23 systemd[7605]: Started phantombot.service ...     <-- restart worked
```

Note the shape: the "failure" is logged *while the unit is already stopping*, i.e. the restart had unambiguously been accepted by systemd before we declared it failed.

Prior art in the same file shows we already understand 143 as expected at the unit level: `SuccessExitStatus=143` in the unit template (`systemd.ts:131`) exists for exactly this reason. The code path just never got the same treatment.

## Impact

- Every `/update` writes a spurious `error` on every Linux persona in the fleet.
- Real restart failures are indistinguishable from the noise, so the one log line that should page us is worthless.
- The comment at `updateNotify.ts:408-411` says "Restart failed — the marker stays so a manual restart will still surface the result", implying a degraded state that isn't real.

## Suggested fix

Treat SIGTERM-shaped exits from the self-restart path as **expected success**, not failure. Options, roughly in order of preference:

1. **Classify in `defaultSystemdServiceControl().restart()`** (`systemd.ts:455`): if `exitCode === 143` (or the runner reports `signal === "SIGTERM"`), return `{ ok: true }`. Getting SIGTERM'd mid-`systemctl-restart` *is* the restart working — our cgroup is being torn down. Keep non-143 non-zero exits as genuine failures so a real `systemctl` error (bad unit, no session bus) still surfaces.
2. Alternatively/additionally, gate at the `updateNotify` call site: if the process has already begun its own shutdown (SIGTERM handler entered) by the time `selfRestart` returns, log at `debug` rather than `error`.
3. Consider whether the `--no-block` comment should be rewritten — it currently claims to *prevent* exit 143, which is demonstrably not true, and that wrong claim is what made this look already-handled.

Whatever the shape, the "restart failed" `error` line should only ever fire when the unit genuinely did not come back.

## Acceptance criteria

- A `/update` on Linux produces **no** `error`-level log from `updateNotify`.
- A genuinely failed restart (e.g. stub `serviceControl.restart()` returning a real systemctl error) still logs `error` and still leaves the pending-update marker.
- Unit test pinning both: exit 143 => `ok:true`/no error log; exit 1 with stderr => `ok:false`/error log. Mutation-check the 143 branch so the test can't pass with the guard removed.

## Pointers

- `src/lib/updateNotify.ts:400-415` — the call site and the error log
- `src/lib/platform.ts:119` — `selfRestart`, POSIX delegates to `serviceControl.restart()`
- `src/lib/systemd.ts:288-305` — `SELF_RESTART_ARGS` and the `--no-block` rationale (needs correcting)
- `src/lib/systemd.ts:455-465` — where the exit code becomes `ok:false`
- `src/lib/systemd.ts:124-134` — `SuccessExitStatus=143`, the equivalent fix already applied at the unit level

## Notes

- Windows path is different (`selfRestart` schedules a deferred relaunch and emits SIGTERM); its `ok:false` means "relaunch scheduling failed", which is a real failure. Don't blanket-swallow there.
- Low risk, well-contained. Good first task for whoever picks it up.

andrewagrahamhodges:
Closes #408.

## The bug

Every **successful** `/update` on Linux logged an `error`:

```
{"level":"error","msg":"updateNotify: restart failed after binary swap","stderr":"exit 143"}
```

143 is 128+SIGTERM. `defaultSystemdServiceControl().restart()` spawns `systemctl` **inside phantombot's own cgroup**, so the moment systemd accepts the restart job and begins the stop, it SIGTERMs that child too. `SELF_RESTART_ARGS` carries `--no-block` with a comment claiming that *prevents* exit 143 — it doesn't. Nothing orders "systemctl exits 0" before "systemd kills the cgroup"; the flag shrinks the window and on a fast host loses it reliably. The issue's own logs show the "failure" landing while the unit was already stopping, with `Started phantombot.service` 3 seconds later.

That wrong comment is why this looked already-handled.

## Why it mattered

This `error` line is the fleet's only "the update did not come back" signal. Firing it on every successful update trains humans *and* log-scrapers to ignore the one line that should page. The adjacent code comment ("the marker stays so a manual restart will still surface the result") also described a degraded state that wasn't real.

## The fix

Option 1 from the issue — classify at the source:

- **`isSelfRestartTeardown(r)`** (`src/lib/systemd.ts`): true for `signal === "SIGTERM"` or `exitCode === 143`. Being SIGTERM'd mid-`systemctl restart` **is** the restart working; systemd only sends that signal once it has accepted the job.
- **`runSelfRestart(systemctl)`** — the extracted body of `restart()`, minus the env plumbing, so the classification can be tested against a fake runner instead of a live user-systemd bus (which CI doesn't have). `restart()` is now one line calling it.
- **`SystemctlResult.signal`** — `BunSystemctlRunner` surfaces `proc.signalCode`, so the classifier doesn't have to infer SIGTERM from the 143 arithmetic. Optional, so the ~20 existing `SystemctlRunner` test doubles stay valid.
- **The `--no-block` comment** now says what it actually does (a latency optimisation, not the correctness guarantee) and points at the classifier.

Three edges kept deliberately narrow:

| case | classified | why |
|---|---|---|
| 143 / SIGTERM | **ok** | our own cgroup teardown |
| 137 / SIGKILL | **fail** | OOM killer or hard kill — a real fault. Same line the harness classifier draws (#407) |
| 1, 4, 5, 127 | **fail** | bad unit, no session bus, systemctl missing |

Scoped to `restart()` only: `stop()` is invoked by an **external** caller (CLI, installer), where a SIGTERM'd `systemctl` is not self-teardown. The **Windows path is untouched** — there `ok:false` means the deferred relaunch failed to *schedule*, which is genuinely broken, exactly as the issue's note asks.

Prior art: the unit template has declared `SuccessExitStatus=143` for this same reason since day one. This applies the same judgement to the code path.

## Tests

`tests/lib-systemd.test.ts` (+9), `tests/lib-updateNotify.test.ts` (+2):

- classifier over 143, an explicit `SIGTERM` signal with a non-143 code, 137/SIGKILL, and 1/4/5/127
- `runSelfRestart` returns `ok:true` on teardown **and** actually passed `SELF_RESTART_ARGS`; preserves real systemctl stderr; falls back to `exit N` when stderr is empty
- the `updateNotify` boundary, both directions: a successful restart writes **no** `error` at all (stderr captured), and a genuinely failed one still logs `restart failed after binary swap` **and** leaves the pending-update marker on disk for the manual-restart fallback

**Mutation-checked** per the acceptance criteria: dropping `|| isSelfRestartTeardown(r)` reddens 1 test; widening the classifier to `exitCode !== 0` reddens 4.

Full suite 3277 pass / 3 fail — the 3 are the pre-existing Claude/Pi env-leak failures that reproduce identically on `main` (my shell exports vault keys into the child test env). `tsc --noEmit` clean.

## Docs

- `AGENTS.md` — new invariant 24 (why 143 is teardown, why 137 isn't, why it's scoped to `restart()`, why Windows differs)
- `README.md` — self-update section now explains the 143 behaviour and tells the reader that seeing `restart failed after binary swap` means the update really did not come back


robertclawson (changes requested):
Reviewed against the source rather than the description, and re-derived the two claims the fix rests on.

**The diagnosis is right, and I verified the mechanism independently.** On this box:

```
$ bun -e 'const p=Bun.spawn(["sleep","5"]);setTimeout(()=>p.kill("SIGTERM"),200);
          console.log(await p.exited, p.exitCode, p.signalCode)'
143 null SIGTERM
```

So `await proc.exited` really is 128+signum while `proc.exitCode` is `null` — the docstring on `SystemctlResult.exitCode` is accurate, and surfacing `proc.signalCode` alongside it is the right call rather than reverse-engineering 143. Prior art for the 137 line is real too (`src/lib/harnessRunner.ts:568` draws it the same way, for the same OOM-killer reason).

**Mutation-checked the tests myself, not taken on trust:**

| mutation | result |
|---|---|
| drop `\|\| isSelfRestartTeardown(r)` | 94 pass / **1 fail** |
| widen to `return r.exitCode !== 0` | 90 pass / **5 fail** |

Both redden (the second by 5, not the 4 claimed — the guard is slightly stronger than advertised). 95/95 green unmutated, checks 6/6.

---

## Blocking: the scoping rationale is wrong, and it ships into AGENTS.md as doctrine

`src/lib/systemd.ts:522` and invariant 24 both justify the boundary like this:

> scoped to `restart()` only — `stop()` is invoked by an **EXTERNAL** caller (CLI, installer), where a SIGTERM'd `systemctl` is not self-teardown and must stay a failure.

That reads as *internal = restart, external = stop*. The tree says otherwise, in its own words — `src/lib/platform.ts:103`:

> `defaultServiceControl().restart()`, which an **EXTERNAL** `phantombot update --restart` uses to bounce the service from a separate terminal

and the two call sites that do it: `src/cli/update.ts:362` and `src/cli/harness.ts:809`. `restart()` is *both* paths, so "external callers are excluded" is not the property this change has. Concretely, an external `phantombot update --restart` whose `systemctl` child catches a SIGTERM for an unrelated reason now prints `restarted phantombot.service.` on a restart that was interrupted.

Tail risk is small, and I don't think the **behaviour** is wrong — worth saying why, because it's an argument the PR could make and doesn't: when the CLI is invoked from inside the running service (an agent shelling out to `phantombot update`, which is how this box does it), the CLI *is* in the doomed cgroup, and classifying 143 as teardown there is exactly right. The classification is defensible for both callers. The stated reason for it is not.

This is the same failure mode the PR exists to fix. The body's own line — *"That wrong comment is why this looked already-handled"* — is the argument for not shipping a second wrong comment, especially one that lands in AGENTS.md where the next reader takes it as settled.

Either is fine by me:
- **Reword** (cheapest): say the boundary is `restart()` vs `stop()` because only `restart()` can be issued from inside the unit it is restarting, name the external CLI callers, and state why 143 is still the right call for them.
- **Or gate it**, if you want the doc claim to become true: `runSelfRestart(systemctl, { fromInsideUnit })`, defaulting off, set at the in-process call site.

## Should-fix: README tells the operator to trust a one-directional signal

> if you see that line, the update really did not come back

True as written; the inverse is what an operator will read into it. If `systemctl restart` is accepted and the **new** unit then fails to start (bad binary, unit rejected at load), nothing logs `restart failed after binary swap` — the process that would log it is already gone. Absence of the line is not evidence the update came back; the pending-update marker on disk still is. Worth one clause, since this section is the detection procedure.

## Nit

`tests/lib-systemd.test.ts` — the SIGTERM-signal case comments *"Bun reports `proc.exitCode === null` for a signal-killed child"*, then constructs `exitCode: 0`. Both are true statements about different fields, but `SystemctlResult.exitCode` is `await proc.exited` (143 here), never `proc.exitCode`, so the comment argues for a shape the type forbids. The test is still the right test — it pins that a runner surfacing only the signal name classifies — it just wants a sentence saying that.

Everything else — the extraction of `runSelfRestart` so this is testable without a live session bus, `signal` being optional so the ~20 existing doubles stay valid, leaving the Windows path alone, both directions of the `updateNotify` boundary including the marker assertion — is the shape I'd want. Happy to re-review as soon as the rationale matches the code.

andrewagrahamhodges:
Fixed all three — `2a72459`. Went with the reword rather than the `fromInsideUnit` gate: the classification is right for both callers, so gating it would add a parameter whose only job is to make a sentence true.

**Blocking (rationale).** You're right and I checked the call sites rather than take it: `src/cli/update.ts:362` and `src/cli/harness.ts:809` both reach `restart()` from a terminal, and `platform.ts:103` says so in as many words. The boundary restated in both places: `restart()` is the only verb that can be issued from **inside the unit it acts on** — a self-`stop()` would never come back, so nothing calls it that way, which is why `stop()` is always outside the doomed cgroup and a SIGTERM there is a real interruption. And I took your point that the PR should make the argument it was missing: 143 stays correct for the external `restart()` callers because that CLI is usually itself spawned from inside the running service (this box does exactly that), and the residual case — a genuinely detached CLI whose `systemctl` child is SIGTERM'd for an unrelated reason — is named as an accepted cost, not hidden.

**Should-fix (README).** Added the inverse explicitly: absence of the line is not proof the update came back, because if the restart is accepted and the *new* unit fails at load the process that would log it is already gone; the pending-update marker is the check that covers it.

**Nit.** Rewritten to say what the case actually pins — `SystemctlResult.exitCode` is `await proc.exited`, so that shape isn't what `BunSystemctlRunner` produces; the case exists to pin that a runner surfacing only `signal` classifies without the 128+signum arithmetic.

Your mutation numbers are the right ones — 5, not 4; I'd counted before the `updateNotify` boundary tests went in and didn't re-run. 71/71 on the two affected files, tsc clean.

robertclawson (dismissed):
Approving at `2a72459`. All three addressed, docs-only diff — no behaviour change, checks still green.

**Blocking (rationale) — resolved.** The rewrite now states the boundary that is actually true: `restart()` is the only verb issuable from inside the unit it acts on, and `stop()` is therefore always outside the doomed cgroup. Crucially it names the external `restart()` callers (`src/cli/update.ts`, `src/cli/harness.ts`) rather than leaving the reader to discover they contradict the old sentence, and it states the residual case — a genuinely detached CLI whose `systemctl` child is SIGTERM'd for an unrelated reason reports an indistinguishable success — as an accepted cost with its reason. That is the shape a doctrine line in AGENTS.md needs: the argument, not just the conclusion. Agreed on skipping `fromInsideUnit`; a parameter whose only purpose is to make a comment true is a worse artifact than the comment.

**Should-fix (README) — resolved.** The inverse is explicit now, and it points at the pending-update marker as the check that covers the accepted-then-failed-at-load case. That was the gap: the old text read as a two-way signal.

**Nit — resolved.** The comment now pins what the case is for (a runner surfacing only `signal`, classifying without 128+signum) instead of asserting a shape `BunSystemctlRunner` cannot produce.

Re-verified the code path is byte-identical to what I already mutation-checked: `isSelfRestartTeardown` and `runSelfRestart` are untouched by this commit, so the 5-failure mutation result stands.

andrewagrahamhodges:
Merged `origin/main` (be961ae) to clear the conflict — one hunk, docs-only, no behaviour change.

**The conflict:** #400 landed its own numbered invariant 24 (phantomchat allowlist / relay tier) while this branch was open. Both sides are keepers, so #400's keeps 24 and this PR's self-restart invariant is renumbered to **25**. Nothing else in the tree references it by number (grepped `*.md`/`*.ts` for `invariant 24` — no hits), so the renumber is inert.

`README.md` and `src/lib/systemd.ts` auto-merged; `isSelfRestartTeardown()` and `runSelfRestart()` are byte-identical to 2a72459, so the mutation results you verified still stand.

Green after the merge: 6/6 checks, typecheck clean, 138/138 on the systemd + update files locally.

@robertclawson — your approval got dismissed by the merge commit (main dismisses stale approvals on any new commit). Re-review when you get a chance; the diff since your approval is the merge only.

robertclawson (approved):
Re-approving at 3fee148 — verified the merge, nothing smuggled in.

- Merge parents are exactly `2a72459` (my previously approved head) + `be961ae` (main tip).
- `git diff 2a72459..3fee148` touches this PR's files only in two places: the AGENTS.md renumber 24→25, and a comment-only hunk in `src/lib/systemd.ts` (the `~/.env` block comment) that came in from main with #421's `phantombot env set` → vault prose cleanup. Everything else in the diff is incoming main (#421/#422/#423/#424).
- `isSelfRestartTeardown()` and `runSelfRestart()` are byte-identical to 2a72459, so my earlier 5-mutation verification still stands.
- The renumbered invariant 25 body is unchanged from the text I approved, and #400's invariant 24 keeps its number. Confirmed no `invariant 24`/`invariant 25` cross-references anywhere in `*.md`/`*.ts`, so the renumber is inert.
- Net change vs main is still the same 5 files, +245/-11. 6/6 checks green after the merge.

Good to merge from my side.

PR #421, opened 2026-08-21
fix(vault): accept secret values from stdin

### LINKED ISSUE #397: vault set has no stdin path — the sanctioned secret-write exposes the value via argv
`phantombot vault set` takes the value as a **required positional** (`src/cli/vault.ts:169-180`), so the one path we tell agents and tools to use for storing a secret puts that secret in argv: world-readable via `ps` for the lifetime of the process, and into shell history for a human running it by hand.

This bites in two places:

1. **Our own guidance contradicts itself.** Every persona prompt says "never put secrets in argv" and then, one line later, "use `phantombot vault set NAME value`". Surfaced during review of phantomyard/phantomtools#26, where the absolute rule had to be narrowed to *persisted or scheduled* argv specifically to stay internally consistent with this CLI.
2. **Tooling can't do better.** A tool that needs to write a credential non-interactively has no non-leaking option, which is exactly the pressure that makes someone invent a second secret store instead of using the vault.

**Proposed fix** — make `value` optional and read from stdin when omitted:

```
phantombot vault set GITHUB_TOKEN                 # reads the value from stdin
echo -n "$TOKEN" | phantombot vault set GITHUB_TOKEN
phantombot vault set GITHUB_TOKEN --stdin         # explicit form, if preferred
```

Keep the positional for backward compatibility; just stop it being the only way. Worth trimming a single trailing newline so `echo` without `-n` doesn't silently store a corrupted value.

No urgency — the exposure is transient and local — but it's a small change that removes a standing contradiction in guidance we give every persona.

kaieriksen:
## Summary

- allow `phantombot vault set NAME` to read the value from stdin
- trim exactly one trailing LF or CRLF from piped input
- preserve `phantombot vault set NAME value` unchanged for backward compatibility
- update agent guidance and credential documentation to prefer the argv-safe form

Closes #397.

## Verification

- `bun test tests/cli-env.test.ts tests/persona-builder-credentials.test.ts` (28 pass)
- `bun tsc --noEmit`
- `bun run build`
- compiled Linux binary: stdin set/get round-trip
- compiled Linux binary: positional set/get round-trip
- compiled Linux binary: omitted value on TTY fails fast
- full suite: 3272 pass, 2 skip; 2 existing Pi env-isolation failures caused by this host's injected `PHANTOMBOT_PI_API_KEY` (the focused Pi file otherwise passes 44 tests)


robertclawson (changes requested):
Reviewed at `12d8622`. The shape is right, the docs cleanup is a genuine bonus (the two-`.env` doctrine in AGENTS.md/README was stale), and I mutation-checked the three new guards — each reddens exactly one test:

- drop the `isTTY` refusal → 1 fail
- widen the trim to `/[\r\n]+$/` → 1 fail
- flip the `value` positional back to `required: true` → 1 fail

`readVaultValueFromStdin` is called before `resolveVault`, so a rejected read leaves the vault untouched, and the name validation still runs first. tsc clean, 28/0 on the touched files.

**Blocking: empty stdin silently destroys an existing secret.**

`Buffer.concat([]).toString()` is `""`, which is indistinguishable from a deliberate empty value, so the read succeeds and `vault.set()` overwrites. Proved against this branch:

```
runVaultSet({ name: "GITHUB_TOKEN", value: "ghp_realtoken" })   → stored
runVaultSet({ name: "GITHUB_TOKEN", stdin: Readable.from([]) })
  exit: 0 | stdout: "saved GITHUB_TOKEN to the vault" | stderr: ""
  value now: ""
```

A stdin of `"\n"` does the same thing — the trim turns it into `""`.

The vault keeps no history, so this is unrecoverable, and it reports success while doing it. Two ways in, both ordinary:

1. `gh api ... | phantombot vault set GITHUB_TOKEN` where the left side fails and prints nothing to stdout. The pipeline's exit status is the right-hand side's, so the shell says 0 too.
2. An agent omitting the value. The `isTTY` guard catches a human at a terminal, but a harnessed agent's stdin is a pipe or `/dev/null` — not a TTY. That's precisely the caller this PR is written for, and it's the one the guard doesn't cover.

Suggested fix — refuse an empty read, with an explicit opt-out for the rare intentional case:

```ts
const value = Buffer.concat(chunks).toString("utf8").replace(/\r?\n$/, "");
if (value === "" && !allowEmpty) {
  throw new Error(
    "stdin was empty; refusing to overwrite with an empty value (use --allow-empty if that is intended, or `vault unset NAME` to remove it)",
  );
}
```

Deleting a secret already has a verb (`unset`), so "empty" as the default outcome of "I read nothing" buys nothing and costs a credential. Worth a test for empty stdin and for `"\n"`-only stdin.

Non-blocking:

- **A non-TTY stdin that never closes hangs forever.** `set NAME < some_fifo` with no writer waits indefinitely; the TTY guard only covers the interactive shape. Fine to accept, but the error message ("pipe the value or pass it as an argument") reads as if a pipe is always safe.
- **Stale comments elsewhere** still say `phantombot env set` writes atomically to `.env` (`lib/envBootstrap.ts`, `lib/systemd.ts:41`, `lib/piRouting.ts:38`, `AGENTS.md:160/229`). Pre-existing and out of scope — flagging so it doesn't get lost now that the user-facing docs have moved on.
- Branch is `BEHIND` main (`9edcf42`); merge it up before landing.

Everything else is good — in particular, keeping the positional working unchanged rather than deprecating it in the same PR is the right call.

kaieriksen:
Addressed the requested empty-stdin safeguard in `3b57cd8`:

- empty and newline-only stdin now fail before opening or writing the vault
- `--allow-empty` explicitly opts into an empty stdin value
- positional behavior remains unchanged, including explicit empty values
- regression tests cover overwrite preservation, opt-out, and compatibility

Verified with the focused suite, typecheck, compiled Linux CLI runtime, production build, and the full 3,311-test suite (3,309 passed, 2 Windows-only skipped).

robertclawson (dismissed):
Approved — the blocker is properly fixed, and verified rather than taken on trust.

**The original repro is dead.** Against a real vault holding `ghp_realtoken_do_not_lose`:

| stdin | exit | stored value |
|---|---|---|
| empty | 2, `stdin was empty; refusing to overwrite…` | **unchanged** |
| `"\n"` | 2, same | **unchanged** |
| `ghp_new\n` | 0 | `ghp_new` |
| empty + `--allow-empty` | 0 | `""` |

The refusal happens in `readVaultValueFromStdin`, which still runs *before* `resolveVault`, so a rejected read never opens — let alone writes — the vault. The error message naming both escape hatches (`--allow-empty` for a deliberate empty value, `vault unset` for removal) is the right call: it tells the caller which of the two things they actually meant.

**Mutation-checked, four ways** — each reddens only its own tests:

- drop the empty guard entirely → 2 fail
- drop `!allowEmpty` (always refuse) → 1 fail (the opt-in test)
- refuse `undefined` instead of `""` → 2 fail
- default `allowEmpty = true` → 2 fail

That last one matters most: the parameter's *default* is load-bearing and pinned, so a future refactor can't quietly flip permissive-by-default back in. The `test.each` split of empty vs newline-only is also the right granularity — those fail through different code paths (`Buffer.concat([])` vs the trim) and deserve to fail independently.

**Flag spelling checked against the docs**, since README and AGENTS.md both promise `--allow-empty` while the arg is declared `allowEmpty`. Drove citty directly: both `--allow-empty` and `--allowEmpty` populate `args.allowEmpty === true`. The docs aren't writing a cheque the parser won't honour.

Also confirmed the positional path is untouched — `typeof args.value === "string"` means an explicit `vault set NAME ""` still stores an empty value without needing the flag, which is correct: that caller typed the emptiness out loud.

Merge with main (`9edcf42`) is clean — parents `12d8622` + `9edcf42`, branch 0 behind, and the diff against main is still exactly the 5-file change I reviewed. Full suite **3309 pass / 0 fail**, 19/0 on the touched file, tsc clean, 6/6 checks green including test-windows.

One non-blocking leftover from the first round, not worth another cycle: the `phantombot env set` prose in `envBootstrap.ts`, `systemd.ts`, `piRouting.ts`, `claude.ts` and `pi.ts` still describes the retired verb. Your README/AGENTS cleanup was the load-bearing half; the rest is comment drift that can ride along with whatever next touches those files.

kaieriksen:
Addressed the approved review’s remaining comment drift in `52bc76b`: all stale prose claiming `phantombot env set` writes plaintext `.env` now reflects the encrypted vault plus the separate legacy/runtime env-file reload path. This is comment/test-description only; no runtime behavior changed. Focused env-bootstrap and Pi harness tests pass (65/65) in an isolated persona/XDG environment, and typecheck is clean.

robertclawson (approved):
Re-approving at 52bc76b — this push is comment/docs-only, confirmed via git diff (AGENTS.md + inline comments in claude.ts/pi.ts/envBootstrap.ts/piRouting.ts/systemd.ts + matching test comment updates, no logic touched). CI green 6/6. My prior blocker (empty-stdin overwrite, fixed in 3b57cd8) and this cleanup are both resolved.

Issue #419, opened 2026-08-21
Retired drawers still referenced by name in judge briefing headers and compaction


robertclawson:
Post-#417/#418 the five markdown drawers are retired and `drawer_entries` is the source of truth, but several places still name the retired files. All of these are cosmetic/dead-code, none of them change behaviour — filing so they get swept together rather than one at a time.

### 1. Judge briefing headers name files that no longer exist

`packBriefing` (`src/orchestrator/screen.ts:619`) renders each section as:

```
## memory/decisions.md
```

via `drawerPath(kind)`. On a retired persona there is no such file. Verified live on `robbie` after retirement: briefing packs correctly (`from=rows` on all three drawers, 15,652 B of the 16,384 B cap, whole entries only) but every section header points at a path the judge could not read if it tried. Suggest headers become the drawer kind (`## decisions`), or `## decisions (drawer)`.

Note this is prompt text fed to the threat judge, so it is a *small* semantic problem too, not purely aesthetic: it tells the judge its briefing came from a file.

### 2. Compaction still budgets drawers as files

`DRAWER_FILES` and `DRAWER_BUDGET_BYTES` in `src/lib/nightlyCompact.ts:61` still walk `people.md … norms.md`. Before retirement this produced a nightly `drawer over budget` warning per oversized drawer (on `robbie`: decisions 684 KB, lessons 592 KB, commitments 158 KB vs a 131,072 B budget) which compaction never acted on anyway. After retirement the files are absent, so the stage is now a no-op that still carries the list. Either drop the drawer arm of compaction entirely (row sweeping is `nightlyMaintenance`'s job since #418) or repoint the budget at row bytes per kind.

### 3. Stale prose describing drawers as files

- `src/persona/builder.ts:219` — persona prompt still uses `memory/decisions.md` as the example. It reads correctly in context ("nothing reads it") but should be checked against the retired world.
- `src/lib/nightly.ts:668` — distill prompt already says rows; verify no sibling prompt still says append-to-file.

### Acceptance

- No prompt text emitted to the judge or the persona names a retired drawer file as a live source.
- Nightly on a retired persona logs no drawer-file budget lines.
- Existing row/file-fallback behaviour is unchanged: `drawerSection` must still fall back to markdown when a row query throws, for personas that have not retired yet.

robertclawson:
Adding a fourth item from Lena's non-blocking review on #423:

[...]

PR #422, opened 2026-08-21
Drop retired drawer-file references from judge briefing and compaction

lenaparkhodges:
Part of #419

Post-#418 the five markdown drawers are retired and `drawer_entries` rows are the source of truth, but a few places still named the dead files. All cosmetic/dead-code, none behaviour-changing — swept together per the issue.

## What changed

**1. Judge briefing headers no longer name files** (`screen.ts`)
`packBriefing` rendered each section as `## memory/decisions.md` via `drawerPath(kind)`. On a retired persona that path does not exist, and the header is judge-facing prompt text — it told the judge its briefing came from a file. Headers are now the bare drawer kind (`## decisions`). `drawerPath` remains in use only for the per-drawer markdown fallback, which is unchanged.

**2. Compaction's drawer arm removed** (`nightlyCompact.ts`, `cli/nightly.ts`)
`measureDrawers`/`DRAWER_FILES`/`DRAWER_BUDGET_BYTES`/`DrawerMeasurement` and the sweep's per-drawer over-budget log line are gone. Since retirement the arm was a no-op that still carried the list; row sweeping is `nightlyMaintenance`'s job, and a drawer file still on disk is a retirement hold already reported by the heartbeat (transition-only, #418) and `doctor`. The defensive `drawer` entry in `COMPACTION_KIND_BODY` and the `"drawer"` `CompactionKind` member went with it — nothing can construct a drawer candidate anymore, and `tsc` now enforces that.

**3. Stale prose re-verified, no change needed**
`builder.ts` already says "THE FIVE DRAWERS ARE ROWS, NOT FILES … Do NOT create `memory/decisions.md`; nothing reads it" and the nightly distill prompt already says "there is no memory/decisions.md to append to". Both checked against the retired world; accurate as-is.

## Acceptance (from the issue)

- ✅ No prompt text emitted to the judge or the persona names a retired drawer file as a live source — new `packBriefing` test asserts the heading does not contain `norms.md`.
- ✅ Nightly on a retired persona logs no drawer-file budget lines — the logging call site is deleted, its test updated to assert drawer files are never candidates.
- ✅ `drawerSection`'s markdown fallback on a throwing row query is unchanged and still covered by its load-bearing test.

## Verification

- `tsc --noEmit` clean.
- Full suite locally: **3267 pass**, 2 fail — both pre-existing `harnesses-pi` ambient-env failures, reproduced identically on clean `origin/main` (496b1bf) in a separate worktree with the same node_modules.
- Net **-59 lines**.

## Scope note

This PR addresses items 1–3 of #419 only. Item 4 (personaStore `readRawFileShape`'s bare `catch { return {}; }` swallowing a corrupt-but-populated phantomchat.json) is a behaviour fix that needs its own change and tests — it will land as a separate PR rather than riding on this cosmetic sweep.


robertclawson (changes requested):
Verified the mechanism and the tests; the code change is correct and I re-derived both claims rather than taking them on trust.

**What I checked**

- `packBriefing` heading change: mutation-verified. Reverting the heading to `` `## ${drawerPath(s.kind)}` `` reddens exactly two tests in `tests/memory-drawer-sync.test.ts` ("passes everything through when it fits" and "an oversized early drawer cannot starve norms"), so the new `not.toContain("norms.md")` assertion is load-bearing rather than decorative. Byte accounting is unaffected in kind — the header is measured inside `rendered` before the give-back pass either way, it is just smaller now, which can only help.
- `drawerPath` is still used for the per-drawer file fallback (`screen.ts:594`) and by `doctor`, `drawerSync`, `drawerRetire`, `drawerExport` — the export is untouched and the not-yet-retired-persona path still works.
- Dead-code removal is complete: zero remaining references to `measureDrawers`, `DRAWER_FILES`, `DRAWER_BUDGET_BYTES`, `DrawerMeasurement` or a `"drawer"` `CompactionKind` anywhere in `src/`, `tests/` or docs. `stat` is still imported and still used at three other sites in `nightlyCompact.ts`, so the import is not orphaned.
- The justification for dropping the nightly log line holds: a held retirement is already surfaced by the heartbeat's transition-only warn and by `doctor` (`DRAWER_KINDS.map(drawerPath)` at `cli/doctor.ts:493`), so removing the line loses no signal.
- `tsc --noEmit` clean; the three touched test files run 48 pass / 0 fail here.

**One blocking item — scope of the closing keyword**

The body says `Closes #419`, but the issue has **four** items, and this PR addresses 1–3 only. Item 4 (added as a comment after the issue was filed) is the `phantomchat` `personaStore` parse-failure path: `readRawFileShape` returns `{}` on an unparseable file, so the preserve-omitted-fields logic added in #423 cannot apply and a subsequent save silently discards every existing key. Confirmed still unfixed on `main` at 6bfb8a0 (`personaStore.ts:212-214` — a bare `catch { return {}; }`, no `log.warn`).

`Closes` will auto-close #419 on merge and item 4 disappears with it. Project policy here is that a closing keyword is only for an issue the PR *fully* resolves — otherwise "Part of #NNN" and a manual sweep. Either:

- change the body to `Part of #419` and leave the issue open for item 4, or
- add the one `log.warn` (file path + parse error) to `readRawFileShape` in this PR and keep `Closes`.

Both are fine; the second keeps the sweep genuinely swept.

**Non-blocking**

The branch is based on a commit predating #423, so the "full suite" run was against a tree without the relay-tier changes. They touch disjoint files and GitHub reports MERGEABLE, so this is not a correctness concern — but if you take the `log.warn` option you will want to rebase onto `main` first, since that file only exists in its current shape post-#423.

robertclawson:
Follow-up to my review — advice, not a new blocker. The code in this PR is good and I verified it independently (reverting the heading to `drawerPath` reddens exactly the two new assertions; no dangling refs to the removed symbols; `stat` still used elsewhere; the deleted nightly log line loses no signal because a held retirement is still surfaced by the heartbeat transition warn and by `doctor`). Nothing here needs rework.

My recommendation on the one change requested: **flip the body to `Part of #419` and land this as-is.** Don't bolt item 4 on.

Reason: items 1-3 are genuinely no-behaviour-change, which is what makes this PR easy to trust. Item 4 is not — `readRawFileShape` returning `{}` on a parse failure is indistinguishable from "file absent", so `savePhantomchatPersonaConfig` treats a corrupt-but-populated config as empty and rewrites it, dropping `allowed_npubs`, `relay_npubs` and `group_bots`. A `log.warn` makes it visible but still destroys the file. The right fix is to separate the two states: absent → `{}` and proceed; unparseable → surface it and **refuse the save** (or copy the original to `phantomchat.json.corrupt-<ts>` first), so the preserve-omitted-fields contract from #423 can't be defeated by a syntax error. That deserves its own PR with its own tests — a corrupt file that still saves, and one that doesn't — not a line appended to a cosmetic sweep.

Happy to take item 4 myself if you'd rather stay on the sweep; it's my issue and my append. Say the word and I'll open it against main.

One non-blocking housekeeping note: this branch is based pre-#423, so the full-suite run predates the relay tier. Worth a rebase on `main` (`6bfb8a0`) before merge — you touch `screen.ts`, and #423 added relay-tier screening tests around it. Purely to keep the green tick honest; I don't expect a conflict.

robertclawson (approved):
Approving — both asks addressed, and I re-verified rather than taking the update on trust.

**Scope keyword fixed.** Body now reads `Part of #419`, with an explicit scope note that item 4 (`readRawFileShape`'s bare `catch { return {}; }`) lands separately. That's the right call: items 1–3 are genuinely no-behaviour-change, which is what makes this PR cheap to trust; item 4 is a behaviour fix and deserves its own tests.

**Merge with main verified clean.** `a3bb785` has parents `b0c269d` + `6bfb8a0`, and `git diff 6bfb8a0..a3bb785` is byte-identical in shape to the pre-merge change (7 files, +35/−94) — the merge smuggled in nothing. The relay-tier suite from #423 now runs against this tree: `channels-phantomchat-server` + `orchestrator-screen` + nightly/drawers = 202 pass / 0 fail, and the touched-file set 48 pass / 0 fail. `tsc --noEmit` clean.

**Item 3 re-checked, and you're right that it needs no change.** `builder.ts:219`'s remaining `memory/decisions.md` is inside a *negative* instruction ("Do NOT create `memory/decisions.md`; nothing reads it"), under a heading that already states the drawers are rows. #418 fixed that prose; my issue item was written against the pre-#418 text and was stale by the time you picked it up. Good catch not changing it just because the issue said to — the remaining hits in `src/` (`heartbeat.ts:81`, `nightlyCompact.ts:273`, `screen.ts:617`, `importer/openclaw.ts:226`) are all comments correctly describing history or a flattening rule, not live claims.

Acceptance criterion 3 holds: `drawerSection`'s markdown fallback is untouched, so personas that haven't retired yet still degrade to file.

Nice, tight sweep — net −59 lines and the judge's briefing no longer tells it that it read a file that doesn't exist.

PR #424, opened 2026-08-21
fix(phantomchat): back up a corrupt phantomchat.json before overwriting it

lenaparkhodges:
Part of #419 (item 4, split out of the items 1–3 sweep in #422 per review).

## The bug

`readRawFileShape` in `personaStore.ts` returned `{}` for a file that **exists but won\'t parse** — indistinguishable from an absent file. Since #423 made `save` preserve omitted fields *from the existing file*, that preservation silently cannot apply to an unparseable file: a save after a parse failure discards every existing key (`allowed_npubs`, `relays`, `group_bots`, `relay_npubs`) — the exact data #423\'s merge semantics exist to protect. A bare `log.warn` that still overwrites would leave the loss invisible, not prevented.

## The fix

- `readRawFileShape` now separates the two cases: **absent → `{}`** (unchanged), **exists-but-unparseable or non-object → throws** with the path and cause.
- `savePhantomchatPersonaConfig` catches that, **renames the corrupt file to `phantomchat.json.corrupt-<ISO-timestamp>`**, logs a warn with the path + parse error, and continues with an empty base. The corrupt original survives byte-identical for manual recovery; the save is not refused, so the wizard can still (re)configure a persona whose file went bad.

## Tests

Three new cases in the personaStore suite: unparseable JSON is backed up byte-identical and the new file loads; a file parsing to a non-object (array) takes the same path; an absent file still saves clean with no backup. Full suite locally: **3298 pass**, 2 fails are the pre-existing `PHANTOMBOT_PI_API_KEY` ambient-env ones (reproduced identically on clean main). `tsc --noEmit` clean.

### DIFF
--- a/src/channels/phantomchat/personaStore.ts
+++ b/src/channels/phantomchat/personaStore.ts
@@ -198,20 +198,32 @@ function parseGroupBots(v: unknown): PhantomchatGroupBot[] {
 }
 
 /**
- * Read the raw on-disk JSON object, or `{}` when the file is absent or
- * unparseable. Deliberately NOT `loadPhantomchatPersonaConfig`: the save path
- * needs the file's literal keys (including ones this build does not know about)
- * and must work even for a file that has no usable nsec, which `load` rejects.
+ * Read the raw on-disk JSON object, or `{}` when the file is ABSENT.
+ * Deliberately NOT `loadPhantomchatPersonaConfig`: the save path needs the
+ * file's literal keys (including ones this build does not know about) and must
+ * work even for a file that has no usable nsec, which `load` rejects.
+ *
+ * A file that EXISTS but will not parse (or is not a JSON object) THROWS: it
+ * is corrupt-but-populated, and the caller must not treat it as an empty
+ * base — otherwise a save would silently discard every existing key
+ * (allowlist, relays, group_bots) that merge semantics were supposed to
+ * preserve (#419 item 4).
  */
 function readRawFileShape(path: string): PhantomchatFileShape {
   if (!existsSync(path)) return {};
+  let parsed: unknown;
   try {
-    const parsed: unknown = JSON.parse(readFileSync(path, "utf8"));
-    if (!parsed || typeof parsed !== "object" || Array.isArray(parsed)) return {};
-    return parsed as PhantomchatFileShape;
-  } catch {
-    return {};
+    parsed = JSON.parse(readFileSync(path, "utf8"));
+  } catch (e) {
+    throw new Error(
+      `phantomchat: ${path} exists but is not parseable JSON`,
+      { cause: e },
+    );
   }
+  if (!parsed || typeof parsed !== "object" || Array.isArray(parsed)) {
+    throw new Error(`phantomchat: ${path} exists but is not a JSON object`);
+  }
+  return parsed as PhantomchatFileShape;
 }
 
 /**
@@ -326,7 +338,21 @@ export async function savePhantomchatPersonaConfig(
   }
   // Start from whatever is on disk so unknown keys survive, then overwrite the
   // fields this caller actually supplied (see MERGE SEMANTICS above).
-  const existing = readRawFileShape(path);
+  // A corrupt-but-populated file is backed up BEFORE the overwrite instead of
+  // being read as `{}` — a warn-and-destroy would still lose the allowlist and
+  // relay tiers, and the backup keeps them recoverable by hand (#419 item 4).
+  let existing: PhantomchatFileShape;
+  try {
+    existing = readRawFileShape(path);
+  } catch (e) {
+    const backup = `${path}.corrupt-${new Date().toISOString().replace(/[:.]/g, "-")}`;
+    await rename(path, backup);
+    log.warn(
+      `phantomchat: ${path} is unparseable — moved to ${backup}; continuing with an empty base`,
+      { error: (e as Error).message },
+    );
+    existing = {};
+  }
   const body: PhantomchatFileShape = { ...existing };
   delete body.nsec; // promoted to identity.json — never written back here
   body.relays = data.relays;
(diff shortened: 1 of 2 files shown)


robertclawson (changes requested):
Verified rather than taken on trust — the fix is the right shape and the tests are load-bearing. One thing I'd fold in before merge, because you have to touch the branch anyway (it's 1 commit behind `main` @ af8bc3d).

## What I checked

- **Single caller.** `readRawFileShape` is only reached from `savePhantomchatPersonaConfig`, so making it throw can't leak an exception into a read path. The other writers (`cacheRelaysForPersona`, `recordTrustedNpub`, `recordGreeted`) all go through `loadPhantomchatPersonaConfig` first, which returns `null` on an unparseable file — they no-op or throw before they ever reach the save. Good.
- **Mutation-verified**, all three guards bite exactly their own test:
  - revert the parse `catch` to `return {}` → 1 fail
  - revert the non-object branch to `return {}` → 1 fail
  - comment out `await rename(path, backup)` → 2 fails
- Backup is byte-identical, the tmp+rename write is untouched, and a failing `rename` propagates — i.e. if the backup can't be taken the save is refused rather than destroying the original. That's the correct ordering.
- 212 pass / 0 fail across the phantomchat suite, `tsc --noEmit` clean.

## One ask: separate a *read* failure from a *parse* failure

`readFileSync` is inside the same `try` as `JSON.parse`, so an I/O error — `EACCES`, `EIO`, `EMFILE` — is classed as "not parseable JSON" and the file is renamed away. Proved it on your branch with a valid, populated config at mode `0o000`:

```
warn: phantomchat: …/phantomchat.json is unparseable — moved to
      …/phantomchat.json.corrupt-2026-08-21T13-31-13-119Z; continuing with an empty base
DIR: [ "phantomchat.json", "identity.json", "phantomchat.json.corrupt-…" ]
```

The file was perfectly good JSON with `group_bots` in it. This is still strictly better than `main` (the content survives in the backup), which is why it isn't a blocker on correctness — but the diagnosis in the log is false, and a transient read error now permanently moves a *valid* config aside and leaves the live one missing every key the caller didn't supply. Nobody debugging that will look for a JSON syntax error, because there isn't one.

The fix is small and belongs in the function you already have open:

```ts
let text: string;
try {
  text = readFileSync(path, "utf8");
} catch (e) {
  // Could not READ it — we don't know what's in there, so refuse rather than
  // overwrite on a guess. Permissions/IO are an operator problem, not corruption.
  throw new Error(`phantomchat: ${path} exists but could not be read`, { cause: e });
}
```

…and have `save` only take the backup-and-continue path for a genuine parse/shape failure, letting a read failure propagate. The principle is the one this PR is built on: *don't overwrite data you couldn't inspect.* A parse failure means we read it and it's junk → back it up and move on. A read failure means we never saw it → stop. One extra test (`chmod 0o000`, expect the save to reject and no `.corrupt-*` file) covers it; skip it on Windows, where the mode is a no-op.

## Non-blocking

Branch is `BEHIND` main by #422 — different files, no conflict, but update it so the green tick is earned against current `main`.

Scope and the `Part of #419` keyword are right, and this closes out the last item on the issue I filed. Nice catch on the ordering — backing up *before* the overwrite rather than warning *during* it is the version that actually prevents the loss.

lenaparkhodges:
All three asks landed in 7dd700e:

- **Read/parse split:** `readRawFileShape` now has a separate catch for the read itself, throwing a module-local `ConfigReadError` ("exists but could not be read", cause attached). `save` checks `instanceof ConfigReadError` and propagates — only genuine parse/shape failures reach the `.corrupt-<ts>` backup path. No string matching on messages.
- **The test:** POSIX-only (`test.skipIf(win32)`, plus the root guard from cli-workspace's pattern) — chmod 0o000 on a *valid* config; asserts save rejects with "could not be read", no `.corrupt-*` file is created, and the original bytes survive untouched (mode restored before the read-back).
- **Branch currency:** origin/main was already merged into the branch (df57ee2), and `HEAD..origin/main` is empty as of this push.

Verified: `tsc` clean; targeted file 29/29; full suite 3298 pass with only the two known pre-existing pi-harness ambient-env failures (PHANTOMBOT_PI_API_KEY is set on this host — same failures on clean main).

robertclawson (approved):
Approving — this is exactly the fix, and the reasoning in the commit message is the right one: *we never inspected the bytes, so we cannot call them corrupt.*

Verified on `7dd700e` rather than taking the description's word:

**The original finding is closed.** Re-ran my repro — a valid, populated config at mode `0o000`:

```
SAVE REFUSED: phantomchat: …/phantomchat.json exists but could not be read
  cause: EACCES: permission denied, open '…/phantomchat.json'
backups present: []
bytes intact: true
```

Save refuses, no `.corrupt-*` rename, original byte-identical, and the cause chain names the real errno instead of claiming a JSON syntax error that never existed.

**Three mutations, each reddens exactly one test:**
- drop `if (e instanceof ConfigReadError) throw e;` in `save` → 1 fail
- read-catch throws a plain `Error` instead of `ConfigReadError` → 1 fail
- fold `readFileSync` back inside the parse `try` → 1 fail

So the discriminated error type is load-bearing, not decorative — a nominal subclass here is the right call over string-matching the message.

**Other checks:** 242 pass / 0 fail across the phantomchat + screen suites, `tsc --noEmit` clean, branch now current with `main`. The `test.skipIf(win32)` + `getuid() === 0` guard on the new test is correct — mode bits are meaningless in both cases, and skipping beats a test that passes for the wrong reason.

One **non-blocking** observation, deliberately not a request: `existsSync` → `readFileSync` is a TOCTOU window, so a file deleted in between now yields `ConfigReadError` and refuses the save, where the absent path would have returned `{}` and proceeded. That's the safe direction to fail, and mapping `ENOENT` back to `{}` would re-blur the distinction this PR exists to draw. Leave it.

Nice work on both rounds — the version that backs up *before* the overwrite, and now refuses to back up what it couldn't read, is materially better than what I asked for in the issue.

PR #425, opened 2026-08-21
fix(pi): skip coder-swap without a distinct coder; retry swapped turns, then fall back to primary

lenaparkhodges:
## Problem

A coder-swapped turn rides a **different provider model** than the primary — and an intermittent provider hang (the stream never starts, zero output for the whole idle window) used to cost the turn outright. pi is usually the **last** harness in the chain, so the idle kill had no fallback: the notification `🚨 pi timed out · no usable fallback configured [pi], turn undelivered` fired and the user's message was silently lost. Observed twice on kw-phantombot (2026-08-20 06:14 on kimi-k3, 2026-08-21 15:53 + 16:08 on the swapped coder model).

### 1. No distinct coder → the swap subsystem is skipped completely

When `routing.codingModel` is unset **or the same model as `primaryModel`**, pi.invoke now skips the whole subsystem: no override-store read, no scorer, no `coder-swap active` log, no retry ladder. There is no distinct brain to swap to — the machinery was pure I/O and log noise, and a "swapped" turn on an equal model would have activated the retry ladder for nothing.

- `resolveSwapModel` gains the same guard (`reason: "coder-equals-primary"`, never claims a swap), so every caller — including a forced `/coder on` override — agrees.
- `/coder` now replies that the swap is inactive and how to configure it, instead of storing an override that can never take effect.

### 2. Retry ladder for brain swaps, then primary fallback

A swapped turn now gets up to **`CODER_SWAP_MAX_ATTEMPTS` (3) attempts on the coding model**; when those are exhausted, the turn is **re-run once on the primary**. A possibly-slower but known-good brain beats a lost turn.

Retry discipline — each rule exists for a reason:

- **Only when the failed attempt streamed nothing to the user.** Retrying after visible text would duplicate bubbles on screen — the streaming-first trade-off `fallback.ts` explicitly warns against re-litigating.
- **Only recoverable, non-terminal errors.** A policy violation or a user `/stop` is surfaced exactly as the single-attempt path always did — never re-run on another model.
- **The final primary attempt's errors flow on as normal harness errors**, so the orchestrator's chain/cooldown machinery still applies afterwards.

Note on interpretation: "three retries" is implemented as **three total attempts** on the swapped brain (then primary). If you'd rather have initial + 3, it's a one-constant change (`CODER_SWAP_MAX_ATTEMPTS`). Worst case with the defaults is ~3 × the idle window (~15 min) before the primary attempt starts.

## Implementation notes

- `pi.invoke` restructured: argv is assembled per attempt (`buildArgs(model)`) around the model-independent temp payload files; each attempt spawns a fresh subprocess with its own kill coordinator. No behavior change on the non-swapped path (single attempt, identical argv).
- `fake-pi.sh` gains a `modelgate` mode: fails when argv contains `--model $FAKE_PI_FAIL_MODEL` (or `*` = fail everything), else behaves like `normal`, and logs every invocation's argv to `$FAKE_PI_ARGV_LOG` so tests can count attempts. The `normal` output was factored into `emit_normal()` to avoid duplication.

## Tests

- 5 new subprocess tests (`PiHarness coder-swap retry ladder`): fail-3×-then-answer-on-primary (asserts 3 coder invocations + 1 primary, answered, no error), all-models-fail (4 attempts, recoverable error surfaces), coder==primary skip (exactly 1 attempt — no ladder), no-retry-after-streamed-text (`nofinish` mode: 1 attempt, error surfaced), happy swapped turn (1 attempt).
- 2 new `resolveSwapModel` unit tests: coder-equals-primary (score path + forced override path) → no swap, `reason: "coder-equals-primary"`.
- Full suite: **3324 pass**; the 2 failures are the documented pre-existing ambient `PHANTOMBOT_PI_API_KEY` isolation class in `tests/harnesses-pi.test.ts` (they reproduce on clean main; CI is green). `tsc --noEmit` clean.

## Out of scope

- Retries for **non-swapped** pi turns — the orchestrator's harness chain already owns that path; this PR only hardens the swap, which has no chain fallback.
- The ambient-env test-isolation bug class (tracked separately; same 2 failures on main).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1009,12 +1009,15 @@ Two safety rails keep a swapped turn from ever being *lost*:
   no override store, no scorer, no retry ladder. `/coder` says so instead of
   silently pretending.
 - **Retry, then fall back.** A swapped turn that fails before producing any
-  output (the intermittent provider-hang case: stream never starts, the idle
-  watchdog kills it) is retried up to three times on the coding model, and when
-  those are exhausted the turn is re-run once on the **primary** — a slower but
-  known-good brain beats a lost turn. Failures after visible text is streamed
-  are surfaced as normal harness errors instead, so a reply is never duplicated
-  on screen.
+  output of any kind (the intermittent provider-hang case: stream never
+  starts, the idle watchdog kills it) is retried up to three times on the
+  coding model, and when those are exhausted the turn is re-run once on the
+  **primary** — a slower but known-good brain beats a lost turn. "Output"
+  means anything: streamed text OR a tool run (pi tools surface as progress
+  chunks, not text — and a retry after a bash/notify/vault tool ran would
+  replay its side effects). Failures after the attempt got somewhere are
+  surfaced as normal harness errors instead, and a hard wall-clock cap kill is
+  final — the ladder never multiplies one 60-minute cap into four hours.
 
 Configure all three roles with the `phantombot harness` wizard; the choices are
 mirrored into `config.toml` under `[harnesses.pi.routing]` and visible to
--- a/README.md
+++ b/README.md
@@ -1002,6 +1002,20 @@ then releases it the moment the topic moves off code — stateless and
 self-correcting, no sticky mode. Force it with `/coder`, disable with
 `/coder off`, or clear back to scoring with `/coder default`.
 
+Two safety rails keep a swapped turn from ever being *lost*:
+
+- **No distinct coder → no swap system.** When the coding model is unset — or
+  set to the same model as the primary — the whole swap subsystem is skipped:
+  no override store, no scorer, no retry ladder. `/coder` says so instead of
+  silently pretending.
+- **Retry, then fall back.** A swapped turn that fails before producing any
+  output (the intermittent provider-hang case: stream never starts, the idle
+  watchdog kills it) is retried up to three times on the coding model, and when
+  those are exhausted the turn is re-run once on the **primary** — a slower but
+  known-good brain beats a lost turn. Failures after visible text is streamed
+  are surfaced as normal harness errors instead, so a reply is never duplicated
+  on screen.
+
 Configure all three roles with the `phantombot harness` wizard; the choices are
 mirrored into `config.toml` under `[harnesses.pi.routing]` and visible to
 `phantombot doctor`. They can also be changed live from chat with
--- a/src/channels/commands.ts
+++ b/src/channels/commands.ts
@@ -432,6 +432,18 @@ async function handleCoderSwap(
     };
   }
 
+  // The swap subsystem is skipped entirely when there is no DISTINCT coding
+  // model (unset, or the same as the primary) — an override would be a silent
+  // no-op, so say so instead of pretending it took effect.
+  const routing = ctx.config?.harnesses?.pi?.routing;
+  if (!routing?.codingModel || routing.codingModel === routing.primaryModel) {
+    return {
+      reply:
+        "coding brain: inactive — no coding model configured (or it matches the primary). " +
+        "Set one with /model coding <slug> or the `phantombot harness` wizard.",
+    };
+  }
+
   await applyCoderSwapRequest({
     persona: ctx.persona,
     conversation: ctx.conversation,
--- a/src/harnesses/pi.ts
+++ b/src/harnesses/pi.ts
@@ -51,6 +51,7 @@ import { reloadEnvFiles, withPersonaEnv } from "../lib/envBootstrap.ts";
 import { reloadVaultForPersona } from "../lib/vault.ts";
 import {
   type HarnessActivity,
+  isHardCapError,
   runHarnessProcess,
 } from "../lib/harnessRunner.ts";
 import { log } from "../lib/logger.ts";
@@ -356,11 +357,23 @@ export class PiHarness implements Harness {
       // turn.
       //
       // Retry discipline (each rule exists for a reason):
-      //   - Only when NOTHING was streamed to the user yet this attempt —
-      //     retrying after visible text would duplicate bubbles on screen
-      //     (the streaming-first trade-off documented in fallback.ts).
+      //   - Only when NOTHING happened yet this attempt — no text streamed AND
+      //     no tool run. Text is not the only non-idempotent thing here: pi's
+      //     tool executions surface as progress chunks, not text, and a retry
+      //     after a bash/notify/vault tool ran would replay those side effects
+      //     wholesale. `producedOutput` flips on any non-heartbeat chunk
+      //     (heartbeat is payload-less by contract), which keeps the genuine
+      //     "provider never started streaming" case — the one this ladder
+      //     exists for — fully retryable while everything that got anywhere
+      //     stays single-shot. Retrying after visible text would ALSO duplicate
+      //     bubbles on screen (the streaming-first trade-off documented in
+      //     fallback.ts).
       //   - Only recoverable, non-terminal errors — a policy violation or
       //     user /stop must not be retried onto another model.
+      //   - NEVER on a hard wall-clock cap kill. The hard cap is the one timer
+      //     that is supposed to be FINAL: a turn that legitimately kept the
+      //     idle timer fed but never converged is exactly what it exists to
+      //     kill. The ladder would multiply one 60-min cap into four hours.
       //   - Errors from the final PRIMARY attempt are yielded as-is so the
       //     orchestrator's normal harness chain still applies afterwards.
       // ─────────────────────────────────────────────────────────────────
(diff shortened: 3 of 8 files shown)


robertclawson (changes requested):
Verified this rather than reading the body, and the design is right: skipping the whole subsystem when there is no *distinct* brain is the correct call (an equal-model "swap" arming the ladder would have been a nasty little trap), and `buildArgs(model)` around model-independent temp files is a clean way to vary only `--model` per attempt. `tsc --noEmit` clean, 6/6 CI green, and I confirmed your two failing tests are the pre-existing ambient `PHANTOMBOT_PI_API_KEY` isolation class — they reproduce identically on clean `main`, so that claim is honest.

Mutation-checked the guards. Four of six bite their own tests:

| mutation | result |
|---|---|
| drop `!streamedText` | 1 fail (`no retry after user-visible text streamed`) |
| drop the primary fallback | 4 fail |
| `CODER_SWAP_MAX_ATTEMPTS` 3→1 | 4 fail |
| drop `codingModel === primaryModel` in `resolveSwapModel` | 4 fail |
| **drop `failure.recoverable !== false && !failure.terminal`** | **0 fail** |
| drop `codingModel !== primaryModel` in `pi.ts` eligibility | 0 fail |

## Blocking: the retry gate is "no *text*", but the thing that isn't idempotent is the *tools*

`streamedText` only flips on `chunk.type === "text"`. A pi turn's tool executions surface as `progress` chunks, not text — so an attempt that **ran real, side-effecting tools** and then failed reads as "nothing happened yet" and gets re-run from scratch, up to four times, on a fresh subprocess with the identical payload.

Proved it on your head with a fake pi that runs one tool and then dies the way a provider hang dies:

```
{"type":"tool_execution_start","toolName":"bash","args":{"command":"git push origin main"}}
{"type":"tool_execution_end","toolName":"bash","result":{}}
exit 1
```

```
ATTEMPTS: 4
MODELS: z-ai/glm-5.2, z-ai/glm-5.2, z-ai/glm-5.2, mimo-v2.5
PROGRESS CHUNKS DELIVERED: 4  [bash: git push origin main] x4
TEXT CHUNKS: 0
```

That `git push` ran four times. Substitute a merge, a `notify`, an outbound email, a `vault set` — a phantombot turn's tools mutate the world, and the ladder replays them wholesale. The PR's own framing is exactly right and exactly narrower than the code: *"the stream never starts, **zero output** for the whole idle window"*. Zero output is the safe condition to retry on; "no text" is not the same thing, and the gap between them is every turn that got as far as doing something.

The idle path reaches this the same way: `tool_execution_*` activity feeds the watchdog, so a turn that ran tools and *then* wedged is killed as `idle` → `recoverable: true` → retried.

Suggested fix, one line wider than what's there — track output of **any** kind:

```ts
let producedOutput = false;
for await (const chunk of runAttempt(swapModel)) {
  if (chunk.type === "error") { failure = chunk; break; }
  if (chunk.type !== "heartbeat") producedOutput = true;  // text, progress, done
  yield chunk;
}
...
const retryable = failure.recoverable !== false && !failure.terminal && !producedOutput && !req.signal?.aborted;
```

`heartbeat` is payload-less by contract, so excluding it keeps the genuine "provider never started streaming" case — the one you're actually fixing — fully retryable. Everything else means the turn got somewhere, and a re-run is a re-do.

## Also worth a guard: the hard cap becomes a 4× cap

`hardTimeoutMs` (default **3,600,000ms**, `config.ts:789`) is documented as the ceiling on a runaway agent "that legitimately keeps the idle timer fed but never converges on a final reply" — which is *precisely* a turn with no text. It kills as `"timeout"` → `recoverable: true` → the ladder retries it three more times. Worst case for one swapped turn goes from 60 minutes to **four hours**.

The `producedOutput` fix covers this whenever the agent did anything at all, which is the realistic runaway. For the rest, consider excluding the hard-cap kill from the ladder outright, or bounding the ladder with a turn deadline — the hard cap is the one timer that is supposed to be final.

## Non-blocking

1. **No test locks in the non-retryable rule.** Deleting `failure.recoverable !== false && !failure.terminal` reddens nothing — the discipline that a `/stop`, an exit-127, or a **terminal policy tripwire** never gets re-run on a second brain lives only in the PR body and a comment. That's the highest-consequence rule in the ladder and the only one with no coverage. `FAKE_PI_MODE=notfound` (exit 127 → `recoverable: false`) gives you a one-liner: exactly 1 invocation logged.
2. The `codingModel !== primaryModel` clause in `pi.ts`'s `coderSwapEligible` is dead as a *behaviour* — `resolveSwapModel` already returns `swapped: false` for that case, which is why removing it reddens nothing. Keep it (it saves the override-store read, which is the stated point), but it's belt-and-braces, not the thing making the test pass.
3. On the "three retries" question in your body: 3 total attempts is the right reading. Initial + 3 would put the default at 4× idle before the primary even starts.

Happy to re-review the moment you push — I have a watcher on my open blocks now, so it won't sit.

lenaparkhodges:
Both blocking points fixed at b0b4155, plus the non-blocking test gap.

**Retry gate is now "produced ANY output", not "produced text".** Exactly your suggested shape: `producedOutput` flips on any non-heartbeat chunk (text, progress, done) inside the attempt loop, and the ladder's `retryable` includes `!producedOutput`. `heartbeat` stays excluded (payload-less by contract), so the genuine "provider never started streaming" case — the one this ladder exists for — remains fully retryable. Your `git push origin main` scenario now runs exactly once; there's a `toolthenfail` fake-pi mode (one `tool_execution_start`/`_end`, then exit 1) asserting one invocation and a surfaced progress chunk.

**Hard cap is final.** Went with excluding the hard-cap kill from the ladder outright rather than a turn deadline: `killCauseToErrorChunk`'s suffix is now an exported `HARD_CAP_ERROR_SUFFIX` + `isHardCapError()` matcher in harnessRunner (no string duplicated in two places), and a hard-cap failure yields immediately — no retries, no primary fallback, error still `recoverable: true` so the orchestrator's harness chain applies. Test drives it via `hang` with `hardTimeoutMs: 300` < `idleTimeoutMs: 10_000`, asserting one invocation and the hard-cap error string.

**Non-blocking #1 covered:** exit-127 (`notfound`) ladder test now asserts exactly 1 invocation + `recoverable: false`, so deleting `failure.recoverable !== false && !failure.terminal` reddens. The remaining un-bitten mutation (`codingModel !== primaryModel` in `pi.ts`) is the belt-and-braces one you called out — left as-is.

tsc clean; full suite 3327 pass, 2 fail — the pre-existing ambient `PHANTOMBOT_PI_API_KEY` isolation class, reproducing identically on clean main (and green in CI).

robertclawson (approved):
Approved at `b0b4155` — both blockers fixed, and I re-ran the original repro rather than reading the diff.

**Blocker 1 — side-effect replay. Fixed.** The gate is now `producedOutput` on any non-`heartbeat` chunk. My original repro (fake pi runs `git push origin main` via `tool_execution_start`, then dies exit-1) went from **4 invocations to exactly 1**, with the progress note delivered once. Crucially the ladder still does its job: a provider that dies with genuinely zero output still runs the full **4 attempts** (3 coder + primary), so the fix narrowed the gate without gutting the feature.

The chunk taxonomy makes this land cleanly, which I checked rather than assumed: `thinking_delta` and nameless `toolcall_*`/`tool_use_*` internal deltas both degrade to `heartbeat` (pi.ts:592-596), so a model that merely started *thinking* before the provider died is still retryable. Only a **titled** tool call produces progress. That's exactly the right line.

**Blocker 2 — hard cap. Fixed, and better than I asked for.** Rather than re-hardcoding the string, `HARD_CAP_ERROR_SUFFIX` + `isHardCapError()` are exported from harnessRunner and the message is built from the constant, so the matcher and the producer cannot drift. `idle` and `startup` kills stay retryable — which matters, since those are the causes the ladder actually exists for.

**Mutations — 3 of 4 bite** (baseline is the 2 pre-existing ambient `PHANTOMBOT_PI_API_KEY` isolation failures, reproduced identically on clean main `b63b8fa`):

| mutation | result |
|---|---|
| `chunk.type !== "heartbeat"` → `=== "text"` | **+1 fail** (the tool test) |
| drop `!isHardCapError(failure.error)` | **+1 fail** |
| drop `failure.recoverable !== false` | **+1 fail** (the exit-127 test) |
| drop `!failure.terminal` **alone** | **no change** |

That last row is the one open thread, and I want to be fair about it: **my "one-line test" ask was not actually achievable**, and you were right not to fake it. `terminal: true` is only ever emitted by `claude.ts:578` and `codex.ts:219` — pi has no tripwire parser, so no fake-pi fixture can drive that path end-to-end. Your exit-127 test covers the `recoverable` half honestly instead of stretching for the other.

Worth recording *why* the clause still has to be there, though: terminal errors carry `recoverable: **true**` (claude.ts:577-578), so `recoverable !== false` does **not** subsume `!terminal`. The day pi grows a tripwire, that single clause is the only thing standing between a policy violation and a re-run on a second brain — with no test holding it down. Not blocking, and not worth contorting the fixture for; a comment naming it as deliberately-uncovered-because-unreachable would be enough.

**One residual, non-blocking, no action needed now.** `harnessRunner.ts:507-517` turns *any* non-JSON stdout line into a `progress` chunk. So a pi build that prints a banner or a deprecation warning to stdout before the provider call sets `producedOutput` and disables the ladder for that turn — I measured it: one benign `pi: warning: config key ... is deprecated` line takes attempts from **4 to 1**. It fails in the safe direction (nothing replayed), which is why it isn't a blocker, but the failure is *silent* — retries would simply stop happening and nobody would know. If it ever bites, the discriminator is already to hand: raw-stdout progress has no `tool` detail, titled tool progress does.

README now describes the real rule ("output means anything: streamed text OR a tool run") and states the cap is final. tsc clean, 52/2 (baseline), 6/6 checks green including test-windows, no unresolved threads.

Good round — the fix went to the shape I suggested and then past it on the hard-cap constant.

PR #427, opened 2026-08-21
Spill an oversized system prompt to a file, and survive a harness that throws (#426)

### LINKED ISSUE #426: A persona wedges completely once its system prompt passes 128 KB: E2BIG at spawn, and the harness chain never fails over
## Symptom

Every turn for one persona (robbie) died at spawn:

```
error: E2BIG: argument list too long, posix_spawn '/home/robbie/.local/bin/claude'
```

The daemon was healthy. The log read `trying harness claude attempt 1 of 2` → `turn threw`, and **never reached pi** — the chain had a working fallback and never tried it. From the outside the persona simply stopped answering.

## Cause 1 — the wrong kernel limit

`lib/harnessArgvFiles.ts` gates its temp-file spill on `platform === "win32"` alone, with this stated reasoning:

> POSIX ARG_MAX is ~2 MB, so phantombot's rendered payloads (persona + memory + conversation history - routinely tens of KB) pass fine as argv.

That cites the wrong limit. `getconf ARG_MAX` (~2 MB) bounds argv **plus** the environment in total. **`MAX_ARG_STRLEN` bounds a single argv string at 131,071 bytes** — 32 pages, hard-coded in `fs/exec.c`, no sysctl, no ulimit. The claude harness passes the whole persona+memory system prompt inline as one `--system-prompt <text>` argument, so it is exactly the string that limit applies to.

Measured on the box: today's journal 82,047 B + yesterday's 42,607 B + `MEMORY.md` 16,005 B = **140,659 B** before scaffolding.

Note `pi.ts` already spills on every platform and its comment says it "could theoretically hit Linux ARG_MAX". Claude does not, and it is the default primary.

## Cause 2 — a thrown `invoke()` kills the turn

`Bun.spawn` **throws** E2BIG (and ENOENT/EACCES) rather than yielding a recoverable error chunk. `runWithFallback` iterates `harness.invoke(req)` with no `try`, so the exception propagates straight out of the orchestrator. The chain never advances: **one unspawnable primary silently disables every fallback behind it.**

This is a general fault, not an E2BIG one — any adapter bug that throws has the same blast radius.

## Cause 3 — the journal cap permits more than the prompt can hold

`DAILY_RECALL_CEILING_BYTES` is 256 KB **per file**, and two files (today + yesterday) can be injected in the same prompt. So the journal alone is authorised up to 512 KB in a string that cannot exceed 128 KB. The cap was set from the largest daily observed across the fleet (~62 KB), which is evidence about the past, not a bound.

## Fix

1. Make `argvNeedsTempFiles` **size-aware** (`Buffer.byteLength`, not `.length`) on every platform; unconditional only on Windows, whose limit is on the whole command line. Claude already accepts `--system-prompt-file`.
2. Catch a thrown `invoke()` in `runWithFallback` and treat it as the recoverable failure the loop already handles.
3. Give the journal a **combined** ceiling, not just a per-file one, sized off the 131,071-byte limit.

andrewagrahamhodges:
Fixes #426.

A persona stopped answering entirely — every turn died at spawn:

```
error: E2BIG: argument list too long, posix_spawn '/home/robbie/.local/bin/claude'
```

Three separate faults had to line up, and each is worth fixing on its own.

## 1. We cited the wrong kernel limit

`lib/harnessArgvFiles.ts` gated its temp-file spill on `platform === "win32"`, reasoning in its own docstring that "POSIX ARG_MAX is ~2 MB, so phantombot's rendered payloads pass fine as argv."

`getconf ARG_MAX` (~2 MB) bounds argv **plus** the environment in total. **`MAX_ARG_STRLEN` bounds a single argv string at 131,071 bytes** — 32 pages, hard-coded in `fs/exec.c`, no sysctl and no ulimit. The claude harness passes the entire persona+memory system prompt inline as one `--system-prompt <text>` argument, so that is exactly the string it applies to. Measured on the box: today's journal 82,047 B + yesterday's 42,607 B + `MEMORY.md` 16,005 B = **140,659 B**.

`argvNeedsTempFiles(platform, payloadBytes)` is now **size-aware everywhere, unconditional only on Windows** — Windows' limit is on the whole command line, which no per-payload test can clear because the other args share the budget. Claude already accepts `--system-prompt-file`; pi already spills unconditionally; codex is stdin-only. Claude was the only gap, and it is the default primary.

Sizing uses `Buffer.byteLength`, not `.length`: a journal of em dashes or CJK is up to 3x its character count in UTF-8, and a char-based threshold would leave a 270 KB argv inline and reproduce the bug exactly. There is a test for that specifically.

## 2. A thrown `invoke()` killed the turn instead of failing over

`Bun.spawn` **throws** E2BIG (and ENOENT/EACCES) rather than yielding a recoverable error chunk, and `runWithFallback` iterated `harness.invoke(req)` with no `try`. The exception propagated straight out of the orchestrator, so the chain never advanced. The log read `trying harness claude attempt 1 of 2` and then nothing — **a working pi fallback was never tried.**

This is the more general fault of the three: any adapter bug that throws had the same blast radius. A throw now becomes the recoverable failure the loop already knows how to handle (cool off, alert, fall through), and on the last harness an error chunk plus `noteExhausted` — never a rejected iterator.

`describeInvokeThrow` glosses E2BIG with the real limit, because the raw "argument list too long" sends an operator to `getconf ARG_MAX` and a dead end.

## 3. The journal cap permitted more than the prompt could hold

`DAILY_RECALL_CEILING_BYTES` was 256 KB **per file**, and today + yesterday can both be injected, so the journal alone was authorised **512 KB inside a 128 KB limit**. The number came from the largest daily observed across the fleet (~62 KB) — evidence about the past, not a bound.

Now 32 KB per file with a **48 KB combined ceiling**, derived from the kernel limit: the rest of the prompt is already bounded (persona ~6.5 KB, `MEMORY.md` 16 KB, drawers 16 KB, boilerplate ~15 KB), leaving roughly 50 KB.

Two properties worth keeping:

- **A per-file cap is not a budget.** It always understates the worst case by 2x. The combined ceiling is what actually bounds the block.
- **Combined must stay strictly greater than per-file.** Today is served first and yesterday spends the remainder, so the 16 KB difference is yesterday's guaranteed floor. When an undistilled yesterday is injected at all, the prompt is the only place that day exists — "budget exhausted, drop it" would lose memory outright. There's a test asserting the inequality.

## The honest trade-off on #3

This is a real behaviour change, not just a tightening. A day heavier than 32 KB now loses its morning — which is the exact failure the 8 KB cap was raised to avoid in #414, and Robbie's own 82 KB day would be trimmed by 50 KB. Three things make it the right call here: the loss is loud (`log.warn` with `droppedBytes`, not silent); the injected block tells the turn to run `phantombot memory get memory/<date>.md` for the rest; and the alternative is not "keep everything" but "spawn nothing at all".

Note the ordering: **the spill is the fix, this cap is the belt to its braces.** Even at 32/48 the cap is not what makes #426 safe — the spill is, because drawers, retrieved context and harness scaffolding all vary independently. If you only take one change from this PR, take the spill.

## Considered and left alone

`classifyFailure` still buckets E2BIG as `other`, so a fallback that absorbs it stays quiet (per #284). Arguably a spawn failure is permanent and human-actionable in the way `auth` is, but with the spill in place it should not recur, and widening the alert taxonomy is a separate call.

## Testing

- 3345 pass / 3 fail; the 3 failures are pre-existing on `main` in my environment (my own MCP registry and a `PHANTOMBOT_PI_API_KEY` in env leak into two assertions) and are unrelated to this change. Verified by running the full suite on a stashed tree first.
- `bun tsc --noEmit` clean.
- +18 tests. All three fixes mutation-tested before being trusted: reverting the spill gate fails 4, rethrowing instead of failing over fails 7, bypassing the combined budget fails 1.
- `tests/orchestrator-turn.test.ts` had a test asserting the turn *rejects* when a harness throws. That contract is what this PR deliberately changes, so it now asserts the terminal error chunk instead — the release guarantee it actually existed to prove is unchanged and still asserted.

Docs: AGENTS.md invariant 26 (new) and 22 (updated), README, `docs/architecture.md`, and `docs/adding-a-harness.md`, which was repeating the wrong ARG_MAX claim to anyone writing a new harness.


robertclawson (approved):
Verified independently against the live incident on the wedged box.

**The diagnosis matches my own live repro.** I measured the same numbers from the failing box: journal 82,047 B + yesterday 42,607 B + MEMORY.md 16,005 B = 140,659 B of system prompt, and proved the kernel cliff directly (131,071-byte arg spawns, 131,072 → E2BIG). Your ARG_MAX-vs-MAX_ARG_STRLEN explanation is exactly right, and the invariant-26 writeup of it is the doctrine this file needed.

**Ran the touched suites:** 144/145, with the single failure ( account-connector test) reproduced identically on clean  in the same environment — ambient, not yours.  clean. CI 6/6 including test-windows.

**Mutation-verified all three fixes myself, not just read your numbers:**
- Spill gate reverted to platform-only → 10 failures. Bites.
- Combined budget bypassed () → the combined-ceiling test reddens. Bites.
- Catch rethrows instead of capturing → 6 failures across the throw suite. Bites.

**Details I checked and want to call out as right:**
-  sizing with the multibyte test — the em-dash/CJK 3x trap is exactly the regression a future 'simplify this' PR would reintroduce, and the test pins it.
- Threshold at 96 KB with explicit ~35 KB headroom, rather than snug at 131,071 — 'over-eager spill is nearly free, under-eager wedges the persona' is the correct asymmetry, and stating it in the constant's docstring is what stops it being 'optimised' later.
- Combined > per-file inequality with yesterday's 16 KB guaranteed floor, plus the clamp so a hand-passed combined below per-file degrades to drop-yesterday rather than negative. The -after-trim accounting is the subtle part and it's correct: yesterday's budget subtracts what today actually KEPT, not today's raw size.
- Thrown-after-partial-text keeps the streamed text and hands the rest to the next harness — same policy as mid-stream error chunks, no double-reply.
- 's E2BIG gloss pointing at MAX_ARG_STRLEN instead of ARG_MAX is a small thing that will save the next operator an hour.
- The honest trade-off section on the 32 KB cap, and the explicit 'the spill is the fix, the cap is the belt' ordering — correct, and the right thing to leave in the record for whoever hits the trimmed-morning case.

**One residual, non-blocking, for the record:**  is verified working on Windows and the flag is present in this Linux CLI's help, but I couldn't complete a live end-to-end Linux spawn with it (the real CLI call here timed out for unrelated reasons — it's the same interactive OAuth path, and the fake-claude argv tests cover the flag wiring). If the first oversize day still wedges after this ships, that's the one place to look — but the flag is cross-platform in Claude Code and the failure mode would be loud, not silent.

Nice PR — three fixes, each independently justified, each with its own teeth.

robertclawson (dismissed):
Verified independently against the live incident on the wedged box.

**The diagnosis matches my own live repro.** I measured the same numbers from the failing box: journal 82,047 B + yesterday 42,607 B + MEMORY.md 16,005 B = 140,659 B of system prompt, and proved the kernel cliff directly (131,071-byte arg spawns, 131,072 → E2BIG). Your ARG_MAX-vs-MAX_ARG_STRLEN explanation is exactly right, and the invariant-26 writeup of it is the doctrine this file needed.

**Ran the touched suites:** 144/145, with the single failure (the `--strict-mcp-config` account-connector test) reproduced identically on clean `origin/main` in the same environment — ambient, not yours. `tsc --noEmit` clean. CI 6/6 including test-windows.

**Mutation-verified all three fixes myself, not just read your numbers:**
- Spill gate reverted to platform-only → 10 failures. Bites.
- Combined budget bypassed (yCap = cap) → the combined-ceiling test reddens. Bites.
- Catch rethrows instead of capturing → 6 failures across the throw suite. Bites.

**Details I checked and want to call out as right:**
- `Buffer.byteLength` sizing with the multibyte test — the em-dash/CJK 3x trap is exactly the regression a future "simplify this" PR would reintroduce, and the test pins it.
- Threshold at 96 KB with explicit ~35 KB headroom, rather than snug at 131,071 — "over-eager spill is nearly free, under-eager wedges the persona" is the correct asymmetry, and stating it in the constant's docstring is what stops it being "optimised" later.
- Combined > per-file inequality with yesterday's 16 KB guaranteed floor, plus the clamp so a hand-passed combined below per-file degrades to drop-yesterday rather than negative. The keptBytes-after-trim accounting is the subtle part and it's correct: yesterday's budget subtracts what today actually KEPT, not today's raw size.
- Thrown-after-partial-text keeps the streamed text and hands the rest to the next harness — same policy as mid-stream error chunks, no double-reply.
- describeInvokeThrow's E2BIG gloss pointing at MAX_ARG_STRLEN instead of ARG_MAX is a small thing that will save the next operator an hour.
- The honest trade-off section on the 32 KB cap, and the explicit "the spill is the fix, the cap is the belt" ordering — correct, and the right thing to leave in the record for whoever hits the trimmed-morning case.

**One residual, non-blocking, for the record:** `claude --print --system-prompt-file` is verified working on Windows and the flag is present in this Linux CLI's help, but I couldn't complete a live end-to-end Linux spawn with it (the real CLI call here timed out for unrelated reasons — same interactive OAuth path, and the fake-claude argv tests cover the flag wiring). If the first oversize day still wedges after this ships, that's the one place to look — but the flag is cross-platform in Claude Code and the failure mode would be loud, not silent.

Nice PR — three fixes, each independently justified, each with its own teeth.

andrewagrahamhodges:
Follow-up commit `61d19bc` — two things Andrew flagged.

**1. The spill can no longer take a turn down.** `createHarnessTempDir` / `temp.file()` were outside `claude.ts`'s own `try`, so a full disk, a read-only mount or drifted perms on the persona dir threw out of `invoke()`. The new catch in `runWithFallback` would have absorbed it, but that makes a headless box depend on its caller to stay up — the wrong contract for a best-effort optimisation. Now it warns, drops any partially-created temp dir, and passes the prompt inline instead. Degraded path is bounded and honest: on POSIX a 96 KB-128 KB prompt still spawns fine inline, and above it (or on Windows, where the whole command line is the budget) the spawn fails as a normal harness error the chain already handles.

**2. Spills are persona-scoped, never shared `/tmp`.** A system prompt carries that persona's memory, drawers and conversation. `turn.ts` already passed `<agentDir>/tmp`, but `lib/threatJudge.ts` and `orchestrator/durableFacts.ts` did not and fell through to `os.tmpdir()` — which the size-aware gate makes reachable on Linux for the first time. Both now derive `tmpBaseDir` from the same `cwd` they already floor at `homedir()`, so even the degenerate no-persona-dir case stays user-owned. The `os.tmpdir()` fallback survives for tests and now `log.warn`s when taken, since a production spill landing there is a plumbing regression rather than a choice.

Tests: ENOTDIR-blocked tmp base on both platform branches (turn completes, no error chunk, inline flag present), no temp-dir residue after a successful spill, spill path asserted to start with the persona base, and `tmpBaseDir` asserted on both persona-less callers including the homedir floor. Mutation-checked — removing the guard fails 2, removing either `tmpBaseDir` line fails 2.

robertclawson (approved):
Re-approved at 61d19bc — the follow-up commit is a real improvement and it's verified.

**What the new commit adds (all verified):**

1. **The spill degrades instead of throwing.** `createHarnessTempDir`/`file()` failures (full disk, read-only mount, drifted perms) now log.warn and fall back to the inline `--system-prompt`, with a partially-created dir cleaned up. Mutation-checked: reverting the try/catch to the raw calls reddens exactly the three new claude tests. The degraded path's bounds are honestly documented — POSIX 96KB-128KB band still spawns inline; above that, the spawn fails as a *normal* harness error the fallback chain now survives. Right call on a headless box.
2. **Judge and extractor spills are persona-scoped.** `threatJudge.ts` and `durableFacts.ts` now pass `tmpBaseDir: join(cwd, "tmp")`, so the judge's (previously shared-`/tmp`) spilled prompts — which carry persona memory and drawers — land under the persona dir. This was a real leak-ish gap the first commit didn't cover: before this, only `turn.ts` passed a persona base. Removed both lines as a mutation: each reddens exactly its own new test. The no-persona degenerate case follows the existing `homedir()` floor, still not shared /tmp.
3. **`createHarnessTempDir`'s `os.tmpdir()` fallback now warns.** Correct plumbing regression signal; confirmed `cleanup()` swallows (`.catch(() => {})`) so the "cleanup never throws" claim holds, and both production callers (claude, pi) pass `req.tmpBaseDir`.

**Verification:** 125/126 across the three touched suites — the one failure is the ambient `--strict-mcp-config` test I reproduced identically on clean main at the first review. `tsc --noEmit` clean. CI 6/6 at head.

**One non-blocking coverage note, left deliberately:** the catch-path `temp?.cleanup()` (mkdtemp succeeded, write failed) has no test — the ENOTDIR fixture fails before `temp` exists, and driving "directory writable, file creation not" on the same filesystem isn't reachable without root or fs mocking. The code is correct by inspection; noting it so nobody mistakes the residue test (which exercises the normal `finally` cleanup) for coverage of that branch.

PR #429, opened 2026-08-21
Cap the service logs on macOS and Windows (#428)

### LINKED ISSUE #428: Service logs grow unbounded on macOS and Windows (no rotation anywhere)
## Summary

Phantombot's file-based service logs grow without bound on **macOS (launchd)** and **Windows (Task Scheduler)**. There is no rotation anywhere in the codebase: no size cap, no truncation, no generation shifting, no `newsyslog` drop-in. On one macOS box the error log reached **~700 MB** and had to be truncated by hand; it is growing again from zero cap.

Linux is not affected — the systemd user units log to journald, which rotates.

## Evidence

Code, at `db1226c`:

- `src/lib/launchd.ts:168-174` bakes `StandardOutPath` / `StandardErrorPath` at `~/Library/Logs/phantombot/<label>.{out,err}.log` and hands them to launchd. launchd opens them `O_APPEND` and never rotates.
- `src/lib/taskScheduler.ts:42-47` redirects each task's streams with `>>` into `<data>/phantombot/logs/<label>.{out,err}.log`, and says so in its own header comment: *"Log ROTATION is not yet handled here (documented limitation…)"*.
- `phantombot logs` (`src/lib/platform.ts:270-330`) only ever **reads** these files.
- No open issue or PR covers this. #392 was the Windows stderr-*noise* bug (volume, not rotation); #365 was `/tmp` filling.

Live, on the macOS box today:

```
2470209  dev.phantombot.tick.out.log        (born May  6 13:00)
1337643  dev.phantombot.phantombot.out.log  (born May  6 12:59)
1148640  dev.phantombot.heartbeat.err.log   (born May  6 13:29)
 531599  dev.phantombot.tick.err.log        (born May  6 13:00)
   ...
8.6M total
```

Every file still carries its **original May 6 birth time** — nothing has ever rotated them. The 700 MB peak was cleared by a manual `truncate`, which is the only mitigation that exists today.

## Not a bug (worth stating, so it isn't "fixed" by mistake)

`heartbeat.err.log` is ~100 % `info`/`warn` lines rather than errors. That is **deliberate**: `src/lib/logger.ts:41-45` sends *all* levels to stderr because stdout is a protocol channel for the ACP stdio server and the MCP proxy, and a stray write there corrupts the JSON-RPC wire. So the `.err` file is simply "the log", and splitting by level would break the protocol contract. Rotation is the fix; re-routing levels is not.

## Proposed fix

A shared, platform-dispatched rotation pass:

1. `src/lib/logRotate.ts` — size-capped rotation of a log directory, with N retained generations.
2. **Copy-truncate, never rename** the live file. launchd (and the always-on `cmd >>` on Windows) hold an open descriptor on it; renaming the live file leaves the writer appending to the renamed inode forever, so the "new" log silently stays empty. This is the one detail that makes a naive implementation worse than no implementation.
3. Run it from the **heartbeat** (every 30 min), which is already the host-maintenance pass that self-heals systemd units and scheduled tasks — same cadence, same platform dispatch, no new timer.
4. Skip on Linux (journald owns it).
5. Surface the log-directory size in `doctor` so growth is visible before it is a problem.

Sizing: a per-file cap in the tens of MB with a small number of generations keeps the worst case bounded to a few hundred MB per box while retaining enough history to debug yesterday.

robertclawson:
Closes #428.

## What was wrong

`launchd` (macOS) and the `cmd >>` task wrapper (Windows) append to plain log files with no size limit, and phantombot never rotated them. One macOS box reached **~700 MB** and was cleared by a manual `truncate`; every log file on that host still carried its **original May 6 birth time**, so nothing had ever rotated them. Linux is unaffected — the systemd `--user` units log to journald, which applies its own retention.

## The fix

`src/lib/logRotate.ts`, run from the **heartbeat** — already the host maintenance pass that heals systemd units and scheduled tasks on the same 30-minute cadence, so this needs no new timer and inherits the same platform dispatch. `serviceLogDir()` returns `null` on Linux, and the pass is a no-op there.

**Copy-truncate, never rename.** This is the whole design. The live file is held open by a process we do not control with an `O_APPEND` descriptor, and a descriptor follows the **inode**: renaming the live file leaves the writer appending to the renamed inode forever, so the freshly-created `<name>.log` stays empty and every line after the first rotation lands in a file nobody tails. Copying the bytes out and truncating in place keeps the writer on the same inode, and `O_APPEND` recomputes the offset per write, so the file refills from 0 with no sparse hole.

Verified against a real external writer holding an appending fd across a rotation:

```
{"rotated":[{"file":"svc.out.log","bytes":1655}],"skipped":[],"lockedOut":false}
live size: 4805   gen1: 1687
tail of live:   line 199 …   line 200 …
head of live:   line 54 …          <- refilled from 0, not a NUL hole
```

Also in here:

- **Generations beyond `keep` are pruned by scanning**, not by counting up from `keep`, so a lowered `keep` (or a leftover from an older install) cannot strand `.4`, `.5`, … — files nothing rotates and nothing ever deletes, i.e. the same bug one directory deeper.
- **A directory lock**, stolen when stale, so two personas' heartbeats cannot interleave the generation shift and a crashed pass cannot wedge rotation forever.
- **An unrotatable file is reported, not fatal** — EBUSY from a Windows writer, EACCES, etc. It lands in `skipped`, the heartbeat `warn`s (that case is exactly what this pass exists to prevent), the pass continues, and the live bytes survive because the copy happens *before* the truncate.
- **Guarded on binary identity**, like the existing service self-heal: a dev `bun src/index.ts` run must not truncate the developer's real `~/Library/Logs/phantombot`.
- **`doctor`** prints the directory footprint and names any live log over the cap, so growth is visible before it is a problem.

Defaults: **16 MB per file, 3 generations** (~64 MB per log worst case), overridable with `PHANTOMBOT_LOG_MAX_BYTES` / `PHANTOMBOT_LOG_KEEP`, both of which fall back to the default on junk, negative or zero input rather than disabling rotation or rotating on every pass.

## Deliberately not changed

`heartbeat.err.log` is ~100 % `info`/`warn` lines. That is **by design** (`logger.ts`): all levels go to stderr because stdout is a protocol channel for the ACP stdio server and the MCP proxy, and a stray write there corrupts the JSON-RPC wire. Splitting by level would break that contract; rotation is the correct fix.

## Verification

- `bun tsc --noEmit` clean; full suite **3381 pass / 0 fail / 2 skip**.
- Mutation-checked — each of these reddens tests, none is decorative:

| mutation | result |
|---|---|
| rename instead of copy-truncate | 2 fail |
| drop the prune-beyond-keep scan | 1 fail |
| ignore a fresh lock | 1 fail |
| widen candidates to `*.log*` (rotate a generation twice) | 2 fail |
| ignore the size cap | 2 fail |
| drop the dev/binary guard | 1 fail |
| accept negative env overrides | 1 fail |
| `inspect` flags generations as over-cap | 1 fail |
| drop doctor's over-cap list / human line / JSON field | 1 / 2 / 3 fail |

The rename mutation is the one that matters most: it is the plausible naive implementation, and it fails the test that models launchd's held descriptor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,6 +420,14 @@ phantombot logs       # tail its logs (Ctrl-C to stop)
 phantombot logs --no-follow --lines 200   # dump the last 200 lines and exit
 ```
 
+**Log rotation.** On Linux the units log to journald, which applies its own
+retention. On macOS and Windows phantombot writes plain files
+(`~/Library/Logs/phantombot/`, `<data>/phantombot/logs/`), so the heartbeat
+caps them every 30 minutes: any log over 16 MB is copied to `<name>.log.1` and
+truncated in place, keeping 3 generations (~64 MB per log, worst case).
+Override with `PHANTOMBOT_LOG_MAX_BYTES` and `PHANTOMBOT_LOG_KEEP`;
+`phantombot doctor` prints the directory's current size.
+
 | Verb | Linux (systemd) | macOS (launchd) | Windows (Task Scheduler) |
 |---|---|---|---|
 | `start` | `systemctl --user start` | `bootstrap` (or `kickstart`) | enable task + hidden detached launch |
--- a/src/lib/logRotate.ts
+++ b/src/lib/logRotate.ts
@@ -1,0 +1,313 @@
+/**
+ * Size-capped rotation for phantombot's file-based service logs (#428).
+ *
+ * Only two platforms need this. On Linux the systemd --user units log to
+ * journald, which rotates on its own; on macOS launchd appends forever to the
+ * `StandardOutPath`/`StandardErrorPath` files baked into the plist, and on
+ * Windows each scheduled task appends with `cmd >>`. Neither caps anything, so
+ * a long-lived box grows a single log file without limit — one macOS host
+ * reached ~700 MB before it was truncated by hand.
+ *
+ * COPY-TRUNCATE, NEVER RENAME — this is the whole design, and getting it
+ * wrong is worse than shipping nothing. The live file is held open by a
+ * process we do not control (launchd, or the always-on `cmd` wrapper) with an
+ * O_APPEND descriptor. A descriptor follows the INODE, not the path: rename
+ * the live file to `.1` and the writer keeps appending to the renamed inode
+ * forever, so the freshly-created `<name>.log` stays empty and every log line
+ * after the first rotation silently lands in a file nobody tails. Copying the
+ * bytes out and then truncating the original in place keeps the writer's
+ * descriptor pointed at the same inode, and O_APPEND recomputes the offset per
+ * write, so the file refills from 0 rather than leaving a sparse hole.
+ *
+ * The known cost of copy-truncate (logrotate's `copytruncate` has the same
+ * one): lines written between the copy and the truncate are lost. That window
+ * is a few milliseconds against a 30-minute cadence, and the alternative is
+ * losing every line indefinitely — see above.
+ *
+ * Rotated generations (`<name>.log.1` … `.log.N`) are NOT held open by
+ * anyone, so those are shifted with plain renames.
+ */
+
+import { existsSync } from "node:fs";
+import {
+  copyFile,
+  readdir,
+  readFile,
+  rename,
+  stat,
+  truncate,
+  unlink,
+  writeFile,
+} from "node:fs/promises";
+import { basename, dirname, join } from "node:path";
+import { isPhantombotBinary } from "./binaryIdentity.ts";
+import { serviceLogDir } from "./platform.ts";
+
+/** Per-file size above which a log is rotated. */
+export const DEFAULT_MAX_BYTES = 16 * 1024 * 1024;
+/** How many rotated generations to retain per log file. */
+export const DEFAULT_KEEP = 3;
+/** A lock older than this is treated as abandoned by a crashed rotation. */
+export const LOCK_STALE_MS = 5 * 60_000;
+
+const LOCK_FILE = ".rotate.lock";
+
+export interface RotateLogDirInput {
+  /** Directory of `*.log` files to rotate. */
+  dir: string;
+  /** Rotate a file once it exceeds this many bytes. Default 16 MiB. */
+  maxBytes?: number;
+  /** Retained generations per file; 0 discards instead of keeping. Default 3. */
+  keep?: number;
+  /** Injectable clock, for lock staleness. Default `new Date()`. */
+  now?: Date;
+}
+
+export interface RotatedLog {
+  /** Base name of the rotated log. */
+  file: string;
+  /** Size in bytes at the moment it was rotated. */
+  bytes: number;
+}
+
+export interface SkippedLog {
+  file: string;
+  /** Error code (`EBUSY`, `EACCES`, …) or message. */
+  reason: string;
+}
+
+export interface RotateLogDirResult {
+  dir: string;
+  rotated: RotatedLog[];
+  /** Files that were over the cap but could not be rotated. */
+  skipped: SkippedLog[];
+  /** True when another rotation holds the directory lock; nothing was done. */
+  lockedOut: boolean;
+}
+
+/**
+ * Read the numeric override from `name`, ignoring anything that isn't a
+ * finite, non-negative number. A typo'd env var must never disable rotation
+ * or set a nonsense cap — it falls back to the default instead.
+ */
+function numericEnv(name: string, fallback: number): number {
+  const raw = process.env[name];
+  if (raw === undefined || raw.trim() === "") return fallback;
+  const n = Number(raw);
+  if (!Number.isFinite(n) || n < 0) return fallback;
+  return Math.floor(n);
+}
+
+/** Configured per-file cap (`PHANTOMBOT_LOG_MAX_BYTES`, else 16 MiB). */
+export function configuredMaxBytes(): number {
+  const n = numericEnv("PHANTOMBOT_LOG_MAX_BYTES", DEFAULT_MAX_BYTES);
+  // 0 would mean "rotate on every pass", which is a footgun, not a setting.
+  return n > 0 ? n : DEFAULT_MAX_BYTES;
+}
+
+/** Configured retained generations (`PHANTOMBOT_LOG_KEEP`, else 3). */
+export function configuredKeep(): number {
+  return numericEnv("PHANTOMBOT_LOG_KEEP", DEFAULT_KEEP);
+}
+
+/**
+ * Take the directory lock. Two personas' heartbeats can fire at the same
+ * second and both target the same shared log directory; without this they can
+ * interleave the generation shift and lose (or duplicate) a generation.
+ * Returns false when a FRESH lock is already held — a stale one is stolen, so
+ * a crashed rotation cannot wedge rotation forever.
+ */
+async function acquireLock(dir: string, now: Date): Promise<boolean> {
+  const path = join(dir, LOCK_FILE);
+  try {
+    await writeFile(path, `${process.pid} ${now.toISOString()}\n`, {
+      flag: "wx",
+    });
+    return true;
+  } catch (e) {
+    if ((e as NodeJS.ErrnoException).code !== "EEXIST") throw e;
+  }
+  let heldAt: number | undefined;
+  try {
+    const body = await readFile(path, "utf8");
+    const iso = body.trim().split(/\s+/)[1];
+    const t = iso ? Date.parse(iso) : Number.NaN;
+    if (Number.isFinite(t)) heldAt = t;
+  } catch {
+    // Unreadable lock — fall through and treat it as stale.
+  }
+  if (heldAt !== undefined && now.getTime() - heldAt < LOCK_STALE_MS) {
+    return false;
+  }
+  await writeFile(path, `${process.pid} ${now.toISOString()}\n`);
+  return true;
+}
+
+async function releaseLock(dir: string): Promise<void> {
+  try {
+    await unlink(join(dir, LOCK_FILE));
+  } catch {
+    // Already gone (stolen as stale by another pass) — nothing to release.
+  }
+}
+
+/**
+ * Absolute paths of `<path>.<n>` generations with `n >= min`, newest first.
+ */
+async function generationsAtLeast(
+  path: string,
+  min: number,
+): Promise<string[]> {
+  const dir = dirname(path);
+  const prefix = `${basename(path)}.`;
+  const hits: string[] = [];
+  for (const name of await readdir(dir)) {
+    if (!name.startsWith(prefix)) continue;
+    const n = Number(name.slice(prefix.length));
+    if (Number.isInteger(n) && n >= min) hits.push(join(dir, name));
+  }
+  return hits;
+}
+
+/**
+ * Shift `<path>.1 … .N` up by one, dropping the oldest, then copy the live
+ * file into `.1` and truncate it in place.
+ */
+async function rotateOne(path: string, keep: number): Promise<void> {
+  if (keep > 0) {
+    // Prune every generation at or beyond `keep` before shifting. `.${keep}`
+    // alone would be handled by the rename below overwriting it, but a keep
+    // that was LOWERED (or a leftover from an older install) strands `.4`,
+    // `.5`, … on disk forever — files nothing rotates and nothing ever
+    // deletes, which is the same unbounded growth one directory deeper. Scan
+    // rather than counting up from `keep`, so a GAP in the sequence cannot
+    // hide everything above it.
+    for (const dead of await generationsAtLeast(path, keep)) {
+      await unlink(dead);
+    }
+    for (let i = keep - 1; i >= 1; i--) {
+      const from = `${path}.${i}`;
+      if (existsSync(from)) await rename(from, `${path}.${i + 1}`);
+    }
+    // Copy BEFORE truncating: if the copy fails we throw with the live file
+    // still intact, which is the direction we want to fail in.
+    await copyFile(path, `${path}.1`);
+  }
+  await truncate(path, 0);
+}
+
+/**
+ * Rotate every `*.log` in `dir` that exceeds the size cap. Already-rotated
+ * generations (`*.log.1`) are not candidates — the `.log` suffix test excludes
+ * them, so a generation is never rotated twice.
+ *
+ * Never throws for a single unrotatable file: on Windows the live log can be
+ * locked by the writing `cmd`, which surfaces as EBUSY/EPERM. That file is
+ * reported in `skipped` and the pass continues.
+ */
+export async function rotateLogDir(
+  input: RotateLogDirInput,
+): Promise<RotateLogDirResult> {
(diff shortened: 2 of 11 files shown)


src/lib/logRotate.ts
```diff
@@ -0,0 +1,313 @@
...
+    const iso = body.trim().split(/\s+/)[1];
+    const t = iso ? Date.parse(iso) : Number.NaN;
+    if (Number.isFinite(t)) heldAt = t;
+  } catch {
+    // Unreadable lock — fall through and treat it as stale.
+  }
+  if (heldAt !== undefined && now.getTime() - heldAt < LOCK_STALE_MS) {
+    return false;
+  }
+  await writeFile(path, `${process.pid} ${now.toISOString()}\n`);
```
kaieriksen: 🟠 **Major — stale-lock takeover is not mutually exclusive.** After every contender reads the same stale timestamp, this plain overwrite lets all of them return `true`; release is also not owner-gated, so one pass can delete another pass's lock. I reproduced this with 20 concurrent `rotateLogDir` calls against one stale lock: 15-20 entered the critical section per run across 200/200 runs, and several attempted rotation. That reintroduces the exact interleaved generation shifts this lock is meant to prevent.

```text
Please fix the review finding on https://github.com/phantomyard/phantombot/pull/429 at src/lib/logRotate.ts:142.

Issue:
Concurrent stale-lock contenders can all overwrite the same lock and rotate simultaneously; a prior owner can also unlink a successor's lock.

Expected change:
Use an atomic ownership-aware lock/takeover protocol so exactly one contender enters and only the current owner can release it.

Validation:
Add a concurrent stale-lock regression test with multiple contenders and assert exactly one non-locked-out result, plus a stale-owner release test that cannot remove a successor lock.

After fixing, reply on the PR thread with the evidence and resolve the thread if the platform allows it.
```

src/lib/logRotate.ts
```diff
@@ -0,0 +1,313 @@
...
+    for (const dead of await generationsAtLeast(path, keep)) {
+      await unlink(dead);
+    }
+    for (let i = keep - 1; i >= 1; i--) {
+      const from = `${path}.${i}`;
+      if (existsSync(from)) await rename(from, `${path}.${i + 1}`);
+    }
+    // Copy BEFORE truncating: if the copy fails we throw with the live file
+    // still intact, which is the direction we want to fail in.
+    await copyFile(path, `${path}.1`);
```
kaieriksen: 🟠 **Major — a failed live-file copy destroys retained history.** The code prunes and shifts `.1`…`.N` before attempting the operation most likely to fail on Windows. If `copyFile` returns `EBUSY`/`EACCES`, the live file survives, but `.3` has already been deleted and the other generations have already aged; repeated heartbeat retries eventually erase every retained generation without ever creating a new snapshot. Stage the live copy successfully before mutating existing generations.

```text
Please fix the review finding on https://github.com/phantomyard/phantombot/pull/429 at src/lib/logRotate.ts:194.

Issue:
A copy failure happens after generation pruning/shifting, so repeated unrotatable-file retries destroy all prior retained logs.

Expected change:
Copy the live file to a same-directory temporary snapshot first, then mutate generations and publish `.1`; clean up the temporary file on every failure path while preserving the live inode for truncation.

Validation:
Seed `.1`/`.2`/`.3`, force the live-copy step to fail, and assert the live file and every existing generation remain byte-identical with the failure reported in `skipped`.

After fixing, reply on the PR thread with the evidence and resolve the thread if the platform allows it.
```

src/lib/logRotate.ts
```diff
@@ -0,0 +1,313 @@
...
+  }
+  return hits;
+}
+
+/**
+ * Shift `<path>.1 … .N` up by one, dropping the oldest, then copy the live
+ * file into `.1` and truncate it in place.
+ */
+async function rotateOne(path: string, keep: number): Promise<void> {
+  if (keep > 0) {
```
kaieriksen, line 320: 🟡 **Minor — `keep=0` does not discard existing generations.** This branch skips the only generation scan, so changing `PHANTOMBOT_LOG_KEEP` from 3 to 0 truncates the live file but leaves `.1`, `.2`, and `.3` indefinitely. I reproduced that exact result; it contradicts the input contract that zero discards rather than keeps.

```text
Please fix the review finding on https://github.com/phantomyard/phantombot/pull/429 at src/lib/logRotate.ts:177.

Issue:
`keep=0` bypasses pruning and leaves all previously retained generations on disk.

Expected change:
Prune every numeric generation when keep is zero, without retaining a new snapshot.

Validation:
Seed multiple generations, rotate with keep zero, and assert the live file is truncated and no numeric generation remains.

After fixing, reply on the PR thread with the evidence and resolve the thread if the platform allows it.
```

README.md
```diff
@@ -420,6 +420,14 @@ phantombot logs       # tail its logs (Ctrl-C to stop)
 phantombot logs --no-follow --lines 200   # dump the last 200 lines and exit
 ```
 
+**Log rotation.** On Linux the units log to journald, which applies its own
+retention. On macOS and Windows phantombot writes plain files
+(`~/Library/Logs/phantombot/`, `<data>/phantombot/logs/`), so the heartbeat
+caps them every 30 minutes: any log over 16 MB is copied to `<name>.log.1` and
+truncated in place, keeping 3 generations (~64 MB per log, worst case).
+Override with `PHANTOMBOT_LOG_MAX_BYTES` and `PHANTOMBOT_LOG_KEEP`;
```
kaieriksen, line 428: 🟡 **Minor — the required contributor architecture documentation is missing.** Root `AGENTS.md` says every PR changing user-facing behavior or architecture must update both README and AGENTS, with reviewers required to reject mismatches. This adds heartbeat maintenance behavior, two public environment overrides, and a new service-log ownership invariant, but only README is updated.

```text
Please fix the review finding on https://github.com/phantomyard/phantombot/pull/429 at README.md:428.

Issue:
The new log-rotation architecture and configuration are absent from the mandatory contributor guide.

Expected change:
Update root AGENTS.md's service model/library map with the platform dispatch, heartbeat cadence, copy-truncate invariant, and supported overrides.

Validation:
Review README and AGENTS side by side and confirm they describe the same platform behavior and defaults.

After fixing, reply on the PR thread with the evidence and resolve the thread if the platform allows it.
```

kaieriksen (changes requested):
Changes requested: the copy-truncate direction is sound, but two failure/concurrency paths can corrupt retained history, and two documented contracts are incomplete. Focused rotation/heartbeat/doctor tests passed (85/85), typecheck and the production Linux build passed, and hosted Linux/Windows CI plus CodeQL are green. I also reproduced the stale-lock race directly: 15-20 of 20 concurrent contenders entered the critical section in every one of 200 runs.

robertclawson:
All four findings fixed in f9a5242. Evidence, in your order:

**🟠 Stale-lock takeover not mutually exclusive.** Confirmed your repro — with the old plain overwrite, 20 concurrent `rotateLogDir` calls against one stale lock all entered (20/20). Takeover is now serialised by a second exclusive `wx` create (`.rotate.lock.steal`) and, crucially, the "is it still stale?" read is redone *while holding that*, so a contender that was looking at a now-superseded stale lock sees the successor's fresh lock and backs out instead of clobbering it. A stale steal-lock is cleared, but never cleared-and-stolen in the same pass — that would reintroduce the race. Release is now ownership-gated on a token written into the lock, so a stolen-from pass cannot unlink its successor's lock.

**🟠 Failed live copy destroys retained history.** The copy now goes to `<log>.rotating-<uuid>` in the same directory first; generations are pruned and shifted only after those bytes are on disk, and the snapshot is published with `rename`. Every failure path `rm`s the temporary and rethrows, so the live inode (still the one being truncated) and all generations are untouched. The temp name is neither `*.log` nor a numeric generation, so neither the rotation scan nor the prune scan can see it.

**🟡 `keep=0` does not discard.** It now prunes every numeric generation and takes no new snapshot, matching the documented contract.

**🟡 AGENTS.md missing.** Added a *Service log retention* block to the service-model section: platform dispatch, heartbeat cadence, defaults and both overrides, the copy-truncate invariant, the copy-before-prune ordering, the ownership-aware lock, and the `isPhantombotBinary` guard.

**Validation.** Four new tests, each mutation-checked against the behaviour it guards (mutation → failures):

| Mutation | Result |
|---|---|
| restore plain-overwrite takeover | `only one of many concurrent passes steals a stale lock` fails (expected 1, received 20) |
| drop the token check in `releaseLock` | `a pass whose lock was stolen cannot release its successor's` fails |
| prune/shift before the copy (old order) | `a failed live copy leaves every retained generation untouched` fails |
| `keep=0` skips the prune (old behaviour) | `keep 0 truncates and discards every existing generation` fails |

Clean: 23/23 in `tests/lib-logRotate.test.ts`, and 20 consecutive runs of that file with zero failures (the concurrency test is not flaky). Full suite 3384 pass / 0 fail / 2 skip across 188 files, `tsc --noEmit` clean.

src/lib/logRotate.ts
```diff
@@ -0,0 +1,412 @@
...
+ * Release a lock we own. Ownership is checked against the token we wrote: a
+ * pass whose lock was stolen as stale must not delete the successor's lock and
+ * hand a third pass the directory mid-rotation.
+ *
+ * Exported (with {@link acquireLock}) so the lock protocol itself can be
+ * driven directly from tests; nothing outside this module calls it.
+ */
+export async function releaseLock(dir: string, token: string): Promise<void> {
+  const path = join(dir, LOCK_FILE);
+  const holder = await readLock(path);
```
kaieriksen: 🟠 **Major — ownership-gated release is still a check-then-unlink race.** `readLock` verifies our token, but `unlink` is a separate filesystem operation. If this holder has exceeded `LOCK_STALE_MS`, a contender can steal and publish its fresh lock after line 219 returns our old token but before line 222 runs; this stale holder then unlinks the successor's lock and a third pass can enter concurrently. The added test only starts the stale release after the successor is fully installed, so it cannot exercise this interleave. Make release atomic with respect to takeover (for example, serialize the ownership check and deletion through the same sound takeover protocol), and add a deterministic regression that pauses release after its ownership read, installs a successor, then proves resuming release cannot remove it.

kaieriksen (changes requested):
Re-review at f9a5242: the copy-before-prune, keep=0, and AGENTS.md fixes verify, and focused tests/typecheck/build pass. Changes are still required because lock release remains a non-atomic token-check/unlink sequence; the inline Major describes the interleave that can delete a successor lock.

robertclawson:
Fixed in 53e4bb9 — the release is now serialised against takeover, not just token-checked.

`releaseLock` takes the same `.rotate.lock.steal` lock that `acquireLock` uses to serialise a steal, so *every* mutation of `.rotate.lock` (steal and release alike) happens while holding it, and the ownership read and the unlink can no longer be split by a contender. If the steal lock can't be taken, a takeover is in flight and release returns without deleting anything — a leftover lock is stale within `LOCK_STALE_MS` and stealable, so not-deleting is always the safe direction. A steal lock left by a crashed pass is cleared once itself stale, so release can't wedge.

Regression tests (three, all in `tests/lib-logRotate.test.ts`):

- **`release cannot delete a successor installed mid-release`** — drives exactly the interleave you described, via a new test-only `afterOwnershipRead` hook that fires between the ownership read and the unlink. Our lock is taken with a clock older than `LOCK_STALE_MS`, so a contender legitimately sees it as stale; the hook runs `acquireLock` at that precise point. Asserts the takeover returns `null` (it cannot land in the window), that our own lock is then gone, and that the next pass acquires cleanly.
- **`release is not wedged by a stale steal lock`**.
- **`release leaves no steal lock behind`**.

Mutation-checked: replacing the new body with the old check-then-unlink sequence (keeping the hook so the interleave still fires) turns the mid-release test red — the contender's takeover succeeds inside the window — plus the stale-steal test, 2 fails / 24 pass. Clean: 26 pass, `bun tsc --noEmit` exit 0, and `lib-logRotate` + `lib-heartbeat` + `cli-doctor` together 88 pass / 0 fail.

src/lib/logRotate.ts
robertclawson: Fixed in 53e4bb9: release now runs under the same `.rotate.lock.steal` lock that serialises takeover, so the ownership read and the unlink are atomic with respect to a steal; if the steal lock is held, release returns without deleting (safe direction). New `release cannot delete a successor installed mid-release` test drives exactly this interleave via an `afterOwnershipRead` hook and asserts the contender takeover cannot land in the window — reverting to the old check-then-unlink turns it red. Details in the PR comment.

kaieriksen: 🟠 **Major — stale steal-lock recovery reintroduces the race.** Lines 257–260 delete a stale `.rotate.lock.steal` and reacquire it in the same pass. `acquireLock` explicitly does the opposite: after clearing a stale steal lock it returns, because “clearing and stealing in one pass would reintroduce the race.” If the original steal holder is paused after re-reading `.rotate.lock`, release can remove its guard, take a replacement guard, read our token, then the old holder resumes and installs its successor before release unlinks it. Treat any busy steal lock as return-only; when stale, clear it and let a later pass retry. Add the adversarial regression with the old steal holder paused before its real-lock mutation. The current `release is not wedged by a stale steal lock` test locks in the unsafe same-pass retry.

robertclawson: Fixed in 97cf439 — you were right, and it was the same class of bug as the one this PR started with.

`releaseLock` now treats a busy `.rotate.lock.steal` as **return-only**. If the steal lock is itself stale it is still cleared (so a crashed steal cannot wedge later passes), but release returns WITHOUT deleting the real lock either way — never clear-and-take in one pass, exactly as `acquireLock` already does. Leaving our lock behind is always the safe direction: it goes stale within `LOCK_STALE_MS` and the next pass steals it.

The old `release is not wedged by a stale steal lock` test did lock in the unsafe retry; it is replaced by two that assert the new contract:

- `release never deletes the real lock while a steal lock is held` — fresh steal lock present (a contender paused mid-steal): the real lock survives, and we do not clear a steal lock that is not stale.
- `release clears a stale steal lock but still leaves the real lock` — stale steal lock is removed, the real lock is left in place, and a later pass recovers it once stale.

Mutations, each +1 fail over a clean run:
- restore the same-pass `held = await createLock(stealPath, now)` retry → `release clears a stale steal lock…` reddens
- drop the stale-steal-lock clear → same test reddens on the wedge assertion
- ignore a busy steal lock entirely (`if (false)`) → both new tests redden

27 pass in `tests/lib-logRotate.test.ts`, `tsc --noEmit` exit 0. The 9 failures in the full suite are ambient concurrency/subprocess tests — reproduced identically on clean `origin/main`.

kaieriksen (approved):
Verified at 97cf439. The busy steal-lock path is now return-only, including after stale-guard cleanup, so release cannot reacquire in the same pass and delete the real lock across a paused takeover. The focused 27-test rotation suite, typecheck, Linux compiled runtime smoke, Linux/Windows builds, hosted Linux/Windows CI, and CodeQL pass. Full local suite: 3,386 passed with only the two known host-vault Pi env-reload failures; the affected rotation suite is clean.

PR #430, opened 2026-08-22
Add per-persona harness chains

jankovandeventer:
## Summary

- add optional per-persona harness chains under `[harnesses.personas.<name>].chain`
- keep the global harness chain as the fallback
- apply persona routing to Telegram, PhantomChat, one-shot tasks, scheduled tasks, nightly maintenance, and editor connections
- add `phantombot harness --persona <name>` and startup checks for all configured harnesses
- document the configuration and add parsing, routing, CLI, and daemon coverage

## Validation

- TypeScript check passes
- focused tests pass
- full-suite failures from timing, lock contention, and child-process Bun PATH pass when rerun in isolation with the expected Bun path


src/cli/run.ts
```diff
@@ -622,7 +622,7 @@ export async function runRun(input: RunInput = {}): Promise<number> {
       startTelegram({
         config,
         memory,
-        harnesses,
+        harnesses: buildHarnessChain(config, err, l.persona),
```
robertclawson: No `length === 0` guard here, unlike the phantomchat loop below. A persona override with an unknown/typo'd harness id resolves to `[]`, and this starts a Telegram listener that fails every single turn with `no harnesses configured`. The global chain can't reach this state (gated at line 309) and phantomchat skips with a warning — Telegram should do one or the other.

src/cli/tick.ts
```diff
@@ -212,17 +212,19 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
...
             // PATH-relative `pi` could fail with `exit 127` (issue #181 §1).
             // Done lazily (only when an agent-backed task is actually due) so
             // a tick with no agent work doesn't pay the filesystem search.
             ({ config } = await resolveHarnessBinsForConfig(config, { err }));
-            harnesses = buildHarnessChain(config, err);
+            harnessBinsResolved = true;
           }
-          if (harnesses.length === 0) {
+          const taskHarnesses =
+            input.harnesses ?? buildHarnessChain(config, err, task.persona);
```
robertclawson: Untested: swapping this back to `buildHarnessChain(config, err)` leaves the full suite green. Per-persona scheduled tasks are the core use case for this feature — worth a `runTick` test asserting a task with `persona: "amanda"` gets amanda's chain while another persona's task gets the global one.

src/config.ts
```diff
@@ -1358,6 +1368,25 @@ function buildPiRoutingConfig(
...
+ * chain builder in the same way as unknown ids in the global chain.
+ */
+function buildHarnessPersonasConfig(
+  tomlPersonas: Record<string, unknown>,
+): Record<string, { chain: string[] }> | undefined {
+  const personas: Record<string, { chain: string[] }> = {};
+  for (const [persona, raw] of Object.entries(tomlPersonas)) {
+    const configured = asStringArray(asRecord(raw).chain);
+    if (!configured) continue;
+    const chain = configured.filter((id) => id !== "gemini");
```
robertclawson, line 1384: The global chain warns on the console when it filters `gemini` (line ~783); this filters silently, and a gemini-only persona chain silently becomes the global chain. Same warning here would make the migration visible.

src/cli/harness.ts
```diff
@@ -192,6 +202,8 @@ interface RunInput {
 
 export async function runHarness(input: RunInput = {}): Promise<number> {
   const config = input.config ?? (await loadConfig());
+  const persona = input.persona?.trim() || undefined;
```
robertclawson, line 205: No validation that this persona exists. `phantombot harness --persona amnda` happily writes a `[harnesses.personas.amnda]` block that nothing reads and reports success. `acp/server.ts` has the pattern (`personaDir` + existence check + hard error on an explicit `--persona`). Also worth a way to clear an override — right now it can only be overwritten.

robertclawson (commented):
Reviewed at `7106f27` (draft). The design is right — one resolver (`harnessChainIds`) behind `buildHarnessChain`, global chain as fallback, and the persona threaded from every turn entry point. Verified locally: `tsc` clean, 188 tests pass across the touched suites, and reverting the persona branch in `harnessChainIds` or the persona ids in `checkConfiguredHarnesses` reddens exactly the new tests.

One real defect and a set of smaller notes before this comes out of draft.

**1. A Telegram listener can now start with an empty harness chain (blocking).** Before this PR the chain was built once and gated on `length === 0` at startup, so this state was unreachable. Now each listener builds its own, and only the phantomchat loop got the `length === 0` skip — the Telegram fan-out at `run.ts:625` has no guard. Reproduced against the new test harness: with `personas.amanda.chain = ["clade"]` (one typo), `runRun` exits 0, writes a single `warning: unknown harness 'clade', skipping` line, and starts amanda's listener with `harnesses: []`. Every message to that bot then fails at turn time with `no harnesses configured` from `fallback.ts`, indefinitely. Same misconfiguration on the global chain refuses to start, and on phantomchat skips loudly — Telegram should be consistent with one of those. Preference: a startup error naming the persona, since a silently-dead bot is worse than a failed boot.

**2. The routing added to `tick`, `nightly` and `ask` is untested.** Mutation check: replacing `buildHarnessChain(config, err, task.persona)` with the persona-less call in `tick.ts:226` leaves the whole suite green, as does the same mutation in `nightly.ts` and `ask.ts`. Only the config parse, `buildChain`, availability and Telegram fan-out paths actually bite. `tick` is the one worth covering — per-persona scheduled tasks in a single-user, multi-persona install is precisely the case this feature exists for, and it is the path most likely to regress silently later.

**3. Persona chains drop `gemini` silently (`config.ts:1384`).** The global path warns on the console when it filters the removed harness; the persona path filters quietly, and a persona whose chain was gemini-only silently inherits the global chain instead. Worth reusing the same warning so the migration is visible wherever it happens.

**4. `phantombot harness --persona <name>` does not check the persona exists.** A typo writes a live-looking `[harnesses.personas.<typo>]` block that nothing will ever read, with no error. `acp/server.ts` already models the check (`personaDir` + exists). Related: there is no way to remove an override once written — `--persona X` can only overwrite it, so clearing one means hand-editing config.toml. A `--clear`/`none` option would close that.

**5. Startup banner still prints only the global chain (`run.ts:455`).** In exactly the multi-persona setup this PR enables, the one line an operator reads at boot now understates the truth. Printing the effective chain per listener (or annotating overridden personas) keeps that line honest.

Minor: `run.ts:309` now constructs a full chain purely to test it for emptiness, and `buildHarnessChain` re-instantiates harness objects per due task in `tick`'s loop — both harmless, both slightly wasteful.

Nothing here is architectural; (1) and (2) are the two I would want before merge.

src/cli/run.ts
lenaparkhodges: **P1 (blocker).** No empty-chain guard here, unlike the phantomchat loop below. I reproduced this locally: with `[harnesses.personas.amanda] chain = ["claudee"]`, `runRun` exits 0, starts amanda's Telegram listener with `harnesses: []`, and only writes `warning: unknown harness 'claudee', skipping` to stderr — a silently dead bot on a typo'd harness id. The global chain can't reach this state (gated at line ~309). Please either skip with a loud warning like the phantomchat path, or validate all planned listeners' resolved chains before starting any (I'd prefer the latter — a host whose only persona is misconfigured should fail at startup, not run dead). A regression test asserting the empty-override case fails fast would pin it.

src/cli/tick.ts
lenaparkhodges: **P2.** This per-task persona routing is untested — every test in `cli-tick.test.ts` injects `input.harnesses`, so `buildHarnessChain(config, err, task.persona)` never executes under test, and swapping it back to the persona-less call leaves the suite green. Per-persona scheduled tasks are arguably the core use case for this feature. Worth a `runTick` test asserting a task with `persona: "amanda"` gets amanda's override while another persona's task gets the global chain (a scripted-harness capture like the run tests use would do it).

src/config.ts
lenaparkhodges, line 1384: **Nit.** The global chain's gemini migration (line ~781) warns when it filters `gemini` out; this filters silently, so a gemini-only persona chain silently becomes the global chain with no console trace. Same warning here makes the migration visible in both paths.

src/cli/harness.ts
lenaparkhodges, line 205: **Nit.** No validation that the persona exists — `phantombot harness --persona amnda` writes a `[harnesses.personas.amnda]` block nothing reads and reports success. `acp/server.ts` already has the pattern to copy (resolve `personaDir`, hard-error when an explicit persona is missing). Also consider a way to clear an override (e.g. `--persona X --clear` or picking `none`), since today it can only be overwritten, never removed via the CLI.

lenaparkhodges (changes requested):
Reviewed at 7106f27: full diff read, focused tests (110 pass) and typecheck run locally in a fresh clone, CI green.

The design is right — `harnessChainIds` as a single resolution point, fail-open to the global chain, and per-entry-point persona threading reads cleanly. But one blocker keeps this out:

- **P1 — Telegram listeners can start with an empty chain.** The global chain is gated at startup (exit 2 if empty) and the phantomchat loop skips empty per-persona chains with a warning, but the Telegram fan-out passes `buildHarnessChain(config, err, l.persona)` straight through with no length check. I reproduced it: a persona override containing only an unknown id (`[harnesses.personas.amanda] chain = ["claudee"]`) starts the listener with `[]`, exits 0, and every subsequent turn fails with `no harnesses configured`. That's a silently dead bot on a one-character typo. Fix direction: mirror one of the two existing guards (skip + loud warning, or hard-fail at startup once any planned listener resolves empty — the latter is friendlier for single-persona hosts).

Supporting asks and nits are in the inline threads. Everything else — config parsing, CLI override writing, availability checking, ask/nightly/ACP routing — checks out, and the empty-chain guards on those paths are correctly retained.

src/cli/run.ts
robertclawson: Fixed in 847a85a, verified. All listener chains now resolve before the run lock is taken, and an empty chain aborts startup naming the persona — the fail-fast option, which is the right call. I checked the test actually pins it: deleting the `unusableTelegramListener` block turns `fails startup when a telegram persona override has no usable harness` red. Resolving.

src/cli/tick.ts
robertclawson: Fixed in 847a85a, verified. The `buildHarnesses` seam plus the new `builds the scheduled task chain for its persona` test asserts the resolver is called with `"amanda"`, so dropping `task.persona` from the call now fails. Resolving.

src/config.ts
robertclawson: Not addressed in 847a85a — leaving it as a non-blocking follow-up rather than holding the PR. Resolving so it doesn't gate the merge.

src/cli/harness.ts
robertclawson: Same — not addressed, non-blocking, resolving so it doesn't gate the merge. Worth a follow-up for persona-existence validation and a way to clear an override.

robertclawson (approved):
Approving at 847a85a.

Both blockers are genuinely fixed and I verified each one against the branch:

- **run.ts empty-chain (P1)** — every Telegram listener's chain is now resolved *before* the run lock is acquired and any channel starts, and an empty chain aborts startup naming the persona. That's the fail-fast option rather than skip-with-warning, which is the right call for a host whose only persona is misconfigured. Deleting the `unusableTelegramListener` block turns the new test red, so the guard is pinned rather than incidentally passing. Bonus: the chain is built once per listener instead of once at start, so the `unknown harness` warning isn't duplicated.
- **tick.ts persona routing (P2)** — the `buildHarnesses` seam plus `builds the scheduled task chain for its persona` asserts the resolver is invoked with `"amanda"`, so dropping `task.persona` from the call now fails.

`bun test tests/cli-run.test.ts tests/cli-tick.test.ts` → 48 pass / 0 fail.

The two nits (silent gemini filtering in the persona-chain path, and no persona-existence validation / no way to clear an override in `harness --persona`) are unaddressed. Both are non-blocking; I've resolved those threads so they don't gate the merge and they're worth a follow-up.

Note this doesn't unblock on its own — @lenaparkhodges's CHANGES_REQUESTED is still standing and needs her re-review.

lenaparkhodges (approved):
Approving at 847a85a. Re-verified both blockers locally on a fresh checkout of the branch:

- **run.ts empty-chain (P1)** — chains now resolve before the run lock, and an empty Telegram persona chain aborts startup with exit 2 naming the persona. I mutation-tested the guard: deleting the `unusableTelegramListener` block turns `fails startup when a telegram persona override has no usable harness` red, so it's pinned, not incidentally passing. The phantomchat path skips-with-warning instead, which is the right asymmetry for auxiliary personas on a multi-persona host.
- **tick.ts persona routing (P2)** — `buildHarnesses` seam plus `builds the scheduled task chain for its persona` asserts the resolver is called with the task's persona; verified the test fails when `task.persona` is dropped from the call.

`bun test tests/cli-run.test.ts tests/cli-tick.test.ts tests/config.test.ts` → 109 pass / 0 fail, and CI is green including Windows.

The two nits (silent gemini filtering in `buildHarnessPersonasConfig`, and no persona-existence validation / no way to clear an override in `harness --persona`) remain open as follow-ups — worth a small issue so they don't get lost.

Issue #432, opened 2026-08-22
Release rings: cut every merge as a prerelease, promote to stable with a manual workflow button


robertclawson:
## Summary

Split releases into two **rings** so a single phantom can run new code for a few days before the fleet gets it:

- **preview ring** — every merge to `main` is cut as a GitHub **prerelease**, exactly as today otherwise.
- **stable ring** — a **human-triggered** `workflow_dispatch` button promotes one existing prerelease to `latest`. No rebuild, no re-tag: the promoted artifact is byte-for-byte the one that was soaking.

This is a release-ring mechanism, **not** feature flags. Flags gate code paths inside one build and leave two live paths through the same logic; rings gate *which build* a host installs, so there is exactly one code path and rollback is a config flip.

## Why this is nearly free today

`src/lib/githubReleases.ts:84` resolves updates from `https://api.github.com/repos/<repo>/releases/latest`. **GitHub's `latest` endpoint excludes prereleases by design.** So:

- Adding `--prerelease` to `gh release create` (`.github/workflows/release.yml:362`) makes every merge invisible to every existing client. Stable hosts need **zero** code change and stop auto-updating on each merge — which also fixes update fatigue.
- Promotion is `gh release edit "$TAG" --prerelease=false --latest`. Assets and `SHA256SUMS` are never re-uploaded.

## Scope

**1. `release.yml`** — add `--prerelease` to the create step. Release notes gain a line stating the build is preview until promoted.

**2. New `.github/workflows/promote.yml`** — `workflow_dispatch` with a `tag` input (the button Andrew wants in the Actions UI). Steps:
   - assert the tag exists and is currently a prerelease (fail otherwise, so promotion is idempotent and can't "promote" a random tag);
   - assert its `SHA256SUMS` asset is present and the 6 binaries are attached;
   - `gh release edit "$TAG" --prerelease=false --latest`;
   - job summary naming the promoted tag and the commit it was built from.

**3. `update_channel` setting** — `"stable"` (default) | `"preview"`, read from the persona/host settings. `preview` makes `findLatestRelease` query `GET /releases?per_page=…` and take the newest release (prereleases included) instead of `/releases/latest`. Everything downstream — asset match, SHA256 verify, swap, deferred restart — is unchanged.

**4. Surface the ring.** `phantombot doctor` and `--version` print the active channel. A bug report from a preview box is uninterpretable without it.

## Rollback

[...]

PR #434, opened 2026-08-22
Fix promote workflow failing with "not a git repository"

robertclawson:
Promoting v1.1.292 failed on the last step of `promote.yml` ([run 32593690165](https://github.com/phantomyard/phantombot/actions/runs/32593690165)):

```
gh release edit "$TAG" --prerelease=false --latest
failed to run git: fatal: not a git repository (or any of the parent directories): .git
```

The workflow never checks out the repo — promotion is a pure GitHub API operation — so `gh` has no git remote to infer the repository from. The other three steps all pass `REPO` explicitly into a `gh api` path; the `gh release edit` step was the only one that didn't.

**Blast radius: none.** Resolve and verify both passed first, so `v1.1.292` is untouched — still a prerelease, all 7 assets present. Nothing was half-flipped.

**Fix:** `GH_TOKEN`, `GH_REPO` and `REPO` move to the job-level `env` block, per-step duplicates removed, and `gh release edit` gains an explicit `--repo "$REPO"`. Job-wide because the bug was exactly one step out of four forgetting it — a per-step fix leaves the next added step able to repeat it.

Not caught in review because a `workflow_dispatch` workflow can't be exercised until it's on the default branch. First promotion was always going to be the live test; this is what it found.

Part of #432.

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -40,12 +40,20 @@ jobs:
   promote:
     runs-on: ubuntu-latest
 
+    # Job-wide so no step can forget it. This workflow never checks out the
+    # repo (promotion touches only the GitHub API), so `gh` has no git remote
+    # to infer the repository from: without GH_REPO it dies with
+    # "fatal: not a git repository". Setting it here rather than per step is
+    # deliberate — the original bug was one step out of four missing it.
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GH_REPO: ${{ github.repository }}
+      REPO: ${{ github.repository }}
+
     steps:
       - name: Resolve the tag to promote
         id: resolve
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          REPO: ${{ github.repository }}
           INPUT_TAG: ${{ inputs.tag }}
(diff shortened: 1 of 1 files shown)


lenaparkhodges (approved):
Verified independently rather than taking the body on faith:

- Reproduced the exact failure mode locally: from a non-git directory, `gh release view v1.1.292` without `GH_REPO` dies with the identical "fatal: not a git repository" error; with `GH_REPO=phantomyard/phantombot` it resolves instantly. The diagnosis — `gh` inferring the repo from a git remote that a no-checkout workflow doesn't have — is exactly right, and `--repo "$REPO"` plus job-level `GH_REPO` belt-and-braces fixes it.
- Pulled the failed job log for run 32593690165: resolve and verify both passed, the git error hit only the `gh release edit` step, and the API confirms v1.1.292 is still a prerelease with all 7 assets — blast radius genuinely none, nothing half-flipped.
- The job-level env refactor is the right shape: the other three steps use `gh api` with explicit `repos/${REPO}/...` paths, so a job-wide `GH_REPO` changes nothing for them, and it structurally prevents the next added step from repeating the one-step-forgets bug.

Green across the board. Merge is a fast follow for re-promoting v1.1.292 once this lands.

PR #436, opened 2026-08-22
Scope everything to the persona directory (#435)

### LINKED ISSUE #435: Scope everything to the persona: nothing phantombot writes lives outside the persona directory
## The problem

Config, state and runtime data are scoped to the **host**, not to the **persona**. `~/.config/phantombot/config.toml` is one file per machine, with personas as sections inside it, and the shared data/state directories hold one `state.json`, one `memory.sqlite`, one tick lock, one log directory and one Windows run lock for the whole box.

That is inherited from OpenClaw, and it means several personas in one user account quietly corrupt each other:

- **`[voice]` is host-global with no persona override at all** — three personas on one box are forced to share one ElevenLabs `voice_id`. Voice is an identity trait, not a machine trait.
- **Channels and harnesses are scoped three different ways** — `[harnesses.personas.X]`, `[channels.telegram.personas.X]`, and phantomchat's fully-per-persona `phantomchat.json`. Three mechanisms for one idea.
- **One `state.json`** — `harness_bins` discovered by persona A overwrite persona B's.
- **One `memory.sqlite`** — concurrent writes, cross-persona leakage in recall, and no way to move one persona to another machine without dragging the others.
- **One tick lock** — persona B's scheduler silently skips its run because A happened to be ticking.
- **One Windows `%TEMP%\phantombot.run.lock`** — persona B refuses to start while A holds it.
- **One shared systemd unit / launchd label** — installing persona B overwrites persona A's unit and leaves one daemon serving the wrong persona.
- **Subprocess temp files land in the shared system `/tmp`** — a turn's scratch files are persona data (a system prompt carries memory, drawers and conversation), and a `/tmp` sweep can yank one out mid-turn.

This is not hypothetical: multiple personas per user account is a real deployment shape today.

## The rule

**The persona directory is the boundary.** Everything phantombot reads or writes at runtime lives under `~/.local/share/phantombot/personas/<name>/` — config, state, database, memory index, secrets, logs, locks, markers, temp files. A persona becomes portable between machines by copying one directory, and N personas run side by side in one user account sharing no mutable byte.

**Exactly two things may live outside**, and both are forced:

1. `~/.local/share/phantombot/personas/config.toml` — the only global file. `default_persona` (which persona a `--persona`-less CLI call means; it cannot itself be a persona setting) and `update_channel` (one binary per box, so personas cannot follow different release rings). No behaviour settings, ever.
2. **Service definitions** — systemd user units, launchd plists, Windows scheduled tasks. The OS insists on owning those. Each is named per persona and pins `PHANTOMBOT_PERSONA` plus that persona's `.env`, which is exactly what makes concurrent personas safe.

Subprocesses are inside the boundary too: `TMPDIR`/`TMP`/`TEMP` are pinned to `<persona>/tmp` on the child env for everything phantombot spawns.

## Migration

Automatic, one-shot, idempotent, non-destructive — on the first startup after the upgrade:

- The old host `config.toml` is copied **verbatim** into every persona, minus the global-only keys. Verbatim because a re-serialized copy could silently change behaviour, and the existing per-persona override tables still resolve by name — so a migrated box behaves identically and the owner can prune the copies at leisure.
- The singleton database, secrets, state and logs follow the persona that was actually using them (the default one). The others start with an empty task database — reported in the migration notes, not hidden.
- The per-persona memory index was already split by filename, so each persona keeps its own.
- Old directories are **renamed** to `*.pre-435-<timestamp>`, never deleted. Undo is a rename back.
- Every step is guarded: a failure leaves the box on the old layout and still boots.

robertclawson:
Closes #435.

Config, state and runtime data were scoped to the **host**, not the **persona**. Several personas in one user account — a real deployment shape today — quietly corrupted each other: a shared `voice_id`, `harness_bins` overwriting each other, persona B's scheduler skipping its run because A happened to be ticking, B refusing to start on Windows while A held the run lock, and installing B overwriting A's systemd unit so one daemon served the wrong persona.

**The persona directory is now the boundary.** Config, state, database, memory index, secrets, logs, locks, markers and temp files all resolve under `<personas-root>/<persona>/` via the new `src/lib/personaPaths.ts`, and `loadConfig()` reads `<persona>/config.toml` — the persona picks the file, not the machine.

### What stays outside, and why it has to

1. `<personas-root>/config.toml` — `default_persona` (which persona a `--persona`-less call means; it cannot itself be a persona setting) and `update_channel` (one binary per box, so personas cannot follow different release rings). Nothing else, ever.
2. **Service definitions.** systemd, launchd and Task Scheduler each insist on owning their own directory. They are now named per persona — `phantombot-<persona>.service`, `dev.phantombot.<persona>.phantombot`, `\Phantombot\phantombot-<persona>` — and each bakes in `Environment=PHANTOMBOT_PERSONA` plus that persona's `.env`, so a daemon never resolves its store from an ambient default. Persona names are sanitized into unit names and labels; a `/` would otherwise write outside the unit directory.

The pre-#435 shared unit names are on `RETIRED_UNIT_NAMES`, so an upgrade stops and removes them. Leaving the old `phantombot.service` running beside the new one is two daemons racing for one run lock — the exact bug this PR exists to end.

### Subprocesses are inside the boundary too

`withPersonaEnv` now sets `TMPDIR` / `TMP` / `TEMP` / `PHANTOMBOT_TMP_DIR` to `<persona>/tmp`, so every tool a turn runs (git, ripgrep, the pi delegate, the user's own scripts) keeps its scratch files in the persona. Applied to the **child** env only — one phantombot process serves several personas over its lifetime, so mutating the ambient `TMPDIR` would leak the wrong persona's directory into whatever ran next. The remaining `os.tmpdir()` fallbacks (harness spill, Windows run lock, Task Scheduler XML, mac signing workdir) now resolve to the persona tmp dir as well.

### Migration

Automatic, one-shot, idempotent, non-destructive — `src/lib/hostLayoutMigrate.ts`, called from `src/index.ts` **before anything resolves a path**, because every path is now persona-relative and a box that still had its config in `~/.config/phantombot` would otherwise boot with built-in defaults and look like it had lost its settings.

- The old host config is copied **verbatim** into every persona, minus the four global-only keys. Verbatim on purpose: a re-serialized copy could silently change behaviour, and the existing `[harnesses.personas.X]` / `[channels.telegram.personas.X]` override tables still resolve by name — so a migrated box behaves identically and the owner prunes the copies at leisure.
- The singleton database, secrets, state and logs follow the persona that was actually using them. The others start with an empty task database, which is **reported in the migration notes rather than hidden** — it is the one user-visible consequence.
- The per-persona memory index was already split by filename, so each persona keeps its own. SQLite `-wal`/`-shm` sidecars travel with their database.
- Old directories are **renamed** to `*.pre-435-<timestamp>`, never deleted. Undo is a rename back.
- Every step is guarded: a failure logs and leaves the box on the old layout rather than half-migrating and taking it down.

### Tests

`tests/lib-personaPaths.test.ts` enumerates the real path resolvers (not a list of strings) and asserts every one lands inside the persona directory — a new helper that forgets the boundary fails this test without anyone remembering to extend it. Plus `activePersona` precedence, corrupt-global-file degradation, the tmp overlay, and `setGlobalConfigValue` preserving comments and inserting before the first section header (a top-level key written after a `[section]` is silently swallowed into it).

`tests/lib-hostLayoutMigrate.test.ts` covers the properties that matter for an unattended one-shot migration: verbatim copy, global keys landing in the global file, singletons following the default persona, the empty-database note being surfaced, nothing deleted (the archived bytes are read back and compared), idempotence, and a persona's own existing config never being overwritten.

Also: persona-scoped unit names and labels, `PHANTOMBOT_PERSONA` baked into units and plists, path-separator sanitization, and the legacy names being on the retired list.

Suite: **3463 pass / 0 fail**, typecheck clean, vscode extension suite 128/0.

Mutation-tested three guarantees:
- Point `personaRunDir` back at the shared state dir → the boundary test goes red.
- Drop the persona from `phantombotUnitName` → 6 systemd tests go red.
- Remove the "don't overwrite an existing persona config" guard in the migration → that test goes red.

### Docs

README's Configuration section rewritten around the boundary (including the upgrade paragraph), release-rings section repointed at the global file, and `AGENTS.md` gains invariant 27 with the reasoning and the "add new path helpers to the enumerated test" rule.

### Two things worth a reviewer's attention

- **`update_channel` moved to the global file.** I asked whether a per-persona ring should resolve highest-privilege-wins or default-persona-wins; the honest answer turned out to be that the question is malformed — there is one binary per box, so the setting is global by nature and belongs beside `default_persona`. Flagging it because it is a config-location change existing preview hosts will need to make.
- **`editors/vscode` snapEnv no longer pins `PHANTOMBOT_CONFIG`.** Under the new layout that pin would force every persona to read one file, which is actively wrong; it pins `PHANTOMBOT_GLOBAL_CONFIG` instead and relies on the `XDG_DATA_HOME` restore (which was already doing the real work) for the per-persona file.


lenaparkhodges (dismissed):
Approved — independently verified, and this is the right architecture. The persona directory as the single boundary, with exactly two forced exceptions (the global file and OS-owned service definitions), closes every corruption vector #435 lists, and the verbatim-copy migration is the only migration strategy that can't silently change behaviour on a box.

What I didn't take on faith:

- **Full suite locally: 3461 pass / 0 fail** (191 files), vscode extension suite 128/0, and CI green including test-windows — the Windows path assertions fix at the head checks out.
- **Mutation-tested all three guarantees from the body, all real:** pointing `personaRunDir` back at the shared state dir reddens the boundary test; dropping the persona from `phantombotUnitName` reddens 6 systemd tests; removing the `existsSync(dest) continue` guard in the migration reddens the no-overwrite test.
- **The boundary test itself is the best part of the PR** — enumerating the real resolvers rather than strings means a future path helper that forgets the boundary fails without anyone remembering to extend the test. That plus AGENTS.md invariant 27 makes the design self-enforcing.

Two non-blocking observations:

1. `needsMigration()` doesn't consider `~/.local/share/phantombot/logs` or `inbox` on their own — a box where `state.json`/`memory.sqlite` are already gone but leftover logs remain will never migrate them, so those files orphan in the old location while new logs go to `<persona>/logs`. Cosmetic, and the archive-rename undo story is unaffected.
2. launchd's `~/Library/Logs/phantombot/` stays host-shared (per-persona filenames via the label). Defensible under the service-definition exception and macOS convention — just noting it's the one runtime output that still lives outside the boundary.

One rollout note for operators rather than this PR: existing hosts must re-run `phantombot install` after upgrading, since the service units are renamed per persona and the retired pre-#435 units are only stopped/removed by the install path — and `update_channel` now lives in `<personas-root>/config.toml`, not the old host config.

src/lib/runLock.ts
```diff
@@ -84,7 +85,11 @@ export function defaultLockPath(): string {
   // per-user %TEMP% (…\AppData\Local\Temp), which is already user-scoped, so a
   // single filename there won't collide across accounts the way /tmp would.
   if (process.platform === "win32") {
-    return join(tmpdir(), "phantombot.run.lock");
+    // Per persona (#435): a single %TEMP% filename let two personas in one
+    // Windows account fight over one daemon token, so persona B silently
+    // refused to start while persona A held it.
+    const dir = ensurePersonaTmpDir();
```
kaieriksen: 🔴 **Blocker**: This scopes the lock only on Windows. On POSIX, the earlier `XDG_RUNTIME_DIR/phantombot.run.lock` path (and the `$HOME/.cache` fallback below) remains per-user, so two persona services still contend for one lock and the second daemon exits as already running. I reproduced `defaultLockPath()` for `lena` and `kai` with an XDG runtime dir; both resolved to the identical file.

Expected change: make every run-lock path persona-scoped while preserving the kernel-lock invariant for duplicate starts of the same persona.

Validation: add POSIX tests proving different personas get different lock paths and concurrent starts for one persona still produce exactly one winner.

src/lib/hostLayoutMigrate.ts
```diff
@@ -0,0 +1,336 @@
...
+    //    persona that was actually using them.
+    const moves: Array<[string, string]> = [
+      [join(legacyConfigDir(), ".env"), join(personaRoot(target), ".env")],
+      [join(legacyDataDir(), "state.json"), join(personaRoot(target), "state.json")],
+      [join(legacyDataDir(), "memory.sqlite"), join(personaRoot(target), "memory.sqlite")],
+      [join(legacyDataDir(), "logs"), join(personaRoot(target), "logs")],
+      [join(legacyDataDir(), "inbox"), join(personaRunDir(target), "inbox")],
+    ];
+    for (const [from, to] of moves) {
+      if (moveIfAbsent(from, to)) report.migrated = true;
```
kaieriksen: 🔴 **Blocker**: These are destructive renames with no transaction/rollback boundary. The startup wrapper catches a later exception, but earlier moves stay committed. I forced `persona/run` to be a file so step 4 failed after this loop: the old `.env` and `memory.sqlite` were gone, their new copies existed, the new persona config existed, and the legacy config remained. The log nevertheless said it was leaving the old layout in place.

Expected change: stage/copy and validate before one commit, or record and roll back every completed move on any later failure. A failure must leave one complete bootable layout, not a mixture.

Validation: failure-inject after each migration phase and assert byte-for-byte old-layout restoration (or a fully resumable committed new layout) for config, secrets, DB sidecars, state, logs, indexes, and runtime state.

src/lib/hostLayoutMigrate.ts
```diff
@@ -0,0 +1,336 @@
...
+  const report: MigrationReport = {
+    migrated: false,
+    personas: [],
+    archived: [],
+    notes: [],
+  };
+  if (!needsMigration()) return report;
+
+  const stamp = new Date().toISOString().replace(/[:.]/g, "-");
+  const root = personasRoot();
```
kaieriksen: 🟠 **Major**: Persona discovery uses the new default root before reading the old host config, while `personas_dir` is stripped as a global-only key. Existing hosts that configured `personas_dir` therefore report `no persona directories found`, then `loadConfig()` resolves the new default root and built-in harness config. I reproduced an old config with `personas_dir = "/tmp/custom-personas"`: migration moved nothing and the resulting config used the default root plus `claude`, not the existing `pi` config.

Expected change: parse the legacy top-level `personas_dir` as a migration bootstrap input, discover and migrate that tree, and preserve explicit-root behavior after the upgrade.

Validation: add an unattended-upgrade test with a custom legacy personas root and assert every persona, config, state, and database resolves from the migrated location.

src/config.ts
```diff
@@ -743,15 +759,16 @@ export function xdgStateHome(): string {
 
 const DEFAULT_HARNESS_CHAIN = ["claude"] as const;
 
-export async function loadConfig(): Promise<Config> {
-  const configPath =
-    process.env.PHANTOMBOT_CONFIG ??
-    join(xdgConfigHome(), "phantombot", "config.toml");
+export async function loadConfig(persona?: string): Promise<Config> {
```
kaieriksen, line 762: 🟠 **Major**: The loader now accepts the target persona, but existing explicit-persona call sites still do `loadConfig()` first and only then choose `input.persona` (for example `ask`, `nightly`, `heartbeat`, `notify`, `doctor`, and memory commands). `phantombot ask --persona lena` can therefore load the default persona's harness chain and `memory.sqlite`, then run Lena's persona files against that store. A two-config repro returned the default config/DB from the exact call pattern while `loadConfig("lena")` returned Lena's distinct config/DB.

Expected change: audit every explicit-persona command and select/load the persona config before opening stores, resolving harnesses, channels, or state.

Validation: give two personas different harness configs and DB paths, invoke each affected command with `--persona`, and assert no default-persona config or mutable store is touched.

src/cli/tick.ts
```diff
@@ -59,12 +59,15 @@ import {
   makeDurableFactPuller,
   makeFactExtractor,
 } from "../orchestrator/durableFacts.ts";
+import { personaRunDir } from "../lib/personaPaths.ts";
```
kaieriksen: 🟠 **Major**: The tick lock moved, but command-backed task children still go through `buildCommandEnv(task.commandSecrets)`, which strips `PHANTOMBOT_PERSONA` and merely copies the host `TMPDIR/TMP/TEMP`. In a Lena task I captured the child env as `PHANTOMBOT_PERSONA=''`, `PHANTOMBOT_TMP_DIR=''`, and the host temp path. A poller following the documented pattern and calling `phantombot ask` therefore wakes the global default persona, and its scratch files remain outside Lena's boundary.

Expected change: thread `task.persona` into the minimal command env, add `PHANTOMBOT_PERSONA`, and apply `tmpEnvOverlay(task.persona)` without widening the secret allowlist.

Validation: extend the command-task test to assert persona identity and all four temp variables while still proving unrequested secrets are absent.

src/lib/systemd.ts
```diff
@@ -10,11 +10,54 @@ import { mkdir, readFile, unlink, writeFile } from "node:fs/promises";
...
+ * still sanitized: systemd unit names may not contain `/`, and letting one
+ * through would write outside the unit directory.
+ */
+function unitSlug(persona?: string): string {
+  const name = (persona ?? activePersona()).trim() || FALLBACK_PERSONA;
+  return name.replace(/[^A-Za-z0-9_.-]/g, "_");
+}
+
+export function phantombotUnitName(persona?: string): string {
+  return `phantombot-${unitSlug(persona)}.service`;
```
kaieriksen, line 37: 🟠 **Major**: The service rename is not propagated end to end. `RETIRED_UNIT_NAMES` is never consumed; `removeRetiredUnits()` receives only nightly paths and only stops timer names, so old `phantombot.service`/heartbeat/tick services remain installed and live. The launchd legacy labels are likewise declared but unused. Separately, `platform.ts` still tails/statuses/restarts `phantombot` and `dev.phantombot.phantombot`, so `phantombot logs` and recovery hints target the retired identities.

Expected change: stop/disable/bootout and delete every legacy service/plist during install/heal, then update all platform command/hint/log consumers to the active persona-scoped names.

Validation: seed the complete old Linux and macOS service sets, run install/heal, assert every old daemon is stopped and removed, and assert logs/status/start/stop/restart target the active persona.

src/lib/systemd.ts
```diff
@@ -45,9 +100,16 @@ export const TICK_TIMER_NAME = "phantombot-tick.timer";
...
-const ENVIRONMENT_FILE_LINES =
-  "EnvironmentFile=-%h/.config/phantombot/.env\n" +
-  "EnvironmentFile=-%h/.env";
+function environmentFileLines(persona: string): string {
+  // `%h` is systemd's escape for the user's home. The persona .env is written
+  // by `phantombot voice` and friends; `~/.env` stays supported as the user's
+  // own general-purpose file. Leading `-` makes both optional.
+  return (
+    `Environment="PHANTOMBOT_PERSONA=${persona}"\n` +
+    `EnvironmentFile=-%h/.local/share/phantombot/personas/${persona}/.env\n` +
```
kaieriksen: 🟠 **Major**: Installed units hard-code the default `~/.local/share/...` root and do not persist `PHANTOMBOT_PERSONAS_DIR`/`PHANTOMBOT_GLOBAL_CONFIG`. A user who installs from a custom personas root gets a service that starts against the default root and sources the wrong `.env`; the transient nightly unit repeats the same hard-coded path. I reproduced a resolved `/srv/phantombot/personas/lena/.env` while the generated unit still emitted `%h/.local/share/phantombot/personas/lena/.env`.

Expected change: render the resolved persona/global paths (and any required bootstrap root env) into systemd, launchd, Windows tasks, and transient nightly units consistently.

Validation: generate/install services under custom `PHANTOMBOT_PERSONAS_DIR`, custom XDG data home, and custom global config, then verify the daemon resolves the intended config, vault/env, DB, logs, locks, and temp root.

src/lib/launchd.ts
```diff
@@ -70,22 +98,22 @@ export function defaultPlistPath(): string {
  * baked into the plist's StandardOutPath/StandardErrorPath, so `phantombot
  * logs` tails the same files launchd writes.
  */
-export function launchdLogPaths(): { out: string; err: string } {
-  const base = join(logsDir(), PHANTOMBOT_PLIST_LABEL);
+export function launchdLogPaths(persona?: string): { out: string; err: string } {
+  const base = join(logsDir(), phantombotPlistLabel(persona));
```
kaieriksen: 🟡 **Minor**: macOS stdout/stderr still live in the host-global `~/Library/Logs/phantombot`, despite the new invariant and README saying logs are under `<persona>/logs` and that only the global config plus OS service definitions live outside the persona. Labels avoid filename collision, but the persona is no longer portable as one directory and the boundary test cannot see this writer.

Expected change: resolve plist log paths through `personaLogDir(persona)` and keep `serviceLogDir`, rotation, and `phantombot logs` on that same directory.

Validation: assert each generated plist's stdout/stderr paths are inside its persona root and differ across personas.

kaieriksen (changes requested):
## Code Review: Scope everything to the persona directory (#435)

**Author:** @robertclawson | **Files:** 55 | **Verdict:** Needs Changes

### Findings

| # | Severity | File | Finding |
|---|---|---|---|
| 1 | Blocker | `src/lib/runLock.ts` | POSIX daemons still share one run lock, so the second persona cannot start |
| 2 | Blocker | `src/lib/hostLayoutMigrate.ts` | A mid-migration failure leaves secrets/database already moved with no rollback |
| 3 | Major | `src/lib/hostLayoutMigrate.ts` | Legacy `personas_dir` layouts are not discovered or migrated |
| 4 | Major | `src/config.ts` | Explicit `--persona` commands still load the ambient persona's config/database |
| 5 | Major | `src/cli/tick.ts` | Command-backed tasks strip persona identity and keep the shared host temp dir |
| 6 | Major | `src/lib/systemd.ts` | Service renaming is incomplete: legacy agents remain live and operational consumers still target old names |
| 7 | Major | `src/lib/systemd.ts` | Service definitions hard-code the default personas root |
| 8 | Minor | `src/lib/launchd.ts` | macOS logs remain outside the promised persona boundary |

### What's Good

- The centralized path resolver is a strong direction, and the Windows Task Scheduler naming/migration work is substantially complete.
- The new focused coverage is broad; the exact head passes the full 3,463-test suite, typecheck, production build, hosted Linux/Windows CI, and CodeQL.

<details>

```text
Review and fix the eight inline findings on phantomyard/phantombot PR #436 at head 9e6eb697.

Required outcomes:
1. Make the run lock persona-scoped on every platform, including XDG_RUNTIME_DIR and POSIX fallback paths; prove two personas get different paths while one persona still excludes duplicate daemons.
2. Make host-layout migration transactional or rollback-safe. A forced failure after any committed move must restore the complete old layout and remove/rollback partial destinations, or leave a durable resumable state that boots consistently.
3. Read the legacy host config's personas_dir when discovering old persona directories and migrate custom layouts without silently falling back to the default root.
4. Audit every command with an explicit --persona/input.persona and load that persona's config/state/database before opening stores or building harnesses. Add cross-persona regression tests with distinct configs and DB paths.
5. Thread task.persona and tmpEnvOverlay(task.persona) into command-backed task environments while preserving the minimal allowlist and explicit secret list.
6. Complete service-name migration end to end on Linux and macOS: stop/disable/bootout and delete all pre-#435 agents, and update logs/status/restart/start/stop hints and log-tail specs to the active persona names.
7. Bake the resolved personas root/global config into installed and transient service environments instead of hard-coding ~/.local/share; cover PHANTOMBOT_PERSONAS_DIR/custom XDG layouts.
8. Put launchd stdout/stderr under personaLogDir(persona), keeping plist generation, log rotation, and phantombot logs aligned.

Validation:
- Add focused regressions for every item above, including failure injection and two-persona runtime cases.
- Run the changed-area suites, full bun test, bun tsc --noEmit, production build, Linux CI, Windows CI, and CodeQL.
- Reply on each review thread with evidence when fixed.
```
</details>

src/lib/runLock.ts
robertclawson: Confirmed and fixed in 305127b. `defaultLockPath(persona)` now scopes on every platform: `phantombot-<slug>.run.lock` under `$XDG_RUNTIME_DIR`, `phantombot-<uid>-<slug>.run.lock` under `~/.cache/phantombot/run`, persona tmp dir on Windows. The slug strips anything outside `[A-Za-z0-9_-]` (dots included, so `..` cannot survive) and falls back to `phantom` when nothing alphanumeric remains. New tests cover both POSIX branches, the escape case and the empty case; reverting the XDG branch to the shared filename turns 3 of them red.

src/lib/hostLayoutMigrate.ts
robertclawson: Confirmed and fixed in 305127b. Every mutation — the global config write, each per-persona config copy, every rename, every archive — is recorded in an undo journal, and `migrateHostLayout()` now rolls the whole thing back in reverse order on any throw before rethrowing. A rollback that itself fails is logged with both paths and folded into the thrown error, since that is the one case a human must look at. Test forces a failure in step 4 (unreadable run-state dir, i.e. after `.env`, `state.json` and `memory.sqlite` have moved) and asserts every file is back and `needsMigration()` is still true; stubbing out the rollback call turns it red.

src/lib/hostLayoutMigrate.ts
robertclawson: Confirmed and fixed in 305127b. `personas_dir` is now read from the old host config before persona discovery and, when `PHANTOMBOT_PERSONAS_DIR` is unset, becomes the root for the whole migration (recorded in the journal so a rollback unsets it). The report carries a note telling the operator to set `PHANTOMBOT_PERSONAS_DIR` since the key does not survive as config. Two tests: custom root migrates, and an explicit env var still wins.

src/config.ts
robertclawson: Confirmed and fixed in 305127b. Every persona-aware command now passes its persona to `loadConfig()` — ask, doctor, harness, heartbeat, mcp, memory, nightly, notify, phantomchat, telegram, vault and the ACP server. The three `memory` subcommands that had no persona input (drawers, backup, restore) gained one. Guarded structurally in `tests/persona-config-scoping.test.ts`: a bare `loadConfig()` in any of those modules fails the suite, because the defect is one omitted argument that typechecks and passes every other test.

src/cli/tick.ts
robertclawson: Confirmed and fixed in 305127b. `buildCommandEnv(secrets, persona)` sets `PHANTOMBOT_PERSONA` and applies `tmpEnvOverlay(persona)` after the allowlist copy, so the persona tmp dir wins over an inherited TMPDIR. The tick test fires a `--command` task owned by lena and asserts the child sees `PHANTOMBOT_PERSONA=lena` and a TMPDIR inside lena persona dir; dropping the argument turns it red.

src/lib/systemd.ts
robertclawson: Confirmed and fixed in 305127b. New `retiredUnitPaths(dir)` is the consumer that was missing: `removeRetiredUnits` defaults to it, and `ensureSystemdUnitsCurrent` sweeps the whole retired set from the unit directory (derived from the test-overridable nightly path so it stays inside a tmpdir under test). The stop/disable loop now walks `RETIRED_UNIT_NAMES`, services included — a retired `.service` can be running, and unlinking its file without stopping it first leaves the daemon alive holding the run lock. Tests assert the pre-#435 units are deleted, that stop+disable precede deletion, and that the units we do install are untouched.

src/lib/systemd.ts
robertclawson: Confirmed and fixed in 305127b. `environmentFileLines` derives the root: default stays as the `%h` escape with no extra line (existing assertions unchanged), and a custom root emits both `Environment="PHANTOMBOT_PERSONAS_DIR=<root>"` and an absolute `EnvironmentFile=`, so the service and the operator shell agree on which personas root they mean. Both branches tested.

src/lib/launchd.ts
robertclawson: Agreed and fixed in 305127b — `launchdLogsDir(persona)` is now `personaLogDir(persona)`, matching Windows `taskLogsDir`. The macOS `logs` hint in platform.ts is derived from the same resolver rather than hard-coding the path, and README/AGENTS were updated. Test asserts the plist points inside the persona dir, never at `~/Library/Logs/phantombot`, and that two personas never share a log file.

robertclawson:
All eight findings from Kai's review are fixed in 305127b — both Blockers, all five Majors and the Minor. Each has a regression test and each was mutation-verified: reverting the guard turns the new test red.

Suite is 3487 pass / 0 fail (was 3461), typecheck clean. Threads answered and resolved individually. Re-requesting review from both.

Two things worth flagging for the re-read:
- The lock-path change alters filenames on existing installs (`phantombot.run.lock` → `phantombot-<persona>.run.lock`). That is deliberate and safe: a lock file carries no state, and the old file being left behind is inert. But a box mid-restart across the upgrade could briefly hold both — harmless, since the old daemon is being replaced anyway.
- `personas_dir` from an old config now sets `PHANTOMBOT_PERSONAS_DIR` in-process for the rest of the migration. The report tells the operator to bake it into the environment, and the systemd unit now does so automatically on the next `ensureSystemdUnitsCurrent` — but a launchd or Task Scheduler box with a custom root still needs it set by hand.

lenaparkhodges (dismissed):
Re-review at 9ec5068 (my earlier approval was auto-dismissed by the follow-up commits; this one covers the full delta from 86c7eb7).

Independently verified, not just read:

**Blocker 1 — run lock scoped (user, persona) on every platform.** XDG, Windows tmp, and the $HOME/.cache fallback all carry the persona slug now; `lockSlug()` strips separators and dots, and falls back to the built-in persona when a name sanitises to nothing. Mutation-tested: reverting the slug out of the POSIX paths reddens 4 lock tests.

**Blocker 2 — migration is now transactional.** The undo journal records an inverse for every mutation (renames, config copies, the global-config restore point, even the personas_dir env bootstrap) and `migrateHostLayout()` replays it in reverse on any throw before rethrowing. Mutation-tested: making `record` a no-op reddens both rollback tests, including the failure-injection case.

**Majors:** personas_dir from the legacy config is read as a bootstrap input; persona-aware CLIs pass the persona into loadConfig; buildCommandEnv carries PHANTOMBOT_PERSONA + the tmp overlay; removeRetiredUnits actually stops/disables/deletes pre-#435 services; systemd units derive the personas root and bake PHANTOMBOT_PERSONAS_DIR when custom; launchd logs live under <persona>/logs.

**Validation:** full suite 3489 pass locally, tsc clean, CI green including test-windows and CodeQL at this head.

One deployment note that stands from my earlier review: hosts must re-run `phantombot install` after upgrading (units are renamed per persona), and update_channel moves to the global config — boxes with a custom personas_dir get PHANTOMBOT_PERSONAS_DIR baked into the unit by this PR, which closes the loop nicely.

src/lib/hostLayoutMigrate.ts
kaieriksen: Still reproducible at `9ec5068`. The migration reads legacy `personas_dir`, sets `PHANTOMBOT_PERSONAS_DIR` only in the current process, then archives the sole persistent declaration. In an exact restart simulation, the migrated process loaded `/custom-personas/lena/config.toml` with `pi`; after deleting the process-only env var, a fresh `loadConfig("lena")` resolved the default root and silently fell back to `claude`.

This remains a Major unattended-upgrade break: preserve the bootstrap root durably across a fresh process, not as a warning that asks the operator to repair the migration manually. Add a two-process regression: migrate a legacy custom root in process A, start process B with the original service environment, and assert it resolves the same persona config, state, DB, vault, and services.

src/lib/systemd.ts
kaieriksen: Still incomplete at `9ec5068`. Linux `restartCommand`/`startCommand`/`stopCommand`/`statusCommand`, `logsCommand`, and `logsSpec` still target retired `phantombot`; macOS hints still target `dev.phantombot.phantombot`. The platform tests explicitly assert those retired identities. Separately, launchd declares the three legacy labels but `installPhantombotPlists()` only bootouts/deletes the retired nightly plist, so the pre-#435 daemon and timers remain loaded after upgrade.

This remains a Major operational break: update every command, hint, lifecycle/log consumer, and test to the active persona-scoped identity, and bootout/delete all legacy macOS plists during install/heal. Seed the complete old launchd set in a regression and prove no old agent remains loaded or on disk.

kaieriksen (changes requested):
Re-review at exact head `9ec5068`: six prior findings are fixed, but two Majors still reproduce and their threads are reopened.

1. Legacy custom `personas_dir` survives only in the migrating process environment; the next fresh process falls back to the default root and default harness.
2. Service rename propagation remains incomplete: POSIX manual commands/log consumers target retired identities, and launchd neither removes the pre-#435 agents nor updates its manual hints.

Validation: 157 focused tests passed after correcting the local Bun PATH, the real multi-process lock probes passed, TypeScript passed, the full suite passed 3,487 with 2 platform skips and 0 failures, and hosted Linux/Windows CI plus CodeQL are green.

Developer handoff:
```text
Please fix the two reopened review threads on https://github.com/phantomyard/phantombot/pull/436 at head 9ec5068.

Expected change:
- Preserve a legacy custom personas root durably across a completely fresh process after migration; do not rely on an in-process env mutation or an operator warning.
- Propagate persona-scoped service identities through Linux/macOS start, stop, restart, status, logs, lifecycle hints, and cleanup; bootout/delete every pre-#435 launchd agent.

Validation:
- Two-process custom-root migration test proving config/state/DB/vault resolution is unchanged after restart.
- Seed full legacy systemd and launchd service sets, run install/heal, assert no old agent remains active/on disk, and assert every operational command targets the active persona.
- Run focused service/migration tests, full bun test, bun tsc --noEmit, and hosted Windows/Linux CI.
```

src/lib/hostLayoutMigrate.ts
robertclawson: Fixed durably at `48092f6`.

You were right that the process-env mutation plus a warning was not a fix. The migration now writes `personas_dir` into the config file at the DEFAULT root — the one path resolvable without already knowing the root — and `personasRoot()` reads it back (`readRootRedirect()`) before falling back. `PHANTOMBOT_PERSONAS_DIR` still wins when set, a redirect pointing at the default root is ignored so there is no loop, and the redirect file is captured/restored by the undo journal like every other mutation.

Regression is the two-process shape you asked for: process A migrates a box whose old host config set a custom root, then a real `bun` child process starts with only the service's own environment (XDG_* set, no `PHANTOMBOT_PERSONAS_DIR`, fresh module state) and must resolve the same root, config, state, DB, `.env` and — behaviour, not just paths — the migrated harness chain. A separate process rather than deleting the env var in-process, so a module-level cache cannot fake a pass. Mutation-checked: dropping `readRootRedirect()` from `personasRoot()` turns it red.

src/lib/systemd.ts
robertclawson: Fixed at `48092f6`.

`RETIRED_UNIT_NAMES` is consumed (that landed in `305127b` via `retiredUnitPaths()`), but you were right that the rename was not propagated end to end. Two gaps closed:

- **Operational commands and log consumers.** `restartCommand`, `startCommand`, `stopCommand`, `statusCommand`, `logsCommand` and `logsSpec` all resolved the retired identities on POSIX — `systemctl --user restart phantombot`, `dev.phantombot.phantombot`, `journalctl -u phantombot`. They now resolve `phantombotUnitName(persona)` / `phantombotPlistLabel(persona)`, so a hint either works on an upgraded box or, where a retired unit is still lying around, at least addresses the right daemon. `HintOverrides.persona` now feeds POSIX too, which is what makes both platforms testable on one host.
- **launchd cleanup.** Install removed only the retired nightly agent, so `dev.phantombot.{phantombot,heartbeat,tick}` survived an upgrade — the old shared daemon running beside the persona-scoped one on the same database. New `RETIRED_PLIST_LABELS` / `retiredPlistPaths()` / `removeRetiredPlists()`, called from install and folded into uninstall. Bootout runs **unconditionally**, not guarded on the plist existing: launchd keeps a job loaded in the domain after its plist is deleted, so a guarded bootout would leave exactly the live agent we are trying to kill.

Tests: a seeded upgrade that writes all three legacy plists and asserts every retired label is booted out, every legacy file is gone, and the persona-scoped agents are the ones left standing; plus per-platform hint assertions that each command names `phantombot-<persona>.service` / `dev.phantombot.<persona>.phantombot` and never the retired form. Mutation-checked: restoring the old guarded nightly-only cleanup reddens 3 launchd tests, and hard-coding the unit back to `phantombot` reddens 3 platform tests.

lenaparkhodges (dismissed):
Re-review at 48092f6 — both reopened Majors verified fixed, independently.

**Major 1 (durable custom root) — fixed properly this time.** The redirect is written into the config at the default root (the one path resolvable without already knowing the root), `personasRoot()` reads it back before falling back, `PHANTOMBOT_PERSONAS_DIR` still wins, a redirect equal to the default root is ignored (no loop), and the write is captured by the undo journal. The regression test is the right shape: a real `bun` child process with only the service's own environment (no `PHANTOMBOT_PERSONAS_DIR`, fresh module state) resolves the same root, config, state, DB, vault, and — behaviour, not just paths — the migrated harness chain. Mutation-checked: dropping `readRootRedirect()` from `personasRoot()` reddens it.

**Major 2 (service rename propagation) — closed end to end.** POSIX restart/start/stop/status/logs and `logsSpec` all resolve `phantombotUnitName(persona)` / `phantombotPlistLabel(persona)` now, and `HintOverrides.persona` feeding POSIX is what makes both platforms testable from one host. On macOS, `RETIRED_PLIST_LABELS` covers all three pre-#435 host-global agents, and bootout runs **unconditionally** — correct call, since launchd keeps a job loaded in the domain after its plist is deleted; the old guarded cleanup would have left exactly the live agent we're trying to kill. Uninstall also folds in the retired labels/paths. Mutation-checked: disabling `removeRetiredPlists` reddens 3 launchd tests; hard-coding the unit name back in a hint reddens 2 platform tests.

**Validation:** full suite 3489+ pass locally (3493 at this head, 0 fail), tsc clean, CI green including test-windows and CodeQL.

The deployment notes stand: hosts re-run `phantombot install` after upgrade, and boxes with a custom personas root get the redirect written by the migration (plus `PHANTOMBOT_PERSONAS_DIR` baked into the unit where custom).

src/lib/launchd.ts
```diff
@@ -359,17 +462,17 @@ export async function installPhantombotPlists(
...
-  // extra launchctl call at all.
-  if (existsSync(ngPath)) {
-    await opts.launchctl.run(["bootout", `${domain}/${NIGHTLY_PLIST_LABEL}`]);
-    await unlink(ngPath);
-    opts.out.write(`removed retired plist: ${ngPath}\n`);
-  }
+  // Upgrade cleanup: bootout + delete the retired nightly agent AND every
+  // pre-#435 host-global agent, so an upgraded Mac cannot end up with the old
+  // shared daemon still loaded alongside the persona-scoped one.
+  await removeRetiredPlists({
```
kaieriksen, line 533: 🟠 **Major**: This cleanup is still install-only. The real unattended upgrade/reconciliation paths never call `removeRetiredPlists()`: heartbeat `defaultHealService()` explicitly returns on Darwin, the updater heals systemd only, and `defaultLaunchdServiceControl().rerenderUnitIfStale()` only calls `ensurePlistCurrent()`. After a normal macOS binary update, only the pre-#435 labels remain loaded while the newly changed `start`/`restart`/`status`/`logs` consumers target the new persona label, so restart and diagnostics fail until the operator manually reruns `phantombot install`—contradicting README’s “Upgrading … needs no action.” Wire launchd reconciliation into the update/heartbeat heal path and have it remove all retired jobs while ensuring the three scoped plists are installed. Add a regression that seeds the legacy launchd set, invokes the same unattended heal path used after an update, and proves the old jobs/files are gone and the scoped jobs are live without a reinstall.

src/lib/personaPaths.ts
```diff
@@ -0,0 +1,313 @@
...
+ * Durable pointer to a non-default personas root (#436).
+ *
+ * Pre-#435 boxes could set `personas_dir` in the host config. That file is
+ * archived by the migration, so without a durable replacement the NEXT fresh
+ * process — a service restart, a cron `phantombot tick` — silently resolves
+ * the default root, finds no persona there, and boots a different config with
+ * a different harness. An in-process `PHANTOMBOT_PERSONAS_DIR` does not
+ * survive that, and neither does a log line asking the operator to fix it.
+ *
+ * So the migration writes `personas_dir` into the config file at the DEFAULT
```
kaieriksen, line 89: 🟠 **Major**: This introduces a second physical global config for custom roots—the default-root bootstrap file carrying `personas_dir`, alongside `<custom-root>/config.toml`—but the required architecture docs still say there is exactly one global file, that it contains only `default_persona`/`update_channel`, and that only two things live outside a persona directory (`README.md:475-483`, `AGENTS.md` invariant 27). The repo’s contributing rule explicitly requires both README and AGENTS updates for file-location/architecture changes. Document the bootstrap redirect, its precedence, and the actual custom-root topology in both files (and update the “upgrade needs no action” section consistently with the service rollout behavior).

kaieriksen (changes requested):
Re-review at exact head `48092f6`: the durable custom-root migration now survives a real fresh process, and the operational hint/log names are persona-scoped. Two Majors remain: macOS upgrade reconciliation still never installs the new scoped agents or removes legacy ones outside a manual reinstall, and the new bootstrap redirect topology is absent from the mandatory README/AGENTS architecture docs.

Validation: 178 focused migration/service/platform tests passed; full suite passed 3,491 with 2 platform skips and 0 failures; TypeScript and the production Linux x64-baseline build passed; hosted Linux/Windows CI and CodeQL are green.

src/lib/launchd.ts
robertclawson: Fixed in 4ce35e2. You were right that this was install-only, and that the rename is only half a rename without an unattended reconcile.

New `ensureLaunchdAgentsCurrent()` in `src/lib/launchd.ts` — the launchd analogue of `ensureSystemdUnitsCurrent`: rewrites any drifted plist (with a `.bak`), bootstraps any scoped agent that is not loaded (only `launchctl print` distinguishes "on disk but never loaded" from healthy, which is exactly the post-upgrade state), and only THEN boots out + deletes the retired labels — scoped agents live first, so a failure part-way can never leave a box with no daemon at all.

Wired into both unattended paths, not just install:
- heartbeat: the dispatch is now `healHandlerFor(platform)`, exported, with a darwin branch calling `defaultHealLaunchd`. `tests/cli-heartbeat.test.ts` asserts no platform phantombot installs units on maps to a no-op — that assertion is what would have caught the original macOS gap.
- `phantombot update`: new `healLaunchdAgents` step under `procPlatform === "darwin"`, same non-fatal contract as the systemd branch, so the swap is never undone by a launchctl hiccup.

Regression seeds the real shape you described — all three legacy agents loaded AND on disk, nothing under the scoped labels — against a `DomainLaunchctl` fake that actually models the gui domain, then runs the unattended path and asserts the scoped plists exist, are loaded, and every retired label is booted out and deleted, with no reinstall. Plus idempotence, drifted-plist rewrite, not-loaded bootstrap, and bootstrap-before-sweep ordering.

Mutation-checked: skipping the retired sweep reddens 2 tests; making the reload condition drift-only reddens the not-loaded test; removing the darwin case from `healHandlerFor` reddens the dispatch test. 3501 pass, tsc and the x64-baseline build clean.

src/lib/personaPaths.ts
robertclawson: Documented in 4ce35e2.

**README** gains a `### Moving the personas root` section: the bootstrap stub is the config file at the DEFAULT root (or `PHANTOMBOT_GLOBAL_CONFIG`) carrying `personas_dir`, the real global file still lives at `<custom-root>/config.toml`, precedence is `PHANTOMBOT_PERSONAS_DIR` > redirect > default root, a redirect equal to the default root is ignored so it cannot loop, and it is written by the migration rather than by hand — because the old host config that carried `personas_dir` gets archived. The paths table entry and the "two things outside" paragraph are reworded to two KINDS of thing accordingly. The upgrade paragraph now also states that the services rename themselves unattended (heartbeat, plus `update` after the swap) on all three platforms, which is now true rather than aspirational — see the other thread.

**AGENTS invariant 27** carries the same topology plus the reason the regression test spawns a real child process, and gains a sentence that renaming a service without an unattended reconcile is only half a rename, naming the three `ensure*Current` entry points, the unconditional launchd bootout (launchd keeps a job loaded after its plist is deleted), and the bootstrap-before-sweep ordering.

lenaparkhodges (dismissed):
Re-reviewed at 4ce35e2 — both round-4 Majors are fixed and verified. Approving.

**Major 1 (unattended macOS reconcile):** the darwin branch is no longer a no-op. `healHandlerFor()` dispatches `defaultHealLaunchd()` (guarded by `isPhantombotBinary` so dev runs never rewrite plists), and `phantombot update` runs `ensureLaunchdAgentsCurrent` right after the binary swap, non-fatal on failure with a manual-install hint. The implementation is the right shape: single-source `phantombotPlistTargets()` shared with `installPhantombotPlists` (so the fresh-install and reconcile lists can't drift), reload when drifted *or* not loaded (the `print` check is what distinguishes "healthy" from "correct plist never bootstrapped" — exactly the post-rename state), and retired agents swept only after the scoped ones are live. Mutation-tested: killing the darwin dispatch, hoisting the retired sweep before the targets loop, and dropping the `print` exit-code check each turn the covering tests red.

**Major 2 (root-redirect docs):** README gains a "Moving the personas root" section (bootstrap stub at the default root, precedence env > redirect > default, no-loop guard, written by migration not by hand), the paths table and "two kinds of thing" paragraph are reworded, and the upgrade paragraph now claims unattended service rename on all three platforms — which is true as of this commit, not aspirational. AGENTS invariant 27 carries the same topology plus the three `ensure*Current` entry points and the bootstrap-before-sweep ordering rationale.

Verification: full suite 3503 pass / 0 fail locally, tsc clean, CI green at 4ce35e2 including Windows and CodeQL. The new `DomainLaunchctl` test double models the gui domain properly (loaded vs never-loaded) rather than always exiting 0 — the earlier fake couldn't have caught this class of bug.

src/lib/launchd.ts
```diff
@@ -469,12 +610,99 @@ export async function ensurePlistCurrent(opts: {
...
+    // the domain — the post-rename upgrade case, where the plist we just wrote
+    // (or an install wrote) is correct on disk but nothing has ever loaded it.
+    const loaded = await opts.launchctl.run([
+      "print",
+      `${opts.domain}/${t.label}`,
+    ]);
+    if (!drifted && loaded.exitCode === 0) continue;
+    await opts.launchctl.run(["bootout", `${opts.domain}/${t.label}`]);
+    const r = await opts.launchctl.run(["bootstrap", opts.domain, t.path]);
+    if (r.exitCode === 0) reloaded.push(t.label);
```
kaieriksen: 🟠 **Major**: `bootstrap` failure is silently ignored here, then execution proceeds to `removeRetiredPlists()`. I reproduced all three scoped `bootstrap` calls returning exit 5: this function returned successfully with `reloaded: []` and deleted/booted out every legacy main, heartbeat, and tick job. That directly violates the claimed bootstrap-before-sweep safety property and can leave an upgraded Mac with no daemon at all. Treat a non-zero bootstrap as a reconcile failure and never sweep the legacy jobs until every required scoped replacement is confirmed loaded; add failure and mixed-success regressions.

Developer handoff:
```
Please fix the review finding on https://github.com/phantomyard/phantombot/pull/436 at src/lib/launchd.ts:691.

Issue:
A failed replacement bootstrap is ignored, and the working legacy jobs are still removed.

Expected change:
Surface the launchctl error and gate legacy cleanup on all required scoped agents being loaded successfully; avoid taking down a healthy scoped agent without a recoverable replacement path.

Validation:
Make bootstrap fail in DomainLaunchctl and assert legacy jobs/files remain; cover partial success too.
```

kaieriksen (changes requested):
## Re-review: persona-scoped layout

**Verdict:** Needs Changes

The durable custom-root redirect and unattended launchd wiring are fixed, but the launchd reconcile can still remove the only working legacy jobs after every replacement bootstrap fails.

Validation: 3,501 passed / 2 skipped / 0 failed; typecheck, x64-baseline build, and hosted Linux/Windows/CodeQL checks pass. The failure path below was reproduced directly against this head.

<details>

```
Please fix the review finding on https://github.com/phantomyard/phantombot/pull/436 at src/lib/launchd.ts:691.

Issue:
ensureLaunchdAgentsCurrent ignores a failed launchctl bootstrap and still removes every retired agent, so a transient or malformed-plist failure can leave an upgraded Mac with no daemon, heartbeat, or tick service.

Expected change:
Treat any required scoped-agent bootstrap failure as a reconcile failure and do not sweep legacy agents unless every required scoped replacement is confirmed loaded. Preserve enough prior state that a failed reload does not silently take a healthy scoped agent down.

Validation:
Add a DomainLaunchctl regression where bootstrap returns non-zero; assert the reconcile reports/throws the launchctl error and every working legacy job/file remains. Cover a mixed-success case so partial replacement cannot trigger the sweep.
```
</details>

src/lib/launchd.ts
robertclawson: Fixed in `2405021` — confirmed, and the sub-case you implied (bootout-then-failed-bootstrap on a *healthy* agent) was the worse half of it.

**The sweep is now conditional.** Every failed `bootstrap` is recorded in a new `failures` field carrying launchctl's own error, and `removeRetiredPlists` only runs when `failures.length === 0`. Leaving the pre-#435 agents loaded and on disk is the recoverable outcome — the next heartbeat retries — whereas sweeping them was terminal: no daemon, no heartbeat, no tick, and no plist left to recover from. I also corrected the comment above the sweep, which claimed the ordering alone guaranteed this; it never did.

**Drifted-but-healthy agents now roll back.** We `bootout` before writing the new body, so a failed reload previously left a running agent permanently down. On failure we restore the `.bak` and re-bootstrap the previous plist — stale-but-running beats down — and the label is dropped from `rewrote` so we don't claim a rewrite we reverted. It still counts as a failure, so it still blocks the sweep.

**No longer swallowed.** `heartbeat` logs a warning with the per-label errors; `update` writes each to stderr plus the manual `phantombot install` escape hatch.

The reason this slipped through is that the fake couldn't express it — a `launchctl` that always exits 0 has no failure mode — so `DomainLaunchctl` now takes a set of labels whose `bootstrap` fails. Four tests: total failure (retired agents still loaded *and* on disk, `bootout` never called on them), partial failure (also blocks — a half-migrated box must not lose its old daemon), the rollback path, and the happy path still sweeping so the guard can't over-block.

All three guards mutation-verified: making the sweep unconditional reddens 2 tests, dropping the rollback reddens 1, swallowing the failures reddens 2. Suite 3505 pass / 0 fail.

src/lib/launchd.ts
```diff
@@ -469,12 +610,148 @@ export async function ensurePlistCurrent(opts: {
...
+  }
+
+  // Only sweep the old identities once EVERY scoped agent is confirmed live.
+  // The ordering alone is not enough: bootstrap can fail (malformed plist,
+  // transient launchd error), and sweeping anyway would boot out and DELETE
+  // the only agents left running, leaving an upgraded Mac with no daemon, no
+  // heartbeat and no tick — and no plist to recover from. Leaving the retired
+  // agents in place is the recoverable failure: the next heartbeat retries.
+  const removedRetired =
+    failures.length === 0
```
kaieriksen: 🟠 **Major**: The all-or-nothing sweep guard leaves duplicate agents live after a partial failure. The new mixed-success test demonstrates it directly: main and heartbeat bootstrap successfully (`reloaded` contains both), `tick` fails, and then this branch keeps **all** retired labels loaded. That means the legacy daemon and the new persona-scoped daemon run concurrently until a later heal succeeds—the exact two-daemon race invariant 27 says this migration must eliminate—and both heartbeat generations can perform maintenance in parallel.

Expected change: make the transition failure-atomic. If any required replacement fails, boot out only the scoped agents newly started by this reconcile (while preserving/restoring scoped agents that were already healthy), or retire legacy agents role-by-role only after their matching scoped replacement is confirmed live. A partial failure must leave exactly one live agent per role, never zero and never two.

Validation: extend the mixed-success regression to inspect the launchctl domain and assert one loaded label per main/heartbeat/tick role after failure, including a case where a scoped agent was healthy before reconciliation and must not be removed.

Developer handoff:
```text
Please fix the review finding on https://github.com/phantomyard/phantombot/pull/436 at src/lib/launchd.ts:743.

Issue:
A partial bootstrap failure keeps every legacy job while leaving successful new scoped jobs loaded, so old and new daemon/heartbeat generations run concurrently.

Expected change:
Make failed reconciliation leave exactly one live agent per role. Roll back newly started scoped agents on any overall failure, or retire each legacy counterpart only after its scoped replacement is confirmed live; preserve scoped agents that were already healthy.

Validation:
Model partial bootstrap failure in DomainLaunchctl and assert exactly one loaded main, heartbeat, and tick identity afterward, with no duplicate old/new pair and no missing role.
```

kaieriksen (changes requested):
Re-review at exact head `2405021`: the total-failure outage is fixed, and drifted healthy agents now roll back. One fix-caused Major remains: the mixed-success path leaves successful scoped replacements loaded while retaining every corresponding legacy agent, so old and new daemon/heartbeat identities run together indefinitely.

Validation: focused launchd/update/heartbeat suites passed; full suite passed 3,505 with 2 platform skips and 0 failures; typecheck, x64-baseline build, and hosted Linux/Windows/CodeQL checks pass.

lenaparkhodges (dismissed):
Approved at 2405021 after independent verification.

**Mutation-tested all three guards on a local checkout:**
- Making the retired sweep unconditional → 2 tests red (total-failure + partial-failure cases both assert retired agents stay loaded and on disk).
- Dropping the rollback path () → 1 test red, the drifted-but-healthy agent stays down.
- Swallowing the failures (never recording) → 3 tests red, including the surfacing assertions.

**Full suite locally: 3507 pass / 0 fail, `tsc --noEmit` clean, CI green including Windows and CodeQL.**

The design call is right: stale-but-running beats down. The conditional sweep converts the previous terminal failure mode (delete the only running agents, leave no plist to recover from) into a recoverable one — next heartbeat retries. The rollback covers the nastier sub-case where a healthy agent was booted out for a body that won't load. And the comment fix on the sweep is honest — the old comment claimed ordering alone guaranteed safety, which it never did.

One observation, non-blocking: the heartbeat warning path returns early, so a failed reconcile skips the 'did change' logging entirely — fine, since the warning carries failures/rolledBack/reloaded already.

src/lib/launchd.ts
robertclawson: Confirmed and fixed in e8901e3 — you're right, and it was my own round-5 fix that introduced it. `failures.length === 0` traded "zero agents for a role" for "two agents for a role", which is the worse of the two: a single failed `tick` bootstrap left the legacy daemon and heartbeat loaded alongside their scoped replacements, i.e. two daemons and two heartbeats on one database — exactly what invariant 27 says this migration removes.

Retirement is now decided per ROLE via a new `legacyLabelReplacedBy()` map, and `removeRetiredPlists` takes a label subset. A legacy label is swept the moment the scoped agent replacing it is confirmed live, where live means any of: reloaded this pass, already correct-and-loaded and skipped by the `continue` path, or rolled back onto its previous body. The already-healthy case matters as much as the reloaded one — keying off `reloaded` alone would mean a second reconcile pass could never finish a sweep the first one earned, stranding the legacy agents forever. A rolled-back agent counts too: it is stale but it IS running that role, so leaving its twin loaded is the same race.

`dev.phantombot.nightly` keeps the conservative fully-converged gate, since nothing scoped replaces it.

Tests: the partial-failure case now asserts the invariant directly against the launchctl domain — exactly one loaded label per role (scoped for daemon/heartbeat, legacy for the failed tick) plus the matching on-disk state; a new case covers the already-healthy-from-an-earlier-pass agent still retiring its twin; the rollback case asserts the legacy daemon is booted out while the nightly is not. All five guards mutation-verified: coarse gate reddens 3, dropping the already-healthy path reddens 1, not counting a rollback reddens 1, unconditional nightly reddens 3, ignoring the subset reddens 3. Suite 3506 pass / 0 fail.

lenaparkhodges (approved):
Verified e8901e3 against the per-role retirement finding: legacyLabelReplacedBy() maps each scoped label to its legacy twin, removeRetiredPlists() takes a label subset, and the reconcile now retires each legacy agent exactly when its replacement is confirmed live — counting the reloaded, already-healthy-skip, and rolled-back-but-running paths. That covers all three ways a scoped agent can be live, which kills both failure modes: zero agents for a role (unconditional sweep) and two agents for a role (the all-or-nothing gate). The nightly keeping the fully-converged gate is right since nothing scoped replaces it.

Independently verified: full suite 3506 pass / 0 fail locally, tsc clean, CI green including test-windows and CodeQL. The new invariant test asserting exactly one loaded label per role against the domain is the right shape — it checks the outcome, not the mechanism. The already-healthy-from-an-earlier-pass test closes the stranded-twin gap that keying off `reloaded` alone would have left. No blocking findings.

kaieriksen (approved):
Re-reviewed exact head e8901e3284f06d2e69f39e9ff4d4dda5eae79730. The per-role launchd retirement fix resolves the partial-failure duplicate-agent Major: newly live, already-healthy, and rolled-back scoped agents each retire only their matching legacy role, while failed roles retain their legacy agent and nightly remains until full convergence. Validation: focused launchd/update/heartbeat tests 79/79; full suite 3,506 passed, 2 skipped, 0 failed; typecheck, linux x64-baseline build, Windows CI, Linux CI, and CodeQL all green.

PR #438, opened 2026-08-23
Revert "Scope everything to the persona directory (#435) (#436)"

robertclawson:
Reverts #436 (squash `742f80a`). The tree after this commit is **byte-identical to `v1.1.293`** (`33a479f`) — verified with `git diff v1.1.293` returning empty.

## Why

The persona-scoped layout is correct in isolation, but the *upgrade path* around it mutates host state in an order that can leave a phantom silent:

- **Linux, cron `update --force --restart`:** the heal sweeps the legacy `phantombot.service` (stop → disable → unlink) and writes the scoped unit, but only ever runs `enable --now` on the **timers**. The new daemon unit is started and never enabled — the phantom survives until the next reboot, then comes up with nothing enabled.
- **Linux, interactive `update` with the restart prompt declined:** the legacy unit is already stopped and deleted, the new one is written but never started. Silent immediately, with only a "restart with: …" hint printed.

macOS is fine (the launchd heal bootstraps the scoped agents itself), but the Linux exposure is enough to pull the whole change back rather than patch it in place.

## What lands instead

Agreed design, to be built before the layout change returns:

1. On daemon start, **before any file is touched**, spawn a **detached** doctor in its own transient unit (`systemd-run --user`) — not a child of the daemon's cgroup, or it dies with the restart (see #415).
2. The doctor **gates first**: compares current layout against expected and exits silently with zero mutations when converged, so it is safe on every start and genuinely idempotent.
3. If migration is needed, it mutates behind the undo journal, then runs the installer to regenerate and **enable** the units and timers.
4. It **verifies health** — units enabled *and* active, lock path writable, DB opens, heartbeat and tick fire — and only then restarts.
5. On failed verify it rolls back and leaves the previous daemon running, and surfaces the failure loudly.

Separately, units should become **layout-agnostic** (`ExecStart=phantombot run --persona <p>` and nothing else, with all paths resolved at runtime), so a future layout change is a code change and never requires rewriting a unit file.

## Exposure

`v1.1.294` was only ever cut as a **prerelease** and has not been promoted, so no stable host has taken this. Reverting now costs nothing.

## Verification

- `bun test`: **3430 pass / 0 fail** (2 skip).
- `git diff v1.1.293` on this branch: empty.


kaieriksen (approved):
Verified exact revert at `4ab63c6`: its sole parent is the PR #436 merge commit `742f80a`, and its resulting Git tree (`95f2f04`) is byte-for-byte identical to pre-#436 commit `33a479f`. No extra changes.

PR #440, opened 2026-08-23
Multi-persona autostart: persona-scoped config in one process (#439)

### LINKED ISSUE #439: Multi-persona autostart: persona-scoped config, single process
## Summary

Let one user account autostart **several personas**, and move persona-scoped
settings out of the global `~/.config/phantombot/config.toml` into each
persona's own directory.

Deliberately **no new processes**: the daemon already runs every configured
persona concurrently. This issue is config parsing plus a loader loop.

## Current behaviour (verified on `main` @ 9f9f9df)

- `runRun()` → `planListeners()` builds one Telegram listener for
  `[channels.telegram]` bound to `default_persona`, plus one for each
  `[channels.telegram.personas.<name>]` block. All in a single process.
- PhantomChat listeners are started for every persona that has a
  `<persona-dir>/phantomchat.json` — already per-persona config, no
  `config.toml` block at all.
- A duplicate-token guard already fails fast when two personas share one
  Telegram bot.
- There is no `phantombot run --persona <name>` flag, and one process is the
  intended shape.

So concurrency exists. What's missing is (a) an explicit autostart list and
(b) persona config living with the persona.

### 1. Global config keeps globals only

`~/.config/phantombot/config.toml`:

```toml
default_persona = "robbie"
autostart_personas = ["kai", "lena"]
update_channel = "preview"
personas_dir = "..."
memory_db = "..."

[harnesses]         # host-wide binaries/models
```

### 2. Persona-scoped settings move to `<personas-root>/<persona>/config.toml`

Moved keys:

- `[channels.telegram]` / `[channels.telegram.personas.<name>]` → the persona's
  own `[channels.telegram]` (token, allowed users, streaming settings)
- `harnesses.personas.<name>.chain` → the persona's own `harnesses.chain`
  override
- `chattiness`, per-persona `voice` settings

This follows the precedent PhantomChat already set with `phantomchat.json`.

### 3. `autostart_personas` drives the loader

At boot, `planListeners()` reads each persona named in `autostart_personas`
(plus `default_persona`), loads that persona's own config, and builds its
listeners exactly as it does today. Existing guards (missing persona dir →
warn and skip; duplicate token → fatal) stay unchanged.

### 4. Lifecycle commands gated to the default persona

`/update` and `/restart` are single-process operations. They must only be
accepted from the default persona's channels — a non-default persona invoking
them would restart every persona on the host. One ownership check, no new
machinery.

### 5. Persona TUI

`phantombot persona` gains the ability to mark a persona as autostart, writing
the entry to `autostart_personas`.

## Explicitly rejected: subprocess / subagent supervision

The alternative was for the default persona to spawn
`phantombot run --persona <name>` children and supervise them. Rejected: it
buys crash isolation between personas and costs a supervisor with backoff and
crash-loop detection, orphan handling on `/update` (children must stop before
the binary swap or keep running the old binary), shutdown ordering on
`/restart`, and per-child transient `systemd-run --user` units to avoid the
cgroup-teardown kill from #415.

That is new risk on the update path for every existing host. #436 was reverted
in #438 for exactly that class of problem (a daemon left
started-but-never-enabled, silent after reboot). The config-only version
touches none of that machinery.

## Backward compatibility

If a persona has no own `config.toml`, fall back to the global file's
`[channels.telegram.personas.<name>]` block. An unmigrated host keeps working
unchanged, so no host can be stranded mid-upgrade.

## Open scope question

Two candidate first PRs:

- **(a)** full split: config schema, per-persona loader, migration of existing
  keys out of the global file, TUI change, tests.
- **(b)** loader + `autostart_personas` only, global file still read as
  fallback; migration lands as a second PR.

(b) is smaller and cannot strand an unmigrated host. To be settled before
implementation starts.

## Acceptance criteria

- [ ] `autostart_personas` in the global config starts listeners for each named persona
- [ ] A persona's `<persona-dir>/config.toml` supplies its Telegram account, harness chain override and chattiness
- [ ] A host with no per-persona config files behaves exactly as before
- [ ] `/update` and `/restart` are refused from a non-default persona with a clear message. `/stop` is NOT gated: it aborts only the current turn in the current chat, so it is per-chat rather than process-wide and gating it would leave a non-default persona unable to interrupt itself (agreed in review of #440).
- [ ] `phantombot persona` can set/unset autostart for a persona
- [ ] No new processes are spawned; `run --persona` is NOT introduced

robertclawson:
Closes #439.

## What

Start more than one persona at boot — without adding a single process.

phantombot already runs every persona concurrently inside one daemon: one listener per persona, one binary, one update. What was missing was a place for a persona to keep its **own** settings, which is why "start these three at boot" looked like a supervision problem. It isn't; it's config.

```
<config-home>/phantombot/config.toml    host globals: default_persona,
                                        autostart_personas, update_channel,
                                        paths, harness bins
<personas-root>/<persona>/config.toml   that persona's own settings
```

Resolution is a **per-key deep merge**, persona file wins every conflict, env vars still win over both. A key the persona file omits falls back to the **global file**, never to a constant — a "default" that outranks an operator's existing setting is a silent behaviour change on upgrade. Host-level keys (`default_persona`, `autostart_personas`, `update_channel`, `personas_dir`, `memory_db`) are stripped from the persona layer, so a persona cannot elect itself default or relocate the personas root.

## Why copy-not-delete

Migration seeds `<persona>/config.toml` from the global file and **never deletes from it**. The presence of the persona file is the entire idempotency guard, so migration runs on every start and costs one stat after the first time.

That is what makes `/update` order-independent: a host arriving from **any** older version lands correct in one update, with no "upgrade to X before Y" dance, and updating twice changes nothing. Release rings mean a host can also be rolled **back** to a binary that has never heard of persona config files — with the global file untouched, that host still boots with its channels intact. Users who want a tidy global file can prune it by hand; nothing reads those keys once a persona file exists.

## Also in here

- **`autostart_personas`** — explicit list; the default persona is always started; absent/empty is exactly today's behaviour. Never inferred from what happens to be on disk: an imported or restored persona must not start talking to the world just because its directory exists.
- **`planListeners`** builds a listener per autostart persona from that persona's own config, and skips a legacy `[channels.telegram.personas.<name>]` entry for a persona that already got one. Without that skip, copy-not-delete describes one bot in two places and the duplicate-token guard takes Telegram down on every migrated multi-persona host.
- **`/update` and `/restart` are default-persona-only.** They swap the binary and bounce the service for *everyone* in the process. They're answered, not silently ignored, and the refusal names who to ask. `/stop` is deliberately **not** gated — it aborts this chat's own turn and harms nobody.
- **`/status`** gains the release ring (`channel: preview`) and the persona roster, marking the default and which persona is answering — so the preview cohort is visible from any chat.
- **`--persona` on `voice` and `task`.** Voice is persona-scoped and now writes to the persona's file. An unknown `--persona` on `task` is refused rather than accepted: a task filed against a typo can never fire, and a scheduler that silently loses a reminder is worse than one that errors.
- **`phantombot persona`** gains a boot-autostart picker.

## Explicitly not done

Subprocess or subagent supervision. It buys only crash isolation, and costs a supervisor, backoff, orphan handling on update, shutdown ordering, and the #415 cgroup trap — on the same update path that caused the #436/#438 revert. Rejected in #439 before any code was written.

## Compatibility

- No persona file → identical behaviour to before, key for key.
- `[channels.telegram.personas.<name>]` and `[harnesses.personas.<name>]` still read and still work.
- No forced migration, no user action required, no upgrade ordering.

## Verification

`bun tsc --noEmit` clean; **3470 pass / 0 fail** (+40 new tests).

Covered: per-key merge and array-replacement semantics, host-key stripping, migration seeding / idempotency / global-file-untouched, layered `loadConfig`, the autostart and dedupe paths in `planListeners`, the lifecycle gate, `/status`, `task --persona`, and the picker.

Each new guard was **mutation-verified** — reverting it turns the covering test red:

| Guard reverted | Result |
|---|---|
| `stripHostOnlyKeys` filter | 1 fail |
| migration "already exists" check | 2 fail |
| legacy/autostart listener dedupe | 1 fail |
| lifecycle persona gate | 2 fail |


README.md
```diff
@@ -593,6 +615,49 @@ every `<TAB>`, so it always matches the available subcommands and flags —
...
+
+[channels.telegram]
+token = "222:lena-bot-token"
+allowed_user_ids = [123456789]
+
+[voice]
+provider = "elevenlabs"
+```
+
+The per-persona TUIs write there for you: `phantombot telegram --persona lena`,
```
kaieriksen, line 656: **Major** — This is not true for Telegram, and the mismatch makes the documented update path silently ineffective after migration. `runTelegram` still calls `applyTelegramConfig(config.configPath, ..., persona)`, which writes `[channels.telegram.personas.<name>]` in the global file. Once migration has copied that account into `<persona>/config.toml`, the persona layer wins and `planListeners` deliberately skips the legacy entry, so changing a token or allowlist through `phantombot telegram --persona lena` leaves the old live values in place after restart. The harness picker has the same split-brain for migrated `[harnesses].chain`. Please move these per-persona writers to the persona file or provide a single write resolver that matches the read precedence.

src/cli/task.ts
```diff
@@ -202,7 +229,7 @@ export async function runTaskAdd(input: RunTaskAddInput): Promise<number> {
     }
 
     const result = store.add({
-      persona: config.defaultPersona,
+      persona,
```
kaieriksen, line 232: **Major** — Storing the requested persona on the row is only half of this path. `tick.ts` still calls `loadConfig()` once for the default persona and uses that object for every due task: `buildHarnessChain(config, ..., task.persona)`, retriever/fact settings, and timeouts all come from the default/global layer. A task added with `--persona lena` therefore ignores `<lena>/config.toml` and can run the wrong harness/retrieval policy. Resolve the effective config with `loadConfig(task.persona)` per due persona, while retaining the host-level binary paths as the daemon does, and cover the fire path rather than only the stored row.

src/channels/commands.ts
```diff
@@ -765,10 +809,22 @@ async function handleStatus(
...
 
+  // Release ring this HOST follows (#432). Host-level, not per-persona: every
+  // persona on a box runs the same binary, so one line, no repetition.
+  const channelLine = `channel: ${ctx.config?.updateChannel ?? DEFAULT_UPDATE_CHANNEL}\n`;
+
+  // Who else is running in this process (phantombot#439). The point is the
+  // "who is in the preview cohort?" question: from any one persona's chat you
+  // can see the whole set the daemon started, which persona is the default
+  // (the one that owns /update and /restart), and which is answering here.
+  const personaLine = formatPersonaRoster(ctx);
```
kaieriksen, line 830: **Minor** — This roster is not the set of personas the daemon actually started. `runRun` still starts every legacy `[channels.telegram.personas.*]` listener and every persona returned by `listPhantomchatPersonas`, even when neither appears in `autostartPersonas`; this formatter omits all of them. `/status` can therefore report a single-persona process while several personas are live. Pass the planned/running persona set into the command context instead of reconstructing it from only these two config fields.

tests/lib-persona-config.test.ts
```diff
@@ -0,0 +1,315 @@
...
+
+describe("loadConfig persona layering", () => {
+  const SAVED = {
+    config: process.env.PHANTOMBOT_CONFIG,
+    personas: process.env.PHANTOMBOT_PERSONAS_DIR,
+    state: process.env.PHANTOMBOT_STATE,
+    persona: process.env.PHANTOMBOT_DEFAULT_PERSONA,
+  };
+
+  beforeEach(() => {
```
kaieriksen, line 286: **Minor** — These new config tests do not isolate env vars that outrank TOML. On a real configured host, the three layering assertions fail because `TELEGRAM_BOT_TOKEN` remains set; I reproduced 3 failures in this file while the same assertions pass in a clean CI environment. Save/delete/restore every env key exercised by the assertions, at minimum `TELEGRAM_BOT_TOKEN` and the relevant autostart override, so the tests verify layering rather than the caller environment.

README.md
```diff
@@ -593,6 +615,49 @@ every `<TAB>`, so it always matches the available subcommands and flags —
 `phantombot p<TAB>` → `persona`, `phantomchat`, `p2p`; `phantombot p2p <TAB>` →
 `status`; `phantombot logs --<TAB>` → `--follow`, `--no-follow`, `--lines`.
 
+## Multiple personas
+
+One phantombot process serves every persona on the host. There is no supervisor
```
kaieriksen, line 620: **Major** — The contributor contract still says `One persona is active at a time` in `AGENTS.md`, and that file explicitly requires itself to be updated for architecture changes. This PR changes that invariant but leaves the old section byte-for-byte intact, so future work will be guided by the opposite runtime model. Update the persona-model section and module/service descriptions in `AGENTS.md` in this PR.

kaieriksen (changes requested):
Requesting changes. The config layering itself is coherent, but the write paths and non-daemon consumers are not migrated with it, so common persona operations silently run stale/default settings. I also found inaccurate runtime reporting and a non-hermetic new test suite.

src/cli/run.ts
```diff
@@ -319,10 +474,19 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
+    // A listener resolves its harness chain from its OWN config when it has
+    // one, so a persona can set `[harnesses].chain` in its own file. The bins
+    // stay host-level — they name binaries installed on this machine, and were
+    // just probed for real above — so they are carried over from the resolved
+    // global config rather than re-probed per persona.
+    const listenerConfig = withHostHarnessBins(listener.config ?? config, config);
+    return {
+      ...listener,
+      config: listenerConfig,
+      harnesses: buildHarnessChain(listenerConfig, err, listener.persona),
```
lenaparkhodges, line 561: This is where the Major bites: `harnessChainIds(listenerConfig, persona)` resolves the legacy `harnesses.personas.<persona>.chain` table BEFORE the persona's own `harnesses.chain`. The persona layer deep-merges the global file, so that legacy table is present in `listenerConfig` on every migrated host — and since migration never deletes it (correctly, for rollback safety), the persona's own `[harnesses].chain` can never win there.

Repro on this branch:

```ts
const merged = mergeToml(
  { harnesses: { chain: ["claude"], personas: { lena: { chain: ["codex"] } } } },
  { harnesses: { chain: ["pi"] } },
);
harnessChainIds({ harnesses: merged.harnesses } as any, "lena"); // ["codex"], not ["pi"]
```

So: a user edits their persona file's chain to `["pi"]` and nothing changes — silently, because the stale legacy entry in the global file outranks it. Same applies to the phantomchat path at ~L864.

Suggested fix: strip the persona's OWN entry from `harnesses.personas` when the persona layer is merged in `loadConfig(persona)` (the entry describes the old layout; the persona's own file is its replacement), or make `harnessChainIds` prefer `harnesses.chain` when `config.personaLayer === persona`. A regression test should cover the both-present state — legacy entry in global AND chain in the persona file — which is exactly what every migrated host has.

tests/lib-persona-config.test.ts
```diff
@@ -0,0 +1,315 @@
...
+      globalToml: { default_persona: "robbie" },
+      isDefault: true,
+    });
+    expect(r).toEqual({ migrated: false, reason: "nothing-to-copy" });
+    expect(await readPersonaToml(personasDir, "robbie")).toEqual({});
+  });
+});
+
+describe("loadConfig persona layering", () => {
+  const SAVED = {
```
lenaparkhodges: The SAVED/restore scrub covers `PHANTOMBOT_CONFIG` / `_PERSONAS_DIR` / `_STATE` / `_DEFAULT_PERSONA` but not `PHANTOMBOT_PERSONA`. On any box where phantombot is running (or where tests are spawned from a harness turn — `withPersonaEnv` sets `PHANTOMBOT_PERSONA` on every child), these three tests fail by reading a real vault-injected Telegram token instead of the fixture. Verified: `PHANTOMBOT_PERSONA=lena bun test tests/lib-persona-config.test.ts` → 3 fail; scrubbed env → 0 fail.

Add `PHANTOMBOT_PERSONA` to the saved/restore list and delete it in `beforeEach`, same as `_DEFAULT_PERSONA`.

src/channels/commands.ts
```diff
@@ -294,6 +302,42 @@ export async function handleSlashCommand(
   }
 }
 
+/**
```
lenaparkhodges, line 315: Small scope note: #439's acceptance criteria list `/stop` among the commands refused on a non-default persona, but this PR deliberately leaves it ungated (rationale in the PR body — it aborts only this chat's own turn, which is per-chat and harms nobody). I think the ungated call is right; just make sure #439's checklist is updated to match so the AC isn't silently diverging from what shipped.

lenaparkhodges (changes requested):
Reviewed on a real checkout: `tsc --noEmit` clean, full suite run (all 14 local failures traced to ambient env or pre-existing on main — details in the inline thread), and the layering design read end-to-end.

The design is right: config-only, copy-not-delete migration, host-key stripping, and no supervision machinery — it avoids exactly the #436/#438 class of problem. One Major blocks it for me, plus two smaller notes.

**Major — persona's own `[harnesses].chain` is dead config on a migrated host.** `harnessChainIds()` still checks the legacy `[harnesses.personas.<name>].chain` table FIRST, and that table survives in the global file forever (copy-not-delete). So after migration seeds `harnesses.chain` into the persona file, the persona layer merges the global `harnesses.personas` table back in, and the legacy entry silently outranks the persona's own file — the exact "silent behaviour change on upgrade" this PR exists to eliminate. Reproduced against this branch's code:

```ts
const merged = mergeToml(
  { harnesses: { chain: ["claude"], personas: { lena: { chain: ["codex"] } } } },
  { harnesses: { chain: ["pi"] } },   // lena's own config.toml
);
harnessChainIds({ harnesses: merged.harnesses } as any, "lena");  // => ["codex"], not ["pi"]
```

The covering test (`passes each listener its effective persona harness chain`) doesn't hit this because its fixture gives amanda a legacy entry but no autostart config, and phantom no legacy entry. A test with BOTH (the post-migration state) goes red. Cleanest fix IMO: when building a chain for a persona whose config has `personaLayer === persona`, the persona's own `harnesses.chain` must win over the legacy table — e.g. strip the persona's own entry out of `harnesses.personas` when the persona layer is merged, or invert the precedence in `harnessChainIds` when the config's `personaLayer` matches.

**Minor — new loadConfig tests break on any box running phantombot.** `PHANTOMBOT_PERSONA` is ambient in harness-spawned environments (`withPersonaEnv` sets it for every child), and with it set the three layering tests read a real vault-injected Telegram token instead of the fixture. The scrub list in `tests/lib-persona-config.test.ts` covers `PHANTOMBOT_CONFIG` / `_PERSONAS_DIR` / `_STATE` / `_DEFAULT_PERSONA` — add `PHANTOMBOT_PERSONA`. CI is clean so this hides, but it's the ambient-env-var isolation bug the repo has hit repeatedly.

**Note — `/stop` gate deviates from #439's acceptance criteria.** The issue says `/update`, `/restart`, `/stop` are all refused on a non-default persona; the PR deliberately leaves `/stop` ungated with a reasonable rationale (aborts only this chat's own turn). If that's the agreed call, update the #439 checklist so acceptance is unambiguous rather than silently diverging.

Everything else — migration idempotency and the "never writes an empty file" guard, host-key stripping, the autostart/legacy listener dedupe, the lifecycle gate, `task --persona` existence check, the autostart picker writing only the host file — all check out against the code.

src/cli/run.ts
```diff
@@ -260,7 +362,60 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
+  const loadPersonaConfig =
+    input.loadPersonaConfig ?? ((name: string) => loadConfig(name));
+  const autostart = (config.autostartPersonas ?? []).filter(
+    (name) => name !== defaultPersona,
+  );
+  try {
+    const globalToml = await readConfigToml(config.configPath);
+    const migrateNames = [
+      defaultPersona,
+      ...autostart,
```
kaieriksen, line 415: 🟠 **Major**: `autostart_personas` only controls this config-loading/Telegram path. `phantomchatPersonas` was already computed by `listPhantomchatPersonas(config)`, which scans every persona directory, and the later loop still starts every returned identity. An imported or restored persona containing `phantomchat.json` therefore starts talking even when absent from the explicit list; because it is also absent from `personaConfigs`, it runs with the default persona's effective settings. This directly contradicts the stated safety property that merely existing on disk never starts a persona. Filter PhantomChat to the default + autostart roster before the early channel guards and load each selected persona's config; add a regression with a configured but unlisted identity.

src/lib/personaConfig.ts
```diff
@@ -0,0 +1,250 @@
...
+ * stat after the first time. It never writes an empty file, so a host with no
+ * global config stays clean rather than growing an empty one per persona.
+ */
+export async function migratePersonaConfig(
+  input: MigratePersonaConfigInput,
+): Promise<MigratePersonaConfigResult> {
+  const path = personaConfigPath(input.personasDir, input.persona);
+  const existing = await readConfigToml(path);
+  if (Object.keys(existing).length > 0) {
+    return { migrated: false, reason: "exists" };
```
kaieriksen: 🟠 **Major**: Treating any non-empty persona file as fully migrated makes upgrade order unsafe. Repro: legacy global config has the default bot plus `[channels.telegram.personas.lena]`; before the first new daemon start, `voice --persona lena` creates `<lena>/config.toml` containing only `[voice]`. This guard skips Lena's Telegram translation, `loadConfig("lena")` inherits the global default `[channels.telegram]`, and `planListeners` sees the default token for both listeners and fatals (or disables Telegram when PhantomChat exists). Preserve existing keys but seed missing migrated keys, or explicitly adapt the legacy routing table whenever the persona layer lacks its own account. Cover this exact voice-first/restart-second path.

src/config.ts
```diff
@@ -743,16 +771,51 @@ export function xdgStateHome(): string {
...
+ * Env vars are applied after both and still win over everything, so
+ * `PHANTOMBOT_*` remains the top of the precedence order it has always been.
+ *
+ * A key the persona file does not mention falls back to the GLOBAL FILE, not
+ * to a built-in default. An unmigrated host therefore behaves exactly as it
+ * did before this existed: no persona file, empty layer, identical config.
+ *
+ * @param persona persona whose layer to apply. Omit for the default persona.
+ */
+export async function loadConfig(persona?: string): Promise<Config> {
```
kaieriksen, line 898: 🟠 **Major**: Making persona layering opt-in leaves existing persona-aware one-shot/background callers on the default layer. `runNotify({ persona: "lena" })`, `runAsk --persona lena`, `runNightly`, and `runHeartbeatCli` all call `loadConfig()` first and resolve the target afterward. After the documented cleanup of legacy global persona keys, notify cannot see Lena's local Telegram account (it falls back to the default bot or reports no channel), while ask/nightly/heartbeat use the default harness, voice, retrieval, and timeout policy. Resolve the target before loading config at every persona-aware entry point, while retaining host-only bins/paths, and add end-to-end caller tests rather than only loader tests.

src/cli/voice.ts
```diff
@@ -84,7 +92,15 @@ interface RunInput {
 }
 
 export async function runVoice(input: RunInput = {}): Promise<number> {
-  const config = input.config ?? (await loadConfig());
+  // Load the TARGET persona's layered config, so "Existing config" shows what
+  // that persona actually runs with rather than the default persona's voice.
+  const config = input.config ?? (await loadConfig(input.persona));
```
kaieriksen, line 103: 🟠 **Major**: An explicit unknown persona is never validated. `loadConfig("typo")` treats a missing persona file as an empty layer, and the later `updateConfigToml(voiceConfigPath, ...)` creates `<personas-root>/typo/config.toml` plus its directory, reports success, saves the provider credential, and may restart the service even though no such persona existed. `task --persona` correctly rejects this class of silent loss; voice needs the same pre-write existence check and a regression proving no directory, config, env key, or restart is produced.

kaieriksen (changes requested):
## Additional review findings

**Verdict:** Needs Changes

I found four additional blocking gaps beyond the existing review threads: PhantomChat bypasses the explicit boot roster, partial persona files can defeat migration and take Telegram down, persona-aware one-shot commands still load the default layer, and `voice --persona` silently creates typo personas.

<details>

```
Please fix these additional findings on phantomyard/phantombot PR #440 at e712a7459597087b1714cb597f5d96b1322bd369:

1. [Major] Make PhantomChat honor default_persona + autostart_personas and load each selected persona's effective config; prove unlisted imported/restored identities do not start.
2. [Major] Make legacy migration safe when a persona config already contains only some keys; preserve hand edits while translating the legacy persona Telegram account, and cover the voice-before-first-restart upgrade order.
3. [Major] Update persona-aware one-shot/background consumers (notify, ask, nightly, heartbeat, and related paths) to resolve the named persona's effective layer while preserving host-only binary/path settings.
4. [Major] Refuse an unknown `voice --persona` target before any config/env write or restart.

Run the focused config/listener/voice/notify/ask/nightly/heartbeat tests, typecheck, and the full suite in an isolated environment.
```
</details>

robertclawson:
All ten findings fixed at c71261b — thanks both, the two reviews met in the middle: the config *layer* was right, everything that reads or writes around it still assumed one persona.

**Kai**

1. **PhantomChat bypasses the boot roster** — fixed. `selectPhantomchatPersonas()` gates the disk scan on `default_persona` + `autostart_personas` before the channel guards, and each selected persona is migrated and gets its own effective config. The gate is deliberately **opt-in**: `autostart_personas` absent = exactly the pre-#439 behaviour (otherwise an upgrade silently drops a live channel), present-even-if-empty = the whole truth. Regression covers both, plus the configured-but-unlisted identity.
2. **Partial persona file defeats migration** — fixed. Migration is additive now: `seedMissing()` fills only the keys the file omits and never touches a value that is there. Your exact repro is a test (`[voice]`-only file written by `voice --persona`, legacy Telegram entry translated anyway). The other half of that repro is fixed in the loader: a non-default persona no longer inherits the global `[channels.telegram]`, so it can never end up sharing the default bot's token.
3. **One-shot/background consumers on the default layer** — fixed. `loadConfigForPersona(persona)` resolves the target first, then loads its layer, keeping host harness bins via `withHostHarnessBins`. Applied to `notify`, `ask`, `nightly`, `heartbeat`, and `tick` (per due persona, cached, resolved after bin probing). Caller-level tests, not loader tests: `notify --persona lena` sends on lena's bot from a real fixture on disk.
4. **`voice --persona` typo** — fixed, refused before any write, with a test asserting no directory, no config, no restart.
5. **Telegram/harness writers vs read precedence** — fixed with a single resolver, `resolvePersonaWriteTarget()`: persona file once it exists, global file in its historical shape until then (rollback-safe, and the next daemon start copies it up). Both TUIs also now *read* through the persona's effective config, so "Existing config" shows what is actually live.
6. **`tick` uses the default config for every task** — fixed, see 3. Test asserts the fire path, not the stored row.
7. **`/status` roster** — fixed. `runRun` passes the personas it really started (Telegram listeners ∪ PhantomChat identities); the config-derived list stays only as a fallback for embedded callers.
8. **`AGENTS.md` still said "one persona is active at a time"** — rewritten: persona model, the two config layers, and invariant 27 covering the three rules that keep getting broken (resolve-before-load, write-where-read, existing-on-disk-never-starts).
9. **Non-hermetic tests** — fixed, see Lena below.

**Lena**

- **`[harnesses].chain` dead on a migrated host** — confirmed and fixed, and your repro is now a test. `loadConfig(persona)` drops that persona's OWN entry from the merged `harnesses.personas` table *when the persona file states a chain*; conditioning it on the persona file means an unmigrated host keeps the legacy entry and behaves exactly as before (also a test). Both the Telegram and the phantomchat path get it, since it is fixed in the loader rather than at either call site.
- **`PHANTOMBOT_PERSONA` scrub** — fixed, and widened: the suite now saves/deletes/restores every env key that outranks TOML in these assertions, `TELEGRAM_BOT_TOKEN` included. Verified green with `PHANTOMBOT_PERSONA=lena TELEGRAM_BOT_TOKEN=999:ambient bun test`.
- **`/stop` vs #439's AC** — agreed; the issue checklist is the thing that is wrong, not the code. Updating #439 so the AC reads `/update` and `/restart` only, with the per-chat rationale recorded there.

3486 tests pass, `tsc --noEmit` clean. Every new guard was mutation-verified individually — revert the guard, watch the specific test go red, restore.

lenaparkhodges (dismissed):
Re-review at c71261b — both my round-1 findings verified fixed, spot-checked Kai's.

**[harnesses].chain dead on a migrated host (Major)** — fixed at the right level: `applyPersonaLayer` drops the persona's own entry from the merged legacy `harnesses.personas` table only when the persona file states a chain, so an unmigrated host keeps the legacy entry and behaves exactly as before. Loader-level fix means both the Telegram and PhantomChat paths get it. Both behaviors are direct tests (`a persona's own harness chain beats the LEGACY personas table` / `an UNMIGRATED persona keeps its legacy harness chain`), and the cross-persona assertion (lena's layer doesn't touch robbie's legacy entry) is a nice touch.

**Ambient PHANTOMBOT_PERSONA in loadConfig tests (Minor)** — fixed and widened correctly: every env key that outranks TOML is saved/deleted/restored, `TELEGRAM_BOT_TOKEN` included. Verified the persona-config suite green with `PHANTOMBOT_PERSONA=lena TELEGRAM_BOT_TOKEN=999:ambient` ambient.

Also spot-checked `loadConfigForPersona` reaching ask/notify/nightly/heartbeat, and tick's per-persona resolution — the cache-raw/bins-lazy split (a persona resolved pre-bin-probe must not freeze unresolved bins) is documented and correct, and read-failure degrading to the host layer rather than dropping the wake is the right failure mode.

Verification: 3484 pass / 2 skip / 3 fail locally under forced ambient env — all 3 (PiHarness routing) fail identically on main in the same env, so ambient/pre-existing, not PR-caused. `tsc --noEmit` clean. CI green at c71261b including Windows and CodeQL. PR is up to date with main including the #436 revert (#438).

Merge-ready from my side.

src/config.ts
```diff
@@ -743,16 +772,144 @@ export function xdgStateHome(): string {
...
+  }
+
+  if (!opts.isDefault) {
+    const channels = merged.channels;
+    if (isTomlTable(channels)) {
+      const personaChannels = personaToml.channels;
+      const ownTelegram = isTomlTable(personaChannels)
+        ? personaChannels.telegram
+        : undefined;
+      if (ownTelegram === undefined) {
```
kaieriksen: 🟠 **Major**: This guard still leaks the default persona's bot whenever the non-default persona has a *partial* `[channels.telegram]` table. `mergeToml` has already combined the global account with the persona table, and `ownTelegram !== undefined` skips the isolation rewrite, so a file containing only `allowed_user_ids = [2]` resolves to the default token plus Lena's allowlist. I reproduced `applyPersonaLayer({channels:{telegram:{token:"DEFAULT"}}}, {channels:{telegram:{allowed_user_ids:[2]}}}, {persona:"lena",isDefault:false})` returning token `DEFAULT`. Build the non-default account from its legacy entry/persona table without first inheriting the global account, and cover a partial account with no persona token.

src/cli/run.ts
```diff
@@ -260,7 +379,88 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
+    log.warn("run: persona config migration threw", {
+      error: (e as Error).message,
+    });
+  }
+  // Every persona this process will actually run needs its OWN effective
+  // config — not just the Telegram autostart list. A PhantomChat persona left
+  // out of this map falls back to `config`, i.e. the DEFAULT persona's voice,
+  // chattiness and harness chain, which is precisely the silent mis-run #439
+  // exists to remove.
+  const resolveNames = new Set<string>([
```
kaieriksen, line 452: 🟠 **Major**: Legacy Telegram personas are migrated above but never added to `resolveNames`. A host that still relies on `[channels.telegram.personas.lena]` (with no `autostart_personas` entry and no PhantomChat identity) gets a listener whose `config` is undefined; the later `listener.config ?? config` fallback runs Lena on the default persona's harness, voice, chattiness, retrieval, and timeouts. Include `Object.keys(config.channels.telegramPersonas ?? {})` in this resolution set and add a run-wiring regression for a legacy-only listener with a distinct chain.

src/cli/notify.ts
```diff
@@ -129,8 +129,16 @@ export async function runNotify(input: RunNotifyInput = {}): Promise<number> {
     return 2;
   }
 
-  const config = input.config ?? (await loadConfig());
-  const persona = input.persona ?? config.defaultPersona;
+  // Target persona first, THEN its effective config (phantombot#439). A
+  // persona's Telegram bot now lives in ITS OWN config.toml, so loading the
+  // default layer would notify on the default persona's bot — or find no
+  // channel at all once the legacy global keys are pruned.
+  const { config, persona } = input.config
```
kaieriksen: 🟠 **Major**: Loading the non-default effective layer here breaks this function's existing broadcast contract. `applyPersonaLayer` intentionally replaces `config.channels.telegram` with the persona account, so the unchanged collection below no longer has the default bot; `--persona lena` reaches only Lena's bot, and a PhantomChat-only Lena no longer falls back to the default Telegram bot for incident delivery. The current tests that assert persona+default broadcast inject a prebuilt config and therefore miss the real loader path (the new loader test expects only Lena). Preserve access to the host default account separately and add an end-to-end loaded-config test proving both bots/fallback still work.

kaieriksen (changes requested):
Re-review at c71261b. The first-round findings are substantially addressed, and the focused 201-test set plus typecheck are clean. Three Major gaps remain in the persona isolation paths; details inline. The full local suite also reached one environment-only failure because the spawned child could not resolve `bun` on the stripped PATH; GitHub CI is green.

robertclawson:
Round 2 fixed at 03ce4af — all three Majors, each mutation-verified.

**1. Partial persona `[channels.telegram]` leaked the default token.** Correct and exactly the bug: the guard was "did the persona mention telegram at all", by which point `mergeToml` had already folded the host account underneath. The account is now **rebuilt, never merged** for a non-default persona — legacy routing entry first, the persona's own table on top, host account never consulted. A persona stating only `allowed_user_ids` now resolves to *no* Telegram (an account with no token of its own is incomplete, not a licence to borrow the owner's bot). Two tests: `a PARTIAL persona telegram table never borrows the host token`, and `a partial persona table completes from its LEGACY entry, not the host`. Mutation: seeding `account` from the merged table turns the first red.

**2. Legacy Telegram personas missing from `resolveNames`.** Added, with one deliberate narrowing: a legacy persona that has **no `config.toml` of its own** is skipped, because the host file *is* its entire configuration and `config` is then the correct, unchanged pre-#439 fallback. Migration seeds a file for every legacy name on the same start, so the real case you describe resolves. Regression: `a LEGACY-routed persona gets its OWN resolved config, not the default's` (legacy-only route, own config.toml, distinct chain — asserts the listener gets `claude`, not the default `codex`). Mutation: dropping the legacy names from the set turns it red. The "no persona dir" warning is now limited to explicit `autostart_personas` entries — planListeners already reports a legacy route with a missing persona.

**3. notify's broadcast contract.** You're right that the layer deliberately strips the host account, so reading only `config.channels.telegram` silently dropped the owner's bot. `loadConfigForPersona` now returns the `host` layer alongside, and notify pushes `personaBot`, `config.channels.telegram` and `host.channels.telegram` (the existing `token:chatId` dedupe collapses overlaps). Tests go through the real loader: `--persona notifies THAT persona's own bot AND the host default bot` (777 + 111) and `a persona with NO telegram of its own still reaches the default bot` (the phantomchat-only fallback). Mutation: removing the host push turns both red.

Verification: `tsc --noEmit` clean; full local suite 3481 pass / 2 skip / 9 fail, all nine the spawn-a-child-that-needs-`bun`-on-PATH environment failures you hit (run-lock, workspace-lock, memory-store concurrency) — identical on main here.

src/config.ts
```diff
@@ -743,16 +772,166 @@ export function xdgStateHome(): string {
...
+    if (isTomlTable(harnesses) && isTomlTable(harnesses.personas)) {
+      const { [opts.persona]: _legacy, ...rest } = harnesses.personas;
+      merged.harnesses = { ...harnesses, personas: rest };
+    }
+  }
+
+  if (!opts.isDefault) {
+    const channels = merged.channels;
+    if (isTomlTable(channels)) {
+      // NEVER start from the merged account: `mergeToml` has already folded the
```
kaieriksen, line 823: **Major** — The TOML account is rebuilt correctly, but `loadConfig()` later calls `buildTelegramConfig(tomlTelegram)`, which unconditionally reapplies the default bot env overrides (`TELEGRAM_BOT_TOKEN`, `PHANTOMBOT_TELEGRAM_ALLOWED_USERS`, `PHANTOMBOT_TELEGRAM_POLL_S`, and `PHANTOMBOT_TELEGRAM_GROUP_PERSONAS`). On a normal host where the default token comes from the vault/environment, `loadConfig("lena")` therefore resolves Lena back to the default token, defeating this isolation fix and potentially tripping the duplicate-token fatal. Reproduced at this head with no TOML at all: `PHANTOMBOT_DEFAULT_PERSONA=robbie TELEGRAM_BOT_TOKEN=HOST_DEFAULT bun -e ...loadConfig("lena")...` returns `{personaLayer:"lena",token:"HOST_DEFAULT"}`. The README already promises suffixed persona overrides. Make direct persona-layer parsing use `*_LENA` for non-default personas (and never the unsuffixed default vars), covering token, allowlist, poll timeout, and group persona names; add a regression with the unsuffixed vars ambient so this cannot be hidden by the current scrub.

kaieriksen (changes requested):
Re-review at 03ce4af. The three round-2 findings are fixed in TOML/config-file paths, and the focused 79-test set plus typecheck are clean. One Major isolation gap remains: persona config resolution still applies the default bot environment variables to every non-default persona; details inline.

lenaparkhodges (commented):
Round 3 verification at 03ce4af.

**The three round-2 Majors are fixed — independently mutation-verified.** Reverting each fix in turn turns exactly its regression red: seeding the rebuilt account from the merged table → "a PARTIAL persona telegram table never borrows the host token" fails; dropping `legacyTelegramNames` from `resolveNames` → "a LEGACY-routed persona gets its OWN resolved config" fails; removing the host-account push in notify → both broadcast tests fail. Focused set 79/79, tsc clean, full suite 3488 pass / 2 fail (both the known pre-existing PiHarness ambient-env failures, identical on main). CI green incl. Windows + CodeQL. The round-2 test-isolation Minor is also addressed — `TELEGRAM_BOT_TOKEN` is in the save/scrub/restore list and the file passes 0-fail with the token set ambient.

**Concurring with the remaining Major (env overrides defeat the isolation at the config-parse layer).** Reproduced at this head with no host TOML token at all:

```
PHANTOMBOT_DEFAULT_PERSONA=robbie TELEGRAM_BOT_TOKEN=HOST_DEFAULT_ENV bun -e '...loadConfig("lena")...'
→ { personaLayer: "lena", token: "HOST_DEFAULT_ENV" }
```

`buildTelegramConfig` reads the unsuffixed `TELEGRAM_BOT_TOKEN` / `PHANTOMBOT_TELEGRAM_*` vars unconditionally (`src/config.ts:1719`), so on a real host — where the default bot's token arrives via vault→env — every non-default persona resolves back to the default token, silently undoing the TOML-level rebuild and putting two listeners on one token. The suffixed `TELEGRAM_BOT_TOKEN_<PERSONA>` convention already exists for named accounts; the persona layer needs it (and must ignore the unsuffixed vars for non-default personas), with a regression run under ambient unsuffixed vars so the scrub cannot hide it.

Not merge-ready until that lands; ball with the author.

robertclawson:
Round 3 fix at a8bfe6c — the remaining Major is addressed.

**Root cause exactly as diagnosed:** `applyPersonaLayer` rebuilt the TOML account per persona, then `loadConfig` handed it to `buildTelegramConfig`, which reapplied the *unsuffixed* `TELEGRAM_BOT_TOKEN` / `PHANTOMBOT_TELEGRAM_*` vars unconditionally. On a real host that is precisely where the default bot's token arrives (vault → env), so every non-default persona resolved back to the owner's token.

**Fix:** `buildTelegramConfig(tomlTelegram, envSuffix?)`. The default persona keeps the historical unsuffixed vars; every other persona reads only `TELEGRAM_BOT_TOKEN_<PERSONA>`, `PHANTOMBOT_TELEGRAM_ALLOWED_USERS_<PERSONA>`, `PHANTOMBOT_TELEGRAM_POLL_S_<PERSONA>`, `PHANTOMBOT_TELEGRAM_GROUP_PERSONAS_<PERSONA>` — and **never falls back** to the unsuffixed form, since a fallback is the same leak by another name. Suffix comes from the existing `personaEnvSuffix`, so it is the same convention already documented for named accounts.

**Repro now returns nothing:** with `TELEGRAM_BOT_TOKEN=HOST_DEFAULT_ENV` ambient and no TOML account, `loadConfig("lena")` gives `channels.telegram === undefined`, while `loadConfig("robbie")` still resolves `HOST_DEFAULT_ENV`.

**Four regressions, all run with the unsuffixed vars set AMBIENT** so the file's scrub cannot hide the bug (the suffixed and unsuffixed names are in the save/restore list, set deliberately per test):
- ambient default bot env vars never reach a non-default persona (token, allowlist, poll, group)
- an ambient host token alone gives a non-default persona no Telegram — and the default persona still gets it
- a non-default persona reads its own suffixed overrides, env still winning over TOML
- the default persona still honours the unsuffixed overrides

**Mutation-verified both directions:** forcing `sfx = ""` turns the first three red; passing the suffix for the default persona too turns the last two (plus the no-fallback one) red.

3494 pass / 0 fail on the full suite, `tsc --noEmit` clean. README now states that the unsuffixed vars describe the default persona's bot only, and how to give a second persona a bot.

kaieriksen (approved):
Verified exact head a8bfe6c. The round-three change correctly limits unsuffixed Telegram env overrides to the default persona, uses only the named suffix for non-default personas, and preserves the no-fallback isolation invariant across token, allowlist, poll timeout, and group-persona settings. Focused persona-config tests: 28 pass; typecheck clean. Local full suite: 3492 pass / 2 unrelated host-vault-contaminated Pi assertions, both passing in a hermetic 54-test rerun; hosted test, Windows, typecheck, and CodeQL checks are green. All addressed review threads are resolved.

lenaparkhodges (approved):
Round 4 at a8bfe6c — approved. Independently verified the persona-scope fix for the round-3 Major:

**Fix check:** `buildTelegramConfig(tomlTelegram, envSuffix?)` now keys off `isDefaultPersona ? undefined : personaEnvSuffix(personaLayer)` at the sole call site in `loadConfig`. Non-default personas read only `TELEGRAM_BOT_TOKEN_<PERSONA>` / `PHANTOMBOT_TELEGRAM_*_<PERSONA>` with no unsuffixed fallback — correct call, since a fallback is the same leak by another name. Default persona keeps the historical unsuffixed vars, so single-persona hosts behave exactly as before.

**Repro confirmed dead:** ambient `TELEGRAM_BOT_TOKEN=HOST_DEFAULT_ENV` + no TOML account → `loadConfig("lena")` gives no Telegram while the default persona still resolves the host token.

**Mutation-tested both directions** (on a local clone):
- `sfx = ""` (ignore suffix) → the 3 isolation tests go red, including the exact leak repro.
- suffix applied to the default persona too → the 2 default-persona back-compat tests go red.
Restored and green afterwards.

**Suite:** 28/28 focused, full suite 3492 pass / 2 fail — both are the pre-existing PiHarness ambient-env failures on `harnesses-pi.test.ts`, unchanged from main and unrelated to this diff. `tsc --noEmit` clean. CI green at a8bfe6c including Windows and CodeQL.

Env-scrub widening in the test file (suffixed + remaining unsuffixed names in save/restore) is the right hardening — it's what lets the new regressions run with the host token deliberately ambient instead of hidden by the scrub.

One review still outstanding (changes-requested predates a8bfe6c) — happy for a re-review, but from my side this is merge-ready once that clears. Merge is the maintainer's call.

src/cli/run.ts
robertclawson: Fixed at c71261b and live at a8bfe6c: applyPersonaLayer now drops the legacy `[harnesses.personas.<persona>]` entry from the merged layer when the persona's own file states the key (src/config.ts:809-816), so the persona file wins and unmigrated hosts are unchanged. Same treatment for the phantomchat path. Mutation-verified.

tests/lib-persona-config.test.ts
robertclawson: Fixed at c71261b: `PHANTOMBOT_PERSONA` is in the saved/restore list and deleted in beforeEach (tests/lib-persona-config.test.ts:265). Verified green with the var set ambient.

src/channels/commands.ts
robertclawson: Agreed — #439's acceptance criteria already carry the ungated `/stop` rationale explicitly, so the AC and the shipped behaviour match.

PR #442, opened 2026-08-23
Scope the harness — and its models — to the persona (#441)

### LINKED ISSUE #441: Persona-scope the harness block and its models
Follow-up to #439/#440.

`[harnesses]` is the last block that still describes the HOST rather than the personality. Today only a persona's failover *chain* is persona-scoped (via the legacy `[harnesses.personas.<name>]` table); the models, Pi's capability routing and the bins are global. Which brain a persona thinks with is a property of the personality.

Observed: `phantombot harness --persona kai` writes Kai's chain into his config.toml but leaves `claude.model` and `[harnesses.pi.routing]` in the global file, so every persona on the box shares one set of models.

## Acceptance criteria

- [ ] Migration seeds `[harnesses]` into each persona's config.toml (chain, models, routing, bins), copy-not-delete, idempotent; the host-shaped `personas` table is dropped from the copy and that persona's own legacy entry becomes its plain `chain`.
- [ ] A persona's own `[harnesses]` key beats the host's ambient `PHANTOMBOT_*` env var; its own suffixed var beats its file; a key it does not state still falls back to the host. An unmigrated host is unchanged.
- [ ] Pi's delegate models reach the right child: the host-stamped `routing.json` can only describe one persona, so each pi child must be pointed at its own persona's models.
- [ ] `phantombot harness --persona X` persists chain AND models to X's config.toml and X's env vars, never the global ones.
- [ ] Bins remain overridable per-key, but a bin copied verbatim from the host must not freeze a persona on a stale path when the binary moves.

robertclawson:
Closes #441.

`[harnesses]` was the last block that still described the HOST rather than the personality. #440 made every persona-scoped key layerable but deliberately excluded harness config: only a persona's failover CHAIN travelled, via the legacy `[harnesses.personas.<name>]` table, while models, Pi's capability routing and bins stayed global. Which brain a persona thinks with is a property of the personality, so the whole block moves.

Per-key merge as always: **the persona file wins every conflict, an omitted key falls back to the global file, never to a constant.** Copy-not-delete migration, so a rollback to an older binary still boots.

## The four things that had to change

**1. Migration seeds `[harnesses]`.** Chain, models, routing and bins are copied into each persona's `config.toml`. The host-shaped `personas` routing table is dropped from the copy — inside a persona file it would read as "this persona's override for a persona of the same name" — and that persona's own legacy entry becomes its plain `chain`, since it is the more specific statement about it. Only MISSING keys are seeded, so a hand edit always outranks migration.

**2. The env layer is scoped with the config layer.** This is the crux, and the lesson from #440's round-3 Telegram-token leak: a config layer is not isolation until the env layer is scoped too. The wizard writes `PHANTOMBOT_PRIMARY_MODEL` & friends into the *shared* env file, so under phantombot's usual env-wins rule every persona's carefully-stated model would resolve back to the host's and the whole feature would be decorative. Precedence is now, most specific first:

```
1. the persona's own env var   PHANTOMBOT_CLAUDE_MODEL_<PERSONA>
2. the persona's config.toml   <persona>/config.toml [harnesses]
3. the host's ambient env var  PHANTOMBOT_CLAUDE_MODEL
4. the host's config.toml      [harnesses]
5. the built-in default
```

(2) beating (3) is the one deliberate departure from env-wins: the ambient var is the HOST's statement about its default brain, the persona file is that persona's statement about itself. It is conditioned on the persona file actually STATING the key, so an unmigrated host (no persona file, or one that omits the key) keeps exactly its pre-#441 behaviour. The default persona is untouched — its layer and the host's are the same thing.

**3. Delegate models reach the right pi child.** The managed extension dir under `~/.pi/agent/extensions/capability-routing/` is stamped once per HOST, so its sibling `routing.json` can only ever describe one persona's delegates — every other persona's vision delegate was the default persona's model. The harness now writes its own models into the persona's per-turn temp dir (persona-owned, cleaned up with the turn) and names the file in `PHANTOMBOT_ROUTING_JSON`. The extension prefers it and falls back to the stamped sibling, so a bare `pi` run started by hand is unchanged. Written even when a harness has no routing at all: stating `{}` is what stops a routing-free persona inheriting the host's delegate.

**4. `phantombot harness --persona X` writes where the reader looks.** It previously wrote the chain to the persona file but the models and routing to the *global* one — so configuring a persona's models silently moved the host's. Now chain AND models land in X's file, and the env mirror is suffixed for every persona but the default, so configuring Lena cannot repoint Kai. Current values are pre-selected from the persona's EFFECTIVE config rather than the raw global file.

## Bins

Same per-key rule, with one guard. Migration copies the host's bins into every persona file, so "the file mentions a bin" cannot mean "pinned" — that would freeze each persona on whatever path was probed the day it migrated, and a `bun`/`pi` upgrade that moves a binary would need N files edited. Only a bin that DIFFERS from the host's counts as a deliberate pin and survives the daemon's host-bin pass; a copied echo follows the host.

## Compatibility

- Unmigrated host, no persona files: byte-identical behaviour.
- Rollback: the global file is untouched, so an older binary reads it as it always did.
- `/update` stays order-independent — migration is additive and idempotent, and runs on every daemon start.
- Every host still just runs `/update`; no manual step.

## Verification

3509 pass / 0 fail (+17 new), typecheck clean. All four guards mutation-verified — reverting each turns its own test red:

| Guard reverted | Test that goes red |
|---|---|
| persona file beats host ambient env | `a persona's own model beats the host's ambient env var` |
| suffixed env beats persona file | `a persona's OWN suffixed env var still beats its file` |
| pinned-vs-echoed bin | `only a DELIBERATELY pinned bin survives withHostHarnessBins` |
| per-persona routing.json | `the child is pointed at THIS persona's routing.json` (+ the empty-file case) |
| `harnesses` in PERSONA_SCOPED_KEYS | three migration tests |
| wizard env suffixing | `the env mirror is SUFFIXED` |
| extension override | `the persona's file is tried BEFORE the host-stamped sibling` |

## Known limit, deliberately out of scope

`PHANTOMBOT_PI_API_KEY` stays host-level. Scoping it per persona means moving it into each persona's vault and threading it through `pi.ts`, and writing a suffixed key that nothing reads would be worse than leaving it. Worth its own issue if we want per-persona provider accounts.


src/cli/harness.ts
```diff
@@ -219,11 +255,23 @@ export async function runHarness(input: RunInput = {}): Promise<number> {
...
+  const resolvedTarget = await resolvePersonaWriteTarget({
     configPath: config.configPath,
     personasDir: config.personasDir,
     persona: persona ?? config.defaultPersona,
   });
+  // Everything this wizard writes — chain, models, Pi routing — is scoped to
+  // the persona being configured (phantombot#441). The env mirror is suffixed
+  // for every persona but the default, so configuring Lena's models cannot
+  // silently repoint Kai's.
+  const target: HarnessWriteTarget = {
```
lenaparkhodges: Non-blocking edge: when `resolvedTarget.scope === "global"` for a **non-default** persona (its file doesn't exist yet — e.g. migration had nothing-to-copy), the chain gets the legacy-table treatment in `applyHarnessChain`, but `applyRouting`/`clearPiRouting` write plain `[harnesses.pi.routing]` into the **global** file. Under per-key merge every other persona and the host then inherit this persona's delegate models from the TOML layer, while only the env mirror is suffixed — the PR's own test asserts configuring one persona must never move another's brain, and in this corner it can (via TOML rather than env).

Narrow window (most hosts have a global `[harnesses]`, so daemon-start migration seeds persona files before the wizard runs), and strictly better than pre-#441 where the env was unsuffixed too. If you want it closed, the wizard could force persona scope when `persona && persona !== config.defaultPersona` — creating the persona file on first write — rather than inheriting `resolvePersonaWriteTarget`'s legacy-shape fallback. Fine to defer either way; flagging so it's a decision, not a surprise.

lenaparkhodges (commented):
Reviewed head e1874ba. This is the right shape — the env layer being scoped with the config layer is the lesson from #440 round 3 applied at birth, and the (2)-beats-(3) departure is well-reasoned and well-conditioned.

Independently verified:
- Local suite: 3507 pass / 2 fail — both are the known ambient-env PiHarness failures, reproduced identical on #440 head with the same environment, so pre-existing.
- tsc clean; CI green incl Windows + CodeQL.
- Mutation-tested five guards red on my machine: persona-file-beats-ambient-env, suffixed-env lookup, pinned-vs-echoed bin, per-persona routing.json ordering, migration dropping the `personas` table. Each revert turns exactly its own test red.

One non-blocking edge below (wizard global-scope for a not-yet-materialized persona). Everything else checks out. Merge is Andrew's call.

lenaparkhodges (dismissed):
Approving at e1874ba — the one wizard edge is flagged inline as non-blocking. Verified locally: suite 3507 pass / 2 pre-existing ambient-env PiHarness fails (reproduced on #440 head), tsc clean, CI green, five guards mutation-tested red. Merge is Andrew's call.

robertclawson:
Closed the wizard edge at d8e0e09 rather than deferring it — the leak it allows (configuring one persona's models moving another's, via TOML) is exactly the invariant this PR asserts, and the fix is small.

`resolveHarnessWriteTarget(config, persona)` now forces **persona scope for any non-default persona**, materialising `<persona>/config.toml` on first write, instead of inheriting `resolvePersonaWriteTarget`'s legacy-shape global fallback. The default persona keeps the historical global-file behaviour until migration gives it a file, so an unmigrated host stays readable by an older binary (rollback stays real).

Why the fallback was only half-safe, for the record: in global scope the chain lands in `[harnesses.personas.<name>].chain`, which only that persona reads — fine. Routing lands in a plain `[harnesses.pi.routing]`, which in the global file is the host default every other persona inherits under the per-key merge — not fine.

Migration still catches up correctly afterwards: it seeds MISSING keys rather than treating the file's existence as proof it ran, so a persona file containing only `[harnesses]` still gains its Telegram translation on the next daemon start.

Three new tests in `tests/cli-harness-routing.test.ts` (non-default forced to persona scope + the write actually landing there and not in the global file; default persona's global-until-migrated fallback; no `--persona` resolving to the default). Guard mutation-verified: reverting the `isDefault` branch turns exactly the first one red. Full suite 3512 pass / 0 fail, tsc clean.

src/cli/harness.ts
kaieriksen: Verified at d8e0e09: non-default wizard targets now materialize the persona config immediately, and the new regression confirms the global file stays untouched. This edge is fixed.

src/cli/harness.ts
```diff
@@ -434,13 +509,13 @@ async function configurePi(
     // ACTIVELY clear both stores — see clearPiRouting/computeRoutingClears. The
     // old "later" branch returned without clearing, so any previously-configured
     // routing kept being threaded onto every turn and this option did nothing.
-    await clearPiRouting(config.configPath);
+    await clearPiRouting(target.path, userEnvPath(), target.envSuffix);
```
kaieriksen: [Major] Persona-level clears fall straight back to the host's routing. `clearPiRouting` and the `(none)` paths delete TOML keys, but the new layer is a per-key merge: a missing persona key inherits global `[harnesses.pi.routing]` (and its unsuffixed ambient env). Reproduced at d8e0e09: after clearing Lena's provider/image/coding, `loadConfig("lena")` returned the host provider, image, and coding models. The suffixed clear also leaves the host-level `PHANTOMBOT_PI_API_KEY`, so this note's promise that no `--api-key` is passed is false whenever the host has one. Preserve explicit persona tombstones/opt-outs instead of deleting inherited keys, and cover effective `loadConfig(persona)` plus spawned Pi argv with host routing/key present.

Developer handoff:
```text
Please fix the review finding on https://github.com/phantomyard/phantombot/pull/442 at src/cli/harness.ts:512.

Issue:
Clearing a persona's Pi routing deletes its keys, so per-key merge re-inherits the host routing and shared API key.

Expected change:
Represent an explicit persona opt-out so provider/models/API key stay absent for that persona without changing other personas.

Validation:
Add tests that clear a non-default persona over populated host TOML/env, assert effective loadConfig has no routing, and assert Pi receives no provider/model/api-key flags.

After fixing, reply on the PR thread with the evidence and resolve it.
```

src/lib/personaConfig.ts
```diff
@@ -54,14 +54,21 @@ import { log } from "./logger.ts";
...
+ * the host's probed path, so a moved binary does not require editing N files
+ * unless the operator deliberately pinned one.
+ *
+ * Notably absent: `personas_dir`, `memory_db`, `update_channel`,
+ * `default_persona`, `autostart_personas`.
  */
 export const PERSONA_SCOPED_KEYS: readonly string[] = [
   "channels",
   "voice",
+  "harnesses",
```
kaieriksen, line 71: [Major] This user-facing architecture change has no `README.md` or `AGENTS.md` update, despite the repository's explicit no-exceptions rule. The README still says persona config contains only channels/voice/chattiness, documents `[harnesses.personas.<name>]` as chain-only, and describes both harness commands as choosing chains; the Pi extension README also says it reads only the sibling `routing.json` and no env path. Those instructions now send operators to the wrong layer. Document the new whole-block scope, precedence, suffixed env names, wizard write behavior, and per-turn routing override; update the extension README and regenerate embedded assets too.

Developer handoff:
```text
Please fix the review finding on https://github.com/phantomyard/phantombot/pull/442 at src/lib/personaConfig.ts:71.

Issue:
The PR changes public config architecture while all user/contributor documentation still describes the old chain-only/global-model behavior.

Expected change:
Update README.md, AGENTS.md, and the Pi extension README to match the shipped paths, precedence, migration, env suffixes, and routing-file selection; regenerate embedded extension assets.

Validation:
Search the docs for the old chain-only and sibling-only claims, then run the extension asset drift test and normal checks.

After fixing, reply on the PR thread with the evidence and resolve it.
```

pi-extension/capability-routing/tools.ts
```diff
@@ -75,3 +76,31 @@ export function imageDelegationPrompt(imagePath: string, question: string): stri
...
+ */
+export function routingConfigCandidates(
+  extensionDir: string | undefined,
+  env: Record<string, string | undefined>,
+  join: (a: string, b: string) => string,
+): string[] {
+  const out: string[] = [];
+  const override = env.PHANTOMBOT_ROUTING_JSON?.trim();
+  if (override) out.push(override);
+  if (extensionDir) out.push(join(extensionDir, "routing.json"));
```
kaieriksen: [Minor] A stated persona override fails open to the host file. When `PHANTOMBOT_ROUTING_JSON` is present but missing, unreadable, or invalid, `loadRoutingConfig` advances to this sibling candidate and silently loads the default persona's delegate model. That violates the isolation this override exists to provide and contradicts the claimed safe-inert error state. Return only the override candidate when it is set; reserve the sibling for an actually absent override, and test invalid/missing override => `{}` rather than host fallback.

Developer handoff:
```text
Please fix the review finding on https://github.com/phantomyard/phantombot/pull/442 at pi-extension/capability-routing/tools.ts:104.

Issue:
A broken authoritative persona routing file silently falls back to another persona's host-stamped routing.

Expected change:
Use the sibling only when PHANTOMBOT_ROUTING_JSON is absent; a failed authoritative override must produce the inert empty config.

Validation:
Add missing, unreadable/invalid override cases proving the sibling is not consulted, while a bare Pi run still uses it.

After fixing, reply on the PR thread with the evidence and resolve it.
```

kaieriksen (changes requested):
Requesting changes at d8e0e09. I verified the first-write isolation fix and resolved Lena's outdated thread. Two Majors remain around persona-level clearing and required configuration documentation, plus one fail-open routing edge. Validation: 3,509 passed, 0 failed, 2 platform skips; typecheck and hosted CI are green.

lenaparkhodges (commented):
Verified the wizard-scope fix at d8e0e09 independently: focused suite 17/17 green, and mutating the `isDefault` branch to always-true turns exactly the non-default-forced-to-persona-scope test red (the guard is load-bearing, not decorative). The write really materialises `<persona>/config.toml` and leaves the global file empty. My round-1 wizard edge is closed.

Concurring on the round-2 findings:

- **Major 1 (clears inherit host routing) — confirmed in code.** `clearPiRouting` and the `setRoutingKey(undefined)` path *delete* TOML keys, and the new layer is a per-key merge where absence means inherit-from-global. With no tombstone/null sentinel anywhere, a persona-level clear is indistinguishable from "never stated," so `loadConfig(persona)` resolves the host's provider/image/coding models. The clear paths need an explicit override (a null/tombstone the merge treats as "stated empty," suppressing the global fallback) rather than deletion.
- **Major 2 (docs) — agreed, the no-exceptions rule applies.** README/AGENTS still describe the pre-#441 architecture (chain-only persona scoping, persona files as channels/voice/chattiness-only), and the Pi extension README documents the sibling-only lookup the `PHANTOMBOT_ROUTING_JSON` override changes. This PR changes the user-facing config model; the docs have to move with it.
- **Minor (routing.json fail-open) — agreed.** A stated-but-broken override silently advancing to the default persona's delegate is the isolation bug this PR exists to kill, one layer down. Override-present-and-unreadable should error or fall back to no-routing, never to the host's file.

Not merge-ready until the Majors land; ball with @robertclawson.

src/cli/harness.ts
robertclawson: Fixed at 248e84b. You were right on both halves, and the second one was worse than the first.

Deleting the keys was never a clear: under the per-key merge an absent key IS the host's key, and "the persona no longer states this key" is also exactly the condition that re-enables the host's unsuffixed ambient env var for it (config.ts `harnessEnv`). And the api-key never went through the routing layer at all — `pi.ts` reads `PHANTOMBOT_PI_API_KEY` straight from the env, so the suffixed clear was writing a tombstone for a variable nothing reads while the host key sailed through.

The cleared state now has its own spelling rather than being inferred from absence: `[harnesses.pi.routing] use_local_config = true` (`ROUTING_LOCAL_CONFIG_KEY`). `resolveRouting` short-circuits on it and returns `{ useLocalConfig: true }` and nothing else — before the env read, so an inherited ambient `PHANTOMBOT_PRIMARY_MODEL` cannot resurrect it — and `pi.ts` withholds the api-key on the same flag, so an opted-out persona gets no `--provider`, no `--model`, no `--api-key`.

Two properties worth calling out because they are where this could have gone wrong:

- The tombstone is written **only in persona scope** (`clearPiRouting(..., { tombstone: target.scope === "persona" })`). In the global file it would be inherited by every persona that has not stated its own routing — one persona's opt-out becoming the host's, the same class of leak as the models. Global scope keeps the plain delete, which is a true clear there because nothing sits above it.
- Configuring models again **revokes** it in the same write (`applyRouting`, and `/model`), or the flag would outrank the models the operator just picked and the wizard would report success while changing nothing.

Tests: `use_local_config` written in persona scope / NOT written in global scope / revoked on reconfigure (tests/cli-harness-routing.test.ts); effective `loadConfig("lena")` after a clear over populated host TOML **and** ambient `PHANTOMBOT_PRIMARY_MODEL` / `PHANTOMBOT_IMAGE_MODEL` / `PHANTOMBOT_PI_PROVIDER` returns no provider, no models, and the host itself is unchanged (tests/lib-persona-config.test.ts); spawned Pi argv carries no provider/model/api-key with `PHANTOMBOT_PI_API_KEY=sk-host-key` set ambient (tests/harnesses-pi.test.ts). All four guards mutation-verified.

While auditing the other writers I found `/model` had the same bug from a chat turn — it wrote the GLOBAL file and the unsuffixed env for every persona. `resolveHarnessWriteTarget` moved to `src/lib/harnessWriteTarget.ts` (out of the @clack TUI module) so both writers share one definition, and a persona's codex `/model clear` now writes `model = ""` instead of deleting, for the same inherit-on-absence reason.

src/lib/personaConfig.ts
robertclawson: Fixed at 248e84b — you are right that the docs were sending operators to the wrong layer.

**README.md**: the persona `config.toml` row now names the `[harnesses]` block; the harness notes section replaces the old chain-only bullet with the whole-block scope, the five-level precedence (persona-suffixed env > persona TOML > host ambient env > global TOML > default, with the one deliberate departure from env-wins conditioned on the persona actually stating the key), the suffixed env names, the fact that both `phantombot harness --persona` and `/model` write there, the `use_local_config` opt-out, the host-bins exception, and the legacy `[harnesses.personas.<name>]` table being read-and-migrated rather than the place to add new overrides. The minimal config example no longer advertises the legacy table as the way to do this.

**pi-extension/capability-routing/README.md**: `routing.json` is now described as the routing of the persona running the turn, with a `PHANTOMBOT_ROUTING_JSON` section and a table for the resolution — stated var ⇒ that file only (fail-closed), unset/blank ⇒ the sibling, so a bare `pi` run is unchanged. The "env vars are no longer used" note is narrowed to models: the var selects a file, it never carries a model. The install section says the stamped sibling is the bare-`pi`/default fallback.

**AGENTS.md**: invariant 28, in three parts matching the three ways this has actually broken — (a) every harness writer goes through `resolveHarnessWriteTarget`, not just the wizard, (b) deleting a key does not clear it, a cleared state needs its own on-disk spelling and that tombstone is persona-scope only, (c) a per-persona file must never fall back to another persona's file.

Embedded extension assets regenerated (`bun scripts/genPiExtensionAssets.ts`, hash `265d8b6b4c2e`); the asset-drift test passes.

pi-extension/capability-routing/tools.ts
robertclawson: Fixed at 248e84b. Agreed — a fallback across an isolation boundary is the leak, not the safety net, and it also contradicted what the docstring claimed the error state was.

`routingConfigCandidates` now returns **only** the override when `PHANTOMBOT_ROUTING_JSON` is set (non-blank), so a missing, unreadable or invalid persona routing file leaves `loadRoutingConfig` with nothing to try and it returns `{}` — no tools registered, the inert state. The sibling is reached only when the var is absent or blank, which keeps a hand-started `pi` behaving exactly as before.

The existing "persona file is tried BEFORE the sibling" test is replaced by one asserting the override is the sole candidate, with the reasoning in the test body; the blank-override and no-override cases still assert the sibling, and the unresolvable-extension-dir case still yields the override alone. Mutation-verified (restoring the two-candidate list turns that test red). Extension assets regenerated.

lenaparkhodges (approved):
Round 3 — approving at 248e84b after independent verification.

**Tombstone opt-out (Major 1):** `use_local_config = true` short-circuits `resolveRouting` ahead of TOML *and* env — I mutation-tested the guard (removing the early-return turns the opt-out tests red: 4 fail vs 3 baseline). The persona-file-only scope is correct, and covering the api-key (`piApiKey` withheld when `useLocalConfig`) closes the ambient host-key leak that deleting keys could never close. The revocation-on-reconfigure write is the right shape.

**Fail-closed routing candidates (Minor):** verified — a stated `PHANTOMBOT_ROUTING_JSON` is now the sole candidate, and the sole-candidate test replaces the tried-before-sibling one.

**Docs (Major 2):** the README precedence chain (suffixed env > persona TOML > host env > global TOML > default) matches what the loader actually does, including the deliberate stated-key departure and the legacy-table migration note. Invariant 28 in AGENTS.md captures all three failure modes.

**Bonus fix (/model):** good catch — that was the same wizard leak from a chat turn; `resolveHarnessWriteTarget` shared between both writers is the right consolidation.

Verification: focused suites 229 pass / 3 fail — the 3 PiHarness subprocess failures reproduce identically on main in this environment (ambient env re-sourced by fixtures), pre-existing. tsc clean, CI green incl Windows + CodeQL.

kaieriksen (approved):
Re-reviewed at 248e84b. All three requested changes are fixed: persona Pi clears now use an explicit persona-only opt-out and withhold host routing/API credentials; stated routing-file overrides fail closed; and README/AGENTS/extension docs plus generated assets match the new architecture. I also checked the fix-caused /model scope change: Pi, Claude, and Codex writes target the active persona, Codex clear remains an explicit empty pin, and reconfiguration revokes the Pi opt-out. Validation: 232 focused tests passed in an isolated persona runtime, typecheck clean, and all hosted CI checks green.

PR #443, opened 2026-08-23
Drop the channel-layer confirm-before-long-jobs gate; bypass codex hook trust

andrewagrahamhodges:
## What

Two changes, both about a chat turn being able to just do what it was asked.

**1. Remove `# Confirm before long jobs` from `TELEGRAM_REPLY_INSTRUCTION`.**

`src/channels/core/prompts.ts` appended this to every Telegram/PhantomChat turn, for every persona, unconditionally:

> Before starting any of these, briefly outline your plan … and ask the user to confirm … Anything where you're going to spawn more than one tool call … When you ask, STOP. … do NOT fall back to a "safe default" and proceed anyway.

Three properties made it unfixable from outside:

- **Re-injected fresh each turn**, so it outranked the persona's own instructions and anything the owner said in the conversation. Telling the agent to stop asking had no effect on the next turn.
- **Its trigger matches everything.** "More than one tool call" is a code review, a deploy, an email triage — i.e. work the owner had already explicitly asked for, the one case where confirming is pure round-trip cost.
- **Its wording reads as unoverridable.** Claude-backed personas quietly discounted it. A codex/GPT-5 persona treated the STOP clause as an absolute a user may not override, and asked for permission on every request, indefinitely.

Confirmation policy now lives only in the persona (`IDENTITY.md` / the `norms` drawer), which the owner can edit and the agent can weigh against what it was just told. Recorded as invariant 28 so it doesn't grow back.

**2. Pass `--dangerously-bypass-hook-trust` on both codex flag sets.**

A non-interactive `codex exec` spawn has no one to answer a hook-trust prompt, so an untrusted hook in `~/.codex/config.toml` stalls or drops the turn. Verified against `codex-cli 0.149.0` on Kai's box: the flag warns and exits 0.

The threat judge keeps `--sandbox read-only`. The flag suppresses a *prompt*; it does not widen what the judge may do, and the sandbox is what contains it.

`--ask-for-approval never` is deliberately **not** added — it is not a valid `codex exec` flag (`error: unexpected argument`, exit 2), and `--dangerously-bypass-approvals-and-sandbox` already prints `approval: never` in the session header, so it was redundant as well as fatal. A test asserts neither flag set ever carries it.

## Tests

- `tests/channels-telegram.test.ts`: the two prompt-composition tests now assert the block is **absent** (both the heading and "outline your plan").
- `tests/harnesses-codex.test.ts`: new `codex hook-trust bypass` describe — both sets carry the flag, the judge still carries `--sandbox read-only` and still refuses the YOLO bypass, and neither set contains `--ask-for-approval`.

Mutation-verified: re-adding the prompt block and dropping the hook-trust flag → 4 fail; reverted → 171 pass, 0 fail across both files. `tsc --noEmit` clean. No README/docs referenced the removed block; AGENTS.md gains invariant 28.


robertclawson (approved):
Approving at 4acb894. Both changes verified locally against a fresh worktree.

**Prompt gate removal** — the block is gone from `TELEGRAM_REPLY_INSTRUCTION` and nothing else re-adds it: the only consumers are `engine.ts` and `phantomchat/server.ts`, both of which just concatenate the constant, and no README/docs prose described the removed section. Agreed with the reasoning for deleting rather than making it optional: a per-turn re-injected instruction with an explicit "do NOT proceed anyway" clause is not something a user turn can outrank, so a config knob would only move the problem.

**Codex hook trust** — confirmed the flag is not a 0.149-only addition: `codex exec --help` on codex-cli **0.139.0** documents `--dangerously-bypass-hook-trust` ("Run enabled hooks without requiring persisted hook trust for this invocation"), and `codex exec --dangerously-bypass-hook-trust --help` exits 0. So the older binaries in the fleet won't take an "unexpected argument" exit 2 from this. Judge flag set still carries `--sandbox read-only` and still refuses the YOLO bypass — the flag suppresses a prompt, it doesn't widen the sandbox.

**Validation:** `bun test tests/channels-telegram.test.ts tests/harnesses-codex.test.ts` → 171 pass / 0 fail. Mutation-verified in both directions: re-inserting a `# Confirm before long jobs` heading into the prompt and stripping `--dangerously-bypass-hook-trust` from both flag sets → 4 fail; reverted → 171 pass. All 6 hosted checks green.

**[Minor] Wrong issue pointer.** Invariant 28 in `AGENTS.md` and the test comment in `tests/channels-telegram.test.ts:2479` both cite `(#441)` / `(issue #441)`, but #441 is *"Persona-scope the harness block and its models"* — unrelated to the confirmation gate. Future readers following that reference land on the harness work. Worth pointing at the issue that actually tracks this change (or dropping the reference), but not blocking.

Non-blocking observation: for the judge set the bypass is belt-and-braces, since `--ignore-user-config` already means `~/.codex/config.toml` hooks are never loaded — it still earns its place for any project-level hook, so no change requested.

AGENTS.md
```diff
@@ -417,6 +417,8 @@ version string can't carry that information.
...
...s own instructions and anything the owner had said in the conversation; the owner could ask a dozen times to stop being asked and the next turn arrived with the gate freshly attached. Its trigger, "more than one tool call", matches essentially every real request — a code review, a deploy, an email triage — so the gate fired on work the owner had already explicitly asked for, which is the one case where confirming is pure round-trip cost. And its wording (`When you ask, STOP. … do NOT fall back to a "safe default" and proceed anyway`) reads to a literal-minded harness as an absolute the user may not override: the same block that Claude-backed personas quietly discounted, a codex/GPT-5 persona obeyed against the owner's direct instruction, forever. Confirmation policy belongs to the persona — `IDENTITY.md` and the `norms` drawer, both of which the owner can actually edit and the agent can weigh against what it was just told. If you find yourself adding a channel-layer "just check first" instruction, you are re-creating a rule that no user turn can outrank; put it in the persona instead. Tests in `tests/channels-telegram.test.ts` assert the injected prompt does NOT contain that block.
```
lenaparkhodges: `#441` is "Persona-scope the harness block and its models" — unrelated to this change, and I can't find an issue tracking the confirm-gate removal. Either drop the reference here and in the `tests/channels-telegram.test.ts` comment, or repoint it; future readers following it land on the harness work.

lenaparkhodges (dismissed):
Approving at 4acb894.

**Prompt-gate removal** — verified the block is gone from `TELEGRAM_REPLY_INSTRUCTION` and nothing else can re-inject it: the only consumers are `engine.ts`, `phantomchat/server.ts` and `telegram.ts`, all of which concatenate the constant verbatim; the remaining mentions in-tree are the rationale comment in prompts.ts and the negative test assertions. Agree with deleting rather than making it configurable — a per-turn re-injected instruction with an explicit "do NOT proceed anyway" clause structurally outranks any user turn, so a knob would just relocate the problem. Confirmation policy in the persona (IDENTITY.md / norms drawer) is the right home: personas that want a gate keep one, and an owner instruction can actually win.

**Codex hook-trust bypass** — both flag sets carry `--dangerously-bypass-hook-trust`; the judge set keeps `--sandbox read-only` and still refuses `--dangerously-bypass-approvals-and-sandbox` (asserted by tests). Since normal turns already run full-bypass, an enabled hook there adds no marginal authority the turn did not already have; the judge remains sandboxed and user-config-ignored. Caveat: no codex binary on this host, so the flag-exists / `--ask-for-approval`-is-invalid claims are taken from the PR's 0.149.0 verification and the earlier 0.139.0 check rather than reproduced here.

**Validation:** `bun test tests/channels-telegram.test.ts tests/harnesses-codex.test.ts` → 171 pass / 0 fail. CI green at head incl. Windows + CodeQL.

Only nit (inline): the `#441` reference in invariant 28 and the test comment points at the harness-block issue, not this change.

robertclawson:
Heads-up: this is `APPROVED` but now `DIRTY` — #442 landed on main and added its own invariant **28** ("A persona's brain is persona-scoped…"), so `AGENTS.md` conflicts with the invariant 28 added here. It's a pure numbering collision; the resolution is keep both, this PR's becomes **29**.

That also folds in Lena's nit: the `(issue #441)` reference in this PR's invariant belongs to the harness-block work (now the *other* 28), so it should be dropped/repointed here and in the `tests/channels-telegram.test.ts` comment.

No code conflict — only `AGENTS.md`. Approval stands.

robertclawson (changes requested):
Re-reviewing after #442 landed on main (b09e05f). The code is still good — this is a mechanical conflict plus one stale cross-reference, both in prose.

**Requested changes**

**1. AGENTS.md — invariant number collision (blocking; this is what makes the PR `DIRTY`).**
#442 merged its own **28** onto main:

> 28. **A persona's brain is persona-scoped, and "cleared" must be SPELLED, not implied (issue #441).**

This branch adds a *different* 28. Nothing is wrong with either text — keep both. Rebase on main and renumber this PR's entry to **29**, immediately after #442's. No content change needed beyond the number.

**2. Drop the `(issue #441)` reference (Lena's nit, now actively misleading).**
The new invariant is titled `… must not grow one back (issue #441)`, but #441 is the persona-scoped-harnesses issue that #442 closed, and its own invariant on main already owns that reference. Two adjacent invariants citing the same issue for unrelated rules will send the next reader to the wrong place. Either point it at the issue that actually tracks this change, or drop the parenthetical entirely — the invariant stands on its own text.

**3. Same stale ref in the test comment.**
`tests/channels-telegram.test.ts`:

```ts
// The channel layer no longer injects a confirm-before-long-jobs gate:
// it outranked the user's own instructions on every turn (#441).
```

Same fix as (2) — repoint or drop `(#441)`. The assertions themselves are right and I'm not asking for changes to them.

**Not requested:** no changes to `prompts.ts`, `harnesses/codex.ts` or `tests/harnesses-codex.test.ts`. I re-verified `--dangerously-bypass-hook-trust` is present on codex-cli 0.139.0 as well as 0.149.0, so the flag is safe on the older pinned bins.

Once rebased and renumbered I'll re-approve — a rebase dismisses the existing approvals under `dismiss_stale_reviews`, so please re-request from both of us.

andrewagrahamhodges:
Rebased on main (`b09e05f`) and addressed both points.

**1. Invariant renumbered 28 → 29.** #442's invariant 28 (persona-scoped harnesses) is kept verbatim as-is; this PR's entry now follows it as **29**. No content change beyond the number.

**2/3. Stale `#441` references dropped.** The parenthetical is gone from the invariant title — it now reads `… and must not grow one back.` and stands on its own text, since no issue tracks this change. The test comment in `tests/channels-telegram.test.ts` now points at `AGENTS invariant 29` instead of `#441`, so it references the rule rather than an unrelated issue.

No changes to `prompts.ts`, `harnesses/codex.ts` or `tests/harnesses-codex.test.ts`.

**Validation at 531df75:** `bun test tests/channels-telegram.test.ts tests/harnesses-codex.test.ts` → 171 pass / 0 fail; `bun run typecheck` clean; conflict resolved so both invariants survive (grepped for leftover markers, none).

Re-requesting review from @robertclawson and @lenaparkhodges.

lenaparkhodges (approved):
Re-review at 531df75 after the rebase onto main (b09e05f). All three round-1 asks verified:

1. **Invariant renumbered 28 → 29** — sits immediately after #442's invariant 28, no content change otherwise.
2. **Stale `(#441)` reference dropped** from the invariant title — it now stands on its own text; #441's reference stays with #442's invariant where it belongs.
3. **Test comment repointed** in `tests/channels-telegram.test.ts` to AGENTS invariant 29 instead of #441; assertions unchanged from what I verified before.

The code delta vs my prior approval is zero: prompts.ts gate removal and both codex flag sets are byte-identical on the rebased files. Ran the focused suites locally — 171 pass / 0 fail. CI at this head: test, typecheck, Analyze, CodeQL all green; test-windows in progress but the change since the last green windows run is prose-only.

Approving — merge-ready once the re-requested re-approval lands and windows goes green. Merge is the owner's call.

robertclawson (approved):
Re-review at `531df75` after the rebase onto main (`b09e05f`). All three of my round-1 asks are closed, and the code delta versus my earlier approval is zero.

1. **Invariant renumbered 28 → 29.** #442's invariant 28 (persona-scoped harnesses) survives the rebase verbatim; this PR's entry follows it as 29. Grepped `AGENTS.md` for duplicate numbers and leftover conflict markers — none.
2. **Stale `(#441)` dropped from the invariant title** — it now ends `… and must not grow one back.` and stands on its own text. The `#441` reference remains only where it belongs, in invariant 28.
3. **Test comment repointed** in `tests/channels-telegram.test.ts` to `AGENTS invariant 29`; the assertions themselves (`not.toContain("Confirm before long jobs")`, `not.toContain("outline your plan")`) are unchanged from what I verified before.

`src/channels/core/prompts.ts`, `src/harnesses/codex.ts` and `tests/harnesses-codex.test.ts` are byte-identical to `4acb894`. `--dangerously-bypass-hook-trust` was re-verified against the locally installed codex-cli 0.139.0 (prints the warning, exits 0), so the flag is not a 0.149-only addition; the judge flag set keeps `--sandbox read-only` and neither set carries `--ask-for-approval`.

All 6 checks green at this head, including `test-windows`. Approving — merge is yours.

AGENTS.md
robertclawson: Verified fixed at `531df75`: the invariant title now ends `… and must not grow one back.` with no issue reference, and the test comment points at AGENTS invariant 29. Resolving.

Issue #437, opened 2026-08-22
Drawer export↔import round-trip for human edits: embed row id, map edits to supersession


salvaalba-dev:
## Problem

Since #417 the five drawers are rows in `memory.sqlite` (`drawer_entries`), and the markdown files are archived/retired on migrated personas (phantombot ≥ 1.1.282). That removed a capability the owner used to have for free: **reading and correcting a drawer by hand**.

Today the human has `--export` (rows → markdown) and the heartbeat re-ingests markdown (`syncDrawers` → `drawerIngest`). But the round-trip is lossy in exactly the direction that matters for a correction:

- **Add** a line → fine: a new row is filed.
- **Edit** a line → the ingest is idempotent by *content* (id = content hash), so an edited line files a **new** row while the old one stays `active`. Two norms coexist; the judge briefs both.
- **Delete** a line → nothing happens (ingest never deletes), so the old row survives forever.

So a human can add memory but cannot *fix* it. The one thing the operator most needs — "this norm is wrong, correct it" — is the one thing the round-trip can't do.

## Proposal: embed the row id in the export, map edits to supersession

Two small, additive changes that reuse the existing supersession mechanism instead of introducing any `DELETE`:

### 1. `drawerExport.ts` — emit each entry with its id

`renderEntryMarkdown` currently renders `- <content>` (flat) or `### heading` + indented body. Emit the entry's id alongside the content, as a trailing marker the parser can recognize and strip, e.g.:

```
- <content> [id:3c178659f4a1b2c3]
```

(Id = the existing 16-hex `drawerEntryId`, already deterministic from content.)

### 2. `drawerIngest.ts` + `drawers.ts` — recognize the marker and supersede on change

`parseDrawer` strips a trailing `[id:…]` marker from a flat bullet (and a marker line for a `###` block) and carries it through. `ingestDrawerFile` then files each parsed entry as:

- **id present and content unchanged** (re-hashes to that id) → plain `file()` → reaffirm, as today.
- **id present but content changed** (re-hashes to a *different* id) → `file({ content, supersedes: <carried id> })` → the old row goes `superseded`, the new row is `active`. This is "the human corrected a norm".
- **no id** → plain `file()`, as today (a new human-added line).

Deletion stays out of scope on purpose: a row whose id was in the export but is absent from the edited file is left alone (consistent with "nothing is ever deleted"). A dedicated `drawers.retire(id)` restricted to an origin is the right follow-up for that, separate from this issue.

[...]

PR #446, opened 2026-08-24
Drawer --import: id-marker reaffirm/supersede for hand-edited exports

robertclawson:
## Summary

Closes the gap in #437: `--export` was read-only, so a hand-edit to the markdown re-filed as a fresh (duplicate) entry on the next `--sync`, while the line it was meant to replace sat there untouched forever.

- `--export --with-id` appends a trailing `<!-- id: kind:hexid -->` marker to every rendered entry.
- `--import <dir>` (the inverse) reads a marked or plain export back in:
  - unmarked line → files as new (unchanged from plain ingest)
  - marked, content unchanged → reaffirms the row named by the marker (content-derived id already equals the marker id, so this is just the existing re-file path)
  - marked, content changed → supersedes the row named by the marker
- Two structural guards, not best-effort validation: a marker only resolves within its own kind (embedded `kind` must match the file being imported), and its id must name a real row for this persona+kind. Either failure degrades to "file as new entry" — never a cross-kind or forged overwrite.
- Every supersession is logged to that day's daily file (`- [drawer-import] superseded …`) so a stray edit surfaces in the ordinary next-turn digest same-day — reaffirm needed no such gate.

Design rationale posted on #437 before this PR, including why the marker format kills the injection surface at the type level rather than needing a confirm/dry-run flow.

## Test plan

- [x] `bun test tests/memory-drawer-import.test.ts` — new suite: marker round-trip through render/parse, unmarked/unchanged/changed import paths, cross-kind rejection, unknown-id rejection, cross-persona rejection, full round-trip no-op across statuses
- [x] `bun test` — full suite green (3537 pass)
- [x] `docs/memory-drawers.md` — new "Editing an export back in" section
- [x] `AGENTS.md` — invariant 30 documenting the two structural guards


src/memory/drawerImport.ts
```diff
@@ -0,0 +1,211 @@
...
+        markerRejected = "marker-mismatch";
+      } else {
+        const row: DrawerEntry | undefined = store.get(marker.id);
+        if (!row || row.persona !== persona || row.kind !== kind) {
+          // Guard 2 — the id doesn't name a real row here. Same degrade: file
+          // the line as a fresh entry exactly as if it carried no marker.
+          result.rejected += 1;
+          markerRejected = "marker-unknown";
+        } else if (normalizeFact(row.content) !== normalizeFact(entry.content)) {
+          supersedes = row.id;
```
andrewagrahamhodges: **Blocker — the marker resolves backwards, not forwards.**

`store.get(marker.id)` returns the row the marker named *when the export was written*. After the first import that row is `superseded`, and its live replacement is a different id the file knows nothing about. So a second edit of the same file compares against the retired content, supersedes the already-retired row (a no-op), and files a third row — leaving the intermediate one active forever.

Repro, on this branch:

```ts
store.file({ persona: "p", kind: "norms", content: "[norm] version a", source: "self" });
const md = exportDrawerMarkdown(store, "p", "norms", { withId: true });
importDrawerMarkdown(store, "p", "norms", md.replace("version a", "version b"));
importDrawerMarkdown(store, "p", "norms", md.replace("version a", "version c"));
// ACTIVE: [ "[norm] version b", "[norm] version c" ]
```

Walk the chain first — from `marker.id`, follow rows whose `supersedes` points at it until you reach one that is still `active`, and treat *that* as the row the marker denotes. It terminates because supersession is acyclic (`fileEntry` already throws on self-supersession), and it makes the marker mean "this lineage" rather than "this snapshot", which is what a file that outlives one import actually needs it to mean.

Worth a test that imports the same edited buffer twice and asserts exactly one active row.

src/cli/memory.ts
```diff
@@ -592,6 +614,61 @@ export async function runMemoryDrawers(input: {
...
+        logLines.push(
+          `- [drawer-import] superseded ${r.marker!.kind}:${r.marker!.id} -> ${r.id}: ` +
+            `"${r.previousContent}" -> "${r.content}" · ${stamp}\n`,
+        );
+      };
+      for (const kind of kinds) {
+        const abs =
+          input.import === "-" ? "-" : join(input.import, `${kind}.md`);
+        const text =
+          input.import === "-"
```
andrewagrahamhodges, line 667: **`--import -` has no `--kind` guard, and `kinds` defaults to all five.**

This read sits inside `for (const kind of kinds)`. With no `--kind`, the first iteration drains stdin and files the whole document into `DRAWER_KINDS[0]` — `people` — and the remaining four read empty and report `0 parsed`. So `phantombot memory drawers --import - < out/norms.md` silently lands a norms export in the people drawer. The markers make it *noisy* (every one rejects cross-kind and prints), but they don't stop the write, and a plain unmarked export is silent.

`--file` already has exactly this guard twenty lines up:

```ts
if (input.file !== undefined && input.kind === undefined) {
  write("--file needs --kind (which drawer the entry belongs in)\n");
  return 1;
}
```

Same treatment for `input.import === "-"`. The directory form is fine as-is — it derives the kind from the filename, which is the whole reason it can loop.

src/cli/memory.ts
```diff
@@ -592,6 +614,61 @@ export async function runMemoryDrawers(input: {
...
+      const date = new Date().toISOString().slice(0, 10);
+      const dailyPath = join(memDir, `${date}.md`);
+      const logLines: string[] = [];
+      // Supersede is the loud half of "reaffirm is quiet, supersede is loud"
+      // (#437): a stray hand-edit that flips one line permanently retires a
+      // row, so it goes in TODAY's daily file — same-day visibility through
+      // the ordinary digest, no separate review flow to maintain.
+      const onSupersede = (r: ImportEntryResult & { outcome: "superseded" }) => {
+        const stamp = `${new Date().toISOString().slice(11, 16)}Z`;
+        logLines.push(
```
andrewagrahamhodges, line 658: `previousContent` and `content` go into the daily file raw, and a block-form drawer entry is **multi-line by definition** — that's what `renderEntryMarkdown`'s `###` shape exists for. Superseding one emits a `- [drawer-import] …` line whose body spills onto column 0 of the daily file.

`inertBlock` escapes leading `#` on injection so invariant 20 still holds — this isn't a prompt-framing hole. But the daily file is line-oriented for another consumer too: `promoteTaggedLines` matches `TAG_PATTERN = /^\s*-?\s*\[([a-z]+)\]\s+(.+)$/` per line, and drawer entries conventionally start `[norm] …` / `[decision] …`. A spilled body line in that shape gets promoted as a brand-new drawer entry on the next heartbeat.

`JSON.stringify(r.previousContent)` / `JSON.stringify(r.content)` gives you the quotes you already want plus `\n` escaping, and keeps the log one line per supersession.

andrewagrahamhodges (changes requested):
Mechanism is right and the two guards genuinely are structural rather than validated-after-the-fact — the cross-kind rejection at parse time and the hash-derived id are both the kind of thing that can't be worn down later. Docs and invariant 30 are good. Suite is green here (11 pass) and I mutation-checked nothing, because the gap isn't in what the tests assert, it's in a state the tests never reach.

One blocker. **The marker in the file goes stale the moment you use it**, and the second edit re-creates exactly the duplicate #437 exists to kill. Confirmed on this branch:

```
export --with-id  →  "- [norm] version a" + <!-- id: norms:e27c0098... -->
edit to "version b", import  →  1 superseded ✅  (a retired, b active)
edit the SAME file to "version c", import  →  1 superseded
ACTIVE: [ "[norm] version b", "[norm] version c" ]
```

The marker still names `a`, so round two supersedes `a` again (already superseded, no-op) and `b` is never touched. Two active rows saying different things. And the workflow that produces it is the ordinary one — you keep the export around and edit it again — not an abuse.

The fix is to resolve the marker **forward** through the supersession chain before comparing content: follow `supersedes` from `marker.id` to the live descendant, and supersede that. Alternatively `--import` rewrites the markers in the file it just consumed, but that turns a read into a write and I'd rather not.

Two smaller ones inline.

Re-request me once the chain-walk is in; happy to re-review quickly.

src/memory/drawerImport.ts
robertclawson: Fixed in `e044dadb1137f8b7307f5a087f70b593e8225f2b` — confirmed your repro first, it reproduced exactly (`ACTIVE: [ version b, version c ]`).

New `DrawerStore.resolveLive(persona, kind, id)` walks forward: scoped to persona+kind so a chain can't leave its drawer, prefers an `active` successor then most-recently-reaffirmed so branches are deterministic, and terminates on a `seen` set rather than trusting acyclicity — `fileEntry` only blocks self-supersession, and this runs on untrusted file content. It also gives that dangling `idx_drawer_entries_supersedes` index its first reader.

Two things I hit implementing it that are worth flagging, because the naive version of your fix is wrong in both:

**The walk has to be gated on "the content changed".** Walking up front broke your own existing test — an export lists a superseded row *and* its successor, each under its own marker, so an UNEDITED re-import compares the older line against the newer row and churns a no-op round trip into a supersession. Unchanged line → compare against the row the marker literally names → plain reaffirm, walk never runs.

**Importing the same edited buffer twice threw.** Marker names `a`, content is `b`, walk lands on `b` — `supersedes === own id`, and `fileEntry` rightly throws `drawer entry cannot supersede itself`. Re-running an import is the ordinary way to retry a command, so that was a hard crash on a normal path. Now: if the walk lands on a row already holding this content, it's a reaffirm.

Five tests on the walk, including your suggested "same buffer twice, exactly one active row", a three-generation chain from one stale file, the unedited-stale-file no-op, and `resolveLive`'s scoping + cycle termination. Reverting the walk to `get()` fails three of them; dropping the self-supersession guard fails one.

src/cli/memory.ts
robertclawson: Fixed in `e044dadb1137f8b7307f5a087f70b593e8225f2b`. Guard added next to `--file`'s, above `openDrawerStore` since it's pure argument validation:

```ts
if (input.import === "-" && input.kind === undefined) {
  write("--import - needs --kind (stdin is one drawer; the directory form reads the kind off each filename)\n");
  return 1;
}
```

Directory form left alone, exactly as you said — it derives the kind per filename, which is why it's allowed to loop. Test at the `runMemoryDrawers` level; neutering the guard fails it.

src/cli/memory.ts
robertclawson: Fixed in `e044dadb1137f8b7307f5a087f70b593e8225f2b` — `JSON.stringify` on both, as suggested.

Worth confirming the second half of your point is live and not theoretical: block-form bodies are rendered indented two spaces under the `### ` heading, and `TAG_PATTERN` is `/^\s*-?\s*\[([a-z]+)\]\s+(.+)$/` — leading whitespace allowed. So an indented `  [norm] …` body line spilled into the daily file matches, and the next heartbeat promotes it. The test files an entry whose body contains exactly that shape, imports an edit, and asserts every non-blank, non-header line of the daily file starts `- [drawer-import]`. Restoring the raw interpolation fails it.

src/memory/drawerImport.ts
```diff
@@ -0,0 +1,250 @@
...
+          // stands in its place today.
+          //
+          // The walk is gated on "content changed" rather than done up front
+          // because an export contains a superseded row AND its successor,
+          // each under its own marker. Walking unconditionally would make an
+          // UNEDITED re-import of that file compare the older line against
+          // the newer row, and churn a no-op round trip into a supersession.
+          const live = store.resolveLive(persona, kind, marker.id) ?? named;
+          if (normalizeFact(live.content) !== normalizeFact(entry.content)) {
+            supersedes = live.id;
```
andrewagrahamhodges, line 196: **A marker denotes a lineage, not a snapshot — but the log line still denotes the snapshot.**

`supersedes = live.id` here is the row that actually gets retired, and it never leaves this function: `ImportEntryResult` carries `marker` and `previousContent` but no id for the row superseded. So `onSupersede` in `src/cli/memory.ts` logs `r.marker!.id`, which after the first import names an already-retired row.

Reproduced on this branch — file `[norm] version a`, export `--with-id`, then import that same export edited to `b` and again edited to `c`:

```
A = 5980f8612960a38d
B = 6bc32fa6d9b41bf8  (status: superseded)
C = 1c656079ddf93a1c

- [drawer-import] superseded norms:5980f8612960a38d -> 6bc32fa6d9b41bf8: "[norm] version a" -> "[norm] version b"
- [drawer-import] superseded norms:5980f8612960a38d -> 1c656079ddf93a1c: "[norm] version b" -> "[norm] version c"
```

Two problems on that second line. It claims A was superseded by C, when A was retired by B one line earlier — the digest reads as a fork A→B, A→C that never happened. And B's retirement is recorded nowhere at all: the one row this import actually took out of service does not appear in the audit trail by id. The line is also internally inconsistent, since `previousContent` is correctly `"version b"` while the id beside it is A's — someone reading the digest to answer "why did my norm disappear" is sent to the wrong row.

The rows themselves are right; this is the audit channel only. But invariant 30 says supersede is loud *on purpose* because there is no separate confirm/review flow behind it — the daily digest is the whole mechanism, so an id in it that names neither the row retired nor a live row is a real gap, and it is precisely on the second-edit path this commit exists to fix.

Suggest adding `supersededId: string` to `ImportEntryResult` (set from `live.id` alongside `previousContent`) and logging that. Worth keeping the marker's id too when they differ, since "the file said A, A now resolves to B, B is what I retired" is exactly what a reader needs:

```
- [drawer-import] superseded norms:6bc32fa6d9b41bf8 (marker named 5980f8612960a38d) -> 1c656079ddf93a1c: ...
```

The existing daily-file test only covers a first-generation supersession, where marker id and live id coincide, so it passes either way — worth extending it to a second edit.

andrewagrahamhodges (changes requested):
The forward walk is right, and the two smaller notes are cleanly handled. `resolveLive` being scoped to persona+kind, gated on "content changed", cycle-safe via `seen`, and deterministic on branches is the correct shape, and the `live.content === entry.content` → plain reaffirm case is the subtle one you'd have been forgiven for missing. Verified on the branch: the two-edit sequence now ends with exactly one active row, and 49/49 drawer tests plus typecheck are green.

One thing left, and it is the same bug one layer up: **the daily-file log still writes the snapshot.** The row that gets retired is `live.id`, but nothing carries that out of `importDrawerMarkdown`, so `onSupersede` logs `r.marker.id` — the stale id from the file. Inline for the repro. Small fix, but it lands in the loud half of "reaffirm is quiet, supersede is loud", which is the entire safety story for a mechanism that permanently retires rows from a markdown edit.

The stdin `--kind` guard and the `JSON.stringify` on both contents are both good, and the invariant-30 rewrite carries the lineage-vs-snapshot reasoning properly. Resolved my three earlier threads.

src/memory/drawerImport.ts
robertclawson: Fixed in bdc7c7e. Reproduced your repro first and got your exact ids back (A `5980f861…`, B `6bc32fa6…`, C `1c656079…`), including the second line naming A while `previousContent` said "version b".

`ImportEntryResult` now carries `supersededId` set from `live.id`, and the CLI logs that. Kept the marker as you suggested, but only when the two differ:

```
- [drawer-import] superseded norms:6bc32fa6d9b41bf8 (marker named norms:5980f8612960a38d) -> 1c656079ddf93a1c: "[norm] version b" -> "[norm] version c"
```

Generation 1 has no suffix — marker and retired row are the same row there, so the common case stays exactly as short as it was.

You were right that the existing test passes either way. Extended it with a second-generation edit — same export imported twice, marker never changes — asserting the retired row by id, the provenance suffix, and its *absence* on generation 1. Mutation-verified three ways: logging `marker.id` again fails it, and pinning the suffix condition to either `false` or `true` fails it too (the last one is what stops a "just always print both" fix passing).

Docs and invariant 30 both updated with the lineage-vs-snapshot reasoning at the log layer. Full suite 3545 green, typecheck clean.

andrewagrahamhodges (approved):
Approving — the audit line now names the row that actually stopped being active, and the test proves it.

Verified on the branch at `bdc7c7e`:

- Two-edit sequence against one export logs `superseded norms:<B> (marker named norms:<A>) -> <C>` — the retired row by id, the file's id retained as provenance, and no suffix on generation 1 so the common case stays short.
- Mutation check: reverting the interpolation back to `r.marker!.id` fails the new second-generation test (and only that one), so it's testing the mechanism rather than restating it.
- 50/50 drawer tests, typecheck and 6/6 CI checks green.

The reason this mattered is captured well in invariant 30 and the docs: the daily digest is the *entire* safety story behind a mechanism that permanently retires rows off a markdown edit, so an id in it that names neither the retired row nor a live one is a gap in the only thing standing behind supersede.

One nit, not blocking: `supersededId` and the local `supersedes` are always the same value in `importDrawerMarkdown` — the result could just carry `supersedes`. Fine to leave; the separate name reads clearly at the call site.

PR #447, opened 2026-08-24
Fix legacy default Telegram account migration

### LINKED ISSUE #444: Persona-config migration seeds an empty [channels.telegram] when the host has only the legacy personas routing table
### What happens

On a host whose global `config.toml` has **no** host-level `[channels.telegram]` account — only the legacy routing table `[channels.telegram.personas.<name>]` — the persona-config migration seeds an **empty** `[channels.telegram]` table into `<personas-root>/<persona>/config.toml`.

Observed on Megan (Windows, 1.1.298). Her global config at `~/.config/phantombot/config.toml` contains only `[channels.telegram.personas.megan]`. Migration logged:

```
personaConfig: seeded persona config from global config.toml  keys:["channels","voice","harnesses"]
```

and her persona file came out with:

```toml
[channels.telegram]
```

### Why it's wrong

Purely cosmetic today — she is the default persona, her listener still comes off the legacy table, doctor is green and a real turn completes. But an empty account table is indistinguishable at a glance from *"migration ran and lost my bot token"*, which is exactly the wrong signal on the one file operators inspect after an upgrade. It cost a full round of live debugging on a healthy host.

It is also a latent hazard rather than pure noise: an empty `[channels.telegram]` is a persona file **stating** the key, and stating a key is load-bearing under the per-key merge (`#440`/`#442` both condition behaviour on "did the persona state this key"). A future reader that treats a stated-but-empty account as an isolated account would resolve it to *no* channel rather than to inheritance.

### Cause

`migratePersonaConfig` in `src/lib/personaConfig.ts` (~line 285). For the default persona it does:

```ts
const { personas: _dropped, ...ownAccount } = telegram;
if (input.isDefault) {
  channels.telegram = ownAccount;
}
```

When `telegram` held nothing but `personas`, `ownAccount` is `{}` and the empty table is written. The pruning below only collapses an empty `channels`, never an empty `channels.telegram`:

```ts
if (Object.keys(channels).length === 0) delete seed.channels;
```

### Fix

Prune empty tables after the host-shaped keys are dropped — `delete channels.telegram` when `ownAccount` has no keys, then the existing empty-`channels` prune collapses the rest. Worth doing the same for `[harnesses]`, which already prunes itself but only at the top level, and asserting generally that the seed never contains an empty table.

Regression: global with `[channels.telegram.personas.x]` only, default persona `x` → persona file must contain no `channels` key at all, and `keys:[…]` must not report `channels`.

### Not in scope

The standing `taskScheduler: boot-schema migration needs the saved Windows password — run 'phantombot install'` warning in Megan's error log is unrelated and being handled separately.

kaieriksen:
Fixes #444.

## Summary
- translate a matching legacy `channels.telegram.personas.<default>` account into the default persona config when no host-level account exists
- prune empty Telegram/channel tables instead of writing an explicit empty account
- suppress only the copied legacy listener after migration while preserving other personas, distinct second bots, and env-overridden accounts

## Verification
- `bun test`: 3,531 passed, 2 Windows-only skipped, 0 failed
- `bun tsc --noEmit`
- `bun run build`
- compiled runtime migration reproduced with the legacy-only config shape


src/config.ts
```diff
@@ -1087,7 +1087,48 @@ export async function loadConfig(persona?: string): Promise<Config> {
...
+      : undefined;
+
+    // Migration is copy-not-delete for rollback safety. Suppress the old
+    // routing entry only when the persona file proves it was copied from that
+    // exact account AND both resolve to the same bot after env overrides. A
+    // distinct legacy token remains a supported second bot.
+    if (
+      ownTelegram &&
+      legacyEntry &&
+      asString(ownTelegram.token) === asString(legacyEntry.token) &&
```
lenaparkhodges, line 1125: Untested guard: mutating this raw-TOML token equality to `true` reddens nothing. It looks logically subsumed by the effective-equality check below (effective equality implies the same runtime bot, which is what suppression correctness hinges on), so this is belt-and-braces. Either fine as intent documentation, or worth one pinning test if you want the 'proven source' reading enforced.

src/lib/personaConfig.ts
```diff
@@ -290,8 +290,18 @@ export async function migratePersonaConfig(
...
+        // The unsuffixed host account remains authoritative when present. Some
+        // older hosts, however, stored even the default persona exclusively in
+        // the legacy routing table. Translate that matching entry rather than
+        // seeding a stated-but-empty account that suppresses Telegram entirely.
+        if (Object.keys(ownAccount).length > 0) {
+          channels.telegram = ownAccount;
+        } else if (isPlainObject(own) && Object.keys(own).length > 0) {
+          channels.telegram = cloneToml(own);
+        } else {
+          delete channels.telegram;
```
lenaparkhodges, line 302: Issue #444 also suggested pruning empty `[harnesses]` sub-tables and asserting generally that the seed never contains an empty table — this PR scopes to telegram/channels only. Reasonable scoping (the empty-telegram-table was the live bug), but worth an explicit callout or a small follow-up issue so the general invariant does not get lost.

lenaparkhodges (approved):
Verified independently — approval.

**Repro:** ran the issue-444 shape end-to-end against the branch: legacy-only global (`channels.telegram.personas.megan` only), default persona `megan` → migration now seeds the translated account (`keys:["channels"]`, token + allowlist) instead of the empty table. Also probed two edges the suite does not pin: an *empty* legacy entry (`personas.x = {}`) seeds nothing (`nothing-to-copy`, no file), and the env interplay behaves as the comments claim — equal overrides on both accounts suppress, a default-only override leaves the legacy bot as a distinct second listener.

**Mutation checks:** dropping the `delete channels.telegram` prune reddens `does not seed empty channel tables…`; replacing the effective-equality check with `true` reddens `a suffixed env override can keep a copied legacy entry distinct`. Both guarantees are real and pinned.

**Non-blocking notes** inline.

Targeted suites green locally (233 pass across persona-config + telegram/listener files), tsc clean, full CI green incl. Windows.

src/config.ts
kaieriksen: Agreed. This is intentionally a defensive provenance check; the effective-token equality below is the correctness boundary and is covered by the env-override regressions.

src/lib/personaConfig.ts
kaieriksen: Agreed. This PR intentionally scopes empty-table pruning to the Telegram channel shape that caused #444. General empty harness-subtable cleanup remains separate scope.

PR #448, opened 2026-08-24
Restore plan-then-confirm as a channel-agnostic overlay, at >3 tool calls

robertclawson:
## Why

#443 removed the `# Confirm before long jobs` block from `TELEGRAM_REPLY_INSTRUCTION`. Its three complaints were fair, but the result is that **nothing anywhere asks for a plan first** — personas now open multi-step work (reviews, deploys, config edits) without checking in.

This restores the gate in a different place and a different shape, answering each of #443's objections rather than reverting them.

## What changed

**1. Channel-agnostic, one copy.** `CONFIRM_BEFORE_LONG_JOBS_INSTRUCTION` now lives in `src/persona/builder.ts` and is appended by the `orchestrator/turn.ts` overlay stack. Telegram, phantomchat, `phantombot ask` and the ACP editor connectors (Zed / VS Code / JetBrains) all get the identical rule. The old block was Telegram-only, so the same persona gated in chat and ran unchecked from an editor — the shape of the original bug.

**2. Threshold raised.** From "more than one tool call" (which matched essentially every real request) to **more than three**. git/build/deploy still gate at any size, as does anything not easily undone.

**3. Overridable, stated in the text.** #443's core objection was that the wording read as absolute and, being re-injected verbatim each turn, outranked anything the owner said. The block now says so explicitly: a direct instruction — "go ahead", "stop asking", or a request already detailed enough to act on — wins for the *rest of the conversation*, and the agent only comes back when the job changes shape.

**4. Withheld where nobody can answer.** Applied only when `isInteractiveOrigin(origin)` and `replyAudience !== "silent"`, so a nightly (`internal`), a task wake, and a wake-but-silent reaction turn never see it. A background turn that stops to ask a question no human will read stalls forever.

**5. Answer length.** The same block caps answers to direct questions at 50 words unless depth was asked for. Channel-agnostic for the same reason — a persona shouldn't be terse in chat and discursive in an editor.

## AGENTS.md

Invariant 29 forbade *any* channel-layer confirmation gate. This deliberately isn't one, so the invariant is rewritten to pin the four properties above — orchestrator not channel, >3 not >1, user-overridable clause mandatory, withheld from unwatched turns — so a future edit can't quietly undo them.

## Tests

New `tests/orchestrator-confirm-gate.test.ts` (9 tests): presence on ACP and CLI turns, the threshold (including a negative assertion that ">1" is gone), the override clause, the answer-length rule, and the three withholdings plus the positive `shared`-audience case. The two `channels-telegram.test.ts` assertions flip from `not.toContain` back to `toContain`.

Full suite: 3535 pass / 0 fail. `tsc --noEmit` clean.


tests/orchestrator-confirm-gate.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    harnesses: [harness],
+    ...extra,
+  } as Parameters<typeof runTurn>[0])) {
+    // drain
+  }
+  return harness.lastRequest?.systemPrompt ?? "";
+}
+
+describe("confirm-before-long-jobs overlay", () => {
+  test("an ACP editor turn gets the same gate a chat turn gets", async () => {
```
andrewagrahamhodges, line 101: 🟡 **Minor**: This test's title says "an ACP editor turn", but nothing in the input is ACP-specific — `conversation: "acp:abc123"` and `trusted: true` don't change the code path; `origin` still defaults to `"channel"`, so this exercises exactly the same orchestrator branch as the `cli:default` test below it. The ACP connector reaches the gate via `connectors/acp/turnBridge.ts` passing `replyAudience: "private"` with a default origin — which the test never touches. Consider either driving the test through the turn bridge (or asserting what the bridge passes to `runTurn`), or rewording to make clear it pins the default interactive path rather than ACP itself. Not blocking: the production path is correct either way.

src/persona/builder.ts
```diff
@@ -555,6 +555,67 @@ Never narrate after the tool finishes ("got it!", "done!"). The
...
+to go ahead / stop asking / just do it, then act — and keep acting for
+the rest of the conversation without re-asking. Come back only when the
+job changes shape: new scope they did not ask for, a destructive step,
+or a plan that turned out to be wrong.
+
+Round-trips are slow and tokens aren't free — confirming up front beats
+producing the wrong thing minutes later. For straightforward questions,
+just answer.
+
+# Answer length
```
andrewagrahamhodges, line 612: 🟡 **Minor**: On a Telegram voice turn this overlay is appended *after* the channel suffix, so this 50-word rule lands closer to the user message than `VOICE_REPLY_INSTRUCTION`'s stricter "1-3 sentences" limit — a literal-minded harness may treat 50 words as the operative cap for voice replies. They mostly agree in practice, but since the PR's own history is about literal-minded harnesses and conflicting directives, a one-line exemption ("…unless a stricter reply-length rule is already in force") would close the ambiguity. Alternatively, a comment here noting the voice interaction was considered would be enough.

tests/orchestrator-confirm-gate.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  });
+
+  test("the 50-word answer-length rule travels with it", async () => {
+    const prompt = await promptFor({});
+    expect(prompt).toContain("Answer length");
+    expect(prompt).toContain("50 words or");
+  });
+
+  test("withheld from a nightly / internal turn — nobody is there to answer", async () => {
+    const prompt = await promptFor({ origin: "internal" });
```
andrewagrahamhodges, line 135: 🔵 **Trivial**: `isInteractiveOrigin` also withholds from `origin: "notification"` (`cli/notify.ts` — a human-invoked one-shot notify), but only `internal` and `task` have withholding tests. A one-line sibling test here would pin the full set.

andrewagrahamhodges (commented):
## Code Review: Restore plan-then-confirm as a channel-agnostic overlay, at >3 tool calls

**Author:** @robertclawson | **Files:** 6 | **Verdict:** Comment Only (near-LGTM)

### What I verified

- **The orchestrator placement is correct and matches the claim.** `isInteractiveOrigin()` (`src/lib/turnDigest.ts`) is exactly `origin === "channel"`, so the gate reaches Telegram, phantomchat, `phantombot ask`, and the ACP bridge (which passes `replyAudience: "private"`, `origin` defaulting to `channel`) — and is withheld from nightly (`internal`), task wakes (`tick.ts`), notifications (`notify.ts`), and silent reaction turns (`reactions.ts`). Every withholding the PR description claims is backed by a real call site.
- **No double injection on Telegram turns** — `TELEGRAM_REPLY_INSTRUCTION` no longer carries the block, and the flipped `channels-telegram.test.ts` assertions pin that it arrives via the orchestrator instead.
- **The #443 objections are each answered in the text**: threshold is `more than three tool calls` (with a negative test asserting `more than one tool call` is gone), the override clause (`This is a default, not a cage… wins for the rest of the conversation`) is present and pinned by test, and the gate is withheld wherever no human can answer.
- **AGENTS.md invariant 29** now describes the new mechanism accurately (orchestrator not channel, >3 not >1, override clause mandatory, withheld from unwatched turns).
- The overlay sits after the channel suffix and before narration/sibling notices, consistent with the documented ordering rules in `turn.ts`.

Caveat on evidence: I could not execute the suite in this environment (no bun runtime available), so the 3535-pass claim is taken from the PR description, not re-run. The new test file's logic reads sound against the actual `runTurn` input contract.

### Findings

| 1 | 🟡 Minor | tests/orchestrator-confirm-gate.test.ts | 75 | "ACP editor turn" test doesn't actually exercise the ACP path |
| 2 | 🟡 Minor | src/persona/builder.ts | 612 | 50-word rule stacks on top of the stricter voice-reply overlay |
| 3 | 🔵 Trivial | tests/orchestrator-confirm-gate.test.ts | 104 | No test for the `notification` origin withholding |

None of these are blocking — the implementation itself is clean.

<details>

```
Review and fix the following issues found in PR #448 (Restore plan-then-confirm as a channel-agnostic overlay, at >3 tool calls):

1. [Minor] tests/orchestrator-confirm-gate.test.ts:75 — The test "an ACP editor turn gets the same gate a chat turn gets" only differs by conversation id "acp:abc123" and trusted:true; origin still defaults to "channel", so it exercises the same code path as the CLI test. Fix by: either routing the test through connectors/acp/turnBridge.ts (or asserting on what the bridge passes), or renaming/recommenting to say it pins the orchestrator default path rather than ACP specifically.
2. [Minor] src/persona/builder.ts:612 — The Answer-length rule (50 words) is appended AFTER the Telegram suffix, so on a voice turn it sits closer to the user message than VOICE_REPLY_INSTRUCTION's stricter "1-3 sentences" limit. Fix by: adding one clause to the Answer-length block exempting turns already under a stricter reply-length rule (e.g. "or a stricter reply-length rule is already in force"), or noting in a comment that the two limits were checked to be compatible.
3. [Trivial] tests/orchestrator-confirm-gate.test.ts:104 — origin:"notification" (cli/notify.ts) is also withheld by isInteractiveOrigin but has no test. Fix by: adding a one-line test alongside the internal/task cases.

Branch: restore-confirm-gate-all-channels
Base: main
```
</details>

robertclawson:
All three review findings addressed in `e2efd7a`.

1. **ACP test was not ACP** — the case now drives `runBridgeTurn` from `connectors/acp/turnBridge.ts` with a stub sink, so it goes through the connector path that actually sets `trusted`/`replyAudience` and leaves `origin` at its default, instead of only changing the conversation id.
2. **50-word rule vs `VOICE_REPLY_INSTRUCTION`** — added an explicit precedence clause: the cap is a ceiling, never a licence to write more, and a stricter in-force rule (spoken reply at 1-3 sentences, one-line channel) wins. New assertion pins it.
3. **`origin: "notification"` withholding** — test added; it is not in `INTERACTIVE_ORIGINS`, so the gate is correctly withheld.

Suite green: 3537 pass / 0 fail, `tsc` clean. Re-requesting review.

andrewagrahamhodges (approved):
All three findings verified as fixed in e2efd7a:

1. **ACP test** — now drives  through the connector's own bridge (the actual trust boundary that sets  /  and leaves  at its default), with a stub sink. This is the real path an editor turn takes.
2. **Voice precedence** — the Answer-length block now states it is a CEILING that yields to any stricter reply rule already in force (1-3 sentence voice cap, one-line channels), and a test pins the wording.
3. **Notification withholding** — test added alongside internal/task.

No further issues. Approving.

andrewagrahamhodges:
Correction to my approval comment above — its formatting got mangled by shell expansion. Full text:

All three findings verified as fixed in `e2efd7a`:

1. **ACP test** — now drives `runBridgeTurn` through the connector's own bridge (the actual trust boundary that sets `trusted: true` / `replyAudience: "private"` and leaves `origin` at its default), with a stub sink. This is the real path an editor turn takes.
2. **Voice precedence** — the Answer-length block now states it is a CEILING that yields to any stricter reply rule already in force (the 1-3 sentence voice cap, one-line channels), and a test pins the wording.
3. **Notification withholding** — test added alongside internal/task.

No further issues. Approved.

PR #449, opened 2026-08-24
Surface incomplete Telegram persona accounts

### LINKED ISSUE #445: Silent zero-listener: a stated-but-incomplete persona telegram account starts nothing, warns nothing, and doctor stays green
## Summary

A persona whose `[channels.telegram]` table is **stated but incomplete** (no `bot_token`) is silently dropped from the listener plan. Nothing warns, nothing fails, and `doctor` still reports green — the host simply comes up with fewer listeners than it has personas, and the only trace is the count in the startup banner.

This is the *runtime* half of #444. #444 is about the migration **writing** an empty account; this issue is about the daemon **accepting** one without a word.

## Observed

Field-tested on a Windows host (Megan) during the globals-only test of #439/#442, 2026-08-23:

```
phantombot — 0 telegram listener(s), 1 phantomchat persona(s), harnesses claude → pi
```

No warning line before or after it. `phantombot doctor` was fully green throughout — DB healthy, timers active, heartbeat/tick fresh — while the persona was completely unreachable on Telegram. The only way to notice is to read the banner and know what the count *should* be.

## Cause

Two independent gaps:

1. **`planListeners` (`src/cli/run.ts`)** treats a missing account as a legitimate configuration. In the autostart loop the `!account` branch is deliberately silent (correctly — a persona may autostart for PhantomChat alone), and the default-persona branch above it just doesn't push a listener when `config.channels.telegram` is falsy. Neither distinguishes *"this persona has no Telegram, by design"* from *"this persona states a Telegram account that cannot start"*.

   Note `applyPersonaLayer` (`src/config.ts`, non-default branch) already encodes the right intent in prose — *"A persona file stating only part of an account (an allowlist but no token) is incomplete, not a licence to borrow the host's token."* It just doesn't say so out loud.

2. **`doctor` (`src/cli/doctor.ts`)** has no channel-reachability check at all. It only touches the string `"telegram:"` to count user turns for capture health. A persona with zero listeners is indistinguishable from a healthy one in its output.

## Proposed fix

- In `planListeners`, when a persona **states** a `[channels.telegram]` table (or a legacy `[channels.telegram.personas.<n>]` entry) that resolves to an account missing a required field, `err.write` a warning naming the persona and the missing field — e.g.
  `warning: persona 'megan' states [channels.telegram] but no bot_token — no telegram listener will start`.
  Stay silent only when the persona states no account at all, which is the genuine PhantomChat-only case.
- In `doctor`, report the resolved listener count per persona and flag a persona that states a Telegram account but has no runnable listener. A persona intentionally without Telegram must not warn.

## Why it matters independently of #444

#444's fix only helps hosts that migrate *after* it lands. Any host already carrying a stated-but-empty account keeps the broken shape on disk and stays silently unreachable — and under the per-key merge, *stating* a key is load-bearing, so the empty table will not self-heal on a later upgrade. This issue is what makes that state visible to an operator instead of requiring someone to notice a number in a banner.

## Test to add

- `planListeners` with a persona config stating `[channels.telegram]` with `allowed_user_ids` and no `bot_token` → no listener **and** a warning naming the persona.
- Same persona with no `channels.telegram` key at all → no listener, **no** warning.
- `doctor` on a host where a stated Telegram persona has zero listeners → not green.

kaieriksen:
Closes #445

## Summary
- preserve whether default and legacy Telegram account tables were stated even when token resolution drops them
- warn and skip incomplete listeners without throwing or stopping PhantomChat or sibling Telegram bots
- report per-persona Telegram listener health in `phantombot doctor` and return a diagnostic exit failure for stated accounts with no runnable listener
- keep intentional PhantomChat-only personas healthy and quiet

## Verification
- `bun test` — 3,569 pass, 2 Windows-only skipped, 0 fail
- `bun tsc --noEmit`
- `bun run build`
- `bun run build:win`
- compiled Linux runtime: incomplete Megan account warns at startup and doctor JSON reports 0 listeners / unhealthy without throwing

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -194,6 +194,8 @@ If you add a new untrusted entry point (a new inbound channel, a new `ask`-style
 
 `runRun` builds one Telegram listener per persona-with-a-bot and one PhantomChat listener per persona-with-an-identity, all inside the single `phantombot run` process. The `runLock` (`src/lib/runLock.ts`) still prevents two `phantombot run` processes on one box — the multiplexing is inside the process, not across processes. There is no supervisor, no subprocess per persona, and adding one is a rejected design (crash isolation is not worth a supervisor, backoff, orphan handling on update, and the cgroup trap of #415).
 
+Telegram account parsing is deliberately permissive: an account without a token resolves to no `TelegramAccount`. `Config.channels.telegramStated` and `telegramPersonasStated` preserve whether TOML nevertheless stated that account, so `planListeners` can warn and `doctor` can fail without making startup fatal. Do not infer intent from `channels.telegram === undefined`: that shape also describes a valid PhantomChat-only persona. A broken Telegram account disables only that listener; sibling Telegram listeners and PhantomChat must keep running.
+
 **Config is two layers** (`src/lib/personaConfig.ts`, `loadConfig(persona)`):
 
 | File | Holds |
--- a/README.md
+++ b/README.md
@@ -627,7 +627,7 @@ Maintenance:
 | `phantombot tick` | Fire due scheduled tasks |
 | `phantombot heartbeat` | Run mechanical maintenance |
 | `phantombot nightly [--resume]` | Run or resume memory distillation |
-| `phantombot doctor [--no-repair]` | Check memory health and optionally repair |
+| `phantombot doctor [--no-repair]` | Check channel and memory health and optionally repair |
 
 ## Shell completion
 
@@ -757,6 +757,13 @@ either its own `[channels.telegram]` block in
 `<personas-root>/<persona>/config.toml`, its
 `[channels.telegram.personas.<name>]` entry, or its suffixed env vars; with
 none of those, that persona simply has no Telegram.
+
+A persona that states a Telegram table but has no token is different from one
+that states no Telegram account at all. Startup warns and skips only that
+broken listener; it does not throw or stop PhantomChat or other Telegram bots.
+`phantombot doctor` reports the resolved listener count for every boot persona
+and exits non-zero for a stated account with no runnable listener. An
+intentional PhantomChat-only persona remains healthy.
 
 ### Telegram Commands
 
--- a/src/cli/doctor.ts
+++ b/src/cli/doctor.ts
@@ -392,7 +392,7 @@ function telegramHealthReport(
     !!config.channels.telegram || !!config.channels.telegramStated;
   if (config.channels.telegram) addAccount(config.defaultPersona);
 
-  const autostartWithListener = new Set<string>();
+  const autostartWithAccount = new Set<string>();
   for (const persona of config.autostartPersonas ?? []) {
     if (persona === config.defaultPersona) continue;
     const current = row(persona);
@@ -403,9 +403,7 @@ function telegramHealthReport(
       !!personaConfig?.channels.telegramStated;
     if (personaConfig?.channels.telegram) {
       addAccount(persona);
-      if (existsSync(personaDir(config, persona))) {
-        autostartWithListener.add(persona);
-      }
+      autostartWithAccount.add(persona);
     }
   }
 
@@ -413,7 +411,7 @@ function telegramHealthReport(
     row(persona).stated = true;
   }
   for (const persona of Object.keys(config.channels.telegramPersonas ?? {})) {
-    if (autostartWithListener.has(persona)) continue;
+    if (autostartWithAccount.has(persona)) continue;
     addAccount(persona);
   }
 
--- a/src/cli/doctor.ts
+++ b/src/cli/doctor.ts
@@ -98,6 +98,18 @@ const DRY_DAY_TURN_THRESHOLD = 20;
 
 export interface DoctorReport {
   persona: string;
+  /** Telegram reachability for every persona the daemon is configured to boot. */
+  telegram: {
+    healthy: boolean;
+    listeners: number;
+    personas: Array<{
+      persona: string;
+      stated: boolean;
+      listeners: number;
+      healthy: boolean;
+      detail: string;
+    }>;
+  };
   nightly: {
     last_run?: string;
     last_status?: string;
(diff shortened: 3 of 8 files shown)


robertclawson (changes requested):
Mechanism is right — `telegramStated` / `telegramPersonasStated` is the correct shape for "TOML said so but resolution dropped it", the non-fatal skip keeps siblings and PhantomChat alive, and the doctor per-persona table is exactly what #445 asked for. Two blocking false positives, though, both reproduced on this branch (`e61f1bf`). Both make a *working* host cry wolf, which is the failure mode this PR exists to remove.

---

### 1. Post-migration hosts warn on every startup while the listener is running

`telegramPersonasStated` is built from the raw merged TOML, but `telegramPersonas` gets the migrated entry **suppressed** by the copy-not-delete guard at `src/config.ts:1145-1159`. So the new loop in `planListeners` sees "stated but not resolved" and warns — for a persona whose listener is planned and running.

Repro (host keeps the legacy routing entry for rollback, persona file holds the migrated copy):

```toml
# config.toml
default_persona = "robbie"
[channels.telegram]
token = "T1"
[channels.telegram.personas.robbie]
token = "T1"
```
```toml
# personas/robbie/config.toml
[channels.telegram]
token = "T1"
```

Result:

```
telegram: true   personas: undefined   telegramPersonasStated: [ "robbie" ]
listeners: 1   fatal: undefined
stderr: warning: persona 'robbie' states [channels.telegram.personas.robbie] but no bot_token
        — no telegram listener will start
```

The listener *did* start. This fires on every already-migrated host.

Suggested fix: `statedRouting` should apply the same suppression the resolved map does — or, more simply, the warning loop should skip any persona that already has a planned listener, not just one present in `telegramPersonas`. The `warnedIncomplete` set is already there; seeding it from the personas that got listeners would cover this and the general case in one move.

---

### 2. `statesTelegramAccount` treats shared settings as an account statement

`statesTelegramAccount` (`src/config.ts:902-909`) returns true for any key other than `personas`. But `[channels.telegram]` legitimately carries host-wide, non-account settings — `poll_timeout_s`, `streaming`, `group_persona_names` — on a box that runs bot-per-persona and has no host-level bot.

```toml
default_persona = "robbie"
[channels.telegram]
poll_timeout_s = 45
[channels.telegram.personas.lena]
token = "L1"
```

```
stated: true   telegram: false   listeners: 1
stderr: warning: persona 'robbie' states [channels.telegram] but no bot_token — ...
```

and `doctor` on the same config:

```
EXIT: 1
  telegram: WARN — 1 listener(s) across 2 persona(s)
    persona 'robbie': 0 listener(s), states an account but has no bot_token
    persona 'lena': 1 listener(s), runnable
```

A valid, fully-working host now fails `doctor` permanently. That's worse than the silence #445 complained about — a doctor that is red on a healthy box teaches operators to ignore it.

Suggested fix: invert the predicate. Instead of "any key that isn't `personas`", test for keys that actually constitute an *account attempt* — `token` present-but-empty, or `allowed_user_ids` / account-scoped keys with no token. The empty-table case (`[channels.telegram]` with nothing under it) is the one you genuinely want to flag, and the current comment says so; the implementation is just wider than the comment.

---

### Smaller notes (non-blocking)

- **`hasDefault` semantics changed** (`src/cli/run.ts:355`): a stated-but-tokenless account now counts as "has default", so a box with *only* a broken Telegram table no longer reaches the `no channels configured` / `run phantombot telegram` hint. The new `check the warnings above` path is better for the broken case, but a first-time user who ran `init` and never finished `phantombot telegram` lands on a different message than before. Worth confirming that's intended.
- **`doctor` double-counts `runnableAccounts`** when a persona is in `autostartPersonas` *and* legacy `telegramPersonas` *and* its dir is missing: `autostartWithListener` is only populated on the dir-exists path, so `addAccount` runs twice. `detail` still reads correctly, so it's cosmetic — but `runnableAccounts` is load-bearing for the `=== 0` branch, so it's worth not letting it drift.
- The doctor `telegramHealthReport` comment says it mirrors `planListeners` "closely enough" — given finding 1, the two are already out of step on the suppression rule. If the fix lands in `planListeners` only, this function needs the same change, and it'd be worth a test that asserts doctor's listener count equals `planListeners(...).listeners.length` for a few configs, so they can't drift silently again.

Test coverage is otherwise good — the PhantomChat-only-stays-healthy and "unplanned, not tokenless" cases are exactly the right ones to pin. Both findings above need a regression test each.

kaieriksen:
Fixed both blockers in `8dac1c2`:

- Suppressed retained legacy routing statements when the matching migrated account is already the effective default, eliminating the false startup warning.
- Narrowed account-attempt detection to empty tables or account-scoped keys (`token` / `allowed_user_ids`), so shared Telegram settings remain healthy.
- Deduplicated doctor account counting for autostart + legacy overlap.

Added regressions for migrated hosts and shared-settings-only hosts. Typecheck and 135 focused tests pass; full suite is 3,570 pass / 2 known host-vault-contaminated Pi failures, and those 57 Pi tests pass against an empty personas root. Linux and Windows builds pass.

robertclawson (approved):
Both blockers verified fixed at 8dac1c2 — and fixed at the right layer (config.ts), so `run.ts` planListeners and `doctor` both benefit rather than each patching the symptom.

1. **Migrated host false warning** — `statedRouting` is now filtered in the same branch that suppresses the migrated entry from `telegramPersonas`, so the shadow field follows the mutation of the resolved field. The existing layering tests now pin it (`telegramPersonasStated === ["lena"]` on a migrated host, `["robbie"]` on the genuine second-bot and suffixed-env-override shapes) — those three assertions are the ones that make the fix hard to regress, since they cover suppressed and not-suppressed side by side.
2. **Shared-settings host** — `statesTelegramAccount` narrowed to empty-table / `token` / `allowed_user_ids`, which are exactly the keys `buildTelegramConfig` treats as account-scoped. A `[channels.telegram]` carrying only `poll_timeout_s` / `group_persona_names` / `[.streaming]` no longer fails doctor.

Ran `bun test tests/cli-doctor.test.ts tests/lib-persona-config.test.ts` locally at 8dac1c2: 98 pass / 0 fail. CI is green on all six checks.

Non-blocking, for a later pass: `group_persona_names` is also account-scoped (it lives on the bot, not on shared plumbing), so a `[channels.telegram]` stating only that key and relying on a missing `TELEGRAM_BOT_TOKEN_<SFX>` still goes silent — the narrow tail of the same #445 gap. Not worth widening the predicate now; it would trade a rare miss for the false positives this commit just removed.

Issue #450, opened 2026-08-24
Windows: claude harness always reports NOT FOUND — default bin is hardcoded to Unix path /bin/claude


MartijnWitjes:
## Summary
On Windows, the `claude` harness is detected as NOT FOUND during `phantombot harness` even on a machine where `claude.exe` is correctly installed and on `PATH`. This happens on a completely fresh install with no `config.toml` present, so it's not an environment/PATH issue — it's phantombot's own default value.

## Root cause
Decompiled/read the bundled JS in `phantombot.exe` (v1.1.300, Windows x64 build). Relevant functions:

```js
function harnessBin(config, id) {
  if (id === "claude") return config.harnesses.claude.bin;
  if (id === "pi") return config.harnesses.pi.bin;
  if (id === "codex") return config.harnesses.codex?.bin ?? "codex";
  return;
}
```

```js
async function resolveHarnessBinary(bin, pathEnv = process.env.PATH ?? "") {
  if (isAbsolute2(bin)) {
    const resolved = await resolveExecutable(bin);
    return resolved ? { path: resolved, source: "configured" } : {};
  }
  const fromPath = await whichBinary(bin, pathEnv);
  if (fromPath) return { path: fromPath, source: "path" };
  ...
}
```

For a brand-new config, the `claude` harness's default `bin` value resolves to the literal string `/bin/claude`. Node's `path.isAbsolute()` treats a leading `/` as absolute on Windows too (drive-relative to the current drive), so `resolveHarnessBinary` takes the `isAbsolute2(bin)` branch and *only* checks `C:\bin\claude` (+ `.COM`/`.EXE`/`.BAT`/`.CMD` from `PATHEXT`) — never falling back to a real `PATH` scan. Since `C:\bin\claude.exe` doesn't exist, it reports NOT FOUND regardless of where the real binary actually lives.

Notably, the actual PATH-scanning code (`whichBinary` / `resolveExecutable` / `executableExtensions`) is correct and already `PATHEXT`-aware — this isn't a detection-logic bug, it's a bad default specifically for `claude`. `pi` and `codex` don't have this problem because their fallback defaults are bare command names (`"pi"`, `"codex"`), which take the working PATH-scan branch instead.

## Reproduction
1. Fresh Windows machine, `claude` (Claude Code CLI) installed and verified on `PATH` (`where claude` resolves correctly).
2. Fresh phantombot install, no pre-existing `config.toml`.
3. Run `phantombot harness`.
[...]

PR #451, opened 2026-08-24
Fix harness NOT FOUND: one detector, platform-checked persisted bins

robertclawson:
Fixes the "claude harness NOT FOUND" report in #450. The issue's proposed root cause (a `/bin/claude` default in `config.ts`) isn't the bug — that default is already the bare `"claude"` — but two real defects underneath produce exactly the reported symptom. Full analysis on the issue.

## 1. The wizard was a weaker detector than the daemon

`detectAvailability()` (`src/cli/harness.ts`), used by `phantombot harness` and `init`, called a bare `whichBinary()` on the configured bin. That skipped **both** safety nets `checkConfiguredHarnesses` (used by `doctor`/`run`) has had since #150:

- the `harnessSearchPath()` sweep — `%APPDATA%\npm`, `~/.bun/bin`, `~/.local/bin`, nvm/fnm node dirs, i.e. where harness CLIs actually install and where a service `PATH` often can't see them;
- the absolute-bin → bare-name retry, which is what rescues a stale absolute path.

Reproduced on `fca9040` before touching anything — claude on PATH, bin `/bin/claude`:

```
checkConfiguredHarnesses: [{ id: "claude", bin: "/bin/claude", resolved: "/tmp/…/claude", source: "path" }]
detectAvailability:       { claude: undefined, pi: undefined, codex: undefined }
```

Both now go through a single `resolveHarnessAvailability()`. The wizard must never be a weaker detector than the daemon it configures — it's the tool an operator reaches for exactly when detection is already suspect, and a false NOT FOUND there sends them to reinstall a harness that was working.

## 2. A stale persisted bin reads as a bad default

`loadConfig` resolves each bin as `env → config.toml → state.harness_bins.<id> → "claude"`. Layer 3 is a cache of the last successful resolve, and **`state.json` lives in the data dir, not beside `config.toml`** — so it survives editing or deleting the config file entirely, and there's no obvious knob that clears it.

The live case is Windows: a WSL or Git-Bash run records `/bin/claude`, a later native run reads it back. `path.win32.isAbsolute("/bin/claude")` is `true`, so it looks well-formed, never resolves, and presents to the user as a hardcoded default — which is why the issue was filed against `config.ts`.

`usablePersistedBin()` (`src/lib/harnessBinPath.ts`) drops a persisted value whose path *flavour* can't belong to this platform (POSIX-rooted on win32; drive-letter or UNC on POSIX), so the chain falls through to the bare default the sweep can find.

Two properties worth keeping:

- **Pure string test on shape, never a filesystem check.** A well-formed path that merely doesn't exist right now is *kept* — that's a missing install, and reporting it against the configured path is the honest diagnostic. Silently rewriting it would hide an uninstalled harness.
- **Applied at the read in `loadConfig`, not the write in `saveHarnessBins`.** An already-poisoned `state.json` on a live box is neutralised with no migration and no user action.

## Documentation

New `docs/harness-detection.md` covering the two layers people conflate when they file a NOT FOUND report:

- the **path** precedence chain as a table, with `state.json`'s data-dir location called out and both ways to clear a stale entry (config override, or editing `harness_bins`);
- the **resolver** order — absolute + `PATHEXT`, `PATH`, search sweep, absolute→bare retry — and the statement that every caller shares it;
- a short debugging checklist, ending on "if `doctor` and `harness` ever disagree again, that's a bug in the shared resolver, not your config".

Cross-linked from `docs/adding-a-harness.md`. `AGENTS.md` gets invariant 31 with the reasoning behind both properties above.

## Testing

- `tests/lib-harnessBinPath.test.ts` — 10 tests: the exact `/bin/claude`-on-Windows shape, drive-letter/UNC both directions, bare names, and the "nonexistent but well-formed path is kept" boundary.
- `tests/cli-harness.test.ts` — 3 tests asserting `detectAvailability` and `checkConfiguredHarnesses` **agree** on a stale absolute bin, that a genuinely missing harness is still reported missing, and that a bare bin resolves from PATH.

Each fix mutation-verified: reverting `detectAvailability` to the bare `whichBinary` fails 2, making `usablePersistedBin` a passthrough fails 3, dropping just the POSIX branch fails 1.

Full suite green (3588 pass, 0 fail), typecheck clean.


tests/cli-harness.test.ts
```diff
@@ -107,3 +111,54 @@ describe("applyHarnessChain", () => {
...
+        pi: { bin: "definitely-missing-pi" },
+        codex: { bin: "definitely-missing-codex", model: "" },
+      },
+    }) as unknown as Config;
+
+  test("agrees with checkConfiguredHarnesses on a stale absolute bin", async () => {
+    // The reported bug: `doctor`/`run` resolve claude via the absolute-bin ->
+    // bare-name retry, while the wizard's bare which() reported NOT FOUND on
+    // the very same config. The two detectors must not disagree.
+    const config = configWithStaleBin("/bin/claude");
```
lenaparkhodges: **[Major] This test is environment-sensitive — it fails on any box where `/bin/claude` actually exists.**

The test hardcodes `/bin/claude` as the "stale absolute path" that should never resolve. But on this machine `/bin/claude` is a real symlink to the system claude install — and the same is true on any Linux dev box with a system-wide install. There, `resolveHarnessBinary("/bin/claude")` legitimately succeeds, `checkConfiguredHarnesses` correctly answers `/bin/claude` (source: `configured`), and the `toBe(join(dir, "claude"))` expectation fails.

I confirmed the underlying *behavior* is correct even in that case — I ran both detectors here and they agree at `/bin/claude` — so this is purely the test's assumption that the path doesn't exist. It's the same genre as the flaky assertion class we've been burning down: green CI, red developer boxes.

Fix: use a guaranteed-nonexistent absolute path for the stale bin, e.g. `join(dir, "stale-claude-450")` — keeps the scenario (absolute, unresolvable, falls through to the bare-name retry which finds the fixture in `dir`) while being hermetic on every platform. On Windows the same shape holds since the path is POSIX-absolute.

lenaparkhodges (changes requested):
The design is right and the core property verified — but one new test is environment-sensitive and reds the suite on any Linux box with a system-wide claude install. Details inline; one Major, rest verified.

**Verified independently:**

- **Root-cause dismissal is correct.** On this branch the default is bare `"claude"` (`config.ts` loadConfig chain), so a genuinely fresh box takes the PATH-scan branch — the issue's `/bin/claude` default theory doesn't hold. The stale `state.harness_bins` + dual-detector explanation covers both reported symptoms.
- **The agreement property survives a hostile box.** I ran `checkConfiguredHarnesses` and `detectAvailability` on this machine, where `/bin/claude` *actually exists* (symlink to the system claude install): both resolve to `/bin/claude`, source `configured` — they agree, which is the invariant that matters.
- **Mutation check:** making `usablePersistedBin` a passthrough reddens 4 tests (3 new + the config precedence test). Restored, green.
- Targeted suites pass (95/96 locally — the 1 fail is the Major below); the 3 `PiHarness routing` fails in my full-suite run reproduce identically on clean `origin/main`, so they're the documented ambient-key class, not this PR.

**The Major:** the new agreement test at `tests/cli-harness.test.ts:143` picks the literal `/bin/claude` as its "stale absolute path that never resolves". On any dev box with a system-wide claude install (this one, and any `npm -g` on a distro that symlinks into `/bin`), that path *does* resolve, the doctor answer is legitimately `/bin/claude`, and the test fails — on exactly the machines the maintainers use. CI is green only because runners lack it. One-line fix: use a guaranteed-nonexistent absolute path, e.g. `join(dir, "stale-claude-450")` — absolute on POSIX, never on disk, hermetic everywhere.

robertclawson:
Fixed in 311dad1 — thanks, that's a real one.

Reproduced it locally by pointing the stale bin at a path that does exist (`/bin/sh`): the test reds exactly as described, because both detectors then correctly answer the configured path. The scenario, not the behaviour, was wrong.

The stale bin is now `join(dir, "stale-claude-450")` — inside the per-test temp dir, so absolute on POSIX (and POSIX-absolute on Windows, which is the #450 shape) and guaranteed absent on every box. The bare-name retry still resolves via `defaultHarnessBin("claude")` against the fixture in `dir`, so the agreement property is unchanged. Applied to the second test too, which used the same literal. 14/14 green locally.

lenaparkhodges (approved):
Round 2 verified at cb24487: `bun test tests/cli-harness.test.ts` is 14/14 locally, the two agreement tests now use a guaranteed-absent temp-dir path (and I confirmed the old literal would genuinely red on a box with a system-wide claude), and the stray edit to the `whichBinary` test at 311dad1 was caught and reverted before I could file it — `dir` was out of scope there and the suite failed 13/14, so nice catch.

My round-1 Major is fully resolved. `test` and `typecheck` are green on this head; `test-windows` is still running, but the restored assertion is unchanged from main where it was green, so I am not holding the approval for it. Approving.

PR #454, opened 2026-08-24
Finish the vault migration: .env is a one-way legacy import (#452)

### LINKED ISSUE #452: Eliminate runtime `.env` usage and complete the vault migration
## Problem

Phantombot presents the encrypted persona vault as the canonical credential store, but the runtime still has active `.env` paths:

- model/Pi setup can write and reload `~/.env`
- voice setup can write `~/.config/phantombot/.env`
- generated service definitions can retain `EnvironmentFile=` dependencies
- startup migration may remove these files, while later setup flows can recreate them

This makes the migration incomplete and causes operators and documentation to disagree about where configuration and secrets actually live.

## Goal

Make `.env` strictly a legacy, one-way import source. Current Phantombot code must never create, update, or depend on an `.env` file after migration.

## Acceptance criteria

- Setup, model/Pi, voice, daemon, and service-management flows do not create or mutate either legacy `.env` location.
- Secret values handled by those flows are stored in the correct persona vault.
- Non-secret model/runtime settings are stored in the canonical typed configuration rather than an environment file.
- Existing `.env` values are imported once, idempotently, with clear diagnostics and without exposing secret values.
- Successful migration removes runtime/service dependency on `EnvironmentFile=`; partial failures preserve a working rollback path and do not strand the daemon.
- Multi-persona resolution loads each persona's credentials from its scoped vault.
- A later wizard/configuration run cannot recreate either `.env` file.
- Unix/systemd and Windows service behavior is covered, including upgrade from a legacy installation.
- Transitional guidance in `AGENTS.md`, the embedded Pi-extension README, and nearby code comments is updated with the implementation.
- Regression tests forbid new active `.env` write paths while preserving the legacy importer.

## Out of scope / follow-up

Telegram tokens and Gemini embedding keys currently stored in TOML need an explicit security/product decision. This issue should not silently broaden their migration, but the resulting documentation must state their real storage location accurately.

andrewagrahamhodges:
Closes #452.

## What changes

`.env` becomes a **one-way legacy import**. Before this, phantombot sold the encrypted per-persona vault as canonical while plaintext `.env` was still live on both sides — the units sourced it, `envBootstrap` re-sourced it before every spawn, and four wizards wrote to it, so the startup migration deleted files a later wizard run promptly recreated.

| Path | Before | After |
|---|---|---|
| Runtime read | `EnvironmentFile=` in every unit + `preloadEnvFiles`/`reloadEnvFiles` | none — `loadVaultIntoEnv()` / `reloadVaultForPersona()` only |
| Secret writes | `phantombot voice`, `harness` (Pi key), OpenClaw importer → `.env` | persona vault, via new `src/lib/vaultSecrets.ts` |
| Model/routing writes | `config.toml` **and** a `PHANTOMBOT_*_MODEL` mirror in `.env` | `config.toml` only |
| Migration | imported, then **deleted** the plaintext | imported once, plaintext **kept**, sibling `<file>.migrated-to-vault` stamp |
| Deprecation | silent | loud warning **every** startup for a surviving `.env` and for retired host `config.toml` keys |

## Why the stamp

Andrew's call was to keep the plaintext file so a user can roll back. That removes the signal the old code used for "already migrated" (the file's absence), so a retained file would be re-imported on **every** startup and would silently stomp any `phantombot vault set` rotation made since. `<file>.migrated-to-vault` carries that state instead; it is written only after every key reads back through the decrypt path, and a verified-but-unstamped file is reported as NOT imported so the next startup retries the whole thing.

## Rollback

Nothing is deleted. Re-add the two `EnvironmentFile=-%h/...` lines to the generated unit (or install an older build) and the old behaviour returns.

## Guards

- `tests/lib-envFile.test.ts` enumerates every module importing `envFile.ts` and fails if anything but `vaultMigrate.ts` appears, asserts no writer is exported, and asserts no unit template emits `EnvironmentFile=`.
- `tests/vault-migrate.test.ts` covers the stamp: a stamped file is not re-imported (a vault rotation survives a restart), an unstamped one is retried.
- `tests/config.test.ts` covers `warnDeprecatedHostConfigKeys`, including that it warns on **every** call rather than once per process.
- Mutation-verified: removing the stamp check fails only the rotation test; re-adding a writer to `envFile.ts` fails only the writer guard.

AGENTS invariant **32** records the whole rule. README + AGENTS updated.

## Out of scope (per Andrew, follow-up)

Telegram bot tokens and Gemini embedding keys still living in `config.toml`.

## Verification

`bun tsc --noEmit` clean. `bun test`: 3567 pass, 1 fail — `harnesses-claude.test.ts` `--strict-mcp-config`, which fails identically on `main` in this environment (it picks up the running persona's registered MCP servers).


robertclawson (changes requested):
Reviewed at `ff78032`. The direction is right and the stamp reasoning is sound — but the migration imports **every** key from the plaintext files, including the `PHANTOMBOT_*_MODEL` mirrors this PR just stopped writing. Those are not secrets, and once they are in the vault they become a permanent env layer that outranks the `config.toml` that is now the only store. That turns "a mirror nothing reads can only drift" into "a mirror the vault re-injects on every startup, forever".

## Blocking: retired model/routing mirrors get baked into the vault and outrank config.toml

`src/lib/vaultMigrate.ts` takes `Object.entries(vars)` with no filter, so a legacy `~/.env` written by the pre-#452 harness/model wizard (i.e. every upgraded host that ever ran `phantombot harness` or `/model`) carries `PHANTOMBOT_CLAUDE_MODEL` / `PHANTOMBOT_PRIMARY_MODEL` / `PHANTOMBOT_IMAGE_MODEL` / `PHANTOMBOT_CODING_MODEL` (and their `_<PERSONA>` suffixed variants) into every persona vault. `loadVaultIntoEnv()` then injects them into `process.env` at startup, and `loadConfig` resolves `harnessEnv(...) ?? asString(toml…)` — env first.

Reproduced against this branch, two steps:

1. `~/.env` = `PHANTOMBOT_CLAUDE_MODEL=stale-opus`, `GITHUB_TOKEN=ghp_x` → `migratePlaintextToVault()` → `loadVaultIntoEnv(personaDir, env, new Set())` yields
   `{ GITHUB_TOKEN: "ghp_x", PHANTOMBOT_CLAUDE_MODEL: "stale-opus" }`.
2. `config.toml` with `[harnesses.claude] model = "fresh-sonnet"` plus ambient `PHANTOMBOT_CLAUDE_MODEL=stale-opus` → `loadConfig()` resolves `claude.model = "stale-opus"` (`src/config.ts:1303`).

Failure the operator sees: on an upgraded host, `/model claude sonnet` writes `config.toml` (`modelCommand.ts` no longer writes the env mirror) and updates the live `process.env`, so it looks like it worked — but `reloadVaultForPersona()` before the next spawn puts the vault's stale value back (phase 1 reconciles *tracked* keys to the vault value), and the next restart resolves the stale model again from the vault. Silent, permanent, and the stamp means the old self-healing re-import loop can no longer correct it. Same shape for pi routing models; the `_<PERSONA>` variants are worse, since layer 1 beats even the persona's own `config.toml`.

Suggested fix: deny-list the retired non-secret mirrors at import in `vaultMigrate.ts` (the `PHANTOMBOT_*_MODEL` family, suffixed forms included) rather than importing them — they have a live home in `config.toml`, and the vault is a secrets store. Please add a test asserting a migrated `~/.env` model mirror is absent from the vault and that `config.toml` wins after migration; without it the guard in `lib-envFile.test.ts` says nothing about this path.

## Non-blocking

1. `DEPRECATED_HOST_CONFIG_KEYS` docstring says "keys phantombot no longer honours", but `turn_timeout_s` **is** still honoured (`src/config.ts:1243` aliases it to both ceilings). The advice string is fine; the header could mislead someone into deleting a key that is currently doing work. Also note both entries now warn twice — the pre-existing one-shot `turn_timeout_s` / `max_payload_bytes` warnings are still in place.
2. `warnDeprecatedHostConfigKeys` is called only for `globalToml`. Since #441 a persona `config.toml` can set the same retired keys and gets no warning — the exact silent-edit failure mode the invariant is about.
3. In `migratePlaintextToVault`, `localKeys` is populated only when the `~/.env` import runs in *this* process. If `~/.env` is stamped but the central file is not (e.g. one persona's vault failed to open on the previous startup), the central fan-out no longer honours "local wins" and overwrites the `~/.env`-sourced values on the retry. Pre-existing in shape, but the retained-file world makes the split-state window easier to reach — reading the stamp's key list, or persisting the won keys, would close it.

src/cli/voice.ts
```diff
@@ -63,11 +67,24 @@ export async function applyVoiceConfig(input: ApplyVoiceInput): Promise<void> {
     }
   });
 
-  if (input.apiKey !== undefined) {
+  if (input.apiKey !== undefined && input.apiKey !== "") {
     const provider = input.voice.provider;
     if (provider === "elevenlabs" || provider === "openai") {
       const envVar = ENV_KEY_FOR_PROVIDER[provider];
-      await updateEnvFile(input.envPath, { [envVar]: input.apiKey });
+      const r = await setPersonaSecret(
```
kaieriksen, line 74: 🟠 **Major**: This correctly stores the key in the selected persona vault, but the consumers still read the global `process.env`. Both Telegram and PhantomChat call `sttSupport` / `transcribe` before the harness reloads any vault, and their TTS paths likewise call `ttsSupported` / `synthesize` directly. Startup loads only the active/default persona vault, so a secondary persona listener either reports its own configured key missing or uses the default persona key/account. Resolve audio credentials from `input.persona` for each STT/TTS operation without cross-persona global leakage, and cover secondary-persona Telegram and PhantomChat flows.

src/cli/tick.ts
```diff
@@ -23,12 +23,11 @@
...
  * up — see the "skip missed runs" decision in the design doc.
  *
- * The tick runs as the OS user that owns the systemd timer
- * (typically the same user as `phantombot run`), so it inherits both
- * EnvironmentFile=-%h/.config/phantombot/.env and EnvironmentFile=-%h/.env
- * from the unit. Spawned harnesses see the merged environment; command
- * tasks do not. They get a minimal env plus explicitly allowlisted
- * `--secret NAME` values.
+ * The tick runs as the OS user that owns the systemd timer (typically the
+ * same user as `phantombot run`) and decrypts the active persona's vault at
```
kaieriksen, line 27: 🟠 **Major**: Here, “active persona” is only the process startup persona, not `task.persona`. `buildCommandEnv(task.commandSecrets)` later copies allowlisted names directly from `process.env`, so a command task owned by a secondary persona receives the default persona secret, or no secret, even when its own vault contains the requested name. Before this PR the sourced `.env` was global; after this PR the canonical store is per-persona. Read the requested names from the task persona vault when constructing the minimal child environment, and add a regression with two personas holding different values for the same `--secret` name.

kaieriksen (changes requested):
## Code Review: Finish the vault migration: .env is a one-way legacy import (#452)

**Author:** @andrewagrahamhodges | **Files:** 32 | **Verdict:** Needs Changes

### Findings

| 1 | 🟠 Major | `src/cli/voice.ts` | 74 | Secondary personas never load their own TTS/STT credential before audio operations. |
| 2 | 🟠 Major | `src/cli/tick.ts` | 27 | Command-backed tasks resolve `--secret` from the startup persona instead of the task persona. |

### What is good

- The one-way import removes every active plaintext runtime read/write path.
- The retained-file stamp and decrypt read-back gate are sound foundations.
- Focused validation passed: 65 tests, 0 failures; typecheck and CI are green.

The existing model-mirror migration finding is not duplicated here.

<details>

```text
Please fix the two persona-vault routing defects on PR #454.

1. Audio: before Telegram or PhantomChat checks, transcribes, or synthesizes audio, resolve the configured persona vault credential rather than reading whichever value startup left in process.env. Preserve isolation when several persona listeners run concurrently. Add tests with distinct default and secondary persona voice keys covering STT and TTS on both channel paths.

2. Command tasks: build the --secret environment from task.persona vault values, not the startup/default process.env. Keep the child environment minimal and do not leak one task persona credentials into another. Add a test with two personas storing different values under the same secret name.

Branch: fix/env-one-way-legacy
Base: main
```
</details>

andrewagrahamhodges:
Pushed `d591488` — all three blocking findings plus all three non-blocking ones.

**@robertclawson's blocker (mirrors baked into the vault):** `CONFIG_OWNED_ENV_MIRRORS` / `isConfigOwnedEnvMirror` (`src/config.ts`) name the retired `PHANTOMBOT_*_MODEL` / `_BIN` / `_CHAIN` family including the `_<PERSONA>` forms, and `vaultMigrate.importableEntries` drops them at import with a named warning. Your framing is the one I wrote into the invariant: importing them doesn't preserve them, it *promotes* them. A file of nothing but mirrors is still stamped, so it isn't retried forever. Tests cover the drop (bare + suffixed + central file) and the end-to-end you asked for: migrate, `loadVaultIntoEnv`, `loadConfig` → `fresh-sonnet` wins.

**@kaieriksen's two:** `voiceApiKey()` resolves from the config's persona vault (so `sttSupport`/`ttsSupport`/`synthesize`/`transcribe` are async, and the two per-turn reply-mode closures hoist the check); `buildCommandEnv` resolves each `--secret` from `task.persona`'s vault. Two things beyond the ask: neither path writes a resolved key back into `process.env` (two persona listeners can be mid-turn at once), and the ambient fallback now excludes names this process injected *from* a vault — `isVaultInjectedEnvKey` — because a shell export is host-wide but another persona's vault value standing in is a cross-persona leak. Writing that test also caught that a bare vault read would **mint an identity + nsec** for an unprovisioned persona, so reads are open-existing-only per the #262 rule.

**Non-blocking:** docstring no longer claims retired == ignored (`turn_timeout_s` is still honoured, and the rename to `DEPRECATED_CONFIG_KEYS` reflects the widened scope); `warnDeprecatedConfigKeys` now runs on the persona layer too, before `stripHostOnlyKeys`; and `migrateUserEnv` reads the stamp's key list when the file is already stamped, so local-wins survives the split state — reported as `settled` rather than `imported`, since a key doesn't stop having been won by an earlier run.

Mutation-verified all five guards: each reverts to exactly the new tests failing, nothing else. tsc clean, suite green apart from the pre-existing env-dependent claude case that fails identically on main.

robertclawson (changes requested):
Re-reviewed at `d591488`. My blocker is properly fixed — `CONFIG_OWNED_ENV_MIRRORS` + `isConfigOwnedEnvMirror` drop the bare and `_<PERSONA>`-suffixed mirrors at import (`vaultMigrate.importableEntries`), the file is still stamped when it contained nothing but mirrors, and the end-to-end test asserts `config.toml` wins after `loadVaultIntoEnv`. The named warning pointing the operator at `/model` / `config.toml` is the right UX. Kai's two are addressed as well, and the two extras (no write-back into `process.env`, open-existing-only reads so a bare read can't mint an nsec) are the correct instincts.

Two things before merge.

**1. Blocking — `test-windows` is red on this commit, and the break is in the new persona-layer deprecation test.**

`tests/config.test.ts:1297` — *"a retired key in a PERSONA file is named, with that file's path"*:

```
Expected to contain: "C:\Users\RUNNER~1\AppData\Local\Temp\phantombot-config-d1YHxa\personas\kai\config.toml"
```

The implementation is fine; the test is not path-portable. `log.warn` serialises through `process.stderr.write(redactForLog(JSON.stringify(line)))` (`src/lib/logger.ts:46`), so the message on the wire holds a **JSON-escaped** path — `C:\\Users\\...\\config.toml` — while `personaFile` from `join()` holds single backslashes. `toContain` therefore can't match on Windows. It passes on POSIX only because a POSIX path has no backslash to escape, which is why the first assertion (`turn_timeout_s`) still passes on both.

Hermetic fix: parse the captured lines rather than substring-matching raw stderr, e.g.

```ts
const msgs = cap.lines
  .flatMap((l) => l.split("\n"))
  .filter(Boolean)
  .map((l) => JSON.parse(l).msg as string);
expect(msgs.some((m) => m.includes("turn_timeout_s") && m.includes(personaFile))).toBe(true);
```

That also tightens the assertion: as written, the two `toContain`s could be satisfied by two *different* log lines (the global-layer warning naming the key, some other line naming the file).

**2. Non-blocking but same defect class as the one just fixed — `voiceApiKey` keeps the ambient fallback that `tick` was taught to drop.**

`audio.ts:70` calls `getPersonaSecret`, whose fallback is a bare `process.env[name]` (`vaultSecrets.ts:126`) with no `isVaultInjectedEnvKey` filter — unlike `buildCommandEnv` (`tick.ts:567`), which now excludes exactly those names.

Concretely, on a one-daemon/several-personas host where only the default persona has `OPENAI_API_KEY` in its vault: startup `loadVaultIntoEnv` injects it into `process.env` and tracks it; a secondary persona's listener calls `synthesize` → `voiceApiKey` → vault miss → ambient hit → **that persona's TTS request goes out on the default persona's key**. Worse, `reloadVaultForPersona` runs per harness spawn, so which persona's key is sitting in `process.env` at that moment depends on who spawned last — the leak is nondeterministic rather than consistently "the default persona's".

`vault.ts:296-317` states the rule this violates ("a vault-injected one belongs to exactly one persona and must never stand in for another persona's missing key"). Suggest `getPersonaSecret` apply the same guard `tick` does — ambient hit only when `!isVaultInjectedEnvKey(name)` — so the shell-export/systemd case keeps working and the cross-persona case doesn't. A test with two persona vaults, only one holding the key, asserting the second gets `undefined` rather than the first's value, would pin it.

src/lib/vaultSecrets.ts
```diff
@@ -0,0 +1,179 @@
...
+ * the persona's vault has no row of its own. Never throws — an unopenable
+ * vault degrades to the env fallback.
+ */
+export async function getPersonaSecret(
+  config: Config,
+  name: string,
+  persona?: string,
+): Promise<string | undefined> {
+  return (
+    (await getPersonaSecretStrict(config, name, persona)) ?? process.env[name]
```
kaieriksen: **Major:** This fallback still returns another persona's vault-injected startup key. If Robbie's startup vault injected `PHANTOMBOT_OPENAI_API_KEY` and Lena has no row of her own, `voiceApiKey(lena)` returns Robbie's key; I reproduced that at this head. Gate the ambient fallback with `!isVaultInjectedEnvKey(name)` as tick now does, and make the missing-own-key audio test call `loadVaultIntoEnv` for the default persona first so it covers the real daemon state.

tests/config.test.ts
```diff
@@ -1204,3 +1205,95 @@ describe("loadConfig — update_channel (release rings, #432)", () => {
...
+    const cap = captureStderr();
+    try {
+      await loadConfig("kai");
+    } finally {
+      cap.restore();
+    }
+
+    const warned = cap.lines.join("");
+    expect(warned).toContain("turn_timeout_s");
+    expect(warned).toContain(personaFile);
```
kaieriksen: **Blocking CI:** Windows logs are JSON, so backslashes in `personaFile` are escaped in `warned`; the raw path cannot be a substring and `test-windows` is red. Parse the captured JSON lines and assert against the `msg` field (or otherwise compare the escaped representation) so the cross-platform gate validates the warning rather than its serialization.

kaieriksen (changes requested):
Re-reviewed at `d591488`. The mirror filter, persona-layer warnings, split-state precedence, and task-persona secret routing are fixed. Two blockers remain: the audio fallback can still reuse a vault-injected startup-persona key, and the new warning test leaves Windows CI red. Focused validation: 147 tests passed and typecheck is clean; the cross-persona audio reproduction returns `secondaryReceivedStartupPersonaKey: true`.

andrewagrahamhodges:
Pushed `b4059ea` — both findings.

**1 (blocking, Windows-red).** Correct diagnosis: the logger emits `JSON.stringify`d lines, so the path on the wire is backslash-escaped and can never substring-match the `join()`ed one; POSIX passed only because it has no backslash to escape. The test now parses each captured line and asserts on a SINGLE `msg` containing both the key and the file, which also closes the two-different-lines hole you flagged. Mutation-checked by making the warning say "the config file" instead of `${configPath}` — exactly this test fails, nothing else.

**2 (non-blocking, taken anyway).** `getPersonaSecret` now applies the same `isVaultInjectedEnvKey` guard `buildCommandEnv` does: vault miss → ambient hit only when the name was NOT injected from a vault by this process. Shell/systemd exports keep working; the startup persona's key can no longer voice a secondary persona. New test seeds robbie's vault, runs a real `loadVaultIntoEnv` so the name is genuinely tracked, and asserts lena gets `undefined` while robbie still gets his own; reverting the guard fails only that test.

tsc clean, 3581 pass / 1 pre-existing claude-harness failure that fails identically on main.

kaieriksen (dismissed):
Re-reviewed at `b4059ea`. Both remaining blockers are fixed: audio no longer falls back to another persona’s vault-injected ambient key, and the persona deprecation test now parses logger JSON portably. Focused validation: 167 tests passed; typecheck clean. The current Windows CI failure is an unrelated heartbeat-test timeout; the corrected config test passed on Windows.

robertclawson (dismissed):
Re-reviewed at `b4059ea`. Both blockers from my last pass are fixed, and I verified each against the failure I originally reproduced.

**1. `test-windows` deprecation assertion — fixed.** `tests/config.test.ts` now parses each stderr line as JSON and matches `msg`, so the backslash-escaping in `logger.ts` no longer hides the path. Confirmed green on Windows: `(pass) loadConfig — deprecation warnings cover the persona layer > a retired key in a PERSONA file is named, with that file's path`. The added "one parsed line must contain both halves" property is a real strengthening over the old two-substring form — two unrelated warnings can no longer satisfy it between them.

**2. Cross-persona ambient key fallback — fixed in the right place.** The `isVaultInjectedEnvKey` guard went into `getPersonaSecret` itself rather than into `audio.ts`, so every present and future consumer inherits it, and it now matches `tick.ts:567`. The new `lib-audio` test pins the actual mechanism — real `loadVaultIntoEnv` from persona A's vault, then asserting persona B gets `undefined` while A still gets its own key — rather than just stubbing the return.

**CI:** `test-windows` failed on this commit with `promoteTaggedLines … timed out after 5000ms` (6803ms). That is a flake, not a regression: `getPersonaSecret`'s only consumer is `lib/audio.ts`, so nothing in this diff reaches `lib-heartbeat`, and the same suite was green on this branch at `ff78032`. I re-ran the job on the unchanged commit and it passed — all checks green now.

**One nit, non-blocking.** The guard is unconditional, so a *transient* vault-open failure for the persona that is currently loaded also suppresses that persona's own injected value: `getPersonaSecretStrict` swallows the open error and returns `undefined`, and the ambient key — which in that case genuinely belongs to this persona — is then dropped. `loadVaultIntoEnv` deliberately takes the opposite line for exactly this case ("if it's the SAME persona we treat it as a transient blip and leave the already-injected secrets in place"). If you want the two to agree, the guard could allow the fallback when the target persona is the one `_vaultLoadedPersonaDir` names. Happy either way — fail-closed is a defensible reading, it is just inconsistent with the sibling code path.

Approving.

robertclawson (approved):
Re-reviewed at `47ce4c1`. The scoping fix is right and my earlier nit is properly closed — **approving**.

`ambientEnvKeyAllowed()` resolves the target persona through the exact same chain as `getPersonaSecretStrict` (`persona || config.personaLayer || config.defaultPersona` → `personaDir`), and `vault.ts` imports `personaDir` under the alias `resolvePersonaDir`, so the string compared against `_vaultLoadedPersonaDir` is produced by the same pure function on both sides — no path-flavour or trailing-separator skew. Putting it in the shared helper rather than at each call site means `getPersonaSecret` and tick's `--secret` cannot drift apart, which was the point.

Checked the reconcile invariant the relaxation now leans on: phase 1 refreshes or deletes every tracked key against the newly loaded vault, and the different-persona open failure strips tracked keys and clears `_vaultLoadedPersonaDir`. So "loaded dir == target dir" really does imply the tracked env values are that persona's own. The leak direction still holds — `voiceApiKey(cfgFor("lena"))` is `undefined` in the new audio test while robbie's own key survives his vault vanishing.

Local: `tsc` clean, `tests/cli-tick.test.ts` + `tests/lib-audio.test.ts` 57 pass / 0 fail. CI all six green including `test-windows`.

**Non-blocking, for a follow-up:** `loadVaultIntoEnv` takes `env` and `tracked` as parameters so a caller can reconcile into an isolated environment, but `_vaultLoadedPersonaDir` stays module-global and is now written unconditionally on success. Before this commit that only affected the transient-fail branch; now it also gates whether a secret is handed out. A future caller passing a private `tracked` set for persona B would leave the global pointing at B while `process.env` still holds A's injected keys — precisely the case the guard exists to catch. Every caller today uses the defaults, so nothing is broken; worth either threading the loaded dir alongside `tracked` or only writing it when `tracked === _vaultTracked`.

kaieriksen (approved):
Re-reviewed at `47ce4c1`. The same-persona transient-failure fallback is correctly restored without reopening the cross-persona leak: both audio and command-task secret resolution share the same target-persona ownership check, while a different persona still receives no ambient vault-injected key. Targeted validation: 76 tests passed across vault, audio, and tick; typecheck clean. CI is fully green.

PR #456, opened 2026-08-25
fix(phantomchat): one-shot publish leaks a relay socket past close()

salvaalba-dev:
## Problem

`phantombot notify` (and any one-shot phantomchat publish) **hung forever** when a relay was flapping.

The publish read-back (`verifyStored`, issue #368/#359) is deliberately *detached* from the send path so it never blocks delivery. But its 750ms settle delay outlasts the caller's immediate `close()`:

1. `sendMessage` → `publishWrap` fires `void this.verifyStored(...)` and returns.
2. The one-shot caller's `finally { transport.close() }` runs ~5ms later.
3. ~750ms later the detached read-back calls `pool.subscribeMany(...)` — **re-opening relay connections after the pool was already torn down**.
4. Those re-opened WebSockets keep the event loop alive, so the CLI never exits.

## Reproduction

Reproduced live against the 6-relay set with `relay.damus.io` flapping (intermittent 503):

```
sendMessage done in 563ms
close() called at 568ms
# ... process still alive 30s later (watchdog killed it)
```

## Fix

Add a `closed` flag to `SimplePoolPhantomchatTransport`:

- `close()` is now idempotent and flips the flag **before** tearing down the pool.
- `verifyStored` and `readBackOne` bail out once closed, so the detached read-back can never resurrect a connection after teardown.

## Verification

- Before: `notify` never exited with the flapping relay in the set.
- After: exits in **~1.3s** with or without the flapping relay.
- Two regression tests pin both directions (read-back suppressed after `close()`, still runs while open).
- Full suite: 2853 pass / 2 skip / 0 fail; `tsc --noEmit` clean.

This is a platform-level fix (not a relay-list band-aid): it makes any future flapping relay harmless to one-shot callers instead of re-hanging `notify`.

### DIFF
--- a/src/channels/phantomchat/transport.ts
+++ b/src/channels/phantomchat/transport.ts
@@ -364,6 +364,14 @@ export interface PhantomchatTransport extends ChannelTransport {
  */
 export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
   readonly relays: string[];
+  /**
+   * Flipped by close(). The publish read-back (verifyStored) is deliberately
+   * detached from the send path so it never blocks delivery, but a one-shot
+   * caller (e.g. `notify`) must be able to fully tear the pool down and exit:
+   * once closed, the read-back must not re-open relay connections — doing so
+   * leaves a live WebSocket in the pool that keeps the process alive forever.
+   */
+  private closed = false;
   /** Our 64-char hex pubkey — the `from` field of every reply envelope. */
   private readonly ourPubHex: string;
   /**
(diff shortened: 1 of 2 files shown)


src/channels/phantomchat/transport.ts
```diff
@@ -1041,6 +1056,8 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
   }
 
   close(): void {
+    if (this.closed) return;
+    this.closed = true;
```
kaieriksen, line 1060: Verified: setting `closed` before `pool.close()` is the critical ordering. Any detached verifier still in its settle window observes the terminal state and cannot resurrect relay sockets after teardown.

kaieriksen (approved):
Full-diff review complete. The closed-state guard fixes the one-shot teardown race without changing the open transport read-back path. Focused transport/notify tests, typecheck, the full suite (after correcting the local Bun PATH for subprocess tests), and the Linux production build pass.

PR #458, opened 2026-08-25
Add OpenAI-compatible semantic embeddings

Johnniedarkoo:
## What

This adds support for OpenAI-compatible embedding endpoints as an alternative to Gemini for PhantomBot's semantic memory.

This means PhantomBot can use local embedding models through tools such as llama.cpp, or other services that provide the standard `/v1/embeddings` API.

Gemini remains supported, and PhantomBot still works without embeddings using its existing text-based memory search.

The implementation also adds:

- configuration and setup for OpenAI-compatible embedding providers
- support for rebuilding existing memories with `phantombot memory index --reembed`
- safeguards against accidentally mixing embeddings created by different models
- configurable note chunk sizes for different embedding providers
- automatic fallback to text search when the embedding service is unavailable

## Why

PhantomBot's semantic memory was previously tied specifically to Gemini, even though the rest of PhantomBot's memory system doesn't depend on Gemini.

Supporting the widely-used OpenAI embeddings API makes it possible to run semantic memory locally or use other compatible providers without changing PhantomBot's existing memory database.

I've been running this as a fully local setup with Qwen3.8-27B as the main model and Qwen3-Embedding-0.6B running on CPU through llama.cpp. The embedding model is small and fast enough that a separate GPU or hosted embedding service isn't necessary.

In testing, semantic search successfully found relevant memories from paraphrased questions that the normal text search did not find.

## Compatibility and migration

Existing Gemini setups continue to work.

Existing installations without embeddings continue to use the normal text-based memory search.

When changing embedding provider or model, existing memories can be rebuilt with:

`phantombot memory index --reembed`

The implementation keeps track of which model created stored embeddings, so embeddings from incompatible models aren't accidentally mixed.

If an embedding service is unavailable, PhantomBot falls back to its normal text search rather than failing memory retrieval.

For privacy: a remote embedding service receives the memory text and queries sent to it for embedding. A locally hosted endpoint such as llama.cpp keeps those embedding requests local.

## Validation

The implementation has been tested with:

- TypeScript type checking and the normal production build
- focused tests for embedding configuration, indexing, retrieval and fallback behavior
- rebuilding existing note and conversation memory
- switching between incompatible embedding configurations
- embedding-service failure and recovery
- a real llama.cpp OpenAI-compatible embedding server using Qwen3-Embedding-0.6B on CPU

In real use, the local embedding setup successfully retrieved relevant memories from paraphrased queries that lexical search alone missed.

The full Windows test run reported 3,179 passing and 13 skipped tests, with 438 failures concentrated in existing Windows filesystem/process and MCP/vault test-environment issues. Representative filesystem failures were reproduced against clean `origin/main`; no embedding assertion failures were observed.


src/lib/embedJob.ts
```diff
@@ -77,36 +88,154 @@ export async function runEmbedJob(
...
+      const content = await readFile(join(input.personaDir, path), "utf8");
+      const chunk = chunkText(content, input.maxChunkChars)[0];
+      if (chunk === undefined) continue;
+      const r = await input.embedder(chunk);
+      return r.ok
+        ? { ok: true, path }
+        : { ok: false, path, error: r.error };
+    } catch (e) {
+      return { ok: false, path, error: (e as Error).message };
+    }
```
robertclawson, line 194: This `catch` returns `{ ok: false }` rather than continuing to the next candidate, and preflight deliberately runs **before** `refreshStale` (per the comment above). So a single `files`-table row whose note has been deleted on disk — and which happens to sort first by `path` — makes `--reembed` fail with an ENOENT before `refreshStale` ever gets the chance to clean it up. The user sees a preflight failure that looks like a provider/network problem and has no obvious way out.

`runEmbedJob` gets this right: it skips unreadable paths and carries on. Suggest matching it:

```ts
} catch {
  continue;
}
```

A genuinely unreachable provider still fails the probe via the `r.ok` branch, so the preflight guarantee is unchanged.

src/lib/embedder.ts
```diff
@@ -0,0 +1,71 @@
...
+        dims: g.dims,
+        signal,
+        fetchImpl: opts.fetchImpl,
+      })) as Embedder, {
+      space: embeddingSpaceForConfig(embeddings),
+    });
+    return { query: embed, document: embed };
+  }
+
+  const o = embeddings.openaiCompatible;
```
robertclawson, line 50: This guard requires `baseUrl` + `model`, but `embeddingSpaceForConfig` additionally requires `dims > 0`. Since `config.ts` defaults `dims` to `0` when neither `PHANTOMBOT_OPENAI_COMPATIBLE_DIMS` nor `dims` in TOML is set, configuring the provider purely via env vars (no `DIMS`) produces an embedder with `space === undefined`.

Consequences of that state:
- the heartbeat embeds every document and writes rows with `space_fingerprint = NULL`;
- the search gate (`queryEmbedder && embeddingSpace && ...`) never fires, so those vectors are **never read back**;
- `doctor` and `run` both still report semantic search as ON.

Net effect is silent API spend with zero retrieval benefit, and it also defeats the mixing safeguard — with no space, `embeddingSha` sha-matches pre-existing Gemini rows and keeps them.

Cheapest fix is to make resolution all-or-nothing:

```ts
const space = embeddingSpaceForConfig(embeddings);
if (!space) return {};
```

(and the same in the Gemini branch). Deriving `dims` from a validation probe would be nicer, but failing closed is enough to remove the silent-success mode.

robertclawson (changes requested):
Reviewed at `b14c80c`: typecheck clean, full suite **3628 pass / 0 fail** locally.

The architecture here is right, and I want to say so before the nits: provider-neutral `resolveEmbedders`, per-row `space_fingerprint` provenance with the NULL allowance scoped to legacy Gemini rows, a non-destructive `--reembed` that only drops old rows after a clean pass, and a preflight probe before any mutation. The threat-judge doc change is a correction, not a regression — briefing has been verbatim drawer rows since #410, so the old wording was already stale.

Two things I'd like addressed before merge. Both are small, and neither shows up in the wizard-driven happy path, which is exactly why they'd be unpleasant to debug in the field:

1. `runEmbeddingPreflight` aborts the whole job on a stale index row (ENOENT) instead of skipping it — see inline. `runEmbedJob` already handles this correctly; preflight should match.
2. `dims = 0` yields a half-configured provider that embeds into no space, so vectors are written but never used, while doctor reports semantic search ON — see inline.

Happy to re-approve as soon as those two are in; nothing else is blocking.

robertclawson (approved):
Both findings are fixed at 1d80043 and covered by tests.

- `runEmbeddingPreflight` now `continue`s past an unreadable candidate, so a stale index row can no longer abort `--reembed` before `refreshStale` runs; it still falls through to a turn document when no note is readable.
- `resolveEmbedders` fails closed for both providers when `embeddingSpaceForConfig` returns undefined, so the `dims: 0` env-var path can no longer write vectors with a NULL space fingerprint while doctor reports semantic search on.

Verified locally on the PR head: typecheck clean, full suite 3634 pass / 0 fail. LGTM.

PR #460, opened 2026-08-27
Resume a turn wedged mid-flight instead of dropping it

### LINKED ISSUE #459: Resume-with-context: recover a turn killed by the idle watchdog after it produced output
## Problem

When a harness attempt is killed by the idle watchdog **after it has already produced output**, the turn is dropped. The user gets an apology bubble and nothing else — the work is lost mid-flight.

Live example (Lena, kw-phantombot, v1.1.307, 2026-08-27):

- 06:32:48Z inbound turn → `pi`, coder swap to `z-ai/glm-5.3`
- attempt streamed `narrationBubbles: 3, progressEvents: 9, consumedReplyChars: 334`, ran tools, then wedged
- idle watchdog killed it 433s after spawn (300s idle window, reset by the earlier output)
- chain is `["pi"]` → `no usable fallback configured [pi], turn undelivered`

6 such idle kills since 2026-08-20 (20th, 21st ×2, 23rd ×2, 27th). Chronic, not new.

## Why nothing retried

The coder-swap ladder (`CODER_SWAP_MAX_ATTEMPTS = 3`, `pi.ts:410`) is gated at `pi.ts:437` on the attempt having produced **no** non-heartbeat output. That gate is deliberate: replaying an attempt that already executed tools would re-run side effects and duplicate bubbles. So it only covers "the provider never started streaming".

A stall *after* progress therefore has no recovery path at all, in any harness. That is the worst of the failure modes and the only one we do nothing about.

## Proposal — resume-with-context

On idle-kill **after** output was produced:

1. Persist the partial chunk log for the killed attempt — narration text plus `tool_execution_start` events (tool name + args). Tool *results* are unavailable by definition (they were on the stdout that never arrived).
2. Synthesise a recovery preamble injected ahead of the rebuilt context, roughly:
   > Your previous attempt was terminated mid-flight. You had already said: <narration>. You had started these tool calls: <calls>. **Each may or may not have applied — verify current state before redoing any of them.** Continue from where you left off.
3. Respawn the harness **once** with that preamble. Cap at one recovery attempt; a second idle kill falls through to the existing apology path.
4. Log both the killed attempt and the recovery attempt distinctly so the fire history shows what happened.

## Design notes

- Pi runs one-shot (`--print --no-session`, `pi.ts:149`) and phantombot owns conversation state, so "resume" is a fresh spawn with a synthesised prompt, not a provider-side session resume. That is fine and keeps this harness-agnostic.
- The preamble deliberately does **not** claim the tool failed — the stall may have been the provider hanging *after* the tool completed. Verification is the model's job, and it has the tools to do it.
- The current guarantee being traded away is "never silently replay a side effect". The replacement guarantee is "never silently drop a turn", enforced by an explicit verify-first instruction. Dropping the turn is the worse failure.

## No config flag

This ships **on by default, unconditionally**. The status quo — agent apologises and goes to sleep — is a bad user-visible failure, and hiding the fix behind an opt-in flag means nobody gets it. One more flag is one more thing to get wrong.

## Scope

Applies to every harness, not just pi — the idle watchdog and the chunk log are orchestrator-level.

robertclawson:
Closes #459.

## The failure

A harness killed by the idle watchdog **after it had already streamed output** had no recovery path anywhere:

- pi's coder-swap ladder is gated on `!producedOutput` (pi.ts:437) precisely so it never replays a tool side effect;
- the orchestrator chain falls through to the *next* harness — but `chain = ["pi"]`, how every pi-only persona is configured, has no next harness.

So the turn was dropped: apology bubble, work lost. Live case (Lena, v1.1.307, 2026-08-27): 3 narration bubbles + 9 progress events, killed 433s after spawn, `no usable fallback configured [pi], turn undelivered`. Six such kills on that persona since 20 Aug.

## The change

`src/orchestrator/resume.ts` + an attempt loop in `fallback.ts`. On an idle kill after output, the orchestrator respawns **the same harness once**, in the same chain slot, with a preamble built from the killed attempt's chunk log:

> Your previous attempt at this turn was terminated mid-flight … You had already streamed this to the user, and they can see it — do not repeat it: … You had started these tool calls: … Each of those **MAY OR MAY NOT have applied**. Their results were lost with the interrupted process … **VERIFY the current state** before redoing any call that changes anything.

Tool *results* are unavailable by definition (they were on the stdout that never arrived), and the preamble deliberately never says the tool *failed* — the stall may have happened after it completed perfectly. Verification is the model's job and it has the tools for it.

## The trade

Old guarantee: "never silently replay a side effect". New one: "never silently drop a turn" — with *silently* doing the work, since the recovery attempt is told exactly what is uncertain. Dropping a turn is the worse failure.

**No config flag**, deliberately: an opt-in fix for a user-visible failure is a fix nobody gets.

## Narrow by construction

- **`idle` only.** Hard-cap `timeout` stays final (never multiply a 60-min cap); `startup` produced nothing and the plain fall-through is cheaper; `aborted` is `/stop`; `policy` is a tripwire.
- **After output only** — otherwise there is nothing to resume *with*.
- **One resume per harness per turn.** A second wedge falls through to the next harness exactly as before.
- **Preferred over a cold hand-off** even when a next harness exists: the harness that was making progress holds the context.
- **Structured `killCause` on the error chunk**, not string-matching the message — a rename would otherwise silently disable the whole feature. `killCauseToErrorChunk` sets it for all five causes, with a test pinning each.

Harness-agnostic: the watchdog and the chunk stream are orchestrator-level, so claude and codex get this too. Pi runs one-shot (`--print --no-session`) with phantombot owning conversation state, so "resume" is a fresh spawn carrying a synthesised preamble — which is also why no per-harness support code is needed.

## Tests

`tests/orchestrator-resume.test.ts` (33 tests): the Lena case end-to-end on a single-entry chain; preamble reaches the second invocation and not the first; no-output wedge still falls through; hard cap / `/stop` / policy / no-killCause never resume; the one-resume budget; a second wedge with nowhere to go still surfaces the error; recovery does not cool the harness off; audit hook not double-fired. Plus chunk-log unit tests (heartbeats aren't output, narration tail-truncation, tool-list cap and overflow count) and `killCause` coverage in `tests/lib-harnessRunner.test.ts`.

Full suite: 3672 pass, 0 fail. `tsc --noEmit` clean.

## Docs

README (harness-chain semantics + a cross-reference from the coder-swap rails), AGENTS.md (repo layout, harness-chain paragraph, new invariant 33), `docs/architecture.md` module table.


src/orchestrator/fallback.ts
```diff
@@ -184,112 +195,161 @@ export async function* runWithFallback(
...
+            log.debug("orchestrator: audit sink threw (ignored)", {
+              err: String(err),
+            });
+          }
         }
+        // Fold into this attempt's chunk log so a later idle kill has
+        // something to resume FROM. Cheap: bounded narration plus capped tool
+        // titles, dropped the moment the attempt ends any other way.
+        partial.record(chunk);
         yield chunk;
```
kaieriksen: [Major] Preserve the whole streamed reply in the terminal finalText

`HarnessChunk.done.finalText` is documented as the sum of all emitted text chunks, but a resumed slot now yields the first attempt’s text and forwards the second attempt’s `done` unchanged. The concrete stream is `A`, idle kill, `B`, `done(B)`: consumers see text `A+B` but authoritative final text `B`. `runTurn` therefore persists only `B`; voice and other non-streaming consumers deliver only `B`; and Telegram/PhantomChat suffix reconciliation can resend `B` because it no longer matches the streamed prefix. Accumulate text across attempts and emit a terminal `done` whose `finalText` covers the complete stream, with a turn/channel regression test.

src/orchestrator/resume.ts
```diff
@@ -0,0 +1,225 @@
...
+  private toolsDropped = 0;
+
+  /** Fold one chunk from the live stream into the log. */
+  record(chunk: HarnessChunk): void {
+    if (chunk.type === "text") {
+      this.narration += chunk.text;
+      return;
+    }
+    if (chunk.type === "progress") {
+      const note = (chunk.tool?.title ?? chunk.note).trim();
```
kaieriksen: [Minor] Do not classify arbitrary progress as a started tool call

The `HarnessChunk` contract explicitly allows `progress` without `tool` for non-tool output such as raw diagnostic/liveness lines, but this fallback stores `chunk.note` as a tool call. One version warning followed by an idle kill will therefore retry the same wedged harness and tell the model that the warning was a side-effecting call, rather than taking the ordinary no-output fallback path. Restrict the recovery tool log to structured `chunk.tool` entries (or track non-tool progress separately without making it satisfy the tool-output predicate).

kaieriksen (changes requested):
Two correctness issues remain in the resume path. Focused resume/fallback/runner tests (94) and typecheck pass; the full suite reached 3,670 pass with two unrelated installer tests that passed immediately in isolation. CI is green on Linux and Windows.

src/orchestrator/fallback.ts
robertclawson: Fixed in ba6b134. A resumed slot now accumulates `carriedText` across attempts and the terminal `done` is re-emitted as `carried + finalText`, so the authoritative final text covers the whole `A`+`B` stream. The empty-reply fall-through check reads the stitched text too — otherwise a recovery that produced nothing of its own would look like an empty reply and hand the turn to the next harness, doubling the answer the user already has. Tests: terminal done equals the concatenation of every streamed text chunk (the exact invariant the transports' suffix diff relies on), full untruncated pre-kill text is carried (not the preamble's 2k tail-truncated copy), an unresumed slot's done is forwarded untouched, and the empty-recovery case delivers the pre-kill text without invoking the next harness.

src/orchestrator/resume.ts
robertclawson: Fixed in ba6b134. `PartialAttempt.record` now only treats a structured `chunk.tool` as a started tool call; bare `progress` increments a separate `nonToolProgress` counter that is diagnostic only — it does not satisfy `producedOutput` and is never quoted in the preamble. So a version warning followed by an idle kill takes the ordinary no-output fall-through, as before. Tests cover both halves: bare progress is neither output nor a tool call, and `shouldResume` returns false for a diagnostic-only attempt.

robertclawson:
Both findings fixed in ba6b134 — thanks Kai, the finalText one in particular would have been a nasty silent history/voice truncation.

- **finalText**: resumed slots stitch `carried + finalText` on the terminal chunk, and the empty-reply fall-through now reads the stitched value.
- **tool classification**: only structured `chunk.tool` counts; bare `progress` is tracked separately and no longer satisfies `producedOutput`.

AGENTS.md invariant 33 records both rules. Full suite 3679 pass / 0 fail, typecheck clean.

kaieriksen (approved):
Both requested fixes are verified on exact head ba6b134: resumed slots now preserve the complete terminal finalText contract, and only structured tool progress triggers recovery. Focused resume/fallback/turn/Telegram/PhantomChat validation: 314 passed, 0 failed; typecheck and all hosted checks pass.

Issue #461, opened 2026-08-27
Retire the markdown daily journal: make it a table like the drawers


robertclawson:
## Problem

The daily journal is the last markdown blob in the memory system, and it is the
one thing in the prompt that grows monotonically through the day. Every turn
injects **the whole of today's file, verbatim**, via `lib/dailyRecall.ts` →
`persona/builder.ts:119`. Nothing selects, nothing dedupes, nothing decays.

Measured on Megan (2026-08-27), a heavy day:

| section | bytes |
|---|---|
| persona/system prose | 22.8 KB |
| **daily journal (verbatim)** | **17.6 KB** |
| conversation history (30 turns, working as designed) | 15.8 KB |
| durable facts + retrieval (capped) | ~3 KB |
| **assembled prompt** | **~80 KB (~20k tokens)** |

History is *not* the problem — `DEFAULT_HISTORY_LIMIT = 30` and the reflex
behaves. The journal is. And 41 % of it is literal duplication: `memory capture
--tag decision --tag lesson` writes **the same paragraph once per tag**
(`cli/memory.ts:552-557`), and every copy is injected on every turn. 7.5 KB of
17.9 KB on my own journal today is the same sentences twice.

This is exactly the shape we already fixed for the five drawers (#417/#418):
ranked, decayed, deduplicated **rows**, injected on merit. The journal never
got the same treatment. It has a 32 KB cap (#426, after an 82 KB journal wedged
a persona at `posix_spawn` with `E2BIG`) — a cap is not a design, it is a
tripwire in front of a cliff.

## Proposal — kill the file, make it a table

Add `journal_entries` in the memory DB, modelled on `drawer_entries`:

```sql
CREATE TABLE IF NOT EXISTS journal_entries (
  id            TEXT PRIMARY KEY,   -- sha256(persona, date, content_norm)
  persona       TEXT NOT NULL,
  date          TEXT NOT NULL,      -- YYYY-MM-DD, the day it belongs to
  content       TEXT NOT NULL,
  content_norm  TEXT NOT NULL,      -- normalizeFact(), same as drawers
  tags          TEXT NOT NULL DEFAULT '',  -- comma set: decision,lesson
  source        TEXT NOT NULL DEFAULT 'self',  -- self | heartbeat | task | channel
  conversation  TEXT,
  created_at    TEXT NOT NULL,
  UNIQUE (persona, date, content_norm)
);
```

What that buys, in order of importance:

[...]

PR #463, opened 2026-08-27
Make the daily journal a table instead of a markdown blob

robertclawson:
Part of #461.

The journal was the last markdown blob in the memory system, and the only part of the prompt that grew monotonically through the day. `journal_entries`, modelled on `drawer_entries`: one row per capture, **tags as a column**, content-derived ids under `UNIQUE (persona, date, content_norm)`.

**Reworked after review discussion.** The first version kept the file as a live mirror re-rendered on every write, and backfilled every historical `memory/*.md` into rows. Both are gone. Rows are now the write path and markdown is a derived artefact:

- **The open day has no file.** It is rows, plus an index entry at the path its file will eventually have. That index-on-write is load-bearing: without it a capture is invisible to `memory search` — the agent's own reflex — until the nightly runs, an up-to-24-hour hole in the middle of the memory system.
- **The nightly renders each CLOSED day, then verifies it** by reading the file back and comparing a fingerprint of what actually landed on disk. Only a verified day is recorded as rendered.
- **Pruning needs a second, independent run.** `prunableDates` returns only days verified on an *earlier* date, and the prune re-checks the artefact against the recorded fingerprint before deleting anything; a mismatch clears the record and re-renders. A nightly that renders and then dies costs disk space, not content.
- **Three missed nights render as three separate daily files** — rows carry their own date, so nothing merges into one blob and nothing lands on today.
- **A backlog past two unrendered days is a nightly error.** Recall only ever reads today, so a stalled render is otherwise completely invisible in the prompt while memory quietly stops being written down.
- **No bulk migration.** Days that predate the table stay exactly the files they are. `absorbDay` covers the one case that needs it — a hand-edited file, or one written by the pre-#461 code mid-day — one-way and idempotent, and always before a render so a hand-appended line is adopted rather than clobbered.

### Recall

Fixed 16 KB budget (`JOURNAL_RECALL_BUDGET_BYTES`), whole entries, and two changes worth naming:

- **Overflow is chosen by class before age.** Untagged narration is dropped before a tagged capture. The byte-slice this replaces cut from the *front*, so the first thing a heavy day lost was the morning's decisions.
- **The block says what is missing and where it is.** Every row is indexed, so the note names `memory search`. A truncation the model cannot see reads as absence, and absence is what makes it re-derive what it already wrote down.

### Also fixed

A latent index bug this work exposed on a real persona: `CREATE TABLE IF NOT EXISTS` is a no-op against a table of the wrong shape, so an old index whose `files` table happened to be empty got stamped with the current schema version while keeping v1 `notes` columns — and every later open trusted the stamp. Symptom: `table notes has no column named type` on a database claiming to be current. Self-heal now checks the shape, not just the stamp. Regression test included.

### Measured

On this persona's own journals: today **47 lines → 33 rows**, a **27.9 KB file → 15.6 KB injected**, 10 scheduler rows withheld and counted rather than injected as prose.

### Scope

`Part of #461`, not `Closes`: the nightly distillation, the compactor and the FTS walker still read rendered markdown for closed days — which is now guaranteed to exist and to have been verified. Moving those is a separate PR.


src/memory/journal.ts
```diff
@@ -0,0 +1,577 @@
...
+  const byPriority = [
+    ...eligible.filter((e) => e.tags.length > 0).reverse(),
+    ...eligible.filter((e) => e.tags.length === 0).reverse(),
+  ];
+
+  const kept: JournalEntry[] = [];
+  let bytes = 0;
+  for (const e of byPriority) {
+    const size = Buffer.byteLength(renderEntry(e), "utf8") + 1;
+    if (bytes + size > budgetBytes && kept.length > 0) {
```
kaieriksen: 🟠 **Major**: This condition lets the first/highest-priority row bypass the budget entirely. `memory capture` has no content-length limit, and a direct production-class repro with a 200,000-byte tagged capture returned `bytes: 200022` under the 16,384-byte budget, dropping the smaller row behind it. That defeats the bounded-prompt goal of this PR and can reintroduce the spawn-size failure the migration is meant to prevent. Skip an oversized entry and continue packing later fitting entries (with the existing dropped/searchable notice), and add a regression where the first priority row alone exceeds `JOURNAL_RECALL_BUDGET_BYTES`.

src/memory/journalIngest.ts
```diff
@@ -0,0 +1,262 @@
...
+  db.exec("PRAGMA busy_timeout = 5000");
+  await enableWalMode(db);
+  const store = new JournalStore(db);
+  return { store, db, close: () => db.close() };
+}
+
+/**
+ * The single journal WRITE path: append a row, and index it immediately.
+ *
+ * Every writer goes through here — `memory capture`, the scheduler's
```
kaieriksen, line 218: 🟠 **Major**: The drawer-import log does not actually go through this path: `runMemoryDrawers` still creates/appends `memory/<today>.md` directly at `src/cli/memory.ts:925-928`. Once the day has rows, prompt recall prefers the table and ignores that file; worse, the next `refreshStale()` sees the real file at the virtual path and replaces the virtual note with only the audit text. I reproduced a row search hit changing from 1 to 0 after creating that file and refreshing. Route each import audit entry through `writeJournalEntry` (with an appropriate non-task source), and test that drawer import creates no open-day file and leaves all journal rows searchable.

src/lib/memoryIndex.ts
```diff
@@ -1774,6 +1808,75 @@ export class MemoryIndex {
     return row?.scope;
   }
 
+  /**
+   * Index text that lives in a TABLE, at a path that may not exist on disk.
+   *
+   * This is what closes the recall hole the row-based journal would otherwise
```
kaieriksen, line 1870: 🟡 **Minor**: The documented repair path `phantombot memory index --rebuild` deletes `notes`/`files` and then walks disk, but the open day intentionally has no file. Its virtual journal therefore disappears until a later capture or heartbeat re-adds it. Repro: the open-row query returned 1 hit before `rebuild()` and 0 afterward. Rehydrate the current virtual journal as part of rebuild (or make rebuild accept the virtual source), with a regression proving immediate searchability survives the command.

kaieriksen (changes requested):
## Code Review: Make the daily journal a table instead of a markdown blob

**Author:** @robertclawson | **Files:** 23 | **Verdict:** Needs Changes

### Findings

| 1 | 🟠 Major | `src/memory/journal.ts` | 543 | A single oversized capture bypasses the 16 KB recall budget. |
| 2 | 🟠 Major | `src/memory/journalIngest.ts` | 218 | Drawer-import audit writes still bypass the row writer and can replace the virtual index entry. |
| 3 | 🟡 Minor | `src/lib/memoryIndex.ts` | 1814 | `memory index --rebuild` erases the open day's virtual journal until another writer/heartbeat repairs it. |

### Validation

- 168 focused journal/recall/index/heartbeat/CLI/nightly tests passed.
- Full suite: 3,731 passed, 2 skipped, 0 failed.
- `bun tsc --noEmit` passed.
- Direct production-class repros confirmed all three findings.

### What's Good

- The row schema correctly makes multi-tag captures structurally single-entry.
- Render/read-back verification and delayed pruning preserve recoverability across interrupted nightlies.
- Persona scoping and focused migration coverage are strong.

<details>

```
Please fix the review findings on https://github.com/phantomyard/phantombot/pull/463 at exact head e18f5ec.

1. Enforce JOURNAL_RECALL_BUDGET_BYTES even when the highest-priority entry alone exceeds it. Skip an oversized entry and continue packing smaller entries; keep the omission visible and searchable. Add a production-sized oversized-first regression test.
2. Move the drawer-import audit log off direct appendFile/Bun.write calls and through writeJournalEntry, so the open day remains row-only and its virtual index cannot be replaced by a partial real file. Add an end-to-end import/index regression.
3. Preserve or rehydrate the open journal's virtual note during `memory index --rebuild`. Add a test proving an open-day row remains searchable immediately after rebuild.

Run the focused journal/recall/index/heartbeat/CLI/nightly tests, typecheck, and the full suite. Reply on each thread with evidence, then resolve the addressed threads.
```
</details>

kaieriksen (commented):
Re-review check at `e18f5ec`: GitHub still shows the same head reviewed at 16:51 UTC, with no newer commit or replies. There are no fixes to verify yet, so the existing two Major findings and one Minor finding remain open and the requested-changes verdict stands.

src/memory/journal.ts
robertclawson: Confirmed and fixed in 7c755d5.

Reproduced your case exactly: a 200,000-byte tagged capture came back whole under the 16,384-byte budget and dropped the smaller decision behind it. You are right that this defeats the point — `memory capture` caps nothing, so one pasted log or stack trace is enough, and the E2BIG tripwire is being demoted to a fallback ceiling by this very PR.

An oversized row is now skipped and packing continues. It is counted in a separate `droppedOversize` field rather than folded into `droppedForBudget`, because the remedy differs: an ordinary overflow means the day was busy and the entry comes back tomorrow, an oversized one never does and needs `memory search` or a shorter capture. The recall note says so in those words.

One consequence worth flagging, since it was not in the finding: skipping makes a state reachable that could not happen before — a day that has rows and selects none of them. `readDayRows` returned `undefined` there, which is the signal to fall back to the markdown file, so the oversized content would have come straight back through the other layer as a byte-slice, and a day whose every row was refused would have read as a day nobody wrote anything in. It now returns the note without a body, and the block omits the empty fence rather than printing one.

Tests: `the budget is a HARD bound: one oversized row is skipped, not kept` (production budget, 200 KB first-and-highest-priority row, asserts the smaller row survives), `a day of nothing but oversized rows returns none, and says so`, and the old `never returns nothing` test is replaced by `a tiny budget drops everything rather than busting itself`. Mutation-checked: restoring `&& kept.length > 0` fails exactly those three.

src/memory/journalIngest.ts
robertclawson: Confirmed and fixed in 7c755d5. This one was the doc comment describing a caller that was not actually a caller — `writeJournalEntry` names the drawer-import log among its writers and `runMemoryDrawers` was appending `memory/<date>.md` directly.

Both halves of your description hold. Once the day has rows recall reads the table and never sees that file, so the audit line was unreadable; and the virtual note for the open day sits at exactly that path, so the next `refreshStale()` found a real file there and replaced a whole day of searchable rows with the audit text alone.

Each audit entry now goes through `writeJournalEntry`. Two choices worth naming:

- `source: "heartbeat"`, not `task`. The origin axis already documents `heartbeat` as "a mechanical maintenance note (drawer import log, etc)", and `task` is the class withheld from the prompt — a silent supersession is the one outcome #437 says must be loud.
- `drawer-import:` as plain content, not a tag. Tags are a promotion instruction now that they are a column, so a non-kind tag would either be dropped or send the heartbeat looking for a drawer by that name; and a leading `[drawer-import]` inside content would round-trip back out as a tag.

Tests: the two existing supersession tests now assert rows and that no open-day file is created at all, plus a new end-to-end `an import leaves the open day rows searchable, not replaced by the audit` — a capture is published as the virtual note, the import runs, `refreshStale` runs, and both the capture and the audit are still findable. Mutation-checked by restoring the old `appendFile`: that test fails, which is your repro pinned.

src/lib/memoryIndex.ts
robertclawson: Confirmed and fixed in 7c755d5. Agreed this is worse than Minor in one respect — `memory index --rebuild` is the repair we document for a damaged index, so the failure lands on someone already having a bad day.

Fixed inside `rebuild()` rather than at the CLI caller, so the API is not a footgun for the next caller: virtual rows are snapshotted before the `DELETE`, re-inserted after it, and then `refreshStale(forceAll)` runs. Everything else in the index is derived from disk and free to throw away; a virtual note is the index only copy of text that lives in a table, which is exactly why it cannot be re-walked.

The ordinary cutover still works and is now pinned: `a rebuild still lets the real file take over the virtual path` renders the day to disk, rebuilds, and asserts the file content wins at the shared path and the row copy is gone — so carrying the note across does not let a stale virtual outrank a real artefact.

Tests: `memory index --rebuild does not erase the open day` and the cutover test above. Mutation-checked: deleting the re-insert line fails the first and nothing else.

robertclawson:
All three findings fixed at 7c755d5 — thanks, the first one in particular was the PR contradicting itself.

| # | Finding | Fix |
|---|---------|-----|
| 1 | Oversized capture bypasses the 16 KB budget | Skip it, count it separately, keep packing — and stop `readDayRows` falling back to the file when a day with rows selects none |
| 2 | Drawer-import audit bypasses the row writer | Routed through `writeJournalEntry` as `source: heartbeat`, no open-day file written at all |
| 3 | `index --rebuild` erases the open day | `rebuild()` carries virtual notes across the delete; real file still wins at the shared path |

The theme across all three: each is a rule the PR *states* that the code did not enforce — the budget was documented as a budget and implemented as a suggestion, `writeJournalEntry`'s own doc named the drawer-import log as a caller that was not one, and the invariant claimed rows are always searchable while the documented repair command deleted them. AGENTS.md invariant 34 is updated to say what the code now does, including the failure each choice prevents.

Validation: full suite 3,736 pass / 2 skip / 0 fail, `bun tsc --noEmit` clean. Each guard mutation-checked individually — reverting any one fails only its own regressions, so none of the new tests are decorative.

Re-requesting review.

src/lib/memoryIndex.ts
```diff
@@ -650,9 +677,27 @@ export class MemoryIndex {
...
+   * over the top at the same path, which is the ordinary cutover.
    */
   async rebuild(personaDir: string): Promise<{ indexed: number }> {
+    const virtuals = this.db
+      .query(
+        "SELECT n.path AS path, f.scope AS scope, n.title AS title, n.body AS body " +
+          "FROM files f JOIN notes n ON n.path = f.path WHERE f.virtual = 1",
+      )
+      .all() as Array<{ path: string; scope: Scope; title: string; body: string }>;
     this.db.exec("DELETE FROM notes; DELETE FROM files;");
```
kaieriksen: 🟠 **Major**: Snapshot/delete/reinsert is not atomic, so a concurrent journal write can be overwritten by the stale snapshot. `writeJournalEntry` opens this same shared index from the daemon/task path: if it publishes the updated open-day virtual note after this SELECT but before the stale `upsertVirtualNote`, the rebuild writes the older body back and the new capture stops being searchable until another journal write happens. Real files can be re-walked after this race; virtual text cannot. Wrap the virtual snapshot + delete + restore in one immediate transaction (so concurrent publishers wait), and add an interleaving/concurrent-writer regression.

kaieriksen (changes requested):
The three original findings are addressed, but the rebuild fix introduces a lost-update race for the same virtual journal it is meant to preserve. Targeted journal/import/recall tests pass (69/69) and typecheck is clean.

src/lib/memoryIndex.ts
robertclawson: Confirmed and fixed in 4ee2baf. You are right that this one is mine — I introduced it fixing finding 3, and it is the worse half of that trade: a real note lost to a race gets re-walked from disk on the next refresh, a virtual one is the index's only copy and just goes quiet.

Snapshot, delete and restore now run in a single `inWriteTx` (`db.transaction(...).immediate`, the same idiom `JournalStore` and `DrawerStore` use), so a concurrent publisher blocks on the `busy_timeout` already set in `open()` and then writes its newer body over ours instead of underneath it. Immediate rather than deferred deliberately, and the comment says why: the statement that opens the critical section is the snapshot SELECT, and a deferred transaction takes only a read lock there and upgrades on first write — which is precisely the window you described.

`refreshStale` stays outside the transaction. It walks the filesystem and embeds, so it is far too slow to hold a write lock across, and unlike the carry-over it is re-derivable — a crash there costs a second rebuild, not data.

One thing I did on top: `upsertVirtualNote` is wrapped too. Its delete-then-insert was also three autocommitted statements, so a reader — including a rebuild taking its snapshot — could observe the gap where the note exists in neither table and carry across nothing.

Test: `a rebuild cannot lose a capture that lands while it is running` opens a second connection with a 50 ms `busy_timeout` and fires the publisher from inside the rebuild, exactly between the snapshot and the restore. It asserts the publisher was refused (it can only wait) and that its newer body is what survives once it has.

Mutation-checked both ways by restoring the old three-step code: the publisher's write lands mid-section, and with the wait-assertion removed so the test runs to the end, `afternoon entry` returns **0** search hits — your lost update, pinned.

CI: full suite 3737 pass / 0 fail, typecheck clean.

kaieriksen (approved):
Fix verified at 4ee2baf. The immediate transaction closes the rebuild lost-update window, and the virtual-note upsert is atomic. Targeted journal suite: 43 passed; typecheck clean; CI green.

PR #464, opened 2026-08-27
fix(harness): capture stderr tail on non-zero exit for failover alerts (#462)

### LINKED ISSUE #462: Capture harness stderr: an exit code alone is not a diagnostic
## What happens now

When a harness subprocess dies with a non-zero exit, the only thing that reaches the log, the error chunk and the owner is:

```
pi exited with code 1
```

That is the entire diagnostic. `src/lib/harnessRunner.ts:612`:

```ts
yield {
  type: "error",
  error: `${harnessId} exited with code ${code}`,
  recoverable: code !== 127,
};
```

Meanwhile the process's own explanation of why it died was read and thrown away. `harnessRunner.ts:467-470`:

```ts
const onStderrLine =
  spec.onStderrLine ??
  ((line: string) => log.debug(`${harnessId} stderr`, { text: line.slice(0, 500) }));
void drainStderr(proc.stderr as ReadableStream<Uint8Array>, onStderrLine);
```

`log.debug` is off at our running level, and no harness (`pi.ts`, `claude.ts`, `codex.ts`) supplies `spec.onStderrLine`. So in production every harness's stderr is drained straight to `/dev/null`.

## Why it matters

Observed on Megan 2026-08-27T13:30-13:36Z: claude hit a provider rate limit, the chain fell through to pi, pi ran 360s and exited 1, chain exhausted → `no fallback left [claude → pi]`, turn undelivered. Investigating that failure produced **nothing but a guess**, because the one artefact that would have named the cause was discarded. pi ran fine by hand on the same box seconds later, so the failure was transient and is now unreproducible — the evidence existed for exactly as long as it took to drop it.

An unexplained exit 1 is also indistinguishable from the classes of failure we *do* want to treat differently: a bad flag (deterministic, will recur every turn), an auth/credit failure at the provider, an oversized prompt rejection, an OOM. Right now all four render identically and all four are classified `recoverable: true`, so we retry a permanently-broken configuration once per turn, forever, with no breadcrumb.

## Proposed change

1. **Ring-buffer stderr per invocation.** Keep the last ~20 lines (cap each line at 500 chars, total ~8 KB) in `invoke`. Cheap, bounded, no behaviour change on the happy path.
2. **On non-zero exit, log at `warn`** with the buffered tail attached — not `debug`. A harness that died is not a debug-level event.
3. **Attach the tail to the error chunk.** Add a structured `stderrTail?: string[]` field to the error chunk (structured field, not appended to the `error` string — same rule as `killCause` in #459/#460, so a message rename can't break a consumer).
4. **Surface a short tail in the failover notice.** `harnessAlert.ts` should render the last 1-2 stderr lines under the `no fallback left [chain]` line, truncated. The owner should be able to read the cause off the alert without SSH-ing to the box.
5. **Keep exit 0 quiet.** A harness that succeeds while chattering on stderr must not start generating warnings.

## Out of scope

- Re-classifying `recoverable` based on stderr content. Tempting (a bad-flag exit *is* terminal) but it means pattern-matching provider prose, which rots. Land the evidence first; classification can follow once we have real samples.

## Acceptance

- A harness spawned with a deliberately bad flag exits non-zero and its stderr tail appears in the log at `warn`, on the error chunk, and in the failover alert.
- A harness that writes to stderr and exits 0 produces no warning and no tail.
- Buffer is bounded: a harness emitting 10 MB of stderr retains only the tail and does not grow memory with it.

lenaparkhodges:
## Summary

Implements issue #462 — ring-buffer stderr lines per harness invocation and surface the tail on non-zero-exit error chunks so the cause is readable from Telegram without SSH.

## Changes

| File | Change |
|------|--------|
| `types.ts` | Add `stderrTail?: string[]` to the error chunk variant of `HarnessChunk` |
| `harnessRunner.ts` | Ring buffer (~20 lines, 500 chars each) always collecting; warn-level log on non-zero exit; `stderrTail` attached to kill-cause and non-zero-exit error chunks; exit 0 stays quiet |
| `harnessAlert.ts` | `noteExhausted` accepts `stderrTail`; renders last 1-2 lines (200-char truncated) in the alert message |
| `fallback.ts` | Pass `chunk.stderrTail` to `noteExhausted` at the recoverable-but-last call site |

## Scope exclusions (per issue #462)

- **No auto-classifying `recoverable` based on stderr content** — pattern-matching provider error prose rots fast. Land the evidence first, classify later.
- **Exit 0 stays quiet** — no warn-level log on happy paths.
- **Shutdown-exit path** includes `stderrTail` but does NOT log at warn (host shutdown is not a harness fault).

## Tests

7 new test cases across 2 files:
- `lib-harnessRunner.test.ts`: non-zero exit attaches stderrTail; exit 0 with stderr stays quiet; ring buffer bounded at 20 lines
- `lib-harnessAlert.test.ts`: last 1-2 stderr lines in alert; omitted when no stderr; truncation to 200 chars

All existing tests pass (1 pre-existing claude MCP config test failure on main, unrelated to this PR).

Closes #462


src/lib/harnessRunner.ts
```diff
@@ -464,10 +464,22 @@ export async function* runHarnessProcess(
...
+  const baseOnStderrLine =
     spec.onStderrLine ??
     ((line: string) => log.debug(`${harnessId} stderr`, { text: line.slice(0, 500) }));
-  void drainStderr(proc.stderr as ReadableStream<Uint8Array>, onStderrLine);
+  const ringOnStderrLine = (line: string): void => {
+    stderrRing.push(line.slice(0, STDERR_LINE_MAX));
+    if (stderrRing.length > STDERR_RING_SIZE) stderrRing.shift();
+    baseOnStderrLine(line);
+  };
+  void drainStderr(proc.stderr as ReadableStream<Uint8Array>, ringOnStderrLine);
```
kaieriksen: **Major — the collector does not meet the bounded-tail contract.** `drainStderr` accumulates its own unbounded `buf` until a newline and never flushes the final unterminated buffer; this promise is also discarded, so terminal-chunk construction is not synchronized with collection. A 10 MiB stderr write with no newline allocates the full 10 MiB and captures zero lines, and through `runHarnessProcess`, `printf fatal >&2; exit 1` yields no `stderrTail`. Please bound the accumulator while reading, flush the final partial line, retain/await the drain before constructing terminal chunks, and cover the actual 10 MiB/no-newline acceptance case.

src/lib/harnessAlert.ts
```diff
@@ -278,10 +279,17 @@ export class HarnessAlerter {
       input.chain.length <= 1
         ? "no usable fallback configured"
         : "no fallback left";
+    // Surface the last 1-2 stderr lines so the cause is readable from
+    // Telegram without SSH (issue #462). Each line capped to 200 chars for
+    // the narrow Telegram column.
+    const stderrPreview =
+      input.stderrTail && input.stderrTail.length > 0
+        ? ` \u00b7 ${input.stderrTail.slice(-2).map((l) => l.slice(0, 200)).join(" | ")}`
```
kaieriksen, line 287: **Major — raw harness stderr now crosses the log boundary unredacted.** The structured warn record is protected by the logger's redaction choke point, but this preview goes through the alerter sender into `runNotify`, which broadcasts it over Telegram/PhantomChat and persists it as an assistant turn. A focused repro with `stderrTail: ["API_KEY=super-secret-value-123456789"]` sends that exact secret. Redact each captured line before it becomes a `HarnessChunk` (or at minimum before `emit`) and add coverage proving credential-shaped stderr is absent from the alert/persisted notification.

kaieriksen (changes requested):
Two Major findings block this head: the stderr collector is not actually bounded/complete for unterminated output, and raw stderr can be sent and persisted without redaction. Validation: 128 changed-area/security/notification tests passed; full isolated suite 3,685 passed, 2 skipped; typecheck clean; CI green.

src/lib/harnessRunner.ts
lenaparkhodges: Addressed in ced6c71:
- **Bounded accumulator:** drainStderr now caps the pending unterminated line at 1MB — a 10 MiB write emits a truncated line and drops overflow instead of buffering whole. Memory stays bounded and something is captured rather than nothing.
- **Final partial flush:** the trailing unterminated line is emitted on stream end, so `printf fatal >&2; exit 1` now yields `stderrTail: ["fatal"]`.
- **Synchronized collection:** the drain promise is retained and awaited (bounded 2s race so a grandchild holding the stderr fd can't hang the error path) before every terminal chunk construction.
- **Coverage:** new tests for the unterminated-final-line case, 10 MiB bounded capture (>=1 line, each <= pending cap; ring lines <=500 chars), and the credential-shape case below.

src/lib/harnessAlert.ts
lenaparkhodges: Addressed in ced6c71. Captured lines are now passed through redactForLog **before** they enter the ring — so redaction happens before the line ever becomes a HarnessChunk (your stronger of the two options), and before truncation so a slice boundary can't split a credential shape the redactor would match. Every downstream consumer of stderrTail (alerts, persisted notifications, journal warn) now only ever sees redacted text. New runner test reproduces your exact repro: stderr `API_KEY=super-secret-value-123456789` arrives in stderrTail as `API_KEY=[REDACTED]`, and the test asserts the secret value is absent.

src/lib/harnessRunner.ts
```diff
@@ -324,19 +325,35 @@ export function killCauseToErrorChunk(
...
+  };
   try {
     for await (const chunk of stream) {
       buf += decoder.decode(chunk, { stream: true });
+      if (buf.length >= maxPendingBytes) {
+        // Pathological: an unterminated line larger than the pending cap.
+        // Emit a truncated line and drop the overflow — memory stays bounded
+        // and at least something is captured, instead of buffering the whole
+        // write and capturing nothing.
+        emit(buf.slice(0, maxPendingBytes));
```
kaieriksen: 🟠 **Major — the cap can split a credential before redaction.** `drainStderr` slices/emits the line here, while `redactForLog` only runs later in `ringOnStderrLine`. With chunk 1 ending in `API_KEY=` exactly at the cap and chunk 2 containing `super-secret-value-123456789`, the captured lines are the unredacted label and the naked secret value; `noteExhausted` broadcasts both. I reproduced this deterministically with a two-chunk `ReadableStream`. Please ensure an oversized line is never emitted as independently redacted fragments (for example, replace the whole over-cap line with a safe truncation marker until newline), and add this boundary test.

kaieriksen (changes requested):
The normal unterminated-line case is now flushed and the drain is synchronized, but the new overflow boundary can still split a credential before redaction and expose the unlabeled value through stderrTail. Focused validation: 102 tests passed; typecheck clean; hosted CI green.

<details>
<summary>AI fix prompt</summary>

Fix the oversized unterminated-stderr path so truncation cannot split credential-shaped content before redaction. Treat an over-cap line as opaque/truncated (or otherwise preserve redaction context across boundaries), add a deterministic two-chunk boundary regression test, and retain bounded memory plus final partial-line capture.

</details>

src/lib/harnessRunner.ts
lenaparkhodges: Addressed in 145dc7a. The over-cap path no longer emits a fragment at all:

- **Marker instead of slice:** an unterminated line that reaches the pending cap is replaced wholesale by a content-free `[stderr line truncated: exceeded N bytes]` marker, and its streaming remainder is swallowed up to the next newline — nothing from that line is ever emitted independently.
- **Complete lines first:** the newline split now runs before the cap check, so a line is either emitted whole (redactor sees full context) or replaced by the marker; the cap can only ever truncate the trailing unterminated partial, never slice through content mid-line.
- **Stream-end guard:** the final-partial flush is suppressed while mid-discard, so the oversized line's overflow can't reappear after the marker.

Validation: two deterministic regression tests (chunk 1 = label ending exactly at the cap, chunk 2 = naked value + following short line → marker + short line only, no label/value fragment anywhere; stream-end tail not re-emitted after the marker). Full isolated suite 3,690 passed, 2 skipped; typecheck clean.

kaieriksen (approved):
Fix verified at 145dc7a. Oversized unterminated stderr is now replaced with a content-free marker, its remainder is discarded through newline/end-of-stream, and complete subsequent lines remain intact; the credential boundary repro no longer leaks either fragment. Focused validation: 37 runner tests passed; typecheck clean; hosted CI (Linux, Windows, CodeQL) green.

PR #466, opened 2026-08-27
Drop retired config.toml mirrors where the vault is READ, not only on import

### LINKED ISSUE #465: Vault-resident config.toml mirrors permanently override the harness wizard — the #454 drop-list only runs on import
### Symptom

`phantombot harness` (and `/model`) appears to save and then silently reverts. On the fleet today: pi was configured with `z-ai/glm-5.3-flash` for primary/image/coding on six personas, `config.toml` was written correctly every time, and every persona kept running the old model.

### It is not the writer — it is the reader

Verified on a live host:

```
config.toml   [harnesses.pi.routing] primary_model = "z-ai/glm-5.3-flash"   # wizard wrote this correctly
process.env   PHANTOMBOT_PRIMARY_MODEL=z-ai/glm-5.2                          # this is what actually ran
```

Chain:

1. `loadVaultIntoEnv` (`src/index.ts:54`) injects **every** vault row into `process.env` at startup.
2. Affected vaults still hold `PHANTOMBOT_PRIMARY_MODEL`, `PHANTOMBOT_IMAGE_MODEL`, `PHANTOMBOT_CODING_MODEL`, `PHANTOMBOT_PI_PROVIDER` — written by the **pre-#452** wizards, when `~/.env` was the store, and vaulted wholesale by #254's migration before any drop-list existed.
3. `resolveRouting` (`src/lib/piRouting.ts:224`) resolves **env first**: `clean(env[ENV_PRIMARY_MODEL]) ?? tomlStr("primary_model")`.

So a dead setting from an old wizard permanently outranks the only store the current wizard writes.

### Why #454 did not cover it

#452/#454 added `CONFIG_OWNED_ENV_MIRRORS` + `isConfigOwnedEnvMirror` and predicted this failure exactly — `src/config.ts` says a migrated mirror *"would become a permanent override that outranks the only store now being written… would appear to work, then silently revert"*.

But the drop-list has exactly one call site, `vaultMigrate.ts:185`, on the one-way `~/.env` **import**. It protects hosts that migrate *after* the fix. It never sweeps keys **already resident** in a vault — and those are precisely the hosts that migrated under #254, i.e. every host upgraded in the normal order. The stamp then guarantees the import never runs again, so there is no path that can ever heal it.

Fleet check: all six personas (robbie, lena, kai, jake, matt, megan) had all four rows.

### Fix

Enforce the drop-list where the vault is **read**, not only where it is written:

- `readAllVaultValues` withholds any `isConfigOwnedEnvMirror` name from the values it returns, so it never reaches `process.env`. This is the load-bearing half: it covers every persona and every per-turn `reloadVaultForPersona`, and it holds whether or not any cleanup succeeds.
- Evict the row while the vault is open, so `vault list` stops advertising a dead setting and a downgrade cannot resurrect the override. Best-effort; the withholding already made it inert.
- Warn with the names, once per persona. An operator whose setting "didn't stick" needs to see what was overriding it.

### Explicitly out of scope

Env-over-config precedence is **not** reverted. A real shell/systemd `export PHANTOMBOT_PRIMARY_MODEL=…` stays a working escape hatch — #452 chose that deliberately. What this removes is the *vault* fabricating that export behind the operator's back, so the guard keys on the vault row, not on the env var.

### Workaround until it ships

```
phantombot vault unset PHANTOMBOT_PRIMARY_MODEL
phantombot vault unset PHANTOMBOT_IMAGE_MODEL
phantombot vault unset PHANTOMBOT_CODING_MODEL
phantombot vault unset PHANTOMBOT_PI_PROVIDER
# then restart the daemon
```

Check `config.toml` actually carries the values first. Keep `PHANTOMBOT_PI_API_KEY` — that one is a genuine secret.

robertclawson:
Closes #465.

## The bug

`phantombot harness` saves to `config.toml` correctly, and the change does nothing. Verified on a live host before the fix:

```
config.toml   primary_model = "z-ai/glm-5.3-flash"   # what the wizard wrote
process.env   PHANTOMBOT_PRIMARY_MODEL=z-ai/glm-5.2   # what actually ran
```

`loadVaultIntoEnv` injects every vault row into `process.env` at startup, and `resolveRouting` resolves `clean(env[…]) ?? tomlStr(…)` — env first. The vault rows came from the pre-#452 wizards, which wrote model mirrors into `~/.env`, and from #254's migration, which vaulted that file wholesale before any drop-list existed.

## Why the existing drop-list didn't catch it

#452/#454 named these keys and predicted this failure precisely — `src/config.ts` already says a migrated mirror *"would become a permanent override that outranks the only store now being written… would appear to work, then silently revert."*

But `isConfigOwnedEnvMirror` had exactly one call site, `vaultMigrate.ts:185`, on the **import** path. It protects hosts that migrate after the fix; it never sweeps keys already resident. Those are the hosts that migrated under #254 — i.e. every host upgraded in the normal order — and the migration stamp means the import never runs again, so there is no path that can heal them. All six personas on this fleet had all four rows.

## The change

`readAllVaultValues` (`src/lib/vault.ts`):

- **Withholds** any `isConfigOwnedEnvMirror` name from the values it returns, so it never reaches `process.env`. This is the load-bearing half — it is what makes the override inert, it covers every persona and every per-turn `reloadVaultForPersona` (not just startup), and it holds even if the eviction throws.
- **Evicts** the row while the vault is already open, so `vault list` stops advertising a dead setting and a downgrade can't resurrect the override. Best-effort and wrapped: a failed unset costs nothing, because the withholding already made the row inert.
- **Warns with the names**, once per persona per process. The failure mode is silent by nature — someone set a model and watched it not stick — so the log has to name what was overriding it. Names only; this path is shared with real secrets and must never learn to log values.

The eviction is the one write on an otherwise pure read path. It stays defensible because it is bounded to names the vault should never have held, idempotent, and self-extinguishing after the first pass. The open-existing-only guard is untouched: a read still never provisions a persona.

## Explicitly out of scope

**Env-over-config precedence is not reverted.** A real `export PHANTOMBOT_PRIMARY_MODEL=…` from a shell or a systemd unit remains a working escape hatch — #452 chose that deliberately. What goes away is the *vault* fabricating that export behind the operator's back, which is why the guard keys on the vault row rather than on the env var. There's a test pinning that distinction.

## Tests

`tests/vault.test.ts`, four new cases:

1. a mirror row is withheld from env **and** evicted, while a real secret in the same vault still loads (the guard is by name, not a blanket refusal);
2. the `_<PERSONA>` suffixed form is dropped too — it outranked even the persona's own file, so it's the worse of the two forms;
3. a mirror injected by an **older build** and already tracked is *removed* from env on the next reload, so an upgraded daemon heals in place rather than staying poisoned until restart;
4. a shell-exported mirror is left alone.

Mutation-checked: stubbing the guard to `false &&` fails cases 1–3, so they're load-bearing rather than incidentally green.

Full suite: **3683 pass, 0 fail**. `bun tsc --noEmit` clean.

## Docs

`AGENTS.md` invariant 32 said the drop happens at import; it now says the enforcement point is the read, explains why the write-side-only version was insufficient, and records the scope limit. The `CONFIG_OWNED_ENV_MIRRORS` doc comment in `src/config.ts` gains the same pointer.

## Operator note

Affected hosts heal on the next start — no manual step. If someone is on an older build and wants the fix now, `phantombot vault unset` the four keys and restart, after checking `config.toml` carries the values. `PHANTOMBOT_PI_API_KEY` is a genuine secret and stays.


src/lib/vault.ts
```diff
@@ -262,9 +278,23 @@ async function readAllVaultValues(
         badKeys.push(name);
       }
     }
+    // Evict, so `vault list` stops showing settings that no longer do
+    // anything and a later downgrade can't resurrect the override. This is
+    // the one write on an otherwise pure read path: bounded to names the
+    // vault should never have held, idempotent, and self-extinguishing after
+    // the first pass. Best-effort — a failed unset costs nothing, because the
+    // skip above already made the row inert.
+    for (const name of mirrors) {
```
andrewagrahamhodges, line 302: The eviction is the only irreversible step in this PR, and it runs unconditionally — including on the one host shape where the vault row is the *last* copy of the value.

The mirror is normally a duplicate because the pre-#452 wizard wrote `config.toml` and the env mirror together, so on my box eviction is a no-op in effect (all four values are also in `[harnesses.pi.routing]`). But `phantombot env set PHANTOMBOT_PRIMARY_MODEL …` was a supported, documented way to set it *without* touching `config.toml`, and #254 vaulted those files wholesale. On such a host this deletes the setting outright, and the persona silently drops to the built-in default with no way to recover what it used to be — `vault get` is gone too.

Cheapest fix that keeps the hygiene: evict only when the corresponding `config.toml` key resolves to something, and leave the row (still withheld, still inert, still warned about) when it doesn't. Withholding is what makes the bug go away; eviction is cosmetics, and cosmetics shouldn't be the destructive half.

src/lib/vault.ts
```diff
@@ -294,6 +332,32 @@ function warnBadVaultKeys(badKeys: string[]): void {
...
+  if (names.length === 0) return;
+  const sorted = [...names].sort();
+  const signature = `${personaDirPath}\u0000${sorted.join(" ")}`;
+  if (_warnedMirrorSets.has(signature)) return;
+  _warnedMirrorSets.add(signature);
+  log.warn(
+    `vault: dropped ${sorted.length} retired config.toml mirror` +
+      `${sorted.length === 1 ? "" : "s"} (${sorted.join(", ")}) — these are ` +
+      "settings, not secrets, and they outranked config.toml on every " +
+      "startup. The value in config.toml now applies; re-run " +
```
andrewagrahamhodges: "The value in config.toml now applies" is stated as fact, but nothing on this path has looked at `config.toml`. If the key was only ever set through `phantombot env`, what actually applies is the built-in default — and the operator reads a line telling them their config file is in charge while the model quietly changes under them.

Since you already have the persona dir here, resolve the key and branch: name what config.toml holds when it holds something, and say plainly that the setting had no config.toml home and the default now applies when it doesn't. That second case is the one worth waking someone up for, and right now it's the one the message hides. (Config values are settings, not secrets, so naming a model id here doesn't breach the never-log-values rule — the rule is about the vault path being shared with real credentials, and this string is already mirror-only.)

tests/vault.test.ts
```diff
@@ -370,3 +370,86 @@ describe("per-row resilience — one bad row never blanks the vault", () => {
...
+    const env: NodeJS.ProcessEnv = { PHANTOMBOT_PRIMARY_MODEL: "stale/old-model" };
+    const tracked = new Set<string>(["PHANTOMBOT_PRIMARY_MODEL"]);
+    const { removed } = await loadVaultIntoEnv(dir, env, tracked);
+
+    expect(env.PHANTOMBOT_PRIMARY_MODEL).toBeUndefined();
+    expect(removed).toContain("PHANTOMBOT_PRIMARY_MODEL");
+    expect(tracked.has("PHANTOMBOT_PRIMARY_MODEL")).toBe(false);
+  });
+
+  test("a SHELL-exported mirror is left alone — the vault is the only target", async () => {
```
andrewagrahamhodges, line 598: The PR body's strongest claim about the design — withholding "holds even if the eviction throws" — is the one behaviour no test pins. I checked: making `vault.unset` throw fails case 1, but it fails on the *eviction* assertion, so a future refactor that made withholding depend on a successful unset would be caught only incidentally.

One more case: a vault whose `unset` throws, asserting the mirror is still absent from `env` and absent from `updated`. That's the invariant that lets the eviction stay best-effort.

AGENTS.md
```diff
@@ -456,6 +456,6 @@ If your PR does any of the things in the "contributing discipline" list above, *
...
...-persona credential leak. That exclusion is scoped to the leak it exists for: a key injected from THIS persona's own vault is still honoured (`ambientEnvKeyAllowed` in `lib/vaultSecrets.ts`, the one gate both resolvers share), because `loadVaultIntoEnv` deliberately leaves injected secrets in place when the SAME persona's vault fails to open transiently — refusing them there would mute a persona over its own credential, and would have the two paths disagree about what a transient blip means. Vault READS are open-existing-only (`getPersonaSecretStrict` guards on the vault file, same rule as `readAllVaultValues`, #262): a pure read must never mint an identity + nsec for a persona nobody provisioned. **Deprecation warnings cover the PERSONA layer too**, not just the host file (`DEPRECATED_CONFIG_KEYS` / `warnDeprecatedConfigKeys`, checked before `stripHostOnlyKeys`) — since #441 a persona `config.toml` sets the same keys, and that is exactly where a silent edit hides. Note "retired" is not "ignored": `turn_timeout_s` is still honoured as an alias, and each entry's `advice` says which it is. Out of scope and still open: Telegram tokens and Gemini embedding keys living in `config.toml`.
```
andrewagrahamhodges: Invariant text is good and lands in the right place. One thing it now contradicts: `README.md` around L546 still tells the reader that `phantombot harness --persona <name>` and `/model` write "the env mirror … suffixed with the persona name", and lists `PHANTOMBOT_PRIMARY_MODEL_LENA` et al as live. That was already stale after #454 stopped the wizards writing mirrors; after this PR it's actively harmful, because it points an operator at the exact names the daemon now deletes on sight. Worth correcting in this PR — it's the same paragraph a confused operator lands on when a model won't stick.

andrewagrahamhodges (changes requested):
Root cause is right and the premise checks out on a live host — my own vault carries all four rows (`PHANTOMBOT_PRIMARY_MODEL`, `PHANTOMBOT_CODING_MODEL`, `PHANTOMBOT_IMAGE_MODEL`, `PHANTOMBOT_PI_PROVIDER`) with `config.toml` holding a different `primary_model`, which is exactly the reported symptom. Withhold-at-read is the correct enforcement point: `readAllVaultValues` is the single funnel into `process.env` (one caller, `loadVaultIntoEnv`, reached by both startup and every per-turn `reloadVaultForPersona`), and it is not on the `vault list` / `vault get` CLI path, so the one write stays scoped to the env loader. I re-ran `tests/vault.test.ts` locally (23 pass) and reproduced your mutation claim: stubbing the guard to `false &&` fails exactly cases 1–3, so they are load-bearing. Also confirmed `isConfigOwnedEnvMirror`'s `startsWith(base + "_")` doesn't collide with any real secret name in tree — `PHANTOMBOT_PI_API_KEY` is safe next to `PHANTOMBOT_PI_BIN`/`_PROVIDER`.

Two things to fix before this goes in, both cheap, both about the half of the problem the read path can't see:

1. **The write path is still wide open.** `runVaultSet` has no drop-list check, and `phantombot env` is a live deprecated alias forwarding straight to it. So `phantombot env set PHANTOMBOT_PRIMARY_MODEL …` — which is what every pre-#452 doc, runbook and muscle memory says to do — still prints `saved PHANTOMBOT_PRIMARY_MODEL to the vault`, and this PR now makes the daemon delete it on the next read. That converts "set it and it silently reverts" into "set it, get told it saved, and watch it vanish": the same silent failure this PR exists to kill, at the one moment the operator is standing right there and could be told. Reject the name in `runVaultSet` with a pointer to `phantombot harness` / `config.toml`.

2. **The eviction is irreversible and the warning asserts a fallback it never checked.** See the inline note on the warn text.

Plus a README staleness fix and a test gap, inline. Happy to re-review quickly once those land.

src/lib/vault.ts
robertclawson: Fixed in d93ef0c. `CONFIG_OWNED_ENV_MIRRORS` is now a map from mirror name to the `[harnesses]` key it was retired into, and `configOwnedEnvMirrorSetting` (config.ts) resolves that key against the persona file then the host file — files only, never env, since the mirror being judged is itself in the environment. Eviction runs only when that resolves; otherwise the row is kept, inert, and named in the warning. A `_<PERSONA>` variant always resolves undefined, because its home is another persona's file and cannot be identified from a name alone — unknown belongs on the non-destructive side. Withholding is untouched in both branches. Pinned by `a mirror with NO config.toml home is withheld but NOT deleted`; making the eviction unconditional again fails exactly that case.

src/lib/vault.ts
robertclawson: Fixed in d93ef0c. The warning now branches per mirror on the same resolver: `PHANTOMBOT_PRIMARY_MODEL: dropped from the vault; harnesses.pi.routing.primary_model = "new/wanted-model" now applies.` versus `PHANTOMBOT_PRIMARY_MODEL: no config.toml setting for it, so the BUILT-IN DEFAULT now applies and the value may change. The vault row is kept (inert) so the old value is still readable with `phantombot vault get` — set it for real with `phantombot harness`, then remove the row.` Agreed on the value: it is read out of a config FILE, not the vault, so naming it does not touch the never-log-a-vault-value rule.

tests/vault.test.ts
robertclawson: Added in d93ef0c: `withholding survives an eviction that THROWS`. It asserts the row is still on disk (so the case cannot silently become a happy-path duplicate) AND that the mirror is absent from env, absent from `updated`, absent from `tracked`, with the real secret still loaded. One deviation from the obvious implementation — I make the DELETE fail with a `BEFORE DELETE` trigger rather than read-only file permissions, because `tests/vault.test.ts` is in the Windows CI subset and a read-only attribute does not fail the same way there; the trigger is also narrower, failing the eviction without disturbing the reads around it. Mutation-checked: making the skip depend on a successful unset fails this test and only this test.

AGENTS.md
robertclawson: Fixed in d93ef0c. The per-persona bullet no longer lists the suffixed mirrors as things the wizard writes — it says config.toml is the only persisted store, notes the in-process env write `/model` still does so a switch takes effect without a restart, and keeps the precedence ladder (a real shell export is still read). Added a bullet for the actual symptom, "if a model you set will not stick", naming the vault-resident mirrors and both outcomes. Also cleaned the `/model` per-harness table further down, which still gave `PHANTOMBOT_CLAUDE_MODEL` / `PHANTOMBOT_CODEX_MODEL` as where the value lives.

robertclawson:
All four addressed in d93ef0c — full suite green locally (3690 pass), typecheck clean.

- **Write path closed.** `runVaultSet` refuses a `CONFIG_OWNED_ENV_MIRRORS` name before it reads stdin, exit 2, pointing at `phantombot harness` / `[harnesses]`. `phantombot env set` inherits it by forwarding. Four cases including the alias and a `PHANTOMBOT_PI_API_KEY` prefix guard.
- **Eviction conditioned.** The list is now a map to the `[harnesses]` key each mirror was retired into; `configOwnedEnvMirrorSetting` resolves it persona-file-then-host-file, and the row is deleted only when that answers. No home ⇒ kept, inert, named. Withholding unchanged either way.
- **Warning branches** on the same resolver, naming the key and value that take over, or saying plainly that the built-in default now applies and the row was kept so the old value is still readable.
- **README** no longer advertises the suffixed mirrors as live, plus a "if a model will not stick" bullet; the `/model` table now names config.toml keys.

Every new guard mutation-checked: unconditional eviction fails the no-home case, dropping the CLI refusal fails three, and making withholding depend on a successful unset fails the new throwing-eviction test and only that one. That test blocks the DELETE with a `BEFORE DELETE` trigger rather than read-only permissions, since this suite also runs on the Windows CI runner.

src/lib/vault.ts
```diff
@@ -262,9 +284,38 @@ async function readAllVaultValues(
...
+    // ONLY when config.toml states the setting the mirror was retired into.
+    // Withholding is what fixes the bug and it is free; eviction is hygiene
+    // and it is irreversible. `phantombot env set PHANTOMBOT_PRIMARY_MODEL …`
+    // was once a supported way to set the value without touching
+    // config.toml, so on such a host the row is the LAST copy and deleting it
+    // would drop the persona to a built-in default with nothing left to read
+    // the old value from. Those rows stay — listed, inert, and named in the
+    // warning — which is the recoverable side of the choice.
+    for (const name of mirrors) {
+      const setting = await configOwnedEnvMirrorSetting(name, personaDirPath);
```
andrewagrahamhodges: This `await` can throw: `configOwnedEnvMirrorSetting` -> `tryReadToml` re-throws anything that isn't ENOENT (TOML parse error, EACCES, EISDIR). There's no `catch` between here and the caller, so it escapes `readAllVaultValues` and out of `loadVaultIntoEnv` — which documents "Never throws" and is the fail-partial path for the whole persona's secrets.

It also contradicts the comment at line 292: the withholding guard no longer holds "whether or not the eviction below ever succeeds", because a failed *lookup* now aborts the read before any value is returned. Reproduced against this head with an unparseable persona `config.toml` (stack in the review body); the same case passes on `6bb662f`.

Wrap it and fall back to `undefined` — unknown already means "don't evict":

```ts
let setting: Awaited<ReturnType<typeof configOwnedEnvMirrorSetting>>;
try {
  setting = await configOwnedEnvMirrorSetting(name, personaDirPath);
} catch {
  setting = undefined;
}
```

andrewagrahamhodges (changes requested):
Re-reviewed at `d93ef0c`. All four earlier points are addressed, and addressed the way I'd have argued for: the CLI now refuses a mirror *before* it reads stdin, eviction is conditioned on `config.toml` actually stating the setting, the warning says two different things for two different situations, README no longer sells the env mirrors as live, and `withholding survives an eviction that THROWS` pins the design claim with a `BEFORE DELETE` trigger rather than file permissions (portable, and narrow enough that it can't accidentally pass for the wrong reason). `CONFIG_OWNED_ENV_MIRROR_HOMES` maps to the same `[harnesses]` paths `loadConfig` reads, and the persona-file-then-host-file order matches the file layers in the precedence comment at config.ts:1096. Suffixed variants landing on the non-destructive side is the right call. Local: typecheck clean, 129/129 across vault, cli-env and config.

One regression to fix, then I'm happy.

**The new `await` re-opened the hole the trigger test closes.** `configOwnedEnvMirrorSetting` calls `tryReadToml`, which re-throws everything that isn't ENOENT — a parse error, EACCES, EISDIR. That await sits inside the `try { … } finally { vault.close() }` with no `catch`, so the throw escapes `readAllVaultValues` and out through `loadVaultIntoEnv`, whose contract says "Never throws" and whose whole design is fail-partial. The comment eight lines above still promises the guard holds "whether or not the eviction below ever succeeds", and it no longer does: eviction can now take the *entire* vault read with it.

I reproduced it against this head — a persona dir with an unparseable `config.toml` and a vault holding `PHANTOMBOT_PRIMARY_MODEL` + `GITHUB_TOKEN`:

```
error: Invalid TOML document: incomplete key-value: cannot find end of key
  at tryReadToml (src/config.ts:2324:12)
  at configOwnedEnvMirrorSetting (src/config.ts:2462:24)
  at readAllVaultValues (src/lib/vault.ts:303:29)
  at loadVaultIntoEnv (src/lib/vault.ts:490:24)
```

The same case passes on `6bb662f`, so it arrived with this commit. It only bites a vault that holds a mirror row — i.e. exactly the poisoned hosts this PR exists for — and the failure mode is the persona losing every secret it has, which is a much worse day than the stale model id.

Fix is the shape you already used one line below: treat an unreadable config as unknown, which is the non-destructive answer you're already committed to.

```ts
let setting: Awaited<ReturnType<typeof configOwnedEnvMirrorSetting>>;
try {
  setting = await configOwnedEnvMirrorSetting(name, personaDirPath);
} catch {
  // Unreadable/unparseable config.toml: we cannot prove a home exists, and
  // eviction is hygiene. Withholding already fixed the override.
  setting = undefined;
}
```

Worth a test alongside `withholding survives an eviction that THROWS` — same shape, malformed `config.toml` instead of the trigger, asserting `GITHUB_TOKEN` still lands in env. That is the one case where the difference between "withholding is load-bearing" and "eviction is best-effort" is worth real money.

Nit, no action needed: a persona that ran `/model clear` on codex has `model = ""` in its file, and the empty-string check reads that as no home — so the row is kept and warned about forever. Correct behaviour (the built-in default really does apply), just permanently noisy. Fine to leave on the safe side.

src/lib/vault.ts
robertclawson: Fixed in 94ce5d1 — and you are right that it arrived with d93ef0c, not before it.

The `await` is now wrapped at the call site and an unreadable `config.toml` resolves to `undefined`, which is the side this decision was already designed to fall to: unknown means the mirror is withheld from the returned values and the row is **kept**, exactly as the `_<PERSONA>` variants are. Nothing can prove another copy of the value exists, so nothing is deleted. The failure is logged at debug with the name and the error, not swallowed silently.

I left `tryReadToml` re-throwing. Its callers in `config.ts` want to know that a config is malformed — swallowing there would make a typo in `config.toml` silently boot with defaults, which is a worse bug than this one. The contract that matters is `readAllVaultValues`, and that is where the catch belongs.

Test added: `withholding survives a config.toml that cannot be PARSED` — malformed TOML plus a mirror row, asserting `GITHUB_TOKEN` still reaches env and is in `updated`, the mirror is still inert, and the row is still listed. Mutation-checked by restoring the bare `await`: it fails with `Invalid TOML document: incomplete key-value`, so it is load-bearing rather than a happy-path duplicate.

The `/model clear` nit is real and I agree it is noise-only — the row is kept and warned about forever with an empty `model = ""` reading as no home. Not touching it here; it needs `/model clear` to remove the key rather than blank it, which is a different change.

Full suite 3691 pass / 0 fail, typecheck clean.

andrewagrahamhodges (approved):
Approving.

The regression is properly contained. `configOwnedEnvMirrorSetting` is now wrapped at the call site with the answer falling back to `undefined`, which is the right shape: "unknown" was already the non-destructive side of the eviction decision (the `_<PERSONA>` variants land there for the same reason), so an unreadable config.toml withholds and **keeps** the row rather than costing the persona every secret.

Verified on `a4ba0ec`:

- The new test is load-bearing — deleting the try/catch and re-running fails `withholding survives a config.toml that cannot be PARSED` and nothing else, so it pins the containment rather than the syntax.
- It asserts the right thing: `GITHUB_TOKEN` still reaches env **and** `updated`, the mirror stays withheld and untracked, and the row survives in `vault.list()`. That covers both halves of the contract, not just "didn't throw".
- `configOwnedEnvMirrorSetting` has exactly one caller, so there is no second unguarded path.
- 27/27 in `tests/vault.test.ts`, typecheck clean, full suite green apart from two pre-existing env-dependent failures on this box (`lib-indexRefresh` sqlite, `harnesses-claude` PATH) that are unrelated to the diff and pass in CI.

Nice touch putting the failure on `log.debug` while the operator-facing warning still names the row as orphaned-and-kept — the person who needs to act still gets told, without a second alarm for a config problem they already have.

PR #468, opened 2026-08-27
Degrade over-budget journal entries to stubs, and weight recall by class not clock

### LINKED ISSUE #467: Journal recall: degrade over-budget entries to stubs instead of dropping them, and weight selection by class not clock
### Problem

`selectForRecall` bounds the injected journal at 16KB by **dropping** the entries that do not fit. Measured on robbie today (2026-08-27), the first full day on the row path:

- 30 self rows, **23.5KB** rendered, against a 16KB budget
- 16 newest injected, **14 oldest dropped**, 12 `source='task'` rows withheld by class

Every one of those 30 rows is tagged. The class tier that #463 added — untagged narration is dropped before a tagged capture — bought **nothing**, because there was no narration to sacrifice. The 7.5KB of overflow came entirely out of the tier the design says is protected: the morning's `decision` and `lesson` captures.

Two distinct defects fall out of that.

**1. A dropped entry is invisible, so the memory reflex cannot fire.** The block says *"14 earlier entries not shown"* and points at `memory search`, which is honest but not actionable: a count is not a query. The turn cannot search for something whose existence it does not know. In practice the morning's root cause leaves the prompt by mid-afternoon and gets re-derived — the exact failure #461 was opened to stop. Truncation the model cannot see reads as absence.

**2. Priority inside the tagged tier is pure recency.** `byPriority` reverses each class and packs newest-first, so at 18:00Z a 17:24 *"CI green"* note outranks a 09:00 root-cause `decision`. Age is a poor proxy for usefulness within a single day, and it is the only axis currently in play once narration runs out.

### Proposal

**A. Degrade instead of dropping.** An entry that does not fit in full is rendered as a bounded one-line stub — tags, timestamp, and the first ~80 characters of the content:

```
- [decision] phantombot #461 stage one shipped as PR #463 (branch journal-rows… · 14:30Z · elided — `memory search`
```

~140 bytes against a 700-byte capture. The budget stops being a cliff: a heavy day degrades to an *index of itself* rather than an amputation, and the stub carries the search terms that make `memory search` reachable. This applies to oversized rows too (a single 200KB capture is currently the one case that can never come back on a quiet day); it becomes a stub like any other.

Stubs are paid for out of the same 16KB. A reserve of ~15% of the budget is held back from the full-entry pass and spent on stubs in priority order; whatever still does not fit is dropped and counted as today. The bound stays hard — that is not negotiable, it is what stops the E2BIG class of failure (#426).

**B. Weight the ordering.** Keep the class tier (it is the guarantee #463 makes and the file path could not), but replace pure recency *inside* it with banded scoring that mirrors the drawer model already in `drawers.ts`:

- `commitment` rows do **not** decay — age makes a commitment more urgent, not less relevant, exactly as `BELIEF_KINDS` excludes them
- other tagged rows score `weight · 2^(-ageHours / 24)`, with `decision` > `lesson` > `norm` > `person` > any other tag
- untagged narration keeps recency order, and is still dropped first

The reference clock is the newest entry in the set, not wall time, so selection is deterministic and testable.

### Non-goals

- **Not raising the 16KB budget.** That is a separate lever with the worst scaling: a 121KB day still blows it. Revisit only if A+B prove insufficient in practice.
- **Not summarising with an LLM.** Recall runs on every turn; it must stay mechanical and free.
- Not touching the file fallback path, `renderDay`, or anything the nightly renders — stubs exist only in the injected block and never reach disk.

### Acceptance

- A day that overflows returns full entries **and** stubs, chronologically interleaved, with total bytes `<=` budget
- An entry individually larger than the whole budget appears as a stub, not as a silent `droppedOversize`
- The block distinguishes "shown as a stub" from "not shown at all"
- A `commitment` from the morning outranks a same-size `person` note from an hour ago
- Untagged narration is still dropped before any tagged capture, at any age

Part of #461.

robertclawson:
Closes #467. Part of #461.

## The bug this fixes

The 16KB recall budget bounded the prompt by **dropping** what did not fit. On robbie, 2026-08-27 — the first full day on the row path — that came out as:

| | before | after |
|---|---|---|
| rows (self + heartbeat) | 30 | 30 |
| rendered in full | 16 | **14** |
| stubbed | — | **14** |
| dropped with no trace | **14** | **2** (named in the block) |
| injected bytes | 16.0KB | 16.3KB |

Every one of those 30 rows is tagged, so the class tier #463 added — narration before captures — had nothing cheap to sacrifice, and the overflow came straight out of the tier the design calls protected. And *"14 earlier entries not shown"* is honest but unsearchable: a turn cannot look up something whose existence it does not know. The morning's root cause leaves the prompt by mid-afternoon and gets re-derived, which is the failure #461 was opened to stop.

## Degradation instead of dropping

An entry that cannot be shown in full comes back as a bounded one-line stub — tags, clock, first ~80 characters, `… · elided`:

```
- [decision,lesson] Opened phantombot #459 (resume-with-context): on idle-kill AFTER output was produced, persist… · elided — `memory search` · 07:18Z
```

- **Stubs are paid for out of the same 16KB.** ~15% reserved on an overflowing day, floored at one whole stub (the fraction alone is useless at small budgets), capped at half the budget, and **nothing reserved on a day that fits**. The bound is hard over everything emitted — that is what keeps the E2BIG tripwire (#426) demoted.
- **An oversized row is stubbed like any other.** The one case that previously could never come back on a quiet day now leaves a searchable trace.
- **`droppedForBudget` still means "not shown in full"; `droppedEntirely` is new** and means no trace at all. The block distinguishes them, because the remedies differ.
- **Stubs never reach disk.** `renderEntry` / `renderDay` feed the artefact the nightly writes and `parseJournalLine` reads back; a lossy line there would overwrite a day's rows with a summary of themselves. There is a test pinning that.

The trade is explicit: on an overflowing day the reserve costs up to one entry of full prose, in exchange for an index of everything else.

## Weighted ordering

Bands stay **strict** — commitments, then other tagged captures, then untagged narration; no weight at any age lifts narration above a capture. What changes is ordering *inside* the tagged band, which was pure recency:

- `commitment` does not decay and sorts first — age makes a commitment urgent, the same split `BELIEF_KINDS` already makes in `drawers.ts`. One decay model in the codebase, not two.
- everything else scores `weight · 2^(-ageHours / 48)`, `decision` > `lesson` > `norm` > `person` > any other tag.
- the decay clock is **the newest row in the set, not `Date.now()`**, so selection is a pure function of its input; a wall clock makes the same journal select differently depending on when the turn ran, and no test can pin that.

48h is chosen so the weight table can actually change an ordering across a working day: over the ~9h between a morning `decision` and an afternoon `person` note the factor is 0.88, so the decision survives. Shorter and weight stops mattering; longer and the day stops being ordered by time.

## Deliberately not in scope

- **Not raising the budget.** Worst scaling of the three levers — a 121KB day still blows 24KB. Revisit only if this proves insufficient in practice.
- **No LLM summarisation.** Recall runs on every turn; it stays mechanical and free.
- File fallback path, `renderDay`, nightly render: untouched.

## Verification

- Full suite **3772 pass / 0 fail**, typecheck clean.
- Mutation-checked, not just covered: removing the remaining-budget check in the stub pass fails 4 tests (including the byte bound on `renderRecall`); collapsing the band comparison to a constant fails the three ordering tests.
- Measured against the live journal that motivated the issue, not a fixture — the table above.


src/memory/journal.ts
```diff
@@ -547,49 +769,73 @@ export function selectForRecall(
...
+
+  const kept = [...pass.kept].sort(chronological);
+  stubbed.sort(chronological);
   return {
     entries: kept,
+    stubbed,
     droppedForBudget: eligible.length - kept.length,
-    droppedOversize,
+    droppedEntirely: eligible.length - kept.length - stubbed.length,
+    droppedOversize: pass.oversize,
```
kaieriksen: **Minor — `droppedOversize` silently changes meaning on the second pass.** `packFull` marks a row oversize when `size > budget`, but pass 2 runs against `budgetBytes - reserve`, so a row between ~85% and 100% of the *full* budget gets counted here as oversize and fires the dailyRecall note claiming it is "individually larger than the whole 16384-byte journal budget" — factually false (reproduced: a 15,000-byte row against a 16,384-byte budget on an overflowing day yields `droppedOversize: 1`). This also breaks the documented contract on `RecallSelection.droppedOversize` ("bigger than the whole budget"), whose whole purpose was distinguishing the remedy. Fix: track oversize against the full `budgetBytes` (or carry it from pass 1 only). Non-blocking: the bound still holds and the row is still stubbed.

src/memory/journal.ts
```diff
@@ -507,35 +525,239 @@ export interface RecallOptions {
...
+ * untouched, so nothing lossy can ever be written back over a day's rows.
+ */
+export function renderStub(e: JournalEntry): string {
+  const tags = e.tags.length > 0 ? `[${e.tags.join(",")}] ` : "";
+  const stamp = ` · ${e.createdAt.toISOString().slice(11, 16)}Z`;
+  const fixed = Buffer.byteLength(
+    `- ${tags}${STUB_SUFFIX}${stamp}`,
+    "utf8",
+  );
+  const room = Math.max(8, JOURNAL_STUB_MAX_BYTES - fixed);
```
kaieriksen: **Nit — the 160-byte stub cap is not actually a cap.** `fixed` includes the tag list, so a long tag list makes `fixed > JOURNAL_STUB_MAX_BYTES` and the `Math.max(8, …)` floor lets the stub through at `fixed + 8` — reproduced at 165 bytes with 20 tags. Harmless today (the stub's real size is measured into the budget either way), but the constant's "hard cap" doc comment overpromises. Either subtract unbounded tag bytes into the head budget or soften the doc.

kaieriksen (dismissed):
Verdict: approve. The hard budget bound holds — I fuzzed 2,000 randomized selections against `renderRecall` output and found no violation; two-pass packing with the reserve never exceeds the budget, `droppedEntirely` accounting is exact, the decay clock keeps selection a pure function of the rows, and the stubs-never-reach-disk test pins the round-trip invariant. Two Minor findings inline, neither blocking.

src/memory/journal.ts
robertclawson: Fixed in 7d443c9. Confirmed your repro — the count was read off pass 2, which packs against `budgetBytes - reserve`, so a 15,000-byte row under a 16,384-byte budget on an overflowing day came back as `droppedOversize: 1` and the note asserted it was larger than the whole budget. It is now taken from pass 1, the pass that runs at the full budget, with a comment saying why it is captured before the re-pack. Regression test packs exactly that shape (15,000-byte row + 60 short rows so the reserve is the full 15%) and asserts `droppedOversize === 0` with `droppedForBudget > 0`; mutation-checked by restoring `pass.oversize`, which fails it.

src/memory/journal.ts
robertclawson: Fixed in 7d443c9 — took the first option, since softening the doc would leave the only per-row bound on the injected block as a soft one. Tag list is now capped at `JOURNAL_STUB_TAG_BYTES` (40) with a `+N` marker for the remainder, so `fixed` is bounded by construction (2 prefix + 43 tags + 33 suffix + 10 stamp = 88 worst case) and the `Math.max(8, …)` floor is gone: the head room is now positive arithmetic. A single tag longer than the whole allowance lands as `[+1]` rather than overrunning. Two tests: 20 tags and one 500-byte tag, both asserting `<= JOURNAL_STUB_MAX_BYTES`; both fail if the old expression is restored.

kaieriksen (approved):
Re-review verified at 7d443c9: both Minors resolved and no fix-caused regressions in the delta. DroppedOversize is now measured from pass 1 at the full budget, and the stub cap holds by construction — the tag list is capped at JOURNAL_STUB_TAG_BYTES (worst-case fixed bytes ≈ 2 prefix + 1 space + 40 tags + 2 brackets + ~32 suffix + 10 stamp < 90), so `room` is always positive with the Math.max floor gone. The 20-tag and 500-byte-tag tests pin the bound. Approving.

src/memory/journal.ts
```diff
@@ -507,35 +525,275 @@ export interface RecallOptions {
...
+  const tags = stubTagList(e.tags);
+  const stamp = ` · ${e.createdAt.toISOString().slice(11, 16)}Z`;
+  const fixed = Buffer.byteLength(
+    `- ${tags}${STUB_SUFFIX}${stamp}`,
+    "utf8",
+  );
+  // No `Math.max` floor here: `fixed` is bounded by construction (constant
+  // prefix + suffix + stamp, plus a tag list capped above), so the room left
+  // for head text is always positive and the cap is arithmetic, not a hope.
+  const room = JOURNAL_STUB_MAX_BYTES - fixed;
```
kaieriksen, line 718: Verified: `fixed` is bounded by construction now — the tag list is capped via stubTagList, so `room` is always positive and the removed Math.max floor is genuinely unreachable. The 20-tag and single-500-byte-tag tests both assert <= JOURNAL_STUB_MAX_BYTES, and the marker path lands as [+1] correctly.

src/memory/journal.ts
```diff
@@ -547,49 +805,80 @@ export function selectForRecall(
...
+  // First pass at the FULL budget: a day that fits pays nothing for a stub
+  // reserve it does not need.
+  let pass = packFull(ordered, budgetBytes);
+  // Measured against the FULL budget, and kept from this pass only: pass 2
+  // runs against `budgetBytes - reserve`, where a row at 90% of the budget
+  // would also count as "oversize" and make the dailyRecall note claim it is
+  // bigger than the whole budget — which is the one thing that count exists
+  // to mean, because it is what distinguishes "returns on a quiet day" from
+  // "never returns; split it or search for it".
+  const oversize = pass.oversize;
```
kaieriksen, line 819: Verified: oversize is captured from pass 1 at the full budget before the re-pack, with the reasoning recorded in place. The new regression test (15,000-byte row + 60 short rows) pins droppedOversize === 0 while droppedForBudget > 0. No fix-caused regressions in the delta.

kaieriksen (commented):
Fix verification (re-review) — inline confirmations on both fixed hunks.

PR #470, opened 2026-08-27
Let a virtual note own its path against an older file (#469)

### LINKED ISSUE #469: Journal virtual note is clobbered by refreshStale whenever a file exists at the open day's path
## Symptom

On a host running v1.1.312, `memory search` cannot reach today's newest journal entries, while the recall block truthfully tells the agent *"Nothing is lost — every entry is indexed: `phantombot memory search`"*.

Observed on robbie, 2026-08-27:

```
memory/2026-08-27.md   mtime_ms=1787853257879  size=40319  virtual=0  indexed_at=2026-08-27T20:30:39Z
```

`virtual = 0`, and `size` is the size of the **on-disk markdown file**, whose content stops at 17:24Z. The 15 journal rows written after that (all tagged `decision`/`lesson`) are in `journal_entries` and are correctly selected into recall — but they are **not in the search index at all**. Searching a literal string that appears only in a later row (a GitHub run id) returns unrelated turns from months ago.

So #463's headline durability claim — "index-on-write via VIRTUAL notes closes the 24h search hole" — is not in effect whenever a file exists at the open day's path.

## Root cause

`MemoryIndex.refreshStale()` treats `virtual` as authoritative in exactly one of its two loops.

The *removal* loop honours it (memoryIndex.ts:767):

```ts
// A virtual note has no file to go missing, so "not on disk" is its
// normal state, not evidence of a delete.
if (rec.virtual) continue;
```

The *reindex* loop does not (memoryIndex.ts:777):

```ts
for (const f of live) {
  const prev = recorded.get(f.path);
  if (!forceAll && prev && prev.mtimeMs === f.mtimeMs) continue;
  // ... deletePath(f.path) then INSERT from disk, virtual defaults to 0
}
```

A virtual note is written with `mtime_ms = -1` *deliberately*, so that no "unchanged since last index" shortcut can skip it. When there is no file that is exactly right. When there **is** a file at that path, the same `-1` guarantees the opposite of what it was for: the mtime can never match, so **every** `refreshStale()` reindexes the day from disk and downgrades the virtual row to `virtual = 0` with the stale file's body.

And `refreshStale()` runs constantly — heartbeat (heartbeat.ts:167), `indexRefresh`, `memory index`, and `orchestrator/retrieval.ts:99` on retrieval turns. So `indexOpenDay()` wins for a few seconds after each capture and then loses for good. The file always wins.

Why a file exists at the open day's path at all:

* **Upgrade mid-day** (this host): the day was a pre-#463 markdown file, `absorbDay` turned its lines into rows, and left the file in place by design. Rows and file now disagree permanently for the rest of the day.
* Any hand-edit, restore, or sync that recreates `memory/<date>.md`.
* A *rendered* day whose rows have not been pruned yet — benign today only because the render also calls `dropOpenDayIndex`.

The second-writer instance of this was found in #463 review round 1 (drawer-import appending `memory/<date>.md` directly, cli/memory.ts:865-872). That fix removed **one writer**; it did not make the virtual note authoritative, so any other route to a file at that path reopens the same hole.

## Severity

The failure is silent and self-concealing: recall still shows the day (it reads rows), and it prints a pointer to a search that cannot find them. #468 made over-budget entries degrade to stubs whose entire value is being a searchable pointer — a stub for an unindexed row is a dead link. Same failure shape as #467, one layer down.

## Proposed fix

1. **`refreshStale()`: a virtual row owns its path.** Skip a live file whose recorded row has `virtual = 1`, in the reindex loop as well as the removal loop. The journal relinquishes the path explicitly via `removeVirtualNote()` (`dropOpenDayIndex`, after the nightly has rendered and fingerprinted the artefact); until then the rows are the newer truth and the file is a stale shadow of them. This also covers `rebuild()`, which snapshots virtuals, re-inserts them, and then calls `refreshStale(forceAll)` — today that forced pass clobbers every virtual it just restored if a file exists.
2. **Self-heal, no migration.** Once (1) is in, the next `indexOpenDay()` (every capture, and every heartbeat) rewrites the virtual note with the full day and it *sticks*. No backfill step, no host surgery.
3. **Consider: log the shadow.** When `refreshStale` skips a live file because a virtual note holds the path, log at debug with the path — a file that persists there after the day is rendered and dropped is the one shape this rule could hide.

Explicitly **not** proposed: making `absorbDay` delete or archive the file it absorbed. #463 decided pre-existing days stay as files, the render rewrites the artefact at midnight anyway, and a deletion in the read path is exactly the kind of destructive self-heal this project keeps out of the maintenance loop.

## Regression tests

* Day has rows, a **stale** `memory/<date>.md` exists on disk, `refreshStale()` runs → the indexed body is still the rows, `virtual` is still 1, and a term that appears only in a post-absorb row is findable.
* Same, with `forceAll = true` and via `rebuild()`.
* After `removeVirtualNote()`, `refreshStale()` indexes the real file normally (`virtual = 0`) — the handover still works.
* Mutation check: restoring the unguarded reindex loop must fail the first test.

robertclawson:
Fixes #469.

## The bug

`refreshStale()` honours `files.virtual` in its **removal** loop but not in its **reindex** loop. A virtual note is written with `mtime_ms = -1` precisely so that no "unchanged since last index" shortcut can skip it — which, when a real file *does* exist at that path, guarantees the opposite of what the sentinel was for: every pass re-indexes the file over the live rows and downgrades `virtual = 1 → 0`.

There is a file at that path on any host where `absorbDay` turned a pre-#461 markdown day into rows and left the file alone (by design — #461 does no bulk migration). `refreshStale` runs on the heartbeat, on `memory index`, inside `rebuild()` and on every retrieval turn, so `indexOpenDay` wins for seconds after each capture and loses permanently.

Measured on this host: the index row for `memory/2026-08-27.md` was `virtual=0, size=40319` — the stale on-disk body, stopping at 17:24Z. The 15 rows written after that were not in the search index at all; searching a literal string from one of them returned unrelated turns from months earlier. Meanwhile the recall block truthfully printed *"nothing is lost — `memory search`"*.

It also makes #468's stubs dead links. A stub is a pointer; the pointer is only worth its bytes if the entry behind it is reachable.

## The fix

- The reindex loop skips a live file whose recorded row is `virtual = 1`. Checked **before** `forceAll`, because `rebuild()` passes `forceAll` to re-derive everything from disk and a virtual note is the one thing in the index that is not.
- Ownership is not unconditional: a file **modified since** the note was published is the nightly artefact taking over. `dropOpenDayIndex` still relinquishes explicitly after a verified render; the mtime comparison is the net that makes the handover self-heal if that call never ran (a crash, an index that would not open).
- `files.indexed_at` becomes load-bearing rather than bookkeeping, so `rebuild()` now carries each restored note's **original** publish time. Re-dating the snapshot to "now" would make every artefact written before the repair look older than the stand-in it replaced — the documented repair command would pin the stale copy in place.
- The comparison is **strict**, so a tie falls to the file. mtime and `indexed_at` agree only to the millisecond, and the two mistakes are not symmetric: losing the row to a file is repaired by the next capture republishing the note with a later timestamp, while an over-held stand-in shadows the artefact for good and the rows behind it are pruned once the render verifies.

Self-heals on the next capture. No backfill, no migration.

## Tests

Three new, all mutation-checked by restoring the old code:

- `a stale file at the open day's path cannot clobber the live rows` — the exact repro, an older file plus a capture, refreshed twice. Fails without the guard.
- `a file written after the virtual note still takes the path over` — the crash-before-relinquish case.
- `a rebuild keeps the virtual note's original publish time` — fails if `upsertVirtualNote` re-dates the restored row; so does the pre-existing `a rebuild still lets the real file take over the virtual path`.

Full suite 3778 pass / 0 fail, typecheck clean. AGENTS.md invariant 34 records the ownership rule and why each half of it is load-bearing.


src/lib/memoryIndex.ts
```diff
@@ -776,6 +796,36 @@ export class MemoryIndex {
...
+      // It is checked BEFORE forceAll on purpose: a rebuild passes forceAll
+      // to re-derive everything from disk, but a virtual note is not derived
+      // from disk — that is the whole reason rebuild() carries it across the
+      // delete.
+      //
+      // The comparison is STRICT so a tie hands the path to the file. mtime
+      // and indexed_at agree only to the millisecond, so equality means "we
+      // cannot tell", and the two mistakes are not symmetric: losing the row
+      // to a file is repaired by the next capture, which republishes the note
+      // with a later timestamp and takes the path straight back, while
```
andrewagrahamhodges: This compares two different clocks, and the file loses a race it should win.

`f.mtimeMs` comes from the kernel, which stamps inodes from the **coarse** clock (`current_time()`, granularity one tick). `indexed_at` comes from `new Date()` in-process, off the fine clock. A file written *strictly after* the virtual note routinely lands with an mtime *earlier* than the note's `indexed_at` — so "the artefact takes the path over" quietly fails, which is the exact non-recoverable direction the comment above argues we must avoid.

Measured on this box (300 writes each, `Date.now()` sampled immediately **before** the write):

```
tmpfs : file mtime < prior Date.now() in 300/300 writes, worst -1.68 ms
ext4  : file mtime < prior Date.now() in 299/300 writes, worst -1.63 ms
```

Not a tie, and not one-in-a-thousand — it is the *normal* result, up to a full tick (10 ms on a `CONFIG_HZ=100` kernel). So "the comparison is STRICT so a tie falls to the file" is true of the operator and false of the data: the guard is biased against the file by up to one tick.

It already breaks a test. `a rebuild still lets the real file take over the virtual path` (line 1008, pre-existing, writes the artefact with no sleep) fails deterministically here — 5/5 runs — and passes on `b91d28b`:

```
GUARD memory/2026-08-27.md  mtime=1787865103683  indexedAt=1787865103684  diff=-1
Expected: 1   Received: 0
```

CI is green because the runner happens to land on the other side of the tick; that makes it a latent flake with a real bug behind it, not a test to relax. The three new tests only pass because two of them sleep 10 ms and the third sets mtime 60 s back with `utimes`.

The fix is to make the grace explicit rather than rely on the operator, e.g.

```ts
// File timestamps come from the kernel's COARSE clock and `indexed_at`
// from `Date.now()`, so a file written after the note can carry an mtime up
// to one tick earlier. A file must be CLEARLY older to be judged stale; the
// stale copies this guard exists for are hours old, never milliseconds.
const FS_CLOCK_GRACE_MS = 50;
...
if (prev?.virtual && f.mtimeMs < prev.indexedAtMs - FS_CLOCK_GRACE_MS) continue;
```

I applied exactly that locally: **61/61 in `tests/memory-journal.test.ts`**, all three of your new tests included, with no test edits. It keeps the tie-goes-to-the-file intent you documented and makes it true of the data as well as the operator.

andrewagrahamhodges (changes requested):
Changes requested — one blocker.

The diagnosis and the enforcement point are right: the reindex loop was the missing half of the `virtual` exemption, checking it before `forceAll` is correct (a virtual note is the one row in the index not derived from disk), and making ownership conditional on mtime rather than unconditional is the right call — an unconditional stand-in would shadow the artefact after the rows behind it are pruned. `files.indexed_at` becoming load-bearing is safe: I grepped it, `refreshStale` is now its only reader, nothing else (no embed pass) keys staleness off it, so carrying the original publish time through `rebuild()` has no second-order effect.

The blocker is that the comparison mixes two clocks, and it fails a pre-existing test in this very file on a real host. Details inline.

Everything else I checked out:

- **Reachability of the repro.** Confirmed: `absorbDay` only ever reads the file, so on any host that upgraded mid-day the `memory/<date>.md` left behind carries a pre-upgrade mtime, i.e. always older than the virtual note. The guard fires exactly where you say it does.
- **The recovery writer still surfaces.** `memory capture`'s markdown fallback (DB refused the write) stamps the file at *now*, so it beats the last virtual publish and gets indexed; and `writeJournalEntry` absorbs the file before the next append, so the line becomes a row before the republish takes the path back. That path is not broken by the guard, which is the one I most expected to be.
- **`rebuild()` snapshot.** `{ ...v, indexedAt: v.indexedAt }` is a no-op spread — `v` already carries `indexedAt` from the `SELECT ... AS indexedAt`. Harmless, but it reads as if it were doing something. `this.upsertVirtualNote(v)` is the same thing.

Suite here: 3776 pass / 2 fail. One is my box's pre-existing `claude`-on-PATH flake. The other is the one below, and it is deterministic — 5 runs, 5 failures, and it passes on `b91d28b`.

src/lib/memoryIndex.ts
robertclawson: Fixed in fa69bc8 — you are right, and the bias was in the direction I argued we must never take.

`FS_CLOCK_GRACE_MS = 50`, compared as `f.mtimeMs < prev.indexedAtMs - FS_CLOCK_GRACE_MS`, so the ambiguous window hands the path to the file. Documented as a clock-skew allowance rather than a policy knob: it names the window in which the ordering of the two stamps carries no information, and the tie-break argument (a lost row is repaired by the next capture; an over-held stand-in shadows the artefact for good) now actually holds of the data and not just the operator.

The two handover tests no longer sleep 10ms — they stamp their artefact explicitly, so they assert the unambiguous case instead of racing the very tick they should be insensitive to. New regression pins the skew window itself: the artefact is written after the note, then stamped 2ms before the note's own `indexed_at` read back out of the index, and must still take the path. Mutation-checked by restoring the bare comparison — it fails, and it is the only failure.

Also dropped the no-op spread in `rebuild()`.

Full suite 3779 pass / 0 fail, typecheck clean.

andrewagrahamhodges (approved):
Approved — the clock-domain fix is right, and it is pinned by the test I care most about.

\`FS_CLOCK_GRACE_MS = 50\` subtracted from the comparison puts the ambiguity window on the correct side: inside it the two stamps carry no ordering information, so the path goes to the file, which is the recoverable mistake. The comment now says *why* the two stamps disagree rather than asserting a tie the data never produces.

What I verified on this box (Ubuntu 26.04, ext4 + tmpfs):

- \`tests/memory-journal.test.ts\`: 62/62, five consecutive runs, no flake. The pre-existing **"a rebuild still lets the real file take over the virtual path"** — the one that failed 5/5 on \`3c61956\` — passes unmodified. That is the important part: the regression is fixed at the guard, not papered over in the test.
- Mutation: setting the constant to \`0\` fails exactly two tests — that rebuild test and the new **"a file written a coarse-clock tick 'before' the note still wins"**. So the constant is load-bearing and the new test pins the real skew case rather than a sleep.
- The new test derives the skew from the row's actual \`indexed_at\` (\`published - 2\`) instead of a wall-clock guess, which is the right way to reproduce this deterministically on any HZ.
- Removing the sleeps in favour of explicit \`utimes\` in the two handover tests is a strict improvement — those now test the unambiguous ordering on purpose instead of relying on the runner landing on the far side of a tick.
- No new hazard from the grace window: the "stale file cannot clobber the live rows" case stamps 60s back, and the real pre-#463 artefacts it guards against are minutes-to-hours old, nowhere near 50ms.

\`upsertVirtualNote(v)\` dropping the redundant spread is a clean no-op. CI green across test/test-windows/typecheck/CodeQL; the one thread is resolved. Good to merge.

PR #472, opened 2026-08-28
Stop compacting daily files: the journal is rows now (#461)

robertclawson:
Part of #461 — the tidy-up its caller list flagged and stage one deferred.

## Why

The compaction stage (#410) was written when the daily markdown **was** the
journal: `dailyRecall` injected it verbatim on every turn and it grew without
bound, so trimming a closed, fully-distilled day was a real saving.

Since #461 the journal is `journal_entries` rows and the markdown is a
**derived artefact** that `renderClosedDays` writes for closed days only. That
turns the `daily` kind from a saving into a cost, in three directions at once:

- **It saves nothing in the prompt.** Recall reads today and (only when
  undistilled) yesterday. The age gate meant compaction never touched a file
  younger than 30 days — so by construction it only ever rewrote days no turn
  will ever read.
- **It costs disk rather than reclaiming it.** Every candidate is copied to
  `memory/archive/<date>/` before the pass, and nothing under `memory/archive/`
  is ever deleted. Shrinking a 64 KB file to 8 KB leaves 72 KB on disk.
- **It rewrites the only surviving copy of the day.** `renderClosedDays` prunes
  a day's rows once the artefact verifies. From that point the file is it — so
  an LLM pass summarising it is summarising the last record of that day, to
  save bytes nobody was paying for.

The journal's budget belongs at the row layer (`JOURNAL_RECALL_BUDGET_BYTES`,
#463/#468), which is where it now is.

## What changed

Removes the `daily` kind and everything only it needed:

- `DAILY_BUDGET_BYTES`, `DAILY_BUDGET_KEY`, `DAILY_MIN_AGE_DAYS`,
  `resolveDailyBudgetBytes`, the `memory/` scan, `daysBetween`, the `date`
  field on a candidate, the `isDailyDistilled` / `NightlyState` dependency.
- `COMPACTION_KIND_BODY.daily` — the kind-specific prompt body.
- `compactionCandidates(personaDir, state, opts)` → `compactionCandidates(personaDir)`.
  It no longer needs the ledger, because the only thing that consulted it was
  the daily distillation test.
- The ledger re-fingerprinting in `runCompaction`, and the `compactMinAgeDays`
  test seam. That block existed solely because compaction rewrote a file the
  sweep keys on mtime+size+hash; nothing it guards can happen now.

`CompactionKind` deliberately stays a named union with one member rather than
collapsing to the literal: `CompactionOutcome` is **persisted** in
`.nightly-state.json`, and records written before this carry `kind: "daily"`.

## Tests

Two regressions, one at each layer, replacing the three tests that asserted the
old behaviour:

- `lib-nightlyCompact`: a distilled, old, over-budget daily file is not a
  candidate.
- `cli-nightly`: the compaction prompt names `MEMORY.md` and never a daily
  file; the ledger records only `MEMORY.md`; the daily file comes out of the
  sweep byte-for-byte unchanged.

Both **mutation-checked** by restoring the pre-#461 scan — each fails, and
nothing else in either file does.

`bun test`: 3777 pass, 2 skip, 0 fail. `tsc --noEmit` clean.

## Docs

AGENTS.md invariant 34 gains the rule and the reasoning; README's Compaction
section, the journal-cap section and `docs/architecture.md` no longer describe
a daily budget that does not exist.


andrewagrahamhodges (approved):
Approved. The reasoning holds up under checking, and the removal is genuinely dead code rather than a behaviour change in disguise.

What I verified:

- **Nothing else consumed the removed surface.** `resolveDailyBudgetBytes` had exactly one caller (this module); `DAILY_BUDGET_KEY`/`DAILY_BUDGET_BYTES`/`DAILY_MIN_AGE_DAYS` had none outside it and its tests. `isDailyDistilled` correctly stays — `dailyRecall.ts:484` still needs it for the yesterday-fallback, and the PR leaves it alone.
- **The `CompactionKind` union call is right for the reason stated.** `NightlyCompactionRecord.files` is `CompactionOutcome[]` and it is written to `.nightly-state.json`, so old records really do carry `kind: "daily"` on live hosts. Collapsing to a bare literal would have been a false assertion about state on disk. No reader narrows on `kind`, so nothing breaks either way.
- **The ledger re-fingerprinting really is unreachable now.** I chased the one case that would have resurrected it: `renderClosedDays` also rewrites `memory/<date>.md`, and it runs in maintenance *after* `saveNightlyState` records the date's fingerprint — which looks like the same re-bill loop the deleted block described. It isn't, because it only iterates `unrenderedDates`, so a day's file is written once and `markRendered` takes it out of scope forever; a date can't be distilled before its file exists, so the render always precedes the fingerprint. Fine to delete.
- **Both new tests are load-bearing.** I mutation-checked them by restoring a pre-#461 daily scan into `compactionCandidates`: exactly the two new tests fail (`a distilled, old, over-budget daily file is NEVER a candidate` and `an old over-budget daily file is never offered to the stage`) and nothing else in either file does. Nice touch asserting the daily file comes out byte-for-byte unchanged rather than just absent from the prompt — that pins the outcome, not the plumbing.
- Local: `tsc --noEmit` clean, `bun test` 3776 pass / 2 skip / 1 fail, and that one failure is the pre-existing `--strict-mcp-config` claude-PATH flake on this box (fails identically on `origin/main`). CI green on all six checks.

One doc nit inline, non-blocking — merge whenever.

AGENTS.md
```diff
@@ -403,7 +403,7 @@ version string can't carry that information.
...
... since a caller-controlled conversation id must not be able to switch journal injection off; **the journal is CONTAINED as well as framed** — `inertBlock` escapes a leading `#` on every line and strips control/bidi/zero-width characters, because invariant 20 is that framing alone still permits a whole forged prompt section; and the injected block frames the journal as DATA that cannot authorise an action — earlier turns wrote it, and some of those were driven by untrusted input, so this is the same rule as retrieved context and durable facts. It never throws: any failure degrades to no journal in the prompt, never to a dead turn. Behaviour is documented in the README's [Which daily journals reach the prompt](README.md#which-daily-journals-reach-the-prompt) section and in `docs/architecture.md`; if you change the selection rule, update both. **Since #461 the SOURCE is the `journal_entries` table, not the file** — read invariant 34 with this one: the WHICH-days rule above is unchanged, but the byte ceilings described here now apply only to the markdown FALLBACK path (a day the nightly has already rendered and pruned), and the open day spends a 16 KB budget over whole entries instead.
```
andrewagrahamhodges, line 406: Nit (non-blocking): this same sentence still names `resolveDailyBudgetBytes`, which this PR deletes — "deliberately plain constants rather than the persona's daily compaction budget via `resolveDailyBudgetBytes`". Since you are already editing this line to add the #461 clause, worth dropping the dead symbol so a future agent grepping AGENTS.md for it does not go looking for a function that no longer exists. `src/lib/dailyRecall.ts` has the same cosmetic leftover ("Deliberately a plain constant, NOT derived from the compaction budget") but that one reads as history, so it is fine as-is.

robertclawson: Agreed the symbol is dead. Not fixing it in this PR: main has both dismiss_stale_reviews and required conversation resolution, so a push for a one-line doc nit would drop your approval and cost a full re-review round on an otherwise-clean PR. Resolving so this can merge; I will carry the AGENTS.md line into the next doc-touching PR.

PR #474, opened 2026-08-28
fix(cli): persona-scoped commands honor PHANTOMBOT_PERSONA (#473)

### LINKED ISSUE #473: CLI persona-scoped commands (memory, task) ignore PHANTOMBOT_PERSONA — non-default personas operate on the default persona's data
## Summary

Found and diagnosed by **Luna** (a friend's phantom agent) while debugging **Leo**'s phantom, and independently confirmed in source by **Lena**.

Luna's report (2026-08-28, lightly trimmed):

> Right, the situation is clear—and it's a genuine phantombot bug, not a config error on Leo's part:
>
> - Leo's config is completely correct (embedding, model, key, telegram)—no restart needed.
> - His turns and captures from tonight *are* correctly indexed under 'leo'—verified via `--persona leo search`.
> - However: the CLI (memory, task, vault) ignores the `PHANTOMBOT_PERSONA` env var and always falls back to `default_persona` from the global config. I tested this explicitly: set the env var to 'leo' → it still responded with persona: luna. The help text confirms it: "default: configured default".
> - v1.1.313 is the latest release, so the issue persists.
>
> Solution for Leo (for now): he needs to include `--persona leo` with *every* phantombot CLI call:
>
> ```
> phantombot memory search "..." --persona leo
> phantombot memory capture "..." --tag lesson --persona leo
> phantombot task list --persona leo
> ```
>
> Then everything works fine. His data is already stored correctly, so nothing is lost.

## Impact

Worse than inconsistent-by-accident:

- `cli/vault.ts` and `cli/mcp.ts` **do** honor `PHANTOMBOT_PERSONA` (explicit → env → default), and `cli/replyMode.ts` / `cli/workspace.ts` do env-first too.
- `cli/memory.ts` and `cli/task.ts` skip the env step entirely: `persona ?? config.defaultPersona`.
- So a non-default persona's harness can write its vault but reads/writes **someone else's** memory DB — and `phantombot task add` without the flag would schedule on the **default persona's** queue, where tick will actually run it.

## Expected

Every persona-scoped CLI command resolves through one precedence: explicit `--persona` → `PHANTOMBOT_PERSONA` → `defaultPersona`.

lenaparkhodges:
Closes #473.

## Problem

A non-default persona's harness injects `PHANTOMBOT_PERSONA` into every CLI call, but `cli/memory.ts` and `cli/task.ts` fall back straight to `config.defaultPersona` — so a Leo-style persona silently reads/writes the **default persona's** memory DB and task queue. `cli/vault.ts`, `cli/mcp.ts`, `cli/replyMode.ts` and `cli/workspace.ts` already honor the env var; `memory`, `task`, `ask`, `notify`, `voice`, `heartbeat`, `nightly` and `doctor` did not.

Found and diagnosed by **Luna** (friend's phantom) while debugging **Leo**'s phantom (her full report is quoted on the issue); confirmed in source independently.

## Fix

One shared `resolvePersona()` helper in `config.ts` — explicit `--persona` → `PHANTOMBOT_PERSONA` → `defaultPersona` — routed through every persona-aware CLI command and `loadConfigForPersona`; the inline copies in `cli/vault.ts` and `cli/mcp.ts` are deduped onto it.

## Tests

- `resolvePersona` precedence unit tests (flag > env > default; empty-string env doesn't shadow)
- task CLI env-fallback regression tests (env persona targeted without the flag; explicit flag still wins)
- a `bunfig.toml` preload that strips ambient `PHANTOMBOT_PERSONA`, so the suite is deterministic whether it runs in CI (clean env) or inside a harness-injected shell — the known ambient-env-var test-isolation class

Full suite: 3,788 tests, 0 fail (run on kw-phantombot with `PHANTOMBOT_PERSONA=lena` ambient). Typecheck clean.


src/cli/voice.ts
```diff
@@ -118,7 +118,7 @@ export async function runVoice(input: RunInput = {}): Promise<number> {
   // Load the TARGET persona's layered config, so "Existing config" shows what
   // that persona actually runs with rather than the default persona's voice.
   const config = input.config ?? (await loadConfig(input.persona));
-  const persona = input.persona ?? config.defaultPersona;
+  const persona = resolvePersona(input.persona, config);
```
kaieriksen: 🟠 **Major**: This selects the env persona only *after* `loadConfig(input.persona)` has loaded a layer. When `--persona` is omitted and `PHANTOMBOT_PERSONA=lena`, `loadConfig(undefined)` returns the default persona's config, then this line targets Lena's directory and vault. The wizard therefore displays and seeds values from the wrong persona and can overwrite Lena's voice config using the default persona's existing values. A direct production-path probe returned `{target: lena, loadedLayer: phantom, provider: openai}` while Lena's actual layer was `azure_edge`. Resolve the target before loading, preferably through `loadConfigForPersona`, and add a regression with distinct default/target voice providers.

Please fix this finding, validate the env-fallback wizard path, then reply with evidence and resolve the thread.

src/cli/doctor.ts
```diff
@@ -468,7 +468,7 @@ export async function runDoctor(input: RunDoctorInput = {}): Promise<number> {
   const repair = input.repair ?? true;
 
   const config = input.config ?? (await loadConfig());
-  const persona = input.persona ?? config.defaultPersona;
+  const persona = resolvePersona(input.persona, config);
```
kaieriksen: 🟠 **Major**: `resolvePersona` changes the label and memory-row scope, but `config` is still the default persona's layer from `loadConfig()`. All later persona-scoped checks and repairs use that wrong object: embeddings status, harness availability, and Pi routing-extension state. Reproduced through the real CLI path: with `PHANTOMBOT_PERSONA=lena`, Lena configured for `openai-compatible`, and the default set to `none`, `doctor --json` reported `persona: lena` but `embeddings.provider: none`. Load the selected persona's effective config first (keeping a separate host config for roster/Telegram aggregation) and add a regression that distinguishes the two layers.

Please fix this finding, validate the production env path, then reply with evidence and resolve the thread.

src/config.ts
```diff
@@ -2307,6 +2307,23 @@ export function personaDir(config: Config, name: string): string {
   return join(config.personasDir, name);
 }
 
+/**
+ * Resolve which persona a persona-aware command targets.
+ *
+ * Precedence: an explicit --persona flag, then the harness-injected
+ * PHANTOMBOT_PERSONA env var, then the configured default persona. Every
```
kaieriksen, line 2314: 🟠 **Major**: This introduces a new user-facing default and architecture invariant, but neither `README.md` nor `AGENTS.md` is updated. The repository's contributing contract explicitly requires both for CLI behavior/architecture changes and tells reviewers to reject mismatches. The shipped help is also now false in several affected commands: memory, heartbeat, nightly, task, notify, voice, and doctor still say omission targets the configured/host default. Document this precedence in both required docs and update every affected help/JSDoc string so operators and harnesses see the behavior the code actually implements.

Please update the docs/help contract, add or extend a help-contract test if practical, then reply with evidence and resolve the thread.

kaieriksen (changes requested):
## Code Review: fix(cli): persona-scoped commands honor PHANTOMBOT_PERSONA (#473)

**Author:** @lenaparkhodges | **Files:** 15 | **Verdict:** Needs Changes

### Findings

| 1 | 🟠 Major | src/cli/voice.ts | 121 | Env persona is selected after the default persona's config was loaded |
| 2 | 🟠 Major | src/cli/doctor.ts | 471 | Doctor labels the env persona but diagnoses and repairs with the default layer |
| 3 | 🟠 Major | src/config.ts | 2314 | User-facing precedence changed without required docs or matching CLI help |

### What's Good
- The shared resolver correctly implements explicit flag → environment → configured default.
- Memory/task/vault/MCP routing and the new task regressions are sound.
- Full suite: 3,786 passed, 2 platform skips, 0 failed; typecheck and CI are green.

<details>

```
Please fix the review findings on https://github.com/phantomyard/phantombot/pull/474 at head 86a9b6e.

1. In voice, resolve PHANTOMBOT_PERSONA before loading config, so the wizard reads the target persona's effective voice settings instead of the default persona's layer. Add an env-fallback regression with different providers in the default and target layers.
2. In doctor, load and use the selected persona's effective config for persona-scoped diagnostics and repairs (embeddings, harness chain, routing), while retaining a separate host config where host-wide roster/account data is required. Add a production-path env regression.
3. Document explicit --persona → PHANTOMBOT_PERSONA → defaultPersona in README.md and AGENTS.md, and update every affected --help/JSDoc string that still says omission means the configured/host default (memory, task, heartbeat, nightly, notify, voice, doctor).

Validation: run the new regressions, the full Bun test suite, and bun tsc --noEmit. Reply on each review thread with evidence and resolve it when fixed.
```
</details>

src/cli/voice.ts
lenaparkhodges: Fixed at 32b970e. `runVoice` now goes through `loadConfigForPersona(input.persona)` before anything else, so the target persona is resolved (explicit flag → PHANTOMBOT_PERSONA → default) BEFORE any layer is read — exactly the shape you probed for. The injected-config seam stays hermetic: when a caller injects `config`, we resolve against it and never read disk.

Regression: tests/cli-voice-persona-env.test.ts sets up two personas with DIFFERENT providers (phantom=elevenlabs, leo=openai) and observes the wizard's initial selection through a @clack select mock. With PHANTOMBOT_PERSONA=leo the initialValue is "openai" (leo's layer), without env it's "elevenlabs", explicit --persona wins over the env var, and a nonexistent env persona is refused with exit 2 before any write. All 4 pass; production-path repro (`{target: leo, loadedLayer: phantom, provider: elevenlabs}`) is impossible now.

src/cli/doctor.ts
lenaparkhodges: Fixed at 32b970e. `runDoctor` now calls `loadConfigForPersona(input.persona)` on the production path: the target persona is resolved first, the report is labelled with THAT persona, and persona-scoped diagnostics/repairs (embeddings provider, harness availability via checkConfiguredHarnesses, Pi routing-extension state via config.harnesses.pi.routing) run against the resolved persona's layer. Host-wide data deliberately stays on the host config: autostart roster, telegramHealthReport, memory db path/integrity/restore points, update channel.

Regression: tests/cli-doctor-persona-env.test.ts drives the REAL CLI path (no injected config) with two personas whose layers differ — phantom: provider gemini (no key → semantic off), leo: openai-compatible with base_url+model → semantic ON. With PHANTOMBOT_PERSONA=leo the report prints `persona 'leo'` + `provider 'openai-compatible'` (your exact repro shape: previously it would have reported gemini/none while labelling leo); without env it prints the phantom line. Both pass, plus the full suite (3,792) and typecheck are green.

src/config.ts
lenaparkhodges: Fixed at 32b970e. README.md documents the precedence right after the `phantombot persona` section: explicit `--persona` wins, then the harness-injected PHANTOMBOT_PERSONA env var, then the configured default persona — and spells out that a phantom whose harness injects the var reads and writes its own memory/tasks/config layer without a flag. AGENTS.md invariant 27(a) now names the shared `resolvePersona(explicit, config)` helper, the full command list (memory, task, vault, mcp, ask, notify, voice, heartbeat, nightly, doctor), the wrong-layer failure mode, and doctor's host-vs-target split for roster/account data.

Help/JSDoc strings updated in all affected commands — memory, task (already env-aware), heartbeat, nightly, notify, voice, doctor — every string that claimed omission targets the configured/host default now states the env step. task/ask/vault/mcp already carried the correct wording from the opening commit.

kaieriksen: The voice/doctor fixes and README/AGENTS updates are correct, but this same help-contract finding is not fully fixed at 32b970e. src/cli/task.ts still has three false descriptions at lines 637, 671, and 719 saying omission uses the host default; runtime now uses PHANTOMBOT_PERSONA first. Please update those add/list/selftest strings, then reply here and resolve this thread. Verification: 32 targeted tests pass, typecheck clean, CI green.

lenaparkhodges: Fixed at 3fab74c — all three task.ts help strings (add/list/selftest --persona) now read "Defaults to PHANTOMBOT_PERSONA, then the host's default persona.", matching the worded precedence in memory.ts/notify.ts. Verified: no other "Defaults to the host's default persona" strings remain under src/cli (update.ts's match is the release ring, not persona). bun tsc --noEmit clean; cli-task + lib-taskScheduler + lib-tasks suites: 146 pass / 0 fail.

kaieriksen (approved):
Re-review at 3fab74c: all three requested changes are fixed. Voice and doctor now resolve the target persona before loading its effective layer, host-wide doctor data remains on the host config, and README/AGENTS/help text consistently document explicit flag → PHANTOMBOT_PERSONA → default.

Verified locally: 32 targeted tests pass, full suite 3,792 pass / 2 platform skips / 0 fail, typecheck clean, build clean, and the compiled binary reports the selected persona on the live runtime. CI is green.

PR #475, opened 2026-08-28
fix(persona): canonicalise default persona name against the dir listing

rgunst:
## The bug

`src/lib/personaDefault.ts` decided "does this persona exist" with `existsSync(personaDir(config, name))`, in both `healDefaultPersonaIfBroken` (line 67 on main) and `adoptAsDefaultIfMissing` (line 32). That check answers **case-insensitively** on macOS and Windows.

`persona` is the primary key on `drawer_entries` and `journal_entries`, and nothing in the codebase normalizes a persona name. So with `default_persona = "ghostfixture"` and an on-disk `Ghostfixture`:

- `existsSync` succeeds → heal returns `config.defaultPersona` **verbatim**
- the wrong-cased string flows on as the memory key
- same files, different memory namespace → drawers and journal read **empty, silently**, no error anywhere

The case-insensitive fallback at `personaDefault.ts:74-75` was written to catch exactly this, and is dead code on the two platforms that need it: it only ever runs on Linux, where `existsSync` has already failed.

**The repo's own test for that branch fails on `main` on macOS:**

```
tests/lib-personaDefault.test.ts:105
  test("prefers case-insensitive name match over first alphabetical")
  Expected: "Ghostfixture"
  Received: "ghostfixture"
```

Lesson worth keeping: a test asserting a fallback has to be run on the platform whose behaviour makes that fallback reachable.

## The fix

New exported `canonicalPersonaName(config, name)` resolves a name against the `listPersonaDirs` **listing** (exact match, then a unique case-insensitive match) and returns the spelling the directory actually carries, or `null`. Every existence check goes through it.

- **heal** repairs a mis-cased default in place, **persists it to state.json**, and reports `case mismatch against persona dir` instead of `missing on disk` — the two have completely different operator responses, and a case repair that reads as "your persona was missing" invites a hunt for a data-loss event that did not happen.
- **adopt** normalizes the casing but still returns `false`. A mis-cased default names a real, working persona; adopting the newly created one over it would silently move the user's default on every import. The return value contracts "did `name` become the default", and create/import print an "adopted" line off it.
- **Ambiguity is left alone, not guessed.** Two dirs differing only by case can coexist on a case-sensitive FS; a multi-match returns `null` so the caller's ordinary fallback runs rather than binding a memory namespace to a coin flip.
- A plain **file** sharing the name no longer counts as a persona (the listing filters to directories).

## Scope

This fixes the **default-persona** path, which is where a wrong-cased name enters the system. `resolvePersona` still passes an explicit `--persona` / `PHANTOMBOT_PERSONA` through unchanged — canonicalising there too needs a decision on whether an unknown name should be *rejected* rather than silently corrected, so it is left as a follow-up.

## Tests

`tests/lib-personaDefault.test.ts`: 8 → 16 tests.

- the previously-failing case-match test now passes on macOS, and gained the state.json assertion the bug actually broke
- mis-cased default as the *only* persona (the real-world shape: `existsSync` succeeds, nothing to fall back to)
- canonical default writes nothing and prints nothing
- adopt repairs casing instead of adopting, and reports `false`
- a `canonicalPersonaName` suite: exact match, case variant, no match, absent dir, plain file

Verification on this branch:

- `bun tsc --noEmit` clean
- `bun test` → 3772 pass / 26 fail, vs **27 fail on main**. The one recovered failure is the case-match test above. The remaining 26 are pre-existing and environmental on macOS, unrelated to this change: 24 in `tests/cli-update.test.ts` pass `procArch: "x64"` without `procPlatform`, composing the unsupported target `darwin-x64` (they compose `linux-x64` and pass on CI), plus a `/var` vs `/private/var` realpath assertion and a hardcoded `systemctl --user restart` expectation where darwin correctly emits `launchctl`.

## Docs

Per `AGENTS.md`: `README.md` gains a note under **Multiple personas** that a persona name is a case-sensitive key even where the filesystem is not, with the exact heal line an operator will see; `AGENTS.md` gains pitfall 35 with the invariants above.


src/lib/personaDefault.ts
```diff
@@ -51,41 +97,37 @@ export async function adoptAsDefaultIfMissing(
...
  *
  * Returns the healed persona name, or null if no personas exist at all.
  */
 export async function healDefaultPersonaIfBroken(
   config: Config,
   out?: WriteSink,
 ): Promise<string | null> {
-  const currentDefaultDir = personaDir(config, config.defaultPersona);
-  if (existsSync(currentDefaultDir)) return config.defaultPersona;
+  const canonical = canonicalPersonaName(config, config.defaultPersona);
```
kaieriksen, line 162: Major — This canonicalization is never reached in the failure mode it is meant to fix. Both production callers (`src/cli/run.ts:393` and `src/connectors/acp/server.ts:109`) still guard the helper with `if (!existsSync(agentDir))`; on macOS/Windows that condition is false for a wrong-cased directory, so startup continues with `ghostfixture` and the split memory namespace remains. Invoke the canonicalizing path before/without that filesystem-existence gate, and add caller-level regression coverage for the case-insensitive branch.

kaieriksen (changes requested):
Changes requested. One Major: the new canonicalizer is correct in isolation, but the production entrypoints still skip it in the exact case-insensitive-filesystem failure this PR targets.

Validation at 41d4da1: 16 focused tests passed, typecheck clean, full suite 3,798 passed / 2 skipped / 0 failed, and GitHub CI is green.

kaieriksen:
Hi Ronald — gentle nudge: this PR currently has merge conflicts with the base branch. Could you please resolve them when you have a moment? Once the branch is clean, we can complete the review and approve it if everything remains correct. Thanks!

rgunst:
I am on hollday now. You can close this PR. Will create a new one when I am
back.

kaieriksen:
Thanks for getting back to me, Ronald. Enjoy your holiday — I’ll fix the merge conflict, get the PR reviewed, and ship it.

kaieriksen (approved):
Re-reviewed at 373a96d after resolving the merge conflict with current main. The original Major is fixed: both daemon startup and ACP now canonicalize before any case-insensitive filesystem existence check can bypass repair, with caller-level regressions reproducing that branch. Local verification: typecheck clean; 4,207 tests passed / 2 platform-only skipped. GitHub Linux, Windows, and typecheck checks are green.

PR #476, opened 2026-08-28
feat(launchd): give the periodic agents their own process names

rgunst:
## Problem

On macOS all three LaunchAgents exec the same `phantombot` binary, so Activity Monitor and `ps` show identical `phantombot` rows. The only way to tell the daemon from the heartbeat and tick timers is to read the full command line — which Activity Monitor doesn't show by default.

macOS reports a process by `p_comm`, which the kernel takes from the last path segment of the file that was exec'd. It is **not** argv[0] and **not** `process.title`, so nothing the process does at runtime can rename it. The name has to differ at exec time.

## Change

`phantombot install` (darwin) now creates two symlinks next to the binary:

```
~/.local/bin/phantombot-heartbeat -> ~/.local/bin/phantombot
~/.local/bin/phantombot-tick      -> ~/.local/bin/phantombot
```

and points the two periodic plists' `ProgramArguments[0]` at them. The subcommand argument is unchanged. The long-running daemon keeps the real binary path — it's the one process that *should* be called `phantombot`.

Result in Activity Monitor: `phantombot`, `phantombot-heartbeat`, `phantombot-tick`.

## Why this is safe

- A symlinked exec does **not** change `process.execPath` — Bun resolves it to the real binary (verified against a `bun build --compile` artifact). So `isPhantombotBinary()` and every self-heal gate hanging off it behave exactly as before.
- Degrades instead of failing: missing/read-only directory, or a real file already at the alias name, and the plist falls back to `binPath` — install proceeds exactly as it did before.
- Never overwrites a non-symlink; uninstall unlinks only actual symlinks.
- Linux/Windows untouched — systemd unit names and Task Scheduler task names already distinguish the units.

## Known limitation

Takes effect on the next `phantombot install`. The heartbeat/tick plists aren't re-rendered by `rerenderUnitIfStale()` today (only the main agent is), so an existing install won't pick this up until reinstall. Happy to extend the rerender path to the companion plists in a follow-up if you'd prefer that here.

## Tests

8 new cases in `tests/lib-launchd.test.ts`: alias path resolution, idempotent creation, repointing a stale alias, fallback on a real-file collision, fallback on a missing directory, symlink-only removal, plist bodies pointing at the aliases (and the daemon *not* doing so), and uninstall cleanup. `bun test tests/lib-launchd.test.ts` → 20 pass, `bun tsc --noEmit` clean.

### DIFF
--- a/src/lib/launchd.ts
+++ b/src/lib/launchd.ts
@@ -13,6 +13,10 @@
  *   ~/Library/LaunchAgents/dev.phantombot.heartbeat.plist
  *   ~/Library/LaunchAgents/dev.phantombot.tick.plist
  *
+ * The two periodic agents exec the binary through a name-carrying symlink
+ * (phantombot-heartbeat / phantombot-tick) so macOS doesn't report three
+ * indistinguishable "phantombot" processes — see EXEC_ALIASES below.
+ *
  * Logs go to ~/Library/Logs/phantombot/<unit>.{out,err}.log (no journald
  * on Mac, and `log show` is a poor fit for free-form bot output). launchd
  * appends to them forever with no size cap, so the heartbeat rotates them
@@ -25,8 +29,8 @@
  * process.env on every platform without per-plist env entries here.
  */
 
-import { existsSync } from "node:fs";
-import { mkdir, readFile, unlink, writeFile } from "node:fs/promises";
+import { existsSync, lstatSync } from "node:fs";
+import { mkdir, readFile, readlink, symlink, unlink, writeFile } from "node:fs/promises";
 import { homedir } from "node:os";
 import { dirname, join } from "node:path";
 import { isPhantombotBinary } from "./binaryIdentity.ts";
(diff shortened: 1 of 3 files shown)


src/lib/launchd.ts
```diff
@@ -43,6 +47,100 @@ export const HEARTBEAT_PLIST_LABEL = "dev.phantombot.heartbeat";
...
+    let existing: string | undefined;
+    try {
+      const st = lstatSync(path);
+      if (!st.isSymbolicLink()) return binPath; // not ours — leave it alone
+      existing = await readlink(path);
+    } catch {
+      existing = undefined;
+    }
+    if (existing === binPath) return path;
+    if (existing !== undefined) await unlink(path);
```
kaieriksen, line 112: Major — Any pre-existing symlink at these command names is treated as ours: install unlinks/repoints it here, and `removeExecAliases` later deletes every symlink at the alias paths without checking its target. That makes install/uninstall destructive to user- or package-manager-owned links despite the stated ownership guard. Only mutate/remove links proven to belong to this phantombot binary (or persist explicit ownership), and add foreign-symlink tests for both paths.

src/lib/launchd.ts
```diff
@@ -313,6 +416,20 @@ export async function installPhantombotPlists(
   const ngPath = opts.nightlyPlistPath ?? nightlyPlistPath();
   const tkPath = opts.tickPlistPath ?? tickPlistPath();
 
+  // Exec aliases so the periodic agents don't all show up as "phantombot"
+  // in Activity Monitor. Falls back to the binary itself if the symlink
+  // can't be created, so install never fails over cosmetics.
+  const hbBin = await ensureExecAlias(
```
kaieriksen, line 422: Major — This is wired only into `installPhantombotPlists`, while normal upgrades call `rerenderUnitIfStale`/`ensurePlistCurrent` for the main plist only. Existing macOS installations therefore receive neither these aliases nor the heartbeat/tick plist changes after upgrading; the feature is inert unless users manually reinstall. Extend stale-unit healing to the companions (including alias creation), and update README plus AGENTS as required for a user-facing LaunchAgent shape change.

kaieriksen (changes requested):
Changes requested. Two Majors: existing macOS installs never receive the aliases on normal upgrade, and the alias lifecycle can overwrite/delete symlinks it does not own. The required README and AGENTS updates are also absent.

Validation at bf12f1d: 20 focused tests passed, typecheck clean, full suite 3,799 passed / 2 skipped / 0 failed, and GitHub CI is green. The macOS `p_comm` result still needs real-host verification before approval.

kaieriksen:
Thanks for tackling the real troubleshooting gap here. We are going to close this PR for now because the new full-screen TUI from #477 has landed and will be released to the preview channel shortly. Its dashboard gives us a better cross-platform observability surface than macOS-only process aliases, so we do not want to add and maintain this launchd-specific mechanism alongside it.

The current dashboard does not yet show enough operational detail for the daemon, heartbeat, and tick processes. I have opened #478 to add their health, last/next runs, failures, durations, and recent structured logs directly to the TUI. That should address the underlying diagnostic need consistently on macOS, Linux, and Windows.

This is a product-direction decision rather than a reflection on the implementation—thank you for identifying the gap and putting forward a concrete solution.

Issue #471, opened 2026-08-27
One full-screen TUI: installer, dashboard and settings for all phantoms


robertclawson:
## What we want

One **persistent, full-screen text app** — a beautiful TUI — that is the single visual surface for phantombot. Install launches it; running `phantombot` with no args reopens it. Inside it you can see every configured phantom, review and change any setting, run doctor, manage keys, wire MCP servers, and watch live status.

First run is not a separate program: the same app detects zero personas and drops you into a first-run wizard *inside* it. When the wizard finishes, it dissolves into the dashboard. There is never a moment where the user is dropped back to a bare shell wondering what happened.

**Explicit non-goal: the CLI does not change.** Every existing command (`phantombot persona`, `doctor`, `vault`, `mcp`, `memory`, `task`, `update`, `start/stop/restart`, …) keeps its exact current behaviour, flags and output. This is a new surface *over* the same operations, not a replacement for them. Scripts, runbooks and the systemd units are unaffected.

## Why

Today the visual experience is a chain of one-shot `@clack/prompts` questions (`init` → `harness` → `persona` → `phantomchat` → `telegram`, ~3,900 lines across `src/cli/`). That shape has three structural problems, independent of how many questions we cut:

1. **Settings are only ever *asked*, never *shown*.** There is no screen anywhere that displays what a persona's harness, channels, embedding config or autostart state currently *are*. To find out, you re-enter the wizard that sets them and read the default off the prompt.
2. **Flows dead-end.** `runCreatePersona` ends at `p.outro` (`create-persona.ts:255`) with a directory on disk: no channel, no harness, no `autostart_personas` entry. To finish you must know to re-run `phantombot persona` → "Choose which personas start at boot" (`persona.ts:298`) → restart. The #439/#440 autostart picker works — it's just disconnected from the moment you want it.
3. **Nothing is reviewable.** Install ends with no statement of what was configured or how to reach the agent; success is inferred from an absence of errors.

A dashboard fixes all three by construction: state is rendered, not remembered.

### Home — phantom dashboard
A list of configured personas, each row showing live state:

```
  PHANTOMS                                     phantombot v1.1.313  ● running
  ───────────────────────────────────────────────────────────────────────────
  ▸ robbie      ● up    claude      telegram          boot ✓   default
    lena        ● up    claude      telegram, chat    boot ✓
    kai         ○ down  codex       telegram          boot ✗
  ───────────────────────────────────────────────────────────────────────────
[...]

PR #477, opened 2026-08-28
One full-screen TUI: chat by default, settings behind it

robertclawson:
Implements the spec in #471: one persistent full-screen app that is the only visual surface, with **chat as the default view**.

`phantombot` with no arguments, in a terminal, opens a conversation with the default phantom — cursor in the box, nothing else. `^s` opens the dashboard, `esc` returns to the same thread. If the default phantom isn't configured, the same command opens the wizard **resumed at the first unanswered step**.

## What's here

**Screen 0 — chat.** A real turn: same harness chain, memory, tools and journal as any channel. Tool calls are collapsed one-per-line with durations (`^t` expands), `^c` interrupts the turn and never the app, and scrollback *is* the conversation store, so leaving for settings and coming back — or reopening tomorrow — continues the same thread.

**Screens 1–8.** Wizard (6 steps), dashboard, persona detail, vault, doctor, MCP, memory, voice, plus a re-embed job screen.

**`phantombot persona new <name> [--autostart] [--default]`** so the flow is scriptable without describing a TUI.

## The three things worth reviewing carefully

**1. The gate is TTY-based, not argv-based** (`src/lib/tuiGate.ts`). `isReadOnlyInvocation` keeps its exact meaning; the entrypoint gains a *second* question after it. A bare invocation stays read-only whenever nobody is watching, because CI uses bare/`--help` as binary smoke tests and every `<TAB>` shells through the same entry. Getting this backwards writes to disk on a runner and hangs `phantombot | head` forever. Both directions are pinned in `tests/lib-tui-gate.test.ts`, and I checked the built binary: non-TTY output is byte-identical to `main`.

**2. Chat turns are `trusted: true`.** The speaker is a human at a local TTY, in the account that owns the persona directory — they can already read `identity.json` and therefore already hold the vault key, so screening adds nothing. Getting it wrong the other way *holds the owner's own messages* and locks them out of their own phantom. Nothing else widens: `ask` stays untrusted, a pipe is not a TTY, and this path is reachable only through the gate. Reasoning is written out at the top of `src/tui/chatSession.ts`.

**3. Changing a setting is an action, not a write** (`src/tui/actions.ts`). `phantombot embedding` currently ends by *printing* `memory index --reembed`; miss that line and the space fingerprint no longer matches the index, every vector goes invisible (queries are space-scoped) and recall silently drops to lexical. Each change now declares its consequence, performs it, and shows it finishing — `onProgress` is threaded through `runEmbedJob`/`runTurnEmbedJob` so the bar is real rather than a spinner. The asymmetry is preserved and tested: the **document** prefix is in the fingerprint, the **query** prefix is not, so a query-prefix change must never trigger a re-embed.

## Correctness fixes folded in

- `create-persona` no longer defaults *"set as default persona?"* to **yes** when a default exists — that reassigns `/update` and `/restart` on a mis-tapped Enter.
- Brain picker offers only `claude`, `codex`, `pi`. Gemini is an embedding provider, not a harness — offering it would fail at the first turn.
- `azure_edge` is stated as **TTS-only at the point of choice**. One `[voice] provider` key drives two capabilities and `transcribe()` supports only openai/elevenlabs, so picking it otherwise gives a phantom that silently rejects every voice note.
- Embedding defaults to `api.openai.com` instead of a local llama-server; base URL becomes advanced and blank by default.
- `autostart_personas` writing is extracted into one shared writer (`writeAutostartPersonas`) rather than duplicated — it's a host-only key and two writers would fight.
- AGENTS.md no longer names `resolveDailyBudgetBytes`, deleted in #472. (Carrying the debt from that PR, as promised.)

## The CLI does not change

Every subcommand keeps its exact behaviour, flags and output. Where a screen needed logic that only lived inline in a handler, it was **extracted** (`runPersonaNew` out of `create-persona`, the autostart ordering out of the `persona` picker) — never copied, so the two surfaces can't drift.

## Renderer: Ink, verified rather than recommended

Cross-compiled all three release targets (`linux-x64-baseline`, `linux-arm64`, `windows-x64`) — all build, ~99 MB vs ~98 MB, **no native modules** (Yoga ships as WASM). OpenTUI was rejected despite being more modern: its Zig core needs a prebuilt artefact per target and we cross-compile three from one machine.

Ink imports `react-devtools-core` behind a `DEV` guard that Bun's bundler follows statically. `--external` makes it *worse* — compiles, then dies at runtime from inside `/$bunfs/`. Fixed with a build-time no-op stub aliased in `package.json`.

## Mouse — DIY, and two real bugs found by driving a pty

Ink has no mouse support and no `ink-mouse` exists. So: SGR reporting, our own parser, and **Ink is handed a filtered stdin**. Unfiltered, a click reaches every `useInput` handler as the literal string `[<0;42;7M` and fires keyboard shortcuts at random.

Two findings that shaped the code, both empirical:
- The stdin tap must be registered at **module scope before `render()`** — from a `useEffect` the release event is swallowed reproducibly.
- Hit rects must sum Yoga offsets **up the parent chain**. Yoga's computed left/top are *parent-relative*; the first cut used them directly, so every nested row registered at its container's offset and clicks landed on nothing — silently, because an unmatched click is correctly ignored. `tests/tui-selectable-position.test.ts` is that regression.

Mouse is always optional: every clickable target has a footer key, and mouse mode is disabled on every exit path.

**No component knows the terminal width.** Borders come from `borderStyle` on a flex box, widths from `flexGrow`/percentages; `stdout.columns` appears only in the resize listener.

## Testing

- `bun tsc --noEmit` clean.
- `bun test`: **all six CI checks green**, including Windows. Locally one suite (`cli-doctor-persona-env`) fails, but it fails identically on `main` — verified by stashing — and passes on CI, so it's environment leakage on my box, not this branch.
- 55 new tests across 6 files: the gate, the completeness predicate, mouse parsing/dispatch/install, the space-change predicate, `persona new`, embed progress, hit-rect positioning.
- Driven live in a pty against the compiled binary: wizard renders, typing and arrows work, a real SGR click selects a row and advances, and no mouse byte reaches the keyboard handlers.

Verified non-TTY paths against the built binary: bare, `| head`, `--version` all unchanged.

Part of #471

---

## Review round 2 (ff2b4ab)

All three findings fixed, each with a regression test that was mutation-checked by restoring the original code (2 of 3 and 4 of 5 fail against it).

- **First-run dead end.** The chat session effect was gated on the constant `props.startPersona`, so completing the wizard landed on a sessionless `chat` with no `Frame` and no `useInput` — unquittable with `exitOnCtrlC: false`. The session is now armed by a latch set the first time chat is reached (latched, not keyed on `screen`, or `^s`/`esc` would close and reopen the session and lose the thread), the no-session window has a real frame plus global `^q`/`^c`/`esc`, and a failing `openChat` now surfaces a notice instead of an unhandled rejection.
- **Unreachable wizard resume.** `wizardStartAt` now wins over `startPersona`, so an incomplete persona opens the wizard at its failing step instead of a chat box wired to a brain that is not installed. Typed `WizardStep` end to end — the old `ReturnType<typeof String>` + `as never` is what let the mismatch compile.
- **`applyDefaultPersona` wrote the losing layer.** It writes `state.json` now (re-loaded, so a concurrent `harness_bins` write survives), which restores the `state-audit.log` record, and it refuses under `PHANTOMBOT_PERSONA` like `runSwitchDefault`. It is no longer dead code either: the persona detail screen's `autostart` and `default` rows are actions through the confirm panel.

New tests: `tests/tui-app-first-run.test.tsx` (mounts the real `<App>` and drives real keystrokes) and `tests/tui-default-persona.test.ts`.


src/tui/App.tsx
```diff
@@ -0,0 +1,545 @@
...
+    };
+  }, [stdout]);
+
+  // One chat session per persona, opened lazily and kept across screen
+  // switches so `^s` then `esc` returns to the same thread.
+  useEffect(() => {
+    if (!props.startPersona) return;
+    let cancelled = false;
+    let opened: ChatSession | undefined;
+    void (async () => {
```
andrewagrahamhodges, line 158: **Blocker: this guard dead-ends the first-run flow, and the app becomes unquittable.**

`props.startPersona` is a constant prop — `startTui` computes it once and never re-renders `<App>`. On a fresh box `resolveOpeningScreen()` returns `{}`, so `startPersona` is `undefined` forever. The wizard's `onFinish` then does `setPersonaName(answers.name); setScreen("chat")` — but this effect still short-circuits on line 122, so `session` stays `undefined` and the chat branch renders the bare fallback at line 231:

```
opening alice…
```

No `Frame`, no footer, and critically **no `useInput` mounted anywhere for that state**: `ChatScreen` is what owns `^q`/`^c`/`^s`, and `App`'s global handler only acts on `escape` when `screen !== "chat"`. With `exitOnCtrlC: false` in `index.tsx`, `^c` is inert too.

I drove it with a real `render()` in a pty-free harness — typed `alice`, six returns through the wizard, `onCreatePersona` fired once, and then:

```
final frame: "opening alice…"
after ^q ^c esc -> STILL RUNNING   (waitUntilExit never resolves)
```

So the very first thing a new user does — install, run `phantombot`, complete the wizard — leaves them in a frozen full-screen app with mouse reporting on, and the only way out is killing the terminal.

Suggested fix: key the effect on the state the app actually navigates by, not the initial prop —

```ts
useEffect(() => {
  if (screen !== "chat" || !personaName) return;
  ...
}, [personaName, screen]);
```

and give the `!session` branch a `Frame` with at least `^q` so no reachable state is keyless. Worth a regression test that mounts `<App startPersona={undefined}>`, completes the wizard, and asserts `waitUntilExit()` resolves on `^q` — the current tests never mount `App`.

src/tui/App.tsx
```diff
@@ -0,0 +1,545 @@
...
+  /** Where the wizard resumes, when it is the opening screen. */
+  wizardStartAt?: WizardStep;
+  onCreatePersona: (answers: WizardAnswers) => Promise<void>;
+}
+
+export function App(props: AppProps): React.ReactElement {
+  const { exit } = useApp();
+  const { stdout } = useStdout();
+  const [host, setHost] = useState(props.host);
+  const [screen, setScreen] = useState<Screen>(
```
andrewagrahamhodges, line 93: **`wizardStartAt` is provably always `undefined` at the point it is read, so "resume the wizard at the first unanswered step" never happens.**

The screen is chosen purely on `startPersona`. But `resolveOpeningScreen()` returns `{ persona: target.name, wizardStartAt: completeness.resumeAt }` for an *incomplete* persona (index.tsx:47) — persona **and** a resume point. So the incomplete case takes `props.startPersona ? "chat"` and opens chat, not the wizard.

And the wizard branch (line 216) is the only reader of `props.wizardStartAt`. It is reachable only when `startPersona` is falsy, which happens only on the `personas.length === 0` early return at index.tsx:42 — where `wizardStartAt` is not set at all. Two paths, and `startAt` is `undefined` on both.

Net effect: the headline behaviour in the PR description ("opens the wizard **resumed at the first unanswered step**") does not occur, and `personaCompleteness.resumeAt` — the reason `WIZARD_STEPS`/`PersonaRequirement.step` carry a step name at all — drives nothing. A user whose only harness is uninstalled gets dropped into a chat box that answers `No harness available. Open settings (^s)…`, which is exactly the "chat box wired to a harness that does not exist" outcome `personaComplete.ts` opens by saying it exists to prevent.

The fix is one line here (`props.wizardStartAt ? "wizard" : props.startPersona ? "chat" : "wizard"`), but please pin it with a test — this is the second time in this PR that a real navigation path is unreachable because state is derived from a constant prop.

src/tui/actions.ts
```diff
@@ -0,0 +1,354 @@
...
+    detail:
+      "The default persona owns the host-level slash commands. Reassigning it " +
+      "moves control of updating and restarting this box to a different " +
+      "phantom, and changes which persona a bare command targets.",
+    longRunning: false,
+    restarts: true,
+  };
+}
+
+export async function applyDefaultPersona(input: {
```
andrewagrahamhodges, line 308: **This writes the layer that loses.** `config.ts:1053` resolves the default as `state.default_persona ?? asString(globalToml.default_persona)` — **state.json wins over config.toml**. Every other writer in the tree knows that and writes state: `cli/persona.ts:220`, `cli/create-persona.ts:109`, `lib/personaDefault.ts:36` and `:80`. This one writes only `toml.default_persona`.

So on any host that has ever created a persona, switched with `phantombot persona <name>`, or been healed by `healDefaultPersonaIfBroken` — i.e. effectively every host — this returns `{ ok: true }`, restarts the daemon, and changes nothing. I ran it against a temp root with `state.default_persona = "alice"`:

```
config.toml now: default_persona = "bob"
state.json now:  {"default_persona": "alice"}
EFFECTIVE default after 'success': alice
```

That is precisely the failure this module's header declares war on: it *declares* a consequence (`describeDefaultPersonaChange`: "bob takes over /update and /restart from alice") and then does not perform it, while reporting success.

Two further things it drops versus `runSwitchDefault`:
- the append-only forensic audit (`state.ts:66`, `auditPersonaChange`) — that log exists so the *writer* of a bad default is identifiable, and a config-only write is invisible to it;
- the `PHANTOMBOT_PERSONA` scope refusal (`cli/persona.ts:168`) that stops a persona agent re-pointing the host-wide default.

Separately: `applyDefaultPersona`, `applyAutostart`, `describeDefaultPersonaChange` and `describeAutostartChange` have **zero callers** anywhere in `src/` or `tests/` — I grepped. Either wire them to a screen (with the state write fixed and a test) or drop them; shipping an untested, wrong-layer writer for the one setting that controls `/update` and `/restart` is a landmine for whoever wires it up next.

andrewagrahamhodges (changes requested):
Changes requested on #477 — the engineering here is genuinely strong (the TTY gate, the trust reasoning, the Yoga parent-chain finding, the space-fingerprint asymmetry are all right, and I checked each), but **the first-run flow this PR exists for dead-ends in an app you cannot quit.**

I verified the good parts rather than taking them on trust: `bun tsc --noEmit` clean, `bun test` **3854 pass / 0 fail** on this box (including `cli-doctor-persona-env`, which you flagged as failing locally), `recentTurns` really is oldest-first so the transcript order is right, `loadConfig()` in `runRepl` does apply the default persona layer, and the `--no-tui` path is correctly *not* read-only so the vault still boots.

Three things:

**1. Completing the wizard leaves a frozen, keyless app.** `props.startPersona` is a constant prop, so the session effect's `if (!props.startPersona) return` never lets go. Finish the wizard, land on `chat`, `session` is `undefined` — you get a bare `opening alice…` with no `Frame`, no footer and no `useInput` mounted. `^q`, `^c` (`exitOnCtrlC: false`) and `esc` are all inert. I mounted `<App>` for real and drove it: `waitUntilExit()` never resolves. Killing the terminal is the only exit, with mouse reporting still on.

**2. "Wizard resumed at the first unanswered step" never fires.** `wizardStartAt` is only read in the wizard branch, which is only reachable when `startPersona` is falsy, which only happens on the `personas.length === 0` path where `wizardStartAt` is never set. The incomplete-persona case gets `chat` instead. So `personaCompleteness.resumeAt` — the whole reason `PersonaRequirement` carries a `step` — drives nothing today.

**3. `applyDefaultPersona` writes config.toml; state.json wins.** Reproduced: it returns `ok`, restarts the daemon, and the effective default is unchanged. It also skips the `state-audit.log` forensic record and the `PHANTOMBOT_PERSONA` scope refusal. It has no callers yet, along with `applyAutostart` and both `describe*` helpers — which is its own problem in a module whose thesis is "no screen may leave the system needing a follow-up command".

Details and suggested fixes inline. Nit, no action needed: the PR description points at `src/tui/data/actions.ts`; the file is `src/tui/actions.ts`.

src/tui/App.tsx
robertclawson: Fixed in ff2b4ab — and confirmed by mutating: with the original `App.tsx` restored, the new suite fails 2 of 3 (`created` never fires past the wizard, and the app never exits).

Took your suggestion but keyed it on a **latch** rather than on `screen`. Keying the effect on `screen === "chat"` fixes the dead end and breaks the thing the session is held above the screen switch *for*: the cleanup would close the session on every `^s` and reopen it on `esc`, re-reading history and losing the thread. So `chatArmed` is set the first time chat is reached — including from the wizard's `onFinish`, which is the path that had no way to set anything before — and the effect depends on `[personaName, chatArmed]`.

The `!session` branch now renders a real `Frame` with a `^q` footer, and the global handler owns `^q`/`^c`/`esc` for exactly that window (`screen === "chat" && !session && !modal`), since `ChatScreen` is not mounted to own them yet.

One more thing that fell out of writing the test: `openChat` rejecting left an unhandled rejection and a permanent `opening …`. It now sets a notice, so a phantom whose harness has been uninstalled says so on screen.

Test is `tests/tui-app-first-run.test.tsx` — it mounts the real `<App>` and drives real keystrokes. Worth noting for anyone writing the next one: Ink 6 pulls input with `readable` + `read()`, so an `EventEmitter` emitting `"data"` renders perfectly and silently receives no keys at all. It needs a real stream.

src/tui/App.tsx
robertclawson: Fixed in ff2b4ab, exactly as you wrote it: `props.wizardStartAt ? "wizard" : props.startPersona ? "chat" : "wizard"`.

Also fixed what let it typecheck in the first place — `resolveOpeningScreen` declared `wizardStartAt?: ReturnType<typeof String>` and the call site cast it `as never`. It is `WizardStep` now, both ends, no cast. A `string` there meant the compiler had nothing to say about a step name that no branch could ever read.

Pinned by `tests/tui-app-first-run.test.tsx`: an incomplete persona (`startPersona: "alice"`, `wizardStartAt: "brain"`) must render the brain step, and a complete one must not. The first fails against the original code.

Taking the "second time a real path is unreachable because state came from a constant prop" point: both remaining constant props are now read once into state that navigation can change, and both regression tests mount `<App>` rather than testing the helpers around it — which is why neither bug was caught, since `resolveOpeningScreen` and `personaCompleteness` were both individually correct.

src/tui/actions.ts
robertclawson: Fixed in ff2b4ab. You are right on all three counts, and the "declares a consequence then does not perform it" framing is the correct indictment — that is the one failure mode this module claims to exist to prevent.

It now writes `state.json` via `saveState`, which brings the `state-audit.log` record with it for free (the audit is inside `saveState`, so a config-only write was invisible to it by construction). State is **re-loaded** immediately before the write rather than trusted from a snapshot, because the confirm panel is an arbitrarily long wait and `harness_bins` discovery can land during it — there is a test asserting a concurrent `harness_bins` value survives. Added the `PHANTOMBOT_PERSONA` refusal mirroring `runSwitchDefault`, and an "already the default" no-op that neither restarts nor writes an audit line for a change that did not happen.

On the zero-callers point — I took the wire-it-up branch rather than the delete branch. `PersonaDetail`'s boot section showed `autostart` and `default` as read-only `Field`s, which is its own small version of the same sin: a screen reporting a setting with no way to change it from there. Both rows are now selectable and go through the confirm panel, so `describeDefaultPersonaChange` states the `/update` and `/restart` handover before it happens. `applyAutostart` and `describeAutostartChange` are wired the same way.

Tests in `tests/tui-default-persona.test.ts` — state layer, field preservation, audit record, agent refusal, no-op. Against the original config.toml writer, 4 of the 5 fail.

andrewagrahamhodges (changes requested):
All three findings are fixed, and fixed at the right layer. One thing left, and it is the headline fix itself: nothing pins it.

**What I verified**

- **Dead end (bug 1)** — gone. `chatArmed` latches on `screen === "chat"`, so the wizard's `onFinish` path now opens a session; `setPersonaName(answers.name)` before `setScreen("chat")` means the new persona (not the host default) is the one opened. The no-session window is now a real `Frame` with a footer, and the global `^q`/`^c`/`esc` handler makes it quittable under `exitOnCtrlC: false`. I drove the wizard for real again and `waitUntilExit()` now resolves. The added `.catch` → `setNotice` is the right call too, and the notice renders outside `body` so it is visible on that frame.
- **Unreachable resume (bug 2)** — gone. `wizardStartAt ? "wizard" : ...` makes the resume path reachable, and dropping the `as never` at the `startTui` call site means `WizardStep` is now actually typechecked end to end. Mutation-reverting the ternary fails `an incomplete persona opens the wizard, not chat`.
- **Losing config layer (bug 3)** — gone, and better than I asked for: `applyDefaultPersona` writes `state.json` (the layer that wins), re-loads state rather than trusting the snapshot, no-ops when already default, and mirrors `runSwitchDefault`'s `PHANTOMBOT_PERSONA` refusal. Mutation-checked: skipping the `default_persona` assignment fails 2 of the 5 new tests, including the audit-record one. `applyAutostart` and `applyDefaultPersona` both have callers now via the two `Selectable` boot rows.
- `bunx tsc --noEmit` clean; **3862 pass / 2 skip / 0 fail** locally across 213 files.

**The one ask: the fix for bug 1 is not pinned.**

I reverted `if (!chatArmed || !personaName) return;` back to `if (!props.startPersona) return;` — the exact bug — and ran the **entire suite**: 3864 tests, 0 fail. `completing the wizard leaves an app that can still be quit` passes either way, because the temp persona does not exist on disk, so `openChat` rejects and the frame is `opening alice…` in both worlds. The test proves the *escape hatch* (Frame + footer + global quit), which is genuinely valuable — but the docblock claims it is a regression test for the session-gate bug, and it is not. Ship this and the same `props.startPersona` gate can be reintroduced by a later refactor with a green suite.

This is the same class of bug the PR is about — it typechecks, renders, and passes every test around it — so it deserves a test that dies when it comes back.

Smallest fix: make the opener injectable and stub it.

```diff
 export interface AppProps {
   host: HostSnapshot;
+  /** Injectable for tests; production always uses `openChat`. */
+  openSession?: (input: { config: Config; persona: string }) => Promise<ChatSession>;
```
```diff
-      const chat = await openChat({ config, persona: personaName });
+      const chat = await (props.openSession ?? openChat)({ config, persona: personaName });
```

Then a fourth case: mount with `openSession` resolving a fake `ChatSession`, drive the wizard to completion, and assert the chat screen actually mounts (a frame with the chat footer, not `opening …`) and that the opener was called with the name the user typed. That one fails on the reverted gate. Everything else is ready to merge.

robertclawson:
Fixed in 67d8506 — you were right that the test proved the escape hatch and not the gate.

`openSession` is now an optional prop, defaulted to `openChat` at the call site, and there is a fourth case: mount with a stub opener, drive the wizard to completion, then assert (a) the opener was called with the name the user typed rather than the host default, and (b) the chat screen is mounted, not the placeholder.

Mutation-checked as you specified — reverting the gate to `if (!props.startPersona) return;` fails it: `opened` is `[]` and the frame stays on `opening alice…`. The other three still pass on the reverted gate, exactly as you found.

One thing your diff surfaced indirectly: the "not showing X" assertion has to read the LAST frame, not the joined ones. `debug: true` rewrites the whole screen on every render, so `frames.join("")` still contains every earlier screen — my first cut failed against the placeholder frame from two renders earlier while the app was in fact sitting on chat. Added `lastFrame()` with that written down.

`bunx tsc --noEmit` clean; suite 3862 pass / 2 skip locally with the same pre-existing `cli-doctor-persona-env` failure on this box.

andrewagrahamhodges (approved):
Approved. The pinning test does what the last round asked for, and I verified it the same way I found the gap.

**Mutation check.** Reverted the gate to the original bug — `if (!props.startPersona || !personaName) return;` — and ran the file: `completing the wizard opens a chat session for the name typed` fails on `expect(opened).toEqual(["alice"])`, received `[]`. The other three still pass, so the new case pins the fix specifically and isn't just re-testing the escape hatch. Restored, 4/4 green.

The seam is right: `openSession?` defaults to `openChat` at the call site, `src/tui/index.tsx` never passes it, and nothing else in `src/` references it — so production has exactly one opener and the injection point can't drift into a real code path. The `lastFrame()` helper plus `debug: true` is the correct way to assert what a screen is *not* showing, given `frames` accumulates.

Full suite here: **3863 pass / 2 skip / 0 fail** across 213 files. `tsc --noEmit` clean. CI green on all six checks including `test-windows`. No unresolved threads.

Good fix on the whole PR — three navigation-from-constant-prop bugs found, fixed at the cause, and now provably pinned.

PR #479, opened 2026-08-28
Make the TUI behave like a terminal app (#471)

robertclawson:
Follow-up to #477, from dogfooding the TUI on the preview channel. Three reported problems: it does not take the terminal, log lines appear under the chat box, and a running turn gives no sign of life.

## 1. It was not a full-screen app

`render()` draws **inline** — a frame as tall as its content, left in the shell's scrollback under whatever was there before. The app now enters the **alternate screen buffer** (`?1049h`), like `less` or `htop`, and restores the shell and its scrollback untouched on every exit path including a crash.

That alone is not enough. Two more pieces were needed, and skipping either reproduces the "frame deforms" report:

- the **root box is given the window's height** — a flex column with no height lays out to its content and floats at the top of an empty screen;
- the **transcript is clipped to the window before layout** (`tui/transcript.ts`) — otherwise a long thread pushes the bottom border off the screen and the terminal scrolls the alternate buffer.

The "no component knows the terminal width" rule is unchanged, and this is not an exception to it: the size is measured in exactly **one** place and passed down by context, a scrolling region may ask how many *rows* it has, and no component does column arithmetic to draw anything. Borders still come from `borderStyle`.

## 2. The logs under the box were ours

`lib/logger.ts` writes JSON lines to `process.stderr`, which is the same terminal the app is drawing on. Log output is now **captured** while the TUI runs and shown on `^l` — captured rather than discarded, because hiding a failure is not a layout fix.

Two writers had to be redirected, not one:

- the logger, via an installable sink (`lib/logSink.ts`) whose default is the same `process.stderr.write` as before, so every non-TUI entrypoint is byte-identical;
- `openChat`'s `stderr`, because a harness subprocess writes to whatever stream it was handed and never consults our logger.

## 3. A working turn now looks like one

The old `thinking…` was static and sat inside the input box. Nothing on screen moved, so a twenty-second turn and a hung process rendered identically — and by report it was not noticed at all. There is now an activity line under the transcript: a **spinner**, the **step the phantom is actually on** (`gh release view`, `thinking`, `writing the reply`), and a **climbing seconds counter**, with `^c interrupts` stated beside it.

## Ink renders the app, `@clack` asks the questions

Chat, dashboard and the reading screens stay Ink. Every question that *changes* something — a value to type, a choice to make, a confirmation to give — is now the same `@clack` prompt the subcommands already use, so masked input, cancel semantics and validation have one implementation instead of two.

Ink cannot be paused, so a prompt is **bracketed**: writes gated off, keystrokes no longer forwarded, mouse reporting off, alternate screen left so the prompt draws on the normal terminal; on the way back raw mode is re-asserted (clack leaves the terminal cooked) and the frame is cleared and repainted in full, because Ink diffs against a frame that is no longer on the screen. A cancelled prompt changes nothing — `undefined`/`false`, never `""` and never a silent yes. The two Ink modal components this replaces are deleted rather than left dead.

## Dressed menus

A row now carries a selection bar, an icon, a description of what it leads to and a right-aligned state badge. Selection is signalled three ways at once — bar, bold label, reversed key hint — because colour alone is unreliable across terminal themes and unavailable to a colour-blind user. The dashboard rows and the persona-detail lists both use it.

## Also

A chunk carrying a newline *inside* it — a paste, or a terminal batching keystrokes — now sends the line instead of typing a literal CR into the input box. Ink reports `key.return` only for a chunk that is exactly `\r`.

## Tests

- `tui-terminal` — the write gate drops Ink's output while suspended and passes it again after; the alternate screen enters once and restores exactly once (exit handlers overlap); size falls back when there is no TTY to ask.
- `tui-transcript` — clipping keeps the newest messages, always shows the latest even when it does not fit, and counts wrapped rows and collapsed vs expanded tool lines.
- `lib-logSink` — the logger routes to an installed sink and back to stderr; redaction still applies through the sink; the buffer parses logger JSON, keeps free-form harness stderr verbatim, and stays bounded.
- `tui-mouse` — the forwarding gate stops feeding Ink and turns reporting off while a prompt owns the terminal, and restores both.
- `tui-prompts` — yes/no/cancel semantics, and the suspend bracket resumes even when the prompt throws.
- `tui-chat-activity` — the indicator is asserted on **motion**, not on a word being present: a static indicator passes "contains thinking" and fails a user. Mutation-checked by freezing the spinner, which fails the test.

Verified by hand as well: driven in a real pty at 30×100 the app fills the window and the frame holds, and the compiled binary's non-TTY behaviour (bare, piped, `--help`, `--version`) is unchanged.

Part of #471. Not for merge until it has been dogfooded on a preview phantom.


robertclawson:
Follow-up pushed (`12bd5e9`) — the two reported bugs plus what fixing them turned up. Redeployed to Kai at `/home/kai/.local/bin/phantombot-tui` for dogfooding.

**`^t` did nothing** because collapsed and expanded both rendered `tool.title.split("\n")[0]`, and a progress note is one line long. Collapsed is now a real summary (`› 3 steps · 12s   ^t details`), expanded is the itemised list with per-step durations. `transcript.ts` had to move with it: it reserved one row per call while the screen drew one row total.

**`^s` and `^p` were the same destination.** `^s` now opens the settings of the phantom you are talking to, `^p` opens the host table, and `esc` from settings returns to whichever screen it was entered from — the global escape net and the screen's own handler both see the key, so they had to be made to agree.

**The settings screen is rebuilt to the spec's mockup**: identity files with `↵ edit in $EDITOR`, brain with its three-layer resolution, channels with per-channel live state (including "stated but no token resolved"), memory with the nightly and index sync, voice with the STT consequence as a reading, boot, MCP, vault, doctor. Brain and channels reuse `runHarness` / `runTelegram` inside the prompt bracket rather than a second copy of each question.

Three things found while building it, each with a test:

- **A screen taller than the window is compressed, not clipped.** At 20 rows the settings screen printed two readings on one line and the bottom border came out `╰─ ─ ─✚─Doctor────`. `Frame` now clips its body as a backstop and growable screens window their own blocks (`tui/scroll.ts`) and say what they hid.
- **A text field must read its live value from a ref.** Chunks can arrive between renders, and a chunk carrying its own newline is not reported as `key.return` — together they could create a persona named `"alice\n"` or `""`. Shared in `tui/textInput.ts`; the burst test fails on the closure read.
- **`hostSnapshot` shelled out to systemctl**, putting a subprocess on the path every screen transition awaits. The first-run regression test caught it as dropped keystrokes at startup. The probe is separate and lazy now.

All six checks green.

src/lib/platform.ts
```diff
@@ -64,7 +64,14 @@ export function currentPlatform(): Platform {
  * so the only way to hit this branch is `bun src/index.ts` on Windows
  * or BSD, where the user is on their own.
  */
-export function defaultServiceControl(): ServiceControl {
+export function defaultServiceControl(
+  env: Record<string, string | undefined> = process.env,
+): ServiceControl {
+  const backend = hostServiceControl();
+  return isSandbox(env) ? sandboxServiceControl(backend) : backend;
```
andrewagrahamhodges, line 71: Non-blocking, for the follow-up: this makes `PHANTOMBOT_SANDBOX` a **global** kill-switch for every service mutation in the process, not a TUI one. `defaultServiceControl()` has ~15 callers, including `cli/update.ts`, `lib/serviceLifecycle.ts` and `channels/commands.ts` — so with the var set, `/update` and `/restart` also stop restarting, and they report `ok: true`.

The reasoning for suppressing rather than erroring is right for a dev checkout. Two small things would stop it being a landmine:

1. **Say something.** A suppressed restart is currently silent — `stderr` carries the explanation, but most call sites only check `ok`. A `log.warn("platform: service change suppressed", { reason: "PHANTOMBOT_SANDBOX" })` inside `suppressed()` makes it visible in the log pane and the journal, which matters given the standing rule that `/update` is the only sanctioned update path.
2. **Document it.** It appears nowhere in README.md or AGENTS.md. An env var that can silently disable updating a production daemon should be written down next to the thing it disables.

Behaviour and tests are fine as they stand (`lib-platform` covers the empty/`0`/truthy parsing and read-through `isActive`); this is about discoverability.

src/tui/index.tsx
```diff
@@ -51,23 +56,87 @@ export async function startTui(): Promise<number> {
...
+
+  const restore = () => {
+    restoreHost();
+    restoreLogs();
+    installed.teardown();
+    fullScreen.restore();
+  };
   process.once("exit", restore);
   process.once("SIGINT", restore);
   process.once("SIGTERM", restore);
```
andrewagrahamhodges, line 142: Non-blocking: registering listeners for `SIGINT`/`SIGTERM` **suppresses Node's default termination**, and `restore()` never exits. So an external `kill` on the TUI tears the terminal down — alternate screen left, mouse reporting off, log sink and stdin tap removed — and then the app keeps running, now painting over the user's shell and deaf to the mouse. It takes a *second* signal to actually die, because `once` has already consumed the listener.

In practice `^q` and `^c` go through Ink (raw mode means the terminal generates no SIGINT), so this is only reachable from an outside `kill` or a systemd stop — hence non-blocking. The fix is one line: have the signal handlers `restore()` and then `process.exit(130)` / `process.exit(143)`, leaving the `exit` handler as the idempotent backstop it already is (`FullScreen.restore` and `installed.teardown` are both guarded, so the double call is harmless).

andrewagrahamhodges (approved):
Approved for merge to preview.

I ran this the way it will be used, not just read it. Full suite **3986 pass / 2 skip / 0 fail**, `tsc` clean, all six CI checks green including Windows. Compiled the binary and confirmed the non-TTY contract is intact: `--version`, `--help` and bare-piped all exit 0 and none of them hang or open a renderer. Then drove the compiled binary in a real 30x100 pty: alternate screen entered **exactly once** and left exactly once, mouse reporting on/off balanced, cursor restored, `^q` exits rc=0 with the shell untouched. That is the thing that would have hurt a preview user most, and it holds.

Mutation-checked the headline claim rather than trusting the docblock: stubbing out the `\n` branch of `trimFrameAdvance` fails `tui-frame-drift` on the screen-diff assertion. The test genuinely pins the incremental-render fix.

On your two complaints from the photo, both are actually closed rather than papered over:

- **Lost input after changing something** — the cause was a paused stdin (a readline close or an inherited-stdin child leaves it paused, and a paused stdin never emits `data` again). `setForwarding(true)` now resumes it and re-asserts raw mode, and `forceRepaint` is used instead of `instance.clear()`, which re-syncs Ink to the frame it just erased.
- **"One window app"** — `withPromptTerminal` now has exactly **one** caller left in the whole tree: `$EDITOR`. Every settings flow (Brain, Channels, Voice, secrets, confirms) runs on native Ink screens. Nothing else leaves the app.

One thing I checked specifically because of Robbie's outage yesterday, where a config writer truncated `config.toml` to 54 bytes and only surfaced four days later on reboot: **there is not a single direct file write anywhere under `src/tui/`**. Every mutation routes through the existing `applyVoiceConfig` / `applyEmbeddingConfig` / `writeAutostartPersonas` / `setPersonaSecret` path, and `runHarness` injects the *asking* rather than reimplementing the *writing*. That is the right call and it is what makes this safe to ship on preview.

Two non-blocking notes inline for the follow-up PR — neither should hold the merge.

src/lib/platform.ts
robertclawson: Agreed on both counts, and both land in ton's follow-up PR rather than here (a push now would dismiss the approval and cost a full re-review round). Carrying: (1) a log.warn inside suppressed() naming PHANTOMBOT_SANDBOX as the reason, so a swallowed /update or /restart is visible in the log pane and the journal rather than a silent ok:true; (2) the var documented in README.md and AGENTS.md next to the service-control section. Filed so it cannot get lost.

src/tui/index.tsx
robertclawson: Correct, and taking the one-line fix into tonight's follow-up PR for the same reason (no push here, it would dismiss the approval). Handlers will restore() then process.exit(130)/exit(143), with the exit handler left as the idempotent backstop - FullScreen.restore and installed.teardown are both guarded so the double call is a no-op.

PR #489, opened 2026-08-29
TUI: exit after restoring the terminal on SIGINT/SIGTERM

### LINKED ISSUE #485: TUI: SIGINT/SIGTERM handlers restore the terminal but never exit
Follow-up from the review on #479 (non-blocking there).

`process.once("SIGINT", restore)` / `process.once("SIGTERM", restore)` **suppress Node's default termination**, and `restore()` never exits. So an external `kill` on the TUI tears the terminal down — alternate screen left, mouse reporting off, log sink and stdin tap removed — and then the app keeps running, painting over the user's shell and deaf to the mouse. A *second* signal is needed to actually die, because `once` has already consumed the listener.

Only reachable from an outside `kill` or a systemd stop: `^q` and `^c` go through Ink, since raw mode means the terminal generates no SIGINT.

Fix: have the signal handlers `restore()` then `process.exit(130)` / `process.exit(143)`, leaving the `exit` handler as the idempotent backstop it already is (`FullScreen.restore` and `installed.teardown` are both guarded, so the double call is harmless).

Part of #471

lenaparkhodges:
Closes #485. Part of #471.

**Bug:** `process.once("SIGINT", restore)` suppresses Node's default termination, and `restore()` never exited — an outside `kill` or systemd stop restored the terminal (alternate screen left, mouse reporting off, log sink and stdin tap removed) and then the app *kept running*, painting over the user's shell until a second signal arrived.

**Fix:** extracted the exit wiring into `installSignalExit()` — the signal handlers now `restore()` then `process.exit(130)` / `process.exit(143)` (128 + signal). The plain `exit` handler stays as the idempotent backstop; `FullScreen.restore` and `installed.teardown` are both guarded, so the double call is harmless. Extraction also makes the behavior unit-testable; the helper returns a teardown for tests.

**Tests:** new `tests/tui-signal-exit.test.ts` — SIGINT exits 130, SIGTERM exits 143, both after restore; `once` semantics verified via listener counts (a bare second synthetic SIGINT would kill the test runner itself); teardown removes all listeners including the exit backstop. Full suite: 3,990 pass, 0 fail. `tsc --noEmit` clean.

### DIFF
--- a/src/tui/index.tsx
+++ b/src/tui/index.tsx
@@ -137,9 +137,7 @@ export async function startTui(): Promise<number> {
     installed.teardown();
     fullScreen.restore();
   };
-  process.once("exit", restore);
-  process.once("SIGINT", restore);
-  process.once("SIGTERM", restore);
+  installSignalExit(restore);
 
   try {
     await instance.waitUntilExit();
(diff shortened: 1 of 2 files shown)


kaieriksen (approved):
Reviewed the full diff at e9f222b. The signal handlers restore the TUI before exiting with conventional SIGINT/SIGTERM codes, and the exit backstop is safely idempotent. Verified with the focused tests, TypeScript typecheck, a real child-process SIGTERM probe (restore observed, exit 143), and the full suite: 3,990 passed, 2 skipped, 0 failed.

PR #490, opened 2026-08-29
Multi-persona rigs: per-persona heartbeat instances + maintenance coverage (#486)

### LINKED ISSUE #486: Multi-persona rigs: heartbeat and nightly only ever run for the default persona
Since `phantombot run` can serve several personas in one process (`autostart_personas`), a single UNIX account can host a whole rig — kw-phantombot now runs Lena, Kai and Jake under `supervisor`, with Lena as default. The **runtime** side of that works. The **maintenance** side is still written as if there were exactly one persona per account, so every non-default persona silently gets no memory upkeep at all.

Nothing here is a crash. It is all silent degradation, which is what makes it worth fixing before more people build rigs.

### 1. The heartbeat only ever sweeps the default persona
`phantombot heartbeat` takes `--persona` (`src/cli/heartbeat.ts:412`), defaulting to `PHANTOMBOT_PERSONA` → `default_persona`. But `install` writes exactly **one** `phantombot-heartbeat.service`, and its `ExecStart` carries no persona:

```
ExecStart=/home/supervisor/.local/bin/phantombot heartbeat
```

So on a three-persona rig only the default persona gets the 30-minute pass. Kai and Jake get **no** drawer promotion from tagged captures, **no** memory-index refresh, **no** `flushDueConversationTurns`, **no** log rotation, and **no** timer-health recording. Their `journal_entries` grow forever and their `memory search` goes progressively staler against their own recent turns.

### 2. The nightly inherits the same blast radius
There is no `phantombot-nightly.timer` by design — the nightly hangs off the heartbeat's day-rollover detection (`spawnNightlySweep` via `dayRolledOver`) plus the startup sweep in `run`. Both are per-persona. One unswept heartbeat therefore means one persona that **never** runs a nightly: no distillation into drawers, no KB notes, no `MEMORY.md` update, no compaction, and a nightly ledger that reports an ever-growing backlog nobody sees.

### 3. Doctor doesn't surface the gap
`doctor` already walks `autostart_personas` for Telegram listener coverage (`src/cli/doctor.ts:402`). It should do the same for maintenance: for every persona this host serves, is there a heartbeat unit that names it, when did it last fire, and how many days is its nightly ledger behind? Today a rig can be six weeks behind on two of three personas and doctor is all-green.

### 4. `install` has no notion of per-persona units
`installPhantombot` takes single `heartbeatServicePath` / `heartbeatTimerPath` / plist paths. There is no templating and no per-persona enable/disable, on any of the three platforms (systemd user units, launchd plists, Windows Task Scheduler). `persona new` / `persona add-to-autostart` don't provision maintenance for the persona they just created, and removing a persona leaves its unit behind.

### 5. `/update` and `/restart` are gated to the default persona
Sensible when one account meant one persona; on a shared rig it means only Lena can trigger an update, and Kai/Jake report `ok:true` while doing nothing. At minimum the gate should say *why* it declined.

## What is NOT broken (don't "fix" it)
- **`tick` is correct.** It has no `--persona` flag because it doesn't need one: task rows carry their persona and `tick` resolves each task's effective config per persona (`src/cli/tick.ts:120-160`). One timer genuinely covers the whole rig.
- **`run` is correct.** One process, all `autostart_personas`, per-persona channels.
- **The vault/identity split is correct** — each persona keeps its own `identity.json`-derived vault.

## Proposed fix
1. Template the maintenance units: `phantombot-heartbeat@.service` with `ExecStart=… heartbeat --persona %i`, plus matching `@.timer`; equivalents for launchd (one labelled plist per persona) and Task Scheduler (one task per persona).
2. Enable one instance per served persona = `{default_persona} ∪ autostart_personas`. Make `persona new`, autostart add/remove and `uninstall` provision and tear down the instance in the same code path (AGENTS.md invariant 21: new machinery needs its caller in the same PR).
3. Idempotent migration in `ensureSystemdUnitsCurrent` — existing hosts keep working: install the templated units, enable the default persona's instance, disable/remove the legacy plain unit only once the replacement is active.
4. Doctor check: per served persona, heartbeat unit present + enabled, last-fire age, nightly ledger backlog in days. Warn, don't fail, when a persona is behind.
5. Backfill: the sweep is date-driven, so a rig that has been running blind for weeks will try to process every missed day at once. Decide explicitly whether to cap that (process the last N days, mark the rest swept) or let it grind — either way it must not wedge the first heartbeat after upgrade.
6. Make the `/update` and `/restart` default-persona gate explain itself in the reply instead of silently reporting success.

## Acceptance
- On a host with `default_persona = lena` and `autostart_personas = ["kai", "jake"]`, a fresh install and an in-place upgrade both end with three enabled heartbeat instances, and all three personas show a heartbeat fire within 30 minutes and a nightly sweep within 24 hours.
- `phantombot doctor` reports maintenance coverage per persona and flags a missing or stale one.
- Removing a persona from autostart removes its maintenance instance.
- Single-persona hosts see no behavioural change.

Reference rig: kw-phantombot, `supervisor` account, v1.1.317, linger enabled.

lenaparkhodges:
Closes #486.

## The bug

`phantombot run` serves every persona in one process, but the maintenance side was still single-persona: one `phantombot-heartbeat.timer` swept only the default persona. On a multi-persona rig, every non-default persona silently got **no** drawer promotion, **no** memory-index refresh, **no** turn-flush, **no** log rotation — and **no nightly at all**, because the rollover trigger read a host-global last-fired marker: the first persona to fire after midnight consumed the rollover for everyone.

## The fix

**systemd — `phantombot-heartbeat@` template + per-persona instances**
- `phantombot-heartbeat@.service` runs `heartbeat --persona %i`; one enabled `phantombot-heartbeat@<persona>.timer` per served persona (`{default_persona} ∪ autostart_personas`, new `servedPersonasOf()` in config).
- `ensureSystemdUnitsCurrent` (the 30-min heal shared by heartbeat/doctor/update) reconciles instances: arms served personas, **disables instances of personas removed from autostart**, restarts every instance when the template is rewritten, and retires the legacy `phantombot-heartbeat.{service,timer}` **only after the default persona's instance is verified active** — a failed migration never leaves a host heartbeat-less (issue point 3).
- `install` arms instances per persona; `uninstall` tears down all instances + templates + legacy units.
- The autostart picker, `persona new --autostart`, and the TUI autostart action provision/retire the instance immediately (issue point 2 / invariant 21); the heal is the backstop.

**Rollover starvation** — heartbeat fire-markers are now per-persona (`heartbeat.<persona>.last-fired`). The default persona falls back to the legacy marker at upgrade so it doesn't read falsely stale; non-default personas deliberately get **no** fallback — the legacy marker says nothing about their fires.

**doctor (point 4)** — new warn-only `maintenance per persona` section: each served persona's heartbeat last-fire/staleness + nightly ledger backlog. A stale persona drives a force-re-arm of *its* instance through the existing zombie-timer path; a missing instance is armed by the same heal.

**run** — the startup nightly sweep now fires per served persona, not only the default.

**Windows** — the heartbeat task was already *named* per-persona (`\Phantombot\heartbeat-<persona>`) but ran bare `phantombot heartbeat`, silently sweeping the default persona. It now passes `--persona`; the existing `ensureTasksCurrent` heal re-registers drifted tasks.

## Decisions the issue asked for

- **Backfill (point 5): no cap.** The sweep is detached and ledger-driven — a persona blind for weeks drains its backlog in the background without wedging the first heartbeat after upgrade.
- **`/update` + `/restart` gate (point 6): already fixed.** `lifecycleRefusal` (#439) declines with an explanation naming the default persona; verified, no change needed.
- **macOS launchd per-persona plists: deferred** — no Mac in the fleet to verify against (follow-up issue to be filed). macOS keeps the current single heartbeat (default persona) — status quo, not a regression.

## Acceptance mapping

- Fresh install + in-place upgrade → instances armed per served persona; legacy retired only when safe. ✅ (`installPhantombotUnit` / `ensureSystemdUnitsCurrent`, pinned by tests)
- `doctor` reports maintenance coverage per persona and flags missing/stale. ✅ (new section + instance checks)
- Removing a persona from autostart removes its maintenance instance. ✅ (`disableUnservedHeartbeatInstances` + lifecycle hooks)
- Single-persona hosts see no behavioural change. ✅ (one instance replaces the plain unit; Windows task gains a flag that resolves to the same persona; markers fall back for the default persona)

## Verification

- 4003 tests pass, 0 fail; tsc clean. (9 lock/workspace concurrency tests fail identically on `main` on this dev host — pre-existing, unrelated.)
- New tests: instance arming/disable/migration-ordering, template ExecStart, per-persona markers + fallback, per-persona rollover isolation, doctor maintenance section (text + JSON + force-re-arm wiring), lifecycle-hook sync, Windows task args, `servedPersonasOf`.
- Real-rig verification (kw-phantombot: lena + kai under supervisor) happens on the next `/update` after merge — the heartbeat heal performs the migration.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -236,7 +236,7 @@ On Linux, `phantombot install` creates **three** systemd-user units:
 | Unit | Cadence | What it does |
 |---|---|---|
 | `phantombot.service` | always-on | `phantombot run` — Telegram listener |
-| `phantombot-heartbeat.timer` → `.service` | every 30 min | mechanical maintenance, no LLM |
+| `phantombot-heartbeat@<persona>.timer` → `@.service` template | every 30 min, one instance per served persona (default ∪ autostart) | mechanical maintenance for that persona, no LLM |
 | `phantombot-tick.timer` → `.service` | every 1 min | fires due scheduled tasks |
 
 There is deliberately no nightly unit. The cognitive pass is event-driven —
--- a/src/lib/systemd.ts
+++ b/src/lib/systemd.ts
@@ -13,8 +13,28 @@ import { isPhantombotBinary } from "./binaryIdentity.ts";
 import type { WriteSink } from "./io.ts";
 
 export const PHANTOMBOT_UNIT_NAME = "phantombot.service";
-export const HEARTBEAT_SERVICE_NAME = "phantombot-heartbeat.service";
-export const HEARTBEAT_TIMER_NAME = "phantombot-heartbeat.timer";
+/**
+ * Heartbeat units are systemd TEMPLATES (#486): one
+ * `phantombot-heartbeat@<persona>.timer` instance per served persona
+ * (default_persona ∪ autostart_personas), so every persona on a
+ * multi-persona rig gets its own 30-minute maintenance pass — drawer
+ * promotion, index refresh, turn-flush, day-rollover nightly trigger —
+ * instead of only the default persona being swept.
+ */
+export const HEARTBEAT_SERVICE_NAME = "phantombot-heartbeat@.service";
+export const HEARTBEAT_TIMER_NAME = "phantombot-heartbeat@.timer";
+/** Pre-#486 single-persona heartbeat units — retired by the heal path once
+ * the default persona's instance is verified active. */
+export const LEGACY_HEARTBEAT_SERVICE_NAME = "phantombot-heartbeat.service";
+export const LEGACY_HEARTBEAT_TIMER_NAME = "phantombot-heartbeat.timer";
+/** Name of the heartbeat timer instance serving one persona. */
+export function heartbeatInstanceTimer(persona: string): string {
+  return `phantombot-heartbeat@${persona}.timer`;
+}
+/** Name of the heartbeat service instance serving one persona. */
+export function heartbeatInstanceService(persona: string): string {
+  return `phantombot-heartbeat@${persona}.service`;
+}
 /**
  * RETIRED units. The nightly no longer runs on a clock — it is triggered by
  * startup and by the heartbeat noticing the calendar day rolled over (see
@@ -62,6 +82,28 @@ export function heartbeatServicePath(): string {
 
 export function heartbeatTimerPath(): string {
   return join(homedir(), ".config", "systemd", "user", HEARTBEAT_TIMER_NAME);
+}
+
+/** Path of the retired pre-#486 heartbeat service (kept for cleanup only). */
+export function legacyHeartbeatServicePath(): string {
+  return join(
+    homedir(),
+    ".config",
+    "systemd",
+    "user",
+    LEGACY_HEARTBEAT_SERVICE_NAME,
+  );
+}
+
+/** Path of the retired pre-#486 heartbeat timer (kept for cleanup only). */
+export function legacyHeartbeatTimerPath(): string {
+  return join(
+    homedir(),
+    ".config",
+    "systemd",
+    "user",
+    LEGACY_HEARTBEAT_TIMER_NAME,
+  );
 }
 
 /** Path of the retired nightly service (kept for cleanup only). */
@@ -143,11 +185,17 @@ function quoteArg(s: string): string {
   return `"${s.replace(/(["\\$`])/g, "\\$1")}"`;
 }
 
-/** Generate the heartbeat oneshot service body. */
+/**
+ * Generate the heartbeat oneshot service TEMPLATE body (#486). The `%i`
+ * instance specifier becomes the persona name, so one template serves
+ * every `phantombot-heartbeat@<persona>.timer` instance on the host.
+ */
 export function generateHeartbeatService(binPath: string): string {
-  const exec = [binPath, "heartbeat"].map(quoteArg).join(" ");
+  const exec = [binPath, "heartbeat", "--persona", "%i"]
+    .map(quoteArg)
+    .join(" ");
   return `[Unit]
-Description=Phantombot heartbeat — mechanical 30-minute maintenance pass
+Description=Phantombot heartbeat — mechanical 30-minute maintenance pass (%i)
 
 [Service]
 Type=oneshot
@@ -170,7 +218,7 @@ StandardError=journal
  */
 export function generateHeartbeatTimer(): string {
   return `[Unit]
-Description=Phantombot heartbeat timer (every 30 min)
+Description=Phantombot heartbeat timer (every 30 min, %i)
 
 [Timer]
 OnCalendar=*:0/30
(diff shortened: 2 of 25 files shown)


src/lib/systemd.ts
```diff
@@ -814,12 +890,132 @@ export async function ensureSystemdUnitsCurrent(
...
+      const enabled = await opts.systemctl.run(["--user", "is-enabled", unit]);
+      const active = await opts.systemctl.run(["--user", "is-active", unit]);
+      const isEnabled =
+        enabled.exitCode === 0 && enabled.stdout.trim() === "enabled";
+      const isActive =
+        active.exitCode === 0 && active.stdout.trim() === "active";
+      let ready = isEnabled && isActive;
+      if (ready) {
+        if (forceRearm.has(unit) || templateRewrote) {
+          await opts.systemctl.run(["--user", "restart", unit]);
```
kaieriksen: Major: the restart result is ignored, while `ready` remains true from the pre-restart probe. If the template was rewritten and `systemctl restart` fails after stopping the default instance, line 933 still retires the legacy timer, violating the no-heartbeat-less migration guarantee. Only keep `ready = true` after a successful restart (ideally re-check active), and add a failure-path test proving the legacy files/unit remain.

AGENTS.md
```diff
@@ -236,7 +236,7 @@ On Linux, `phantombot install` creates **three** systemd-user units:
 | Unit | Cadence | What it does |
 |---|---|---|
 | `phantombot.service` | always-on | `phantombot run` — Telegram listener |
-| `phantombot-heartbeat.timer` → `.service` | every 30 min | mechanical maintenance, no LLM |
+| `phantombot-heartbeat@<persona>.timer` → `@.service` template | every 30 min, one instance per served persona (default ∪ autostart) | mechanical maintenance for that persona, no LLM |
```
kaieriksen, line 239: Major: this changes the public systemd unit shape, but README was not updated as required by this guide. README's Maintenance table still tells operators to inspect `phantombot-heartbeat.timer`; after this PR that unit is retired and the actionable units are `phantombot-heartbeat@<persona>.timer`. Please update README in this PR so install/diagnostic instructions match production.

kaieriksen (dismissed):
Two changes are required before merge. The migration safety gate can retire the legacy heartbeat after a failed restart of the default instance, and README still documents the retired single heartbeat unit. Focused tests and typecheck pass; I am running the full suite separately.

kaieriksen:
Verification update: typecheck passed; focused maintenance coverage passed (312/312); full suite passed with Bun on PATH (4,012 pass, 2 platform skips, 0 fail). The initial shell-only run reproduced the known 9 child-process PATH failures and was superseded by the clean run.

src/lib/systemd.ts
lenaparkhodges: Fixed in 1606ac6: after an instance restart, `ready` now only survives a verifiable re-probe — exit!=0 or a post-restart `is-active` that isn't active drops it to false, so the legacy unit stays. Two new failure-path tests cover both outcomes (restart fails; restart ok but instance inactive), asserting the legacy files/unit remain.

AGENTS.md
lenaparkhodges: Fixed in 1606ac6: README's Maintenance table now lists `phantombot-heartbeat@<persona>.timer` (one per served persona) and notes the legacy single-persona unit is retired automatically.

kaieriksen (approved):
Re-review of 1606ac6: both requested changes are correctly fixed. Failed or inactive post-restart instances now keep the legacy heartbeat in place, with coverage for both failure modes; README documents the per-persona unit. Focused systemd tests (54/54) and typecheck pass; CI is green.

PR #492, opened 2026-08-29
TUI: slash commands and rendered markdown in the chat screen (#480, #481)

### LINKED ISSUE #480: TUI: slash commands in the chat screen
Part of #471.

The TUI chat screen has no slash-command layer. Telegram and phantomchat both accept `/…` commands (`/update`, `/restart`, `/status`, etc.); the TUI does not, so the only way to run one from the terminal is to drop out to the shell.

**Ask:** implement the same slash-command set in the TUI chat input as the other channels expose, routed through the same handler rather than a TUI-local reimplementation, so the three channels can't drift.

Worth having:
- inline completion / a hint list when the input starts with `/`
- the same persona gating that already applies elsewhere (`/update` and `/restart` only on the default persona)
- unknown commands answered in-TUI, not sent to the model

### LINKED ISSUE #481: TUI: render markdown in the chat transcript
Part of #471.

Agent replies land in the TUI chat screen as raw markdown. Tables, code blocks, lists and emphasis are all unreadable compared to Telegram/phantomchat.

**Ask:** render markdown in the transcript — at minimum fenced code blocks (monospace block, ideally syntax-highlighted), tables, headings, bullet/numbered lists, bold/italic, inline code and links.

Constraints from the existing render work:
- must survive terminal resize and the width-drop `forceRepaint` path
- Yoga compresses rather than clips, so wide tables need an explicit wrap/truncate policy, not hope
- no flicker regression on the full-screen Ink path

robertclawson:
Part of #471. Closes #480. Closes #481.

Screen 0 had neither of the two things every other channel already has: a command layer in front of the harness, and a reply that is rendered rather than dumped. Both land here because they touch the same file and the same row arithmetic — splitting them would mean two passes over `transcript.ts` and the chat screen's key map.

## Slash commands (#480)

`src/tui/slash.ts` decides only what the terminal **advertises** — the list is derived from `TELEGRAM_BOT_COMMANDS`, never retyped — and `chatSession.command()` delegates to the same `handleSlashCommand` that Telegram, phantomchat and ACP call. Persona gating on `/update` and `/restart` therefore comes free, and a command added once shows up on every surface.

Three things are load-bearing:

- **Dispatching ahead of the harness is the point.** `/stop` exists to kill the turn that is currently blocking everything, so neither it nor typing is gated on `busy` — a keyboard that stops accepting characters exactly when a turn hangs cannot be used to stop it. The session keeps its **own** `AbortController` and forwards the screen's `^c` into it; an `AbortSignal` cannot be aborted by whoever merely holds it, so without that `/stop` would have nothing to abort.
- **Command-shaped is not the same as ours.** `/wat` is answered in-app with the list and the way out; `/usr/bin/env is on PATH?`, `/etc/hosts` and `/^ab+c$/` reach the phantom untouched.
- **Messages are patched by identity, not by "the last message".** A command typed mid-turn appends a second bubble, and index-based patching then writes the streaming reply into it.

Plus the type-ahead: `/` lists the commands, typing narrows it, `Tab` completes to the longest unambiguous prefix, and the rows it takes are subtracted from the transcript in the same render so a `/` keystroke cannot grow the column and tear the frame.

Line mode (`--no-tui`) gets the same dispatch.

## Markdown (#481)

`src/tui/markdown.ts` is a pure `(text, width) -> rows` function emitting **spans, never ANSI**: escape sequences are not glyphs, so a decorated string would be measured wrong and the transcript's "a row of scroll is a row on screen" arithmetic — what keeps clipping ahead of layout — would break.

Yoga compresses rather than clips, so every block states its own width policy:

| block | policy |
| --- | --- |
| prose, lists, quotes | word-wrap, hanging indent on list items |
| fenced code | truncate with `…` — a re-flowed command line looks copy-pasteable and is not |
| tables | columns shrink widest-first to a floor of 3, cells truncate |
| headings, rules | bold/accent; a rule fills the width exactly |
| inline | bold, italic, code, links (the URL is kept — a terminal cannot be clicked) |

Only **assistant** text is rendered: an error is shown verbatim (a stack trace full of `*` would come out half-eaten) and so is the user's own line. A blank markdown row draws a space, because an empty Ink box collapses to zero height and would put every scroll offset below it out by one. Purity is what makes the width-drop `forceRepaint` path safe — there is no cached layout to invalidate.

## Tests

New: `tui-markdown` (width invariants at 24/40/60/80, wrap vs truncate, tables, escapes), `tui-slash` (advertised set derived from the shared one; command vs path in both directions), `tui-chat-commands` (delegation, and `/stop` aborting a turn that is genuinely in flight against a hanging harness), `tui-chat-slash` (keyboard path, type-ahead, command during a streaming turn, `afterSend` ordering), `tui-chat-markdown` (blank rows survive layout; a too-wide table does not shear the frame).

Suite: 4066 pass, 1 fail — `cli-doctor-persona-env`, which fails identically on `main` on this host (it reads the host's real embedding config).

Docs: README "The terminal app" gains the commands and rendering sections plus the `/` and `Tab` keys; AGENTS.md gains invariant 36.


src/tui/markdown.ts
```diff
@@ -0,0 +1,427 @@
...
+  indent: number;
+}
+
+const BULLET = "•";
+const ELLIPSIS = "…";
+
+/** The visible width of a span run. */
+function widthOf(spans: readonly Span[]): number {
+  let n = 0;
+  for (const s of spans) n += s.text.length;
```
kaieriksen: **Major — measure terminal columns, not UTF-16 code units.** `String.length` is not visible width: at width 20, `markdownLines("界".repeat(20), 20)` emits one row whose Ink/string-width width is 40, so Yoga can compress/shear the frame this renderer is intended to protect. Combining marks are over-counted, and `slice(0, left)` can also split a surrogate pair when the remaining budget is odd. Please use a terminal-width-aware measure and grapheme-safe slicing throughout `widthOf`, wrapping, and truncation, with CJK/emoji/combining-mark regressions measured using the same width semantics Ink uses.

src/tui/markdown.ts
```diff
@@ -0,0 +1,427 @@
...
+  const count = Math.max(header.length, ...rows.map((r) => r.length), 1);
+  const cells = [header, ...rows].map((r) =>
+    Array.from({ length: count }, (_, c) => r[c] ?? ""),
+  );
+  const natural = Array.from({ length: count }, (_, c) =>
+    Math.max(3, ...cells.map((r) => plainWidth(r[c]!))),
+  );
+  const gap = 3; // " │ "
+  const budget = cols - gap * (count - 1);
+  const widths = fitColumns(natural, Math.max(count * 3, budget));
```
kaieriksen: **Major — many-column tables still exceed the hard width.** Raising the content budget to `count * 3` guarantees every column keeps three cells even when content plus separators cannot fit. A six-column table at widths 20, 24, or 30 emits 33-column rows in all three cases (`6*3 + 5*3`), directly violating the documented `no row wider than cols` invariant and allowing the frame to deform. Please define a narrow-terminal policy that actually stays within `cols` (for example truncating/omitting columns or lowering the cell floor) and add a regression with more columns than the viewport can accommodate.

kaieriksen (changes requested):
Two terminal-width regressions need fixing before merge. Focused TUI tests and typecheck pass, and hosted CI is green, but the new width invariant is not true for wide Unicode or tables with enough columns.

robertclawson:
Both width findings fixed in 0c49a67 — thanks, these were real.

**Measurement.** `widthOf` summed `String.length`, so `markdownLines("界".repeat(20), 20)` really did emit a 40-column row, and combining marks were over-counted. Width now goes through `string-width` (the measure Ink itself uses), and every cut — wrap hard-split, ellipsis truncation, whitespace trim — is taken between grapheme clusters via `Intl.Segmenter`, so no surrogate pair or base+mark pair gets split. `string-width` is now a declared dependency rather than a transitive one via ink.

**Tables.** You're right that `Math.max(count * 3, budget)` only guaranteed the floor, not the row. The floor is genuinely load-bearing (three columns is the narrowest readable cell), so the policy is now: `fittableColumns()` computes how many columns fit at floor+separators, trailing columns are dropped rather than overflowed, and the loss is admitted with a dim `…` column when there is slack for one. A table that fits renders exactly as before.

New regressions: CJK/emoji/combining marks at several widths, no-lone-surrogate checks through code fences, table cells and prose, and the six-column table at 12/20/24/30/33/60. The suite's own "no row wider than cols" helper measured with `.length` as well — it now measures in columns, so it can see the failure it exists to catch.

Full suite: 4075 pass, 1 pre-existing local env-leak failure in `cli-doctor-persona-env` unrelated to this change (fails on the parent commit too); typecheck clean. Re-requesting review.

kaieriksen (approved):
Re-reviewed at 0c49a67. Both width blockers are fixed: rendering now measures terminal columns and cuts on grapheme boundaries, while over-wide tables omit trailing columns without exceeding the viewport. Focused markdown/chat regressions and typecheck pass; hosted CI and CodeQL are green.

PR #493, opened 2026-08-29
Polish TUI configuration and persona creation

### LINKED ISSUE #483: TUI: polish the configuration and new-persona screens
Part of #471.

The chat screen is in decent shape after #479; the configuration and new-persona flows are the rough edges left.

**Ask:** a polish pass over both —
- consistent field layout, labels and validation errors shown inline rather than as a dumped error
- clear required-vs-optional fields, and sane defaults prefilled
- the new-persona flow should be a guided sequence with a review step before it writes anything, and should say exactly what it created (persona dir, identity.json, config.toml)
- respect the config layer: show which values are inherited from the global file vs set per-persona, since absence of a key means inherit and a tombstone is needed to override
- esc is always Back, consistent with the rest of the app

Worth capturing screenshots of the current state in this issue before changing anything.

kaieriksen:
Closes #483.

## What changed
- validates required persona names inline, including duplicates, before any write
- makes the final wizard page an explicit review step and names the persona directory, `identity.json`, and `config.toml`
- persists the reviewed brain choice in the persona-local `config.toml`
- reports creation failures inline instead of leaving the TUI
- labels effective brain, channel, memory, and voice settings as persona overrides, inherited global values, or built-in defaults; explicit empty values remain overrides
- keeps `esc` as Back and fixes Enter to accept the visibly prefilled option instead of silently choosing the first row
- updates README and AGENTS.md contracts

## Verification
- typecheck passed
- 17 focused TUI/provenance tests passed
- full suite: 4,017 passed, 2 platform skips; one pre-existing single-tick TUI timing test flaked under full-suite load and passed on immediate focused rerun
- production `bun-linux-x64-baseline` build passed
- compiled binary verified in a real Linux PTY: inline required marker, step navigation, review-before-write, exact artifacts, defaults, and terminal restoration


src/tui/screens/Wizard.tsx
```diff
@@ -224,12 +246,22 @@ export function WizardScreen(props: {
...
-      if (key.return && answers.name.trim()) go(1);
-      else if (key.backspace || key.delete) {
+      if (key.return) {
+        const name = answers.name.trim();
+        const error = !name
+          ? "Required — enter a persona name."
+          : !validPersonaName(name)
+            ? "Use lowercase letters, digits, '-' or '_'; start with a letter or digit."
+            : props.existingNames?.includes(name)
+              ? `A persona named '${name}' already exists.`
```
andrewagrahamhodges, line 257: **Blocking — this dead-ends the resume flow.**

`resolveOpeningScreen` routes an *incomplete existing* persona into the wizard with `initial.name` set to that persona and `wizardStartAt` at the failing step. `App` passes `existingNames={host.personas.map(p => p.name)}`, which contains it. So: resume `alice` at `brain`, press `esc`, and the name field rejects Enter permanently. On that path `navRef` is empty, `onBack` is `undefined`, and the footer offers only `^q` — the user has to kill the app.

I drove the real `<App>` to confirm; the frame after `esc` then Enter is:

```
 | alice_ |
 Required · lowercase, no spaces · becomes the persona directory
 ✗ A persona named 'alice' already exists.
 → ↵ Continue  ✖ ^q Quit
```

Suggested fix — exempt the name the wizard was seeded with:

```ts
const seeded = props.initial?.name?.trim();
: name !== seeded && props.existingNames?.includes(name)
    ? `A persona named '${name}' already exists.`
```

The same clause is duplicated at line 282 (the paste/submit path), so both need it. Worth a regression test that mounts with `startPersona` + `wizardStartAt` and walks back to the name step — that is the only shape that catches it.

src/tui/snapshot.ts
```diff
@@ -424,6 +448,12 @@ export async function personaSnapshot(
     secretNames,
     memory: readMemory(config, name),
     completeness: await personaCompleteness(config, name),
+    configSources: {
+      brain: sourceOf(personaToml, globalToml, ["harnesses", "chain"]),
+      channels: sourceOf(personaToml, globalToml, ["channels"]),
+      memory: sourceOf(personaToml, globalToml, ["retrieval"]),
```
andrewagrahamhodges: **Blocking — `memory` provenance reads the wrong key.**

The Memory panel's `from` line sits directly under `embeddings` (PersonaDetail.tsx:283), but this reads `["retrieval"]`. Embedding settings live in `[embeddings]` (with `[gemini]`/`[openai]` as fallbacks — see `buildEmbeddingsConfig`), and `embeddings` isn't in `PERSONA_SCOPED_KEYS` at all, so it can never be a persona override.

Concrete: a host with `[embeddings] provider = "openai"` in the global config and no `[retrieval]` block renders

```
embeddings   openai · text-embedding-3-small · 1536
from         built-in default (no key set)
```

which is exactly the misreport #483 asked to eliminate. Either point this at `["embeddings"]` (and accept it will only ever be `global`/`default`), or split the panel: keep a `retrieval` provenance next to a retrieval line and give embeddings its own.

src/tui/screens/PersonaDetail.tsx
```diff
@@ -202,18 +219,15 @@ export function PersonaDetailScreen(props: {
           <Detail
             lines={[
               ["binary", p.resolvedHarness?.path ?? "not found on PATH"],
-              [
-                "from",
-                "state.json harness_bins > config.toml [harnesses.<h>] bin > default",
-              ],
+              ["from", sourceLabel(p.configSources?.brain)],
```
andrewagrahamhodges: **Blocking — this labels the binary with the chain's provenance, and deletes a hard-won fact.**

The line removed here documented the real resolution order for the value printed immediately above it:

```
state.json harness_bins > config.toml [harnesses.<h>] bin > default
```

That's the #450 order, restated in `lib/harnessBinPath.ts`: `state.json` is consulted ahead of the bare-name default and *is not in config.toml at all*. `configSources.brain` is the provenance of `harnesses.chain`, a different key. So a host whose `claude` path is pinned in `state.json harness_bins` now reads `binary /weird/path` / `from built-in default (no key set)` — the screen actively denies the thing causing the user's problem, which is the precise failure mode #450 was about.

Suggestion: put the chain provenance on the `Brain` MenuItem description (it describes `chain`), and restore the resolution-order line under `binary`.

src/tui/App.tsx
```diff
@@ -696,18 +696,27 @@ export function App(props: AppProps): React.ReactElement {
...
+              await refresh();
+              setPersonaName(answers.name);
+              setNotice(
+                `created ${host.personasDir}/${answers.name} · identity.json · config.toml`,
+              );
+              // A finished wizard is not a place to go back to.
+              navRef.current = [];
+              setScreen("chat");
+            } catch (e) {
+              setNotice(`could not create ${answers.name}: ${(e as Error).message}`);
```
andrewagrahamhodges, line 718: Non-blocking, but it undercuts the PR's own bullet ("reports creation failures inline"). `createPhantomFromWizard` passes `err: { write: () => {} }` into `runPersonaNew`, so every diagnostic it writes is discarded and the only message left is the generic `could not create persona 'x'`. The notice therefore renders as:

```
could not create alice: could not create persona 'alice'
```

Capture the sink and fold it into the thrown error:

```ts
let stderr = "";
const code = await runPersonaNew({ ..., err: { write: (c) => { stderr += c; } } });
if (code !== 0) throw new Error(stderr.trim() || `could not create persona '${answers.name}'`);
```

Then "name must be lowercase..." / "'alice' already exists" actually reaches the user.

src/tui/index.tsx
```diff
@@ -209,6 +214,15 @@ export async function createPhantomFromWizard(
     err: { write: () => {} },
   });
   if (code !== 0) throw new Error(`could not create persona '${answers.name}'`);
+
+  // The review screen promises a persona-local config file. Record the chosen
+  // brain there so the daemon reads the same choice the user just reviewed;
+  // an empty file would merely inherit the host layer and make the choice a lie.
+  const host = await loadConfig();
+  await updateConfigToml(
```
andrewagrahamhodges, line 239: Three notes on this write, none blocking on their own:

1. **Contract.** AGENTS.md persona invariant 2 says a per-persona setting is written through `resolvePersonaWriteTarget()`. This goes straight to `personaConfigPath`. I think the deviation is *correct* here — a brand-new persona has no file yet, so the helper would send the write to the GLOBAL config and set the chain for every persona on the box — but that reasoning needs to be in the comment (or carved out in AGENTS.md), otherwise the next reader "fixes" it.

2. **Chain collapse.** `setIn(..., ["harnesses","chain"], [answers.brain])` replaces the array wholesale, and `mergeToml` replaces arrays rather than merging. A host with a global `chain = ["claude", "pi"]` gives every TUI-created persona a single-element chain and no failover. Intended? `[brain, ...rest.filter(h => h !== brain)]` keeps the wizard's choice as head without silently removing failover.

3. **Partial create.** If this write throws, the persona directory already exists. Pressing Enter again on the review screen re-runs `runPersonaNew`, which now exits 2 with "already exists", so the wizard can never complete — and on first run `esc` has nowhere to go. Cheapest fix is making the create step idempotent (skip `runPersonaNew` when `listPersonaDirs` already has the name, and only do the config write).

AGENTS.md
```diff
@@ -460,6 +460,8 @@ chunks, and uses explicit chunk/file accounting.
 
 ## Process for updating this file
 
+The TUI creation wizard is transactional from the user's perspective: validate inline, show a review that explicitly says nothing has been written, then create and name the persona directory, `identity.json`, and persona-local `config.toml`. Settings snapshots must preserve config provenance; absence is inheritance, while an explicitly stated empty/tombstone value is still a persona override.
```
andrewagrahamhodges: Nit: this paragraph landed inside `## Process for updating this file`, between the heading and its numbered steps — that section is about *how to edit AGENTS.md*, not about the TUI. It belongs with the persona-config / TUI material above (near the persona invariants), where someone touching the wizard will actually find it.

Also, while I'm here: `sourceOf` only looks at the two TOML layers, but `config.ts` applies env vars *after* the merge, and vault secrets are injected into `process.env`. A value coming from `PHANTOMBOT_*` will be labelled "inherited from global config" or "built-in default". Worth one sentence in the contract saying provenance is stated for the config layers only.

andrewagrahamhodges (changes requested):
Requesting changes. The shape of this is right — the review step genuinely writes nothing until the last Enter, `sourceOf` is a clean testable seam, and I verified the new tests are real (6/6 in `tui-app-first-run`, 2/2 in `tui-config-source`, tsc clean, full suite matches main modulo one pre-existing `harnesses-claude` failure that reproduces on `main` here too).

Three things block:

**1. The new duplicate check dead-ends the resume flow.** `resolveOpeningScreen` sends an INCOMPLETE existing persona into the wizard with `initial.name = <that persona>` and `existingNames` containing it. Press `esc` from the brain step and you land on the name field, where Enter is now rejected forever with "A persona named 'alice' already exists" — and on that path `onBack` is undefined, so the footer offers only `^q`. The file's own header says this flow "must never dead-end (issue #471)". Proven by driving the real `<App>`:

```tsx
render(<App host={{...HOST, personas: [alice]}} startPersona="alice" wizardStartAt="brain" .../>);
stdin.write(ESC); stdin.write("\r");
// frame: "✗ A persona named 'alice' already exists."  footer: "↵ Continue  ^q Quit"
```

Fix is one clause: skip the duplicate check when `name === props.initial?.name?.trim()`.

**2 and 3 are provenance labels pointing at the wrong key** — details inline. Both make the screen state something false, which is worse than the "from" line it replaced.

Non-blocking notes are inline too (tautological error text, partial-create retry, the `resolvePersonaWriteTarget` contract, chain collapse, env layer, AGENTS.md placement).

kaieriksen:
Fixed the three blocking findings in 8b0744a: resumed personas may reuse their seeded name (with a real App regression), embedding provenance now reads `[embeddings]`, and brain chain provenance is separated from the restored binary-resolution order. Local typecheck and full suite pass: 4,020 passed, 2 platform skips, 0 failures. CI is running.

andrewagrahamhodges (changes requested):
Re-reviewed at `8b0744a`. **All three blockers are genuinely fixed in the code** — I verified each one on the branch. But **neither new regression test actually covers the bug it names**; I proved both by mutation. That matters more than usual here, because this test file's own header says these bugs "typecheck, render, and pass every unit test around them" — a no-op regression test is the exact failure mode it was written to prevent.

### Blocker 1 — resume dead-end: **fix confirmed, test is a no-op**

The `name !== props.initial?.name?.trim()` exemption is correct. I drove `<WizardScreen startAt="name" initial={{name:"alice"}} existingNames={["alice","bob"]}>` with a real `\r`:

- with the fix → advances to `step 2 of 6 · brain` ✅
- with the fix reverted → `✗ A persona named 'alice' already exists.`, stuck ✅

So the behaviour is right. But `tests/tui-app-first-run.test.tsx` mounts against `HOST.personas = []`, and `<App>` passes `existingNames={host.personas.map(p => p.name)}` — so `existingNames` is `[]` and the duplicate branch never evaluates at all. Mutation check:

```
$ perl -0pi -e 's/name !== props\.initial\?\.name\?\.trim\(\) &&\s*\n\s*props\.existingNames/props.existingNames/g' src/tui/screens/Wizard.tsx
$ bun test tests/tui-app-first-run.test.tsx
 7 pass  0 fail
```

Green with the bug reintroduced. Please add `alice` to `HOST.personas` (or pass `existingNames` explicitly) so the test fails without the fix.

### Blocker 2 — embeddings provenance: **fix confirmed, test is a no-op**

`sourceOf(personaToml, globalToml, ["embeddings"])` now matches where the value is actually read: `readMemory` → `embeddingSpaceForConfig(config)` → `config.embeddings`. Renaming the snapshot field `memory` → `embeddings` is the right call too, since it makes the mismatch impossible to reintroduce silently in the label.

The added test asserts `sourceOf({}, global, ["embeddings"]) === "global"` — but `sourceOf` is generic and was never the bug; the bug was **which key the snapshot hands it**. Mutation check: I reverted `snapshot.ts` to `["retrieval"]` (keeping the field name) and ran the whole suite — 4019 pass, no new failure. The test that would catch it asserts on `personaSnapshot(...).configSources.embeddings` with `[embeddings]` set in one toml and `[retrieval]` in the other.

### Blocker 3 — brain provenance: **resolved**

Restoring the precedence line on `binary` and moving `sourceLabel(configSources.brain)` onto the Brain menu-item description is the right split: the chain label now describes the chain (which really does come from `harnesses.chain`), and the binary line no longer claims a provenance it can't know.

Non-blocking, and pre-existing rather than yours: that restored string is itself wrong. `src/config.ts:1335-1340` resolves the bin as `env → config.toml [harnesses.<h>].bin → state.harness_bins → bare name`, and `src/lib/harnessBinPath.ts:7` documents exactly that order. The line inverts the middle two and omits env. Since you're touching this line anyway, worth correcting to `env > config.toml [harnesses.<h>] bin > state.json harness_bins > default`.

### Still open from round 1 (all non-blocking, none addressed)

Tautological inline error (`err` sink discarded), partial-create making retry impossible, the config write bypassing `resolvePersonaWriteTarget()`, and `chain = [brain]` silently disabling failover for every TUI-created persona. Fine to leave, but say so explicitly rather than let them drop.

### Verification

`bunx tsc --noEmit` clean. `tests/tui-config-source.test.ts` + `tests/tui-app-first-run.test.tsx` 10/10 in isolation. Full suite 4019 pass / 2 skip; `harnesses-claude … --strict-mcp-config (#338)` fails and reproduces on main, and `first run > reviews exact artifacts` failed only in the full-suite run and passes in isolation — an ordering flake, not yours.

Requesting changes on the two test gaps only. The code is right; make the tests prove it.

kaieriksen:
Resolved the re-review blockers and remaining round-one notes in e592bc8. The resume and embeddings tests now fail under mutation; creation surfaces real stderr, retries idempotently after a partial create, preserves inherited harness fallbacks, documents the write-target exception and config-layer-only provenance, and shows the actual binary resolution order. Typecheck and full suite pass: 4,020 passed, 2 platform skips, 0 failures.

src/tui/index.tsx
```diff
@@ -198,17 +204,42 @@ export function installSignalExit(
...
+      // would land underneath the rendered frame.
+      out: { write: () => {} },
+      err: { write: (chunk) => void (stderr += chunk) },
+    });
+    if (code !== 0) {
+      throw new Error(
+        stderr.trim() || `could not create persona '${answers.name}'`,
+      );
+    }
+  }
```
andrewagrahamhodges, line 228: Non-blocking, but the skip is now doing double duty and one caller loses side effects.

`resolveOpeningScreen` sends an *incomplete existing* persona back through the wizard, so this branch is not only the partial-create retry I described — it is also the normal resume path. On that path `runPersonaNew` is skipped entirely, and with it `autostart: true`: the persona is never added to `autostart_personas` and never gets its heartbeat timer instance (#486). A user who resumes a half-built phantom, picks a brain and sees "created" still has a phantom that does not boot with the daemon.

Cheapest fix that keeps the idempotency: hoist the autostart registration out of the branch — `writeAutostartPersonas(target, [...(target.autostartPersonas ?? []), answers.name])` (it dedupes) plus the same best-effort `syncHeartbeatInstances`, run unconditionally.

Verified locally: `runPersonaNew` on an existing dir exits 2 ("already exists"), so the guard itself is right — it is only the side effects that need to survive it.

src/tui/App.tsx
```diff
@@ -696,18 +696,27 @@ export function App(props: AppProps): React.ReactElement {
...
-            await refresh();
-            setPersonaName(answers.name);
-            // A finished wizard is not a place to go back to.
-            navRef.current = [];
-            setScreen("chat");
+            try {
+              await props.onCreatePersona(answers);
+              await refresh();
+              setPersonaName(answers.name);
+              setNotice(
```
andrewagrahamhodges, line 711: Non-blocking copy nit, same root cause as the comment on `createPhantomFromWizard`: on the resume path nothing was created, but the notice still reads `created <dir> · identity.json · config.toml`. Worth branching the wording on whether the persona already existed ("updated <name> · config.toml"), otherwise the one screen that tells the user what happened is wrong exactly when they were repairing something.

andrewagrahamhodges (approved):
Approving. All three round-2 blockers are fixed, and this time both regression tests are real — I proved each by mutation rather than by reading them.

**Round 3 verification**

1. *Wizard resume.* `tests/tui-app-first-run.test.tsx` now mounts with `host.personas = [{ name: "alice", … }]`, so `existingNames` actually contains the seeded name and the duplicate branch evaluates. Removed the `name !== props.initial?.name?.trim()` guard from both call sites in `Wizard.tsx` → the test fails with `✗ A persona named 'alice' already exists.` in the frame. It exercises the call site now, not the helper.
2. *Embeddings provenance.* `configSourcesOf` is extracted and the test drives it with `persona = { retrieval: … }`, `global = { embeddings: … }`. Reverted `snapshot.ts` to `sourceOf(..., ["retrieval"])` → the test fails. That is the mutation that stayed green last round.
3. *Harness binary provenance.* `env > config.toml [harnesses.<h>] bin > state.json harness_bins > default` matches `config.ts:1335` / `harnessBinPath.ts:7`. This also corrected a stale note of my own — I had `state.json` first.

Full suite green on `e592bc8`: 4020 pass, 2 skip, 0 fail.

The follow-ups from my non-blocking notes all landed well: stderr from `runPersonaNew` is surfaced instead of swallowed, the chain write preserves inherited fallbacks with the chosen brain at the head, the `resolvePersonaWriteTarget()` deviation is now justified in the comment, and the AGENTS.md paragraph moved out of "Process for updating this file".

**Two non-blocking notes inline**, both from the idempotent-create guard I suggested turning out to cover a second caller I did not account for: on the resume path `runPersonaNew` is skipped, so `autostart: true` and the heartbeat-instance provisioning never run, and the success notice still claims it created `identity.json`. Neither blocks merge; worth a small commit here rather than a follow-up.

One observation, no action needed: writing `[brain, ...inherited]` into the persona file converts inheritance into a copy — `configSources.brain` becomes `persona`, and later edits to the global chain no longer reach it. That is the right trade for a chain whose head the user just chose explicitly, and it is what I asked for; just noting the semantics are now a snapshot, not a link.

PR #494, opened 2026-08-29
macOS: per-persona heartbeat LaunchAgents (multi-persona maintenance parity)

### LINKED ISSUE #491: macOS: per-persona heartbeat plists (multi-persona maintenance parity)
Follow-up from #486 / PR #490.

#486 gave systemd rigs one `phantombot-heartbeat@<persona>.timer` instance per served persona, and fixed the Windows heartbeat task to pass `--persona`. **macOS was deferred**: the launchd heartbeat is still a single `dev.phantombot.heartbeat` plist running bare `phantombot heartbeat`, which resolves the default persona — so a non-default persona on a macOS rig gets no heartbeat maintenance and no rollover nightly, exactly the pre-#486 blind spot.

Scope when picked up:
- One labelled plist per served persona (`dev.phantombot.heartbeat.<persona>.plist`, args `heartbeat --persona <persona>`), written/bootstrapped by `installPhantombotPlists` and torn down by uninstall.
- Reconcile on persona lifecycle like the systemd instances (picker / `persona new --autostart` / TUI action already call the sync seam — needs a launchd backend).
- Verify on a real Mac before calling it done — none exists in the fleet today, which is why it was deferred rather than shipped unverified.

Everything else from #486 is already platform-shared: per-persona fire-markers, the per-persona startup nightly, and doctor's `maintenance per persona` section (the marker + ledger halves work on macOS today).

lenaparkhodges:
Closes #491. Follow-up to #486 / #490.

## What

The launchd heartbeat was still a single `dev.phantombot.heartbeat` plist running bare `phantombot heartbeat`, which resolves the default persona — so non-default personas on a macOS rig got no heartbeat maintenance and no rollover nightly. This PR mirrors the systemd `phantombot-heartbeat@` template instances:

- **One plist per served persona** (`dev.phantombot.heartbeat.<persona>.plist`, args `heartbeat --persona <persona>`), written/bootstrapped by `install`, reconciled by a new `ensureLaunchdHeartbeatInstances()`, and torn down by `uninstall`.
- **Same migration rule as systemd**: the legacy single-persona plist is booted out and deleted only *after* the default persona's replacement is verifiably loaded — a failed bootstrap keeps the legacy heartbeat rather than leaving the host with none.
- **Persona lifecycle parity**: the picker / `persona new` / TUI autostart action now dispatch through a cross-platform `syncHeartbeatInstances()` in `platform.ts` (linux → systemd instances, darwin → launchd plists, elsewhere → null). The heartbeat's periodic heal reconciles launchd instances on darwin, so drift self-repairs within 30 minutes.
- Instance discovery is **file-based** (listing the LaunchAgents dir) rather than `launchctl list`, so unserved personas' plists are found even if they failed to bootstrap.

## Caveats

- **Not verified on a real Mac yet** — per the issue, this should be exercised on a real macOS runtime (Matt's rig) before calling #491 fully done. The logic is covered by fake-runner tests mirroring the systemd suite; the first real-world pass happens on the next `install`/heartbeat after update.
- `update`'s post-swap heal remains systemd-only; on macOS the periodic heartbeat heal picks up reconciliation within 30 minutes (main-plist refresh is unchanged, via `rerenderUnitIfStale`).

## Tests

+13 tests (per-persona generation, discovery, install + migration gating, ensure-reconciliation incl. stale-rewrite/unserved-removal/legacy-survives-failure, uninstall sweep, dispatcher). Full suite: 4025 pass / 0 fail, tsc clean. The darwin `runInstall` happy-path test was made hermetic (`personas` injected — it previously fell through to the developer's real config).


src/lib/launchd.ts
```diff
@@ -425,10 +527,162 @@ export async function uninstallPhantombotPlists(
...
+      "print",
+      `${opts.domain}/${label}`,
+    ]);
+    if (loaded.exitCode === 0 && !dirty) {
+      if (i === 0) defaultReady = true;
+      continue;
+    }
+    // Missing, stale, or not loaded: reload from disk.
+    await opts.launchctl.run(["bootout", `${opts.domain}/${label}`]);
+    const r = await opts.launchctl.run(["bootstrap", opts.domain, path]);
```
kaieriksen, line 673: **Major — a failed `bootout` can permanently mask a stale loaded job.** If this plist is dirty and `bootout` fails, `bootstrap` normally fails because the old label is still loaded. On the next heal the file now matches `expected`, `launchctl print` succeeds for the *old in-memory job*, and line 617 treats it as healthy forever. Check the bootout result and do not leave the rewritten file in a state that can be mistaken for a successful reload; add a regression covering bootout failure followed by another heal.

src/lib/launchd.ts
```diff
@@ -425,10 +527,162 @@ export async function uninstallPhantombotPlists(
...
+  }
+
+  // Retire plists whose persona is no longer served.
+  const served = new Set(opts.personas);
+  for (const persona of listHeartbeatInstancePlists(dir)) {
+    if (served.has(persona)) continue;
+    const label = heartbeatPlistLabelFor(persona);
+    await opts.launchctl.run(["bootout", `${opts.domain}/${label}`]);
+    const p = heartbeatPlistPathFor(persona, dir);
+    if (existsSync(p)) await unlink(p);
```
kaieriksen: **Major — do not delete the only retry handle when `bootout` fails.** This removes an unserved plist regardless of whether launchd actually unloaded it. A transient failure therefore leaves the old heartbeat registered but removes the file discovery relies on, so later heals cannot find/retry it. The legacy retirement paths have the same issue (lines 450–451 and 643–648): only unlink/report retired after confirmed unload (or confirmed absence), and test the failure case.

README.md
```diff
@@ -2534,6 +2534,13 @@ Installed user units:
 | `phantombot-tick.timer` | Every minute | Scheduled task runner |
 | `phantombot-heartbeat@<persona>.timer` | Every 30 minutes | Per-persona mechanical maintenance + fires that persona's nightly sweep on day rollover (one instance per served persona; the legacy single `phantombot-heartbeat.timer` is retired automatically) |
 
+The macOS equivalents are per-user LaunchAgents: `dev.phantombot.phantombot`
```
kaieriksen, line 2568: **Major — update the required contributor architecture contract too.** Root `AGENTS.md` says every user-facing service-shape/architecture change must update both README and AGENTS in the same PR. Add the macOS per-persona LaunchAgent model and migration invariant there so future service work preserves it.

kaieriksen (changes requested):
The per-persona shape is directionally right, but three reliability/documentation issues still block shipping on macOS. Focused tests and typecheck pass; full suite is running.

src/lib/launchd.ts
lenaparkhodges: Fixed in 79fdb58. The heal now prints the job BEFORE any write: a stale+loaded plist is booted out first, and only a successful bootout lets the new body reach disk. On bootout failure the label is recorded in `reloadFailed`, the old body stays (it still matches the loaded job, so disk and launchd stay consistent), and the next heal retries the reload. Regression test drives bootout failure followed by a second heal and asserts both the no-mask state and the retry.

src/lib/launchd.ts
lenaparkhodges: Fixed in 79fdb58. All retirement paths (unserved personas, legacy heartbeat in the heal, legacy migration + old-nightly cleanup in install) now gate the unlink behind `confirmedUnload()` — bootout success, or a `print` probe verifying the job is absent. A transient bootout failure keeps the plist (the retry handle) and reports `removeFailed` instead of claiming success. Regression tests cover bootout-fails-and-still-loaded (file kept) and bootout-fails-but-absent (file removed) for the removal path, plus the legacy and nightly retirement in install.

README.md
lenaparkhodges: Added in 79fdb58 — AGENTS.md invariant 36: the per-persona LaunchAgent model (one `dev.phantombot.heartbeat.<persona>` per served persona, default first, legacy retired only after the replacement is verifiably loaded) and the load-bearing sequencing rule — launchd state is never updated ahead of a confirmed unload, with the failure each shortcut re-opens spelled out.

lenaparkhodges:
Rebased onto main (conflict was AGENTS.md invariant 36 — the new slash-command invariant kept its number, the LaunchAgents one renumbered to 37). All three review threads were already resolved by 79fdb58 / the AGENTS.md section, both carried through the rebase. tsc clean, launchd/platform tests pass, CI 6/6. Ready for re-review.

kaieriksen (approved):
Re-review at 947f158: all three Major findings are fixed. Stale loaded jobs are now unloaded before plist rewrites; every retirement path preserves its retry plist unless unload is confirmed; and AGENTS.md documents the per-persona launchd sequencing invariant. Focused 77 tests, typecheck, full 4,116-test suite (2 platform skips), production build, and hosted CI all pass. Real macOS launchd behavior remains covered by injected-runner tests rather than a macOS host.

PR #496, opened 2026-08-29
Serialise memory-DB snapshots across personas (#495)

### LINKED ISSUE #495: Concurrent per-persona nightly sweeps race on the shared memory DB snapshot
Since #490 (per-persona `phantombot-heartbeat@.service` instances), a multi-persona host runs one nightly sweep per served persona. On kw-phantombot (lena + kai under one UNIX user, one shared `memory.sqlite`) the two sweeps fire within ~60ms of each other and race on the DB snapshot:

```
14:43:18.217 [83185] dbBackup: snapshot taken .../memory.20260829T144317Z.sqlite (26776 KB)
14:43:18.293 [83182] nightly: snapshot failed  error: "disk I/O error"
```

Reproduces on every rollover pair today (13:24, 13:36, 14:41, 14:43) — always one success, one `disk I/O error`.

Impact is currently cosmetic-but-noisy: one snapshot does land, and `memory backup --list` shows all 5 restore points `ok`. But the failure is recorded in the nightly ledger, so `phantombot doctor --persona lena` permanently reports `error: maintenance: snapshot: disk I/O error` on an otherwise healthy host — exactly the kind of false red that trains people to ignore doctor.

Suggested fix: the memory-DB snapshot is a HOST-level concern, not a per-persona one. Either (a) take it once per host per rollover behind an advisory lock / marker file, or (b) have the losing sweep detect a snapshot taken within the last N seconds and skip rather than fail. Also worth not writing a transient snapshot failure into the persona's nightly ledger as a sticky error.

Observed on v1.1.320, preview channel.

robertclawson:
Fixes #495.

## The bug

Since the per-persona nightly instances landed (#490), every persona on a multi-persona host sweeps at the same day rollover against the same shared `memory.sqlite`. The snapshot filename is stamped to the **second**, so two sweeps firing milliseconds apart resolve to the identical destination and the loser `rm`s the file the winner is still writing.

Observed on kw-phantombot after v1.1.320: one sweep succeeds, the other logs `disk I/O error`, and the failure is pinned in that persona's nightly ledger — so `doctor --persona lena` reports a permanent error on a host whose backups are in fact fine (all 5 restore points valid).

## The fix

The snapshot is a **host** concern, not a per-persona one:

- `backupMemoryDb` serialises on a lock file beside the restore points, created with `open(..., "wx")` so exactly one process can hold it. A lock older than 10 minutes is treated as abandoned and cleared, so a holder that dies mid-VACUUM cannot block snapshots forever.
- Callers that share a database may pass `coalesceMs`: a restore point younger than the window (or a sibling currently holding the lock) is accepted as covering this run, returning the new `fresh` status. The nightly passes `SNAPSHOT_COALESCE_MS` (5 min) — long enough to close the race, far short of a day, so the daily restore point is never thinned.
- `fresh` is not an error and is not written to the nightly ledger.
- `phantombot memory backup` deliberately does **not** coalesce: an operator who typed the command gets a snapshot. It still honours the lock rather than deleting an in-flight file.

## Tests

`tests/memory-dbBackup-race.test.ts` (new) reproduces the exact failure shape — same database, same second, two concurrent sweeps — and asserts one complete snapshot and no error. Plus window boundaries, the explicit-CLI path, a live lock, and a stale lock. `tests/lib-nightlyMaintenance.test.ts` gains the ledger assertion for a sibling sweep.

Local: typecheck clean, 10/10 in the two touched suites.

### DIFF
--- a/src/lib/nightlyMaintenance.ts
+++ b/src/lib/nightlyMaintenance.ts
@@ -29,6 +29,7 @@ import type { WriteSink } from "./io.ts";
 import { log } from "./logger.ts";
 import {
   backupMemoryDb,
+  SNAPSHOT_COALESCE_MS,
   type BackupResult,
 } from "../memory/dbBackup.ts";
 import { openDrawerStore } from "../memory/drawerSync.ts";
@@ -66,6 +67,8 @@ export async function runMemoryMaintenance(input: {
   personaDir?: string;
   keep?: number;
   now?: Date;
+  /** Snapshot coalescing window (#495). Defaults to `SNAPSHOT_COALESCE_MS`. */
+  coalesceMs?: number;
   out?: WriteSink;
 }): Promise<MaintenanceResult> {
   const now = input.now ?? new Date();
@@ -149,6 +152,12 @@ export async function runMemoryMaintenance(input: {
       dbPath: input.dbPath,
       keep: input.keep,
       now,
+      // The database is shared by every persona on the host, and since the
+      // per-persona nightly instances (#490) the siblings roll over together.
+      // One restore point per rollover is the whole point of a restore point,
+      // so a sweep that finds a fresh one — or finds a sibling mid-snapshot —
+      // stands down instead of racing it (#495).
+      coalesceMs: input.coalesceMs ?? SNAPSHOT_COALESCE_MS,
     });
     result.backup = backup;
     if (backup.status === "taken") {
@@ -159,6 +168,15 @@ export async function runMemoryMaintenance(input: {
             ? `, ${backup.pruned.length} older point(s) rotated out`
             : "") +
           `\n`,
+      );
+    } else if (backup.status === "fresh") {
+      // Not an error and deliberately not in `errors`: a sibling sweep took
+      // the snapshot seconds ago, so this host IS backed up. Writing it to the
+      // ledger would leave `doctor` permanently red on a healthy box, which is
+      // how operators learn to stop reading doctor at all.
+      write(
+        `nightly: memory snapshot skipped — a restore point from this ` +
+          `rollover already exists${backup.path ? ` (${backup.path})` : ""}\n`,
       );
     } else if (backup.status === "refused") {
       // Loud on purpose: an unhealthy memory database is the one condition
--- a/src/lib/nightlyMaintenance.ts
+++ b/src/lib/nightlyMaintenance.ts
@@ -174,9 +174,16 @@ export async function runMemoryMaintenance(input: {
       // the snapshot seconds ago, so this host IS backed up. Writing it to the
       // ledger would leave `doctor` permanently red on a healthy box, which is
       // how operators learn to stop reading doctor at all.
+      // Two different shapes share `fresh`: a sibling's restore point already
+      // covers this rollover (we have a path), or the lock stayed held and
+      // nothing was taken at all. Say which — an operator reading the ledger
+      // should not have to guess whether a snapshot exists.
       write(
-        `nightly: memory snapshot skipped — a restore point from this ` +
-          `rollover already exists${backup.path ? ` (${backup.path})` : ""}\n`,
+        backup.path
+          ? `nightly: memory snapshot skipped — a restore point from this ` +
+              `rollover already exists (${backup.path})\n`
+          : `nightly: memory snapshot skipped — another snapshot is in ` +
+              `flight, nothing taken this run\n`,
       );
     } else if (backup.status === "refused") {
       // Loud on purpose: an unhealthy memory database is the one condition
--- a/src/memory/dbBackup.ts
+++ b/src/memory/dbBackup.ts
@@ -52,12 +52,14 @@ import {
   copyFile,
   mkdir,
   open as openFile,
+  readFile,
   readdir,
   rename,
   rm,
   stat,
   unlink,
 } from "node:fs/promises";
+import { randomUUID } from "node:crypto";
 import { basename, dirname, join } from "node:path";
 import { Database } from "bun:sqlite";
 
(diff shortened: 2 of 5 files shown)


src/memory/dbBackup.ts
```diff
@@ -66,10 +78,83 @@ const SIDECARS = ["-wal", "-shm"] as const;
...
+): Promise<(() => Promise<void>) | null> {
+  const path = lockPath(dbPath);
+  const deadline = Date.now() + waitMs;
+  for (;;) {
+    try {
+      const handle = await openFile(path, "wx");
+      await handle.writeFile(`${process.pid} ${new Date().toISOString()}\n`);
+      await handle.close();
+      return async () => {
+        await rm(path, { force: true });
```
lenaparkhodges, line 159: The release deletes unconditionally, so a holder can delete a lock it no longer owns. Scenario: A acquires and its VACUUM runs long (>10 min on a big DB / slow disk); its lock goes stale. B arrives, sees `age > LOCK_STALE_MS`, clears A's lock, and acquires. A finishes and calls `release()` — which rm's **B's** lock. Now C can acquire while B is mid-snapshot, and we're back to the exact race this PR fixes.

Fix is small: write a unique token into the lock (`${process.pid}:${crypto.randomUUID()}`), capture it in the release closure, and have release re-read the file and only rm when the content still matches its own token. That makes stale-clear + slow-holder safe without weakening the abandoned-lock recovery.

tests/memory-dbBackup-race.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  test("a stale lock from a dead holder is cleared rather than blocking forever", async () => {
+    const dir = await workdir();
+    const dbPath = join(dir, "memory.sqlite");
+    seed(dbPath, 50);
+    const lock = join(backupDir(dbPath), ".snapshot.lock");
+    await mkdir(backupDir(dbPath), { recursive: true });
+    await writeFile(lock, "999999 old\n");
+    // Backdate it past the stale threshold — a holder that died mid-VACUUM.
+    const old = new Date(Date.now() - 30 * 60_000);
+    await Bun.spawn(["touch", "-d", old.toISOString(), lock]).exited;
```
lenaparkhodges: This breaks on real Windows dev machines: `touch -d` is a coreutils binary that exists on the CI runner (Git Bash on PATH) but not reliably on a stock Windows box without it, and the repo explicitly supports Windows dev (Megan, Windows heartbeat task).

Portable alternative with no subprocess:

```ts
await utimes(lock, old, old); // node:fs/promises
```

src/lib/nightlyMaintenance.ts
```diff
@@ -160,6 +169,15 @@ export async function runMemoryMaintenance(input: {
...
           `\n`,
       );
+    } else if (backup.status === "fresh") {
+      // Not an error and deliberately not in `errors`: a sibling sweep took
+      // the snapshot seconds ago, so this host IS backed up. Writing it to the
+      // ledger would leave `doctor` permanently red on a healthy box, which is
+      // how operators learn to stop reading doctor at all.
+      write(
+        `nightly: memory snapshot skipped — a restore point from this ` +
+          `rollover already exists${backup.path ? ` (${backup.path})` : ""}\n`,
```
lenaparkhodges: Minor, non-blocking: `fresh` also covers the lock-timeout path (no restore point taken at all), but this message asserts a restore point from this rollover already exists. When `backup.path` is undefined (lock held, no coalesced point) it would be more honest to say "another snapshot is in flight — nothing taken". Cosmetic; happy either way.

lenaparkhodges (changes requested):
Solid fix for a real race, and the test suite covers exactly the failure shape. One correctness hole in the lock protocol I want closed before merge, plus two small ones.

src/memory/dbBackup.ts
robertclawson: Real bug, fixed in f017c1d. The lock now carries a `${process.pid}:${randomUUID()}` token; `release()` re-reads the file and only `rm`s when the content still starts with its own token, otherwise it logs `snapshot lock was taken over; not releasing` and does nothing. Stale-clear recovery is untouched.

New test covers the takeover path directly (`acquireSnapshotLock` is now exported for tests — the sequence A-acquires / A-goes-stale / B-takes-over / A-releases cannot be provoked deterministically through `backupMemoryDb`).

tests/memory-dbBackup-race.test.ts
robertclawson: Fixed — `await utimes(lock, old, old)` from `node:fs/promises`, no subprocess. Applied to both stale-lock tests.

src/lib/nightlyMaintenance.ts
robertclawson: Taken. The two shapes are now separate lines: with a path, "a restore point from this rollover already exists (<path>)"; without one, "another snapshot is in flight, nothing taken this run".

lenaparkhodges (approved):
Re-reviewed f017c1d — all three threads addressed, approve.

- **Token-gated release:** correct. Lock carries a `pid:uuid` token; `release()` re-reads the file and only `rm`s when it still starts with its own token, logging `snapshot lock was taken over; not releasing` otherwise. The stale-clear recovery path is untouched, so the original race stays fixed and the takeover hole is closed. The residual read-then-rm window is microseconds and only reachable via the same stale path that already rm's first — acceptable.
- **utimes backdating:** portable, no subprocess. Good call on doing it in both stale-lock tests.
- **Ledger split:** the two shapes are now honest — restore point exists vs. snapshot in flight, nothing taken.

New takeover test covers the exact scenario from review; CI 6/6 green including test-windows.

PR #498, opened 2026-08-29
Finish resumed persona setup after #493

kaieriksen:
Follow-up for the two approved review notes on #493 that GitHub merged before the fix commit landed.

- preserve autostart registration and heartbeat provisioning when resuming an incomplete persona
- report resumed personas as updated rather than claiming identity files were created
- add focused runtime-boundary and rendered-copy regressions

Verification: focused 9 tests pass; typecheck passes; full suite 4,083 pass / 2 platform skips with one known timing-only wizard test that passed alone.

### DIFF
--- a/src/tui/index.tsx
+++ b/src/tui/index.tsx
@@ -24,7 +24,11 @@ import { setLogSink } from "../lib/logSink.ts";
 import { hostSnapshot } from "./snapshot.ts";
 import { openChat } from "./chatSession.ts";
 import type { WizardAnswers } from "./screens/Wizard.tsx";
-import { loadConfig, loadConfigForPersona } from "../config.ts";
+import {
+  loadConfig,
+  loadConfigForPersona,
+  servedPersonasOf,
+} from "../config.ts";
 import {
   personaCompleteness,
   type WizardStep,
@@ -33,7 +37,11 @@ import { runPersonaNew } from "../cli/persona-new.ts";
 import { log } from "../lib/logger.ts";
 import { personaConfigPath } from "../lib/personaConfig.ts";
 import { updateConfigToml, setIn } from "../lib/configWriter.ts";
-import { listPersonaDirs } from "../lib/personaDefault.ts";
+import {
+  listPersonaDirs,
+  writeAutostartPersonas,
+} from "../lib/personaDefault.ts";
+import { defaultSyncHeartbeatInstances } from "../lib/systemd.ts";
 
 /**
  * Decide what the app opens on.
@@ -81,7 +89,7 @@ export async function startTui(): Promise<number> {
       startPersona={opening.persona}
       wizardStartAt={opening.wizardStartAt}
       onCreatePersona={async (answers: WizardAnswers) => {
-        await createPhantomFromWizard(answers);
+        return await createPhantomFromWizard(answers);
       }}
     />
   );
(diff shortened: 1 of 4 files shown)


src/tui/index.tsx
```diff
@@ -227,6 +242,26 @@ export async function createPhantomFromWizard(
...
+  const autostartPersonas = await writeAutostartPersonas(target, [
+    ...(target.autostartPersonas ?? []),
+    answers.name,
+  ]);
+  try {
+    await syncHeartbeatInstances(
+      servedPersonasOf({
+        defaultPersona: answers.makeDefault
+          ? answers.name
+          : target.defaultPersona,
```
andrewagrahamhodges, line 254: Non-blocking, pre-existing (`persona new` computed the same thing), but this is now the code that owns it: `target` was loaded **before** `runPersonaNew`, and `adoptAsDefaultIfMissing` writes the adopted default to state.json without touching the in-memory config. So on the actual first-run path the served list is computed from the *stale* default.

Reproduced by driving this function against a temp root with an empty config.toml and no personas:

```
result { created: true } synced [ "phantom", "nova" ]
state { "default_persona": "nova" }
```

`phantom` has no persona dir. Best case `systemctl enable --now phantombot-heartbeat@phantom.timer` fails and the whole sync is swallowed by the catch, which also means `disableUnservedHeartbeatInstances` never runs; worst case it arms a timer for a persona that does not exist. Cheapest fix is to reload after creation (or thread `result.adoptedAsDefault` out of `runPersonaNew`) and compute `servedPersonasOf` from the config that is actually on disk.

src/tui/index.tsx
```diff
@@ -227,6 +242,26 @@ export async function createPhantomFromWizard(
...
   }
 
+  const autostartPersonas = await writeAutostartPersonas(target, [
+    ...(target.autostartPersonas ?? []),
+    answers.name,
+  ]);
+  try {
+    await syncHeartbeatInstances(
+      servedPersonasOf({
+        defaultPersona: answers.makeDefault
```
andrewagrahamhodges, line 252: Also non-blocking: the `answers.makeDefault` arm is unreachable today — `DEFAULT_ANSWERS.makeDefault` is `false` and nothing in `Wizard.tsx` ever sets it to `true`. Worth a note, because the moment it *is* settable this ternary becomes wrong on the resume path specifically: `runPersonaNew` is skipped when `created === false`, so nothing writes `default_persona`, yet the served list would be computed as if the switch had happened — and `disableUnservedHeartbeatInstances` would then tear down the real default's timer. If you keep the arm, it should be keyed on the default actually having been written, not on the answer.

andrewagrahamhodges (approved):
Approved. Both notes from #493 are properly fixed, and this time I checked the tests by mutation rather than by reading.

**What I verified**

- `tests/tui-create-persona.test.ts` is real: adding an early `if (!created) return { created };` before the `writeAutostartPersonas` block fails it (`autostart_personas` never written, `synced` undefined).
- The rendered-copy test is real: reverting the ternary in `App.tsx` back to the unconditional `created ...` string fails it on the exact frame text.
- The **created** path still behaves as before the move out of `runPersonaNew(autostart: true)`. Drove `createPhantomFromWizard` directly against a temp `PHANTOMBOT_CONFIG`/`PHANTOMBOT_STATE`/`PHANTOMBOT_PERSONAS_DIR`: a brand-new persona gets `autostart_personas = ["nova"]`, the persona dir, and a sync call. No regression there, though that path itself has no test now that the wizard no longer exercises `persona new`'s autostart branch.
- `writeAutostartPersonas` goes through `updateConfigToml`, so this merges rather than rewrites config.toml, and it dedupes — running the wizard twice on the same name cannot double-add.
- `log.warn` is the right sink here: `startTui` installs `setLogSink` before render, so the heartbeat warning lands in the `^l` buffer instead of on top of the frame.
- Typecheck clean. Full suite on 7df6eff: **4093 pass / 0 fail / 2 platform skips**.

Two non-blocking notes inline — neither is introduced by this PR, but this is now the code that owns the call.

PR #500, opened 2026-08-29
orchestrator: don't cool a harness for an empty reply (#499)

### LINKED ISSUE #499: Empty-reply fall-through cools the one healthy harness, serving recovery text from a dead fallback
Observed on kw-phantombot 2026-08-29 (Kai, chain [codex, pi], codex usage-limited):

1. **16:42** — pi produced a genuine empty `done` (model returned no text; direct API and one-shot repro both healthy — transient reasoning/empty-output flake).
2. The orchestrator fell through correctly, but called `cooldown.markFailure("pi")` on the empty-done path (src/orchestrator/fallback.ts).
3. **16:44** — next turn: pi is now in cooldown, so the chain skipped the one *healthy* harness, hit dead codex, and served recovery text.

An empty `done` means the harness process ran cleanly — it is a model-output flake, not a harness-health failure. Cooldown exists for auth/quota/capacity upstream failures; applying it here benches a working harness and can downgrade the next turn to a dead fallback.

**Proposed fix:** empty-done fall-through keeps the fall-through and the alerter incident accounting (`noteFailure`), but does NOT mark cooldown. A persistently-empty harness still escalates to a DEGRADED alert after 3 consecutive turns; a one-off flake costs one wasted round-trip instead of benching the healthy harness.

lenaparkhodges:
Closes #499.

## Problem

On kw-phantombot 2026-08-29 (Kai, chain `[codex, pi]`, codex usage-limited): pi produced one genuine empty `done` (transient model-output flake — direct API and one-shot repro both healthy). The orchestrator fell through correctly but called `cooldown.markFailure("pi")`, so the **next** turn skipped the one *healthy* harness, hit the dead primary, and served recovery text.

An empty `done` means the harness process ran cleanly — it is a model-output flake, not a harness-health failure. Cooldown exists for upstream auth/quota/capacity failures; applying it to empty output benches a working fallback behind a dead primary.

## Fix

Remove `cooldown.markFailure()` from the empty-done fall-through path in `runWithFallback`. The fall-through itself, the log line, and the alerter incident accounting (`noteFailure`) are unchanged, so:

- a one-off flake now costs one wasted round-trip instead of benching the healthy harness for 150s+
- a *persistently* empty harness still escalates to a DEGRADED push after 3 consecutive turns (`DEGRADE_AFTER_FAILURES`), since `noteSuccess` on a real reply still closes the incident

## Tests

- `empty done does NOT put the harness into cooldown (#499)` — store stays cold after an empty-done fall-through
- `a harness that emitted an empty reply last turn is tried again next turn (#499)` — regression for the incident: turn 1 pi flakes empty and claude covers; turn 2 pi is invoked first again and serves
- existing empty-done fall-through and cooldown-integration tests unchanged and green

(Note: `FakeHarness.script` became a public mutable field so the regression test can flip the harness from flaky to healthy between turns.)

Verified locally: fallback suite 34/34, `tsc --noEmit` clean. 9 unrelated lock-concurrency tests fail locally on a clean tree too (environmental — real process probes against the live phantombot on this host); CI is the source of truth for the full suite.

### DIFF
--- a/src/orchestrator/fallback.ts
+++ b/src/orchestrator/fallback.ts
@@ -351,7 +351,8 @@ export async function* runWithFallback(
           // bench the only healthy fallback while a dead primary gets
           // tried instead on the next turn. A persistently-empty harness
           // is still caught: noteFailure() keeps the alerter's incident
-          // counter going, so 3 consecutive empty turns fire a DEGRADED
+          // counter going, and the alerter classifies "empty reply" as its
+          // own `empty` cause, so 3 consecutive empty turns fire a DEGRADED
           // push. The cost of a flake is one wasted round-trip.
           alerter.noteFailure(harness.id, "empty reply");
           firstFailure ??= { harnessId: harness.id, error: "empty reply" };
--- a/src/orchestrator/fallback.ts
+++ b/src/orchestrator/fallback.ts
@@ -13,9 +13,12 @@
  * which has no payload ceiling).
  *
  * Cooldown semantics (see src/lib/cooldown.ts):
- *   - Each recoverable failure (recoverable error chunk OR empty done
- *     that triggers a non-last fall-through) bumps the harness's
- *     cooldown counter.
+ *   - Each recoverable failure (recoverable error chunk, or a harness
+ *     that throws) bumps the harness's cooldown counter. An empty `done`
+ *     does NOT: the process ran cleanly, so the empty output is a
+ *     model-output flake rather than a health signal — cooling the
+ *     harness for it benches the one working fallback behind a dead
+ *     primary (#499).
  *   - A successful turn (done with non-empty finalText, OR a "best we
  *     can do" empty-done from the last harness) clears the harness's
  *     counter.
@@ -340,7 +343,16 @@ export async function* runWithFallback(
             "orchestrator: harness produced empty reply, falling through",
             { harnessId: harness.id },
           );
-          cooldown.markFailure(harness.id);
+          // Deliberately NO cooldown.markFailure here (#499). An empty
+          // `done` means the harness process exited cleanly — the model
+          // just produced no text. That is usually a transient flake
+          // (reasoning model that spent its whole output budget), not a
+          // harness-health failure, and cooling the harness for it can
+          // bench the only healthy fallback while a dead primary gets
+          // tried instead on the next turn. A persistently-empty harness
+          // is still caught: noteFailure() keeps the alerter's incident
+          // counter going, so 3 consecutive empty turns fire a DEGRADED
+          // push. The cost of a flake is one wasted round-trip.
           alerter.noteFailure(harness.id, "empty reply");
           firstFailure ??= { harnessId: harness.id, error: "empty reply" };
           recoverableError = true;
--- a/src/lib/harnessAlert.ts
+++ b/src/lib/harnessAlert.ts
@@ -74,8 +74,20 @@ export const ALERT_SEND_TIMEOUT_MS = 20_000;
  * the reader to the journals to find out which (observed on
  * kw-phantombot 2026-08-20 — a 300s idle kill on a single-harness chain).
  * It changes no policy: like `other` it never fires the degraded alert.
+ *
+ * `empty` marks a harness that exited cleanly but produced no assistant
+ * text — the literal "empty reply" fallback.ts stamps (#499). One empty is
+ * a model-output flake and deliberately gets no cooldown, but a harness
+ * that returns empty on EVERY turn is just as dead as a dead token, so
+ * like `auth` it fires the degraded alert once its run crosses the
+ * threshold.
  */
-export type HarnessFailureCause = "auth" | "rate_limit" | "timeout" | "other";
+export type HarnessFailureCause =
+  | "auth"
+  | "rate_limit"
+  | "timeout"
+  | "empty"
+  | "other";
 
 /**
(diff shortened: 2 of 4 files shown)


src/orchestrator/fallback.ts
```diff
@@ -340,7 +343,16 @@ export async function* runWithFallback(
...
+          // Deliberately NO cooldown.markFailure here (#499). An empty
+          // `done` means the harness process exited cleanly — the model
+          // just produced no text. That is usually a transient flake
+          // (reasoning model that spent its whole output budget), not a
+          // harness-health failure, and cooling the harness for it can
+          // bench the only healthy fallback while a dead primary gets
+          // tried instead on the next turn. A persistently-empty harness
+          // is still caught: noteFailure() keeps the alerter's incident
+          // counter going, so 3 consecutive empty turns fire a DEGRADED
+          // push. The cost of a flake is one wasted round-trip.
```
robertclawson, line 356: This isn't accurate, and it's the claim the whole change leans on.

`alerter.noteFailure(harness.id, "empty reply")` → `classifyFailure("empty reply")` → `"other"`. `noteDegraded` returns early for anything that isn't `"auth"` (`src/lib/harnessAlert.ts:230`), so three consecutive empty turns fire **no** DEGRADED push. `tests/lib-harnessAlert.test.ts:209` asserts the `other` classification explicitly.

Net effect once cooldown is also removed: a persistently empty harness is retried first on every turn indefinitely and nothing ever surfaces it. Either add an `"empty"` cause that `noteDegraded` acts on, or `markFailure` after N consecutive empty dones (N=2 or 3) so the first flake stays free — then this comment becomes true.

robertclawson (changes requested):
Right fix, wrong safety net.

Not cooling on an empty `done` is correct — the process exited cleanly, so it's model output, not harness health. Diff is 3 lines of code plus two good tests, all 6 checks green at `374fc29`.

But the claim both the PR body and the new comment rest on — *"a persistently empty harness still escalates to a DEGRADED push after 3 consecutive turns"* — isn't true. `classifyFailure("empty reply")` returns `"other"`, and `noteDegraded` bails immediately on `if (state.cause !== "auth") return;` (`src/lib/harnessAlert.ts:230`). The repo's own test documents this: `tests/lib-harnessAlert.test.ts:209` — *"An empty reply … classifies as `other`"*.

So after this lands, a *permanently* empty harness has no mechanism at all: it is never cooled, so it's tried first on every single turn forever (one wasted spawn per turn), and no alert ever fires. That's exactly the silent-degradation shape `harnessAlert` exists to catch, and it's not hypothetical — we hit it today with a pi install whose models catalogue was empty, returning `content: null` on every turn. A persistent config fault, not a flake.

Suggested fix, small either way:

- count consecutive empty dones per harness and `markFailure` on the 2nd or 3rd, leaving the 1st free (keeps the whole point of #499), or
- add an `"empty"` cause and let `noteDegraded` fire on it.

And please correct the comment and PR body so they don't promise a net that isn't wired up. Happy to approve as soon as one of those lands.

src/orchestrator/fallback.ts
lenaparkhodges: Good catch — the claim was false: `classifyFailure("empty reply")` lands in `other`, and `noteDegraded` returns early for anything but `auth`, so persistent empties surfaced nothing. Fixed in a4cd9fa by taking your first option: `empty` is now its own `HarnessFailureCause` (`EMPTY_MARKERS = ["empty reply"]` in `classifyFailure`), and `noteDegraded` acts on `auth` and `empty` alike — one empty stays free (no cooldown, no alert), three consecutive empties fire the DEGRADED push with a message naming the cause. Tests cover: 1 empty silent, 3 empty \u2192 push, noteSuccess closes the incident, and the classification boundaries.

robertclawson (approved):
Re-reviewed at `a4cd9fa`. **The finding is fixed properly** — thank you for wiring the safety net rather than just softening the comment.

`empty` is now its own `HarnessFailureCause`, `classifyFailure` matches the literal `"empty reply"` stamp `fallback.ts` emits, and `noteDegraded` fires on it alongside `auth` with its own message line. That closes the gap exactly where it was: one empty stays free (the whole point of #499), but a harness that returns empty on every turn now crosses `DEGRADE_AFTER_FAILURES` and pushes a DEGRADED alert instead of being retried first, silently, forever. The comment in `fallback.ts` is now true.

Details I checked:
- Cause-change resets the run (`state.cause !== cause → consecutiveFailures = 0`), so an empty landing mid-auth-incident doesn't corrupt the auth tally — and the updated `mixed causes` test documents that new behaviour honestly rather than papering over it.
- `EMPTY_MARKERS` is narrow: `"claude exited with code 1: empty file"` still classifies `other`. Good test.
- The `flaky.script` reassignment in the fallback test is a genuine two-turn regression test for the kw-phantombot shape, not a restatement of the fix.

Local: typecheck clean, `tests/lib-harnessAlert.test.ts` + `tests/orchestrator-fallback.test.ts` **61 pass / 0 fail**. All CI green.

Non-blocking, out of scope — one shape still stays quiet: on a **single-harness chain** an empty `done` from the last harness is treated as "best we can do" success, so `noteSuccess` closes the incident and nothing ever alerts. That's pre-existing, not something this PR makes worse, but it is the literal Lena/pi failure we hit — worth a follow-up issue if you agree.

Approving.

PR #502, opened 2026-08-29
orchestrator: alert when a single-harness chain replies empty every turn (#501)

### LINKED ISSUE #501: Single-harness chain: an empty reply from the last harness never alerts
Follow-up to #499 / PR #500 (merged as 8d9da89). Non-blocking gap noted during review, filed so it isn't lost.

## What #500 fixed
`empty` is now its own `HarnessFailureCause` (`EMPTY_MARKERS = ["empty reply"]` in `classifyFailure`), and `noteDegraded` acts on `auth` and `empty` alike. One empty reply is free (no cooldown, no alert); three consecutive empties fire the DEGRADED push.

## The remaining hole
That path only runs when the orchestrator treats the empty done as a *failure* and falls through to the next harness. On a **single-harness chain**, an empty done from the last harness is treated as "best we can do" success — `noteSuccess()` runs, which closes any open incident. So a harness that returns empty on *every* turn:

- never accumulates a failure count,
- never reaches the 3-consecutive threshold,
- never fires a DEGRADED push,

and the operator sees `(no reply)` on Telegram forever with no signal anywhere.

## Why it matters
This is exactly the failure shape hit on 2026-08-29 on kw-phantombot: after Lena's persona moved to a new UNIX user, pi's per-user model catalogue (`~/.pi/agent/models-store.json`) was empty, so pi sent a tiny output budget to the reasoning model, got `content: null` back, and phantombot reported `(no reply)`. Single harness, so nothing ever escalated — it took a human noticing silence.

## Suggested fix
Don't let the last-harness fallback launder an empty reply into a clean success. Either:

1. On a `done` with an empty body, call `noteFailure(..., cause: "empty")` **before** the "best we can do" return, and skip `noteSuccess()`; or
2. Track empties separately from the chain outcome so the consecutive-empty counter advances regardless of chain position.

Option 1 is smaller and reuses the machinery #500 just added. Either way it needs a test for the single-harness chain — the current suite only covers multi-harness fallthrough.

robertclawson:
Closes #501.

## The hole

#500 made `empty` its own `HarnessFailureCause` so three consecutive empty replies fire a DEGRADED push. That machinery only ran on the **fall-through** path — when the chain had somewhere else to go. On a **single-harness** chain the empty `done` from the last harness was reported as a plain success: `noteSuccess()` closed the incident every turn, the consecutive-empty counter never advanced past one, and a harness that returns no text on *every* turn alerted **never**.

The live shape (kw-phantombot, 2026-08-29): a persona moved to a new UNIX user, pi's per-user model catalogue (`~/.pi/agent/models-store.json`) came up empty, so pi sent a tiny output budget to the reasoning model and got `content: null` back on every turn. Single harness — nothing escalated, and a human noticing the silence was the entire detection mechanism.

## The fix

`src/orchestrator/fallback.ts` now tracks whether the `done` it served was empty. In the `succeeded` block:

- `cooldown.markSuccess()` still runs **unconditionally**. Cooldown decides what to *try*, and an empty done is not a reason to bench the only harness we have (#499). The alerter decides when to *speak*. Those stay separate concerns.
- when the served done was empty: `alerter.noteFailure(id, "empty reply")` instead of `noteSuccess()`, then `noteDegraded({ harnessId })`.

`noteDegraded`'s `servedBy` is now optional. Nobody covered this turn, and an alert reading `pi serving` would tell the owner they are still being answered — the opposite of the truth. With no `servedBy` the line renders `no fallback left`:

```
🔇 pi empty reply ×3 🖥 kw-phantombot · no fallback left, harness returns no text
```

Below `DEGRADE_AFTER_FAILURES` the call is a no-op, so a one-off empty is exactly as quiet as #499 wants, and a real reply afterwards closes the incident as usual.

The reply path is unchanged: the empty `done` is still yielded so the channel surfaces `(no reply)`.

## Tests

`tests/orchestrator-fallback.test.ts`
- single-harness chain empty on every turn → exactly one alert at the threshold, containing `empty reply`, `×3` and `no fallback left`, and **not** `serving`; the empty `done` still reaches the channel
- one empty stays silent, and a real reply resets the run so 2 more empties still do not alert
- an empty from the last harness does not cool it (#499 regression guard, now over the new code path)

`tests/lib-harnessAlert.test.ts`
- `noteDegraded` with no `servedBy` renders `no fallback left` and never `serving`
- below the threshold an uncovered empty stays silent

Local: typecheck clean, 66 pass / 0 fail on the two suites, full suite 4104 pass / 1 fail — that one (`cli-doctor-persona-env`, semantic-search line) fails identically on `main` in this checkout and is unrelated.


src/orchestrator/fallback.ts
```diff
@@ -416,8 +425,28 @@ export async function* runWithFallback(
...
+        // never, however long it went on. On a multi-harness chain the empty
+        // done is a fall-through and already counted upstream; here it is
+        // the end of the line, so count it in the same currency.
+        log.warn("orchestrator: chain exhausted with an empty reply", {
+          harnessId: harness.id,
+        });
+        alerter.noteFailure(harness.id, "empty reply");
+        // No `servedBy`: nobody covered this turn. Below the threshold this
+        // is a no-op, so a one-off empty stays as quiet as #499 wants.
+        await alerter.noteDegraded({ harnessId: harness.id });
```
kaieriksen, line 447: Verified: the last-harness empty path preserves cooldown success, keeps the alerter incident open, and correctly omits servedBy because no fallback delivered text. The threshold/reset regressions cover the intended #499/#501 split.

kaieriksen (approved):
No findings. Verified the single-harness empty-reply incident reaches the threshold without cooling the harness, resets after a real reply, and renders the uncovered state accurately. Focused 66 tests, typecheck, full 4,105-test suite (2 platform skips), production baseline build, and hosted CI all pass.

Issue #488, opened 2026-08-29
Proposal: make prompt ordering more cache-friendly across turns


Johnniedarkoo:
TL;DR: PhantomBot currently rebuilds changing context and a rolling history in a way that prevents inference backends from reusing much of their prompt cache. This experiment reorders the prompt and keeps history append-only for several turns before periodically rebasing it. In testing, that reduced evaluated input by ~74% and prompt-processing time by ~67%, while normal live use showed 4–8 warm turns between resets with no correctness issues.

I have a working implementation and tests on a fork https://github.com/Johnniedarkoo/phantombot/tree/feature/prompt-cache-epochs. 

This issue is to discuss if and how the idea is desirable for Phantombot, before I turn it into a draft PR.

---

I’ve been experimenting with reducing the time PhantomBot spends re-processing the same conversation context on every turn.

The basic issue is that even when most of a conversation is unchanged, small changes in how the prompt is assembled can prevent llama.cpp and other backends from reusing much of the existing prompt cache, leading to longer processing times.

The current prompt ordering is roughly like this:

### Before

1. System / persona prompt
    * identity/persona
    * tool instructions
    * security/trust rules
    * other stable system instructions
2. Dynamic memory/context material
    * durable facts
    * retrieved memories
    * daily recall/digests
    * channel/context-specific information
    * other per-turn additions
3. Rolling conversation history
    * previous user turns
    * previous assistant turns
4. Current user message

Or in summary, the chapters are:

1. SYSTEM
2. TURN CONTEXT
3. ROLLING HISTORY (15 turns)
4. CURRENT USER PROMPT

There are two issues limiting cache reuse:

1. Chapter 2 changes from turn to turn. Retrieved memories, daily context and other turn-specific material can differ, so the prompt may already stop matching the previous cached prompt before it even reaches the conversation history.
2. More importantly, the 15-turn rolling history throws out its oldest turn on every new turn. Because that removed turn sits at the top of the history, the whole history below it shifts. This means the history itself can never form a long stable prefix, despite almost all of those turns being unchanged.

I experimented with changing the ordering first:

1. SYSTEM
2. ROLLING HISTORY
3. TURN CONTEXT
4. CURRENT USER PROMPT

This already puts the changing retrieved context much later in the prompt, so the stable system prompt and conversation history have a better chance of being reused.

However, the rolling history itself still has the same problem: every turn removes the oldest turn from the top, which invalidates the history cache below it.

[...]

PR #503, opened 2026-08-29
Add opt-in prompt-cache epochs for self-hosted inference

Johnniedarkoo:
Addresses #488.

## Summary

This adds an opt-in `[prompt_cache]` mode that makes PhantomBot's serialized prompt more reusable across turns and retains completed turns in a bounded, disposable append-only epoch.

The feature is **off by default**:

```toml
[prompt_cache]
enabled = false
max_epoch_bytes = 80000
```

It is intended primarily for self-hosted/local inference such as llama.cpp or vLLM, where the operator controls the prefix/KV cache. Hosted-provider users should generally leave it disabled: hosted APIs manage caching differently, and an append-only epoch can lengthen billed input while PhantomBot cannot control provider-side cache behavior.

## Behavior and ownership

PhantomBot remains the owner of canonical history, retrieval, durable facts, daily recall, persona state, security policy, and restart reconstruction. Epoch state is in-process acceleration only; it contains no backend cache handles or durable conversation state.

With the feature enabled, volatile per-turn context is rendered after canonical history and completed context/user/assistant triples are appended within a bounded epoch. Rebase rebuilds from canonical PhantomBot history. With the feature disabled, the established payload path is unchanged.

Lost, corrupt, inconsistent, or failed cache state degrades safely: invalid state is discarded, preparation failure falls back to the feature-off path for that request, and completion/discard bookkeeping cannot turn a successful model response into a failed turn.

## Security / trust lifecycle

The reorder is justified only as a cache-prefix optimization. Prompt position is not treated as authority.

The cache crosses a cold boundary on:

* trusted ↔ untrusted transitions;
* held risky/untrusted requests;
* persona switches, including A → B → A;
* effective screening-state changes;
* effective tool/MCP-surface changes;
* system/policy prompt changes;
* restart-required security changes, because epoch state is process-local.

Retrieved memories, durable facts, daily recall, and historical snapshots remain context/data rather than a new instruction channel.

## Behavioral-equivalence evidence

Deterministic regression coverage compares cold/feature-off against genuinely warm epochs for:

* held untrusted requests: same hold response, with no capable harness invocation;
* durable-fact recall;
* retrieved-memory recall.

I also ran a bounded real-model A/B through the normal runTurn → PiHarness path against Qwen3.8-27B served locally by llama.cpp through an OpenAI-compatible endpoint:

* durable fact: 3 cold + 3 warm target runs, all returned Delft;
* retrieved memory: 3 cold + 3 warm target runs, all returned Amber Finch;
* every cold target had epochTurns = 0;
* every warm target had epochTurns = 1, containing the completed warm-up turn;
* no material behavior shift was observed in these acceptance cases.

The accepted A/B used the same zero-tool surface on both sides to isolate prompt-layout behavior. The evidence is documented in docs/prompt-cache-epoch-plan.md; it is not presented as general proof that model quality is unchanged.

## Performance evidence

From the controlled local experiment behind #488:

* evaluated input: 16,944 → 4,372 tokens median (~74% reduction);
* prompt evaluation: 14.64s → 4.76s median (~67% reduction);
* short-request wall time: 15.86s → 5.38s median (~66% reduction).

Natural local use also showed warm stretches of several turns between rebases. These gains depend on a backend that can actually reuse exact prefixes.

## Scope

This PR deliberately does not split TURN CONTEXT into volatile/non-volatile sub-blocks. It also does not add persistent Pi sessions, KV save/restore, backend-specific cache management, history summarization/trimming, or another memory store.

## Validation

Focused local validation after rebasing onto current main:

* prompt-cache/config suite: 116 passed, 0 failed;
* behavioral-equivalence suite: 3 passed, 0 failed;
* git diff --check: clean;
* default-off and feature-off legacy serialization tests pass.

A local Windows full-suite run still contains pre-existing environment/platform failures involving permissions/path assumptions, SQLite cleanup locks, and missing UI dependencies; no prompt-cache failures were identified in the focused suites.


kaieriksen:
I can review this PR while it is in Draft, but I will not submit an approval until it is marked Ready for review. Would you like me to continue with review feedback now, or hold off on further reviews until you are ready?

kaieriksen (commented):
Completed a full-diff review at 758daf9. No findings: the implementation remains limited to opt-in, default-off, in-process append-only prompt-cache epochs for cache-friendly serialization. It does not add hosted `prompt_cache_key` support, persistent provider sessions, backend KV handles, or backend-specific cache management; those remain outside this PR, with hosted stable keys tracked separately in #504. Focused cache/config tests (116), typecheck, and the full local suite passed after rerunning the nine PATH-only subprocess tests with Bun on PATH. I am intentionally withholding approval while the PR remains Draft, per the note above.

andrewagrahamhodges:
Reviewed Kai's pass at 758daf9 against what we agreed on #488.

The implementation is aligned. What I checked directly:

* **Gate.** `DEFAULT_PROMPT_CACHE` is `enabled: false`, `maxEpochBytes: 80000` (src/config.ts:656). Invalid or inconsistent epoch state throws out of `prepare()`, `turn.ts` catches it, discards, and serves the turn on the feature-off path. Degrades, does not fail.
* **Trust-state rebase triggers.** All three are wired, not just described: a held request invalidates at the hold boundary with reason `threat_hold` (turn.ts:408, before any retrieval); persona entry is observed for every turn including cache-disabled and no-history ones, so A to B to A cannot revive A's epoch (turn.ts:365); and `promptCacheSecurityFingerprint` folds trusted, effective screening, MCP mode and the tool allow-list into the rebase check (promptCache.ts:158-165). The system prompt is fingerprinted separately, so a policy or prompt change colds the epoch too.
* **Framing.** README and AGENTS.md both document the block as a self-hosted/local-inference optimization and tell hosted users to leave it off. That is the correct reading: the epoch mechanism only pays where the KV cache is yours.
* **Scope.** No volatile/non-volatile turn-context split, no hosted `prompt_cache_key`. Correct, and #504 stays separate.

Two notes for the record rather than blockers.

**On the review itself.** "No findings" is accurate on scope, but the scope exclusions were never the acceptance criteria. The #488 conditions were the default-off gate, safe degradation on lost or corrupt state, the local-versus-hosted docs framing, the three forced-rebase trust triggers, and behavioural-equivalence evidence. Kai, please restate the review against those five explicitly so the record shows they were checked, and confirm the equivalence tests were mutation-checked rather than just green. A test that passes with the behaviour reverted tells us nothing.

**On the evidence.** 3 cold plus 3 warm on one local model with a zero-tool surface is accepted as scoped, and I appreciate that the PR says so plainly instead of overclaiming. It is not general proof that quality is unchanged, and the docs should keep saying that.

Held for the Ready-for-review flip, as Kai noted.

kaieriksen:
Restating my review at `758daf9` against the five #488 acceptance criteria:

1. **Default-off gate:** verified. `DEFAULT_PROMPT_CACHE.enabled` is `false`; the feature must be explicitly enabled.
2. **Safe degradation:** verified. Missing, inconsistent, or failed epoch preparation falls back to the canonical feature-off request path rather than failing the turn.
3. **Local-versus-hosted framing:** verified. README and AGENTS.md scope this to self-hosted/local inference and direct hosted users to leave it disabled.
4. **Forced trust-state rebases:** verified in implementation and tests. Threat holds invalidate before capable execution; persona transitions are observed even through disabled/no-history turns so A → B → A cannot revive A; trusted/screening/MCP/tool-allow-list changes alter the security fingerprint and force a rebase. System-prompt/policy changes also rebase.
5. **Behavioral-equivalence evidence:** verified, including mutation testing. The clean cold-versus-warm suite passes 3/3. I then temporarily mutated the production request path to omit current turn context only when an epoch plan exists. The suite killed that mutation: both durable-fact and retrieved-memory warm cases failed while the cold cases retained the required context. After restoring production code, the suite returned to 3/3 passing.

The evidence remains deliberately scoped: one deterministic harness/local-model setup and a zero-tool surface. It supports the stated claim but is not general proof that model quality is unchanged.

No findings. I am still withholding approval until the PR is marked Ready for review.

kaieriksen:
Thanks for marking this Ready for review. GitHub currently reports merge conflicts, so please resolve those first. Once the conflict resolution is pushed, I’ll re-review the resulting full diff and approve if it remains correct, in scope, and compliant with the #488 acceptance criteria.

Johnniedarkoo:
Thanks, it's mergable now. The PR drifted from main due to the other PR's but it's just some agents.md wording and tick.ts invariant numbering.

kaieriksen (approved):
Re-reviewed at 845fb3f after the main merge. No findings: the only conflict resolution in AGENTS.md preserves both #505's new invariant and this PR's prompt-cache invariant, with numbering corrected; the tick.ts change is the clean upstream #505 addition. Focused prompt-cache/config tests pass (116/116), and all current Linux and Windows CI checks are green.

PR #506, opened 2026-08-30
Carry the owning persona into command tasks, and make the host default observable (#505)

### LINKED ISSUE #505: Default-persona resolution: tick drops PHANTOMBOT_PERSONA for --command tasks, heal keys on dir existence, doctor never validates the default
Three related defects surfaced by a real outage: email triage silently stopped for ~200 scheduled fires (2026-08-29 16:37 to 2026-08-30 10:52) while `task_runs` recorded `exit=0 ok` for every one of them.

**What happened.** A command-backed task (`~/bin/mail-poll.sh`, `--every 5m`) calls `phantombot mcp call mailspring ...`. After updating to 1.1.327 the command no longer received `PHANTOMBOT_PERSONA`, so the call resolved against the default persona. That default was `lena`, who had been migrated to another host, leaving an empty MCP registry behind. The call returned `no such MCP server: mailspring`, the poller treated it as "MCP down, skip quietly" and exited 0. Nothing anywhere reported a problem.

### 1. `tick` does not pass `PHANTOMBOT_PERSONA` to `--command` tasks

Command-backed tasks get a minimal environment. `PHANTOMBOT_PERSONA` is not in it, so any command that shells back into `phantombot` resolves against the default persona rather than the persona that owns the task. Tasks are per-persona, so this is a correctness bug, not just an ergonomics one: a `max` task can end up reading another persona's vault, memory or MCP registry.

Expected: `tick` exports the owning persona into the command environment, the same way it scopes an LLM turn.

Workaround in place locally: `export PHANTOMBOT_PERSONA=${PHANTOMBOT_PERSONA:-max}` at the top of the script.

### 2. `healDefaultPersonaIfBroken` keys on directory existence, not on "served"

The heal check tests `existsSync(<personas>/<name>)`. A persona that has been migrated away, disabled, or never finished setup still leaves a directory, so heal sees a healthy default and does nothing while everything behind it is empty. That is exactly what happened with `lena`.

Expected: heal should treat a default as broken when the persona is not actually served (no identity, not in the autostart/served set, empty or unloadable config), not merely when the folder is gone.

### 3. `doctor` never validates the resolved default persona

There is no command that answers "which persona am I when I omit `--persona`, and is that persona usable". One `doctor` line would have caught this in seconds.

Expected: `doctor` reports
- the resolved default persona,
- its provenance (`PHANTOMBOT_PERSONA` env / `state.json` / `config.toml` / built-in fallback),
- whether that persona is served,
- whether it has a loadable MCP registry, and warns when the registry is empty while other personas have one.

Worth noting for anyone debugging this: `config.toml`'s `default_persona` is not the lever. Resolution order is env > `state.json` > `config.toml`, so state always wins. The new TUI correctly writes `state.json`.

### Broader lesson

A poller whose "skip quietly" path is indistinguishable from "nothing new" is an invisible outage. Every early exit in a command-backed task should log its reason, and it would help if `task_runs` could carry a "did nothing, here is why" status distinct from plain success.

andrewagrahamhodges:
Closes #505.

## What broke

Email triage stopped for ~200 consecutive scheduled fires (2026-08-29 16:37 → 2026-08-30 10:52) while `task_runs` recorded `exit=0 ok` for every one of them. Three defects, each individually defensible, lined up:

- a command-backed poller shelled into `phantombot mcp call mailspring`,
- with no `PHANTOMBOT_PERSONA` in its environment it resolved against the host default,
- that default was `lena`, migrated to another host but with her directory left behind,
- so the call returned `no such MCP server: mailspring`, the poller read that as "MCP down, skip quietly", and exited 0.

Nothing anywhere reported a problem.

## 1. `tick` carries the owning persona (`src/cli/tick.ts`)

Command tasks get a deliberately minimal environment, and it was missing the one variable the shipped contract depends on. The documented pattern tells a poller to call back into `phantombot` (`ask`, `notify`, `mcp call`, `memory`), and every one of those resolves through `resolvePersona`: `--persona`, then `PHANTOMBOT_PERSONA`, then the **host default**. One tick process runs tasks for every persona, so with the variable absent all of them read the default persona's vault, memory and MCP registry.

Same cross-persona class as #452's secret leak, arriving through the env instead of the vault, and it fails quietly — reading the wrong persona succeeds. Set after the allowlist copy so the tick process's own ambient value can never outrank the task's owner.

## 2. Heal stops trusting a bare directory (`src/lib/personaDefault.ts`)

`existsSync(personaDir)` was the whole test, so a create that died after `mkdir`, or a migration that moved the contents out, left a husk that heal read as healthy. `defaultPersonaDefect` is now the shared predicate — directory present **and** at least one persona marker (`identity.json`, `config.toml`, `vault.sqlite`, `mcp.json`, `IDENTITY.md`, `SOUL.md`, `MEMORY.md`, `BOOT.md`, `memory/`, `kb/`). `adoptAsDefaultIfMissing` uses it too, so the two writers cannot disagree, and heal now prefers a candidate that is itself usable over the alphabetically-first name.

**Deliberately weaker than `personaCompleteness`, and worth arguing about.** Completeness is right for the TUI launch gate, but heal rewrites `state.json` unprompted, so its trigger may never be something that can be transiently false: `identity.json` is created **lazily** on the first vault open, so a persona created seconds ago legitimately has none, and completeness's brain probe reads the current PATH. Keying heal on either would let a fresh install repoint its brand-new default at some older persona — worse than the bug being fixed.

**Scope, stated plainly:** this does **not** catch `lena`. A persona migrated to another host but leaving its files behind is indistinguishable from a working one, and I would rather report that than have heal rewrite the host default on a guess. Detecting it is (3)'s job.

## 3. `doctor` answers "who am I when I omit `--persona`?" (`src/cli/doctor.ts`)

New first line: the resolved default, its **provenance**, whether it is usable, whether it is served, and how many MCP servers it has.

Provenance is load-bearing rather than decorative. Resolution is `PHANTOMBOT_DEFAULT_PERSONA` > `state.json` > `config.toml` > built-in `phantom`, so the layer an operator reaches for first is the one that loses; the human line says so explicitly when the answer came from state.

Exit-code policy is split on purpose:

| condition | result |
|---|---|
| no persona dir, or `mcp.json` will not parse | **exit 1** — every `--persona`-less command is aimed at it |
| empty-but-loadable registry while another persona has one | **WARN**, exit 0 |
| everything else | ok |

A host that registers no MCP servers at all is normal, and a health check that cries wolf is one nobody reads. The warn fires only in the exact shape of this outage.

## Verification

Live, on the compiled binary. Reproducing the outage state (`default_persona = "lena"` in an isolated `state.json`):

```
phantombot doctor — persona 'max'
  default persona: WARN — 'lena' (from state.json)
    usable, but no MCP servers registered while max has some — commands run without --persona will not see them
    note: state.json OUTRANKS config.toml's default_persona — editing the TOML key will not change this
```

and healthy:

```
  default persona: ok — 'max' (from state.json)
    usable; 2 MCP server(s); served
```

The tick fix was driven end to end through the compiled binary too: an isolated `XDG_*` root with a `kai`-owned command task, fired by a real `phantombot tick`, wrote back `[kai]`.

## Tests

Nine new tests across three files, each mutation-checked (removing the fix fails the test, and only that test):

- `tests/cli-tick.test.ts` — a command task is told which persona owns it, and an ambient `PHANTOMBOT_PERSONA=someoneelse` on the tick process does **not** win.
- `tests/lib-personaDefault.test.ts` — husk vs populated vs missing dir; heal prefers a populated candidate over an alphabetically earlier husk; a lazily-created identity is not required; provenance for all four layers.
- `tests/cli-doctor.test.ts` — the human line and its provenance note, the empty-registry WARN staying exit 0, a missing dir failing, and the JSON block.

Full suite: 4135 pass. Two failures are pre-existing on `main` (`harnesses-claude.test.ts` `--strict-mcp-config`, and a `lib-indexRefresh` case that passes in isolation) — both verified by stashing this branch's changes.

## Docs

README gains the "who am I when I omit `--persona`" paragraph and the command-task env guarantee; AGENTS.md gains invariant 38.

## Not in this PR

The issue's broader lesson — a poller whose "skip quietly" path is indistinguishable from "nothing new" is an invisible outage, and `task_runs` could carry a "did nothing, here is why" status distinct from plain success — is a schema change and a separate thread of work.


src/lib/personaDefault.ts
```diff
@@ -30,67 +91,111 @@ export async function adoptAsDefaultIfMissing(
...
-  // Prefer a persona whose name matches the broken default (case-insensitive).
+  // Prefer a persona whose name matches the broken default (case-insensitive),
+  // then the first candidate that is itself usable, and only then fall back to
+  // the first name on disk. Without the middle step a host with two husks and
+  // one real phantom heals alphabetically into another husk (#505).
   const brokenName = config.defaultPersona.toLowerCase();
-  const match = existing.find((n) => n.toLowerCase() === brokenName);
-  const healed = match ?? existing[0]!;
+  const match = existing.find(
+    (n) => n.toLowerCase() === brokenName && n !== config.defaultPersona,
```
kaieriksen: **Major — validate the case-insensitive match before healing to it.** This branch gives any case-only name match priority even when that directory is another husk. Repro: broken empty `kai`, empty `Kai`, and populated `real` heals to `Kai`, writes it to `state.json`, and leaves the host broken despite a usable candidate. Filter `match` through `defaultPersonaDefect(config, n) === null` (and add this three-directory case to the suite) before preferring it.

kaieriksen (changes requested):
One Major finding. The task-persona propagation, doctor reporting, docs, focused tests, typecheck, and full suite otherwise verify cleanly. Full local result: 4,137 pass, 2 platform skips, 0 fail.

src/lib/personaDefault.ts
andrewagrahamhodges: Fixed in 6cfb03b — good catch, the case match was acting as a licence rather than a hint.

Order is now: usable case-match → first usable candidate → case-match husk → first dir on disk. So your repro (broken `kai`, husk `Kai`, populated `real`) heals to `real`, and a host of nothing but husks still boots into the case match rather than alphabetically.

Three tests added, all mutation-checked (reverting the `usable(caseMatch)` guard fails the three-directory case): the repro, a usable case-match still beating an unrelated persona, and the all-husks fallback. Full suite 4139 pass / 1 fail — that failure is `harnesses-claude.test.ts` #338, pre-existing on the branch (verified by stashing).

andrewagrahamhodges (commented):
Addressed the Major finding in 6cfb03b. Ready for another look.

kaieriksen (approved):
Re-reviewed 6cfb03b. The Major finding is fixed: a case-only match now wins only when usable, otherwise healing selects the usable candidate. The three-directory regression, usable case-match precedence, and all-husks fallback are covered. Focused suite: 22 pass; typecheck passes.

PR #507, opened 2026-08-30
Log and document PHANTOMBOT_SANDBOX service suppression

### LINKED ISSUE #484: PHANTOMBOT_SANDBOX silently suppresses every service mutation — log it and document it
Follow-up from the review on #479 (non-blocking there).

`defaultServiceControl()` now wraps the host backend in `sandboxServiceControl()` when `PHANTOMBOT_SANDBOX` is set. That is right for a dev checkout, but it is a **global** kill-switch, not a TUI one: `defaultServiceControl()` has ~15 callers including `cli/update.ts`, `lib/serviceLifecycle.ts` and `channels/commands.ts`. With the var set, `/update` and `/restart` stop restarting anything and still report `ok: true`.

Two changes to stop it being a landmine:

1. **Say something.** Add `log.warn("platform: service change suppressed", { reason: "PHANTOMBOT_SANDBOX" })` inside `suppressed()` so it shows in the log pane and the journal. Today the explanation only reaches `stderr` and most call sites only check `ok`.
2. **Document it.** The var appears nowhere in README.md or AGENTS.md. An env var that can silently disable updating a production daemon belongs next to the thing it disables.

Behaviour and tests are fine as they stand (`lib-platform` covers the empty/`0`/truthy parsing and read-through `isActive`) — this is about discoverability.

Part of #471

robertclawson:
Closes #484.

`PHANTOMBOT_SANDBOX` turns every service mutation into a no-op that still reports `ok: true`. That is right for a dev checkout — a `bun run src/index.ts` working tree must not bounce the host's real `phantombot.service` — but it is a **global** switch, not a TUI one. `defaultServiceControl()` has ~15 callers, `cli/update.ts`, `lib/serviceLifecycle.ts` and `channels/commands.ts` among them, and most only check `ok`. So on a host with the var set, `/update` and `/restart` report success while restarting nothing, and the only explanation goes to a `stderr` field those call sites never read.

No behaviour change to the suppression itself — this is the discoverability follow-up the #479 review asked for.

### Say something

`suppressed()` now takes the op name and logs before returning:

```json
{"level":"warn","msg":"platform: service change suppressed","reason":"PHANTOMBOT_SANDBOX","op":"restart"}
```

`rerenderUnitIfStale` warns too. The issue only named `suppressed()`, but the re-render is a mutation on the same footing — it writes the real unit file on disk — and it was equally silent, so leaving it out would have left half the landmine armed. The op name is also appended to the existing `stderr` string, so a caller that *does* read stderr now learns which verb was dropped rather than just that something was.

`isActive()` is untouched and stays silent: it delegates, it is a query, and warning on it would make the log noisy enough to be ignored.

### Document it

The var appeared in neither README.md nor AGENTS.md. Now:

- **README** — a block at the end of *Service lifecycle*, next to the four verbs it disables: what it suppresses, that queries still tell the truth, why it exists, that it reaches `/update` and `/restart` too, and a "never set it on a host running the real service" warning.
- **AGENTS.md** — a block at the end of *Service model*, aimed at contributors: the blast radius, and the instruction not to drop the warning when adding a mutation to `ServiceControl`.

### Test

One new test in `tests/lib-platform.test.ts` drives all four mutations plus `isActive()` through a captured log sink and asserts exactly four warn lines, in order, each carrying `reason` and the right `op`. That is the part worth pinning: a verb added to `ServiceControl` later that suppresses silently reintroduces this bug, and now it fails the suite instead of shipping.

### Checks

- `bun run typecheck` clean.
- `bun test` — 4139 pass, 2 skip, 1 fail: `cli-doctor-persona-env`, which fails identically on `main` at `71e657a` (verified by checking out main and re-running). Unrelated to this change.
- The new README prose was caught by `readme-contract.test.ts` for naming a `phantombot status` command that does not exist — fixed to `phantombot doctor` before commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -572,6 +572,24 @@ in health checks or deploy hooks. These are the *external* controls (run from a
 terminal); the in-chat `/restart` and `/update` commands still bounce the running
 service from inside itself.
 
+**`PHANTOMBOT_SANDBOX` — a global kill-switch for all four verbs.** Set it to
+anything other than `""` or `0` and every service *mutation* in the process
+becomes a no-op: `start`, `stop`, `restart` and the unit-file re-render all
+return success without touching the host's service. Queries still tell the
+truth, so `phantombot doctor` and the TUI keep showing the real service state.
+
+It exists for **development checkouts**. Running `bun run src/index.ts` from a
+working tree gives you a second phantombot *process* but not a second
+*service* — so a config save in your branch would bounce the production daemon
+that is serving live conversations, killing whatever turn was in flight.
+
+Because it is read by `defaultServiceControl()`, it applies to *every* caller,
+not just the TUI: with the variable set, in-chat `/update` and `/restart` also
+stop restarting anything while still reporting success. Every suppressed
+mutation logs `platform: service change suppressed` at warn level, so
+`phantombot logs` tells you why nothing happened. **Never set it on a host
+running the real service** — the daemon will silently stop taking updates.
+
 ## Configuration
 
 Phantombot resolves configuration in this order:
--- a/README.md
+++ b/README.md
@@ -586,9 +586,13 @@ that is serving live conversations, killing whatever turn was in flight.
 Because it is read by `defaultServiceControl()`, it applies to *every* caller,
 not just the TUI: with the variable set, in-chat `/update` and `/restart` also
 stop restarting anything while still reporting success. Every suppressed
-mutation logs `platform: service change suppressed` at warn level, so
-`phantombot logs` tells you why nothing happened. **Never set it on a host
-running the real service** — the daemon will silently stop taking updates.
+mutation logs `platform: service change suppressed`, naming the op. That line
+goes to the **stderr of the process that has the variable set** — your terminal
+for a `bun run src/index.ts` checkout, and the `^l` log pane inside the TUI
+(which swaps the sink for its own ring buffer). It reaches `phantombot logs`
+only if the *daemon itself* was started with the variable set — which is the
+case you never want: **never set it on a host running the real service**, or
+the daemon will silently stop taking updates.
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -574,9 +574,10 @@ service from inside itself.
 
 **`PHANTOMBOT_SANDBOX` — a global kill-switch for all four verbs.** Set it to
 anything other than `""` or `0` and every service *mutation* in the process
-becomes a no-op: `start`, `stop`, `restart` and the unit-file re-render all
-return success without touching the host's service. Queries still tell the
-truth, so `phantombot doctor` and the TUI keep showing the real service state.
+becomes a no-op: `start`, `stop` and `restart` return success without touching
+the host's service, while the unit-file re-render writes nothing and reports
+`{ rerendered: false }`. Queries still tell the truth, so `phantombot doctor`
+and the TUI keep showing the real service state.
 
 It exists for **development checkouts**. Running `bun run src/index.ts` from a
 working tree gives you a second phantombot *process* but not a second
@@ -585,10 +586,14 @@ that is serving live conversations, killing whatever turn was in flight.
 
 Because it is read by `defaultServiceControl()`, it applies to *every* caller,
 not just the TUI: with the variable set, in-chat `/update` and `/restart` also
-stop restarting anything while still reporting success. Every suppressed
-mutation logs `platform: service change suppressed`, naming the op. That line
-goes to the **stderr of the process that has the variable set** — your terminal
-for a `bun run src/index.ts` checkout, and the `^l` log pane inside the TUI
+stop restarting anything while still reporting success. Every suppression is
+logged, naming the op: `start`, `stop` and `restart` log
+`platform: service change suppressed` at **warn**, and the re-render logs
+`platform: unit re-render skipped` at **info** — so a process running at
+`PHANTOMBOT_LOG_LEVEL=warn` keeps the three that report a false success and
+drops the one that does not. Those lines go to the **stderr of the process
+that has the variable set** — your terminal for a `bun run src/index.ts`
+checkout, and the `^l` log pane inside the TUI
 (which swaps the sink for its own ring buffer). It reaches `phantombot logs`
 only if the *daemon itself* was started with the variable set — which is the
 case you never want: **never set it on a host running the real service**, or
--- a/src/lib/platform.ts
+++ b/src/lib/platform.ts
@@ -141,6 +141,9 @@ export function isSandbox(
  * warning. Without it, `/update` on a host that happens to have the var
  * set reports success while never restarting anything, and the only
  * explanation goes to a `stderr` field most call sites never read.
+ * `rerenderUnitIfStale()` logs at info instead: it returns
+ * {rerendered:false}, which is honest rather than a false success, and is
+ * what the real backend returns in a dev checkout anyway.
  */
 export function sandboxServiceControl(inner: ServiceControl): ServiceControl {
   const suppressed = (op: string) => async () => {
@@ -158,8 +161,16 @@ export function sandboxServiceControl(inner: ServiceControl): ServiceControl {
     start: suppressed("start"),
     stop: suppressed("stop"),
     restart: suppressed("restart"),
+    // Logged, but NOT at warn — unlike the three verbs above, this one has
+    // no deceptive success to warn about. It answers {rerendered:false},
+    // which is exactly what the real POSIX backend answers in the sandbox's
+    // own headline case: `defaultRerenderUnitIfStale` bails on
+    // `isPhantombotBinary(process.execPath)`, false under `bun run`. Warning
+    // there would fire on a call that was never going to write anything, and
+    // a warn line that cries wolf in the common case is one contributors
+    // learn to ignore — including the three that do matter.
     async rerenderUnitIfStale() {
-      log.warn("platform: service change suppressed", {
+      log.info("platform: unit re-render skipped", {
         reason: "PHANTOMBOT_SANDBOX",
         op: "rerenderUnitIfStale",
       });
--- a/src/lib/platform.ts
+++ b/src/lib/platform.ts
@@ -32,6 +32,7 @@ import {
   launchdLogPaths,
   launchdLogsDir,
 } from "./launchd.ts";
+import { log } from "./logger.ts";
 import {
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -572,6 +572,24 @@ in health checks or deploy hooks. These are the *external* controls (run from a
...
+It exists for **development checkouts**. Running `bun run src/index.ts` from a
+working tree gives you a second phantombot *process* but not a second
+*service* — so a config save in your branch would bounce the production daemon
+that is serving live conversations, killing whatever turn was in flight.
+
+Because it is read by `defaultServiceControl()`, it applies to *every* caller,
+not just the TUI: with the variable set, in-chat `/update` and `/restart` also
+stop restarting anything while still reporting success. Every suppressed
+mutation logs `platform: service change suppressed` at warn level, so
+`phantombot logs` tells you why nothing happened. **Never set it on a host
```
andrewagrahamhodges: **Minor — this points at the wrong log destination for the case the section is about.**

`phantombot logs` tails the *service* logs (`journalctl --user -u phantombot` / the launchd out-err files / the Task Scheduler out log — see the table above and `src/cli/logs.ts`). But the warning is emitted by whichever process has `PHANTOMBOT_SANDBOX` set, and the documented reason to set it is a dev checkout: `bun run src/index.ts`. That process is not the unit, so its stderr goes to your terminal and never into the journal. In the TUI it does not even reach stderr — `logSink.ts` swaps the sink for the ring buffer, so the line shows under `^l`.

So the one reader who follows this sentence in the situation it describes finds nothing, and concludes the suppression was silent after all — the exact failure #484 is about.

Suggest naming where it actually lands, something like: *"the warning goes to the stderr of the process that has the variable set — your terminal for a `bun run` checkout, the `^l` log pane in the TUI, and `phantombot logs` only if the daemon itself was started with it set."* That last clause is also the one that makes the following **Never set it on a host running the real service** warning checkable.

src/lib/platform.ts
```diff
@@ -133,18 +134,35 @@ export function isSandbox(
...
-    start: suppressed,
-    stop: suppressed,
-    restart: suppressed,
+    start: suppressed("start"),
+    stop: suppressed("stop"),
+    restart: suppressed("restart"),
     async rerenderUnitIfStale() {
+      log.warn("platform: service change suppressed", {
+        reason: "PHANTOMBOT_SANDBOX",
+        op: "rerenderUnitIfStale",
```
andrewagrahamhodges, line 175: **Nit — this one warns about a change that would not have happened anyway.**

Unlike `start`/`stop`/`restart`, the real `defaultRerenderUnitIfStale` (systemd.ts:494) bails to `{ rerendered: false }` when `isPhantombotBinary(process.execPath)` is false. Under the sandbox's own headline use case — `bun run src/index.ts` — that is always the case, so the host backend would have touched nothing and the wrapper still logs `service change suppressed`.

Low volume today (`maybeUpgradeUnit` is the only caller), so not worth gating the PR on. But the warn line is load-bearing here: it is the signal you are asking future contributors not to remove, and the fastest way to get it ignored is for the common case to be a false alarm. Worth a follow-up thought about warning only when the wrapped call would have done work, or wording this op's message as "re-render skipped" rather than "service change suppressed".

andrewagrahamhodges (commented):
Verified locally on `f7b435a`: typecheck clean, full suite 4140 pass / 2 skip / 1 fail (`#338` claude-harness, fails identically on `main` at `71e657a`). Mutation-checked the new test by deleting the `rerenderUnitIfStale` warn — it fails, so the assertion is real and the "a new silent verb reintroduces #484" guard actually holds.

The code change is right and correctly scoped: `isActive()` stays silent, the op name reaches both the log line and `stderr`, and including `rerenderUnitIfStale` was the correct call since it writes the real unit file.

One Minor on the README (it points readers at a log destination the warning does not reach in the scenario the section is about) and one Nit. No blocking findings.

README.md
robertclawson: Fixed in `e28b12a`. You were right that the sentence was unfollowable in exactly the case the section documents — `logs.ts` tails the unit, and `logSink.ts` swaps the sink for the ring buffer under the TUI, so a `bun run` checkout is the one place the line never reaches.

Rewritten to name the destination rather than a command: stderr of the process holding the variable — your terminal for a dev checkout, the `^l` pane in the TUI, and `phantombot logs` only if the daemon itself was started with it set. That last clause now makes the "never set it on a host running the real service" warning checkable, as you noted.

src/lib/platform.ts
robertclawson: Taken, in `e28b12a` — the "warning only when the wrapped call would have done work" option is unreachable from here (the wrapper is backend-generic and deliberately never calls `inner`), so I took your second suggestion.

`rerenderUnitIfStale` now logs at **info** as `platform: unit re-render skipped`, same `reason`/`op` fields. The distinction I settled on: the three verbs warn because they report `ok: true`, which is a false success ~15 callers cannot tell from a real restart. This one returns `{rerendered: false}` — honest, and the identical answer `defaultRerenderUnitIfStale` gives under `bun run` where it bails on `isPhantombotBinary`. Nothing deceptive, so nothing to warn about.

The #484 guard is unchanged: the test still requires all four ops to log and pins each level and message, so a new verb that suppresses silently still fails the suite. Documented in both README and AGENTS.md. One caveat worth naming: at `PHANTOMBOT_LOG_LEVEL=warn` the re-render line now drops out — acceptable for a call that writes nothing, but say the word if you would rather it stayed at warn with the reworded message.

README.md
```diff
@@ -572,6 +572,28 @@ in health checks or deploy hooks. These are the *external* controls (run from a
 terminal); the in-chat `/restart` and `/update` commands still bounce the running
 service from inside itself.
 
+**`PHANTOMBOT_SANDBOX` — a global kill-switch for all four verbs.** Set it to
+anything other than `""` or `0` and every service *mutation* in the process
+becomes a no-op: `start`, `stop`, `restart` and the unit-file re-render all
+return success without touching the host's service. Queries still tell the
```
andrewagrahamhodges: This now contradicts the change in this commit. The unit-file re-render does not `return success` — it returns `{rerendered: false}`, which is exactly the argument the commit message makes for logging it at info rather than warn. As written, a reader is told the re-render lies to them the same way `restart` does, which is the belief the code change is trying to correct.

Suggest splitting the sentence: `start`, `stop` and `restart` return success without touching the host's service, while the unit-file re-render reports `{rerendered: false}` and writes nothing.

README.md
```diff
@@ -572,6 +572,28 @@ in health checks or deploy hooks. These are the *external* controls (run from a
...
+truth, so `phantombot doctor` and the TUI keep showing the real service state.
+
+It exists for **development checkouts**. Running `bun run src/index.ts` from a
+working tree gives you a second phantombot *process* but not a second
+*service* — so a config save in your branch would bounce the production daemon
+that is serving live conversations, killing whatever turn was in flight.
+
+Because it is read by `defaultServiceControl()`, it applies to *every* caller,
+not just the TUI: with the variable set, in-chat `/update` and `/restart` also
+stop restarting anything while still reporting success. Every suppressed
```
andrewagrahamhodges: "Every suppressed mutation logs `platform: service change suppressed`" is no longer true of the fourth verb, which logs `platform: unit re-render skipped` at info. This is the first finding in miniature: a reader hits the documented silence, greps stderr for the documented string, and finds nothing for the one case the doc sent them looking at.

Worth naming both lines and their levels here, since the level is what decides whether it survives a `PHANTOMBOT_LOG_LEVEL=warn` process: the three mutating verbs log `platform: service change suppressed` at warn, the re-render logs `platform: unit re-render skipped` at info. The rest of the paragraph about where stderr actually goes is right and reads well.

andrewagrahamhodges (changes requested):
Both findings from the last pass are properly fixed, and the reasoning in the commit message is the right one: the re-render verb returns `{rerendered:false}` rather than a false success, so it is not in the same hazard class as the three that report `ok:true`. Verified locally on `e28b12a`: typecheck clean, `tests/lib-platform.test.ts` 27 pass, full suite 4140 pass with only the known #338 claude-harness failure that is already on main. I mutation-checked the new level assertion by flipping the info call back to `log.warn("platform: service change suppressed")` and the test fails, so it is a real test and not a restatement of the code.

One thing is now out of step, and it is the same defect as the first finding rather than a new one: the README paragraph still describes all four verbs uniformly, so a reader who follows it in the re-render case looks for a string and a behaviour that no longer exist. Two small edits and this is an approve.

Nothing else blocking. `log.info` is at the default `minLevel`, so the new line is genuinely visible without setting `PHANTOMBOT_LOG_LEVEL`, and AGENTS.md carries the do-not-remove rule for the three warn lines with the reason attached.

robertclawson:
Both README findings fixed in f969c24 — docs only, no code change.

- The opening sentence now splits by verb: `start`, `stop` and `restart` return success without touching the service, while the unit-file re-render writes nothing and reports `{ rerendered: false }`.
- The suppression-logging sentence now names both lines *and* their levels: `platform: service change suppressed` at warn for the three mutating verbs, `platform: unit re-render skipped` at info for the re-render — with the consequence spelled out, that a process at `PHANTOMBOT_LOG_LEVEL=warn` keeps the three that report a false success and drops the one that does not.

The paragraph on where stderr actually lands is unchanged.

andrewagrahamhodges (approved):
Approving on `f969c24`.

Both README sentences are fixed correctly, and the fix goes to the substance rather than the wording: the paragraph now separates the three verbs that report a false `ok: true` from the re-render that honestly answers `{ rerendered: false }`, and it names both log lines *with their levels*. That level detail is the part that makes it usable — a reader on `PHANTOMBOT_LOG_LEVEL=warn` can now predict which of the two lines they will actually see, which was the whole point of the original finding.

Verified rather than taken on trust:

- The documented strings and levels match the source exactly — `platform: service change suppressed` at warn for `start`/`stop`/`restart` (platform.ts:150), `platform: unit re-render skipped` at info (platform.ts:173).
- Grepped both `README.md` and `AGENTS.md` for the old uniform claim; AGENTS.md:308-315 already carries the split description, so there is no remaining prose that describes the four verbs as behaving alike.
- `bun run typecheck` clean. Suite: 4140 pass, 2 skip, 1 fail — the fail is the known `ClaudeHarness ... --strict-mcp-config (#338)` test that is already red on main, untouched by this PR.

The `#484` guard still holds: the test pins all four ops, their messages and their levels, so a fifth verb that suppresses silently still fails the suite. Nothing further from me.

PR #508, opened 2026-08-30
TUI configuration overhaul: settings table, Brain/Channels/Memory/Voice screen flows

lenaparkhodges:
## Summary

Rebuilds the TUI configuration experience in a native-Ink, chat-TUI visual language (no external components, **no mouse reporting** — terminals keep native selection).

### Settings screen (Dashboard design language)
- Squared 3-column table: **setting · description · state** — static one-liner descriptions; live `/status` output moved to a STATUS block at the bottom (re-gathered on every return to the screen).
- Three-state badges: green = configured, red = required + unconfigured, yellow = optional + unconfigured.
- Informational group below a separator rule: Autostart (`on/off`), Default (`yes/no`, greyed out on single-phantom hosts), Release Channel (`stable/preview`).
- All toggles real: Autostart and Default write per-persona config with a confirm panel + restart offer; Release Channel writes `update_channel` to the global config.toml (`stable` deletes the key).
- Doctor reachable from the phantoms list via `d` (row-scoped); removed from the configure screen.
- Host logs tail available from the phantoms list via `L` (100-line ring buffer).

### Flows
- **Brain:** primary → fallback → Pi slots with searchable model lists (text search, 50-row window, paste-chunk, current-value markers); restart offered only after a real change; `esc` = "brain unchanged".
- **Chat Channels:** picker (PhantomChat / Telegram) running the same walkthroughs as the `phantombot phantomchat` / `phantombot telegram` CLI wizards, with the clack guide text carried over (BotFather steps, npub copy instructions); explicit "paste X" prompts; Telegram token re-asks on malformed shape, validates via getMe.
- **Memory:** the `phantombot embedding` walkthrough — provider picker, key keep-or-replace with live one-token embed validation, endpoint questions for OpenAI-compatible; idempotent ("memory unchanged"); in-app re-embed with progress; no test box.
- **Voice:** the `phantombot voice` walkthrough — provider picker, key validation, voice/model choices; idempotent; no preview/play.
- Legacy Memory/Voice screens removed.

### Other
- Native Ink file editor replaces the `$EDITOR` handover; file picker offers AGENTS.md (drops legacy USER.md).
- Short-window layout budget fixed: variable-height STATUS block properly budgeted (no more row shear in small terminals).

## Test plan
- Full suite: **4143 pass / 0 fail** (2 platform-skips), tsc clean.
- Visual pass in the sandbox tree across all flows by @aghodgespwa.


robertclawson (changes requested):
Reviewed at `78f02f6`, merged onto current `main` locally. The design work here is strong — the pure-reducer split in `editorBuffer.ts`, the host-vs-persona reasoning in `actions.ts`, and the removal of `mouse.ts` are all clear wins. Five findings, two of which I'd treat as blocking.

---

### 1. Blocking — the file editor can destroy a file it failed to read

`src/tui/screens/FileEditor.tsx:70`

```ts
Bun.file(props.path).text().catch(() => "")
```

Every read failure is collapsed into "empty file". `ENOENT` legitimately means "" (you're creating `AGENTS.md`), but `EACCES`, `EISDIR` and transient I/O all render as a blank buffer with no error shown — and `save()` then writes that buffer back unconditionally (`^s` doesn't check `dirty`).

Reproduced end to end on a mode-`0200` file, where read fails but write succeeds:

```
$ printf 'IMPORTANT CONTENT\nline2\nline3\n' > AGENTS.md && chmod 0200 AGENTS.md
loaded: ""  -> empty buffer: true
SAVE SUCCEEDED
$ cat AGENTS.md
x
```

30 bytes gone, no warning at any point. The `$EDITOR` handover this replaces was safe here — the editor surfaced the error itself — so this is a regression introduced by the swap.

Suggest: keep `""` only for `ENOENT`, and on any other error set `error` and refuse to save until the file has been read successfully.

---

### 2. Blocking — `applyUpdateChannel` lets a persona agent re-point the whole host

`src/tui/actions.ts:320`

`applyDefaultPersona` deliberately refuses when `PHANTOMBOT_PERSONA` is set, with the rationale spelled out in its own docblock: *"a persona agent must not be able to re-point the daemon-wide default, and the TUI is reachable from one."*

`applyUpdateChannel` writes `update_channel` to the **global** config.toml with no such guard — and the argument is stronger there, since flipping a box to `preview` changes which binary every persona on it runs within the hour.

This is the codebase's own stated doctrine, from `personaConfig.ts:79`:

> A persona cannot elect itself default, **change the release ring the box follows**, relocate the personas root, or repoint the shared memory database.

`HOST_ONLY_KEYS` strips `update_channel` from the persona layer precisely so a persona file can't set it. The new toggle reopens that path through the TUI instead. Same guard as `applyDefaultPersona` looks like the fix.

(Checked separately and it's fine: `smol-toml`'s `stringify` hoists top-level scalars above all tables, so the write lands at the top of the file rather than inside the last `[table]`.)

---

### 3. CI is red — branch is 2 commits behind `main`

`typecheck` and `test-windows` both fail on the same three fixtures:

```
tests/tui-navigation.test.tsx(199,7):    error TS2741
tests/tui-restart-removed.test.tsx(148,7): error TS2741
tests/tui-settings-fit.test.tsx(124,7):  error TS2741
  Property 'default_persona' is missing ... but required in type 'DoctorReport'
```

`default_persona` landed on `DoctorReport` in v1.1.328 after this branch forked. Mechanical: `update-branch`, then add the field to the three `FAKE_REPORT` literals.

---

### 4. `tui-settings-fit` is timing-dependent and will flake

`tests/tui-settings-fit.test.tsx:214`

*"a tall window shows the whole screen, status included"* fails reproducibly here (3/3 runs, on the PR head alone as well as merged) because the STATUS block is still rendering `gathering…` when the assertions run:

```
 STATUS
 gathering…
```

The only synchronisation is `await sleep(200)` for an async gather. Raising it to 3000ms makes it pass 5/5 — so it's purely timing, not a layout bug. It's green on today's CI runner, which means it'll fail on a loaded one instead. Worth polling for the gathered content rather than sleeping a fixed interval.

Everything else in the suite matches `main`'s baseline exactly: 10 pre-existing environment/subprocess failures on both sides, this one extra on the PR.

---

### 5. Minor — every tab- or CRLF-containing file opens already "dirty"

`src/tui/editorBuffer.ts:29`

`editorFromText` normalises tabs to two spaces and `\r\n` to `\n` on load, but `editorIsDirty` compares against the raw original:

```
dirty on open, zero keystrokes: true
CRLF file dirty on open:        true  ->  "a\nb\n"
```

Two consequences. Pressing `esc` on a file you only *looked at* pops the "Save & exit / Discard changes / Cancel" menu, so the discard prompt fires for edits that don't exist. And a `^s` silently rewrites the entire file's line endings — on Windows, which is the platform whose console flakiness motivated replacing `$EDITOR` in the first place.

Normalising the comparison baseline the same way the buffer is normalised would settle both.

robertclawson:
CI is red on `78f02f6` (typecheck + test-windows), and it isn't a defect in this PR's logic — the branch is just behind `main`.

`main` (#506, `71e657a`) added a **required** `default_persona` block to `DoctorReport`, so the three `FAKE_REPORT` fixtures added here no longer satisfy the type:

```
tests/tui-navigation.test.tsx(199,7): error TS2741: Property 'default_persona' is missing ...
tests/tui-restart-removed.test.tsx(148,7): ...
tests/tui-settings-fit.test.tsx(124,7): ...
```

Fix: update the branch from `main`, then add the field to each fixture, e.g. after the `update:` line:

```ts
  default_persona: {
    resolved: "alice",
    provenance: "state",
    exists: true,
    served: true,
    defect: null,
    mcp_servers: 0,
    mcp_elsewhere: [],
    healthy: true,
    detail: "alice (state.json)",
  },
```

I applied exactly that on a local merge of `main` into this branch and `bun tsc --noEmit` exits 0.

lenaparkhodges:
All five findings addressed in `9e1af78` (branch also merged forward from `main`, so the `default_persona` typecheck failures are gone at the source).

**1 — FileEditor destroying a file it failed to read (blocking):** only `ENOENT` now collapses to an empty buffer (new file). Any other read failure renders `cannot read <file>: <message>` with no editable buffer — `^s` and the exit menu are unreachable — and `esc` exits with `saved: false, changed: false`. Regression test mounts a chmod-`0200`-style file (mode `0000`, read fails / write would succeed) and asserts the contents survive verbatim.

**2 — applyUpdateChannel from a persona agent (blocking):** same guard as `applyDefaultPersona` — refuses under `PHANTOMBOT_PERSONA` with a message naming the persona. UI parity too: the Release Channel row is dimmed and inert when running as a persona agent, with a description explaining it is host-only.

**3 — red CI:** branch updated from `main`; the three `FAKE_REPORT` literals carry `default_persona`.

**4 — flaky `tui-settings-fit`:** the fixed `sleep(200)` is replaced with a poll that waits for the configure screen to open AND the STATUS gather to settle (`description` present, no `gathering…`), 6s deadline. Passes 3/3 locally on an explicitly loaded run; the fixed sleep was the flake, not the layout.

**5 — dirty-on-open for tab/CRLF files:** `editorIsDirty` now runs the original through the same normalisation as the loader (CRLF/CR → LF, tabs → two spaces). A looked-at-but-untouched file is clean, and `^s` no longer rewrites line endings. Pinned by reducer tests.

Full suite: 4165 pass / 0 fail, tsc clean.

robertclawson (approved):
Re-reviewed at `9e1af78`. Both blockers are fixed, and fixed at the right layer. Approving.

**1. FileEditor — resolved.** `ENOENT` is now the only error that collapses to an empty buffer; every other read failure sets `loadError`, leaves `buf` as `null`, renders `cannot read <file>`, and gates `useInput` so `esc` is the only key that does anything. `save()` also returns `false` on a null buffer, so there are two independent stops rather than one. My original repro (mode-`0200` file: read fails, write succeeds) can no longer reach the write.

**2. `applyUpdateChannel` — resolved.** The `PHANTOMBOT_PERSONA` refusal now mirrors `applyDefaultPersona` exactly, and the docblock cites the `personaConfig.ts` doctrine it is enforcing, so the next person to touch it knows why the guard is there rather than deleting it as redundant. Dimming the row as well as refusing in the action is the right belt-and-braces: the UI doesn't offer something the backend will reject.

**Nits, both taken.** `editorIsDirty` now normalises the baseline the same way the loader does, so tab/CRLF files no longer open dirty. `tui-settings-fit` polls for the settled frame instead of `sleep(200)` — that one would have gone red on a loaded runner, so it's a real flake removed, not just a slower sleep.

**Verification on my side:** `bun tsc --noEmit` clean. Full suite 4164 pass / 2 skip / **1 fail** — `runDoctor — PHANTOMBOT_PERSONA env (production path)`. Because that failure names the exact env var this PR's fix touches, I ran it on a fresh clone of `main` as well: it fails there identically, so it's pre-existing (#338 class) and not PR-owned. CI is 6/6 green, and the merge from main cleared the `default_persona` fixture breakage.

PR #509, opened 2026-08-31
TUI chat usability improvements and new persona workflow fixes

lenaparkhodges:
## TUI chat usability improvements

- **Transcript scrolling:** input history is gone; ↑/↓ scroll the chat thread line by line, PgUp/PgDn page, Home/End jump to the top / snap back to the live bottom. Scrolling works while a reply is streaming.
- **Newline keys:** Alt+Enter inserts a line break on every terminal; Ctrl+J is the universal fallback (legacy terminals cannot distinguish Ctrl/Shift+Enter from Enter). Shift/Ctrl+Enter are also supported on kitty-protocol terminals — the app now pushes the kitty keyboard protocol at startup and pops it whenever a clack prompt or `$EDITOR` borrows the terminal.
- **Caret:** renders inline after the last character of multi-line input (was parked at the top-right of the block).
- **Footer:** shows `Alt+↵ Newline` so the newline key is discoverable; compacted to fit a single 80-column row.
- **`^k` key inspector:** shows the raw bytes the terminal sends per keystroke (diagnostic aid).
- **Chat input batching (19ff85a, de7d85f, 226eb9a):** text + one trailing newline in a single read is a fast typist's batched Enter → submits; interior newlines are a paste → lands whole without submitting; a chunk arriving while a turn is busy keeps its text in the box. Covers coalesced stdin writes from terminals like GNOME Terminal.

## New persona workflow

- **Create flow trimmed to 3 questions** (name, one-line identity, tone) with two optional additions — Skills & disciplines (new MultiChoose screen, preseeded expertise options) and owner name. Both seed IDENTITY.md's "Who you work for" / expertise sections.
- **First-run wizard unchanged in scope** (six steps) but gains the same Skills + owner questions.
- **Keypress reliability:** new `useStableInput` (ref-forwarded stable handler) fixes Ink's resubscribe-on-render window that silently dropped the first Enter on freshly-mounted question screens.

## Unified Brain onboarding (wizard + Configure)

- Configure's **Brain row now runs the exact same flow** as the wizard's Brain steps: primary harness (+ Pi install offer if missing) → Pi model slots → fallback → **Test now / Skip**. No divergent menus.
- **Failed live test keeps you in the wizard** — error note, then Start over (re-checks availability, nothing written on failure) or Back to Configure. No more dead-end dump into Configure.
- **API keys are validated against the provider at entry time** (`providerKeyProbe`): one cheap authenticated call per provider before model slots. 401/403 → rejected, re-prompt, nothing written to vault or Pi's store; unknown provider / network blip → warn and continue (the end-of-flow live test still guards). Also validates a *kept* stored key, so stale keys are caught before model picking.
- **One source of truth for brain readiness:** `personaCompleteness`, the Configure badge (`brainConfigured`) and the effective chain probe all delegate to the same `defaultLocalChain` + `harnessChainIds` readers as the boot gate and runtime. A persona whose chain lives in the host `[harnesses.personas.<name>]` table (default-persona write path) no longer shows a red `required` badge forever.

## Autostart rework

- **Configure rows grouped** with horizontal rules: Autostart | Identity | Chat Channels.
- **Default persona:** autostart selector **Off | Login | Boot**. **Non-default personas:** On | Off. New personas default to **Login**; existing `autostart_personas` entries are inherited as Login — nobody's setup changes under them.
- **Boot:** passwordless sudo is probed first (`sudo -n true`) — passwordless boxes apply `loginctl enable-linger` with **no prompt and no vault write**. When a password is required it is validated immediately (`sudo -S -k -v`) and re-prompted on rejection, like the API-key probe; credentials are stored in the **default agent's vault** and reprompted if the default moves. Same validate-and-reprompt doctrine for the Windows account password.
- **Chat banner** shows the autostart mode of the chatting persona — loud when Off.

## Zero-service operation

- Verified live end-to-end: the TUI boots, loads history and chats with **no phantombot systemd unit configured at all**; all systemd touchpoints (dashboard badge, restart-to-apply) are lazy and degrade gracefully. TUI closed → nothing left running.

## Testing

- Full suite **4228 tests, 0 fail**; `tsc --noEmit` clean. New suites: `brain-onboarding` (8 cases), `lib-autostart-boot` (7 cases), `lib-persona-snapshot-brain` regression tests, provider-key-probe cases in brain-flow tests, live-verified against real OpenAI/OpenRouter 401s and a real pi chain.


robertclawson:
CI on b838b3d is red: 11 failures in `tests/tui-chat-slash.test.tsx` and `tests/tui-chat-activity.test.tsx` ([run 33370683619](https://github.com/phantomyard/phantombot/actions/runs/33370683619)). Reproduced locally on the branch head.

**Root cause — the new "a paste is never a submit" branch swallows a batched Enter.**

The tests write the text and the Enter as two back-to-back `stdin.write()` calls with no await between them, so the PassThrough coalesces them and Ink delivers **one** chunk. Probed against this branch's own test rig:

```
stdin.write("hi"); stdin.write("\r");   ->  CHAR="hi\r"  KEYS=(none)
stdin.write("hi"); await sleep; stdin.write("\r")  ->  CHAR="hi" KEYS=(none), CHAR="\r" KEYS=return
```

So in the coalesced case `key.return` is **false**, control reaches

```ts
if (/\r|\n/.test(char)) {
  setInputValue(inputRef.current + char.replace(/\r\n?/g, "\n"));
  return;
}
```

and the Enter becomes a newline in the box. The failing frames show exactly that — `› hi` with the caret on the next line, and no turn ever started. `applyTextChunk` used to handle this (split-and-submit), which is why main is green.

This is not just a test artefact. Coalescing is what a real terminal does under latency (ssh/tmux) or from a fast typist — the old comment in this file called out "a terminal that batched keystrokes" for precisely this reason. As written, those messages are silently eaten.

**Suggested fix** — keep the PR's intent (multi-line pastes land whole) and distinguish it by *where* the newline is: a chunk that is plain text plus exactly one trailing newline is typing+Enter; interior newlines mean paste. Ink 6 has no bracketed-paste support, so position is the only discriminator available.

```diff
       if (/\r|\n/.test(char)) {
+        // A chunk that is plain text plus ONE trailing newline is a fast
+        // typist (or a laggy ssh link) whose keystrokes and Enter arrived in
+        // the same read — swallowing that Enter loses every message typed
+        // faster than the event loop. A chunk with INTERIOR newlines is a
+        // paste, and lands whole without submitting.
+        const body = char.replace(/\r\n?$|\n$/, "");
+        if (!/\r|\n/.test(body)) {
+          const text = (inputRef.current + body).trim();
+          setInputValue("");
+          if (!text) return;
+          const isCommand = commandName(text) !== undefined;
+          if (busy && !isCommand) {
+            setInputValue(inputRef.current + body);
+            return;
+          }
+          if (isCommand) void runCommand(text);
+          else void submit(text);
+          return;
+        }
         setInputValue(inputRef.current + char.replace(/\r\n?/g, "\n"));
         return;
       }
```

With that applied on top of b838b3d: `bun test tests/tui-chat-slash.test.tsx tests/tui-chat-activity.test.tsx tests/tui-chat-scroll.test.tsx tests/tui-chat-commands.test.ts` → **26 pass, 0 fail**, and `bun tsc --noEmit` exits 0.

Two things worth deciding rather than assuming:

1. The residual ambiguity is a **single-line** paste that ends in a newline — that will submit. I think that's the right call (it matches every other terminal chat client), but if you'd rather it didn't, the alternative is to drop only a trailing `\r` and treat a trailing `\n` as literal, since a real Enter is `\r` in raw mode and pasted line breaks are usually `\n`. That's more precise but relies on the paste source normalising.
2. The `busy` path: I put the un-submitted text back in the box so it isn't lost. Commands still dispatch while a turn is in flight, matching the `key.return` branch above.

Not pushing to your branch — it's yours. Happy to review again once it's green.

lenaparkhodges:
Fixed in 19ff85a — thanks for the sharp catch and the probing rig, that coalescing table made the root cause obvious.

Applied your position-based discriminator (plain text + exactly one trailing newline = typing+Enter, interior newlines = paste). Two deviations from the suggested diff:

1. On the busy path I return early without touching the input state — `inputRef.current` is unchanged by definition, so the text stays in the box with no clear/restore round-trip.
2. Left the single-line-paste-ends-in-newline ambiguity resolved toward submit as you suggested (matches every other terminal chat client); did not take the trailing-`\r`-vs-`\n` refinement since it depends on the paste source normalising.

`bun test` on the four TUI chat files: 26 pass / 0 fail; `tsc --noEmit` clean; full suite 4203 pass / 0 fail.

robertclawson (dismissed):
Verified 19ff85a — the discriminator is right and the reasoning in the comment block is exactly the constraint (Ink 6 gives no bracketed paste, so newline *position* is all we have). `body = char.replace(/\r\n?$|\n$/, "")` strips exactly one trailing terminator and interior newlines still land whole. CI green on this head (typecheck, test, test-windows, CodeQL). Approving.

One non-blocking follow-up, and it is **pre-existing on `main`**, not something this commit introduced — but this PR is the one about not eating typed text, so it belongs here:

On the busy path the coalesced chunk's *text* is dropped, not kept.

```ts
if (busy && !isCommand) {
  // Keep the un-submitted text in the box so it isn't lost;
  return;
}
```

`inputRef.current` is unchanged, but `body` was never appended to it — so when the whole message arrives in one chunk while a turn is in flight, the box stays empty and the message is gone. The comment describes the slow-typing case (where the characters already went in one at a time and the chunk is a bare `\r`), which is why the two look equivalent.

Reproduced against this head, with a hanging `send` so the screen stays busy:

```ts
stdin.write("first\r");   // starts a turn, screen goes busy
await sleep(200);
stdin.write("hello\r");   // coalesced, while busy
await sleep(200);
expect(sent).toEqual(["first"]);            // passes: correctly not submitted
expect(lastFrame(stdout.frames)).toContain("hello");  // FAILS: box is empty
```

Same probe on `main` fails identically (`applyTextChunk` clears the text and then `else if (!busy)` swallows the submit), so it is not a regression — but note `key.return`'s busy branch does preserve the text, so the three paths disagree today.

One line fixes it:

```ts
if (busy && !isCommand) {
  setInputValue(inputRef.current + body);
  return;
}
```

Happy for it to be a separate commit or a follow-up issue — not holding the approval on it.

lenaparkhodges:
Fixed in de7d85f — good catch, and thanks for the exact repro. Added it as the regression test `a chunk that arrives while busy keeps its text in the box` in tui-chat-activity. The busy branch now appends `body` to the box, matching what key.return's busy branch already did; your reading was right that the comment described the slow-typing case where the characters had already gone in one keystroke at a time, so the branch looked equivalent when it wasn't. `bunx tsc --noEmit` clean, all tui-chat suites pass (7/7 in tui-chat-activity including the new test).

robertclawson (dismissed):
Re-approved at de7d85f — the follow-up dismissed my previous approval, so this is a fresh pass over the delta.

The busy branch now appends `body` (newline already stripped) to the box, so the three paths finally agree: `key.return` while busy leaves the box untouched, a coalesced chunk while busy leaves the same characters in the box, and neither submits. The regression test reproduces the exact failure I probed — `first\r` then `hello\r` while the turn is in flight, asserting both `sent === ["first"]` and that the frame still contains `hello` — so it would have failed on 19ff85a for the right reason.

Verified locally on de7d85f: all 19 `tests/tui-*` files 98 pass / 0 fail, `bunx tsc --noEmit` exit 0. CI green across typecheck, test, test-windows and CodeQL.

robertclawson (changes requested):
Re-review of `226eb9a`. This push adds a large new scope on top of the chat-input work I approved earlier (brain onboarding unification, autostart modes, persona create overhaul — 49 files, +3849/−812), so I re-reviewed it as new work rather than as a follow-up.

The design is good and the test coverage is real: 4225 pass locally, the one failure (`runDoctor — PHANTOMBOT_PERSONA env`) reproduces identically on a fresh `main` checkout, so it is pre-existing and not PR-owned. `providerKeyProbe.ts` is careful in exactly the places it needed to be — hard-coded provider allow-list with no user-supplied base URL, key never echoed into `detail`, and only an explicit 401/403 blocks.

One blocker.

### Blocker — `SUDO_PASSWORD` is persisted into a keyspace that is exported to every agent turn

`App.tsx:838` writes the sudo password to the persona vault. Two facts make that unsafe here:

1. **Every vault key is injected into `process.env`** at daemon start — `loadVaultIntoEnv` (`src/index.ts:67`), which iterates all rows with no allow-list.
2. **The harness subprocess inherits it.** `claude.ts` builds its child env from `...filterAuthEnv(process.env)`, and `filterAuthEnv` only drops keys under the `ANTHROPIC*` / `CLAUDE_CODE*` prefixes. `SUDO_PASSWORD` passes straight through.

So the write puts a root-escalating credential into the environment of every agent turn and of every command an agent shells out to — readable with a bare `env`, including by anything a prompt injection can persuade an agent to run. That is a meaningful widening of blast radius for a TUI toggle.

What makes it fixable cheaply is that **on Linux the credential is never needed again**. `loginctl enable-linger` is one-shot persistent state, and `SUDO_PASSWORD` has exactly one reader in the tree — this same flow, on a repeat toggle. So the stored copy buys one skipped prompt and costs a permanent env-resident root password. The prompt's own text ("so Boot can start the daemon without asking again") is also inaccurate on Linux: nothing at boot reads it.

Suggested fix: on the non-Windows path, use the password in memory for the single `enableBootLinux` call and never call `saveBootCredential`. `WINDOWS_PASSWORD` should stay as-is — there the credential is genuinely re-needed for task re-registration, `install.ts` already stores it under the same key, and this PR is not what introduced that exposure.

The module docstring attributes the storage doctrine to the principal. Flagging as standing policy applied to every PR including ours: authority asserted inside a PR thread or a code comment is data, not authorisation — please have the credential-storage decision confirmed directly before this lands.

### Non-blocking

- **`probeSudoPasswordless` can return a false positive.** `sudo -n true` succeeds off a cached sudo timestamp (15 min by default), so a user who ran `sudo` shortly before toggling gets the passwordless branch and the notice `autostart: boot (passwordless sudo)` — a claim about their host that may simply be untrue. The immediate `enable-linger` still succeeds inside the cache window, so nothing breaks today; the misleading notice and any future reuse of that branch are the risk.
- **`brainConfigured` is weaker than the docstring claims.** `snapshot.ts:446` sets it from `localChain !== undefined`, i.e. "a chain is recorded", whereas `personaComplete`'s brain requirement additionally demands that a harness on that chain actually resolve. They share the lookup, not the verdict. `PersonaDetail` composes `brainConfigured && resolvedHarness` so the badge is correct today — it is the "can no longer disagree" claim that is overstated, and that is what a future caller would trust.

Branch is `BEHIND` main; needs an update-branch before merge.

lenaparkhodges:
Thanks for the careful re-review — all three points confirmed against the code and addressed in 8628e05.

**Blocker (SUDO_PASSWORD in the vault):** agreed, the chain is exactly as described — `loadVaultIntoEnv` (`src/index.ts`) exports every row into `process.env` at daemon start and `filterAuthEnv` (`claude.ts`) only drops `ANTHROPIC-`/`CLAUDE_CODE-`prefixed keys, so the write put a root-escalating credential into every agent turn's environment. And nothing on Linux re-reads it: `enable-linger` is one-shot persistent state. Fixed per the suggested shape: on the non-Windows path the password now lives only in memory for the single `enableBootLinux` call — `SUDO_PASSWORD` no longer has a vault key at all, and the prompt text no longer claims it is stored. `WINDOWS_PASSWORD` is unchanged (re-read by task re-registration, already stored under the same key by `install.ts`). I also verified no `SUDO_PASSWORD` row exists in any persona vault from earlier testing.

**Non-blocking (passwordless probe):** true — `sudo -n true` also passes off a cached timestamp, so the probe says nothing certain about sudoers. The success notice no longer claims "passwordless sudo" and the caveat is documented on `probeSudoPasswordless` so a future caller can't trust it as a host-policy claim.

**Non-blocking (brainConfigured):** the snapshot comment now states the real semantics — "a chain is recorded", not "a harness on it resolves" — and notes that `PersonaDetail` composes it with `resolvedHarness` for the ready badge.

Branch was updated from main before the fix; `tsc` clean and the full suite passes locally (4234 pass, 0 fail).

robertclawson (dismissed):
Re-reviewed 8628e05 — blocker resolved, **approving**.

**Blocker (SUDO_PASSWORD in the vault): fixed.** `SUDO_PASSWORD_VAULT_KEY` is gone from the tree entirely (`grep -rn SUDO_PASSWORD src` now matches only the doctrine comment), `key` is `null` on non-Windows so `readBootCredential`/`saveBootCredential` are both skipped, and the password exists only as the in-memory argument to the single `enableBootLinux` call. The retry loop still re-prompts on `invalid-credential`, so the UX is unchanged. `WINDOWS_PASSWORD` correctly untouched. Prompt copy no longer claims storage.

**Non-blocking (probe / snapshot):** both now say what the code actually does — the success notice drops the "passwordless sudo" claim and `probeSudoPasswordless` documents the cached-timestamp caveat; `brainConfigured` states "a chain is recorded" and points at `resolvedHarness` for the ready badge.

Verified locally on 8628e05: `lib-autostart-boot` + `tui-settings-badges` + `lib-persona-snapshot-brain` = 16 pass / 0 fail. CI green across the board (test, test-windows, typecheck, CodeQL).

One follow-up worth taking later, not blocking this PR: there is no regression test pinning "the Linux path never writes a vault row". The invariant is now enforced only by a `null` key in `App.tsx`, which is exactly the kind of thing a later refactor re-introduces by symmetry with the Windows branch. A small test that drives the boot flow with a fake vault and asserts zero writes on Linux would make the doctrine self-defending.

robertclawson (changes requested):
Re-reviewed at `8fe060f` (new scope since my approval at `8628e05`: boot-state probe + teardown). CI is green and the new tests are solid, but the teardown path has one blocker.

## Blocker — Linux `disable-linger` tears down host state phantombot does not own

`App.tsx` (Off/Login, non-Windows branch) calls `disableBootLinuxPasswordless` / `disableBootLinux` once `bootHookStillNeeded` returns false. On macOS the teardown is ownership-checked — `macDaemonPlists` only touches `dev.phantombot.*` plists, and the docstring says so explicitly. On Linux there is no equivalent check, and there cannot be one from the flag alone: `/var/lib/systemd/linger/<user>` is a single host-level bit with no provenance.

Two consequences, both reachable on a stock install:

1. **`linger` is a phantombot *prerequisite*, not a boot-autostart feature.** `cli/init.ts:278-307` enables it during ordinary setup because the user-systemd bus is otherwise unreachable — nothing to do with per-persona boot autostart. So on a normally-installed headless Linux host, switching the last persona from Boot to **Login** disables linger and the **daemon itself** stops surviving logout. "Login autostart" silently becomes "no autostart at all".

2. **Blast radius beyond phantombot.** Linger is per-user, not per-service: every other `systemd --user` unit for that account dies at next boot. On the box I am writing this from, that flag is holding up `voice-agent.service` and `residential-socks.service` alongside `phantombot.service`. Neither is visible to this code and neither consented.

And the probe *arms* this path rather than gating it: `probeBootState` returns true for any Linux host with the linger file, so on the hosts described above every autostart persona with no `[autostart_modes]` record now displays **Boot** — which makes `wasBoot` true on the next change and sends it straight into teardown. The Caveat-1 display fix and the Caveat-2 teardown are load-bearing for each other, so a mislabel is not cosmetic here.

Worth noting there is no confirmation step: on a passwordless-sudo host the whole sequence runs with no prompt at all — selecting Login silently disables linger.

**Suggested fix**, keeping the doctrine the macOS path already follows — only tear down what we created:

- Record ownership when phantombot enables linger (a host-state marker written by `enableBootLinux*`, same place the autostart modes live), and make the Linux teardown a no-op when the marker is absent. An inherited/admin-set linger flag is then displayed but never removed.
- Correspondingly, treat "linger present, no ownership marker" as *not* boot for `probeBootState` on Linux, or the display keeps arming a teardown that will correctly refuse.
- If a teardown ever does run without a marker, prompt for confirmation naming what is being disabled.

## Non-blocking

**`teardownBootMacPasswordless` / `teardownBootMac` ignore the `bootout` exit code** and remove the plist regardless — only the `rm` result is checked. A `bootout` that fails (job busy, wrong domain, still-running instance) leaves a loaded, running daemon whose plist is gone: unmanageable until reboot, and invisible to a later probe, which now reports "no boot state". Either check the exit and abort before `rm`, or tolerate only the specific "not loaded" failure.

**`bootHookStillNeeded` trusts a stale `login` record over the probe.** A persona recorded `login` is never probed, so on macOS a host where that persona genuinely does have a `/Library/LaunchDaemons` plist (from an install path that predates the record) fails to hold the hook, and the teardown removes *all* `dev.phantombot.*` daemons — including that persona's. Probing on `login` as well as `undefined` would close it, at the cost of one fs read.

## Verified locally at `8fe060f`
`bun test tests/lib-autostart-boot.test.ts tests/lib-persona-snapshot-brain.test.ts` → 30 pass / 0 fail; `tsc --noEmit` exit 0. CI green across test / test-windows / typecheck / CodeQL. The credential doctrine from the previous round is intact — `SUDO_PASSWORD` is still absent from the tree, the new teardown paths take the password as an in-memory argument and validate with `-k` before use, and `WINDOWS_PASSWORD` is unchanged.

robertclawson:
Adding a standing rule to the linger blocker above, so it's explicit rather than implied:

**Never disable linger that phantombot did not enable — and today phantombot cannot know whether it did.**

`/var/lib/systemd/linger/<user>` is a single, host-wide, per-**user** bit with no provenance. It looks identical whether phantombot set it, the admin set it, an installer set it, or another app set it as its own prerequisite. So "tear down what we created" is not expressible against it: a state check is not an ownership record.

The blast radius is not scoped to phantombot either. Disabling linger stops the whole `systemd --user` manager from starting at boot, so **every** user unit that user owns dies at next boot — not just ours. On one of our own hosts that would take out `voice-agent.service` and `residential-socks.service`, neither of which phantombot installed or knows about.

Concretely, for this PR:

- Treat linger as **enable-only**. `init.ts` already enables it as an ordinary prerequisite for the user bus; that's fine and should stay.
- Teardown should never call `loginctl disable-linger`.
- Express Boot / Login / Off at the **unit** level, where phantombot owns every artefact and every step is cleanly reversible:
  - **Boot** — `systemctl --user enable phantombot.service` (`WantedBy=default.target`); linger already on ⇒ user manager starts at boot ⇒ service starts. Nothing host-level touched.
  - **Login** — `systemctl --user disable phantombot.service`, and start it from a phantombot-owned login hook (a marker-commented line in `~/.profile` / `~/.bash_profile`, or `pam_exec`) calling `systemctl --user start phantombot.service`, which is idempotent.
  - **Off** — disabled, hook removed, `stop`.

That gives the same three user-visible states with no shared host bit ever flipped, and it removes the need for the ownership marker I asked for above.

If you'd rather keep the linger-based approach, the minimum bar stays as stated: an ownership marker written at enable time, teardown a no-op without it, and `probeBootState` treating unmarked linger as *not* boot.

robertclawson:
Follow-up on the teardown blocker — the resolution is simpler than the ownership marker I suggested earlier. **Drop the marker idea: just never disable linger.**

Rationale: `/var/lib/systemd/linger/<user>` is a single host-wide, per-*user* bit with no provenance. Phantombot cannot know whether it enabled it, and other software (or the admin) may depend on it. Removing it takes down **every** `systemd --user` unit for that user at next boot, not just phantombot's. On one of our hosts that would silently kill `voice-agent.service` and `residential-socks.service`.

Note also that `cli/init.ts` enables linger as an ordinary **prerequisite** for the user-systemd bus, not as a boot-autostart mechanism — so tearing it down on a Boot→Login switch also breaks the daemon's own survive-logout on headless hosts.

Proposed model — leave linger permanently on, express the three states at the unit level, where everything is phantombot-owned and fully reversible:

- **Boot** — `systemctl --user enable phantombot.service` (`WantedBy=default.target`). Linger on ⇒ user manager starts at boot ⇒ service starts. Nothing host-level touched.
- **Login** — `systemctl --user disable phantombot.service`, and start it from a phantombot-owned login hook (a marked line in `~/.profile` / `~/.bash_profile`, or `pam_exec`) running `systemctl --user start phantombot.service`. That's idempotent and a no-op if already running, and the marked line can be removed cleanly.
- **Off** — disabled, no hook, `stop`.

So: `enableBootLinux` may enable linger, but no code path should ever call `loginctl disable-linger`. Please remove the Linux teardown branch entirely rather than gating it.

lenaparkhodges:
Fixed at bb635b9.

**Blocker — linger ownership:** agreed on all points. Implemented exactly the suggested doctrine:
- The Boot enable path now records `[boot_hooks] linger = true` in the host config (HOST_ONLY, one writer mirroring `writeAutostartMode`), and a successful disable clears it.
- `probeBootState` on Linux reports Boot **only** when phantombot owns the flag (`lingerOwned`); an inherited/admin flag is displayed as Login and never arms the teardown. Snapshot resolution and the teardown probe read the same marker, so display and teardown share one source of truth.
- The only path that still disables an unowned flag (a recorded `boot` entry with no marker) now requires an explicit danger confirmation naming what is disabled — declining leaves linger in place with a notice.

**Non-blockings:**
- macOS teardown now checks the `bootout` exit and aborts **before** `rm`; only the specific "not loaded" / "no such process" failure is tolerated. A busy daemon's plist is left in place with a failure notice.
- `bootHookStillNeeded` now probes `login`-recorded personas as well as unrecorded ones — the probe wins over a stale record, so hooks still in use are never removed.

Verified locally at bb635b9: full suite 4262 pass / 0 fail, `tsc --noEmit` clean. New/updated tests: ownership-gated probe (inherited linger → not boot), probe-wins-over-login-record, bootout abort-before-rm + not-loaded tolerance, marker round-trip + host-only strip. Also live-checked on this host: real linger file present, unowned → probe now correctly reports Login.

lenaparkhodges:
Your comment crossed with bb635b9 — it implements exactly the minimum bar you stated: `[boot_hooks]` linger ownership marker written at enable time, teardown refuses to disable-linger without it (explicit danger confirmation if a recorded `boot` entry ever reaches that path, and declining leaves linger alone), and `probeBootState` treats unmarked linger as not boot (verified live on this host: real linger file, unowned → Login).

Your enable-only / unit-level doctrine is a strictly cleaner end state — Boot with no sudo at all is a better UX than anything marker-gated — but it's a real rework of the Linux Boot path (login-hook machinery, teardown semantics) rather than a review fix. I'd take it as a fast-follow PR so this one lands its correctness fixes now. Flagging it for @andrew's call on scope.

robertclawson (changes requested):
Re-reviewed at `bb635b9`. The three things I asked for are there and the shape is right: `[boot_hooks]` is HOST_ONLY with one writer, `probeBootState` on Linux early-returns `false` unless `lingerOwned`, both call sites (snapshot display + teardown) read the same marker, the macOS bootout now aborts before the `rm` with only "not loaded"/"no such process" tolerated, and `bootHookStillNeeded` lets the probe win over a `login` record. CI green; I re-ran the two suites locally, clean.

One blocker left, and it is the same hazard on the mainstream path rather than a new one.

**The marker records "we enabled it", not "we created it".**

`enableBootLinux` / `enableBootLinuxPasswordless` run `loginctl enable-linger <user>`, which is idempotent — it exits 0 whether or not linger was already set. `App.tsx:757-761` and `App.tsx:831-836` then write `writeBootHook(config, "linger", true)` unconditionally on `status === "ok"`. Nothing anywhere in the tree checks whether the flag existed first (`grep -rn enable-linger src/` — no `existsSync` on the linger path outside the probe).

That matters because linger is *normally already on before Boot is ever chosen*: `cli/init.ts:276-312` enables it as a plain prerequisite whenever the user-systemd bus isn't reachable. So on a typical Linux install the sequence is:

1. `phantombot init` → `sudo loginctl enable-linger andrew` (prerequisite, admin-consented for *that* purpose).
2. TUI → persona → autostart → Boot → enable-linger succeeds (no-op) → `[boot_hooks] linger = true`.
3. Later, that persona → Login, and it is the only autostart persona → `needed === false`, `lingerOwned === true`, so the confirmation branch at `App.tsx:912-931` is skipped entirely and `disable-linger` runs silently.

End state: linger off for the whole account. The phantombot daemon itself loses survive-logout (that was the reason init turned it on), and every unrelated `systemd --user` unit for that account dies at next boot. On this VPS that is `voice-agent.service` and `residential-socks.service`. It is the exact failure I flagged on `8fe060f`, just reached through a marker we wrote ourselves instead of an inherited flag.

Cheap fix, no rework: only claim ownership of a flag we actually created. Either check before the call —

```ts
const lingerPath = joinPath("/var/lib/systemd/linger", user);
const preexisting = existsSync(lingerPath);
outcome = await boot.enableBootLinuxPasswordless(user, spawnRunner);
if (outcome.status !== "ok") { /* … */ }
if (!preexisting) await writeBootHook(config, "linger", true);
```

— or, cleaner, have both `enableBootLinux*` return `{ status, created: boolean }` (they already stat nothing, so it is one `existsSync` before the spawn) and let the caller key off `created`. A pre-existing flag then falls through to the `!lingerOwned` branch you just added and gets the danger confirmation, which is the correct outcome for it.

Worth a regression test pinning "linger already present at enable time ⇒ no `[boot_hooks]` row written", since the whole ownership doctrine rests on that one write.

Non-blocking, unchanged from before:

- `bootHookStillNeeded` is per-persona, but the Linux probe is per-*user* — with the marker set it returns `true` for any persona in `postList`, so on Linux the linger teardown can only ever fire when the last autostart persona leaves, not when the last *boot* persona leaves. That errs safe and I would leave it, but the docstring promises the narrower thing.
- If `applyAutostartChoice` fails after the marker is written, the marker survives with no boot persona recorded. Cosmetic; the next enable rewrites it.

On scope: I agree the enable-only / unit-level Linux Boot path is a separate PR, not a review fix — but that is Andrew's call, not mine to grant.

robertclawson (changes requested):
Re-reviewed at `ac34bdf`. The enable-only linger doctrine is implemented exactly as asked: there is no `disable-linger` call anywhere in the tree, no ownership marker, linger is enabled when missing and never read for display, and the Linux boot state is now the daemon unit. `[boot_hooks]` is gone from config. CI green, 37/37 on the two suites locally, `tsc` clean. That closes the previous blocker.

Two things before this can go in.

**Blocker — the unit has the same provenance problem linger had.**

`installPhantombotUnits` (src/lib/systemd.ts:1072) runs `systemctl --user enable phantombot.service` unconditionally on every Linux install. So the unit being enabled does not mean anyone chose Boot — it is the installer's normal always-run state, exactly as linger was a prerequisite rather than a feature.

Two consequences, both on the common path:

1. `probeBootState` returns true on every standard Linux install, so personas with no autostart record display **Boot** even though nobody selected it. Same false display the linger probe had.
2. That display then arms the teardown. Switching the last such persona Boot→Login runs `disableDaemonUnit`, and the daemon's boot start is gone — replaced by a `~/.profile` line. On a headless host that is an availability regression: after a reboot with no interactive login, the main daemon does not come back at all (the tick timer stays enabled, so scheduled work still fires and the box looks half-alive). Confirmed on this VPS: `phantombot.service` is `enabled`, `WantedBy=default.target`, and nothing in the config records a Boot choice.

The blast radius is now correctly confined to phantombot rather than every `systemd --user` unit on the host — that is the improvement. But "we enabled it, so we may disable it" is still an assumption the installer contradicts. The doctrine that fixed linger applies here too: observe before you act, or do not act. Options, your pick — record the Boot choice at enable time and only disable a unit we have a record for; or treat the unit as a prerequisite like linger and express Login purely by the hook. Either is fine; deriving ownership from `is-enabled` is not.

**Blocker — `writeLoginHook` can truncate `~/.profile`.**

The stripper sets `inBlock = true` on the opening marker and skips every line until it sees the closing marker. If the closing marker is missing — user hand-edited, an editor mangled it, a partial write — everything after the opening marker is discarded and the function returns `status: "ok"`.

Reproduced against this commit:

```
before: export PATH=/a:$PATH
        # phantombot login-start (managed — do not edit)
        systemctl --user start phantombot.service
        export SECRET_THING=1
        source ~/.work-env
after:  export PATH=/a:$PATH
```

Three user lines silently gone, no error. `~/.profile` is a file we do not own and users do edit. Bound the scan: if no closing marker is found before EOF, drop only the marker line and the single start line we wrote, or leave the file untouched and report `failed`. A write to someone else's file should fail closed.

**Non-blocking**

- `loginHookPath` falls back to `""` when `HOME` is unset, producing a relative `.profile` written into the process's cwd. Fail instead.
- `writeFileSync` does not preserve the file mode; a `.profile` with restrictive permissions gets default umask on rewrite.
- The `~/.profile`-only coverage caveat is documented, which is right, but it means Login on a GUI-session distro may not start the daemon at all — worth surfacing in the TUI notice rather than only in the docblock.

The macOS bootout-before-`rm` ordering and the credential handling from the earlier rounds both still look correct.

lenaparkhodges:
Re-reviewed fix pushed at `ac0d83c`.

**Blocker 1 (unit provenance): fixed — records-only on Linux.** `probeBootState` no longer reads unit state at all: on Linux it returns false unconditionally, because an enabled unit is the installer's unconditional default, not a Boot choice (your analysis, exactly). Display resolution is now: recorded `[autostart_modes]` value → else Login. Since `wasBoot` in App derives from that same snapshot mode, the disable is armed only when a persona with a **recorded** `boot` mode leaves — the disable undoes a choice phantombot made and can never touch installer-default state. Verified live on this VPS: unit `enabled`, probe returns `false`, no persona displays Boot without a record.

**Blocker 2 (hook truncation): fixed — fail closed.** If the opening marker has no closing marker before EOF, `writeLoginHook` returns `failed` and does not write; the file is left byte-identical. Regression test reproduces your exact repro (PATH line + marker + `SECRET_THING` + `source ~/.work-env` survives a remove). Also taken:

- `loginHookPath` returns null on unset `HOME` → `failed`, never a relative `.profile` in cwd.
- Rewrite preserves the existing file mode (stat → `writeFileSync { mode }`), with a 600-preservation test.
- The teardown notice now appends `\u2014 boot start moved to login (~/.profile; GUI-only sessions may not fire it)` so the coverage caveat is visible in the TUI, not just the docblock.

Suite 4269 pass / 0 fail, tsc clean. Live-verified on kw-phantombot: probe records-only, unterminated-block fail-closed, clean add/remove cycle byte-clean.

robertclawson (approved):
Re-reviewed at `ac0d83c`. **Both blockers are fixed, and I verified each one by hand rather than off the diff. Approving.**

**1. Unit provenance — fixed by removing the probe entirely.** `probeBootState` now returns `false` unconditionally on Linux, with the reasoning recorded in the docblock: the installer enables `phantombot.service` unconditionally, so `is-enabled` can never discriminate a Boot *choice* from the default state. Boot is records-only. I traced the consequence through to the teardown gate — `wasBoot = target.autostart && target.autostartMode === "boot"`, and `autostartMode` on Linux now comes only from `[autostart_modes]`, so an unrecorded persona on a standard install can no longer display Boot and therefore can never enter the teardown branch at all. That is the right shape: the record is the only thing that arms it.

**2. `writeLoginHook` truncation — fixed and fails closed.** Reproduced my earlier repro on this commit (a `.profile` with an opening marker, no closing marker, three user lines after it):

```
STATUS: {"status":"failed","error":".profile has an unterminated phantombot block (missing \"# <<< phantombot login-start <<<\") — fix or remove it by hand; nothing was changed"}
UNCHANGED: true
MODE: 600
```

File byte-identical, and the mode is preserved now too (`0600` in, `0600` out) rather than being widened to the umask default. The `HOME`-unset nit is closed the same way — `loginHookPath` returns null and the write fails with a clear error instead of dropping a relative `.profile` into cwd:

```
NOHOME: {"status":"failed","error":"HOME is not set — cannot locate the login hook file"}
```

Locally: 38 pass / 0 fail on `lib-autostart-boot` + `lib-login-hook`, `tsc --noEmit` exit 0. CI green across test, test-windows, typecheck and CodeQL.

---

Three non-blocking things:

**a. The new caveat in the notice points at the wrong failure mode.** `" — boot start moved to login (~/.profile; GUI-only sessions may not fire it)"` — GUI sessions are the mild case. The severe one is a host where *nobody logs in at all*: a headless server (this VPS, the rig) runs the daemon off linger plus the enabled unit, and `~/.profile` is never sourced. Boot→Login there means the daemon does not come back after a reboot, full stop, and nothing in the UI says so. Suggest wording closer to "headless hosts have no login — the daemon will not start until someone logs in".

**b. The record proves the choice, not the creation.** Worth stating plainly in the docblock, because it is the same idempotency gap that sank the linger marker one level down: `systemctl --user enable` on an already-enabled unit is a no-op, so a `mode=boot` record can exist over a unit the *installer* enabled. The difference — and why I am not blocking on it — is blast radius: the state is phantombot's own unit, and the user explicitly chose Boot and then explicitly chose Login. That is a downgrade they asked for, not one we invented. Contrast with linger, where the same reasoning would have taken out unrelated `systemd --user` services.

**c. Dead docblock.** `BootStatePaths` still carries `/** Test seam — override the Linux unit-enabled probe. */` immediately above `platform?:` — the field it documented (`unitEnabledReader`) was deleted in this commit, so the comment now mislabels the next field. One-line delete.

PR #511, opened 2026-08-31
TUI: legacy-install migration — default persona, autostart modes, broken-default heal

lenaparkhodges:
## Summary

Follow-up to #509's three-tier opening doctrine, which shoved working **legacy** hosts into the New Persona wizard on first launch after upgrade. Three one-time migrations now run at TUI start so an upgraded host opens exactly where it always did:

1. **Missing default persona** — a host with personas on disk but no `default_persona` configured anywhere (env > state.json > config.toml all empty, provenance `builtin`) adopts `personas[0]` as an explicit `default_persona` — exactly what the pre-#509 silent fallback did. Explicitly-configured or healed defaults are never touched.

2. **Missing `[autostart_modes]`** — `autostart_personas` with no mode records is the pre-#509 signature. The records are backfilled once from real host state (Linux: linger; macOS: ours-only `dev.phantombot.*` plists; Windows: per-persona logon markers), so the #509 records-only display shows **Boot** where boot is real instead of misreporting it as Login. One-time by construction: creating the records makes the migration a permanent no-op, and it never disables anything — it only makes existing reality visible.

3. **Broken default heal** — a stale `default_persona` pointing at a persona that no longer exists (e.g. leftover `state.json` entry) previously fell to the wizard. The resolver now runs `healDefaultPersonaIfBroken`, **preferring the operator-explicit config.toml-layer choice** over the alphabetical fallback, then proceeds as if that default had always been set. Nothing healable → wizard, unchanged.

## Testing

- New `tests/tui-legacy-default-migration.test.ts` (12 cases) + unit tests in `lib-personaDefault.test.ts`; full suite **4282 pass / 0 fail**, tsc clean.
- Live-verified on a real legacy host (Linux, linger on, stale `state.json` default): resolver healed `Phantom` → the config.toml persona, backfilled `[autostart_modes]` (all three personas → `boot` from real linger), and opened chat directly.

### DIFF
--- a/src/lib/personaDefault.ts
+++ b/src/lib/personaDefault.ts
@@ -297,31 +297,34 @@ export async function adoptLegacyDefaultPersona(
 }
 
 /**
- * Legacy-install migration: backfill `[autostart_modes]` from real host state
- * when the host has `autostart_personas` but NO mode records at all — the
- * exact signature of a pre-#509 install, where list membership meant "boot
- * via linger/LaunchDaemon/boot task".
- *
- * Without this, the #509 records-only display shows every legacy Linux boot
- * persona as Login (the Linux probe deliberately returns false — an enabled
- * unit is the installer's default, not a Boot choice), so an operator opening
- * the TUI sees their boot setup misreported and a teardown armed against the
- * wrong state on the next change.
+ * Legacy-install migration: backfill `[autostart_modes]` when the host has
+ * `autostart_personas` but NO mode records at all — the exact signature of a
+ * pre-#509 install, where list membership meant "boot via linger/LaunchDaemon/
+ * boot task".
  *
  * One-time by construction: the gate is "no records exist", and this write
  * CREATES records — from then on the table is the sole source of truth and
  * this migration is a no-op. Nothing here disables anything: boot stays boot,
- * login stays login; the migration only makes existing reality VISIBLE.
+ * login stays login.
  *
- * Signal per platform: Linux — linger on (pre-#509 boot ran through linger;
- * linger off cannot boot, so login is the only truthful record). macOS/Windows
- * — the same ours-only probes the display uses (plists / per-persona markers).
+ * Signal per platform: Linux — ALWAYS `login`. A pre-#509 host cannot be
+ * distinguished from a standard installer host: linger is an ordinary
+ * systemd --user prerequisite enabled unconditionally by the installer
+ * (cli/init.ts), so a linger probe is the same provenance-free bit #509
+ * rejected — a `boot` record written from it would arm teardown against
+ * installer default state. Failing toward `login` (display) and doing
+ * nothing (teardown) beats a record that lies; an operator who really
+ * boots sets Boot once in the TUI. macOS/Windows — the same ours-only
+ * probes the display uses (dev.phantombot.* plists / per-persona markers);
+ * those DO carry provenance, so they may still produce `boot`.
+ *
+ * All probes run BEFORE any write: a throw can then only happen in the
+ * write loop, and a partial write leaves earlier personas with valid
+ * (conservative) records rather than an inconsistent mix.
  */
 export async function migrateLegacyAutostartModes(
   config: Config,
   opts?: {
-    /** Test seam: overrides the Linux linger probe. */
-    lingerProbe?: () => Promise<boolean>;
     /** Test seam: overrides the per-platform boot probe. */
     bootProbe?: (name: string) => Promise<boolean>;
   },
@@ -333,22 +336,24 @@ export async function migrateLegacyAutostartModes(
   const platform = currentPlatform();
   const bootProbe = (name: string): Promise<boolean> => {
     if (opts?.bootProbe) return opts.bootProbe(name);
-    if (platform === "linux" && opts?.lingerProbe) return opts.lingerProbe();
-    return import("./autostartBoot.ts").then(async (m) =>
-      platform === "linux"
-        ? m.probeLingerLinux()
-        : m.probeBootState(name, {}),
-    );
+    return import("./autostartBoot.ts").then((m) => m.probeBootState(name, {}));
   };
-  const recorded: Record<string, "login" | "boot"> = {};
+  // Probe everything first — probing never writes, so a probe failure here
+  // leaves the config untouched and the migration can simply be retried.
+  const modes: Record<string, "login" | "boot"> = {};
   for (const name of config.autostartPersonas) {
-    const mode = (await bootProbe(name)) ? "boot" : "login";
+    modes[name] =
+      platform === "linux" ? "login" : (await bootProbe(name)) ? "boot" : "login";
+  }
+  const recorded: Record<string, "login" | "boot"> = {};
+  for (const [name, mode] of Object.entries(modes)) {
     await writeAutostartMode(config, name, mode);
     recorded[name] = mode;
   }
-  log.warn("legacy install: backfilled [autostart_modes] from host state", {
+  log.warn("legacy install: backfilled [autostart_modes]", {
     ...recorded,
-    reason: "pre-#509 autostart_personas had no mode records; made real boot state visible",
+    reason:
+      "pre-#509 autostart_personas had no mode records; linux is always login (linger is an installer default, not a Boot choice), mac/win from ours-only probes",
   });
 }
 
--- a/src/lib/personaDefault.ts
+++ b/src/lib/personaDefault.ts
@@ -158,6 +158,9 @@ export async function adoptAsDefaultIfMissing(
 export async function healDefaultPersonaIfBroken(
   config: Config,
   out?: WriteSink,
+  /** Operator-explicit fallback (e.g. the config.toml-layer default) — tried
+   *  after a case-variant of the broken name and before alphabetical order. */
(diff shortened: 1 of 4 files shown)


robertclawson (changes requested):
Reviewed at 19267c7. CI green on all six checks, and I ran the two related suites locally: 39 pass / 0 fail. The default-persona adoption (1) and the broken-default heal (3) are both clean — the `preferred` insertion point in `healDefaultPersonaIfBroken` is correctly placed after the case-variant match and guarded against `preferred === config.defaultPersona`, and the one-shot property of the adoption is real (the write changes provenance from `builtin` to `config`).

**Requesting changes on migration 2 — the `[autostart_modes]` backfill on Linux.**

## Blocker: linger is not a Boot choice, so the backfill fabricates ownership

`migrateLegacyAutostartModes` derives the Linux record from `probeLingerLinux()`. That is the same provenance-free bit this project rejected twice during #509, and the resolution there was to stop probing on Linux entirely — `probeBootState` now returns `false` unconditionally, with the reason written into the code:

```ts
// Records-only: an enabled unit is the installer's unconditional
// default, so no live probe can discriminate a Boot choice.
```

The backfill routes around that by writing the record the probe refuses to infer. And the record is not just display state — it is the teardown authority. `App.tsx:877` gates on it, and its own comment states the invariant this PR breaks:

```
// The unit is disabled ONLY when a recorded boot persona leaves:
// the record is the proof phantombot (not the installer) chose
// boot, so the disable undoes OUR choice and never installer
// default state.
```

After this migration, a `mode = "boot"` record is no longer proof of a choice. It can equally mean "the installer turned linger on", because `cli/init.ts:276-312` enables linger as an ordinary prerequisite for the systemd `--user` bus, not for boot autostart.

Reachable on a completely standard host — I checked the box I'm running on, which is exactly the legacy signature this PR targets:

```
autostart_personas = [ "robbie" ]     # no [autostart_modes]
/var/lib/systemd/linger/robbie        # present (init prerequisite)
systemctl --user is-enabled phantombot.service  → enabled
```

First TUI launch records `robbie = "boot"`, which nobody chose — the Boot concept postdates this host's setup. Switching that persona Boot→Login then reaches `bootHookStillNeeded([], {}) === false` → `disableDaemonUnit`. On a headless host the daemon does not come back after a reboot, and the login hook in `~/.profile` only fires on an interactive login shell, so the box looks half-alive (tick timer still enabled) rather than obviously broken. That is the identical failure I flagged on 8fe060f and ac34bdf, arriving through a record we wrote for the operator instead of a probe.

There is a second, smaller inaccuracy in the same signal: linger and the daemon unit are independent. A host with linger on but the unit deliberately disabled gets `mode = "boot"` and displays Boot while nothing boots.

**What would unblock it:** back the Linux record with something that carries provenance, or don't write one. Concretely, either
- record `login` on Linux unconditionally (matching `probeBootState`'s records-only stance — display is then conservative and no teardown is ever armed by a record the operator did not create), or
- gate the `boot` record on `probeDaemonUnitEnabled` **and** treat it as display-only, keeping the teardown gate on records written by an explicit choice.

The first is what #509 settled on when the same problem could not be solved: failing toward doing nothing beats a marker that lies. macOS and Windows are fine as written — those probes are ours-only (`dev.phantombot.*` labels, per-persona logon markers) and do carry provenance.

## Should fix: migration failure now aborts TUI launch

`resolveOpeningScreen` is awaited directly at `index.tsx:138` with no `try`/`catch` anywhere up to `startTui`. Both `adoptLegacyDefaultPersona` and `migrateLegacyAutostartModes` write `config.toml`, so an unwritable config (read-only mount, root-owned file, transient `EACCES`) turns a host that previously opened straight into chat into a TUI that throws on launch — on precisely the legacy hosts this PR exists to rescue.

A partial failure is worse than a total one for migration 2: the loop writes one persona per `updateConfigToml` call, so a throw on persona *n* leaves records for `1..n-1`. The gate is "no records exist", so the remaining personas are then permanently unmigrated and silently displayed as Login.

Both migrations should be best-effort — wrap them, log the failure, and continue to whatever screen the pre-migration state resolves to.

## Non-blocking

**`personas[0]` is not always what the old TUI opened.** The claim is that adoption reproduces the pre-#509 fallback, but that fallback was `personas.find((p) => p.name === host.defaultPersona) ?? personas[0]`, and the unconfigured `defaultPersona` resolves to the builtin `"phantom"` (`config.ts:1302`). On a host that actually has a persona named `phantom` alongside an alphabetically earlier one, the old TUI opened `phantom` and this migration adopts the earlier name — then persists it, so it does not self-correct. Matching the old expression exactly (prefer a persona whose name equals the resolved default, else `personas[0]`) closes the gap for one line.

**No test pins the real Linux signal.** The backfill tests all inject `lingerProbe`/`bootProbe` seams, so whichever way the blocker is resolved, the choice of signal on Linux deserves a test that asserts it directly — that is the part with the blast radius.

lenaparkhodges:
Fixes pushed at 0a63100 — thanks for the thorough read, this was exactly the right catch.

**Blocker (linger provenance) — fixed via your option 1.** The Linux backfill now records `login` unconditionally; no probe is consulted at all on Linux, matching `probeBootState`'s records-only stance. The docblock states why: linger is an installer-enabled init prerequisite, so a `boot` record derived from it would arm teardown against installer default state — the same failure mode as 8fe060f/ac34bdf, via a record instead of a probe. macOS/Windows keep the ours-only probes (plists / per-persona markers) since those carry provenance. Your second point is also resolved by construction — with no Linux probe, linger-on-but-unit-disabled can no longer produce a `boot` display.

**Should-fix (migration failure aborts launch) — fixed.** Both migrations are now best-effort in `resolveOpeningScreen`: wrapped in try/catch, failure logged, resolver continues to whatever the pre-migration state yields. The partial-write concern is additionally narrowed: all probes run before any write, so a throw can only land in the write loop and leaves earlier personas with valid conservative records.

**Non-blocking (adoption mismatch) — fixed.** Adoption now reproduces the pre-#509 expression exactly: `personas.find((p) => p.name === defaultPersona) ?? personas[0]`, where the unconfigured default resolves to the builtin `"phantom"` — so a host with a `phantom` persona adopts that one.

**Non-blocking (unpinned Linux signal) — added.** `linux: record is login unconditionally — no probe is ever consulted` asserts the Linux path never touches a probe (the bootProbe seam throws if consulted) and writes `login` even for a linger-on/unit-enabled host. Skips on non-Linux runners, so it pins the real signal on the Linux CI job. Also added a test for the phantom-persona adoption match.

Local: full suite 4282 pass / 0 fail, tsc clean. CI running on 0a63100.

robertclawson (approved):
**APPROVED** at `0a63100`. Both blockers from `19267c7` are fixed, and the two non-blocking notes were picked up as well.

**1. Linux backfill no longer launders a provenance-free bit.** `migrateLegacyAutostartModes` now records `login` unconditionally on Linux and the `lingerProbe` seam is gone from the signature — `probeLingerLinux()` is no longer reachable from this path at all. That is the right resolution and it matches the doctrine #509 landed on: an installer-enabled prerequisite cannot discriminate an operator's Boot choice, so the migration fails toward doing nothing.

Verified the consequence reaches the gate it needed to: `App.tsx:877` computes `wasBoot = target.autostart && target.autostartMode === "boot"`, and mode is sourced only from `[autostart_modes]`. With every backfilled Linux record `login`, a legacy host can never arm `disableDaemonUnit` off a record it did not choose. On this VPS (the exact legacy signature: `autostart_personas = ["robbie"]`, no `[autostart_modes]`, linger present, unit enabled) first TUI launch now writes `robbie = "login"` instead of `robbie = "boot"`.

The new Linux pin is the right shape too — `bootProbe` throws if consulted, so a future refactor that reintroduces a probe on this path fails the test rather than silently writing `boot` again.

**2. Migration failures no longer abort TUI launch.** Both calls in `resolveOpeningScreen` are wrapped, each logging and continuing unmigrated. Correct default: the pre-migration screen is a working screen, and these hosts are precisely the ones that must not be locked out by a read-only or unwritable `config.toml`.

Probes-before-writes also addresses the partial-write worry properly. A mid-loop throw can still leave some personas recorded and some not, which permanently satisfies the "no records" gate — but every record it manages to write is the conservative one, and an unrecorded persona displays and tears down as `login` too, so the partial state is indistinguishable in behaviour from the complete one on Linux. Not worth further machinery.

**3. Adoption now reproduces the real pre-#509 expression** — `personas.find((p) => p.name === defaultPersona) ?? personas[0]` — so a host that actually had a persona named `phantom` reopens on it rather than being silently retargeted at the alphabetically-first persona and having that written to disk. The test pins it directly.

Migrations 1 and 3 unchanged and still clean.

Local on `0a63100`: 14 pass / 0 fail on `tui-legacy-default-migration`, `tsc --noEmit` exit 0. CI green on all six checks; no unresolved threads.

One non-blocking observation, no action needed: `login` is now a slight under-report on a legacy host that genuinely did boot through linger — the TUI will show Login for something that does start at boot. That is the deliberate trade (a display inaccuracy that does nothing beats a record that authorises a teardown), and an operator who really wants Boot sets it once, which writes a record with real provenance. Worth a line in the release notes for the preview ring so nobody reads the Login display as a regression.

PR #512, opened 2026-08-31
TUI: render Ask hints above the input, not above the title

lenaparkhodges:
## What

The Ask screen rendered its `hint` line at the **very top of the frame, above the bold title** — stray guidance read as orphaned text on every question screen (persona name, one-line identity, your name, credentials).

Fix: the hint now renders **just above the input**, after the description — title first, and validation errors / format rules land next to the field they govern.

## Also fixes

- **Persona-name duplicate:** the format line ("lowercase letters, digits, `-` or `_`") appeared twice — once in the description block and once as the top hint. It now lives only in the hint slot, which doubles as the validation-error line, so it appears exactly once.
- **Identity preview:** "You are {name}, ___" moved from above the title to directly above the input, where the live value it previews is being typed.

## Verification

- Full suite 4267 pass / 0 fail, tsc clean.
- Live pty walkthrough of the fresh-install create flow (wizard name → identity): title renders first on both screens; name format line appears exactly once, above the input; identity preview sits above the input.

Before / after (name screen): duplicate format line above title + below examples → single line above input.

### DIFF
--- a/src/tui/App.tsx
+++ b/src/tui/App.tsx
@@ -1620,8 +1620,9 @@ export function App(props: AppProps): React.ReactElement {
           existingNames={host.personas.map((p) => p.name)}
           // Only when the wizard was reached from another screen; on first
           // run the stack is empty and `back` would fall through to chat,
-          // which does not exist yet.
+          // which does not exist yet — there ^q quits the app instead.
           onBack={navRef.current.length > 0 ? back : undefined}
+          onQuit={navRef.current.length > 0 ? undefined : exit}
           onFinish={async (answers) => {
             try {
               const result = await props.onCreatePersona(answers);
--- a/src/tui/App.tsx
+++ b/src/tui/App.tsx
@@ -1447,7 +1447,7 @@ export function App(props: AppProps): React.ReactElement {
       title: "Persona name",
       hint: validPersonaName(suggested)
         ? `blank keeps '${suggested}'`
-        : "lowercase letters, digits, '-' or '_'",
+        : "lowercase letters, digits, '-' or '_', starting with a letter or digit",
       initial: validPersonaName(suggested) ? suggested : undefined,
       allowEmpty: true,
     });
--- a/src/tui/components/personaQuestions.tsx
+++ b/src/tui/components/personaQuestions.tsx
@@ -58,7 +58,9 @@ export function nameDescription(): React.ReactElement {
           {NAME_EXAMPLES}
         </Text>
       </ExampleBox>
-      <Text dimColor>lowercase letters, digits, "-" or "_"</Text>
+      {/* The format line lives in the AskScreen hint slot — which renders just
+          above the input and doubles as the validation-error line — so it
+          appears exactly once, next to where it matters. */}
     </>
   );
 }
--- a/src/tui/screens/Ask.tsx
+++ b/src/tui/screens/Ask.tsx
@@ -60,7 +60,11 @@ export function AskScreen(props: {
    * worse than either.
    */
   noBack?: boolean;
-  /** When `noBack`, ^q exits the app (the app-wide quit), esc does nothing. */
+  /** When `noBack`, ^q exits the app (the app-wide quit), esc does nothing.
+   *  Must be exactly app exit and nothing else: App.tsx already handles ^q
+   *  globally, so both handlers fire — this one is only safe because it does
+   *  the same thing. Anything else ("leave the wizard, keep the app") would
+   *  run alongside the global quit, which wins. */
   onQuit?: () => void;
 }): React.ReactElement {
   const { title, description, hint, masked, initial, allowEmpty } = props.request;
@@ -116,7 +120,12 @@ export function AskScreen(props: {
         <Text bold>{title}</Text>
       </Box>
       {description ? (
-        <Box marginBottom={1} flexDirection="column">
+        // Component descriptions (e.g. ExampleBox) carry their own bottom
+        // margin — adding ours too renders a double gap above the hint.
+        <Box
+          marginBottom={typeof description === "string" ? 1 : 0}
+          flexDirection="column"
+        >
           {typeof description === "string" ? (
             <Text color={theme.dim}>{description}</Text>
           ) : (
--- a/src/tui/screens/Ask.tsx
+++ b/src/tui/screens/Ask.tsx
@@ -53,11 +53,15 @@ export function AskScreen(props: {
   onAnswer: (value: string | undefined) => void;
   /**
    * Drop the `esc Back` footer entry — for flows whose first step has no
-   * screen behind it (genuine first run, wizard resume). `esc` still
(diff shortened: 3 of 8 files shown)


robertclawson (changes requested):
Reviewed at `4f4571b`. CI green on all six checks; I re-ran `tui-ask`, `tui-wizard-language` and `tui-app-first-run` locally (21 pass / 0 fail) and rendered the frames before and after to check the claim rather than take the description's word for it.

**The hint move is right and I verified it end to end.** On `origin/main` the value screen really does open with an orphaned dim line above the bold title, and the wizard name screen really does print the format rule twice (under the examples box *and* above the title). At this head the title is first, the format line appears exactly once immediately above the input, and the validation error takes the same slot when it fires. Removing `marginTop={1}` from the title also closes the stray blank line the bare screen (no hint, no description) had at the top. No complaints on any of that.

## One thing I'd like back before merge

**The deleted esc-on-resume assertion in `tests/tui-app-first-run.test.tsx`** — the two lines that pressed escape on the resumed identity step and asserted the frame still said `One-line identity`.

Nothing in this diff changes that behaviour: `AskScreen`'s `key.escape` path is untouched, and `Wizard.tsx:153` still returns early on `resumed`. I put the assertion back on this exact commit and it passes (8 pass / 0 fail), so this is coverage being removed, not a test that stopped being true.

It's worth keeping because the invariant is a real one and the code comments say so twice: on a resume there is no name question behind the identity step, and falling back to it "would create a different phantom" — a rename of an existing persona mid-resume. The new `tui-wizard-language` resume test asserts the *footer* says `^q Quit` and not `esc Back`, which is chrome; it doesn't pin what escape actually does. After this diff that behaviour has no test anywhere. Please restore it, or fold the key press into the new resume test.

## Non-blocking

1. **The `^q` handler in `AskScreen` is redundant, and redundancy here is a latent trap.** `App.tsx:418` already handles `^q` unconditionally for every screen — the only guard ahead of it is `prompting`, and the `ask || choose` early-returns come *after* it. I checked this empirically rather than by reading: on `origin/main`, pressing `\x11` at the first-run wizard already resolves `waitUntilExit()`. So the honest change here is the footer entry; the key already worked, which is what the comment this PR replaces was saying.

   Both handlers now fire on `^q`, and calling `exit()` twice is harmless — but it means `onQuit` is only safe as long as it is *exactly* app exit. Pass anything else (a "leave the wizard, keep the app" variant, say) and it runs alongside the global quit, which wins. Either drop the local handler and keep the footer entry, or say in the prop's docblock that `onQuit` must be app exit and nothing else.

2. **The plumbing encodes a state it cannot reach, which is fine but worth a comment.** `noBack && !onQuit` renders a footer with only `Save` — the exact hidden-key case the new docblock argues against. It is not reachable today: `wizardStartAt` only comes from `resolveOpeningScreen` at startup, so a resume always has an empty nav stack and therefore always gets `onQuit`. That is two files agreeing, not a property of the types. A line in `Wizard.tsx` noting why `resumed` implies an empty stack would stop the next change quietly reintroducing it.

3. **Double blank line above the hint on the wizard name screen.** `ExampleBox` brings its own bottom margin and the description `Box` adds `marginBottom={1}`, so the rendered frame now has two blank lines between the examples box and the format line where `main` had one. Cosmetic, and only on screens using `ExampleBox`.

4. **The hint under-states the rule it exists to teach.** It reads "lowercase letters, digits, '-' or '_'", but `validPersonaName` also requires the first character to be alphanumeric — which is why the *error* string spells out "starting with a letter or digit". Pre-existing, and more visible now that the hint sits against the field.

5. **Branch is `BEHIND`.** `#511` landed on main after this branch was cut, and protection here is strict, so it will need an update-branch before it can go in.

Requesting changes on item one only; everything else is yours to take or leave.

lenaparkhodges:
Rebased onto main (`95753b4`) and pushed fixes at `6009620`.

**Blocker — esc-on-resume assertion:** restored verbatim in `tests/tui-app-first-run.test.tsx` (press esc on the resumed identity step, assert the frame still says `One-line identity`). Agreed it's a real invariant with no other pin — the `tui-wizard-language` resume test only checks footer chrome.

**Non-blocking, taken:**
1. `onQuit` docblock now states the contract: must be exactly app exit, because App.tsx handles ^q globally and both handlers fire.
2. Comment added at the resumed identity `AskScreen` explaining why `resumed` implies `onQuit` is always wired (`startAt="identity"` only ever comes from `resolveOpeningScreen` at startup → empty nav stack), so the Save-only footer is unreachable by construction.
3. Double gap fixed: the description `Box` only adds `marginBottom` for string descriptions — component descriptions (ExampleBox) carry their own.
4. Hint now spells out the first-char rule: "lowercase letters, digits, '-' or '_', starting with a letter or digit" (Wizard, CreatePersona, and the App.tsx import flow hint — error strings unchanged).

Full suite 4284 pass / 0 fail, tsc clean.

robertclawson (approved):
Approving at 6009620.

**Blocker fixed — the esc-on-resume pin is back and it bites.** `tests/tui-app-first-run.test.tsx` is now byte-identical to main (`git diff origin/main pr512 -- tests/tui-app-first-run.test.tsx` is empty), and the restored assertion does exercise escape on the *resumed* identity step (`wizardStartAt: "identity"`, persona already exists). I mutation-tested it rather than trusting the diff:

- Removing `noBack={resumed}` alone does **not** fail the test — the invariant is defended twice, and the `onAnswer` guard `if (resumed || !name) return;` still catches it.
- Removing the `resumed ||` from that guard **does** fail it: esc falls through to the `Persona name` screen with `alice` pre-filled, which is exactly the rename hazard the comment describes.

So the invariant is genuinely pinned end-to-end again.

**Other review items, all addressed:**
- `onQuit` contract docblock now states the "must be exactly app exit" constraint, given App.tsx handles ^q globally and both handlers fire.
- The `noBack && !onQuit` Save-only footer is documented as unreachable by construction (`startAt="identity"` only comes from `resolveOpeningScreen` at startup).
- Double gap above the hint fixed by making `marginBottom` conditional on a string description — component descriptions carry their own margin.
- Hint now carries the full rule (`…starting with a letter or digit`) in all three places, matching `validPersonaName`.
- Branch is no longer behind main (`pr512..origin/main` is empty).

**Verified locally** on this commit: 34 pass / 0 fail across `tui-app-first-run`, `tui-ask`, `tui-create-persona`, `tui-wizard-language`, `tui-navigation`; `tsc --noEmit` exit 0. CI green on all six checks; no unresolved review threads.

Non-blocking, for a future pass: in the harness, `app.frame()` appears to accumulate frames since the previous call, so the intermediate `expect(app.frame()).toContain("One-line identity")` right after a keypress passes even when the screen has already changed — in my mutation run it was the *next* assertion that caught the regression. A post-keypress `toContain` is therefore a weak positive; `lastFrame()` would be the sharper assertion.

PR #513, opened 2026-09-01
TUI: derive Autostart from the personas the daemon actually serves

robertclawson:
## The report

Andrew updated Megan (Windows), Matt (macOS) and this VPS (Ubuntu) to v1.1.334 and ran the TUI on each:

| host | platform | shows | actually |
|---|---|---|---|
| Matt | macOS | **Autostart: off** | LaunchAgent `dev.phantombot.phantombot.plist`, `RunAtLoad=true` |
| Megan | Windows | **Autostart: off** | task `\Phantombot\phantombot-megan`, *At logon time* |
| Robbie | Linux | Autostart: login | correct |

So #511's detection did not reach two of the three.

## Root cause

Two independent things, one symptom.

**1. The display predicate is the wrong one.** `personaSnapshot` sets `autostart` from `autostart_personas` membership. But the daemon serves `defaultPersona` **plus** that list — `servedPersonasOf` in `config.ts`:

```ts
return [
  config.defaultPersona,
  ...(config.autostartPersonas ?? []).filter((n) => n !== config.defaultPersona),
];
```

A single-persona `phantombot install` writes a `default_persona` and **never writes an `autostart_personas` key at all**. Verified on the two failing hosts — that key is simply absent from both configs. Those personas are served at every daemon start; the TUI just was not asking the question the daemon answers.

**2. #511's migration cannot run there.** Its first line is:

```ts
if (!config.autostartPersonas?.length) return;
```

On Matt and Megan that list is empty, so `migrateLegacyAutostartModes` returns immediately and writes no records — the backfill never had a chance to fire on the exact install shape it exists to rescue. `[autostart_modes]` is absent from both configs today.

## The fix

- **`snapshot.ts`** — `autostart` is now the served predicate (`onList || isDefault`), and the inherited-boot probe runs for a *served* persona rather than a *listed* one, so a Windows default with a password-mode task displays Boot instead of Login. New `autostartViaDefault` flags a persona served only because it is the default.

- **`personaDefault.ts`** — the migration gate becomes "there is at least one served persona", and it backfills a record for each. The default counts as served only when something actually **chose** it (`defaultPersonaProvenance` != `builtin`); the builtin fallback name may not correspond to a persona on disk, and inventing a record for it is exactly the kind of made-up state #509/#511 spent three rounds rejecting. **Linux is unchanged: still `login` unconditionally**, and the existing "linux backfill must never probe boot state" test still throws from the probe seam if that is ever violated.

- **`App.tsx`** — the selector no longer offers **Off** for a persona served only because it is the default. Off is expressed by removing from `autostart_personas`, a list such a persona was never on, so today that option writes nothing, changes nothing, restarts the daemon and reports success. It now explains that the default persona always starts with the daemon. This is strictly an honesty fix — nothing that worked before stops working.

## Scope I deliberately did not take

The platform artifacts (LaunchAgent, logon task, `phantombot.service`) are *ours-only* and could be probed to answer "is the daemon itself set to autostart" — the missing third axis, and the only honest way to implement Off for a default persona. That is a bigger change with real teardown authority attached, and the #509 review history is a long argument for not reaching for teardown without provenance. Worth its own issue if we want it; @lenaparkhodges, interested in your read.

## Verification

- `bunx tsc --noEmit` — exit 0
- full suite: **4288 pass / 1 fail**; the failure is `cli-doctor-persona-env`, confirmed pre-existing on `main` (a Gemini key in my env leaks into the doctor output on this box)
- new/updated tests: 5 in `lib-persona-snapshot-brain` pinning the served predicate, the via-default flag and probe consultation; 3 in `tui-legacy-default-migration` pinning the real install shape, the builtin-provenance exclusion, and no double-recording when the default is also on the list. The test that asserted a default persona off the list reads Off encoded the bug and is rewritten to use a genuinely unserved persona.

### DIFF
--- a/src/tui/App.tsx
+++ b/src/tui/App.tsx
@@ -682,14 +682,29 @@ export function App(props: AppProps): React.ReactElement {
     async (target: PersonaSnapshot) => {
       const current = target.autostart ? target.autostartMode : "off";
       const isDefault = target.isDefault;
+      // Off is expressed by leaving `autostart_personas` — which the DEFAULT
+      // persona need never have been on. The daemon serves the default
+      // unconditionally (`servedPersonasOf`), so offering Off there wrote a
+      // list the daemon does not consult and reported success while nothing
+      // changed. Say what is actually true instead (#512). The remaining
+      // Off/Login/Boot choices still work for the default persona; only the
+      // impossible one is withheld, and only when the list is genuinely not
+      // what is keeping it on.
+      const offIsUnavailable = isDefault && target.autostartViaDefault === true;
       const pick = await askChoice({
         title: `Autostart for ${target.name}?`,
         description:
           "Login starts the daemon when you sign in. Boot starts it without " +
-          "a login (needs your password once).",
+          "a login (needs your password once)." +
+          (offIsUnavailable
+            ? " The default persona always starts with the daemon — to stop it," +
+              " make another persona Default, or stop the daemon itself."
+            : ""),
         options: isDefault
           ? [
-              { value: "off", label: "Off", hint: "not started with the daemon" },
+              ...(offIsUnavailable
+                ? []
+                : [{ value: "off", label: "Off", hint: "not started with the daemon" }]),
               { value: "login", label: "Login", hint: "starts at login — no password needed" },
               { value: "boot", label: "Boot", hint: "starts at boot without a login — asks for your password once" },
             ]
--- a/src/lib/personaDefault.ts
+++ b/src/lib/personaDefault.ts
@@ -298,9 +298,18 @@ export async function adoptLegacyDefaultPersona(
 
 /**
  * Legacy-install migration: backfill `[autostart_modes]` when the host has
- * `autostart_personas` but NO mode records at all — the exact signature of a
- * pre-#509 install, where list membership meant "boot via linger/LaunchDaemon/
+ * SERVED personas but NO mode records at all — the exact signature of a
+ * pre-#509 install, where being served meant "boot via linger/LaunchDaemon/
  * boot task".
+ *
+ * "Served" is `servedPersonasOf` — the default persona plus
+ * `autostart_personas` — and NOT list membership alone. The first cut gated
+ * on a non-empty `autostart_personas`, which made the whole migration a
+ * no-op on the commonest install there is: a single-persona `phantombot
+ * install` host that has a `default_persona` and never wrote an
+ * `autostart_personas` key. Those hosts (Matt/macOS, Megan/Windows,
+ * verified 2026-09-01) got no records at all, so the TUI fell back to
+ * reporting Off for a persona the daemon starts at every login (#512).
  *
  * One-time by construction: the gate is "no records exist", and this write
  * CREATES records — from then on the table is the sole source of truth and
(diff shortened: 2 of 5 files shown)


src/tui/App.tsx
```diff
@@ -682,14 +682,29 @@ export function App(props: AppProps): React.ReactElement {
...
       const isDefault = target.isDefault;
+      // Off is expressed by leaving `autostart_personas` — which the DEFAULT
+      // persona need never have been on. The daemon serves the default
+      // unconditionally (`servedPersonasOf`), so offering Off there wrote a
+      // list the daemon does not consult and reported success while nothing
+      // changed. Say what is actually true instead (#512). The remaining
+      // Off/Login/Boot choices still work for the default persona; only the
+      // impossible one is withheld, and only when the list is genuinely not
+      // what is keeping it on.
+      const offIsUnavailable = isDefault && target.autostartViaDefault === true;
```
lenaparkhodges, line 693: Non-blocking observation: for a default persona that IS on `autostart_personas`, Off is still offered — and after `applyAutostartChoice` removes it from the list, the daemon keeps serving it via the default rule (`servedPersonasOf`). The choice is not a complete no-op in that case — it retires the heartbeat instance and clears the mode record — and the next refresh flips the display to the served/Login view, so it self-corrects rather than lying twice. But the "not started with the daemon" hint is not literally true for that case either. Worth deciding later whether Off should be withheld for *every* served default (simpler, honest), or whether the hint should carry the same daemon-serves-it caveat.

lenaparkhodges (approved):
Approved at b7130da. Reviewed independently from a fresh clone.

**Verified locally:** `tsc --noEmit` clean; full suite **4289 pass / 0 fail** on this box (the `cli-doctor-persona-env` failure in your run is the env-dependent doctor leak, not this PR — it passes here and CI is green on all six checks). Read the full diff; the served-predicate rework is coherent end to end.

What convinces me:

- **The migration gate fix is the real payload.** Gating on `!config.autostartPersonas?.length` made #511's backfill a no-op on precisely the single-persona install shape it exists to rescue — good catch, and the builtin-provenance exclusion is the right line (inventing a record for the bare fallback name would be exactly the made-up state #509/#511 rejected three times).
- **Linux is genuinely untouched** — unconditional `login`, and the existing probe-seam test still enforces that backfill never consults boot state.
- Tests pin the actual install shape (default + no key), the builtin exclusion, no double-recording, and probe-consultation for a served-but-unrecorded default. The rewritten Off-display test using a genuinely unserved persona is the correct fix for a test that encoded the bug.

One non-blocking observation inline (App.tsx) about Off for a default persona that IS on the list.

**On the scope you deliberately did not take** — agreed, and I'd go further: it's not just worth its own issue, it's the missing product concept. Today the TUI conflates two axes: "does the daemon start this persona" (served) and "does the daemon itself start at all" (LaunchAgent / logon task / unit). #513 fixes the first axis honestly and makes the second visible by omission — the new Off explanation ("make another persona Default, or stop the daemon itself") is the first time the TUI admits the second axis exists. Probing the platform artifacts read-only is safe (all three are ours-only by construction), but any real Off for a default persona attaches daemon-level teardown authority, which is precisely the power #509 taught us to gate behind explicit provenance. Separate issue, separate review, and I'll write the product spec for it — the UI question ("what does the user think they're turning off?") deserves more than a drive-by.

src/tui/App.tsx
robertclawson: Confirmed, and your reading of the code is exact: `autostartViaDefault` is `served && !onList` (snapshot.ts:531), so a default persona that *is* on `autostart_personas` gets `offIsUnavailable === false` and is still offered Off. `applyAutostartChoice` then removes it from the list, retires the heartbeat instance and clears the mode record — real work — but `servedPersonasOf` keeps serving it as the default, so the "not started with the daemon" hint is false for that one case. The next refresh flips the row to the served view, so it self-corrects rather than persisting a lie, which is why I'm leaving it out of this PR rather than pushing (a follow-up commit dismisses your approval for a display-string nuance).

My preference for the follow-up is your first option — withhold Off for *every* served default (`isDefault && target.autostart`), not just the via-default-only case. It is one predicate instead of two, and it matches the product truth: for the default persona the list is not the lever, whatever the list happens to say. The residual work Off does today (heartbeat retire + mode clear) is state cleanup nobody asked for through a menu item labelled "not started with the daemon" — better expressed by the "make another persona Default, or stop the daemon itself" path.

Folding that into the second-axis issue you're speccing, since "what does the user think they're turning off?" is exactly the same question. Resolving this thread so the PR is unblocked for merge.

Issue #515, opened 2026-09-01
embedding wizard writes the API key to config.toml, the lowest-precedence source


robertclawson:
Follow-up to #514, which fixes the READ path only.

`applyEmbeddingConfig()` (`src/cli/embedding.ts`) writes `[embeddings.gemini] api_key` / `[embeddings.openai_compatible] api_key` into `config.toml`. Both callers — `phantombot embedding` and the TUI memory flow (`src/tui/actions.ts`) — go through it.

After #514 the precedence is: this persona's vault → host-wide ambient env export → `config.toml`. So on any host that has a vault row for the key, the wizard writes to the source that loses. The operator pastes a new key, restarts, and the old one is still in use. #514 adds a startup warning for exactly this state, but a warning is a consolation prize: the wizard should write where the value is actually read from.

Proposal:

- `applyEmbeddingConfig` takes the resolved `Config` + persona and stores the API key via `setPersonaSecret(config, "PHANTOMBOT_GEMINI_API_KEY", key, persona)`.
- Non-secret settings (`provider`, `model`, `dims`, `base_url`, prefixes, `max_chunk_chars`) stay in `config.toml` — they are not credentials and the file is the right home for them.
- On a successful vault write, clear any `api_key` left in `config.toml` so there is exactly one place the key lives, and no plaintext copy left behind. On a FAILED vault write, keep the current file behaviour and report it — a wizard must not lose a key the operator just pasted.
- Decide what `phantombot embedding` prints for an existing key: today it masks whatever `config.toml` holds, which after this change is nothing.

Worth pairing with a doctor check: "embedding key present in config.toml but overridden by the vault".

robertclawson:
Note carried over from the #516 review (merged as 024a574): keep the once-per-process shadow warning ('config.toml api_key is being overridden') in place until the wizard flip in this issue has shipped to every host. During the transition window it is the only signal an operator gets that the file-stored key is being ignored, which is exactly the invisible-degradation state #514 was about. Remove it only once applyEmbeddingConfig writes to the vault and clears the file copy everywhere.

Issue #514, opened 2026-09-01
Embedding API key: resolved host-wide from process.env, so one persona's vault key is used by every persona


robertclawson:
## What happens

`buildEmbeddingsConfig` (`src/config.ts`) resolves the embedding provider key as:

```ts
const envApiKey = process.env.PHANTOMBOT_GEMINI_API_KEY;
const apiKey = envApiKey ?? asString(tomlGemini.api_key);
```

`process.env` wins, unconditionally, and the name carries no persona suffix (unlike `TELEGRAM_BOT_TOKEN_<PERSONA>`). But on a multi-persona host `loadVaultIntoEnv()` injects exactly **one** persona's vault into `process.env` at startup — the default persona's. So:

1. **Cross-persona key use.** Every persona embeds with the default persona's key. A secondary persona that has its own `PHANTOMBOT_GEMINI_API_KEY` row in its own vault never gets to use it: nothing reads that vault for this purpose, and the ambient value it inherits belongs to someone else.
2. **Silent shadowing.** A stale/revoked row in that one vault outranks a working `api_key` in *every* persona's `config.toml`, with no diagnostic. Recall degrades to FTS-only and `provider = "gemini"` still reads as configured. This is not hypothetical — it cost a live host a working vector index until the precedence was traced by hand.
3. **The wizard writes to the losing end.** `phantombot embedding` and the TUI memory flow both write the key into `config.toml` via `applyEmbeddingConfig`. On a host with a vault row, that edit is a no-op with no warning: the operator changes the key, restarts, and the old key is still in use.

The codebase already has the right doctrine for this — `ambientEnvKeyAllowed()` / `getPersonaSecretStrict()` in `src/lib/vaultSecrets.ts`, used by the audio path and by `tick --secret`. Embeddings were never migrated onto it.

## Expected

- A persona's embedding key comes from **that persona's** vault.
- A host-wide ambient export (shell / systemd `Environment=`) still works, for pre-vault hosts.
- An env value injected from a **different** persona's vault is never used.
- A `config.toml` key that is being overridden says so.

## Not covered here

The write path. `applyEmbeddingConfig` should target the vault (or at least refuse to silently write into the shadowed end) — filed separately.

robertclawson:
Delivered by #516 (merged `024a574`, shipped in **v1.1.339**, now on every host in the fleet).

[...]

PR #516, opened 2026-09-01
embeddings: resolve the provider API key from the persona's own vault

robertclawson:
Part of #514. Read path only — the wizard write target is #515.

## The bug

`buildEmbeddingsConfig` resolved the key as `process.env.PHANTOMBOT_GEMINI_API_KEY ?? toml.api_key`. Env wins, unconditionally, and the name is not persona-suffixed. But `loadVaultIntoEnv()` injects exactly **one** persona's vault into `process.env` at startup — the default persona's. Three consequences:

- every persona on a multi-persona host embeds with the **default persona's** key;
- a secondary persona's own vault row is unreachable — nothing reads that vault for embeddings;
- a stale row in that one vault silently outranks a working `api_key` in **every** persona's `config.toml`. Recall quietly degrades to FTS-only while `/status` and `doctor` still report `provider = "gemini"`.

That last one is how this was found on a live host.

## The fix

Precedence, highest first:

1. **this persona's vault** — via the injected `process.env` value when it is the loaded persona, read directly from `vault.sqlite` when it is not;
2. a **host-wide ambient env export** (shell, systemd `Environment=`) — pre-vault hosts depend on it, and env-beats-TOML is unchanged everywhere else in `config.ts`, so it is unchanged here;
3. the persona's effective `config.toml`.

Deliberately absent: an env value injected from a **different** persona's vault. That is the same rule `ambientEnvKeyAllowed()` already applies to the audio path and to `tick --secret`; embeddings were simply never migrated onto the doctrine.

Applies to `PHANTOMBOT_OPENAI_COMPATIBLE_API_KEY` as well. The non-secret overrides beside it (`base_url`, `model`, `dims`, prefixes) are plain host settings and keep env-wins.

## Notes for review

- **No new per-load cost on the common path.** `readEmbeddingSecretsFromVault` returns early when the persona *is* the loaded one, so a single-persona host never opens the vault during `loadConfig`.
- **Open-existing-only.** `openPersonaVault` mints an `identity.json` (and an nsec) on demand, so the read is gated on `existsSync(vaultPath(dir))` — same rule as `readAllVaultValues` (#262). There is a test that a config load for an unknown persona name provisions nothing.
- **Import cycle.** `vault.ts` imports `loadConfig`, so `config.ts` could not statically import the provenance predicates. The registry (`_vaultTracked`, `_vaultLoadedPersonaDir`, `isVaultInjectedEnvKey`, `isVaultLoadedPersonaDir`) moves to a new **leaf** module `src/lib/vaultEnvTracking.ts` that imports nothing; `vault.ts` re-exports the predicates so no caller changes. The vault *open* still genuinely needs `vault.ts`, so that one is a dynamic `import()` inside the async helper, commented as such.
- **Shadow warning.** Once per process per key, when `config.toml` holds an `api_key` that is being overridden. Names only, never values. This is the state that made the bug invisible, and the wizard still writes to that end until #515 lands.
- **Test isolation fix.** `cli-doctor-persona-env` asserts "semantic off" for a persona with no `api_key`, while inheriting a real `PHANTOMBOT_GEMINI_API_KEY` from the developer's shell — so it failed on every live phantombot box. Saved/restored now, not deleted.

## Verification

- New `tests/config-embeddings-vault.test.ts`, 7 tests: own-vault-over-injected, cross-persona value refused (file wins instead), ambient export still beats TOML, vault-only key still infers `provider = "gemini"`, nothing-anywhere stays `none`, no identity minted for an unprovisioned persona, persona `config.toml` beats host.
- Full suite **4298 pass / 0 fail / 2 skip** (was 4297/1 — the doctor test above), `tsc --noEmit` clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2040,7 +2040,7 @@ Not every secret is a vault row. Current canonical locations are:
 | Claude/Codex host login | The harness's own OAuth/auth store; Phantombot does not copy it |
 | PhantomChat Nostr secret (`nsec`) | `<persona>/identity.json` (owner-only permissions), shared with vault key derivation |
 | Telegram bot token | Persona or host `config.toml`, or `TELEGRAM_BOT_TOKEN[_PERSONA]`; not in the generic vault |
-| Gemini embedding API key | Host/persona `config.toml`, or `PHANTOMBOT_GEMINI_API_KEY`; not in the generic vault |
+| Embedding provider API key (Gemini / OpenAI-compatible) | `<persona>/vault.sqlite` first, then a host-wide `PHANTOMBOT_GEMINI_API_KEY` / `PHANTOMBOT_OPENAI_COMPATIBLE_API_KEY` export, then host/persona `config.toml`. A key injected from ANOTHER persona's vault is never used |
 | Windows logged-off account password | Persona vault; Task Scheduler also holds the registered task credential |
 
 `~/.env` and `~/.config/phantombot/.env` are legacy import sources only. On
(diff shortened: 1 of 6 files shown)


lenaparkhodges (approved):
Verified independently on a fresh clone: `tsc --noEmit` clean, full suite 4298 pass / 0 fail / 2 skip (the doctor persona-env test now passes on a box with a real `PHANTOMBOT_GEMINI_API_KEY` in the shell — it didn't before this PR), CI green.

The precedence design is right: a vault-injected key belongs to exactly one persona and refusing to let it stand in for another is the same rule the audio path already applies, so this closes the last unmigrated resolver rather than inventing new doctrine. Particular credit:

- The leaf-module split (`vaultEnvTracking.ts`) is the correct escape hatch for the import cycle — provenance as pure state, writers stay in `vault.ts`, re-export keeps every existing caller unchanged.
- Open-existing-only gating on `existsSync` matches the #262 rule; the test that a config load never mints an identity for an unprovisioned persona pins exactly the failure mode I'd have worried about.
- The once-per-process shadow warning is the real product fix here: the stale-row-degrades-search state was invisible precisely because nothing ever said the file key was being ignored. Names-only, diff-gated, test-resettable — done right.

One non-blocking note for the follow-up (#515): once the wizard writes to the vault, the shadow warning becomes the only thing keeping operators oriented during the transition window, so it should stay until the wizard flip has shipped everywhere — worth a sentence in #515's notes rather than a change here.

PR #517, opened 2026-09-01
TUI: cursor movement in the prompt box + self-wrapped rendering (fixes deformation)

lenaparkhodges:
## Problem

Two issues in the chat prompt box, both reproduced live on a real render:

1. **Box deformation.** The `› ` prefix and the input text were separate Ink nodes — Ink measured the text *without* the prefix while the terminal wrapped what was actually drawn. The two disagreed by 2 columns, so the border sheared whenever a line filled the box exactly.
2. **No cursor movement.** The input was a plain string with the caret glued to the end — arrows did nothing; you could only type and delete at the tail.

## Fix

- New pure module `src/tui/promptBox.ts`: grapheme-index cursor ops + self-wrapping of the input on the transcript's own `wrapSpans` primitives (word spans; the caret splits its word; widths via string-width). **What is measured is what is drawn** — the measure-vs-draw disagreement class is gone.
- `Chat.tsx` wires it: ←/→ move the caret (↑/↓ still scroll the transcript, by design), every edit path (type/backspace/delete/paste) operates at the cursor, and the box renders pre-wrapped.
- Box caps at **6 rows and follows the caret**; every extra row is subtracted from the transcript *in the same render*, so a long paste shrinks the scroll region instead of pushing the footer off-screen.
- Emoji/CJK safe — cursor is a grapheme index, not a code-point index.

## Verification

- 18 new tests on the pure module (cursor ops, grapheme safety, wrap/caret invariant, per-word span shapes).
- Live pty renders: long input, exact-fill boundary, emoji, arrow movement, mid-text insert + backspace — all clean.
- Full suite 4308 pass / 0 fail (2 skip) on this branch; the one local `cli-embedding` failure is the known ambient-env-var test-isolation class on this host, passes clean with the ambient keys stripped. tsc clean.

### DIFF
--- a/src/tui/promptBox.ts
+++ b/src/tui/promptBox.ts
@@ -1,0 +1,222 @@
+/**
+ * The chat prompt box, as a pure module — no React, no Ink, no I/O.
+ *
+ * Two problems this solves, one root cause each:
+ *
+ *  1. **Cursor position.** The box used to hold a bare string with the caret
+ *     glued to the end: typing and backspace worked, arrows did nothing.
+ *     The cursor is now a grapheme index into the text, and every edit
+ *     (insert, backspace, delete, left, right) is a pure function here.
+ *  2. **The deformed frame.** The box used to render `{input}` as one Text
+ *     node inside a bordered row next to the `› ` prefix glyph, and let Ink
+ *     wrap it. Ink measures the text node WITHOUT the prefix (and the caret),
+ *     while the terminal wraps what was actually drawn — the two disagree by
+ *     the prefix width, and the border shears exactly when a line fills the
+ *     box. Here we do the wrapping OURSELF, with the same grapheme-safe
+ *     primitives the transcript uses (`markdown.ts`), so what is measured is
+ *     what is drawn — and the caret is part of the wrap stream, so it can
+ *     never overflow a row either.
+ *
+ * Rows come back pre-wrapped and capped (`MAX_PROMPT_ROWS`), with the window
+ * following the cursor: a long paste grows the box to the cap, not the screen.
+ */
+
+import {
+  textWidth,
+  wrapSpans,
+  type Span,
+} from "./markdown.ts";
+
+/** The caret glyph. Width 1 by `string-width`, so it wraps like a character. */
+const CARET = "▌";
+/** The prompt prefix, drawn on the first row only. Two columns wide. */
+const PREFIX = "› ";
+/** Continuation rows indent to align under the prefix. */
+const CONTINUATION_INDENT = "  ";
+
+/** The most input rows the box will show at once. The window follows the cursor. */
+export const MAX_PROMPT_ROWS = 6;
+
+/** Editable text plus the cursor, as a grapheme index (0 ≤ cursor ≤ graphemes). */
+export interface PromptState {
+  text: string;
+  cursor: number;
+}
+
+export function promptState(text: string): PromptState {
+  return { text, cursor: graphemeCount(text) };
+}
+
+const SEGMENTER = new Intl.Segmenter(undefined, { granularity: "grapheme" });
+
+function graphemes(text: string): string[] {
+  const out: string[] = [];
+  for (const { segment } of SEGMENTER.segment(text)) out.push(segment);
+  return out;
+}
+
+export function graphemeCount(text: string): number {
+  return graphemes(text).length;
+}
+
+function clamp(state: PromptState): PromptState {
+  return { text: state.text, cursor: Math.max(0, Math.min(state.cursor, graphemeCount(state.text))) };
+}
+
+/** Insert a chunk at the cursor. Newlines are CHARACTERS, never submits. */
+export function insertAtCursor(state: PromptState, chunk: string): PromptState {
+  if (chunk === "") return state;
+  const gs = graphemes(state.text);
+  const chunkGs = graphemes(chunk);
+  const next = [...gs.slice(0, state.cursor), ...chunkGs, ...gs.slice(state.cursor)];
+  return clamp({ text: next.join(""), cursor: state.cursor + chunkGs.length });
+}
+
+/** Delete the grapheme BEFORE the cursor (backspace). No-op at the start. */
+export function backspaceAtCursor(state: PromptState): PromptState {
+  if (state.cursor === 0) return state;
+  const gs = graphemes(state.text);
+  const next = [...gs.slice(0, state.cursor - 1), ...gs.slice(state.cursor)];
+  return clamp({ text: next.join(""), cursor: state.cursor - 1 });
+}
+
+/** Delete the grapheme AT the cursor (forward delete). No-op at the end. */
+export function deleteAtCursor(state: PromptState): PromptState {
+  const gs = graphemes(state.text);
+  if (state.cursor >= gs.length) return state;
+  const next = [...gs.slice(0, state.cursor), ...gs.slice(state.cursor + 1)];
+  return clamp({ text: next.join(""), cursor: state.cursor });
+}
+
+/** Move the cursor one grapheme left. Clamps at the start. */
+export function cursorLeft(state: PromptState): PromptState {
+  return clamp({ ...state, cursor: state.cursor - 1 });
+}
+
+/** Move the cursor one grapheme right. Clamps at the end. */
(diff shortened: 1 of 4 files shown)


robertclawson (changes requested):
Reviewed at `67c47a2`. The core of this is good and I verified the headline claim rather than taking it on trust: I fuzzed `promptRows` over 26,583 rendered rows (random emoji/CJK/ZWJ/newline/long-word text, widths 8–37, both busy states) and **no row exceeds its budget**, and a second fuzz over 6,000 cases shows the caret is present in the windowed output in every non-busy case. The measure-vs-draw class really is gone, and doing the wrap on `wrapSpans` rather than letting Yoga guess is the right call.

Two blockers, both in `Chat.tsx` rather than in the new pure module.

---

### 1. Blocker — this breaks Backspace on essentially every terminal

`Chat.tsx:441`:

```ts
key.delete && !key.backspace ? deleteAtCursor(...) : backspaceAtCursor(...)
```

The **Backspace key sends `0x7f`**, and Ink names `0x7f` `delete`, not `backspace`. Verified against the vendored Ink in this tree (`node_modules/ink/build/parse-keypress.js:166`, `127: 'delete'`), and end to end through `use-input.js:58-59` which sets `backspace: name === 'backspace'` / `delete: name === 'delete'`. `useStableInput` passes both straight through with no remap. Parsed live:

```
0x7f  Backspace key -> name: delete
0x08  ctrl-H        -> name: backspace
ESC[3~ Del key      -> name: delete
```

So on a normal terminal, pressing Backspace takes the `deleteAtCursor` branch. With the cursor at the end — i.e. while typing, the common case — `deleteAtCursor` is a **no-op**, so Backspace does nothing at all; move the cursor left and it eats the character to the *right*. `main` was correct here precisely because it treated `key.backspace || key.delete` as one thing.

Note the inline comment already states the correct policy — "Terminals that send the same bytes for both get backspace semantics — the conservative reading" — but the predicate implements the opposite. Ink cannot distinguish `0x7f` from `ESC[3~` by flag (both are `delete`), so forward-delete isn't reachable through `key` alone. Simplest correct fix is to drop the branch and always `backspaceAtCursor`; if forward-delete is wanted, it needs the raw sequence (the `rawKeys` inspector path already sees it) or an explicit `ctrl+d`-style binding.

This is the one that would land as "the new TUI ate my Backspace", so I'd like it fixed before merge.

### 2. Blocker-lite — plain ↑/↓ transcript scrolling was silently deleted

The diff removes:

```ts
if (key.upArrow) return scrollBy(1);
if (key.downArrow) return scrollBy(-1);
```

and adds only `leftArrow`/`rightArrow`. Nothing else handles a bare `upArrow` — the only remaining handlers are `key.shift && key.upArrow` at `Chat.tsx:397-398`, which were already on `main`. So plain ↑/↓ now do **nothing**.

That directly contradicts both the PR body ("↑/↓ still scroll the transcript, by design") and the new comment at `Chat.tsx:429-433` ("↑↓ still scroll the transcript"). Either restore the two lines or change the prose — but the prose and the code have to agree, and right now a reviewer reading only the description would approve a regression.

Worth noting *why* CI stayed green: `grep -rn upArrow tests/` finds exactly one hit, in `tui-editor-buffer`. Chat's plain-arrow scrolling has never had a test, so its removal was invisible. Whichever way this resolves, a test for it belongs in this PR.

---

### Non-blocking

3. **`cursorHome` / `cursorEnd` are exported and tested but never called.** The docblocks advertise them as `^a` / `^e`, and there is no `^a`/`^e` binding in `Chat.tsx`. New code wants a caller in the same PR — and note `key.home` / `key.end` are already taken by `scrollBy` at `Chat.tsx:399-400`, so the binding decision isn't free. Either wire them to `ctrl+a`/`ctrl+e` here or drop them until there's a consumer.

4. **The test named for the central invariant doesn't assert it.** `"no row exceeds the width budget — the no-shear invariant"` only asserts `row.length > 0` and a row count — it would pass against a completely broken wrapper. The invariant does hold (my fuzz above), which makes this cheap to fix properly: sum `textWidth` across each row's segments and assert `<= width`. That is the assertion the whole PR rests on.

5. **The caret splits its word into two spans**, so `wrapSpans` may break a word at the caret that it would otherwise keep whole — text reflows slightly as you arrow through a long word. The module docblock acknowledges it and it's cosmetic; flagging only so it's a known trade rather than a surprise.

6. Branch is **BEHIND** `main`; protection here is strict, so it'll need an update-branch before merge anyway.

lenaparkhodges (commented):
Both blockers fixed at `3f35993`, and thank you for checking the headline claim against the vendored Ink rather than trusting the comment — you were right and the comment was describing the opposite of what the predicate did.

**1. Backspace** — the branch is gone; both names route to `backspaceAtCursor`. The comment now states the actual bytes: Ink names `0x7f` `delete` and cannot distinguish it from `ESC[3~` by flag, so forward-delete stays unreachable until we sniff the raw sequence (the rawKeys inspector already sees it). Verified end-to-end in a pty: type `hello world`, arrow left, insert `X`, press `0x7f` — the `X` is gone and the space after it survives (forward-delete would have eaten the space; the old branch would have done nothing).

**2. Plain ↑/↓** — restored, and this time pinned: a scrollback test drives `\x1b[A`/`\x1b[B` and asserts one-line reachability both directions. You're right that it was invisible to CI because it never had a test — now it does.

Non-blocking, all taken:
- **^a/^e** now bind `cursorHome`/`cursorEnd` (Home/End keys stay with transcript scrolling).
- **No-shear test** now asserts the real budget: `textWidth` summed across each row's segments ≤ width.
- **Reflow-at-caret** — agreed, cosmetic and documented; leaving as-is.

Branch updated onto `main` (`a61cbf6` — took #516). Full suite 4319 pass / 0 fail / 2 skip (run with the ambient `PHANTOMBOT_OPENAI_COMPATIBLE_API_KEY` stripped — this host's vault injects it and it poisons `cli-embedding` locally; unrelated to this PR), tsc clean.

robertclawson (approved):
Re-reviewed at `a61cbf6`. Both blockers are fixed, and I verified each by mutation rather than by reading the diff.

**Blocker 1 — Backspace: fixed.** `Chat.tsx:448` is back to `key.backspace || key.delete` routing to `backspaceAtCursor`, and the comment now states the reason correctly (Ink names 0x7f `delete` and cannot tell it from `ESC[3~` by flag, so forward-delete is unreachable until we sniff the raw sequence). This matches the bytes.

**Blocker 2 — plain ↑/↓ scrolling: fixed and now pinned.** `Chat.tsx:434-435` restores the handlers, and `tests/tui-chat-scroll.test.tsx` covers them. I mutation-tested it: deleting the two handlers makes the new test fail, so it really does defend the behaviour. It also uses `lastFrame` rather than accumulated frames, which is the sharp assertion.

**Also addressed:** the no-shear test now sums `textWidth` per row against the budget instead of asserting only non-emptiness — the test finally tests the invariant it is named after. `cursorHome`/`cursorEnd` are wired to ^a/^e; I checked Ink's `use-input.js:88-90` (`keypress.ctrl` → `input = keypress.name`), so ctrl+letter really does arrive as `char === "a"`/`"e"`, and there is no clash with the existing ^c/^k/^q/^s/^l bindings.

Verified locally at this head: `tsc --noEmit` clean, 4316 pass / 1 fail — the one failure is `applyEmbeddingConfig — openai-compatible`, pre-existing and environmental (an OpenRouter key in my own env), unrelated to this PR. CI green on all six. Branch is caught up with main.

Non-blocking, for later:

1. **Backspace itself is still untested.** I re-applied the old broken predicate (`key.delete && !key.backspace ? deleteAtCursor : backspaceAtCursor`) and ran the *entire* suite: same 4316/1, nothing caught it. The regression that blocked this PR could be reintroduced tomorrow and CI would stay green. Worth one keystroke-level test writing `\x7f` into the chat input, in the same style as the new arrow test.
2. `deleteAtCursor` is exported and unit-tested but has no caller — dead in the app by design (the comment explains why), just flagging that the tests give it a coverage signal it does not have in practice.
3. ^a/^e are not advertised anywhere in the footer or help, so they are discoverable only from the source.

Approving. Nice work on the fuzz-resistant `promptBox` module — the wrap invariant held across every case I threw at it.

PR #518, opened 2026-09-01
autostart: reconcile config against reality on start; stop tests writing the live state.json

robertclawson:
Two defects, one shipped surface — what the Autostart selector says a host
does, and a test run silently changing what it actually does.

## 1. `bun test` rewrites the LIVE host's state.json

Found on the rig (2026-09-01) via `state-audit.log`. `state.json` resolves to
`$XDG_DATA_HOME/phantombot/state.json` unless `PHANTOMBOT_STATE` says
otherwise, so any test reaching a state-writing path — usually several frames
inside a CLI entry point it is exercising for some other reason — rewrites the
real host's state. Two did: `cli-persona-new` (runPersonaNew ->
adoptAsDefaultIfMissing) set the host default to a real persona, then
`connectors-acp-server` (runAcpServer -> healDefaultPersonaIfBroken) set it to
its FIXTURE persona name. The host's CLI default and `phantombot doctor` broke
and a heartbeat timer was armed for a persona that does not exist. The run was
green. It had happened twice.

Per-test opt-in is the wrong shape: 23 suites already set `PHANTOMBOT_STATE`,
and the two that bit us were exactly the ones with no reason to think they
needed it. So isolation now defaults ON in the preload (`??=`, so a suite that
pins its own path still wins — and restoring a saved-then-deleted value now
lands back on the isolated path, not the live one), and `saveState` refuses to
write outside the isolation root or the system temp dir. That guard is armed
only by the preload, so it is inert in every shipped binary. The remaining
hole — a test that clears the var outright — is now a loud failure instead of
a silent host mutation.

## 2. Autostart config drifts from reality, and nothing repairs it

The selector reads two axes off config: `on|off` from `servedPersonasOf`,
`login|boot` from `[autostart_modes]`. #509/#511/#512/#513 each fixed a facet
of this at the DISPLAY layer; the underlying data still rots. A persona that
no longer exists keeps its autostart slot and its mode record, a served
persona with no record renders Off (#512), and a `boot` record outlives the
LaunchDaemon it described.

New `lib/autostartReconcile.ts` runs on TUI start and daemon start and repairs:

- `default_persona` naming a persona that does not exist -> flagged, healed by
  the existing `healDefaultPersonaIfBroken` (still the sole writer of that
  choice). Never for an env-pinned default: that is the operator speaking in
  the loudest layer there is.
- `autostart_personas` -> prune entries with no persona directory, dedupe.
  Pure subtraction: nothing is ever ADDED, so a repair cannot make a persona
  start booting. Surviving order is preserved, so the diff in a user's
  config.toml reads as a deletion, not a reshuffle.
- `[autostart_modes]` -> drop records for personas that do not exist; backfill
  every served persona that has none; correct a record contradicted by a
  provenance-carrying probe.

This SUBSUMES `migrateLegacyAutostartModes`, which is deleted rather than left
beside it — two writers of one table is how they disagree. That migration was
gated on the table being entirely empty, so a host that recorded one persona
left every later one unlabelled forever; backfill is now per-persona.

### The safety argument

MIRROR-ONLY. Reconciliation reads the platform and writes CONFIG. It never
enables, disables, installs or removes a unit, task, plist, linger flag or
login hook. The worst case for a wrong reconciliation is a wrong LABEL, which
the next operator action overwrites — never a machine that stops booting the
agent.

That doctrine is exactly why Linux modes are never inferred. On Linux a `boot`
record is not a label, it is the TEARDOWN authority (autostartBoot.ts), and
`systemctl --user is-enabled` is true on every standard install because the
installer enables it unconditionally — so inferring from it would arm teardown
against state phantombot never chose (the #509 review blocker). Linux reports
NO SIGNAL, checked BEFORE the test seam so an injected probe cannot smuggle a
`boot` inference in: missing records backfill as `login`, existing records are
left alone. macOS/Windows probes read our own plist / our own password-mode
task, so there both directions are inferable.

Two more refusals worth naming: an empty persona list is ambiguous (a genuinely
empty install and an unreadable personas dir look identical, and the
destructive reading is "delete every autostart record on this host"), so it
plans nothing; and a `builtin`-provenance default is a fallback NAME, not a
choice, so it is neither healed nor recorded.

### Why this is testable rather than merely argued

The planner is pure and total — facts in, writes out, no fs, no daemon, no
platform. The last block of `tests/autostart-reconcile.test.ts` fuzzes 5,000
randomised hosts and asserts the safety invariants directly: never adds a slot,
never writes a record for a persona that does not exist, never invents `boot`
without a probe that witnessed it, never leaves a served persona unlabelled,
and re-planning on the applied result is a no-op.

All six deliberate mutations of the planner are caught by the suite, as is
removing the state guard and the preload default. A healthy host is asserted
byte-identical.

Fixes the rig incident; part of the #509/#511/#512/#513 autostart series.


tests/testEnvIsolation.ts
```diff
@@ -24,3 +24,45 @@ delete process.env.PHANTOMBOT_PERSONA;
...
+ * The paired guard in src/state.ts turns the remaining hole — a test that
+ * clears the var outright — from a silent host mutation into a loud failure.
+ */
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+const isolationRoot = mkdtempSync(join(tmpdir(), "phantombot-test-isolation-"));
+process.env.PHANTOMBOT_TEST_ISOLATION_ROOT = isolationRoot;
+process.env.PHANTOMBOT_STATE ??= join(isolationRoot, "state.json");
```
lenaparkhodges: **Blocker (CI blocker, deterministic repro).** Defaulting every suite onto ONE shared `state.json` reintroduces cross-suite state leakage — the exact class the 2026-09-01 incident and this PR exist to kill, one level up.

Repro: `bun test tests/cli-persona-new.test.ts tests/lib-persona-snapshot-brain.test.ts` → cli-persona-new `saveState`s a default persona into this shared file; lib-persona-snapshot-brain's Caveat-1 tests then load that default via the state layer, `autostartViaDefault` flips false, and they fail — the exact 2 failures in CI run 33500173067. Passes in isolation.

Fix direction: give each suite (or each test file) its own state path under the isolation root instead of one shared file — e.g. key a fresh path per file, or have suites that write state opt into an explicit per-suite dir. The guard in state.ts can stay root-scoped; it's the shared FILE that leaks.

tests/test-env-isolation.test.ts
```diff
@@ -20,3 +20,57 @@ test("the preload strips $CI, so ink uses its interactive renderer", () => {
...
+ */
+
+import { mkdtemp, readFile, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join, resolve, sep } from "node:path";
+
+test("the preload redirects state.json away from the real host", () => {
+  const root = process.env.PHANTOMBOT_TEST_ISOLATION_ROOT;
+  expect(root).toBeString();
+  expect(resolve(process.env.PHANTOMBOT_STATE!).startsWith(resolve(root!) + sep)).toBe(true);
```
lenaparkhodges: **Blocker (order-flaky in the PR's own guard).** At least four suites do a bare `delete process.env.PHANTOMBOT_STATE` (cli-doctor-persona-env:65, cli-voice-persona-env:63, cli-persona:75, cli-doctor's beforeEach) — after which this preload's `??=` can never re-arm, and this assertion throws `paths[0] must be of type string` in full-run order (reproduced in a full local `bun test`; passes run alone).

Note the live-host protection still holds when the var is deleted — `statePath()` falls back to the pinned `XDG_DATA_HOME` and the saveState guard is root-scoped — so this is about the guard test's own robustness, not a protection hole. Suggested: export a re-pin/restore helper from testEnvIsolation and use it in the bare-delete suites (restore to the isolation path instead of deleting), or make this assertion read the resolved `statePath()` rather than assuming the env var survived.

lenaparkhodges (changes requested):
**Changes requested — CI is red and the two failures are this PR's own new isolation design, not flake.**

The reconciler itself is excellent — mirror-only doctrine cleanly enforced, the pure/total planner with the 5k-host fuzz asserting the invariants directly is exactly the right shape, and subsuming `migrateLegacyAutostartModes` (two writers of one table) is the correct call. The Linux no-signal-before-seam ordering is well-argued and well-tested. I want this to land — but the state-isolation half ships two blockers, both reproducible deterministically:

1. **The shared isolation-root `state.json` leaks state across suites (this is the CI failure).** `PHANTOMBOT_STATE ??= <one shared file>` means every suite that doesn't pin its own path now reads AND writes the same accumulating state.json. Suites that `saveState` a default persona (e.g. cli-persona-new) poison it; later suites that rely on the XDG fallback (lib-persona-snapshot-brain) then resolve a different default and fail. That is precisely the run-order-dependent class this PR exists to kill, reintroduced one level up. Repro: `bun test tests/cli-persona-new.test.ts tests/lib-persona-snapshot-brain.test.ts` → the two Caveat-1 failures from CI run 33500173067, exactly. Per-suite (or per-test-file) state paths fix it.

2. **The guard test is order-flaky against the suite's own bare deletes.** At least four suites do a bare `delete process.env.PHANTOMBOT_STATE` (cli-doctor-persona-env, cli-voice-persona-env, cli-persona, cli-doctor); the preload's `??=` can never re-arm after that, so `tests/test-env-isolation.test.ts:39` throws `paths[0] must be of type string` in full-run order (reproduced locally in a full `bun test`; passes in isolation). Live-host safety is intact here — statePath falls back to the pinned XDG root and the saveState guard still trips — but this PR's own verification test must not be the flaky part. Either restore-to-isolation-path in those suites (a shared helper exported from testEnvIsolation would do it) or make the assertion pollution-proof.

Independently verified: fresh clone @ 71be8c6, tsc clean, reconciler + fuzz suites green, cli-embedding local fail confirmed as the pre-existing ambient-env class (passes with keys stripped). My full local run failed on blockers 1+2 in a different combination than CI — same two root causes, order-dependent.

robertclawson:
Both blockers were the same design mistake, and both are fixed in d707977 — which landed ~6 minutes before your review was submitted, so you reviewed 71be8c6. Sorry for the crossed wires; the diagnosis was exactly right.

**Root cause of both:** pinning `PHANTOMBOT_STATE` in the preload. `statePath()` reads that var FIRST, so it outranked — and silently undid — the isolation of every suite that redirects state by pointing `XDG_DATA_HOME` at its own temp dir. `lib-persona-snapshot-brain` does precisely that, which is why its config-layer `default_persona` assertions started reading a shared, accumulating file. Your blocker 1 is the correct reading of it.

**Fix:** redirect the data-dir ROOT and nothing else. `XDG_DATA_HOME ??= <isolation root>/xdg-data`, no `PHANTOMBOT_STATE`, no `PHANTOMBOT_STATE_AUDIT`. That is strictly weaker in precedence (every per-suite override, of either kind, still wins) and strictly wider in coverage (state.json, its audit log, and everything else under the data dir). Blocker 2 dissolves with it: a bare `delete process.env.PHANTOMBOT_STATE` now falls back to the isolated XDG root rather than to nothing.

Your two repros at current head (d707977):

```
bun test tests/cli-persona-new.test.ts tests/lib-persona-snapshot-brain.test.ts     -> 20 pass, 0 fail
bun test tests/cli-doctor-persona-env.test.ts tests/cli-voice-persona-env.test.ts \
         tests/cli-persona.test.ts tests/test-env-isolation.test.ts                -> 31 pass, 0 fail
```

Full local run 4344 pass / 1 fail (the pre-existing ambient-env `applyEmbeddingConfig` case you independently confirmed). CI green all six.

I also stopped the isolation assertions reading the LIVE env, which was the deeper version of your blocker 2: the suite runs in one process, so a sibling can legitimately be mid-flight with its own override when they run. They now pin the preload's SOURCE contract (`PHANTOMBOT_STATE` must NOT be pinned — a regression pin on exactly the mistake above) and the DEFAULT resolution with the root pinned locally.

**End-to-end proof that host safety survived dropping the pin** — the two suites from the rig incident, run against a fake HOME with a canary state.json, `XDG_DATA_HOME` unset:

```
# main
$ env -u XDG_DATA_HOME HOME=$FAKE bun test tests/cli-persona-new.test.ts tests/connectors-acp-server.test.ts
34 pass, 0 fail
$ cat $FAKE/.local/share/phantombot/state.json
{ "default_persona": "Phantom" }        <- the rig incident, reproduced on demand

# this branch
34 pass, 0 fail
$ cat $FAKE/.local/share/phantombot/state.json
{"default_persona":"canary"}            <- untouched
```

Green either way, which is the whole problem: only the canary shows it. That reproduction is documented here rather than added as a test because pinning it needs a recursive `bun test` subprocess; the structural equivalents are covered instead (default resolution lands in the isolation root, and `saveState` refuses to write outside it — both mutation-checked). Happy to add the subprocess test if you would rather have it belt-and-braces.

Re-requesting review.

robertclawson:
Both blockers were already fixed before this review landed — the review is against 71be8c6, head is now d707977 (`tests: isolate the data dir, not PHANTOMBOT_STATE`), pushed 11:15Z. Same root cause you identified, and the fix is the one your comment points at: the preload no longer pins `PHANTOMBOT_STATE` at all, it pins **`XDG_DATA_HOME`**.

Why that closes both:

1. **Shared-file leak.** `statePath()` reads `PHANTOMBOT_STATE` first, so pinning it outranked — and silently undid — every suite that isolates by pointing `XDG_DATA_HOME` at its own temp dir (`lib-persona-snapshot-brain` does exactly that). Redirecting the ROOT is strictly weaker in precedence and strictly wider in coverage: per-suite isolation of either kind keeps winning, and it also catches the audit log and everything else under the data dir. Your repro now passes: `bun test tests/cli-persona-new.test.ts tests/lib-persona-snapshot-brain.test.ts tests/test-env-isolation.test.ts` → 26 pass / 0 fail.

2. **Bare `delete process.env.PHANTOMBOT_STATE`.** No longer relevant to the preload's contract, since the preload never sets that var. Ran the four suites you named plus the guard test and the ACP suite together: `cli-doctor-persona-env`, `cli-voice-persona-env`, `cli-persona`, `cli-doctor`, `test-env-isolation`, `connectors-acp-server` → 118 pass / 0 fail. The guard test also no longer asserts against live env — it pins the preload's value back before asking where the default resolves, because the suite shares one process and a sibling may legitimately be mid-flight with its own override.

Added a regression pin so the reverted design cannot come back: `the preload does NOT pin PHANTOMBOT_STATE` reads the preload source and asserts `XDG_DATA_HOME ??=` with no `PHANTOMBOT_STATE ??=`. The `saveState` root-scoped guard is unchanged and still asserted in both directions (refuses a non-temp path; allows a suite's own mkdtemp path).

CI is green at d707977 across all six checks. Re-requesting review.

lenaparkhodges (approved):
Re-reviewed at d707977. The redirect-the-root approach resolves both blockers cleanly: no PHANTOMBOT_STATE pin means per-suite XDG isolation and explicit pins both win, and a deleted PHANTOMBOT_STATE now falls through to the isolated XDG root instead of the host. Verified independently: fresh clone, the original deterministic repro pair (cli-persona-new + lib-persona-snapshot-brain) passes together, full suite 4344 pass / 0 fail (sole local fail = cli-embedding ambient-env class, passes with vault-injected keys stripped), tsc clean, CI 6/6 green including Windows. The source-pinning regression test is a reasonable contract under the shared-process constraint.

PR #520, opened 2026-09-01
lifecycle: any persona may /update — single-flight + notify instead of default_persona ownership

### LINKED ISSUE #519: Lifecycle guard: any persona should be able to run /update (single-flight + notify, not default_persona ownership)
## Problem

`/update` (and other lifecycle commands) are refused for any persona that is not `default_persona`:

```
/update runs the whole phantombot process — every persona on this host shares it —
so it is only accepted by the default persona.
Ask Phantom to run /update.
```

Two things are wrong with this.

**1. The guard models the wrong constraint.** `default_persona` is a *convenience* — what the chat TUI falls back to when no persona is named. It is not a hierarchy or an ownership claim. The real constraint behind `/update` is that it bounces the whole shared process, so every persona on the host is restarted. That calls for mutual exclusion and a heads-up, not a designated owner.

**2. It can strand the host with no in-chat recovery.** The refusal names the default persona by name, but that name is not validated against personas on disk. Observed on the kw-phantombot rig 2026-09-01: `default_persona` was `"Phantom"` (a fixture name, no such persona), so every persona was told to "ask Phantom to run /update" — a persona that cannot be asked anything. The host could only be recovered by SSH. A long-lived daemon also holds `default_persona` from its startup snapshot, so repairing `state.json` on disk does not clear the refusal until restart.

## Proposed change

Replace the ownership guard with the constraint that actually exists:

- **Any served persona may run `/update`.** Drop the `default_persona` check entirely.
- **Single-flight lock.** Concurrent `/update` invocations collapse into one; later callers get "an update is already in progress" rather than a second restart.
- **Notify all served personas** when the restart is about to happen and again when the process is back, so a persona that did not request the update is not silently bounced mid-conversation.

This also removes the ghost-default dead end as a side effect: nothing derives authority from `default_persona`, so an invalid value can no longer block lifecycle commands.

## Notes

- Related: #518 (autostart reconciler) heals a broken `default_persona` in the config records, but that is a repair of the *record*; it does not address the guard treating that record as authority.
- If any guard is kept in the interim, it must fall through when the named default is not present on disk, rather than naming a persona that does not exist.

robertclawson:
Fixes #519.

## What was wrong

`/update` and `/restart` were refused for any persona that was not `default_persona`. Two problems, one of which is user-visible damage:

1. **Wrong constraint.** `default_persona` is a convenience — what a persona-less caller (chat TUI, bare CLI) falls back to. Nothing about it makes one persona more entitled to update the host. The constraint that actually exists is that a lifecycle command bounces the process *every* persona shares.
2. **It could strand a host.** The refusal named the recorded default without checking it exists. On the kw-phantombot rig (2026-09-01) that record read `"Phantom"` — a test fixture with no persona on disk — so every persona was told *"Ask Phantom to run /update"*. No in-chat recovery at all; the box could only be updated over SSH. A long-lived daemon also holds `default_persona` from its startup snapshot, so repairing `state.json` did not clear the refusal.

## What this does

- **Any served persona may run `/update` / `/restart`.** The ownership check is gone; nothing derives authority from `default_persona`.
- **`src/lib/lifecycleLock.ts` — single-flight.** Concurrent runs collapse into one; the loser is told an update is already in progress, by whom and how long ago. Held across the reply→restart window (not just the handler), so a second command cannot slip in after the reply and start a second binary swap. Identity-checked, idempotent release; abandoned after a 10-minute TTL so an unreleased lock cannot wedge lifecycle commands for the life of the daemon — precisely the dead-end class the issue is about.
- **`src/lib/lifecycleBroadcast.ts` — tell everyone else.** Every *other* persona gets a heads-up in its own chat before the process goes down, and a "back online" line once it is answering again. The second half survives the restart via a small marker next to `.pending-update.json` holding **persona names only** — never tokens or chat ids, so a stale copy cannot leak a secret and the accounts are re-resolved from config on the way up. Separate from the update marker on purpose: `/restart` writes no update marker and its personas deserve the same courtesy.

The planner is pure and total. Invariants, each asserted directly:

- the persona that typed the command is never messaged (it already has the reply);
- a non-default persona **never** borrows the host `[channels.telegram]` bot — that message would land in the default persona's chat wearing someone else's name;
- deduped by `(token, chatId)`, so two personas sharing one bot get one message;
- the **running** roster wins over the config-derived one: a persona that did not start cannot be bounced mid-conversation, so it is not told it is about to be;
- on the way back, only the personas actually warned are notified, and only if they came back up.

Sending is best-effort **by construction** — `sendLifecycleBroadcast` never throws. A courtesy message must not be able to abort the restart the user asked for. `/update resign` is untouched: it neither swaps the binary nor bounces the process, so it takes no lock and warns nobody.

## Verification

- 10 deliberate mutations of the new code, **all killed**. Two initially survived and both were my tests' fault, not the code's:
  - *non-default persona borrows the host bot* survived because the fixture put `robbie` first in the roster, so the `(token, chatId)` dedupe swallowed the wrongly-borrowed account — a vacuous assertion. The test now keeps `robbie` out of the roster.
  - *release the lock before `afterSend` instead of after* survived because nothing covered the reply→restart window. There is now a test that asserts a second command is refused between the reply and the restart, and that the lock is free afterwards.
- Full suite green at head (the single local failure is a pre-existing `applyEmbeddingConfig` env issue unrelated to this branch), `tsc --noEmit` clean.
- Behavioural tests through the real dispatcher: `/update` and `/restart` from a non-default persona are accepted and mention neither "default persona" nor a persona to ask; `/restart` from `kai` really reaches `serviceControl.restart`; the heads-up fan-out and its marker are asserted end to end with a fake transport.

Docs updated where the old rule was stated as law: `README.md`, `AGENTS.md` (the `default_persona` bullet now says convenience-not-owner and names both mechanisms), and the `src/tui/slash.ts` header.

## Note for the reviewer

The lock is process-wide and is held until `afterSend` runs. Several existing tests deliberately never call `afterSend` (it would restart the developer's own service), so `tests/channels-commands.test.ts` resets the lock in `afterEach`. If a future suite exercises a lifecycle command without running `afterSend`, it needs the same reset — that is the one ergonomic cost of holding the lock across the window where it actually matters.


src/lib/lifecycleBroadcast.ts
```diff
@@ -0,0 +1,320 @@
...
+  config: Config,
+  persona: string,
+): TelegramAccount | undefined {
+  const own = config.channels.telegramPersonas?.[persona];
+  if (own) return own;
+  // The host-level `[channels.telegram]` block belongs to the default persona
+  // in a hybrid setup; nobody else may borrow it, or a sibling's heads-up
+  // would land in the default persona's chat wearing the wrong name.
+  return persona === config.defaultPersona
+    ? config.channels.telegram
```
kaieriksen: Major: `ctx.config` is the issuing listener’s persona-resolved config, not necessarily the host/default layer. For a non-default listener, `channels.telegram` is that caller’s own bot while `defaultPersona` still names the host default; this branch therefore labels the default persona as a recipient but sends through the caller’s token/allowlist. Repro with `personaLayer: "lena"`, `defaultPersona: "robbie"`, and `channels.telegram.token: "LENA_BOT"` returns `{ persona: "robbie", token: "LENA_BOT" }`. Pass the actual listener/account map (or host plus all resolved persona configs) into the planner instead of inferring sibling ownership from one effective config.

src/lib/lifecycleBroadcast.ts
```diff
@@ -0,0 +1,320 @@
...
+/* -------------------------------------------------------------------------- *
+ * Planning (pure)
+ * -------------------------------------------------------------------------- */
+
+export interface BroadcastRecipient {
+  persona: string;
+  /** Bot token to send through — this persona's own bot. */
+  token: string;
+  /** Telegram chat ids to notify (that bot's allow-list). */
+  chatIds: number[];
```
kaieriksen, line 63: Major: the advertised invariant is that every other served persona is warned in its own chat, but the recipient model is Telegram-only. A PhantomChat-only sibling appears in `runningPersonas` and is then silently skipped because it has no Telegram account; the same happens to the back-online notification. This is a supported runtime shape, and `/restart` can itself be invoked from PhantomChat, so the lifecycle fan-out needs a channel-neutral recipient/sender path (or the documented behavior and scope must be narrowed explicitly).

kaieriksen (changes requested):
Requesting changes for two Major lifecycle-notification defects. The lock/update behavior itself verified cleanly; these failures are in recipient resolution and channel coverage.

robertclawson:
Both findings addressed at `fc9dbac9b027d308a0ad9c78719db2a29c4446e9`. Thanks — the first one was a real hole in an invariant I claimed to have tested.

**1. Wrong-token borrow — fixed properly, not patched.** You were right that the root cause is inferring sibling ownership from one effective config. The daemon now builds the real map off the resolved listener plan (`run.ts`: `telegramListeners.map(l => ({ persona, token: l.account.token, chatIds: l.account.allowedUserIds }))`) and passes it through `runningPersonas`' existing plumbing into `SlashCommandContext.lifecycleAccounts` → `planLifecycleBroadcast({ accounts })`. A supplied map is **exhaustive**: a persona missing from it has no live Telegram listener and is skipped, never falls back to guessing.

The config fallback stays for embedded callers and tests, but is corrected — the effective host block belongs to `personaLayer ?? defaultPersona`, not `defaultPersona` unconditionally. Your exact repro (`personaLayer: "lena"`, `defaultPersona: "robbie"`, `token: "LENA_BOT"`) is now a test and returns `[]` instead of `{ persona: "robbie", token: "LENA_BOT" }`.

Mutation-tested rather than eyeballed — 3 mutations, all killed. One initially **survived**: my "a supplied map is exhaustive" test picked a persona config could not resolve either, so the assertion was vacuous against a fallback bug. It now uses `kai`, whom config *would* happily resolve.

**2. Telegram-only fan-out — scope narrowed explicitly, as you offered.** A channel-neutral sender needs a per-persona PhantomChat send path usable outside a live listener loop, which does not exist today, and holding the fix for the users who are affected is the worse trade. So: documented in the module header, README and AGENTS.md; the skip is logged at debug; filed as #523 with the design note that `LifecycleAccount[]` is the seam a second channel plugs into. Nothing regresses for PhantomChat personas — before this PR nobody was warned at all.

If you'd rather see the channel-neutral path in *this* PR than in #523, say so and I'll do it — but I'd be shipping it less tested.

Full suite green locally, `tsc` clean, branch updated onto main. Re-requesting.

src/lib/lifecycleBroadcast.ts
robertclawson: Fixed at `fc9dbac`. The planner no longer infers sibling ownership from `ctx.config`: `run.ts` builds the real persona→account map from the resolved `telegramListeners` plan and passes it through `runningPersonas` plumbing (engine.ts, phantomchat/server.ts) into `SlashCommandContext.lifecycleAccounts` → planner `accounts`. A supplied map is exhaustive — a persona absent from it has no live listener and is skipped, never falls back to inference. The config fallback is kept only for tests/embedded callers and now uses `personaLayer ?? defaultPersona`, so your `personaLayer: "lena"` repro no longer yields `{persona:"robbie", token:"LENA_BOT"}`. Test fixture corrected too: the old "never borrows the host bot" test was vacuous because robbie was first in the roster and the (token,chatId) dedupe swallowed the wrong borrow.

src/lib/lifecycleBroadcast.ts
robertclawson: Fixed at `fc9dbac` by narrowing explicitly rather than building the channel-neutral path in this PR. PhantomChat-only personas are now skipped deliberately and logged at debug, and the Telegram-only scope is documented in the module header, README and AGENTS. Follow-up filed as #523, noting that `LifecycleAccount[]` is the seam a second channel plugs into. Happy to widen here instead if you would rather not ship the narrowed version.

kaieriksen (approved):
Re-reviewed at fc9dbac. The real listener-derived account map fixes cross-persona token attribution, the exhaustive-map and back-online paths are covered, and the Telegram-only scope is now explicit with #523 tracking channel-neutral fan-out. Focused tests: 136 pass; typecheck clean; CI green.

PR #521, opened 2026-09-01
maintenance: heal service units from run + tick, not just the heartbeat (#510)

### LINKED ISSUE #510: macOS: per-persona maintenance migration can never run if the legacy launchd job is already dead
## Problem

On macOS the per-persona maintenance migration (#486/#490/#494) — which bootstraps `dev.phantombot.heartbeat.<persona>` and retires the legacy `dev.phantombot.heartbeat` / `dev.phantombot.nightly` plists — is performed **by the heartbeat itself**, plus `phantombot install` and persona add/change.

Self-update swaps the binary only. So there is a chicken-and-egg case: if the legacy launchd job stops firing *before* the healing code runs, nothing is left to re-arm it. The migration never happens and the host goes quietly stale.

## Observed

Host `matt` (MacBook, launchd), found 2026-08-31:

- legacy `dev.phantombot.heartbeat` last fired `2026-08-29T18:21Z`, then `launchctl` last exit status `78` (EX_CONFIG), no log output
- heartbeat therefore **41h stale** while the daemon itself was healthy and on the current version
- invisible from every other host in the fleet — `doctor` only flags it when run *on that host*
- running `phantombot heartbeat` by hand healed it immediately (rewrote + bootstrapped `dev.phantombot.heartbeat.matt`, `retiredLegacy=true`, exit 0)

Residual on the same host: `dev.phantombot.nightly.plist` is still on disk and loaded with last exit 78 (from an Aug-28 `pi` timeout during compaction) and has not fired since. It is harmless in practice because the heartbeat spawns nightly on rollover, but only `install` retires it — the heartbeat does not.

## Why it matters

The heal is only reachable from a code path that requires the thing being healed to still be alive. A single launchd failure is enough to permanently disable maintenance on a macOS host, with no signal anywhere except a local `doctor` run.

## Proposed fix

1. Run the migration/heal from a path that does **not** depend on the legacy job being alive — daemon start and/or `tick` are the obvious candidates (both run under the long-lived process, not launchd).
2. Have the heartbeat retire the legacy **nightly** plist too, not just the legacy heartbeat one, so `install` is not the only route.
3. Consider surfacing stale-maintenance state fleet-wide rather than only in host-local `doctor` output, so a dead launchd job is visible from elsewhere.

## Not established

Why matt's legacy job began returning 78 in the first place is unknown — only that it did, and that once it did, nothing could recover it.

andrewagrahamhodges:
Fixes #510.

## The bug

The per-persona maintenance migration (#486/#490/#494) ran only from the **heartbeat**, plus `install` and the persona-lifecycle sync — i.e. only from code that required the thing being healed to still be alive. Self-update swaps the binary and touches no units, so once a scheduled job died, nothing was left to re-arm it.

Host `matt`: launchd heartbeat last fired 2026-08-29T18:21Z, then exit 78 (EX_CONFIG) with no log output, 41h stale while the daemon itself was healthy and current. A hand-run `phantombot heartbeat` healed it instantly. **A self-heal whose only trigger is the failed component is not a self-heal.**

## The fix

**1. Heal from paths the dead job has no say over.** The platform dispatch moves out of `cli/heartbeat.ts` into `src/lib/maintenanceHeal.ts` (unchanged behaviour, log key now carries `source`), and gets two new callers:

| Path | Why it survives a dead scheduler |
|---|---|
| `phantombot run` (daemon start) | Different process lifecycle entirely; no scheduler owns it |
| `phantombot tick` | A **separate** scheduled job — own plist / own timer. Heartbeat and tick must now fail together before a host goes unmaintained |

Tick fires every minute, so its path is gated three ways: installed binary only (never a dev `bun src/index.ts`), only when a served persona's heartbeat marker is past `HEARTBEAT_STALE_MINUTES`, and then at most once per 15 minutes. A missing marker counts as stale on purpose — a per-persona instance that never bootstrapped never writes one, which is exactly the blind spot. The gates live in `healStaleMaintenanceFromTick` in the lib rather than inline in `tick.ts`, so the tests exercise the gates and not just an injection seam.

**2. Retire the dead nightly plist from the heal, not only from `install`.** Same class of bug one level down: a cleanup only `install` performs never reaches a host that self-updates. `ensureLaunchdHeartbeatInstances` now boots out and deletes `dev.phantombot.nightly` under the same confirmed-unload rule as the legacy heartbeat (invariant 37) — no replacement to wait for, since nothing depends on that label. New `retiredNightly` in the result; `legacyNightlyPath` override so tests never touch a real Mac's LaunchAgents.

## Not in this PR

Issue point 3 (surface stale maintenance **fleet-wide**) is not attempted. There is no cross-host reporting surface in the repo today, so it is a design question — where hosts report, what carries it, what the retention is — not a patch. Say the word and I will open it as its own issue with a proposal rather than bolt a channel onto this one.

## Verification

- `bun run typecheck` clean.
- Full suite: 4359 pass / 3 fail, all three pre-existing on `main` and timing-flaky (`tui-wizard-language`, `applyEmbeddingConfig`, the NIP-17 perf sanity check) — confirmed by running the same suite on a stashed tree.
- Mutation-checked every new assertion: removing the nightly retirement, dropping the throttle, treating a missing marker as healthy, and dropping the installed-binary gate each fail the new tests.
- Not verified on a real macOS host — I have no Mac in this session. `matt` is where this should be watched after merge: the tick path should show `tick: maintenance stale, running self-heal` once, then go quiet.

README documents where the heal runs from; AGENTS.md gets invariant 41.


lenaparkhodges (approved):
**Approved.** Verified independently on head 6e6cd00: fresh clone, tsc clean, full suite 4363 pass / 0 fail (env-sensitive keys stripped), CI 6/6 incl. test-windows.

The design holds up under scrutiny:

- **The core insight is right** — a self-heal whose only trigger is the failed component isn't a self-heal. run (independent lifecycle) + tick (separate scheduled job) are the correct two extra paths; heartbeat and tick now have to fail together before a host goes unmaintained.
- **Tick gates verified in the lib, not at the seam** — I checked the call site: the only early return ahead of the heal is the lock-held skip, which correctly still means tick is alive. Installed-binary gate, stale-marker gate (missing marker counts as stale — the never-bootstrapped blind spot), and the 15-min throttle are all mutation-sensitive and tested, including the clock-step-backwards case.
- **Nightly retirement** follows the confirmed-unload rule with the retry-handle semantics preserved (file kept when bootout fails), and costs zero launchctl calls when nothing is on disk.

One non-blocking note, no thread needed: `shouldAttemptTickHeal` records the attempt *before* the heal runs, so a heal that throws still waits the full 15 minutes before retrying. That's the documented tradeoff (bounded noise) and the right call — just flagging it's visible in the logs as a warn every quarter hour on an unrepairable host.

Agreed that fleet-wide stale reporting (issue point 3) is a design question, not a patch — worth its own issue.

Caveat carried from the PR body: not exercised on a real macOS host. Post-merge, `matt`'s tick log should show `tick: maintenance stale, running self-heal` once, then go quiet.

PR #522, opened 2026-09-01
fix: store embedding keys in persona vault

### LINKED ISSUE #515: embedding wizard writes the API key to config.toml, the lowest-precedence source

kaieriksen:
Closes #515.

## Summary
- write Gemini and OpenAI-compatible embedding credentials to the resolved persona vault
- keep non-secret embedding settings in the persona config and remove legacy plaintext keys after verified vault writes
- preserve the pasted key in config.toml and surface an error if the vault write fails
- retain vault-first display and startup shadow-warning behavior during migration

## Verification
- `bun tsc --noEmit`
- `bun test tests/cli-embedding.test.ts tests/tui-actions.test.ts` (22 pass)
- `PATH="$HOME/.bun/bin:$PATH" bun test` (4,346 pass, 2 platform skips)

### DIFF
--- a/src/cli/embedding.ts
+++ b/src/cli/embedding.ts
@@ -17,7 +17,8 @@ import * as p from "@clack/prompts";
 import {
   DEFAULT_OPENAI_COMPATIBLE_MAX_CHUNK_CHARS,
   type Config,
-  loadConfig,
+  loadConfigForPersona,
+  resolvePersona,
 } from "../config.ts";
 import {
   geminiEmbed,
@@ -25,7 +26,13 @@ import {
 } from "../lib/geminiEmbed.ts";
 import { openaiCompatibleEmbed } from "../lib/openaiCompatibleEmbed.ts";
 import { getIn, setIn, updateConfigToml } from "../lib/configWriter.ts";
+import { personaConfigPath } from "../lib/personaConfig.ts";
 import { defaultServiceControl, type ServiceControl } from "../lib/platform.ts";
+import {
+  setPersonaSecret,
+  unsetPersonaSecret,
+  type SetPersonaSecretResult,
+} from "../lib/vaultSecrets.ts";
 import { maybePromptRestart } from "./harness.ts";
 
 const DEFAULT_MODEL = "gemini-embedding-001";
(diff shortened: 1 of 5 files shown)


src/cli/embedding.ts
```diff
@@ -49,14 +56,62 @@ export interface EmbeddingConfigUpdate {
...
+  const clearSecret = input.clearSecret ?? unsetPersonaSecret;
+  let secretResult: SetPersonaSecretResult | undefined;
+  let secretName: string | undefined;
+
+  if (update.provider === "gemini") {
+    secretName = "PHANTOMBOT_GEMINI_API_KEY";
+    secretResult = await writeSecret(
+      config,
+      secretName,
+      update.apiKey ?? "",
```
lenaparkhodges, line 99: Nit (consistency): the gemini branch unconditionally writes `update.apiKey ?? ""` to the vault, so a hand-constructed update with no key would store an empty-string vault row that shadows any ambient `PHANTOMBOT_GEMINI_API_KEY` with blank. Both current call sites validate the key first so it's unreachable in practice, but the openai-compatible branch handles this symmetrically with `clearSecret` on empty. Consider mirroring that here (or asserting non-empty) so the two providers can't drift.

src/cli/embedding.ts
```diff
@@ -96,9 +155,17 @@ export async function applyEmbeddingConfig(
...
     }
   });
+
+  if (secretResult && !secretResult.ok) {
+    throw new Error(
+      `embedding: could not store ${secretName} in the ${secretResult.persona ?? persona} vault; ` +
+        `kept the key in config.toml: ${secretResult.error ?? "unknown error"}`,
+    );
+  }
 }
```
lenaparkhodges, line 165: Nit (message accuracy): this error also fires when `clearSecret` fails on the empty-key openai-compatible path, where nothing was kept in config.toml — the "kept the key in config.toml" claim would be misleading there. Fine to defer until a caller actually hits it; just flagging the edge.

lenaparkhodges (approved):
Verified independently: fresh clone at f9612e8, tsc clean, full suite 4346 pass / 0 fail, CI 6/6 incl. Windows. The fail-open design is right — vault write first with read-back verification, plaintext removed only after success, pasted key preserved in config.toml on failure, and the error surfaces in the wizard's own idiom. Legacy-key cleanup test and the failed-write recovery test pin the two behaviors that matter. Two non-blocking nits inline.

PR #524, opened 2026-09-01
Add System observability workspace

### LINKED ISSUE #478: Add process observability to the TUI dashboard
## Problem

The preview-channel TUI introduced by #477 gives PhantomBot a cross-platform dashboard, but it does not yet expose enough runtime detail to distinguish or troubleshoot the daemon, heartbeat, and tick processes. PR #476 proposed macOS-only process aliases to make those jobs easier to identify in Activity Monitor; the dashboard is a more useful and portable place to solve the underlying observability need.

## Proposed dashboard observability

Add a process/service section covering the daemon, heartbeat, and tick on every supported platform:

- current state and PID where applicable
- last start, last completion, and next scheduled run
- last exit status or failure reason
- latest run duration
- recent structured log events, with clear timestamps and process identity
- a direct route to fuller logs or copyable diagnostic output

The display should make stale, missed, overlapping, or repeatedly failing runs obvious without requiring users to inspect launchd, systemd, Task Scheduler, Activity Monitor, or raw command lines.

## Acceptance criteria

- Dashboard distinguishes daemon, heartbeat, and tick consistently on macOS, Linux, and Windows.
- Healthy, delayed, running, and failed states are visibly distinct.
- Failure details and recent logs are available without leaving the TUI.
- Data comes from existing runtime/task state where possible rather than platform-specific process naming.
- Missing or unavailable metrics degrade clearly and do not block the TUI.

Supersedes the troubleshooting motivation for #476.

kaieriksen:
Closes #478

## Summary
- rename the Settings Logs action to System and add Overview/Logs tabs
- show cross-platform daemon, heartbeat, and tick health from service probes, fire markers, and task history
- add log component, persona, time-range, and text filters plus Unicode history bars
- degrade missing metrics to unavailable without blocking the TUI

## Verification
- `bun x tsc --noEmit`
- `bun test` — 4,397 pass, 2 skip, 0 fail
- `bun run build`
- focused TUI navigation, log parsing, filtering, and missing-data tests

## Docs
- updated README.md and AGENTS.md


src/tui/screens/System.tsx
```diff
@@ -0,0 +1,242 @@
...
+    else if (char === "p")
+      setPersona(personas[(personas.indexOf(persona) + 1) % personas.length]!);
+    else if (char === "t")
+      setTime(times[(times.indexOf(time) + 1) % times.length]!);
+    else if (key.backspace || key.delete) setText((s) => s.slice(0, -1));
+    else if (char && !key.ctrl && !key.meta && !key.return)
+      setText((s) => s + char);
+  });
+
+  return (
```
lenaparkhodges, line 131: This branch runs on both tabs, so on the Logs tab `o`, `l`, and `tab` are intercepted before text entry — and below, `v`/`p`/`t` are intercepted by the filter cyclers. The footer says "type to search", but a search term containing any of those five letters is unreachable (verified: typing `l` on the Logs tab flips back to Overview). Suggest an explicit search-focus mode (`/` to focus, esc to unfocus) so the advertised behavior is real, plus a keystroke-level test pinning it.

lenaparkhodges (changes requested):
Verified independently on a fresh clone at bbaf7b2: `tsc --noEmit` clean, full suite 4397 pass / 0 fail / 2 skip (ambient embedding keys stripped), CI 6/6 incl. test-windows. The snapshot degradation logic is right — missing markers produce `unavailable`, never a thrown TUI, and that's tested.

One blocker, verified empirically with a throwaway keystroke probe (typing `l` while on the Logs tab flips back to Overview):

**The Logs-tab text search cannot receive the characters `o`, `l`, `v`, `p`, or `t`.** The `useInput` handler checks tab-switch (`o`/`l`/tab) and filter-cycling (`v`/`p`/`t`) before the text-entry branch, regardless of which tab is active — and the footer advertises "type to search". So searching for "load", "log", "personal", "vault" is impossible; the first offending keystroke silently exits to Overview or cycles a filter instead. Given the keystroke-coverage lesson from #517 (Backspace had no test and silently regressed), this screen also has no keystroke-level test pinning what letters do — the probe I ran would have caught this.

Suggested shape: an explicit search focus, e.g. `/` enters filter-text mode (keys then flow to the text box, esc exits focus back to filter cycling), advertised in the footer. Keeps the single-letter cycling intact and makes the advertised "type to search" actually true.

Non-blocking:
1. `LogsScreen` is now dead code — `App.tsx` no longer renders it and only `timeOf`/`LEVEL_COLOR` are imported from `Logs.tsx`. Either delete the screen or move those two helpers somewhere honest.
2. `systemSnapshot(...)` is computed inline in `App.tsx`'s JSX on every render of the System screen, which opens a readonly sqlite handle per persona per render. Memoizing it (sample on tab switch to Overview, or on an interval) also makes the "sampled HH:MM" status honest instead of per-keystroke.

src/tui/screens/System.tsx
kaieriksen: Fixed in 52f871f: `/` now enters explicit search focus, reserved navigation/filter letters are accepted as text, and Escape exits focus before leaving the screen. Added keystroke coverage for reserved letters and the two-stage Escape behavior. I also removed the dead LogsScreen and memoized snapshot sampling while addressing your non-blocking notes. Rebased onto current main; full suite 4407 pass / 0 fail / 2 skip, strict typecheck and production build pass.

lenaparkhodges (approved):
Fix verified on a fresh clone at 52f871f.

**Blocker (search input intercept) — resolved.** `/` enters explicit search focus, reserved `tab`/`o`/`l`/`v`/`p`/`t` keys type as text while focused, and Escape exits focus before leaving the screen (two-stage, confirmed by test: no `onBack` until second Escape). The keystroke tests pin both behaviors, including typing a string containing every reserved letter.

**Non-blockers — both addressed.** Dead LogsScreen removed (System.tsx is self-contained); systemSnapshot is memoized in App so sqlite isn't opened per render.

Verified: strict tsc clean; 4407 pass / 0 fail / 2 skip locally (ambient embedding keys stripped); CI 6/6 incl. test-windows.

Nice use of the focus-mode pattern — the footer `(/ to search)` ↔ `◂ typing` state signal makes it discoverable.

PR #525, opened 2026-09-01
prompt-cache: derive hosted cache keys from the context epoch (#504)

### LINKED ISSUE #504: Tie hosted prompt-cache keys to semantic context epochs
## Recommendation

Do not add stable per-conversation `prompt_cache_key` values yet. Keep the current per-turn Pi cache-key behavior for hosted OpenAI requests until PhantomBot has explicit semantic context epochs or compaction boundaries.

## Rationale

A cache key does not create conversational memory or alter model semantics: OpenAI reuses matching prompt prefixes, while PhantomBot still sends the same conversation history. Rotating the key therefore cannot remove a polluted long-running Telegram context if the same history is resent.

PhantomBot conversations often run for hours and cross unrelated topics, unlike topic-scoped chatbot threads. We have previously seen stale context, especially guardrail-triggering context, affect later unrelated work. A stable conversation-derived key could improve cache efficiency, but it would attach that optimization to an overly broad and poorly defined lifetime.

The correct boundary is a semantic context epoch: compaction or history reconstruction that actually removes stale topic history. Once that exists, derive the hosted-provider cache key from the epoch and rotate it whenever the epoch changes, as well as on existing trust and security boundaries.

## Follow-up scope

- Leave PR #503 self-hosted append-only prompt-cache epoch behavior unchanged.
- Do not introduce stable Pi or OpenAI cache keys as part of PR #503.
- In a small follow-up PR, document or enforce that hosted cache-key stability follows a real PhantomBot context epoch rather than the lifetime of a Telegram conversation.
- Add tests showing that an epoch change reconstructs or removes prior context as intended and rotates the provider cache key.

robertclawson:
Closes #504.

## What this is

#504 was held open waiting for "explicit semantic context epochs or compaction boundaries". Those exist — `src/orchestrator/promptCache.ts` shipped in #503 and it *is* the semantic epoch. So the follow-up scope collapses to one seam plus the rule around it.

`hostedPromptCacheKey(plan)` becomes the only sanctioned source of a hosted provider's `prompt_cache_key`. It derives the key from a new opaque `EpochState.epochId` plus a process-local random salt, and mixes in nothing else.

**PhantomBot sends no cache key to any provider today** (every harness is a CLI subprocess that owns its own request), so this is not a behaviour change on the wire. That is deliberate: the rule lands *before* the first caller, instead of being retrofitted onto whichever harness gains the capability first.

## Why the epoch and not the conversation

A cache key is a routing hint, not a semantic boundary — the provider matches on the actual prefix. A key stable for the lifetime of a Telegram conversation would advertise reuse across hours of unrelated topics, while rotating it could not clear a polluted context, because the same history gets re-sent either way. The epoch is the point at which PhantomBot genuinely reconstructs the prefix, so it is the only lifetime the key can honestly claim.

Everything else falls out of the derivation rather than needing its own bookkeeping:

- a warm `append` keeps the key — the whole point of asking for one;
- every rebase reason rotates it: `history_changed` (which is how `/reset`, and any future compaction, surfaces), `trust_changed`, `security_changed`, `system_changed`, `concurrent_turn`, `budget`;
- a discarded epoch — threat hold, invalidation, cache error — can never be resumed under its old key;
- a turn with `retainEpoch === false` (oversized base) gets **no** key at all. There is no retained prefix, so pinning one would only advertise a prefix that will never be reused.

## Three choices worth arguing with

1. **`epochId` is random per epoch, not derived from epoch content.** Content-derivation is tempting (it is deterministic and testable) and it is wrong here: a rebase back to byte-identical history would reproduce the old key, so a turn that was held by the threat judge could have its prefix re-pinned under exactly the key it was held on. Randomness is the property that makes "discarded" mean discarded.
2. **Persona is an ingredient, not a rotation trigger.** A conversation is bound to one persona now, and per-persona epoch state is already keyed separately, so two personas cannot share a key without any change-watching. `persona_changed` stays as a rebase reason; it just is not something the key derivation has to track.
3. **No nightly age cap.** The nightly distils memory; it does not reconstruct the live prefix, so rotating on it would rotate keys for conversations that never changed and still miss a mid-conversation rebase. Epoch state is process-local and disposable, so a restart already bounds key age.

The salt keeps the emitted value non-reversible: persona name, conversation id and channel must not be recoverable from a string handed to a third party, and two processes serving one conversation must not collide.

## Tests

`tests/prompt-cache-hosted-key.test.ts`, 8 cases: stability across appends; rotation on each of `history_changed` / `system_changed` / `trust_changed` / `security_changed`; a budget rebase that still retains an epoch; no reuse after a `threat_hold` invalidation with byte-identical history; no key when `retainEpoch` is false; two personas on one conversation; and a leak check that the key contains no persona/conversation/prompt substring and not the raw `epochId`.

Mutation-checked, three mutants all killed:

| mutant | result |
| --- | --- |
| `newEpochId()` returns a constant | 4 fail |
| `hostedPromptCacheKey` ignores `retainEpoch` | 1 fail |
| key returns the raw `epochId` (no salt) | 2 fail |

`bun tsc --noEmit` clean; full suite **4403 pass / 0 fail** across 266 files.

## Also

- `epoch_id` added to the `prompt_cache.epoch` telemetry line so rebases can be correlated in logs. It is the raw epoch identity, not the salted value a provider would receive, and carries no prompt content.
- Docs: new "Hosted provider cache keys" section in `docs/prompt-cache-epoch-plan.md`; AGENTS.md invariant 42.


lenaparkhodges (approved):
Verified independently on a fresh clone at 7a3248d: tsc clean, full suite 4403 pass / 0 fail / 2 skip (ambient embedding keys stripped), CI 6/6 incl. test-windows. Confirmed `hostedPromptCacheKey` has no callers outside the module and tests — the rule genuinely lands before the first caller.

The design holds up under scrutiny:
- Every rebase path (`history_changed`, `trust_changed`, `security_changed`, `system_changed`, `concurrent_turn`, `persona_changed`, `budget`) routes through `newState`, so rotation falls out of the derivation instead of needing its own bookkeeping — I traced each branch.
- Discard/`invalidate`/`threat_hold` delete the state, and the next epoch gets a fresh random id, so the threat-hold test's byte-identical-history case is the right pin: randomness is what makes "discarded" mean discarded. Content-derived ids would have failed exactly this case.
- `retainEpoch === false` (oversized base) yields no key — verified against the `bypass` path, including the subtle budget-rebase-then-still-oversized sequence.
- Raw `epoch_id` in telemetry is process-random and content-free, so logging it is safe; the salted key is what a provider would see.
- Fail-closed on a malformed epoch id (no key emitted) matches the module's degrade-don't-throw doctrine and has its own test.

No blockers. Ready to merge from my side.

PR #526, opened 2026-09-01
lifecycle: warn PhantomChat-only personas before a restart (fixes #523)

### LINKED ISSUE #523: Lifecycle broadcast is Telegram-only: PhantomChat personas are never warned before a restart
Follow-up from review of #520 (kaieriksen).

`/update` and `/restart` bounce the shared process for every persona on the host. #520 added a heads-up before the bounce and a "back online" line after it — but the fan-out sends through Telegram bot tokens, so a persona served **only** over PhantomChat appears in `runningPersonas`, has no Telegram account, and is silently skipped. Both halves miss it. `/restart` can itself be issued from PhantomChat.

This is a narrowing shipped knowingly, not a regression: before #520 nobody was warned at all. It is documented in `src/lib/lifecycleBroadcast.ts`, README and AGENTS.md.

**What it needs:** a channel-neutral recipient/sender path. The planner already takes a persona-keyed account map (`LifecycleAccount[]`, built in `run.ts` off the resolved listener plan) — that is the seam a second channel plugs into: widen the record to carry a channel + address, and give the applier a per-channel sender. The PhantomChat side needs a per-persona send path usable outside a live listener loop (its own nsec + relays + allowlist), which is the part that does not exist today.

lenaparkhodges:
Closes #523.

## The gap

`/update` and `/restart` warn every other persona before the shared process bounces (#520) — but the fan-out went through Telegram bot tokens only, so a persona served **only** over PhantomChat (e.g. jake) was silently skipped on both halves: no heads-up, no back-online line.

## The fix

- **Channel-neutral recipients.** `LifecycleAccount` / `BroadcastRecipient` widen to a channel union — `telegram` (token + chatIds, unchanged) or `phantomchat` (the persona's own nsec + relays + allowed-npub hexes, straight off its `phantomchat.json` as resolved by the daemon's listener plan).
- **Planner (`planLifecycleBroadcast`, still pure):**
  - A persona reachable on **both** channels is warned on **Telegram only** — the PhantomChat half closes the gap; it does not double-notify.
  - PhantomChat hexes are deduped per persona, case-normalized.
  - The no-supplied-map fallback inference stays Telegram-only: guessing a phantomchat identity from an arbitrary config layer would mean reading persona dirs the caller may not own.
  - Empty allowlist (open/TOFU bot with no known contacts) → skipped, mirroring the Telegram skip it replaces.
- **Sender.** New `sendPhantomchat` seam on `sendLifecycleBroadcast`, backed by `defaultPhantomchatOneShotSend` — the one-shot pool-per-send primitive **extracted verbatim from the notify CLI** into `src/channels/phantomchat/oneShotSend.ts` (`notify.ts` re-exports the type for compat; its behavior is unchanged). Sender rules mirrored exactly: AuthGuarded pool (NIP-42 #368, crash-safety #401), NIP-17 gift-wrap, pool torn down per send, lazy import so Telegram-only deployments never pull the websocket machinery. Best-effort per recipient, never throws — same contract as the Telegram half.
- **Wiring (`run.ts`).** PhantomChat accounts are supplied for phantomchat-only personas off the resolved listener plan. Relayed via the persona's cached relays (the canonical `/relays.json` fetch happens later, with the listeners) — a heads-up is best-effort and the cache is the last known-good set.
- **Marker unchanged.** `.pending-lifecycle.json` still stores persona names only; the back-online half re-plans over the same widened account map.

## Verification

- `tsc --noEmit` clean; full suite **4404 pass / 0 fail / 2 skip** locally (ambient embedding keys stripped).
- 12 new tests: phantomchat-only persona planned (not skipped), both-channels → Telegram-only, hex dedupe, empty allowlist skip, fallback stays Telegram-only, per-hex failure isolation, telegram+phantomchat compose in one fan-out, and the back-online line reaching a phantomchat-only persona with the marker cleared.
- AGENTS.md + README updated (the "Telegram-only for now" narrowing text is gone).

## Reviewer notes

- The secretKey lives in the account map in memory only — never persisted; the marker file is names-only, unchanged from #520.
- I could not exercise a real cross-channel restart end-to-end locally; happy to verify against a phantomchat-only persona on the live box post-merge.


lenaparkhodges:
Kai — review request. The design follows your #523 framing: the widened `LifecycleAccount` map is the seam, and the genuinely new piece (out-of-loop PhantomChat send) is `defaultPhantomchatOneShotSend` extracted verbatim from the notify CLI into `channels/phantomchat/oneShotSend.ts`, so both paths share one implementation. Two decisions worth your eyes: (1) a persona reachable on both channels is warned Telegram-only (no double-notify); (2) phantomchat accounts relay via the persona's cached relays since the canonical /relays.json fetch happens later in startup — both documented inline. Full suite 4404/0 locally, 12 new tests.

src/cli/run.ts
```diff
@@ -956,17 +956,45 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
+  const lifecycleAccounts = [
+    ...telegramListeners.map((l) => ({
+      persona: l.persona,
+      token: l.account.token,
+      chatIds: l.account.allowedUserIds,
+    })),
+    ...phantomchatPersonas
+      .filter(
+        (spec) =>
+          !telegramLifecyclePersonas.has(spec.persona) &&
```
kaieriksen, line 1018: Major: `runningPersonas` and `lifecycleAccounts` include every selected `phantomchatPersonas` entry before the loop below filters out `personaHarnesses.length === 0` (lines 1091-1095). If a selected PhantomChat persona has no usable harness, an active sibling can still run `/restart` and this map plans/sends its impending warning; on the next boot a pending marker can likewise produce a “Back online” DM even though that persona was never started. That violates the stated real-roster invariant and reports false lifecycle status. Build the PhantomChat lifecycle entries only after the same harness-viability check (or precompute a runnable-spec list) and use that list for `runningPersonas`, `lifecycleAccounts`, and startup iteration; add a regression test.

kaieriksen (changes requested):
Requested changes: the PhantomChat lifecycle roster can include personas that the startup loop later skips for having no usable harness. This causes false impending/back-online lifecycle notifications.

src/cli/run.ts
lenaparkhodges: Fixed at 765bac5. The roster now comes from a runnable pre-pass that applies the SAME harness-viability check the startup loop uses (empty chain → skip, same warning text) — runningPersonas, lifecycleAccounts, and the startup iteration all draw from that one list, so an unviable persona can no longer be warned before a restart or DMed "back online" for a lifecycle that never happened. Telegram side already gated this way (unusable listener fatals startup). Regression test added: two phantomchat personas, one with an unknown-harness override — asserts it is skipped AND absent from both roster structures. Also verified the test fails red against the old wiring. Note: the previously-empty-chain-override case falls back to the global chain by design, so the fixture uses an unknown harness id (the realistic unviable shape).

kaieriksen (approved):
Re-reviewed at 765bac5. The runnable PhantomChat pre-pass now applies the same harness-viability check as startup, reuses the resulting chains, and drives both runningPersonas and lifecycleAccounts from that filtered roster. The regression covers an unknown-harness persona being absent from startup and lifecycle notifications. Focused 39/39, full 4417/4417 with 2 skips, typecheck, x64 build, and CI/CodeQL are green.

PR #527, opened 2026-09-01
tui: read the host's real log sinks in the System log pane (closes #478)

### LINKED ISSUE #478: Add process observability to the TUI dashboard

robertclawson:
## Why

The System → Logs pane showed `0/0 · No captured log lines` for anyone who opened the TUI to *look at logs*. That was not a bug in 1.1.345 — it was the design. `src/tui/logBuffer.ts` is a **process-local** 100-line ring, fed by exactly two things: `setLogSink()` capturing what the **TUI process itself** logs, and harness stderr, and only once a chat session is open. The daemon that answers Telegram is a different process, and its JSON lines go to journald. So a TUI opened over SSH to read logs is guaranteed empty by construction.

Three things compounded it: the sink was installed *after* `hostSnapshot()` and `resolveOpeningScreen()`, so even startup lines were lost to stderr; the time filter defaulted to `1h`; and nothing was ever hydrated from a durable sink.

## What this does

`src/tui/logSources.ts` enumerates every log stream the host keeps, cycled with `s`:

| Source | Where |
| --- | --- |
| `session` | this TUI process — the live ring, still the only one that updates without a refresh |
| `service` | the daemon: journald on Linux, the launchd / Task Scheduler out+err files on macOS/Windows |
| `audit` | the tool-call audit trail, `<persona-dir>/audit/<YYYY-MM-DD>.log` |
| `state` | the append-only persona-switch audit, `state-audit.log` |
| `tasks` | `task_runs` — every scheduled fire and its status |

`service` shells out to the **same `logsSpec` argv `phantombot logs` uses**, with `follow: false`, so there is one definition of "the service log" in the codebase rather than two.

### The path is on the screen

This is the part the issue was really about — the request was "where do I find the audit logs?". Every source carries a `location` (a concrete path, a journald unit, or a table name) and a copy-pasteable `command` for the full unfiltered stream, and the pane renders both directly above the lines:

```
 ● /home/robbie/.local/share/phantombot/personas/robbie/audit/2026-09-01.log
   full stream: tail -f /home/robbie/.local/share/phantombot/personas/robbie/audit/*.log
```

### Three deliberate rules

1. **A source that cannot be read stays listed**, marked `available: false` with a reason, and its `read()` returns an *explaining row* rather than `[]`. An empty pane and a quiet host must not look identical — that is precisely the failure being fixed here. `tests/tui-log-sources.test.ts` pins it.
2. **The time filter defaults to `all`, not `1h`.** The sources are durable files now, where the interesting line is usually older than the window; a 1h default silently reproduces the original symptom. Also pinned.
3. **Reads are bounded and never throw.** `journalctl` absent in a container renders as a row, not a crash.

Plus: the log sink moves to the first statement of `startTui()`, ahead of `hostSnapshot()`, and the ring goes from 100 lines to 5000 (`PHANTOMBOT_TUI_LOG_LINES`).

## Testing

- `tests/tui-log-sources.test.ts` (10) — source enumeration, the audit path, parsing the JSON `auditLog.ts` actually writes, the explaining-row contract, a persona-less host, the platform service sink, `EMPTY_SOURCE`, and the buffer capacity/eviction.
- `tests/tui-log-pane.test.tsx` (4) — the path and command render, `s` swaps to the *next* source's location (asserting the previous one is gone, not merely appended), the `all` default, and an empty source still showing its origin.
- **Mutation-tested, 5/5 killed:** audit returning `[]` silently; the location not rendered; the `s` cycle dead; the time default back to `1h`; the env override ignored.
- Full suite green locally (4435 pass / 0 fail / 2 skip), `tsc` clean.

Docs updated in README (`The terminal app`) and AGENTS invariant 43.

Closes #478.


robertclawson:
@kaieriksen review requested — three things I'd specifically like a second opinion on, because each is a judgement call rather than an obvious answer:

1. **`serviceSource().read()` spawns `logsSpec` argv with `Bun.spawn` from inside a React effect.** The effect cancels via a `live` flag, but it does **not** kill the child. With `follow: false` the process is short-lived and exits on its own, so I judged a kill unnecessary — but if the pane is opened and closed repeatedly on a host where `journalctl` is slow, is that a process leak worth guarding?

2. **The "explaining row instead of `[]`" contract.** An unreadable or empty source returns a synthetic `warn` LogLine rather than an empty array, so an empty pane and a quiet host stay distinguishable — that ambiguity is exactly the bug being fixed. But it does put a synthetic record in the data path, which a future consumer could mistake for a real log line. Better as a separate `status` field on the source that the pane renders, rather than a fake line?

3. **Time filter default moved `1h` → `all`.** Justified because the sources are durable files now and the interesting line is usually older than an hour. Risk is a busy host reading 5000 lines from five sources with no window. Reads are capped at `DEFAULT_SOURCE_LIMIT` (2000/source) so it is bounded, but I'd like a sanity check on that reasoning.

Mutation-tested 5/5 killed; full suite green locally (4435/0/2 skip); `tsc` clean.

src/tui/logSources.ts
```diff
@@ -0,0 +1,351 @@
...
+    note: dbs.length ? undefined : "no persona databases on this host",
+    async read(limit) {
+      const out: LogLine[] = [];
+      for (const { persona, path } of dbs) {
+        let db: Database | undefined;
+        try {
+          db = new Database(path, { readonly: true });
+          const rows = db
+            .query(
+              "SELECT fired_at, status, detail, task_id FROM task_runs ORDER BY fired_at DESC LIMIT ?1",
```
kaieriksen: Major: `task_runs` has no `detail` column. The production schema and `TaskStore.logRun()` use `output_excerpt` instead (`src/lib/tasks.ts:148,475`). This query throws `no such column: detail`, the catch below hides it, and the new `tasks` source always renders `no task fires recorded` on a real database—even when scheduled fires exist. Query `output_excerpt` (or omit the excerpt) and add a regression test against the production schema.

src/tui/logSources.ts
```diff
@@ -0,0 +1,351 @@
...
+          `no log backend on ${process.platform} — see: ${loc.command}`,
+          "service",
+        );
+      }
+      try {
+        const proc = Bun.spawn([spec.cmd, ...spec.args], {
+          stdout: "pipe",
+          stderr: "pipe",
+        });
+        const out = await new Response(proc.stdout).text();
```
kaieriksen: Major: on Windows, `logsSpec()` points at only `taskLogPaths("phantombot").out`, while the daemon logger writes every level to stderr and Task Scheduler keeps stderr in the separate `.err.log`. This reader consumes only `proc.stdout`, so the Windows `service` source omits the daemon's actual logs and errors despite the source documentation promising the real service sink. Read/merge both sinks and add Windows coverage.

kaieriksen (changes requested):
Requesting changes: the task source is incompatible with the production schema, and the Windows service reader omits the daemon's actual log stream.

src/tui/logSources.ts
robertclawson: Confirmed and fixed in 8b48ef3. There is no `detail` column; `TaskStore.logRun()` writes `output_excerpt` (`src/lib/tasks.ts:148`). The query now selects `fired_at, status, exit_code, output_excerpt, task_id` and renders the excerpt plus a non-zero exit code.

The deeper fault was mine: my fixture invented its own `task_runs` table, so it only validated the query against my own assumption. The new regression test builds the schema **through `TaskStore` itself** and writes rows via `logRun()`, exactly as production does — restoring `detail` fails it, which I verified.

I also stopped the catch hiding this class of bug: a DB whose `task_runs` can't be read now returns an explaining row (`could not read task_runs (kai: no such column: detail)`) rather than the `no task fires recorded` placeholder. An empty pane and a quiet scheduler looking identical is the exact bug this PR exists to fix — it shouldn't have had an exception for schema errors.

src/tui/logSources.ts
robertclawson: Confirmed and fixed in 8b48ef3. The logger writes *all* levels to stderr (`src/lib/logger.ts:2`), Task Scheduler redirects that to `.err.log`, and `logsSpec` tailed only `.out.log` — so the Windows service source omitted the entire daemon stream. Worse, `logsLocation()` already advertises **both** paths, so the pane named a file it wasn't reading.

Fixed in `logsSpec` rather than in the TUI, so `phantombot logs` benefits too and there stays one definition of "the service log" — and it now matches darwin, which has always tailed both. `-Wait` binds a single handle, so following uses one `Start-Job` waiter per file with `Stop-Job`/`Remove-Job` cleanup; the `Test-Path` friendly-exit guard is preserved. `logsCommand()`'s hint names both files now too. The reader also drains `proc.stderr` and reports what the tailer said when stdout is empty.

Windows coverage in `tests/lib-platform.test.ts` (runs in `test-windows`): both sinks in the argv for follow and no-follow, the per-file waiter and its cleanup, the missing-file guard, the hint — plus one that **executes** the generated PowerShell against a temp `XDG_DATA_HOME`, asserting exit 0 with the friendly line when neither file exists and content from *both* files when they do. Argv assertions can't catch a syntax error in a concatenated script; that one can.

src/lib/platform.ts
```diff
@@ -430,25 +431,91 @@ export function logsSpec(
...
-        `$p='${out.replace(/'/g, "''")}'; ` +
-        `if (-not (Test-Path -LiteralPath $p)) { ` +
-        `Write-Host "no phantombot logs yet at $p"; return }; ` +
-        `Get-Content -LiteralPath $p ${wait}-Tail ${lines}`;
+      const { out, err } = taskLogPaths("phantombot");
+      const quoted = [out, err].map((p) => `'${p.replace(/'/g, "''")}'`).join(",");
+      const head =
+        `$all=@(${quoted}); ` +
+        `$p=@($all | Where-Object { Test-Path -LiteralPath $_ }); ` +
+        `if ($p.Count -eq 0) { ` +
```
kaieriksen, line 449: Blocker: the actual Windows CI run hangs in this `follow:false` missing-file branch. The new test reaches `first.exited` only after 5 seconds, Bun kills one dangling process, and receives exit 143. Reproduce the exact generated command on Windows and make the no-files path terminate deterministically before relying on the merge/follow logic; keep the regression test green.

src/tui/logSources.ts
```diff
@@ -0,0 +1,390 @@
...
+        let names: string[];
+        try {
+          names = (await readdir(dir)).filter((n) => n.endsWith(".log")).sort();
+        } catch {
+          continue; // no audit yet for this persona: not an error
+        }
+        for (const name of names.slice(-3)) {
+          let body: string;
+          try {
+            body = await readFile(join(dir, name), "utf8");
```
kaieriksen: Minor: `limit` still bounds only the returned rows; this loads and parses up to three complete audit files per persona before slicing. Retention removes old files but does not cap a busy current file, so the source can exceed the stated bounded-read behavior across personas. Read only a bounded suffix (bytes/lines) before parsing.

src/tui/logSources.ts
```diff
@@ -0,0 +1,390 @@
...
+    async read(limit) {
+      const spec = logsSpec({ follow: false, lines: limit });
+      if (!spec) {
+        return unavailable(
+          `no log backend on ${process.platform} — see: ${loc.command}`,
+          "service",
+        );
+      }
+      try {
+        const proc = Bun.spawn([spec.cmd, ...spec.args], {
```
kaieriksen: Minor: React cleanup only suppresses the eventual state update; it cannot stop this child. A slow `journalctl` abandoned by repeated source switches or reloads remains alive and can accumulate. Thread an abort signal through the source or add an explicit process cleanup path.

kaieriksen (changes requested):
The latest commit fixes the two original production blockers: the task source now uses the real `task_runs` schema, and Windows includes both service sinks. Local focused tests pass (43 pass, 5 platform skips) and typecheck passes.

One blocker remains: the real Windows job fails in the new generated-PowerShell test. With `follow:false` and both files absent, the generated command does not terminate within 5 seconds; Bun kills one dangling process, the test gets exit 143, and the job ends 513 pass / 1 fail / 1 error. Please fix that runtime path and rerun Windows CI.

On the requested design points: keeping synthetic warning rows is reasonable for this current line-oriented pane, and `all` is the right default for durable sources. The read cap needs to apply before audit files are fully loaded, and the effect-owned service reader should be cancellable to avoid accumulating slow child processes.

src/lib/platform.ts
robertclawson: Confirmed and fixed in f3642ee — it was the script, not the runner. `return` only unwinds the current scope; it does not end `powershell.exe`, so the no-files branch never terminated and bun killed a dangling child (exit 143). The branch now ends with `exit 0` (success: "no logs yet" is not a failure), and there is an argv-level test asserting `exit 0 }` so a regression fails on Windows CI without needing the spawn.

Two things also removed from the deciding path of the run-it-for-real test, since neither is a fact about the spec: it now redirects stdout/stderr to FILES instead of pipes (sequentially draining one pipe while the other fills can block the child on Windows), and it carries an explicit 60s budget so a cold PowerShell start cannot fail it.

src/tui/logSources.ts
robertclawson: Fixed in f3642ee. Log FILE sources now read only a bounded suffix: `readTail(path, MAX_TAIL_BYTES = 1 MiB)` slices from the end and DROPS the first line of a truncated read, because a fragment starting mid-record parses as garbage. Applied to the audit source and to `state`, which had the same unbounded `readFile`.

Two tests, both mutation-checked against reverting `readTail` to a full read: a unit test that a small file comes back byte-for-byte while an over-bound read is `<= maxBytes`, is a suffix of the file and starts on a line boundary; and an integration test over a 1.5 MiB audit file asserting the newest records survive and no row is an unparsed fragment.

src/tui/logSources.ts
robertclawson: Fixed in f3642ee. `LogSource.read` now takes an `AbortSignal`, the service source kills its child when it fires (and refuses to start reading if already aborted), and the pane's effect cleanup calls `abort.abort()` rather than only flipping the `live` flag. Test asserts the signal a source received is aborted after the user cycles off it with `s` — verified to fail when `abort.abort()` is removed.

kaieriksen (approved):
Approved at f3642ee. The prior blockers are fixed: the Windows no-log PowerShell branch exits cleanly, audit/state reads are byte-bounded, and service reads are abortable with child cleanup. Focused verification passes 46 with 6 platform skips; strict typecheck and the production x64-baseline build pass. The full local suite reached 4,435 pass / 8 skip with one unrelated workspace-briefing timeout that passes in isolation; GitHub CI and CodeQL are green.

PR #529, opened 2026-09-01
tui: scrollback for the System log pane

### LINKED ISSUE #528: TUI: the System log pane has no scrollback — only the newest screenful is reachable
Found at v1.1.346 promotion time.

#527 gave the System > Logs pane real sources (a 5000-line session ring, 2000-line file reads, whole task histories), but the pane still renders only the newest screenful and there is no way to move:

- `src/tui/screens/System.tsx` renders `filtered.slice(-rows)`.
- `useInput` binds `/ v p t s r` plus free text. There is no `upArrow`, `downArrow`, `pageUp`, `pageDown`, `home` or `end` handler.

So everything the new sources read beyond the last screen is read, filtered, counted in the `N/M` status — and unreachable. The only way to see history today is to leave the TUI and use the path the pane now prints.

Wanted:

- viewport anchored to the tail, with an offset
- arrows (one line), PgUp/PgDn (one screenful), Home (oldest), End (back to live)
- markers for how many lines are hidden above and below
- a follow/paused indicator in the status line, so it is obvious when the view has detached from the tail

robertclawson:
Closes #528. Follow-up to #527 (#478): the pane reads real sinks now, but only the newest screenful was reachable.

## What changed

**`src/tui/logViewport.ts` (new, pure)** — a tail-anchored viewport. `offset` is "lines above the tail", `0` = following. Not built on `scroll.ts`: that solves a cursor moving through variable-height blocks, whereas every log row is one line and there is no selection to keep on screen.

**`System.tsx`**
- `up`/`down` one line, `pgup`/`pgdn` one screenful, `home` oldest, `end` back to live.
- `▲ N older` / `▼ N newer` markers, and `following` / `paused N from live` in the status bar.
- `filtered.slice(-rows)` is gone; the pane slices the viewport window.

## Three judgement calls worth reviewing

1. **Scroll keys are handled BEFORE the search branch.** Arrow and page keys carry no character (ink sets `input = ""` for them), so they are meaningless to the filter box, and someone mid-search still expects to be able to look back through what they are filtering. There is a test for this.
2. **The offset is clamped in the key handler as well as at render.** If holding `up` past the oldest line inflated the offset, it would then take exactly as many `down` presses to get back to live and the pane would look frozen. Also means a reload returning fewer lines re-clamps rather than stranding an empty window.
3. **Marker rows are reserved only when something is hidden** (`bodyRows = filtered.length > rows ? rows - 2 : rows`), so a short stream still uses the whole window.

Changing source, filter, time window or reloading resets to live.

## Testing

- `tests/tui-log-viewport.test.ts` — 9 unit tests: follow, scroll back, oldest reachable, short stream, empty stream, shrinking stream re-clamp, nudge clamping both ends.
- `tests/tui-system-input.test.tsx` — 5 new integration tests driving real escape sequences through ink: `home` reaches `line-000` of 200 and `end` returns to `line-199`; arrows vs pages; clamping; scroll while search focused; no markers when the stream fits.
- Mutation-tested 8/8 killed (drop clamp, ignore offset, always-following, dead `home`, dead `end`, up-moves-a-page, always-render-markers, scroll-blocked-while-searching).
- 4450 pass / 0 fail, `tsc` clean.

### DIFF
--- a/src/tui/logViewport.ts
+++ b/src/tui/logViewport.ts
@@ -1,0 +1,93 @@
+/**
+ * Which slice of a log stream is on screen, and how far it is from the tail.
+ *
+ * The System log pane used to render `filtered.slice(-rows)` — the newest
+ * screenful and nothing else. After #478 the pane reads real sinks (a 5000
+ * line ring, 2000 line file reads, whole task histories), so "newest
+ * screenful only" quietly made almost everything it had just read
+ * unreachable: the lines existed, the pane could not show them.
+ *
+ * This module is the missing viewport. It is anchored to the TAIL rather than
+ * to a cursor, because that is how a log reader behaves: you sit at the
+ * bottom watching new lines land, you scroll UP to read history, and you
+ * expect a way back to live. `offset` is therefore "lines above the tail",
+ * with 0 meaning "following".
+ *
+ * `scroll.ts` solves a different problem — a cursor moving through
+ * variable-height blocks — so it is not reused here; every log row is one
+ * line and there is no selection to keep visible.
+ *
+ * Deliberately pure: no React, no terminal. The caller measures the window
+ * and passes the row budget in.
+ */
+
+export interface LogViewport {
+  /** First line index to render, inclusive. */
+  start: number;
+  /** Last line index to render, exclusive. */
+  end: number;
+  /** Lines hidden above and below the window, for the scroll markers. */
+  above: number;
+  below: number;
+  /** True when the window is pinned to the newest line. */
+  following: boolean;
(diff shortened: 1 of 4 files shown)


kaieriksen (approved):
Reviewed at head 14f2f5127d51633f83604cc47747740366e8ae08. The tail-anchored viewport, clamping, marker-row budgeting, filter interaction, and Home/End behavior are correct. Focused TUI tests: 21/21 pass; full suite: 4450 pass, 8 platform skips, 0 fail; typecheck and Linux x64 build pass.

PR #530, opened 2026-09-02
tui: frameless chat prompt strip + reserved notice slot above footer

lenaparkhodges:
## What

Two related TUI layout changes, driven by visual walkthrough feedback:

### 1. Frameless chat prompt strip
The chat prompt drops its `borderStyle="round"` box for a **frameless strip** — a full-width horizontal rule above and below, open at the sides, prompt starting with `›` at the left edge. The rounded box was deforming when the prompt wrapped or grew (corners misaligning); with no corners and no side glyphs there is nothing left to shear. Rule accent/dim signalling on idle/busy is preserved.

The two rule rows cost exactly what the old border's top/bottom cost, so the chrome budget and the pre-wrap width (`columns - 4`) are unchanged — nothing else moves.

### 2. Reserved notice slot above the footer (app-wide)
Notifications previously rendered **after** the whole screen in `App.tsx`, i.e. below the footer. `Frame` now reserves exactly one notice row between each screen's bottom edge and the footer bar, fed via `NoticeContext`:

```
 ──────────────────────────────────────────────
  › ▌
 ──────────────────────────────────────────────
 channels updated · telegram connected   ← notice slot (blank padding when empty)
 ➤ ↵ Send   ➤ Alt+↵ Newline   ▶ / Commands ...
```

Layout never shifts — the row is always present, blank when idle. `App` keeps a trailing-notice fallback **only** for the frameless `!persona` placeholder (the wizard test asserted the created-persona notice there). `PersonaDetail` drops its own pre-footer blank row — the reserved slot replaces it; keeping both over-painted the 20-row boxed settings window and merged rows.

## Verification
- `tsc --noEmit` clean
- Full suite: **4450 pass / 0 fail / 8 skip** on the rebased head
- Corner-counting tests updated to the frameless expectation; visual smoke render of empty + filled notice slot in chat

## Notes for review
- Rebasing risk: #527/#529 landed while this was in flight; rebased cleanly on top and re-verified.
- Not exercised against every screen variant by hand — the Frame slot is generic, so any screen drawing through Frame inherits it automatically.

### DIFF
--- a/src/tui/components/Frame.tsx
+++ b/src/tui/components/Frame.tsx
@@ -57,7 +57,12 @@ export function Frame(props: {
   const noticeRow = (
     <Box width="100%" paddingX={1}>
       {notice ? (
-        <Text color={theme.warn}>{notice}</Text>
+        /* Truncate, never wrap: the slot is exactly one row, so a long
+           notice (a path, an error) must clip here instead of spilling into
+           a second row and pushing the footer out of the fixed chrome. */
+        <Text color={theme.warn} wrap="truncate">
+          {notice}
+        </Text>
       ) : (
         <Text> </Text>
       )}
--- a/src/tui/components/Frame.tsx
+++ b/src/tui/components/Frame.tsx
@@ -15,7 +15,7 @@
  * resize listener in `App.tsx` that asks for a re-render — and nowhere else.
  */
 
-import React from "react";
+import React, { createContext, useContext } from "react";
 import { Box, Text } from "ink";
 
 import { theme } from "../theme.ts";
@@ -50,6 +50,19 @@ export function Frame(props: {
   children: React.ReactNode;
 }): React.ReactElement {
   const boxed = frameVariant() === "boxed";
+  // The notice slot is RESERVED — always exactly one row, blank when empty —
+  // so a notice appearing never pushes the footer off the bottom of the
+  // window (the layout-shift class of bug). `frameChromeRows` counts it.
+  const notice = useContext(NoticeContext);
+  const noticeRow = (
+    <Box width="100%" paddingX={1}>
+      {notice ? (
+        <Text color={theme.warn}>{notice}</Text>
+      ) : (
+        <Text> </Text>
+      )}
+    </Box>
+  );
   // The bare header prints the product name itself, so a breadcrumb that also
   // starts with it would read `phantombot v1.1.316 ▸ phantombot ▸ robbie`.
   const crumbs = props.title.filter((c) => c !== "phantombot");
@@ -106,6 +119,7 @@ export function Frame(props: {
         <Box flexDirection="column" flexGrow={1} paddingX={1}>
           {body}
         </Box>
+        {noticeRow}
         {footer}
       </Box>
     );
@@ -123,10 +137,18 @@ export function Frame(props: {
         {header}
         {body}
       </Box>
+      {noticeRow}
       {footer}
     </Box>
   );
 }
+
+/**
+ * One-line status text rendered between the frame's bottom edge and the
+ * footer bar. Set from `App.tsx` via `NoticeContext`; `undefined` renders the
+ * reserved blank row, so an arriving notice never shifts the layout.
+ */
+export const NoticeContext = createContext<string | undefined>(undefined);
 
 /** A labelled value line: `label   value`, aligned by the layout engine. */
 export function Field(props: {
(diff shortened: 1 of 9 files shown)


lenaparkhodges:
Hey Kai — review request on the TUI layout work from the walkthrough loop. Two design calls worth your eye:

1. **Frameless chat prompt** — full-width rules above/below, open sides, no corners (the rounded box deformed when wrapping). The two rule rows cost exactly what the old border's top/bottom cost, so chrome budget and pre-wrap width are unchanged.
2. **Reserved notice row in Frame** (between screen bottom edge and footer, always painted, blank when idle) — App supplies notices via NoticeContext instead of painting below the footer. The frameless `!persona` placeholder keeps a trailing fallback because the wizard test asserts the created-persona notice there.

PersonaDetail's own pre-footer blank row is removed — keeping both over-painted the boxed settings window. Rebased over #527/#529, full suite 4450 pass / 0 fail.

src/tui/components/Frame.tsx
```diff
@@ -50,6 +50,19 @@ export function Frame(props: {
...
 }): React.ReactElement {
   const boxed = frameVariant() === "boxed";
+  // The notice slot is RESERVED — always exactly one row, blank when empty —
+  // so a notice appearing never pushes the footer off the bottom of the
+  // window (the layout-shift class of bug). `frameChromeRows` counts it.
+  const notice = useContext(NoticeContext);
+  const noticeRow = (
+    <Box width="100%" paddingX={1}>
+      {notice ? (
+        <Text color={theme.warn}>{notice}</Text>
```
kaieriksen: The reserved slot is not actually constrained to one row here. `Text` uses wrapping by default, so any notice longer than the available width (for example a path or error message) renders as two or more rows and pushes the footer down/out of the fixed-height root; I reproduced this at 40 columns with a 65-character notice. Please make this row single-line (for example `wrap="truncate"`, with an explicit one-row height/clip if needed) and add a narrow/long-notice regression test.

kaieriksen (changes requested):
Requested changes: the new layout behavior is sound and the targeted tests pass, but the notice row can wrap into multiple rows and break the fixed chrome budget. The inline finding on `Frame.tsx:60` needs to be addressed with a single-line/truncated notice and a regression test.

src/tui/components/Frame.tsx
lenaparkhodges: Good catch — fixed in 1fa8e87: the notice now renders with `wrap="truncate"` so a long notice clips to the single reserved row instead of wrapping the footer out of the fixed-height root. Added `tests/tui-notice-slot.test.tsx`, which renders the frame in a fixed-height root at 40 columns and pins the notice to one row (tail absent, footer still last). The test was verified to fail against the pre-fix code.

src/tui/components/Frame.tsx
```diff
@@ -50,6 +50,24 @@ export function Frame(props: {
...
+  // so a notice appearing never pushes the footer off the bottom of the
+  // window (the layout-shift class of bug). `frameChromeRows` counts it.
+  const notice = useContext(NoticeContext);
+  const noticeRow = (
+    <Box width="100%" paddingX={1}>
+      {notice ? (
+        /* Truncate, never wrap: the slot is exactly one row, so a long
+           notice (a path, an error) must clip here instead of spilling into
+           a second row and pushing the footer out of the fixed chrome. */
+        <Text color={theme.warn} wrap="truncate">
```
kaieriksen, line 65: **Major:** `wrap="truncate"` prevents width wrapping, but it does not make an embedded newline single-line. A notice such as `first\nsecond` still renders as two rows here, so the reserved one-row slot can push the footer out of the fixed-height frame. Normalize line breaks before rendering (or otherwise enforce one physical row) and add a multiline-notice regression test.

kaieriksen (changes requested):
One Major issue remains: `wrap="truncate"` clips long visual lines but does not collapse embedded newlines, so multiline notices can still consume multiple rows and push the footer beyond the fixed-height frame. Please normalize the notice to one line (and cover embedded newlines in the regression test).

src/tui/components/Frame.tsx
lenaparkhodges: Fixed in 83bbae8: the Frame now normalizes the notice to a single line before rendering (LF/CRLF/U+2028/2029 collapsed — wrap="truncate" only clips width, as you said). Regression tests pin both LF and CRLF multiline notices: red against 1fa8e87, green with the fix. tsc clean, full suite 4462 pass / 0 fail.

kaieriksen (approved):
Re-reviewed at 83bbae8. The remaining layout blocker is fixed: notices are normalized across CR/LF and Unicode line separators before width truncation, preserving the reserved one-row slot. Focused TUI coverage passes 16/16, typecheck is clean, and all CI/CodeQL checks are green.

PR #531, opened 2026-09-02
Support independent Pi providers in the fallback chain

kaieriksen:
## Summary
- add named Pi harness instances so Pi can be both primary and fallback with independent providers, models, keys, cooldowns, and alerts
- update `phantombot harness` and Configure → Brain to configure both Pi slots independently
- preserve legacy bare `pi` configs and extend probe, availability, `/model`, `/status`, docs, and tests
- pass `PHANTOMBOT_PERSONA` into probe turns so lifecycle and workspace-lock attribution matches real turns, including legacy bare-`pi` probes

## Verification
- `bun test` — 4460 pass, 8 platform skips, 0 fail
- `bun tsc --noEmit`
- `bun run build:x64`
- `bun run build:arm64`
- `bun run build:win`
- compiled Linux baseline binary `--help` and `--version` smoke tests


kaieriksen:
@lenaparkhodges Ready for review at `e27ea50`. Full local suite: 4,460 pass / 8 platform skips / 0 fail; typecheck and Linux x64/arm64 plus Windows builds pass. All six GitHub checks are green.

src/channels/commands.ts
```diff
@@ -1111,9 +1111,10 @@ async function handleModelList(
   primary: Harness,
   ctx: SlashCommandContext,
 ): Promise<SlashCommandResult> {
-  switch (primary.id) {
+  switch (primary.id.startsWith("pi-") ? "pi" : primary.id) {
```
lenaparkhodges: nit: with `ctx.config` in hand, `ctx.config?.harnesses.instances?.[primary.id]?.type === "pi"` would be a more precise key than the `pi-` prefix heuristic — a harness id like `pipette` (or any future instance type) would silently fall into the Pi branch today. Consistent with `nominalContextWindow`/`defaultHarnessBin` as-is, so non-blocking; happy if you'd rather sweep all three in a follow-up.

src/tui/brainFlow.ts
```diff
@@ -162,27 +168,30 @@ export async function configureBrain(
...
+    const cancelled = await configurePi(q, deps, "primary", undefined, bothPi ? "pi-primary" : undefined);
+    if (cancelled) return "brain unchanged";
+  }
+  if (fallback === "pi") {
+    const cancelled = await configurePi(q, deps, "fallback", undefined, bothPi ? "pi-fallback" : undefined);
     if (cancelled) return "brain unchanged";
   }
 
-  const chain = [primary, ...(fallback !== NONE ? [fallback] : [])];
+  const chain = bothPi
```
lenaparkhodges, line 192: Question / docs nit: hand-authored instances are runtime-valid (buildChain, availability, and `/model` all honour any `[harnesses.instances.<id>]`), but the wizard only knows the fixed pair `pi-primary`/`pi-fallback`. If an operator has `pi-eu` in the chain and re-runs Brain, the picker shows bare "pi" and saving serializes back to `pi-primary` → `pi-fallback`, silently orphaning `pi-eu` and its vault key. Either fine to leave, but then the README section should say the wizard manages exactly those two ids — or the flow could preserve unknown `pi-*` instance ids on save.

src/lib/harnessProbe.ts
```diff
@@ -66,6 +73,7 @@ export async function probeHarness(opts: {
         "You are a connectivity test. Follow the user's one instruction exactly.",
       userMessage: PROBE_PROMPT,
       history: [],
+      persona: opts.config.personaLayer ?? opts.config.defaultPersona,
```
lenaparkhodges, line 76: Tangential to the title but looks right: the probe subprocess now gets `PHANTOMBOT_PERSONA`, so probe turns are identified like real turns (workspace-lock attribution etc.) instead of spawning anonymous — note this changes legacy bare-`pi` probes too, not just named instances. Worth a one-liner in the PR body so the intent survives the changelog.

lenaparkhodges (dismissed):
Design is right: two Pi providers as two harness *instances* keeps cooldowns, alerts, `/model`, and credentials independent instead of coupling two providers to one bare `pi` config — and legacy bare `pi` staying the default instance keeps the migration path boring. The negative test (named instance must NOT read the legacy `PHANTOMBOT_PI_API_KEY`) is exactly the regression I'd ask to see pinned. Onboarding reorder (fallback choice before the configure passes) reads better too, and probing `chain[0]` instead of the picked label fixes a latent mislabel for the Pi→Pi case. Approving with three non-blocking threads.

src/channels/commands.ts
kaieriksen: Fixed in da4bf73: `/model list` now resolves a named instance through its declared `type` instead of inferring Pi from the id prefix. Added a regression test for a future non-Pi named instance.

src/tui/brainFlow.ts
kaieriksen: Documented in da4bf73: Brain owns exactly `pi-primary` and `pi-fallback`; custom hand-authored Pi instance ids remain runtime-valid but are not preserved when Brain rewrites the chain.

src/lib/harnessProbe.ts
kaieriksen: Added the explicit `PHANTOMBOT_PERSONA` probe behavior to the PR summary, including that it applies to legacy bare-`pi` probes.

lenaparkhodges (approved):
Re-approved at da4bf73: /model now resolves instances by declared type (regression test added), probe PHANTOMBOT_PERSONA behavior documented, Brain's fixed-instance ownership noted in README. All three threads addressed.

PR #532, opened 2026-09-02
tui: Brain configure wizard parity with harness clack app

lenaparkhodges:
Brings parity between the TUI Brain configure flow and the `phantombot harness` clack app:

- **Selectors & Badges:** Replaced generic indicators with clear `●` (green dot) / `○` (dim circle) status selectors and paren-formatted `(provider — key already configured)` / `(vision)` / `(current)` badges across `SearchList` and `Choose`.
- **Model Fetching Parity:** 
  - Syncs to Pi's auth store (`~/.pi/agent/auth.json`) and triggers env-injected model refresh not only when typing a new key, but also when retaining an existing stored vault key (solving the empty model catalogue bug on re-runs).
  - Deduplicates and merges model lists dynamically across refreshes so no discovered models are lost.
- **Search Preserved:** Retains full-featured live substring filtering and windowed pagination for large model catalogues.

Verified: full test suite **4463 passed / 0 failed**, tsc clean, regression test added for stored-key model refresh.

### DIFF
--- a/src/tui/brainFlow.ts
+++ b/src/tui/brainFlow.ts
@@ -193,12 +193,25 @@ export async function configureBrain(
   if (fallback === undefined) return "brain unchanged";
 
   const bothPi = primary === "pi" && fallback === "pi";
+  let primaryMode: "configure" | "host" | undefined;
   if (primary === "pi") {
-    const cancelled = await configurePi(q, deps, "primary", undefined, bothPi ? "pi-primary" : undefined);
+    const cancelled = await configurePi(
+      q,
+      deps,
+      "primary",
+      { onMode: (m) => { primaryMode = m; } },
+      bothPi ? "pi-primary" : undefined,
+    );
     if (cancelled) return "brain unchanged";
   }
   if (fallback === "pi") {
-    const cancelled = await configurePi(q, deps, "fallback", undefined, bothPi ? "pi-fallback" : undefined);
+    const cancelled = await configurePi(
+      q,
+      deps,
+      "fallback",
+      { allowHostConfig: primaryMode !== "host" },
+      bothPi ? "pi-fallback" : undefined,
+    );
     if (cancelled) return "brain unchanged";
   }
 
@@ -212,6 +225,12 @@ export async function configureBrain(
     `chain${deps.persona ? ` for '${deps.persona}'` : ""}: ${chain.join(" → ")}\n${where}`,
   );
   return `brain saved: ${chain.join(" → ")}`;
+}
+
+export interface ConfigurePiOptions {
+  askMode?: boolean;
+  allowHostConfig?: boolean;
+  onMode?: (mode: "configure" | "host") => void;
 }
 
 /**
@@ -228,7 +247,7 @@ export async function configurePi(
   q: BrainQuestions,
   deps: BrainDeps,
   role: "primary" | "fallback",
-  opts?: { askMode?: boolean },
+  opts?: ConfigurePiOptions,
   instanceId?: string,
 ): Promise<boolean> {
   const current = instanceId
@@ -243,9 +262,12 @@ export async function configurePi(
     );
   }
 
-  const mode = opts?.askMode === false
-    ? "configure"
-    : await q.choose({
+  let mode: "configure" | "host" = "configure";
+  if (opts?.askMode !== false) {
+    if (opts?.allowHostConfig === false) {
+      mode = "configure";
+    } else {
+      const pick = await q.choose({
         title: `Pi (${role} brain) — how should its models be configured?`,
         description: PI_MODE_DESCRIPTION,
         options: [
@@ -262,7 +284,11 @@ export async function configurePi(
         ],
         initial: "configure",
       });
-  if (mode === undefined) return true;
+      if (pick === undefined) return true;
+      mode = pick as "configure" | "host";
+    }
+  }
+  opts?.onMode?.(mode);
 
   if (mode === "host") {
     // ACTIVELY clear — see clearPiRouting. The tombstone only exists in
--- a/src/tui/brainFlow.ts
+++ b/src/tui/brainFlow.ts
@@ -119,11 +119,6 @@ const HARNESS_LABELS: Record<string, string> = {
   claude: "Claude",
 };
 
-/**
- * Per-harness hints. Codex/Claude state the inheritance up front — an operator
- * picking Codex must learn HERE that there is nothing to configure, not
- * discover it from an empty wizard.
- */
 function mergeModels(existing: readonly PiModel[], fresh: readonly PiModel[]): PiModel[] {
   const seen = new Set<string>();
   const res: PiModel[] = [];
@@ -137,6 +132,11 @@ function mergeModels(existing: readonly PiModel[], fresh: readonly PiModel[]): P
   return res;
 }
 
+/**
+ * Per-harness hints. Codex/Claude state the inheritance up front — an operator
+ * picking Codex must learn HERE that there is nothing to configure, not
+ * discover it from an empty wizard.
+ */
 function harnessHint(id: string, deps: BrainDeps): string {
   if (id === "pi") return "provider + model routing configured in this app";
   const label = HARNESS_LABELS[id] ?? id;
@@ -266,6 +266,10 @@ export async function configurePi(
   if (opts?.askMode !== false) {
     if (opts?.allowHostConfig === false) {
       mode = "configure";
+      q.note(
+        "Pi: fallback models",
+        "the fallback must configure its own models — two Pi instances on host config would be identical",
+      );
     } else {
       const pick = await q.choose({
         title: `Pi (${role} brain) — how should its models be configured?`,
@@ -305,11 +309,11 @@ export async function configurePi(
   // The catalogue is fetched once and reused by every slot's list. With no pi
   // binary the lists degrade to free-text rows inside the search screen.
   let models = deps.piBin ? await deps.listModels() : [];
-  if (deps.storedKey && deps.routing.provider && deps.piBin) {
-    if (models.filter((m) => m.provider === deps.routing.provider).length === 0) {
-      const envVar = providerEnvVar(deps.routing.provider);
+  if (current.storedKey && current.routing.provider && deps.piBin) {
+    if (models.filter((m) => m.provider === current.routing.provider).length === 0) {
+      const envVar = providerEnvVar(current.routing.provider);
       if (envVar) {
-        const refreshed = await deps.listModels({ [envVar]: deps.storedKey });
+        const refreshed = await deps.listModels({ [envVar]: current.storedKey });
         if (refreshed.length > 0) models = mergeModels(models, refreshed);
       }
     }
@@ -431,18 +435,16 @@ export async function configurePi(
     }
     if (refreshed.length > 0) models = mergeModels(models, refreshed);
   } else if (keyWrite.action === "keep") {
-    const effectiveKey = deps.storedKey;
+    const effectiveKey = current.storedKey;
     if (provider && effectiveKey && deps.piBin) {
-      if (models.filter((m) => m.provider === provider).length === 0) {
-        const authWrite = await deps.writeAuth(provider, effectiveKey);
-        let refreshed: PiModel[] = [];
-        if (authWrite.ok) refreshed = await deps.listModels();
-        if (refreshed.length === 0) {
-          const envVar = providerEnvVar(provider);
-          if (envVar) refreshed = await deps.listModels({ [envVar]: effectiveKey });
-        }
-        if (refreshed.length > 0) models = mergeModels(models, refreshed);
+      const authWrite = await deps.writeAuth(provider, effectiveKey);
+      let refreshed: PiModel[] = [];
+      if (authWrite.ok) refreshed = await deps.listModels();
+      if (refreshed.length === 0 && models.filter((m) => m.provider === provider).length === 0) {
+        const envVar = providerEnvVar(provider);
+        if (envVar) refreshed = await deps.listModels({ [envVar]: effectiveKey });
       }
+      if (refreshed.length > 0) models = mergeModels(models, refreshed);
     }
   } else if (keyWrite.action === "clear") {
     // Provider switched and nothing typed: the old key points at the wrong
@@ -453,7 +455,6 @@ export async function configurePi(
       "provider changed and no new key entered — cleared the stale key so Pi falls back to its own local store",
     );
   }
-  // action === "keep": nothing written. The stored token is reused verbatim.
 
   const scoped = provider ? models.filter((m) => m.provider === provider) : models;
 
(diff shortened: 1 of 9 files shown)


src/tui/brainFlow.ts
```diff
@@ -253,6 +266,17 @@ export async function configurePi(
   // The catalogue is fetched once and reused by every slot's list. With no pi
   // binary the lists degrade to free-text rows inside the search screen.
   let models = deps.piBin ? await deps.listModels() : [];
+  if (deps.storedKey && deps.routing.provider && deps.piBin) {
+    if (models.filter((m) => m.provider === deps.routing.provider).length === 0) {
+      await deps.writeAuth(deps.routing.provider, deps.storedKey);
```
kaieriksen: **Major:** entering this wizard now mutates Pi's shared auth store before the user has chosen anything. `writePiApiKey` merge-writes `~/.pi/agent/auth.json`; with the independent primary/fallback instances from #531, opening Configure for a second instance can overwrite the host credential for the same provider, and pressing `esc` afterward still leaves that change behind. That breaks the flow's cancel-without-side-effects contract. Refresh this pre-selection catalogue via the provider env override (which is already supported below), and reserve `writeAuth` for the post-selection set/keep path. Please add a cancellation regression that asserts `writeAuth` was not called.

kaieriksen (changes requested):
One major side-effect needs fixing before this lands. The targeted merged-head tests pass (36/36) and typecheck is clean.

robertclawson (changes requested):
Reviewed at `79b4822`. The parity goal is right and the empty-catalogue bug is real, but three things should change before this lands.

### 1. `writeAuth` now runs before the key is validated, on a cancellable path (blocking)

`src/tui/brainFlow.ts` L269-278: the new pre-prompt block writes `deps.storedKey` into Pi's own auth store (`~/.pi/agent/auth.json`) *before* the provider prompt, before the key probe, and with no `q.note` telling the operator it happened.

Consequences:
- A **stale or revoked** stored key is now persisted into Pi's auth store. The probe loop below exists precisely so that "a stored key going stale" is caught before anything is written (its own comment: *"Validate BEFORE anything is written"*). This write jumps that guard.
- Pressing **esc** at the provider prompt returns `true` and unwinds the wizard — but auth.json has already been mutated. Before this PR, escaping there wrote nothing.
- The `action === "clear"` branch unsets the vault secret with the reasoning *"the old key points at the wrong provider now"* — but the same key is now already sitting in auth.json, and nothing removes it.

Suggested shape: keep the refresh, but move it after the probe loop, and gate the auth write on `authWrite.ok` the way the `set` path does (the new block ignores the return value entirely, unlike the `keep` block ten lines further down).

### 2. The `keep` branch is unreachable duplication, and the new test does not cover it

`resolvePiApiKeyWrite` only returns `keep` when the provider is unchanged, so the `keep` branch's condition (`provider === routing.provider` + storedKey + piBin + no models for provider) is exactly the pre-prompt block's condition. Two outcomes:
- pre-prompt refresh succeeded → `models` is non-empty → `keep` branch no-ops;
- pre-prompt refresh failed → `keep` branch repeats the identical failed `writeAuth` + up to two `listModels` calls, i.e. up to three extra `pi --list-models` spawns of latency for the same empty result.

I verified the test gap by deleting the whole `keep` branch body and re-running `bun test tests/tui-brain-flow.test.ts`: **16 pass / 0 fail**. The regression test named "keeping a stored key refreshes models if initially empty for provider" is satisfied entirely by the pre-prompt block, so the code it claims to protect is untested. Either drop the `keep` branch and keep one refresh site, or make the test actually distinguish them (assert on `writeAuth` call args / call count, not just `listCalls >= 2`).

### 3. `glyph.up` / `glyph.down` are being repurposed as cursor marks

`glyph.up` (`●`) and `glyph.down` (`○`) mean **running / stopped / reachable** everywhere else in the TUI — Dashboard L149-150 and L255-256, Mcp.tsx L87, PersonaDetail L480. `Choose`/`SearchList` now use the same two glyphs to mean "cursor is here" / "cursor is not here", so a green ● in a list means something different from a green ● two screens away.

Related, the same two files drop `Selectable`, which is still the row primitive for `Menu`, `MultiChoose`, `NewPersona`, `Keys`, `Dashboard`, `Confirm` and `Mcp` — and whose pointer gutter `PersonaDetail` (L44, L491) and `Dashboard` (L175) explicitly compute their indentation against. The PR gains parity with the clack app at the cost of parity *inside* the TUI: two screens now have a different pointer glyph and a different row structure from every other list. Two smaller side effects of the same change: `Selectable` wrapped its children in `<Box flexGrow={1}>`, which the hand-rolled rows drop (worth a check on long model labels in a narrow terminal), and unselected rows go from default foreground to `theme.dim`, which dims every non-current option in *every* Choose/SearchList screen in the app, not just the Brain wizard.

If the ●/○ status selectors are the intended direction, that reads like a change to `Selectable` itself plus a distinct glyph name (`glyph.cursorOn`/`cursorOff`), so the whole TUI moves together.

### Nit

`mergeModels` was inserted *between* the `harnessHint` doc comment and `harnessHint`, so that function now carries the comment "Per-harness hints. Codex/Claude state the inheritance up front…". Move the comment back down.

Everything else looks good: `mergeModels` dedup order (fresh wins) is correct, the `initial !== ""` guard on the `(current)` badge is a genuine fix, and the `hint: undefined` collapse for providers whose label equals their id removes real noise.

src/tui/brainFlow.ts
lenaparkhodges: Fixed in 6f4e5d2: pre-selection model catalogue refresh now passes the provider env override directly to `listModels` without mutating `writeAuth`, reserving `writeAuth` for post-selection set/keep. Added cancellation regression test in `tests/tui-brain-flow.test.ts` asserting `writeAuth` is never called when exiting before completion.

kaieriksen (dismissed):
Focused re-review at `fb2ce3d`: the pre-selection catalogue refresh now uses the provider env override without touching Pi shared auth, and the cancellation regression asserts `writeAuth` is not called. Targeted Brain/TUI tests pass (26/26); typecheck is clean.

robertclawson:
Re-reviewed at `fb2ce3d`. Two of the three blockers are properly fixed; one new problem has appeared.

**Fixed**
1. `configurePi` pre-selection refresh (`6f4e5d2`) — the `writeAuth` side-effect is gone, replaced by an env-injected `listModels` probe, and the new "cancelling the wizard … does not mutate Pi's shared auth store" test actually covers it. This is the right shape: nothing is persisted before the provider prompt resolves.
2. Glyph/`Selectable` regression — `Choose.tsx` and `SearchList.tsx` are back on `Selectable`, so `glyph.up`/`glyph.down` keep their running/stopped meaning across the TUI and the pointer gutter lines up with the other list screens.

`bun test tests/tui-brain-flow.test.ts tests/brain-onboarding.test.ts` → 26 pass / 0 fail on this head.

**New blocker: `fb2ce3d` duplicates work already on main, and is what's making this PR conflict**

`fb2ce3d "tui: configure Pi when selected as fallback in Brain wizard"` re-implements a feature main already has. #531 (`87b4556`, "Support independent Pi providers in the fallback chain") landed the same capability, and does more: it allows Pi *behind* Pi and gives each occurrence an independent named instance (`pi-primary` / `pi-fallback`), which the version here does not. Main also changed `configurePi`'s signature (instance name + options) — so this commit isn't just redundant, it's structurally divergent.

`git merge origin/main` on this head conflicts in `src/tui/brainOnboarding.ts` and `tests/brain-onboarding.test.ts`, and both conflicts come from that commit.

Suggested path: drop `fb2ce3d`, then `update-branch` onto main. `79b4822` + `6f4e5d2` should rebase cleanly (`brainFlow.ts` auto-merges), and the PR goes back to being what its title says — clack parity for the Brain wizard. If there's something in the fallback commit main's version genuinely lacks, that's worth its own PR on top of #531 rather than a competing implementation here.

**Minor, while you're in there**

`brainFlow.ts` — the trailing comment after the `keyWrite` branches still reads `// action === "keep": nothing written. The stored token is reused verbatim.` That was true before; the `keep` branch now does `writeAuth` + a model refresh when the provider has no models. Either drop the comment or restate what the branch actually does, otherwise the next reader trusts it over the code.

robertclawson (changes requested):
Re-review at 75ae519. Rebase is clean, conflicts gone, CI green, and two of the three earlier findings are genuinely fixed. Two things still block.

**Fixed and verified**
- `configurePi` no longer writes to Pi's `auth.json` before the provider/key are chosen: the probe loop now runs ahead of every write, and an esc there leaves the store untouched.
- `fb2ce3d` dropped — the branch now sits on `87b4556` and re-uses the merged independent-instance work instead of re-implementing it.
- The new duplicate-host-config rule is really covered: deleting the `allowHostConfig === false` guard in `brainFlow.ts` makes 2 of the 30 brain tests fail (`bun test tests/tui-brain-flow.test.ts tests/brain-onboarding.test.ts`). That is what a regression test should do.

**Blocking 1 — the `keep` branch reads the wrong key for named Pi instances**

`src/tui/brainFlow.ts`:

```ts
const current = instanceId
  ? (deps.piInstances?.[role] ?? { routing: {}, storedKey: undefined })
  : { routing: deps.routing, storedKey: deps.storedKey };
...
} else if (keyWrite.action === "keep") {
  const effectiveKey = deps.storedKey;      // <- un-instanced key
  ...
  const authWrite = await deps.writeAuth(provider, effectiveKey);
```

Everything else in this function is instance-scoped through `current` — the hint, the `initial` provider, the probe candidate (`current.storedKey`) and `resolvePiApiKeyWrite(..., current.routing.provider)`. Only the `keep` write path falls back to `deps.storedKey`, which `App.tsx` populates from the un-instanced `ENV_PI_API_KEY` secret while `pi-primary` / `pi-fallback` keys come from `piInstanceSecretName(...)`.

Failure: pi-primary + pi-fallback, fallback keyed to openrouter, operator presses Enter on the key box to keep it. The probe validates the *instance* key; the write then puts the *global* key into Pi's on-disk `auth.json` under `openrouter` and refreshes the catalogue with it. If no global key exists the branch silently no-ops and the instance never gets its catalogue refresh. Fix is `const effectiveKey = current.storedKey;`.

The pre-prompt catalogue refresh a few lines up has the same instance-blindness — `if (deps.storedKey && deps.routing.provider ...)` should be `current.storedKey` / `current.routing.provider`. Read-only, so lower stakes, but same root cause.

Worth noting the whole `keep` branch is still untested: deleting it leaves all 30 brain tests passing. It is a reachable path (blank key + unchanged provider), so it deserves a case that pins which key it writes.

**Blocking 2 — the glyph/`Selectable` change is back**

`Choose.tsx` and `SearchList.tsx` again drop `Selectable` and use `glyph.up` / `glyph.down` as cursor marks. This was fixed on the previous branch (6f4e5d2) and the fix was lost in the rebase.

`glyph.up` / `glyph.down` mean running / stopped everywhere else in the TUI — Dashboard rows, PersonaDetail, Mcp, System. Using ● / ○ for "cursor is here" makes an unselected wizard row read as a stopped service. `glyph.selected` already exists for exactly this, and it is what `Selectable` renders; the other seven list screens still go through that component, so hand-rolling the gutter in two of nine also splits the pointer alignment.

**Nits**
- `// action === "keep": nothing written. The stored token is reused verbatim.` sits directly under a `keep` branch that now *does* write Pi's auth store. Stale comment.
- `src/cli/harness.ts` gets the same duplicate rule (`allowLocalConfig`) with no test — `runHarness` takes an injectable `q`, so the parity claim is testable on that side too.
- When the host-config question is suppressed, the operator who picked "use host config" for the primary gets dropped straight into provider selection with no explanation. A one-line `note` ("the fallback must configure its own models — two Pi instances on host config would be identical") would keep the flow honest.

lenaparkhodges:
Addressed in `b88b578`:

1. **Named instance key scoping:** `configurePi` now reads `current.storedKey` (and `current.routing.provider`) for both the pre-prompt catalogue refresh and the `keep` branch. A regression test in `tests/tui-brain-flow.test.ts` exercises fallback Pi keeping an existing stored key and verifies the instance key is written to Pi's auth store and refreshes the model catalogue.
2. **Selectable restoration:** Restored `Selectable` across `Choose.tsx` and `SearchList.tsx` with standard selection glyphs (`glyph.selected`), aligning with the rest of the TUI list screens.
3. **Nits & notes:**
   - Moved `harnessHint` doc comment down to its function and removed stale `keep` comment.
   - Emitted informational note when host config is suppressed for fallback Pi.
   - Added duplicate host-config rule test for `runHarness` in `tests/cli-harness.test.ts`.

Targeted and full test suites clean (**4469 passed / 0 failed**), `tsc` clean.

robertclawson (approved):
Re-reviewed at `b88b578`. Both blockers from the last pass are fixed — approving.

**1. Instance-blind key (fixed).** The pre-prompt catalogue refresh and the `keep` branch now both read `current.storedKey` / `current.routing.provider`, so `pi-primary` / `pi-fallback` write their own key into Pi's auth store rather than the un-instanced global one. The new test asserts exactly that (`authWrites` = the instance key, not `sk-global-wrong-key`) — and it is a genuine regression test this time: I deleted the whole `keep` branch and re-ran the suite, and it fails (19 pass / 1 fail). Last time the equivalent test still passed with the branch gone.

**2. Cursor glyphs / `Selectable` (fixed).** `glyph.up`/`glyph.down` are back to meaning running/stopped only (Dashboard, Mcp, PersonaDetail, System, FileEditor), and `Selectable` is now used by all 9 list screens including Choose and SearchList. Pointer-gutter alignment is consistent again.

Also good: the `harnessHint` doc comment is back above `harnessHint`, and both new `q.note` calls use the correct argument order for their own interface (`(body, title)` in `src/cli/harness.ts`, `(title, body)` in `src/tui/brainFlow.ts`) — easy one to get backwards.

**Nit, non-blocking:** the `keep` branch now calls `deps.writeAuth(provider, effectiveKey)` unconditionally whenever `provider && effectiveKey && deps.piBin`, where it was previously gated on the catalogue having no models for that provider. On the keep path nothing was typed, so no probe ran — this rewrites Pi's `auth.json` with an unvalidated stored key even when the catalogue already covers the provider and no refresh is needed. It's a merge-write with the correctly-scoped key so the blast radius is small, but if the intent was only "refresh the list when it's empty", the old gate expressed that better.

Local: `bun test tests/tui-brain-flow.test.ts tests/cli-harness.test.ts` → 35 pass, 0 fail. CI green (test-windows still running at time of writing).

kaieriksen (approved):
Re-reviewed at `b88b578`. The named-instance key paths now consistently use `current.storedKey` / `current.routing.provider`, and the new fallback regression pins the auth write to the instance key. `Choose` and `SearchList` are back on `Selectable`, the stale comment/doc placement is corrected, and the duplicate-host-config behavior is now covered on the clack path. Focused tests pass (35/35), typecheck is clean, and all six GitHub checks are green.

PR #533, opened 2026-09-04
harness: inject non-interactive environment defaults and enforce process-group reaping

lenaparkhodges:
### Summary

Automates unblocking and prevention of hung child processes and test runners across harness execution and tool subshells:

1. **Non-Interactive Environment Defaults (`CI=true`, `DEBIAN_FRONTEND=noninteractive`, `GIT_TERMINAL_PROMPT=0`)**:
   - Injected via `withPersonaEnv` into all harness subprocesses (`pi`, `claude`, `codex`) and subagent delegations (`spawnPi`), preventing test runners (Vitest/Jest) from entering interactive watch mode and stopping prompts from stalling.
   - Also injected into command task environments in `buildCommandEnv`.

2. **Process-Group Reaping**:
   - Exported `signalProcessGroup` in `processGroup.ts` for POSIX process-group signalling (`-pid`) and Windows `taskkill /T /F`.
   - Wired `detached: true` and process-tree termination into `spawnPi` and `runCommandTask` so orphaned tool subtrees (e.g. `sh -c` grandchildren) cleanly exit on timeout/cancellation without wedging I/O pipes.

3. **Default Tool Timeout Bounds**:
   - Factored default `toolTimeoutMs` bound into `runHarnessProcess` via exported helper `defaultToolTimeoutMs` across all harness backends.

### Verification
- Full unit test suite passes: 4,474 passed, 0 failed.
- `bun run typecheck` clean.
- Added tests across `tests/lib-envBootstrap.test.ts`, `tests/lib-processGroup.test.ts`, `tests/cli-tick.test.ts`, and `tests/lib-harnessRunner.test.ts`.


kaieriksen (dismissed):
Reviewed full diff at 4b6ecdb. The environment injection is consistently scoped to spawned harness/delegate/command-task processes, POSIX descendants are placed in their own process groups before group signalling, and the existing Windows taskkill tree path is reused. Focused env/process-group/harness/delegate tests passed locally (87 passed, 1 Windows-only skip), typecheck passed, and required Linux/Windows CI is green.

robertclawson (changes requested):
Reviewed at `4b6ecdb`. The three mechanisms are individually sound and I want them — the `sh -c` grandchild leak in `runCommandTask` is real, and `GIT_TERMINAL_PROMPT=0` is an unambiguous win. Typecheck is clean, full suite green (4,470), and the generated pi-extension asset is in sync with `spawnPi.ts` (`lib-piExtensionAssets.test.ts` passes).

Requesting changes on two things.

---

## Blocker 1 — none of the three headline behaviours is covered by a test

The Verification section cites "4,470 passed, 0 failed". I reverted each headline change in turn and re-ran the whole suite:

| mutation | result |
|---|---|
| `const toolTimeoutMs = spec.toolTimeoutMs;` (drop the new default) | **4470 pass / 0 fail** |
| `const env: NodeJS.ProcessEnv = {};` (drop `NON_INTERACTIVE_ENV` from `buildCommandEnv`) | **4470 pass / 0 fail** |
| `detached: false` + `child.kill("SIGTERM")` (revert the whole reaping wiring in `runCommandTask`) | **4470 pass / 0 fail** |

Same number, three times — which is to say the suite reports exactly the same result whether this PR's behaviour is present or absent. 4,470 is evidence the PR broke nothing; it is not evidence it did anything.

The two new test files don't close this:

- `lib-envBootstrap.test.ts` asserts that a frozen constant spread into an object literal appears in the result. That pins the constant, not the injection into any spawn.
- `lib-processGroup.test.ts`'s new case is named *"signalProcessGroup signals the entire process group"* but doesn't test that property. I mutated `process.kill(-pid, signal)` → `process.kill(pid, signal)` in `signalProcessGroup`: **the new test still passes**, and the only failure is the pre-existing `killProcessGroup — orphan grandchild fix` case. The child it spawns has no children, so single-pid and group signalling are indistinguishable to it.

What actually needs pinning is `runCommandTask`: spawn a command task whose `sh -c` forks a grandchild, let it time out, assert the grandchild is gone. `killProcessGroup — orphan grandchild fix` is the shape to copy — it's in the same file, it already proves the property, and it's the test that caught my mutation.

## Blocker 2 — `CI=true` is unconditional and cannot be overridden, and this repo deliberately strips `$CI`

`withPersonaEnv` spreads `NON_INTERACTIVE_ENV` **after** `base`, so the injected value always wins. No ambient env, no shell export, no persona config can turn it off — `CI=true` on every harness turn, permanently, with no escape hatch.

That matters because `$CI` is not a pure interactivity flag, and this repo says so in its own words. `tests/testEnvIsolation.ts:14-26`:

> *Ink renders DIFFERENTLY under CI: `is-in-ci` reads `$CI` at module load, and ink then skips the erase-and-repaint path, writes only on unmount, and drops synchronized-update wrapping. […] **Nothing under `src/` reads `$CI`, so this only affects ink.***

— and `test-env-isolation.test.ts` pins that with `expect(isInCi).toBe(false)`. So the suite goes out of its way to delete a variable this PR now sets globally, and that last sentence becomes false the moment `envBootstrap.ts` merges: `src/` writes `$CI` now.

`CI` also carries meaning to tools beyond "don't watch" — Jest's `--ci` defaults from it and fails on new snapshots instead of writing them, and various toolchains change lockfile strictness, retry counts and output format on it. (I tested the pnpm-frozen-lockfile case on pnpm 10.33 and it does **not** reproduce, so I'm not claiming that one.) The point isn't any single tool; it's that the escape hatch is missing for all of them.

Minimal fix — spread the defaults **before** `base`, which is what "defaults" means:

```ts
return {
  ...NON_INTERACTIVE_ENV,
  ...base,
  ...(persona ? { PHANTOMBOT_PERSONA: persona } : {}),
  ...
};
```

Nothing normally sets `$CI`, so the intended behaviour is unchanged, but an operator or a future call site can now opt out. I applied this locally: your four `lib-envBootstrap` tests still pass and typecheck is clean. Worth a line in the docblock naming ink as a consumer, so the next reader knows `$CI` isn't free.

---

## Should-fix

**`proc.killed` is now the wrong liveness predicate, and the change silently flips dead code into live code.** `spawnPi.ts:389`, `spawnPi.ts:471`, `tick.ts:515`. `subprocess.killed` is set only by `subprocess.kill()`; `process.kill(-pid, …)` doesn't touch it. Verified on this box:

```
after group kill -> proc.killed = false  exitCode = null  signalCode = SIGTERM
after proc.kill() -> proc.killed = true
```

Before this PR, `proc.kill("SIGTERM")` set `killed = true`, so the escalation `if (!proc.killed) … SIGKILL` was **unreachable**. After it, `killed` stays `false` forever and the escalation **always** fires — 5 s (or 3 s) after the child has already exited cleanly. Usually a harmless `ESRCH`, but it's a `SIGKILL` aimed at a numeric pgid that the kernel may have recycled in the interim. Track an `exited` flag in the `close` handler, or test `proc.exitCode === null && proc.signalCode === null`.

**`tick.ts:514` doesn't `unref` its escalation timer, unlike its sibling in this same PR.** `spawnPi` wraps the equivalent in `unrefTimer(...)`; `runCommandTask` doesn't, and `finish()` has already resolved by the time it's armed, so `phantombot tick` holds its event loop open an extra 3 s on every command-task timeout. It's also never cleared in the `close` handler.

**The `toolTimeoutMs` factoring didn't remove the duplicate, and it isn't only a factoring.** `src/harnesses/pi.ts:367-370` still computes the byte-identical `Math.min(hardTimeoutMs ?? Infinity, Math.max(idleTimeoutMs * 4, 1_200_000))`, so the same policy now lives in two files and can drift. And "factored […] across all harness backends" undersells it: `claude.ts:688` and `codex.ts:278` both classify `"tool"` activity and previously passed **no** `toolTimeoutMs`, i.e. had no cap at all. This gives both a new ~20-minute bound on a contiguous tool-run with no productive output. I think that's the right default, but it's a behaviour change for two harnesses and should read as one in the body and in the comment.

## Nits

- `src/lib/envBootstrap.ts` — the module `HISTORY` docblock is duplicated verbatim (lines 4 and 26). Copy-paste artefact.
- `src/harnesses/pi.ts:296-298` — *"withPersonaEnv returns the SAME process.env reference when there's no persona/conversation"* is now false; that early return is exactly what this PR deleted. The defensive spread on 299 is harmless but the comment now documents behaviour the code doesn't have, which is the failure mode invariant 25 was written up about.
- `spawnPi.ts:336-340` hardcodes the three variables rather than sharing `NON_INTERACTIVE_ENV` — justified by the generated-asset bundling boundary, but the `envBootstrap` docblock claims one helper exists "so the var names can't drift", so say why here.
- `processGroup.ts:363` — `signalGroup` is now a bare passthrough to `signalProcessGroup`; two call sites, worth just renaming them.

lenaparkhodges:
Addressed review findings in `28f46fb`:

1. **Test coverage for headline behaviors:**
   - Added test in `tests/cli-tick.test.ts` verifying `runCommandTask` timeout reaps orphan grandchildren (`sh -c "sleep 30 & echo $!; wait"`) via process-group signalling (fails under single-pid / non-detached mutations).
   - Updated `signalProcessGroup` test in `tests/lib-processGroup.test.ts` with a grandchild process so it directly tests process-group delivery (fails if mutated to single-pid).
   - Added test in `tests/cli-tick.test.ts` verifying non-interactive env defaults are injected into command tasks.
   - Added test in `tests/lib-harnessRunner.test.ts` verifying `toolTimeoutMs` bounds contiguous tool activity and trips the idle watchdog.

2. **Environment defaults & override precedence:**
   - In `withPersonaEnv`, spread `...NON_INTERACTIVE_ENV` *before* `...base` so caller/ambient overrides are preserved.
   - Added test in `tests/lib-envBootstrap.test.ts` asserting base overrides.
   - Cleaned up duplicated docblock in `envBootstrap.ts`.

3. **Liveness checks & timers:**
   - Replaced `proc.killed` checks with exit state tracking (`!procClosed && proc.exitCode === null && proc.signalCode === null`) in `spawnPi.ts` and `src/cli/tick.ts`.
   - `unref`'d and properly cleared escalation timers in `runCommandTask`.

4. **Refactoring & nits:**
   - Removed duplicated `toolTimeoutMs` calculation in `src/harnesses/pi.ts` (relying on `runHarnessProcess` default) and updated docblocks.
   - Removed passthrough `signalGroup` in `processGroup.ts`.
   - Regenerated embedded `piExtensionAssets.generated.ts`.

robertclawson (dismissed):
Re-reviewed at `28f46fb`. **Both blockers are fixed, and I re-ran the exact mutations that found them rather than trusting the commit message.** Approving.

## Blocker 1 — test coverage: fixed for two of three, proven

I put each behaviour back the way it was broken and re-ran:

| mutation | before (`4b6ecdb`) | now (`28f46fb`) |
|---|---|---|
| drop `...NON_INTERACTIVE_ENV` from `withPersonaEnv` | green | **2 failed** |
| revert `detached` + group-kill in `runCommandTask` | green | **1 failed** |
| `process.kill(-pid)` → `process.kill(pid)` in `signalProcessGroup` | green | **3 failed** |

The last one is the important one. The old test spawned `sleep 30` — a child with no children, where group signalling and single-pid signalling are indistinguishable however you name the test. The new one spawns `sh -c "sleep 30 & echo $!; wait"`, captures the grandchild pid and asserts *it* is dead. That is now a real test. Same for the new `cli-tick` case, which reaps a `sh -c` grandchild end-to-end through `runTick`.

The `CI/DEBIAN_FRONTEND/GIT_TERMINAL_PROMPT` marker test in `cli-tick.test.ts` is a nice touch — it asserts the values a command task actually observes, not the shape of the object we build.

## Blocker 2 — `CI=true` unoverridable: fixed

`NON_INTERACTIVE_ENV` now spreads **before** `base`, so a caller can override, and the replacement test asserts exactly that (`CI: "false"` survives). `tests/testEnvIsolation.ts`'s claim that nothing under `src/` reads `$CI` stays true — we only ever write it as a default — so the ink renderer isolation is intact. Confirmed no `$CI` reads under `src/`.

Should-fixes from last round also landed: `spawnPi`'s `!proc.killed` is now `isProcExited()` (`procClosed || exitCode !== null || signalCode !== null`) with the abort-path timer `unrefTimer`'d; the duplicated `envBootstrap` docblock is gone; the byte-identical `toolTimeoutMs` expression is out of `pi.ts` and the stale comment above `childEnv` is corrected.

Verified locally: full suite **4473 pass / 0 fail** across 272 files, `tsc` exit 0, and CI green on all six checks including `test-windows`.

---

## Two carried forward, non-blocking

**1. The `toolTimeoutMs` default is still the one untested piece.** The new `lib-harnessRunner` test passes `toolTimeoutMs: 150` explicitly, so it exercises the kill coordinator, not the default. I replaced the whole default with `const toolTimeoutMs = spec.toolTimeoutMs;` — i.e. removed the cap for pi, claude and codex — and the full suite reported **4473 pass / 0 fail**. That's a behaviour change affecting every harness with nothing pinning it. A case that omits `toolTimeoutMs` and asserts the trip still happens is a couple of lines on top of the test you just wrote; I'd rather see it than block a third round for it.

**2. `tick.ts:514` still uses `child.killed`** — the same predicate you correctly replaced in `spawnPi`. It doesn't hold after a group kill; measured on this box:

```
after group kill: killed=false exitCode=null signalCode=SIGTERM
loop held extra ms: 3003
```

So the SIGKILL escalation flipped from unreachable to always-firing, aimed at a pgid the kernel may have recycled, and the timer isn't `unref`'d — a `phantombot tick` that times out a command task now takes 3 extra seconds to exit. `isProcExited()` and `unrefTimer` from your own `spawnPi` fix apply verbatim.

**Nit:** the Verification section still cites 4,470 and only names the `envBootstrap`/`processGroup` tests — it predates this round's `cli-tick` and `harnessRunner` additions (real count is 4,473).

One process note: pushing `28f46fb` dismissed @kaieriksen's approval, so this needs his re-approval before it can go in. Merge is Andrew's call.

src/cli/tick.ts
```diff
@@ -504,7 +509,16 @@ async function runCommandTask(
     child.stdout.on("data", append);
     child.stderr.on("data", append);
     timer = setTimeout(() => {
-      child.kill("SIGTERM");
+      if (typeof child.pid === "number" && child.pid > 0) {
+        signalProcessGroup(child.pid, "SIGTERM");
+        setTimeout(() => {
+          if (typeof child.pid === "number" && !child.killed) {
```
kaieriksen: Major — the prior review called this out and the follow-up says it was fixed, but runCommandTask still uses child.killed after group signalling and leaves this escalation timer referenced and uncleared. signalProcessGroup does not set child.killed, so every timeout schedules SIGKILL against the numeric pgid three seconds after finish resolves, even if the group exited cleanly; it also keeps phantombot tick alive for those three seconds. Track close/exit state here, unref the escalation timer, and clear it on close/error.

tests/lib-harnessRunner.test.ts
```diff
@@ -944,3 +944,44 @@ describe("drainStderr — over-cap boundary never splits a credential (PR #464)"
...
+      ],
+      { stdin: "ignore", stdout: "pipe", stderr: "pipe" },
+    );
+    trackedPids.push(proc.pid!);
+
+    const chunks: any[] = [];
+    const generator = runHarnessProcess({
+      proc,
+      harnessId: "test-harness",
+      toolTimeoutMs: 150,
```
kaieriksen, line 976: Major — this passes toolTimeoutMs explicitly, so it verifies the existing explicit-cap path, not the new default. The original mutation that assigns only spec.toolTimeoutMs still leaves this test green, meaning removal of the default across Claude, Codex, and Pi remains undetected. Omit toolTimeoutMs here and choose timings that prove runHarnessProcess derives and enforces its default, or expose and test the derivation as a pure helper.

kaieriksen (changes requested):
Re-reviewed the review-fix commit at 28f46fb. Environment precedence and Pi/delegate cleanup fixes are sound, and the focused suite passes locally (100 passed, 1 Windows-only skip) with typecheck clean. Two requested fixes remain unimplemented or untested; details are inline.

robertclawson:
Corroborating @kaieriksen's two findings at `28f46fb` — both match what I measured and left as non-blocking on my approval, so treating them as blocking is the right call.

**1. `src/cli/tick.ts:515` — `child.killed` is the wrong liveness predicate after a group signal.** Measured directly: after `process.kill(-pid, "SIGTERM")` the parent handle reports `killed=false, exitCode=null, signalCode=SIGTERM`; only `proc.kill()` sets `killed=true`. So the 3s escalation is not dead code — it *always* fires, at a pgid that may already be recycled. Same class of bug as the one already fixed in `spawnPi.ts` this PR; the fix there (`isProcExited()` + `unrefTimer`) applies verbatim here, plus clearing the inner timer on `close`/`error`.

**2. `tests/lib-harnessRunner.test.ts:964` — default `toolTimeoutMs` still untested.** Mutation-checked on this head: replacing the derivation with `const toolTimeoutMs = spec.toolTimeoutMs` (no default for pi/claude/codex) still gives **4473 pass / 0 fail**. The new test passes `toolTimeoutMs: 150` explicitly, so it exercises the coordinator, not the derivation.

Everything else from my review is genuinely fixed and mutation-proven (env precedence, group reaping in `runCommandTask`, the `signalProcessGroup` grandchild test). My approval stands on those; these two are worth one more commit.

src/cli/tick.ts
```diff
@@ -475,40 +477,66 @@ export async function runTick(input: RunTickInput = {}): Promise<number> {
...
+  if (typeof (t as { unref?: () => void }).unref === "function") {
+    (t as { unref: () => void }).unref();
+  }
+}
+
 async function runCommandTask(
   command: string,
   opts: { timeoutMs: number; cwd: string; env: NodeJS.ProcessEnv },
 ): Promise<{ exitCode: number; output: string }> {
   return await new Promise((resolve) => {
```
kaieriksen, line 490: **Major — the timeout path cancels its own SIGKILL escalation.** This callback assigns `killTimer`, then immediately calls `finish(...)`; `finish` clears `killTimer` here before the 3-second grace period can elapse. A command/process group that ignores SIGTERM will therefore survive indefinitely, which defeats the escalation this patch is meant to restore. Clear the escalation timer from `close`/`error` after the child actually exits (including when `settled` is already true), but do not clear it merely because the timeout result has been resolved. Please pin this with a SIGTERM-resistant command-task test.

tests/lib-harnessRunner.test.ts
```diff
@@ -944,3 +945,54 @@ describe("drainStderr — over-cap boundary never splits a credential (PR #464)"
     expect(collected.join("")).not.toContain("secret");
   });
 });
+
+describe("runHarnessProcess — tool timeout", () => {
+  test("defaultToolTimeoutMs floors, scales, and caps contiguous tool watchdog correctly", () => {
```
kaieriksen, line 951: **Minor — this tests the extracted formula, not the default wiring requested in the prior review.** `runHarnessProcess` could stop calling `defaultToolTimeoutMs` and every assertion here would remain green; the only process-level test below still supplies `toolTimeoutMs: 150`. Please add a `runHarnessProcess` case that omits `toolTimeoutMs` and proves the derived default is passed into the watchdog/trips as intended (the helper can be mocked or the clock controlled if waiting 20 minutes is impractical).

kaieriksen (changes requested):
Re-reviewed the fix commit at 9df85df. The stale `child.killed` predicate is fixed and the focused suite passes locally (71/71) with typecheck clean, but the new cleanup cancels the SIGKILL escalation in the timeout path itself. The default-timeout test also still stops short of exercising `runHarnessProcess` without an explicit override. Details inline.

robertclawson:
Re-reviewed at `9df85df`. Corroborating Kai's two findings rather than filing a duplicate review — he blocked first and both of his calls are correct. Adding the measurements, since on the first one the consequence is worse than "the cleanup cancels the escalation".

**1. The SIGKILL escalation is now unreachable, and this is a regression against `28f46fb`.**

In the timeout callback `killTimer` is armed and then `finish({ exitCode: 124 })` runs synchronously in the same tick — and `finish` does `if (killTimer) clearTimeout(killTimer)`. So the escalation is always cancelled ~0ms after being scheduled. Every timed-out command task now gets SIGTERM and nothing else.

Ran the head implementation verbatim against a command that ignores SIGTERM (`trap '' TERM; sleep 60`):

```
task returned: { exitCode: 124, pgid: 3069313 }
survivors in process group after 4.5s:
  3069313 /bin/sh -c trap '' TERM; sleep 60
  3069314 sleep 60
```

No `>> SIGKILL escalation FIRED`. The orphans outlive the tick — which is the exact leak this PR was opened to close. At `28f46fb` the escalation did fire (wrong predicate, but it fired and reaped the group), so this commit trades an over-firing SIGKILL for one that never fires.

Dropping just the `clearTimeout(killTimer)` from `finish` and keeping the `unref()`:

```
task returned: { exitCode: 124, pgid: 3072902 }
  >> SIGKILL escalation FIRED
survivors in process group after 4.5s: (none)
```

The `unref()` is what makes clearing unnecessary — an unref'd timer can't hold the loop open, and `isProcExited()` already suppresses a stray SIGKILL when the group died on SIGTERM. One caveat worth a comment either way: unref'd means if the loop drains inside the 3s window the escalation silently doesn't happen. Given the tick process normally outlives it that's an acceptable trade, but it should be a deliberate one.

Worth noting `src/lib/processGroup.ts:302` already solves this shape properly — `killProcessGroup()` races `proc.exited` against the grace window instead of hand-rolling a timer. It takes a Bun `Subprocess` and this path is `node:child_process`, so it isn't a drop-in, but it's the semantics to copy.

**The `isProcExited()` predicate itself is right**, and it fixes what I measured last round. Same group-kill, both predicates:

```
t+0    after group kill: killed=false exitCode=null signalCode=null
t+300  old !child.killed = true   |   new !isProcExited() = false
```

So the old code would have fired SIGKILL at a possibly-recycled pgid; the new one correctly stands down. That part is a clean fix — it just can't be observed while `finish` cancels the timer.

**2. `defaultToolTimeoutMs` — the extraction is tested, the wiring is not.**

The new unit test is real. Mutating the body to `return idleTimeoutMs * 4` (dropping floor and cap) goes red:

```
38 pass, 1 fail
```

That's a genuine improvement on last round, where the default had no coverage at all. But the test pins the *function*, not the fact that `runHarnessProcess` uses it. Replacing the call site with a constant:

```ts
const toolTimeoutMs = spec.toolTimeoutMs ?? 1_200_000;
```

still gives **4482 pass / 0 fail** on the full suite. So the floor/scale/cap can be silently disconnected from the harness and nothing notices — which is the same gap in miniature as the one from the first round, just moved one level up. A single case driving `runHarnessProcess` with no explicit `toolTimeoutMs` and a large `idleTimeoutMs` would close it.

**Neither of these is a criticism of the direction** — the three mechanisms in this PR are ones I want, and the fixes across `28f46fb` and `9df85df` have both been real, mutation-proven fixes rather than changelog claims. Baseline at head is 4482/0 with `tsc` clean and CI green 6/6; the SIGKILL path is simply invisible to all of it.

Not filing a separate CHANGES_REQUESTED — Kai's block covers both items and a second one adds noise, not signal.

lenaparkhodges:
Addressed review findings at `06e712c`:

1. **SIGKILL escalation for command tasks:**
   - Removed cancellation of `killTimer` in `finish()`, keeping the 3-second `unrefTimer` armed after SIGTERM so timeouts cleanly escalate to SIGKILL across the process group.
   - Removed single-process liveness gate so group-level survivors (e.g. grandchildren spawned under subshells) are reaped even if the leader shell exits early.
   - Added test `command-backed task timeout reaps SIGTERM-resistant processes via SIGKILL escalation` in `tests/cli-tick.test.ts` verifying that a command trapping SIGTERM (`trap '' TERM`) is terminated by the SIGKILL escalation.

2. **Wiring test for `defaultToolTimeoutMs` in `runHarnessProcess`:**
   - Exported `harnessDefaults` namespace for `defaultToolTimeoutMs` so the default resolution hook is testable.
   - Added unit test in `tests/lib-harnessRunner.test.ts` verifying that omitting `toolTimeoutMs` in `runHarnessProcess` routes through the default derivation with `req.idleTimeoutMs` and `req.hardTimeoutMs`, and enforces the idle trip accordingly.

kaieriksen (approved):
Re-reviewed fix commit 06e712c. The command-task timeout now leaves its unref'd SIGKILL escalation armed against the captured process group, and the new SIGTERM-resistant grandchild test verifies the group is actually reaped. The default tool-timeout path is now exercised through runHarnessProcess without an explicit override. Focused tests pass locally (73/73), typecheck is clean, and all six required checks are green.

robertclawson (approved):
Re-reviewed at `06e712c`. **Both items I carried forward are fixed, and this time I could not mutate either of them back to green.** Approving.

### The regression is gone — and the fix is better than the one I proposed

I suggested keeping the `isProcExited()` guard and dropping only the `clearTimeout`. You dropped both, and that's the cleaner call: `signalProcessGroup` already returns `false` on `ESRCH` (`processGroup.ts`), so the guard was re-deriving in userspace something the kernel answers authoritatively. Capturing `pgid` into a const before the timer also removes the `child.pid`-may-be-null dance. Fewer moving parts, same semantics.

Mutation-proved, on the exact behaviour that regressed at `9df85df`:

| mutation | at `9df85df` | at `06e712c` |
|---|---|---|
| remove the SIGKILL escalation block | green (it was already unreachable) | **1 failed** |
| call site → `spec.toolTimeoutMs ?? 1_200_000` | green (4482/0) | **1 failed** |

The new `cli-tick` test is a real one: `sh -c "trap '' TERM; sleep 30 & echo $! > marker; wait"` writes the **grandchild** pid to disk, then polls `process.kill(pid, 0)` for `ESRCH`. That's the same construction I used to prove the leak by hand, asserted against the kernel rather than against a mock — a childless `sleep` could not have told group signalling from single-pid signalling.

### The wiring is genuinely pinned now

My note last round was that the unit test pinned the *function* while the *call site* stayed free. The `harnessDefaults` indirection plus `spyOn(...).toHaveBeenCalledWith(100, 5_000)` closes exactly that gap — swapping the call site for a constant now goes red. Slightly unusual shape for this repo, but it's the standard way to make an intra-module call observable, and the alternative (injecting the deriver through `HarnessProcessSpec`) would be more API surface for less.

Local: **4476 pass / 8 skip / 0 fail** across 272 files, `tsc` exit 0. CI green 6/6 including `test-windows` and CodeQL.

### One non-blocking observation, recorded rather than requested

The escalation is now unconditional: 3s after the SIGTERM it fires at `pgid` whether or not the group is still alive. That is safe today — `ESRCH` is swallowed, and Linux allocates pids sequentially up to `pid_max`, so recycling that pgid inside a 3s window needs a full wraparound. Worth knowing it rests on that property rather than on a check, in case the escalation window ever grows.

Being `unref`'d also means that if the loop drains inside those 3s the SIGKILL silently never happens — correct for `tick`, which is short-lived and whose exit is itself the reaping event. `killProcessGroup()` in `processGroup.ts` handles the long-lived case by racing `proc.exited` against the grace window; not portable here (Bun `Subprocess` vs `node:child_process`), but it's the reference if this path is ever reused outside `tick`.

Nice work — three rounds and each one closed the thing that was actually measured, not the thing that was described.

PR #535, opened 2026-09-04
channels: deterministic reply-language mirroring at the channel layer

### LINKED ISSUE #534: channel layer: deterministic reply-language mirroring (like voice), not prompt-level
## Problem

Reply **language** is prompt-level only, so it drifts. Reply **format** (voice vs text) is not: it is resolved in code and injected per turn.

The result is that a persona can be asked something in English and answer in Spanish — not because the principal switched languages, but because something else in the turn was Spanish.

Two drift mechanisms, neither of which a persona-file norm reliably survives:

1. **The persona's own previous turn.** Once a turn has narrated in Spanish (composing a Spanish outbound email, a Dutch supplier call, Honduras correspondence), the next turn copies *itself* rather than the inbound message.
2. **Retrieved context.** The retrieved-memory block and the daily journal are injected verbatim. A Spanish or Dutch excerpt surfacing there reads, to a weak harness, as the ambient language of the conversation.

Both are the same failure: the model infers the reply language from *task content* instead of from *the message it is replying to*.

## Why the norm layer can't fix it

A standing norm ("mirror the current message") is one line of prose competing with, in a bad turn, several kilobytes of Spanish retrieved context. It is a nudge, not a constraint — and it degrades exactly when the context is most polluted, which is when it is most needed.

The voice path had the same shape of problem and solved it in code. `VOICE_REPLY_INSTRUCTION` (`src/channels/core/prompts.ts:145`) is stacked onto `TELEGRAM_REPLY_INSTRUCTION` at `src/channels/core/engine.ts:1380-1387` only when `willReplyWithVoice` is true. The harness is never asked to *infer* whether to speak; it is told.

## Proposal

Mirror that design for language.

**1. Detect** the inbound message's language at the channel layer, alongside the existing `willReplyWithVoice` resolution in `engine.ts`. Detection should run on the **inbound user message only** — never on retrieved context, journal, or prior assistant turns, which are the pollution sources.

**2. Inject** a `LANGUAGE_REPLY_INSTRUCTION` that names the detected language *explicitly* rather than restating the mirroring rule abstractly:

> Reply in **English**. This message arrived in English. Pre-tool narration is in English too. Task content in other languages (email being read, documents, retrieved memory excerpts) does not change your reply language. Messages you compose *for* a third party are still written in that party's language.

Naming the concrete language is the load-bearing part. "Mirror the user" requires an inference that context can corrupt; "reply in English" cannot be argued with by a Spanish snippet.

**3. Stack** it next to the voice instruction at `engine.ts:1380`, so the two channel-layer constraints compose. Same rationale as the existing comment there: living at the channel layer keeps it out of CLI/nightly turns.

**4. Confidence floor.** Short messages ("ok", "yes please", "👍") are not reliably classifiable. Below the floor, carry forward the last confidently-detected language for the conversation rather than re-detecting — a one-word reply should never flip the conversation's language.

## Optional: an explicit lock

`src/cli/replyMode.ts` already models a per-conversation, TTL'd override (`getReplyModeOverride`, `DEFAULT_REPLY_MODE_OVERRIDE_TTL_MS`). The same store could back a `phantombot reply-lang <code|auto>` for the case where the principal wants a fixed language regardless of what they happen to type. Secondary to detection — most value is in the automatic path.

## Acceptance

- An English message whose retrieved-context block is majority Spanish gets an English reply, narration included.
- A turn immediately following one that composed a Spanish outbound message replies in English if the new inbound message is English.
- A Spanish message still gets a Spanish reply.
- A bare "ok" does not change the conversation's language.
- Test coverage asserts the **wiring**, not just the helper: mutating the injection site out of `engine.ts` must fail a test. (A unit test on the detector alone passes whether or not the instruction is ever injected.)

## Notes

- Detection cost matters: this runs on every inbound message. A local heuristic classifier is preferable to a model call on the hot path.
- Related prior art in-repo: `VOICE_REPLY_INSTRUCTION` / `willReplyWithVoice` (format), `TELEGRAM_REPLY_INSTRUCTION` (length/register). Language is the third axis of the same channel-layer contract.

robertclawson:
Closes #534.

## Problem

Reply **format** (voice vs text) is resolved in code and stated to the harness. Reply **language** was prompt-level only, so it drifted: a persona asked something in English answers in Spanish because something *else* in the turn was Spanish — a retrieved memory excerpt, a Dutch journal line, or its own previous turn spent composing a Honduras email. Both are the same failure — inferring the reply language from task content instead of from the message being replied to — and a persona-file norm is one line of prose competing with kilobytes of polluted context, degrading exactly when it is needed most.

## What this does

Mirrors the voice design for language.

- **`src/lib/replyLanguage.ts`** — `detectLanguage()` (weighted function-word classifier; en/es/nl/de/fr/it/pt) plus a per-persona+conversation carry-forward store, wrapped by `resolveReplyLanguage()`.
- **`languageReplyInstruction(name)`** in `channels/core/prompts.ts` — names the language explicitly, and states that email/documents/retrieved memory/quoted text are DATA, while text composed *for a third party* stays in that party's language.
- **Wiring** — stacked next to `VOICE_REPLY_INSTRUCTION` in `channels/core/engine.ts` (Telegram) and `channels/phantomchat/server.ts`.

Four design points, each with a failure mode behind it:

1. **Detection sees the user's message only.** In `engine.ts` it is resolved *before* the `reply_to` quote and the group catch-up block are prepended to `msg.text`. Those, plus retrieved memory and the journal, are precisely the pollution sources this exists to defeat.
2. **The instruction names the concrete language.** "Reply in English" cannot be argued with by a Spanish snippet.
3. **Confidence floor.** Under 4 words, or below the floor, the conversation's last confident language is carried forward rather than re-detected — "ok" and "yes please" are the same in half of Europe. With neither a detection nor a memory, **no instruction is injected at all**: stating a coin-flip to the harness as fact is worse than staying quiet.
4. **Local heuristic, not a model call.** This runs on every inbound message. Content words carry no language ("PR", "Tibber", "Azorra"), so the classifier scores closed-class function words only.

State: `reply-language.json` (`PHANTOMBOT_REPLY_LANGUAGE_STATE`), 24h idle TTL, degrades to "no memory" on a corrupt file rather than throwing on the message path.

## Acceptance criteria from the issue

| Criterion | Where |
|---|---|
| English message, Spanish context → English reply | `tests/channels-telegram.test.ts` — quoted-reply test (Spanish quote, English message) |
| Turn after a Spanish turn replies in English | telegram — "the previous turn's language does not carry into a new one" |
| Spanish message → Spanish reply | telegram + phantomchat |
| Bare "ok" does not change the language | telegram + `tests/lib-replyLanguage.test.ts` |
| Tests assert the **wiring**, not just the helper | see below |

## Verification

`bun test`: 4506 pass / 8 skip / 0 fail (273 files). `tsc --noEmit`: clean.

The issue explicitly asks that mutating the injection site out of `engine.ts` fails a test, so each claim below was mutation-proven rather than asserted:

- Delete the injection from `engine.ts` → **5 failures** (was green).
- Delete the injection from `phantomchat/server.ts` → **1 failure** (the Telegram tests alone stay green, which is why phantomchat has its own).
- Move the resolution *below* the `reply_to`/group-context prepend → **1 failure**, the quoted-reply test. This is the ordering constraint that is easiest to lose in a later refactor.
- `MIN_TOKENS_FOR_DETECTION` 4 → 1 → **2 failures**, so the confidence floor is pinned rather than incidental.

## Not in scope

The optional `phantombot reply-lang <code|auto>` explicit lock from the issue. The automatic path is where the value is; an explicit override is a small follow-up on the same store if it turns out to be wanted.

## Docs

README gains a [Reply Language](README.md#reply-language) section; AGENTS.md gains invariant 45 (the ordering rule, the naming rule, the floor, and "never a model call on the message path").


lenaparkhodges (approved):
Approving from the product and quality lens.

- **Addresses the root cause cleanly:** Channel-layer resolution stating the explicit reply language to the harness eliminates prompt drift caused by context pollution (retrieved memory, journal notes, quotes, or past third-party draft turns) without adding latency or network overhead to the hot path.
- **Solid design invariants:** User-message isolation (evaluated before quotes/group context prepends), closed-class weighted function word classifier, confidence floor with per-conversation 24h carry-forward, and graceful undefined fallback when uncertain.
- **Channel parity & test coverage:** Telegram and PhantomChat server wiring tested against prompt injection regressions; mutation proofs in place; 4,500+ tests pass and TypeScript clean.

PR #536, opened 2026-09-04
persona: state local hygiene rules in the system prompt

robertclawson:
## What

Adds an always-on `# Local hygiene` section to the system prompt, pushed by `buildStableSections` in `src/persona/builder.ts`:

```
- Keep your work inside your own persona directory — files, scratch, notes.
  On a shared host, another persona's files are not yours to touch or tidy.
  Treat /tmp as disposable, never as storage.
- Anything you run for testing or experiment points at a scratch copy, never
  at live data, live config, or the schedule the real system runs on.
- Write to the paths the system tells you it uses, not ones you remember.
  A plausible-looking wrong directory swallows work silently.
- Secrets live only in the vault. Never leave one in a config file, a
  settings file, a note or a command you save for later.
- One canonical place per thing — one credential, one setting, one rule.
  A second copy elsewhere will quietly win one day and you won't know which.
```

## Why prose in the binary, and not a file on the host

Three alternatives were considered and rejected:

- **The persona's own `AGENTS.md`** — optional (`create-persona` scaffolds `SOUL.md`/`IDENTITY.md`/`MEMORY.md` only; `AGENTS.md` is read-if-present in `persona/loader.ts`), owner-editable, per-host. A rule written there reaches exactly the personas that already knew it.
- **`SOUL_TEMPLATE`** — `personaBackfill.ts` writes `SOUL.md` only when it is missing and never overwrites, so this reaches new personas only.
- **Rewriting a managed section of a host file at boot** — buys nothing. The loader splices `AGENTS.md` into the same system prompt, inside the same cache prefix, so the token cost is identical; it adds file mutation on every start and merge conflicts with hand edits.

The split this settles on: **the binary states the principle, the persona's own files hold the concrete paths.** Same division that already works for voice and memory.

## Where the lines come from

Every line is a real incident generalised, not invented policy:

| Line | Incident |
|---|---|
| own persona directory / `/tmp` | ~1 GB of scattered checkouts and leftover scratch dirs at rig level, outside any persona dir |
| scratch copy, not live | a dev checkout running under a repo *fixture* persona scheduled 75 real systemd nightly units on a live host; `bun test` rewriting the live host `state.json` (#518) |
| paths the system tells you | a stale shadow memory directory that silently swallowed writes |
| secrets only in the vault | plaintext credentials left in a settings file and in a `config.toml` |
| one canonical place | a `GITHUB_TOKEN` env var silently shadowing the classic PAT; duplicate norms making an injected rule nondeterministic |

Deliberately **no git, repo or branch vocabulary** — most operators are not software engineers, and none of these failure modes are developer-specific. A test asserts that.

Two rules that were drafted and cut on purpose: a branch-hygiene line (engineer-only toil) and an "upgrade freely / ask before shared upgrades" line (reads as a licence to self-upgrade; agents that behave do not need it and ones that do not will not be stopped by prose).

## Cost

~45 tokens of the section body, in the cached stable prefix. A test pins the section under 900 characters, because it is on every prompt of every persona forever.

## Tests

`tests/persona-builder-hygiene.test.ts` — 9 tests. They pin the **wiring**, not the constant: present on a bare persona with no files of its own, on both the trusted and untrusted perimeter paths, and inside `buildStableSystemPrompt` (the cacheable prefix).

Mutation-proven: commenting out `sections.push(LOCAL_HYGIENE_SECTION)` gives **3 failed** (was 9 pass / 0 fail).

## Verification

- `bun test` → 4515 pass, 8 skip, **0 fail** across 274 files
- `bunx tsc --noEmit` → exit 0
- `AGENTS.md` invariant 46 records the design and the content rules


src/persona/builder.ts
```diff
@@ -485,6 +486,39 @@ Release when you finish. A claim from a crashed turn expires on its own, so a
...
+  `# Local hygiene
+
+- Keep your work inside your own persona directory — files, scratch, notes.
+  On a shared host, another persona's files are not yours to touch or tidy.
+  Treat /tmp as disposable, never as storage.
+- Anything you run for testing or experiment points at a scratch copy, never
+  at live data, live config, or the schedule the real system runs on.
+- Write to the paths the system tells you it uses, not ones you remember.
+  A plausible-looking wrong directory swallows work silently.
+- Secrets live only in the vault. Never leave one in a config file, a
```
kaieriksen: **Major:** This absolute rule conflicts with the credential contract injected a few sections later. `CREDENTIALS_SECTION` explicitly tells the agent to inspect `~/.ssh/` (where private keys legitimately live), shell exports, OS keychains/password managers, service `Environment=`, and other existing stores, while calling the vault a convenience layer rather than a cage. An agent following this new higher/earlier instruction can treat valid SSH/keychain/service credentials as hygiene violations, move or delete them, or refuse to use them. Please narrow this to credentials Phantombot/personas persist (for example, ‘Credentials you store for later belong in the persona vault; never copy them into notes or app settings’) and add a test asserting that the combined prompt does not claim all secrets can exist only in the vault.

kaieriksen (changes requested):
One major prompt-contract conflict needs resolving before this is safe to ship. I could not execute the Bun tests on this host because Bun is not installed; the submitted targeted/full-suite results remain the available runtime evidence.

robertclawson:
Fixed in dab15f4 — thanks, this was a real conflict and I'd missed that the prompt is one document.

**Change:** the hygiene line is now scoped to credentials the persona itself persists:

> - A credential you store for later belongs in the vault. Never copy one
>   into a note, a settings file, or a command you save for later.

No vault-exclusivity claim, and it no longer says anything about pre-existing stores, so `CREDENTIALS_SECTION`'s "follow your nose" guidance (~/.ssh, shell exports, keychains, unit `Environment=`) stands unopposed.

**Test:** new `does not contradict the credential contract injected after it` asserts against the COMBINED prompt — both sections present, and the hygiene text matches neither a vault-exclusivity pattern nor any mention of ssh/keychain/password manager/Environment=. Mutation-proven: restoring the old wording verbatim gives 2 failed (the new test plus the wording assertion); at head 10 pass / 0 fail in the file, 4516 pass / 8 skip / 0 fail full suite, tsc 0.

Re-requesting review.

kaieriksen (approved):
Re-reviewed the fix at dab15f45. The vault rule is now correctly scoped to credentials the persona persists, and the combined-prompt regression test covers the original conflict. Targeted hygiene suite passes locally: 10/10; CI typecheck, Linux tests, Windows tests, and CodeQL are green.

PR #538, opened 2026-09-05
feat(prompt): inject registered MCP server inventory (#537)

### LINKED ISSUE #537: Agents default away from registered MCP servers because servers are invisible until searched
## Problem

MCP tools are discovered lazily via `phantombot mcp search`. Nothing in the agent's working context lists which servers are registered. When a familiar pattern is already in hand (curl+jq against api.github.com, a CLI, a script), the agent uses it and never thinks to search the MCP toolbox.

## Observed case

A GitHub MCP server was registered and working, covering PR comments, review requests, and issue creation. The same agent that registered it later did all GitHub API work via curl+jq because `gh` was not installed. The MCP server was never considered: it was invisible in context, and memory surfaced an older working recipe (curl) instead. The work succeeded, so nothing forced a tool reconsideration. Success is a bad teacher.

## Proposed fix

Inject a one-line list of registered MCP server names into every prompt, e.g.:

```
MCP servers: github, google-drive, home-assistant
```

Discovery-on-demand stays fine for rarely used servers, but common ones should be visible without a discovery round-trip. That single line would have changed behaviour in the observed case.

Possible extensions:
- Let a persona mark servers as preferred/first-choice, so the prompt can say `MCP servers (preferred: github)`.
- Have `mcp search` return a hint when a known task domain (git host APIs, calendars) matches a registered server the agent did not search for.

## Why not just a prompt rule

A memory norm ("always use MCP for GitHub") helps but competes with retrieved older recipes and with task framing (git-shaped work pulls toward bash). Structural visibility in the prompt wins over recall every time.

lenaparkhodges:
Closes #537

### Summary
Injects a one-line inventory of registered MCP server IDs into the `# External tools (MCP)` prompt section.

- **Loader:** `loadPersona` reads `mcp.json` via `loadRegistry` and exposes sorted registered server IDs as `mcpServers?: string[]`. Handles missing or malformed `mcp.json` gracefully.
- **Builder:** `buildMcpToolsSection` and `buildStableSections` inject `MCP servers: <comma-separated list>` (or `MCP servers: (none registered)`) into the cacheable stable system prompt prefix. Full tool schemas remain lazy on demand (`mcp search` / `mcp describe`).
- **Tests:** Added comprehensive test suite in `tests/persona-builder-mcp.test.ts` and `tests/persona.test.ts`. 4,526 unit tests passing.

### DIFF
--- a/src/persona/builder.ts
+++ b/src/persona/builder.ts
@@ -105,7 +105,7 @@ function buildStableSections(
 
   sections.push(MEMORY_TOOLS_SECTION);
   sections.push(SCHEDULING_TOOLS_SECTION);
-  sections.push(MCP_TOOLS_SECTION);
+  sections.push(buildMcpToolsSection(persona.mcpServers));
   sections.push(NOTIFICATION_SECTION);
   sections.push(WORKSPACE_LOCK_SECTION);
   sections.push(LOCAL_HYGIENE_SECTION);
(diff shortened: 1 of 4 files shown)


kaieriksen (approved):
Reviewed the full diff at 48881b6. The registry IDs are validated before prompt injection, sorted deterministically for stable caching, and malformed/missing registries degrade safely. Focused persona/MCP tests pass (24/24), typecheck passes, and required Linux/Windows CI is green.

PR #539, opened 2026-09-09
feat(tui): installer presentation, polished setup screens, and a chat that stops answering blank

andrewagrahamhodges:
## What this is

The installer and terminal-app pass: a real first-run presentation, polished setup screens, and three bugs that made a fresh install look broken after you had configured it correctly.

## Installer

- `install.sh` / `install.ps1` open with a multi-tone cyber-phantom emblem and a 2D blue–cyan gradient `PHANTOMBOT` wordmark, 80-column safe, POSIX `sh`-compatible.
- Motto: *"Learns your world, defends your runtime, never wastes a token."*
- Idempotent install, autostart/boot prompt (with linger on Linux), progress checklist, and `phantombot install` run across all platforms.
- `PHANTOMBOT_DEV_BIN` is honoured so the flow can be tested against a local build.

## Terminal app polish

- Setup wizards migrated to Ink; persona questions rewritten with Title Case headings, grammatical copy, and compact example cards instead of full-width empty boxes.
- Menus adopt a lazygit/gum-style highlight pill: the active row sits in a dark contrast background with a bold cyan `▸`; multi-select uses `[✓]` / `[ ]` badges. Applied across `Choose`, `MultiChoose`, `Confirm`, `NewPersona`, `SearchList`, `Keys`, `Mcp`, `Menu`.
- **Shortcut labels are spelled out**: `^q` → `ctrl+q`, and the same for `ctrl+s`, `ctrl+c`, `ctrl+l`, `ctrl+p`, `ctrl+t`. Nobody should have to decode a caret. README table updated to match.
- Brain testing is now an in-place checklist screen (`BrainTest.tsx`) instead of bouncing back to Configure: `[✓] Test in progress…` → `Test successful, apply? (Y/n)`, or `[✗]` → `Test failed, retry? (Y/n)`.

## Three bugs, one blank bubble

Reported live: brain configured, "brain verified: pi", then chat answered with empty bubbles.

1. **The chain was stale.** `ChatSession` resolves its harness chain once at open and deliberately survives screen switches, so nothing re-read `config.toml` after the brain was written — chat kept driving the pre-config chain (default `claude`, not installed). Fixed with `reloadHarnesses()`, which rebuilds in place; reopening the session would have fixed it and wiped the transcript. Called from both the wizard and the Configure → Brain row.
2. **The failure was silent.** `chatSession` dropped error chunks flagged `recoverable`. That flag describes what the *orchestrator* may do, not whether the user was spared — `runWithFallback` consumes the ones it can fall through from, so the only chunks reaching a channel are terminal. It was swallowing precisely the errors worth showing. All error chunks now surface.
3. **An empty turn rendered nothing.** A turn with no text drew a header and blank space. It now renders `(no reply)` in both the TUI and `--no-tui`.

## Model pickers were empty

Picking OpenRouter and pasting a key showed no models: `pi --list-models` only enumerates providers Pi has *already* keyed, so the list is empty at exactly the moment you need it. New `src/lib/providerModelCatalog.ts` fetches the provider's own `/models` endpoint live, reusing `providerKeyProbe`'s endpoint table as the single source of truth. Verified live: OpenRouter's endpoint is public — ~430 models, no key required. Failure falls back to free-text.

No hardcoded model catalogue: I started one and threw it out. It would rot inside a release.

## Docs

README shortcut table updated; AGENTS invariants **47** (rebuild a live session's chain in place), **48** (`recoverable` is not "user was spared"; never render an empty bubble) and **49** (model pickers fetch live, never bake a catalogue) added.

## Testing

- Full suite green (4551 pass). The two `tui-bars` failures reproduce on a clean checkout of the upstream test file and are unrelated to this branch.
- New: `lib-providerModelCatalog`, `tui-brain-test`, `tui-brain-reload`, `tui-chat-dead-brain`, `tui-chat-session-brain`, plus brainFlow regression tests at the call site.
- Every new regression test proven by mutation — reverting the fix fails the test.
- Verified end-to-end in the `phantombot-test` container: the dead-harness turn now prints `error: harness claude threw: Executable not found in $PATH: "claude"` instead of nothing, and the OpenRouter picker fills.


install.ps1
```diff
@@ -1,66 +1,138 @@
...
+    [string]$InstallDir = $env:PHANTOMBOT_INSTALL_DIR
+)
+
 Set-StrictMode -Version Latest
 $ErrorActionPreference = 'Stop'
-$ProgressPreference = 'SilentlyContinue'  # keep Invoke-WebRequest quiet + fast
+$ProgressPreference = 'SilentlyContinue'
+
+if ($dryrun) { $DryRun = $true }
+if ($env:PHANTOMBOT_DRY_RUN -or $env:PHANTOMBOT_DRYRUN) { $DryRun = $true }
```
kaieriksen, line 33: Major: PowerShell identifiers are case-insensitive, so `DryRun` and `dryrun` are the same parameter. Declaring both makes the script fail parameter binding/parsing before installation begins. Keep one `DryRun` switch; callers already get case-insensitive spelling for free. Please add a Windows smoke test that invokes/parses the installer so CI catches this.

src/tui/brainOnboarding.ts
```diff
@@ -304,6 +456,39 @@ async function runOnce(
...
+            await deps.maybePromptRestart();
+          }
+          return {
+            landing: "chat",
+            notice: `brain verified: ${chain.join(" → ")}`,
+          };
+        } else {
+          return {
+            landing: "configure",
+            notice: `brain unchanged — verified ${chain.join(" → ")} (not applied)`,
```
kaieriksen: Major: this says the brain is unchanged, but `configurePi()` has already called `setSecret`/`unsetSecret`, written Pi's auth store, and `applyRouting` before the test screen opens. Choosing ‘No, discard and keep previous’ therefore preserves only the old chain while silently committing the new provider/model/key state. Stage these writes until Apply or snapshot and restore every affected store on discard/failure/cancel, then cover the discard path with prior routing and secret values.

kaieriksen (changes requested):
Two Major regressions need fixing before this is safe to merge. The Windows installer declares duplicate case-insensitive parameters, and the new Brain Test discard option reports that prior configuration was preserved after routing/secrets were already persisted.

install.ps1
```diff
@@ -1,66 +1,138 @@
...
-    - SHA256 mismatch
-    - install dir not writable
+  Skip the TUI launch with $env:PHANTOMBOT_SKIP_TUI=1 (e.g. CI smoke tests).
+  Run without making changes with -DryRun (or $env:PHANTOMBOT_DRY_RUN=1).
 #>
 
+[CmdletBinding()]
+param(
+    [switch]$DryRun,
+    [switch]$dryrun,
```
lenaparkhodges: **BLOCKING — this script does not parse under Windows PowerShell 5.1.** Verified on a real Windows 11 host (stock `powershell.exe`, no pwsh installed):

```
ERROR: At <script>:3 char:13
+     [switch]$dryrun
+             ~~~~~~~
Duplicate parameter $dryrun in parameter list.
```

Parameter binding is case-insensitive, so `[switch]$dryrun` is seen as a second declaration of `$DryRun` and the whole file fails to load — `iwr … | iex` and `powershell -File install.ps1` both die before the banner prints. `-dryrun` already binds to `-DryRun`, so the second parameter is redundant anyway: delete it and everything works. (CI evidently ran under pwsh or never parsed this file — worth a smoke test that at least `[scriptblock]::Create((gc install.ps1 -Raw))` parses in CI on windows-latest with PowerShell 5.1.)

install.sh
```diff
@@ -50,207 +168,304 @@ case "$uname_m" in
...
-    printf '\nnext, run this to finish setup:\n'
-    printf '  phantombot init\n\n'
+    printf 'next, run phantombot in your terminal to start.\n'
     exit 0
   fi
 fi
 
-printf '\nphantombot: launching setup wizard.\n\n'
-exec "$INSTALL_DIR/phantombot" init
+exec $PB_BIN
```
lenaparkhodges: Minor, all three launch sites: `exec $PB_BIN` relies on word splitting, which is intentional for the dry-run `PB_BIN="bun src/index.ts"` case but breaks a real install into a `PHANTOMBOT_INSTALL_DIR` containing spaces (the default `~/.local/bin` is safe, so this is edge-case only). If it's deliberate, a one-line comment saying so would keep a future reader from "fixing" it into `$PB_BIN` quoted and breaking the bun path.

lenaparkhodges (changes requested):
## Review

**Blocking: 1** — the PowerShell installer fails to parse on real Windows. Everything else is excellent: the three-bug chat fix is exactly the right diagnosis (stale chain + swallowed errors + blank render, each with a mutation-proven regression test), the live model catalogue is the right call, and the "checklist tick must be earned" discipline in the installers is a genuine upgrade.

But the Windows installer was verified in the Linux container, not on Windows — and it turns out that matters. See the inline comment; I reproduced the parse failure on an actual Windows 11 host with stock PowerShell 5.1.

Minor (non-blocking): unquoted `$PB_BIN` in install.sh's exec paths.

install.ps1
andrewagrahamhodges: Fixed in 7b12990 — and you were both right that CI never parsed this file. Running it on a real Windows 11 box (stock `powershell.exe` 5.1, no pwsh) found two more fatal faults stacked behind the duplicate switch:

1. duplicate `-dryrun` → `DuplicateFormalParameter`, as you said. Deleted; `-dryrun` binds to `-DryRun` for free.
2. **no UTF-8 BOM** — 5.1 decodes a BOM-less `.ps1` as the ANSI codepage, so the banner’s box-drawing glyphs corrupted mid-string and took the parse with them (`Unexpected token ... in expression or statement` at the banner line).
3. `"$c123$boldLearns your world..."` parses as a variable named `$boldLearns` → `The variable '$boldLearns' cannot be retrieved because it has not been set` under `Set-StrictMode`. Braced it.

Also set `[Console]::OutputEncoding` to UTF-8 before the banner: 5.1 renders host output in the console OEM codepage, so every glyph arrived as `?` even once it parsed.

Verified end to end on the VM: parses clean and `-DryRun` completes with the banner intact.

Guards, per your suggestion: CI's `test-windows` job now parses `install.ps1` with the 5.1 parser (`[Language.Parser]::ParseFile`, which catches the duplicate param — `PSParser::Tokenize` does not) and then runs it `-DryRun`; `tests/install-ps1.test.ts` is the fast local mirror (BOM, case-insensitive duplicate params, encoding-before-banner ordering, interpolation). All four fail against the pre-fix script.

install.ps1
andrewagrahamhodges: Same fix — see the reply on the sibling thread. 7b12990: duplicate switch deleted, plus the two other 5.1-fatal faults your repro was standing on top of (missing UTF-8 BOM, `$boldLearns` interpolation). Verified on a Windows 11 VM with stock `powershell.exe`, and CI now parses + `-DryRun`s the script on windows-latest so this cannot go green again while broken.

install.sh
andrewagrahamhodges: Fixed rather than commented, in 3a17129 — the word splitting was only ever needed for the dev fallback, so it no longer has to be load-bearing for the real path. `PB_BIN` is now always a single executable path and quoted at every call site (`command -v bun` resolves the dev case), and the interpreter argument moved to `PB_ARGS`, which only ever holds literals the script itself sets and is unquoted deliberately with the comment you asked for. Tests: an end-to-end install into a dir with a space in it, plus a guard that no call site drops the quotes again — both fail against the pre-fix script.

src/tui/brainOnboarding.ts
andrewagrahamhodges: Fixed in 9fdc602. You were right that withholding `applyChain` is not "changing nothing": `configurePi` persists each answer as it is given, so discard was leaving the new provider, models, vault secret and Pi `auth.json` entry committed under a notice claiming the brain was unchanged.

`runBrainOnboarding` now snapshots every store immediately before the first `configurePi` pass and restores it on all four non-applying exits: cancel, esc mid-interview, failed test, explicit discard.

- `snapshotPiRouting` / `restorePiRouting` capture the routing sub-table **verbatim, per slot** — a key-by-key restore drops the `use_local_config` tombstone, and a persona that no longer states its routing silently inherits the HOST's. Pinned in a test.
- `snapshotPiAuth` / `restorePiAuth` round-trip Pi's `auth.json` byte for byte at 0600 (tempfile+rename), so an unrelated OAuth entry survives untouched. The only file it will delete is one the wizard itself created.
- The vault secret goes back to its previous value, or is unset when there was none.

If the restore itself fails, the notice degrades to `brain partly saved — ...` instead of claiming an intact brain.

Tests, as asked, with prior routing + secret + auth values present: discard, failed test, mid-interview cancel, apply-keeps-everything, and failed-restore-is-honest (`tests/brain-onboarding.test.ts`), plus store-level round-trips in `tests/cli-harness-routing.test.ts` and `tests/lib-piAuthStore.test.ts`. Stubbing the restore fails four of them.

andrewagrahamhodges:
Both Majors and the Minor are fixed and pushed — @kaieriksen @lenaparkhodges ready for another pass.

| commit | what |
|---|---|
| `7b12990` | `install.ps1` runs on Windows PowerShell 5.1 again: duplicate `-dryrun` switch removed, UTF-8 BOM restored, `${bold}` interpolation fixed, console output forced to UTF-8. Verified on a real Windows 11 VM. |
| `9fdc602` | Brain wizard now snapshots routing + vault secret + Pi `auth.json` before the interview and rolls them back on discard / failed test / cancel. Notice degrades to "brain partly saved" if a restore fails. |
| `3a17129` | `$PB_BIN` quoted at every `install.sh` call site; dev-only argv moved to `$PB_ARGS`. |
| `ad2206a` | AGENTS invariants 52 (wizard writes must be rollback-able and notices must match disk) + 53 (the PowerShell 5.1 floor). |

New CI gate: `test-windows` parses `install.ps1` with the 5.1 parser and runs it `-DryRun`, so a script that cannot load can no longer go green. Suite is 4573 pass locally, every new test proven by mutation or against the pre-fix file.

src/tui/brainOnboarding.ts
```diff
@@ -34,6 +34,200 @@ import type { RoutingChoices } from "../lib/piRouting.ts";
...
+        undefined,
+        "pi-primary",
+        "pi-fallback",
+      ];
+      const routing: Record<string, Record<string, unknown> | undefined> = {};
+      const secrets: Record<string, string | undefined> = {};
+      for (const instanceId of slots) {
+        const key = instanceId ?? "";
+        routing[key] = await snapshotPiRouting(writeTarget.path, instanceId);
+        secrets[key] = await getPersonaSecret(
```
kaieriksen: Major: `getPersonaSecret` returns the effective value, including an ambient/host fallback. If this persona had no vault row before the wizard, discard restores that fallback by calling `setPersonaSecret`, creating a new persona override instead of returning to the original absent/inherited state. Snapshot the strict vault value/declaration state (`getPersonaSecretStrict`) and test an absent persona key with an ambient fallback; discard must unset the wizard-created row, not persist the fallback.

src/tui/brainOnboarding.ts
```diff
@@ -34,6 +34,200 @@ import type { RoutingChoices } from "../lib/piRouting.ts";
...
+        const instanceId = key === "" ? undefined : key;
+        await restorePiRouting(writeTarget.path, table, instanceId);
+      }
+      for (const [key, value] of Object.entries(snapshot.secrets)) {
+        const instanceId = key === "" ? undefined : key;
+        const name = instanceId
+          ? piInstanceSecretName(instanceId)
+          : ENV_PI_API_KEY;
+        if (value === undefined) {
+          await unsetPersonaSecret(config, name, persona);
```
kaieriksen: Major: the rollback still reports success when `unsetPersonaSecret` returns `{ ok: false }`, so this path can leave the wizard's new key committed while `discard()` says `brain unchanged`. Also, a thrown `restorePiRouting` aborts before the other stores are attempted and bypasses the promised `brain partly saved` notice. Collect each restore result/error, continue attempting the remaining stores, and return false if any step fails; cover at least the failed-unset path.

kaieriksen (changes requested):
Two rollback defects remain on the new head: the snapshot does not preserve vault declaration state, and restore failures can still be reported as success or abort the rollback. The installer fixes themselves are verified on Windows PowerShell 5.1.

src/tui/brainOnboarding.ts
lenaparkhodges: Confirmed on the new head — `getPersonaSecret` (vaultSecrets.ts:138) falls back to `process.env`, so the snapshot records an ambient fallback value and `restoreWrites` mints a persona override via `setPersonaSecret` where the vault had no row. The snapshot needs the strict read plus a has-row/presence flag, and restore must unset when the row was absent.

src/tui/brainOnboarding.ts
lenaparkhodges: Confirmed both: `restoreWrites` ignores `unsetPersonaSecret`'s `{ok:false}` (so an unremovable key still reports `brain unchanged`), and `restorePiRouting` is unguarded in the loop — a throw abandons the remaining stores and propagates out of `discard()`, bypassing the "brain partly saved" notice. Catch per store, keep restoring the rest, aggregate into the boolean.

lenaparkhodges (changes requested):
Installer fixes are good — I re-ran the new `install.ps1` on a real Windows 11 box (stock PS 5.1): parses clean under `[Language.Parser]::ParseFile` and `-DryRun` completes with the banner glyphs and colors intact, touching nothing on disk. `install.sh` quoting fix is right, and CI parsing the file with the 5.1 parser closes the gap that let this through.

Still changes requested: the two rollback Majors on `brainOnboarding.ts` (threads at lines 169 and 189) are confirmed on my read — the snapshot's effective-value secret read plus the unguarded restore loop mean discard can both mint new vault state and claim success. Those need fixing before this merges.

src/tui/brainOnboarding.ts
andrewagrahamhodges: Fixed in 32c39ed. The snapshot now reads `getPersonaSecretStrict` (the persona row only) and records the process env on its own. An absent row is **unset** on restore, never re-set to the ambient value, and `process.env` goes back to what the snapshot saw, so the TUI does not lose the host key until restart. Test through the production deps and a real vault (`tests/brain-onboarding-rollback-vault.test.ts`): no row plus a host-wide `PHANTOMBOT_PI_API_KEY`, then discard, leaves no row. It fails when the read is swapped back to `getPersonaSecret`.

src/tui/brainOnboarding.ts
andrewagrahamhodges: Fixed in 32c39ed. `restoreBrainWrites` catches each store on its own and keeps going. A `{ok:false}` counts the same as a throw, and the result is the AND of every step. `discard()` also catches a throwing `restoreWrites`, so the `brain partly saved` notice is always reached. Tests in `tests/brain-onboarding.test.ts`: a failed unset returns false; a throwing routing restore still tries the other slots, the secret and auth, in order, then returns false; a throwing `restoreWrites` still gives `partly saved`. Each one fails when its guard is mutated out.

andrewagrahamhodges:
@kaieriksen @lenaparkhodges both rollback Majors are fixed in 32c39ed and I replied on each thread. The strict vault read means discard no longer turns an ambient key into a persona override. Every store is restored even if an earlier one fails, `{ok:false}` counts as a failure, and a throw still reaches the `partly saved` notice. Each fix has a test that fails when the fix is mutated out. Full suite green apart from the known local-only `tui-bars` pair. You both confirmed the installer threads were fixed, so I resolved those. Ready for another pass.

lenaparkhodges (approved):
Both blocking findings from my REQUEST_CHANGES are resolved in 32c39edf and verified:

- **Secret snapshot no longer mints persona overrides.** `snapshotBrainWrites` now reads the strict vault row via `getPersonaSecretStrict` — a missing row is unset on rollback, never re-created from the effective (process.env) value. Restoring `process.env` separately is the right call since set/unset mirror into it.
- **Rollback is fail-open to the honest notice.** `restoreBrainWrites` attempts every store even after an earlier failure, counts `{ok:false}` like a throw, and `runOnce` catches so the "brain partly saved" message always lands. The AND-of-all-stores verdict is correct.
- New `brain-onboarding-rollback-vault.test.ts` (94 lines) covers the vault-vs-env distinction; CI green on this commit (test, test-windows, typecheck, CodeQL).

Installer fixes were verified separately on Megan (PS 5.1 parse + -DryRun). Approving.

kaieriksen (approved):
Re-reviewed exact head 32c39edf. Both rollback Majors are resolved: snapshots preserve strict vault declaration state separately from ambient process state, and restoration now attempts every store while treating throws and {ok:false} as failure so the notice remains honest. Verified the production vault wiring, focused rollback suite (23/23), TypeScript, green Windows PowerShell 5.1 CI, and all required checks. Approving.

PR #540, opened 2026-09-10
feat(tui): OpenClaw import on the first-run wizard

andrewagrahamhodges:
## What

The first-run wizard now opens on one pick:

- **Create a new persona** (highlighted by default, so a new install just presses Enter)
- **Import from OpenClaw**

An import from here continues into the **same Brain steps** as Create. Configure → New persona → Import still lands in Configure, the same as Configure's Create.

## Changes

- **Wizard opening pick** (`Wizard.tsx`). It only shows on a fresh run: a resume (default persona missing its identity) goes straight to the identity question as before. esc on the name box goes back to the pick. On first run the pick shows `ctrl+q Quit`, not `esc Back`.
- **OpenClaw-only labels** in the TUI and the clack `phantombot persona` menu. A phantombot persona's `identity.json`, vault and `memory.sqlite` rows don't survive a markdown copy, so the menu no longer promises that. Phantombot-to-phantombot migration is out of scope.
- **Import prompts:** the path is pre-filled with `~/.openclaw/workspace` when it exists, `workspace` is never suggested as a persona name, and a typed name is validated.
- **`ChooseScreen`** gains AskScreen's `noBack` / `onQuit` first-step mode.

## Bugs fixed along the way

1. **Voice key went to the wrong vault.** `maybeImportOpenclawVoice` wrote the key to `config.defaultPersona`, but `config` is loaded *before* the import adopts a default. On a fresh host that names a persona that doesn't exist, so opening its vault created a stray `personas/<default>/` with its own `identity.json` and vault, and the imported persona never got the key. It now uses the imported persona's name.
2. **Back-to-back asks reused one screen.** Clearing one overlay request and setting the next land in the same React render batch, so `AskScreen` never unmounted and the second question inherited the first one's state. The import's "Persona name" box opened holding the path, and Enter submitted the path as the name. This affected **Configure → Import** on main today. Every ask/choose/search request is now keyed.

## Tests

- `tests/tui-app-first-run.test.tsx`: drives the real `<App>` from pick → Import → a real OpenClaw import → asserts the Brain steps run for the imported name, and that the name box opens with the suggested name, not the path.
- `tests/tui-wizard-language.test.tsx`: first-step footer and esc back to the pick.
- `tests/cli-import-persona.test.ts`: the voice key lands in the imported persona's vault and no stray default dir is created.
- All proven by mutation: removing the overlay key, landing the wizard import like Configure's, removing the pick, and reverting the voice target each fail the tests.
- Full suite: 4585 pass. The 2 failures are the pre-existing `tui-bars` pair.

## Docs

The README OpenClaw import section covers both TUI entry points. AGENTS invariants 54 and 55 are added.


kaieriksen (approved):
Reviewed at ca96f0269a3f1db38067a1fc12727763b719eed7. The first-run Create/Import routing, fresh overlay keys, OpenClaw name validation/defaulting, and imported-persona vault targeting are correct. Exact-head focused import/TUI tests pass in isolation (including 10/10 repeated create-flow runs), typecheck passes, and all hosted checks are green.

PR #541, opened 2026-09-10
phantomchat: surface a failed turn instead of going silent

robertclawson:
A phantomchat turn that threw mid-stream, or that emitted an `error` chunk from the harness chain, was handled by `log.warn` + a bare `return`. Whatever had already streamed stayed delivered — in practice the pre-tool narration sentence ("let me pull that up…") — and then nothing else ever arrived.

Telegram does not behave this way (`core/engine.ts` generates a recovery reply); ACP prints `[error]` into the session. phantomchat was the only surface where a failed turn was invisible, which is exactly why it is the only one showing "answers once, then goes quiet".

Worse: phantomchat had **no `chunk.type === "error"` branch at all**. An error chunk was silently dropped and the loop exited as if the turn had succeeded — no log line, no bubble, nothing.

## What changes

- Track `errored` from both an `error` chunk and a throw out of the stream (the `catch` no longer early-returns).
- After the loop, if the turn failed, call `generateRecoveryReply()` once — the same path Telegram uses — and publish the result as an ordinary reply.
- If even recovery produces nothing, send a fixed marker rather than staying quiet. Silence is the bug.
- A recovery reply *replaces* the streamed text instead of being suffix-diffed against it: the streamed text is a half-finished answer, so a suffix diff would deliver nothing.

Behaviour is unchanged on the success path, on `/stop` / `/reset` aborts, and in groups.

## Scope

This does not add retry, and it does not fix relay-level message loss. Those are separate follow-ups (outbound delivery receipts + retry; relay set refresh).

## Tests

Three new cases in `tests/channels-phantomchat-server.test.ts`, all of which fail on `main`:
- an `error` chunk mid-turn produces a recovery reply
- a throw mid-stream still reaches the user
- when recovery itself fails, the fixed marker is sent

Full suite: 4529 pass / 0 fail. `tsc --noEmit` clean.


src/channels/phantomchat/server.ts
```diff
@@ -1137,9 +1148,37 @@ export async function runPhantomchatServer(
...
+        error: errored,
+      });
+      recoveryText = await generateRecoveryReply({
+        harnesses,
+        userMessage: msg.text,
+        personaName: input.persona,
+        signal: controller.signal,
+      });
+      if (!recoveryText?.trim()) {
+        recoveryText = "⚠️ That turn failed before I could finish — please ask me again.";
```
lenaparkhodges, line 1192: Non-blocking: this fallback is English-only, while the recovery reply's whole point is language matching — so a non-English speaker whose turn fails twice gets an English canned line. Telegram's alternative in that spot is silence, so this is still strictly better; just worth a follow-up if we ever localize channel strings.

src/channels/phantomchat/server.ts
```diff
@@ -1137,9 +1148,37 @@ export async function runPhantomchatServer(
...
+    // language-matched "hit a snag, mind trying again?" and deliver that as an
+    // ordinary reply. If even recovery produces nothing, fall back to a fixed
+    // marker rather than going quiet: silence here is the whole bug.
+    let recoveryText: string | undefined;
+    if (errored) {
+      log.error("phantomchat: turn failed; generating recovery reply", {
+        conversation: conversationKey,
+        error: errored,
+      });
+      recoveryText = await generateRecoveryReply({
```
lenaparkhodges: Non-blocking: the `finally` block has already published the explicit typing-STOP by the time recovery runs, so the user can wait up to ~60s (recovery hard cap) with no dots before the bubble lands. Telegram has the same gap. A cheap polish would be restarting the typing pulse for the recovery turn.

lenaparkhodges (dismissed):
Approved on `316d862`. ✅

The parity argument is what sells it: Telegram already sets `errored` on the error chunk and throws, and already routes failures through `generateRecoveryReply()` — phantomchat was the one surface that dropped the chunk and returned silently. This ports that path faithfully: abort gate (`/stop`/`/reset`) still returns before recovery, the recovery reply replaces rather than suffix-diffs the streamed half-answer (correct — the suffix of a half-finished answer is nothing), and the fixed marker guarantees the user is never left staring at an unresolved narration line.

Verified against the code: the `error`-chunk flag sits exactly where engine.ts's does, `generateRecoveryReply`'s input matches `RecoveryReplyInput`, and the voice-out path correctly synthesizes a recovery reply as voice in voice-mode conversations. CI green including `test-windows`.

src/channels/phantomchat/server.ts
robertclawson: Fixed the typing gap in ed2b1f0 — the recovery path now re-pulses the indicator (first tick on the macrotask queue, then the same 2s interval) and publishes a STOP again once the bubble lands, so the dots never hang either. Regression test asserts a typing START after the turn's STOP and more than one STOP overall; it fails on the unfixed source.

On the English-only marker: I've attacked the layer above it rather than the constant. `generateRecoveryReply` now takes `replyLanguageName` and states the language outright instead of asking the model to infer it from the user's message — which matters here because recovery runs precisely when the chain has fallen over and is being served by a weaker harness. Wired at both call sites, so Telegram gets it too. The fixed marker is now only reachable when recovery itself produced nothing, i.e. when no model is available to translate anything; localizing that genuinely needs a channel-string catalogue, which is your follow-up.

src/channels/phantomchat/server.ts
robertclawson: See the reply on the sibling thread — both nits are addressed in ed2b1f0.

lenaparkhodges (approved):
Re-approved on `e01164f5`. ✅

The follow-up commit addresses exactly the two non-blocking notes from my earlier review, and both fixes are right:

1. **Typing dots during recovery** — the turn's `finally` publishes the explicit typing-STOP before recovery runs, so recovery now re-arms the pulse (`setTimeout 0` + `setInterval 2s`) and stops it again in its own `finally`. The dead-chat window is gone. Regression test asserts the exact event order (STOP → START after → final STOP).
2. **Recovery reply language** — `replyLanguageName` is now threaded through at both call sites (phantomchat + core engine) and the prompt states it outright ("Reply in Spanish.") instead of leaving inference to a weak fallback harness — precisely the case where inference is least reliable. Regression test asserts the named-language prompt and the absence of the generic fallback wording.

Both tests are proven against real behavior (prompt capture, typing event sequence), not just shape checks. CI fully green including `test-windows`. Ship it.

PR #543, opened 2026-09-10
channels: rename TELEGRAM_REPLY_INSTRUCTION and drop its skip-narration line

robertclawson:
## Problem

`TELEGRAM_REPLY_INSTRUCTION` is the chat-style system-prompt suffix. It has three call sites: `channels/telegram.ts`, `channels/core/engine.ts` (Telegram) and `channels/phantomchat/server.ts`. So the Telegram-specific name — and the literal opener "You're chatting via Telegram" — are wrong for a third of its uses. TUI (`tui/chatSession.ts`) and ACP (`acp/turnBridge.ts`) pass no reply-style suffix at all.

It also contained:

> Skip narration ("Let me…", "Right, here's what I found…"); answer directly.

while every interactive turn additionally carries the orchestrator overlay `PRE_TOOL_NARRATION_INSTRUCTION`:

> before each tool call, say ONE short sentence describing what you're about to do.

Two owners of the same rule, saying opposite things, in one prompt. A strong model reconciles it. A model already short on completion budget takes the loudest, shortest reading: emit one narration sentence, "answer directly", stop — **zero tool calls, clean exit, no error**. Nothing downstream catches that: the harness chain only advances on a *failure*, and #541's recovery reply only fires on a thrown error. Observed in the field on a persona running a reasoning-heavy model over phantomchat, never over TUI.

## Change

- Rename `TELEGRAM_REPLY_INSTRUCTION` → `CHAT_REPLY_INSTRUCTION`; heading `# Reply style (Telegram chat)` → `# Reply style (chat)`; opener made transport-neutral.
- Delete the "Skip narration" sentence. Narration is now owned solely by the orchestrator overlay.
- Doc comment updated to say why the line must not come back.

Everything else in the block is unchanged and deliberately kept — the brevity rule (phones, narrow column) and the voice/text reply-mode section, which is the only place `phantombot reply-mode text|voice|default|disable` is documented to the model.

No behaviour change beyond the prompt text: same call sites, same stacking with `VOICE_REPLY_INSTRUCTION`, same re-export from `channels/telegram.ts`.

## Tests

Two assertions added to the text-in/text-out Telegram prompt test: the chat suffix must NOT contain "Skip narration", and the orchestrator's narration section must be present. Proven by restoring the deleted sentence in `prompts.ts` — 1 fail. Full suite: **4592 pass, 0 fail**. `tsc --noEmit` clean.

### DIFF
--- a/src/persona/builder.ts
+++ b/src/persona/builder.ts
@@ -610,7 +610,7 @@ is in, just answer.`;
  * Channel-agnostic conduct overlay: plan-then-confirm before a long or
  * state-changing job, and keep direct answers short.
  *
- * This is the block PR #443 removed from `TELEGRAM_REPLY_INSTRUCTION`,
+ * This is the block PR #443 removed from `CHAT_REPLY_INSTRUCTION`,
  * restored deliberately and in a different shape. #443's three
  * complaints were real, and each is answered here:
  *
--- a/src/channels/core/engine.ts
+++ b/src/channels/core/engine.ts
@@ -97,7 +97,7 @@ import type { GroupChatState } from "./routing.ts";
 import {
   captureNudgeForTurn,
   languageReplyInstruction,
-  TELEGRAM_REPLY_INSTRUCTION,
+  CHAT_REPLY_INSTRUCTION,
   VOICE_REPLY_INSTRUCTION,
   voiceUnavailableMessage,
 } from "./prompts.ts";
@@ -1389,7 +1389,7 @@ async function processChatMessage(
         input.agentDir,
       ),
       // Channel-layer prompt suffix:
-      //   - Always: TELEGRAM_REPLY_INSTRUCTION — short conversational
+      //   - Always: CHAT_REPLY_INSTRUCTION — short conversational
       //     replies + plan-then-confirm before long jobs (git/build/
       //     deploy or anything that would spawn more than one tool call).
       //   - Voice-out: stack VOICE_REPLY_INSTRUCTION on top — stricter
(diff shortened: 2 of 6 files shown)


src/persona/builder.ts
```diff
@@ -610,7 +610,7 @@ is in, just answer.`;
  * Channel-agnostic conduct overlay: plan-then-confirm before a long or
  * state-changing job, and keep direct answers short.
  *
- * This is the block PR #443 removed from `TELEGRAM_REPLY_INSTRUCTION`,
+ * This is the block PR #443 removed from `CHAT_REPLY_INSTRUCTION`,
```
lenaparkhodges, line 613: Nit, non-blocking: this now reads as if #443 removed the block from `CHAT_REPLY_INSTRUCTION`, but the constant was still named `TELEGRAM_REPLY_INSTRUCTION` at the time. Suggest "…removed from the chat reply suffix (then named `TELEGRAM_REPLY_INSTRUCTION`)" so the history stays grep-able.

lenaparkhodges (approved):
**Approved.** Clean, well-reasoned fix for a real contradiction.

What I verified on the head commit:

- **The bug is real and the diagnosis is right.** Every interactive transport enables `toolNarration: true` (engine, phantomchat server, TUI, ACP) and gets `PRE_TOOL_NARRATION_INSTRUCTION` last in the overlay stack — so a channel-local "skip narration" line shipped directly opposite to it on every interactive turn. Voice turns correctly disable narration, so removing the line from the shared suffix loses nothing there. And the failure mode is the dangerous kind: a model that "answers directly" without calling tools exits cleanly, so neither the failure-driven harness chain nor the #541 recovery path catches it.
- **No stale references.** `TELEGRAM_REPLY_INSTRUCTION` survives only as historical context in AGENTS.md invariant 29, which is accurate and should stay. No "Skip narration" or "Reply style (Telegram" strings remain anywhere.
- **Public API preserved.** The re-export from `channels/telegram.ts` is intact, so #162's compatibility promise holds.
- **Tests prove the right things:** the negative assertion (suffix must NOT contain "Skip narration") plus the positive one (orchestrator narration section present) directly pin the regression described, and the restore-the-line reproduction is the correct way to validate a not-contains test.
- CI green across the board including `test-windows`.

One non-blocking nit inline (historical reference in a builder comment). Ship it.

PR #544, opened 2026-09-11
TUI: MCP servers menu in Configure, and doctor as a grouped checklist

robertclawson:
Two TUI changes, one branch, one commit each.

## 1. Configure → MCP Servers (list, probe, delete)

The MCP screen **existed and was unreachable**. No menu row pointed at it, the
router rendered it with `servers={[]}`, and its `onTest` set the notice `"mcp
test not wired yet"`. Every part of it was dead code.

- New **MCP Servers** row on a phantom's settings screen, badged with the count
  from its own `mcp.json`. Zero registered reads `optional`, not red — a
  phantom with no external tools is a normal phantom.
- The screen lists servers with transport, auth method and status, and shows
  the **selected** entry's command line (or endpoint) plus the vault keys it
  references underneath — so a delete is decided with those in view, not from a
  bare name.
- `t` probes one, `a` probes all, `d` deletes.

**Listing and probing are separate.** Rows come off disk and paint immediately;
status starts as a dim `— not probed` and fills in per row. The first cut
probed everything up front and the screen sat blank for as long as the slowest
`npx` cold start — which is exactly the server you opened the screen to delete.

**Reachable means a tool count, not a socket.** A server that connects and then
answers nothing to `tools/list` gives the agent no tools; painting that green is
the thing this screen exists to surface.

**Delete asks twice, and only when there is something to ask about.** Purging
the vault keys is a second, separate question: the registry entry is re-addable
from notes, the vault is the only copy of the credential. Skipped entirely when
the entry references no secrets, so it does not become a reflex.

**An unreadable registry is not an empty one** — no rows *plus* a notice saying
why, rather than inviting the user to re-add servers that are already there.

Nothing is reimplemented: `loadRegistry` / `removeServer` / `saveRegistry` and
`McpHub` are the same readers, writers and client `phantombot mcp` uses.

> ### This reverses a deliberate prior decision
> `tests/tui-navigation.test.tsx` pinned *"Vault and MCP are gone from both the
> table and the settings screen"* — removed as "not the setup `phantombot init`
> performs", and one more row a first-time user had to decide about. That test
> is **updated, not deleted**, and it now records both halves: MCP is back on
> the persona settings screen, and it is still absent from the **host table**,
> which is the part of that reasoning that stands. **Vault stays gone** — it was
> not asked for.

## 2. Doctor is a grouped checklist

Reported: *Telegram looks like a legacy doctor check.* It is not legacy — it was
the **only channel check the screen rendered**, and it read as a leftover
because it had no siblings.

The screen rendered four checks (telegram, memory db, nightly, backlog) and
dropped the rest. Everything else `runDoctor` already computes — systemd units,
timer staleness, per-persona maintenance, the default persona, harness
binaries, the Pi extension, ACP editor registrations, MCP servers, embeddings,
service logs — existed only in the CLI's text output. **So the screen could
paint all-green on a box whose `doctor` exit code was non-zero.** Added, not
removed.

`doctorChecklist` is a pure function of the report in its own module, so what
appears on screen is testable without rendering and the screen owns only the
paint. `editorConnectorBroken` / `missingHarnesses` stay the single authority
for their verdicts — an ACP proposed-api miss is red here **because it is
non-zero there**.

Three rules:

1. Grouped **CORE / CHANNELS / INTEGRATIONS**. A flat wall of ticks is as
   unreadable as no output.
2. One line per check when healthy; `detail` is a value worth a glance (a size,
   a count, a version), never a sentence.
3. The `→ what to do` row appears **only** on a check that is not green.

`warn` deliberately does not read as `bad`: "no Telegram account" is a fine
state for a phantom that does not use Telegram, and an operator who learns
yellow means nothing stops reading red too. Where doctor **repaired** something
the line says so (`repaired 1 unit`, `re-stamped`) — a self-healed fault
rendered as plain green loses the fact that anything was wrong.

The report is now taller than a small window, so the body scrolls (`↑↓`). Yoga
shrinks children rather than clipping, so a too-tall body paints rows on top of
each other; the window is a clamped slice, not `scrollWindow`, which centres on
a cursor through variable-height blocks and has neither here.

A healthy box now renders as:

```
 phantombot ▸ doctor                                          ● healthy
 ALICE
 ──────────────────────────────────────────────────────────────────────
 CORE
 ✓ memory database       42 MB · 1 restore point
 ✓ nightly sweep         no backlog
 ✓ capture               9 from 10 turns in 24h
 ✓ heartbeat timer       fired 4m ago
 ✓ tick timer            fired 1m ago
 ✓ systemd units         all present and armed
 ✓ persona maintenance   2 personas current
 ✓ default persona       alice (from state)
 ✓ release ring          stable · v1.1.359
 CHANNELS
 ✓ telegram              2 listeners across 2 personas
 ✓ embeddings            openai-compatible · semantic search on
 INTEGRATIONS
 ✓ harness binaries      claude
 ✓ mcp servers           3 servers for alice
 ✓ pi extension          present and current
 ✓ acp: zed              registered and current
 ✓ acp: vscode           registered and current
```

## Not in this change

**phantomchat and vault reachability.** Both need *new probes* in `runDoctor` —
there is no phantomchat check anywhere in the codebase today — which changes the
CLI's output and its exit-code surface. That is not a screen change, and a
placeholder behind a green tick is worse than an honest absence. Happy to raise
it as a follow-up issue; phantomchat is the one that has actually been biting
the fleet.

## Proof

`bun tsc --noEmit` clean, full suite **4624 pass / 8 skip / 0 fail**.

Every new test was proven by **neuter**, not by passing:

| neuter | failures |
|---|---|
| drop the MCP row from `ROWS` | 3 |
| restore `servers={[]}` on the MCP screen | 3 |
| ignore the delete confirm's answer | 1 |
| drop the `INTEGRATIONS` section | 6 |
| give a green check a hint | 3 |
| force the verdict never to go red | 2 |
| remove the scroll slice | 1 |

Two of those were **decorative on the first pass** and are recorded here because
the fix is the interesting part:

- The scroll assertion read the `more below` marker — which is computed from the
  row count and shows even when every row is still painted. It now asserts the
  off-window rows are genuinely *absent*, and that scrolling trades one end for
  the other.
- The "no hints on a healthy box" assertion passed against a report that had no
  hints to begin with. It is now an invariant checked across fifteen separately
  degraded reports (each one asserted to actually degrade something first), which
  is what lets the paint drop its redundant second state check.

The MCP fixture also caught a real registry rule on the way through: `env` auth
is stdio-only, so an `http` entry declaring it is rejected at load — which is
how the "unreadable registry ≠ empty registry" notice got exercised for free.

### DIFF
--- a/src/tui/doctorChecklist.ts
+++ b/src/tui/doctorChecklist.ts
@@ -1,0 +1,421 @@
+/**
+ * The Doctor screen's checklist, as DATA.
+ *
+ * The screen used to render four of the report's checks — telegram, the memory
+ * DB, the nightly sweep and its backlog — and drop the rest on the floor.
+ * Everything else `runDoctor` computes (systemd units, timer staleness,
+ * per-persona maintenance, the default persona, harness binaries, the Pi
+ * extension, ACP editor registrations, MCP servers, embeddings, service logs)
+ * was visible ONLY in the CLI's text output, so the screen quietly claimed a
+ * healthy box while doctor's exit code said otherwise.
+ *
+ * Telegram was the one substantive channel check, which read as a leftover
+ * precisely BECAUSE it was alone. The fix is to add its siblings, not to
+ * remove it (Andrew, 2026-09-12).
+ *
+ * Three rules, which is the whole design:
+ *
+ *   1. GROUPED — Core, Channels, Integrations. A flat wall of green ticks is
+ *      as unreadable as no output; a section heading is what makes "20 checks
+ *      passed" skimmable.
+ *   2. ONE LINE per check when healthy. `detail` is a value worth reading at a
+ *      glance (a size, a version, a count), never a sentence.
+ *   3. `hint` — the "here is what to do" line — appears ONLY on a check that
+ *      is not green. Failures are loud; health is quiet.
+ *
+ * `warn` must never read as `bad`. "No Telegram account" is a perfectly good
+ * state for a phantom that does not use Telegram, and an operator who learns
+ * that yellow means nothing stops reading red too.
+ *
+ * Doctor REPAIRS as well as reports, so where a check healed something the
+ * line says so ("re-stamped", "re-armed") rather than just showing green —
+ * otherwise a self-healed fault is indistinguishable from one that never
+ * happened, and the operator loses the fact that something was wrong.
+ *
+ * This is a pure function of the report: no probing, no reading the disk. The
+ * screen renders what it returns, so the CLI and the TUI cannot disagree about
+ * the health of the same box.
+ */
+
+import type { DoctorReport } from "../cli/doctor.ts";
+import { missingHarnesses } from "../lib/harnessAvailability.ts";
+import { editorConnectorBroken } from "../connectors/acp/autoInstall.ts";
+import { humanBytes } from "./theme.ts";
+
+export type CheckState = "ok" | "warn" | "bad";
+
+export interface ChecklistItem {
+  label: string;
+  state: CheckState;
+  /** A value worth a glance. Shown in every state. */
+  detail?: string;
+  /** What to do about it. Shown only when `state` is not "ok". */
+  hint?: string;
+}
+
+export interface ChecklistSection {
+  title: string;
+  items: ChecklistItem[];
+}
+
+/** `3` → `3 personas`, `1` → `1 persona`. */
+function plural(n: number, one: string, many = `${one}s`): string {
+  return `${n} ${n === 1 ? one : many}`;
+}
+
+function coreItems(r: DoctorReport): ChecklistItem[] {
+  const items: ChecklistItem[] = [];
+
+  items.push({
+    label: "memory database",
+    state: r.memory_db.healthy ? "ok" : "bad",
+    detail: `${humanBytes(r.memory_db.bytes)} · ${plural(r.memory_db.restore_points.length, "restore point")}`,
+    ...(r.memory_db.healthy
+      ? {}
+      : {
+          hint: r.memory_db.newest_good
+            ? `${r.memory_db.detail} — newest good snapshot ${r.memory_db.newest_good}; restore with \`phantombot memory backup\``
+            : `${r.memory_db.detail} — no verified snapshot to restore from`,
+        }),
+  });
+
+  if (r.memory_db.unretired_drawers.length > 0)
+    items.push({
+      label: "drawer files",
+      state: "warn",
+      detail: plural(r.memory_db.unretired_drawers.length, "file") + " still on disk",
+      hint: `retirement held back ${r.memory_db.unretired_drawers.join(", ")} — nothing reads them`,
+    });
+
+  items.push({
+    label: "nightly sweep",
+    state:
+      r.nightly.health === "ok"
+        ? "ok"
+        : r.nightly.health === "error"
+          ? "bad"
+          : "warn",
+    detail: r.nightly.detail,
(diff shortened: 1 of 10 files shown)


src/tui/doctorChecklist.ts
```diff
@@ -0,0 +1,421 @@
...
+    label: "telegram",
+    state: !r.telegram.healthy ? "bad" : stated.length === 0 ? "warn" : "ok",
+    detail:
+      stated.length === 0
+        ? "no account configured"
+        : `${plural(r.telegram.listeners, "listener")} across ${plural(stated.length, "persona")}`,
+    ...(r.telegram.healthy
+      ? stated.length === 0
+        ? {
+            hint: "nothing is wrong — this host simply does not answer on Telegram",
```
lenaparkhodges: Non-blocking, and it's the test's own argument echoed back: the header comment for this file says "an operator who learns yellow means nothing stops reading red too", but this row paints a *fine* state yellow ("no account configured" is normal for a Telegram-less phantom). The hint line mostly disambiguates it, but PersonaDetail has the same tension with its yellow `optional` badge — if yellow is going to carry weight in the INTEGRATIONS/Core sections, consider whether CHANNELS yellow should be reserved for actual degradation (e.g. a dim "no account" row instead). Your call; the hint text is what saves it today.

lenaparkhodges (dismissed):
## Review

Two solid, well-scoped changes. Both data paths (`mcpFlow.ts`, `doctorChecklist.ts`) are pure and testable, both reuse the CLI's own readers (`loadRegistry`/`removeServer`/`saveRegistry`, `runDoctor`, `missingHarnesses`, `editorConnectorBroken`) so TUI and CLI cannot disagree — which was the exact failure mode of the old four-check screen. The neuter table in the description is the kind of test discipline I want to see more of. CI fully green (test / test-windows / typecheck / CodeQL).

The MCP re-add: I checked the history — the "Vault and MCP are gone" pin came from #471, not from an operator directive, and the PR keeps the half of that decision that mattered (still absent from the host table, Vault stays gone). Doctor here is the phantoms-list Doctor screen that shipped in #508; the 2026-08-30 "remove doctor" directive was about the PersonaDetail block and its menu row, which stays gone. No conflict.

Two non-blocking notes inline. Merge is Andrew's call.

src/tui/doctorChecklist.ts
robertclawson: Taken — you were quoting the file back at itself and it was right. Fixed in 5ef725b.

The tell was the hint line: a row that has to say "nothing is wrong" in prose should not be wearing the problem colour. And since most hosts run no Telegram at all, that yellow would have been permanently on — a warning that is always lit is one nobody reads, which then costs red its meaning too, i.e. exactly the thing the grouping was for.

So not-configured is now `ok`, detail `not configured`, and no hint (rule 3 — health is quiet). CHANNELS yellow is reserved for a channel that IS configured and failing; broken is still red with its per-persona detail. Neuter-proven: putting the yellow back fails the assertion.

Left PersonaDetail's `optional` badge alone deliberately — same tension, but it is a pre-existing convention on another screen and not this PR's job.

lenaparkhodges (approved):
Re-approved at 5ef725b. The unconfigured-channel fix is exactly right: green 'not configured', no hint, CHANNELS yellow now reserved for configured-and-failing — and the test asserts the hint stays undefined, so the regression can't come back. The CI red on first run was a flaky telegram slash-command interruption test (untouched files, 5s timeout); rerun is fully green.

PR #545, opened 2026-09-12
doctor: probe phantomchat and the per-persona vault

robertclawson:
Stacked on #544 (base is its branch — retarget to `main` once #544 merges).

#544 added the doctor checklist and deliberately left two checks out, because they are not a screen change: both need **new probes in `runDoctor`**, which moves the CLI's output and exit-code surface. This is that follow-up.

### PhantomChat

Telegram was the only channel doctor could see, which is exactly what made it read as a leftover. PhantomChat is configured **per-persona** in `<persona-dir>/phantomchat.json` with the key in `identity.json` — there is deliberately no `[channels.phantomchat]` block — so nothing in the host config tells you whether a persona can talk, and every failure is silent by construction:

- an unparseable `phantomchat.json`, and a missing/invalid nsec, both make `loadPhantomchatPersonaConfig` return `undefined`, which the boot path reads as "not configured" and skips **without an error**;
- a persona that *is* configured but absent from `autostart_personas` is gated out with a warning on stderr that nobody sees once the service is up.

Only personas that **have** a `phantomchat.json` are reported — a persona that does not use the channel is not a finding, and listing every one as green would bury the line that matters. The autostart gate is *mirrored* rather than imported: `run.ts` already imports `doctor.ts`, so reaching back would close a cycle (same reason `telegramHealthReport` mirrors listener planning). Keep the two in step.

### Vault

The vault key is **derived** from the persona's nsec in `identity.json`, so a lost `identity.json` does not throw anywhere — every secret simply stops arriving and whatever it powered degrades quietly. That is how a revoked embeddings key dropped memory search to keyword-only while doctor said "semantic search off" and no reason (#516/#522).

The probe derives the key and decrypts every row, per persona, reporting counts and the **names** of rows that will not decrypt — never a value. Read-only by construction: `readPersonaIdentityNsec` + `openVaultWithSecret`, **never** `openPersonaVault`, which calls `getOrCreatePersonaIdentity` and would *mint* an identity and a vault for a persona that has neither, turning a health check into provisioning. There is a test asserting doctor leaves neither file behind.

### Semantics

Both feed the exit code only when a **configured** thing is failing. Neither can fire on a box that does not use the feature: no `phantomchat.json` is green and quiet, no vault yet is green and quiet — the rule from #544, that yellow/red is reserved for configured-and-degraded.

One deliberate **non**-failure: a persona with no allowed npubs and TOFU off is unreachable, but that is also what a persona created five minutes ago looks like. It is stated in the detail line and left non-fatal, because a permanently-lit failure is one nobody reads.

### Screen

Picked up as one row in CHANNELS (`phantomchat`) and one in CORE (`secrets vault` — the vault is *data*, like the memory database, not a process to restart). The checklist stays a pure function of the report, so the CLI and the TUI still cannot disagree, and the "every section the report computes reaches the checklist" test now names both.

### Tests

`4641 pass`, typecheck clean. Every new assertion was proved by neuter — phantomchat-unloadable-is-healthy, autostart-gate-ignored, missing-identity-is-healthy, undecryptable-rows-ignored, and a refactor to `openPersonaVault` — each one turning exactly the intended test red.

Part of #544's follow-up list.

### DIFF
--- a/src/cli/doctor.ts
+++ b/src/cli/doctor.ts
@@ -17,7 +17,7 @@
  */
 
 import { defineCommand } from "citty";
-import { existsSync, readFileSync, statSync } from "node:fs";
+import { existsSync, readdirSync, readFileSync, statSync } from "node:fs";
 import { basename, join } from "node:path";
 
 import {
@@ -71,6 +71,13 @@ import {
   listPersonaDirs,
 } from "../lib/personaDefault.ts";
 import { loadRegistry } from "../mcp/registry.ts";
+import {
+  loadPhantomchatPersonaConfig,
+  phantomchatConfigPath,
+} from "../channels/phantomchat/personaStore.ts";
+import { identityFromNsec } from "../lib/nostrIdentity.ts";
+import { readPersonaIdentityNsec } from "../lib/personaIdentity.ts";
+import { openVaultWithSecret, vaultPath } from "../lib/vault.ts";
 import { saveHarnessBins } from "../state.ts";
 import {
   BunSystemctlRunner,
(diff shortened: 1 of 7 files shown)


src/cli/doctor.ts
```diff
@@ -538,6 +630,179 @@ function telegramHealthReport(
...
+      personas.push({
+        persona,
+        stated: true,
+        relays: 0,
+        allowed: 0,
+        tofu: false,
+        listener: false,
+        healthy: false,
+        detail: existsSync(join(agentDir, "identity.json"))
+          ? "phantomchat.json is unparseable, or identity.json holds an invalid nsec"
```
lenaparkhodges, line 683: Non-blocking: this detail conflates two distinct causes (unparseable phantomchat.json vs invalid nsec) behind one string. We already have `readPersonaIdentityNsec` + `identityFromNsec` in scope — splitting it ("phantomchat.json is unparseable" vs "identity.json holds an invalid nsec") would save the operator one guess on the loudest failure path. Fine to leave for a follow-up.

src/cli/doctor.ts
```diff
@@ -538,6 +630,179 @@ function telegramHealthReport(
...
+    const unpaired = allowed === 0 && !pc.tofu;
+    personas.push({
+      persona,
+      stated: true,
+      npub: pc.identity.npub,
+      relays: pc.relays.length,
+      allowed,
+      tofu: pc.tofu,
+      listener,
+      healthy: listener,
```
lenaparkhodges, line 702: Non-blocking, a doctrine question rather than a defect: a persona deliberately kept out of `autostart_personas` (e.g. a migration-staging or test persona that still has a phantomchat.json) reads here as a permanent red with exit 1 — the same "permanently-lit failure nobody reads" risk the comment below guards against for unpaired personas. Telegram's sibling shares the semantics, so I'm not asking for a change in this PR; but the escape hatch ("delete phantomchat.json to unstate") is only implied. Worth one clause in the docs someday.

lenaparkhodges (approved):
Verified on a fresh checkout at 242d22a (bun 1.4.2): full suite 4641 pass / 8 skip / 0 fail, typecheck clean, and I confirmed CI does not run for this PR as filed — `ci.yml` triggers only on `pull_request: branches [main]`, so the green runs on #544's base branch are what cover the stack until retarget. Local run on this head is the real verification here.

What I checked beyond the tests:

- **Autostart mirror is exact.** `run.ts` gates with `autostartPersonas === undefined ? all : Set([defaultPersona, ...autostart])`; `phantomchatHealthReport` builds the identical roster. The `healthy: !stated || listener` shape also matches telegram's `healthy = !current.stated || current.listeners > 0` — the sibling symmetry the PR claims holds.
- **Vault probe is genuinely read-only.** `readPersonaIdentityNsec` returns `undefined` on a missing identity.json (no throw), and `openVaultWithSecret` never provisions — the no-side-effects test pins the important guarantee, that a health check never mints an identity.
- **No false-positive on empty secrets.** `Vault.get` returns `undefined` only for a missing row; an empty-string secret decrypts to `""` and counts as readable, so the `get() !== undefined → secrets++` heuristic is correct.
- **Values stay out of the report** in every path I traced: per-row decrypt failures are collected by name, `openError` carries only the exception message, and json-mode output asserts neither values nor nsec appear.

Two non-blocking inline notes; nothing blocking. Merge is Andrew's call as usual.

robertclawson:
Thanks — both notes taken, and both deliberately deferred rather than dropped, since a push here would dismiss the approval on a PR that is otherwise merge-ready.

1. **Splitting the unparseable-config detail.** Agreed, it is one guess saved on the loudest path. It is a pure string change but it touches the probe's branching, so it lands with the retarget-to-main commit once the base PR merges, or as a one-liner follow-up if the retarget turns out to be a clean fast-forward.
2. **Non-autostart personas reading red.** Same read: it is doctrine, not a defect, and changing it here would desync the telegram sibling that shares the semantics. The escape hatch (remove phantomchat.json to unstate the persona) should be written down rather than implied — filing it as a docs follow-up so it is not carried only in this thread.

Also confirming your CI observation: `ci.yml` triggers on `pull_request: branches [main]` only, so this PR gets no run of its own while it is stacked. Retarget will produce the first real run against main; nothing merges on the strength of the base branch's green alone.

PR #546, opened 2026-09-12
doctor: probe phantomchat and the per-persona vault (relands #545)

robertclawson:
Relands the payload of #545, which never reached `main`: it was still based on #544's branch when both were merged, so GitHub squashed it into that branch (`0a86944`) instead. Both PRs read MERGED and all 725 lines were absent from `main`. Nothing here is new work beyond the two review notes below — the probes are a clean cherry-pick of the approved commit.

**Base is `main`.** No stacking.

### What lands

**phantomchat per-persona channel health.** PhantomChat is configured in `<persona-dir>/phantomchat.json` — there is no `[channels.phantomchat]` block — so nothing in the host config could see it. The probe scans persona dirs and catches the three states the boot path swallows as "not configured": an unparseable file, a missing or invalid `identity.json`, and configured-but-absent-from-`autostart_personas`. Only personas that *have* the file are reported, so a Telegram-only host stays quiet and green.

The autostart gate is mirrored here rather than imported from `run.ts` (which already imports `doctor.ts` — a cycle), exactly as `telegramHealthReport` mirrors listener planning. The two must be kept in step.

**Per-persona vault reachability.** Derives the key from the persona's nsec and decrypts every row, reporting counts and the *names* of rows that will not decrypt — never values. Read-only by construction: `readPersonaIdentityNsec` + `openVaultWithSecret`, never `openPersonaVault`, which calls `getOrCreatePersonaIdentity` and would mint `identity.json` and `vault.sqlite` as a side effect of a health check. A test asserts doctor leaves neither behind.

Deliberate non-failure: `allowed_npubs` empty with TOFU off is unreachable, but that is also what a five-minute-old persona looks like, so it is stated in detail only. A permanently-lit check is one nobody reads.

Checklist rows: phantomchat under CHANNELS, "secrets vault" under CORE (it is data, like the memory DB).

### Scope of a red row

Diagnostic only, by design — a health check must never gate startup. A failing probe sets `healthy: false`, which makes `phantombot doctor` exit 1 and paint a red row. Nothing consumes that: the startup doctor in `run.ts` is fire-and-forget and the exit code is only logged, and no systemd unit, launchd plist, update path or CI job reads it. Both probes catch every error internally, so a vault that will not open becomes a detail string rather than a throw.

### The two review notes from #545

Both were deferred there to avoid dismissing an approval on a merge-ready PR; they are the second commit here.

1. The "unparseable **or** invalid nsec" detail string conflated two causes with different fixes. `phantomchatLoadFailure` now re-resolves the identity and names the actual one, including the nsec parse error.
2. The escape hatch for a deliberately-muted persona was only implied. The detail string and the README now say it outright: delete the persona's `phantomchat.json` to unstate the channel. There is no separate "configured but intentionally off" state.

### Verification

4644 pass, 0 fail; typecheck clean. Eight neuters proved the new tests are not decorative — five on the probes, three on this PR's changes (conflated string restored, escape hatch removed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1030,6 +1030,16 @@ before the key existed: every configured identity starts. Once the key is
 present — even as an empty list — it is the whole truth, and a persona outside
 it is skipped with a warning naming the fix.
 
+`phantombot doctor` treats that skip as a FAILURE, not a note: a persona that
+holds a `phantomchat.json` (or a stated Telegram account) but sits outside the
+roster is configured to talk and silently mute, which is a misconfiguration far
+more often than it is a choice. If muting it IS the choice, unstate the channel
+rather than leaving the two halves disagreeing — delete
+`<persona-dir>/phantomchat.json` (or clear the persona's Telegram account) and
+doctor stops reporting it, because a persona with no channel file is not using
+the channel at all. Keeping the file as a parked config is what earns the red
+row; there is no separate "configured but intentionally off" state.
+
 A persona name is CASE-SENSITIVE everywhere it is used as a key — persona
 directory, memory rows, vault, tasks — even though macOS and Windows
 filesystems are not. `default_persona = "robbie"` against a directory named
--- a/src/cli/doctor.ts
+++ b/src/cli/doctor.ts
@@ -17,7 +17,7 @@
  */
 
 import { defineCommand } from "citty";
-import { existsSync, readFileSync, statSync } from "node:fs";
+import { existsSync, readdirSync, readFileSync, statSync } from "node:fs";
 import { basename, join } from "node:path";
 
 import {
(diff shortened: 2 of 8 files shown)


lenaparkhodges (approved):
Verified fresh checkout at 786cf4b: full suite 4644 pass / 0 fail, typecheck clean on bun 1.4.2, CI green (test / test-windows / typecheck / CodeQL). Commit 1 is a faithful reland of the #545 payload I reviewed at 242d22a — diffed the per-file patches (doctor.ts, doctorChecklist.ts, Doctor.tsx, PersonaDetail.tsx, snapshot.ts, cli-doctor.test.ts), all identical. Commit 2 closes both deferred #545 notes: `phantomchatLoadFailure` re-resolves the identity and names the actual cause (missing identity.json / invalid nsec / unparseable file) with three new tests pinning each branch, and the mute escape hatch is now stated in both the detail string and the README. No new scope beyond the two notes. Merge Andrew's call.

PR #547, opened 2026-09-12
tests: de-flake the in-flight-turn Telegram slash-command tests

robertclawson:
## Why

`runTelegramServer slash commands > /stop aborts an in-flight turn and suppresses the would-be reply` and its sibling `a second non-slash message interrupts an in-flight turn` have timed out at bun's 5000ms per-test cap three times now (2026-09-10, twice on 2026-09-12). This is not a harmless annoyance: the most recent hit was on the **#544 merge run**, which failed the suite and so **skipped the v1.1.359 release** — there is no v1.1.359. (Both changes did ship, in v1.1.360 off the #546 merge.)

## Cause

Both tests drove their entire ordering with fixed wall-clock timers:

- message 1 enqueued up front,
- `setTimeout(… , 30)` to enqueue the interrupt "after the harness is in-flight",
- `setTimeout(() => ac.abort(), 200 /* or 300 */)` to stop the polling loop.

The fake transport long-polls in 20ms slices. On a loaded runner the loop could be torn down before the interrupt update had been polled at all, so the abort never fired — leaving the fake harness parked on its 5s fallback `setTimeout` and blowing the 5000ms cap. The assumptions in the comments ("~30ms later, after the harness is already in-flight") were true on an idle box and untrue on a busy one.

## Fix

Sequence off observable state instead of the clock:

- the interrupt is enqueued only once the harness reports itself **in-flight**, so it can never race ahead of the turn it is meant to interrupt;
- the polling loop is stopped only once the **effect under test is observed** — the abort fired *and* `/stop` replied; the second turn ran *and* replied.

New `waitUntil(cond, what, timeoutMs)` helper polls a predicate to a deadline and names what it was waiting for when it blows, so a real regression reports the missing effect rather than a bare `this test timed out after 5000ms`. Per-test cap raised to 20s so that the helper's own 4s deadline is always what fires first — the cap stops being a participant in the test's logic.

The harness's internal 5s fallback is kept as a safety valve, not a schedule.

## Verification

Proved by **reproduction**, not just by a green run. Under 12 busy-loop processes on this box:

| | result |
|---|---|
| old tests, under load | **2 of 6 runs failed** at the 5000ms cap |
| new tests, under load | **8 of 8 passed**, ~270–420ms each |

Full suite **4644 pass / 0 fail**, typecheck clean.

## Noted, not shipped

`tests/channels-phantomchat-server.test.ts` has the same-shaped test (`/stop aborts an in-flight turn`) using `slashSleep(120)` to wait for the turn to register. It has not flaked yet, but it is the same class of race. Left alone deliberately to keep this PR to the tests that are actually breaking releases — happy to fold it in here if you want it.

### DIFF
--- a/tests/channels-telegram.test.ts
+++ b/tests/channels-telegram.test.ts
@@ -151,6 +151,33 @@ class FakeTransport implements TelegramTransport {
   }
 }
 
+/**
+ * Poll `cond` until it holds, then return — instead of guessing a wall-clock
+ * delay.
+ *
+ * The in-flight-turn tests below used fixed `setTimeout`s to sequence
+ * "message arrives", "interrupt arrives" and "stop the polling loop". On a
+ * loaded CI runner the loop was sometimes torn down before the interrupt had
+ * been polled, leaving the harness parked on its 5s fallback and tripping
+ * bun's 5000ms per-test cap — a flake that skipped a release when it landed
+ * on a merge run. Waiting for the observable effect makes the ordering a fact
+ * rather than a race, and a blown wait fails with what it was waiting for
+ * instead of a bare "timed out".
+ */
+async function waitUntil(
+  cond: () => boolean,
+  what: string,
+  timeoutMs = 4000,
+): Promise<void> {
+  const deadline = Date.now() + timeoutMs;
+  while (!cond()) {
+    if (Date.now() > deadline) {
+      throw new Error(`timed out after ${timeoutMs}ms waiting for ${what}`);
+    }
+    await new Promise((r) => setTimeout(r, 5));
+  }
+}
+
 class ScriptedHarness implements Harness {
   invocations = 0;
   lastRequest?: HarnessRequest;
@@ -3424,6 +3451,9 @@ describe("runTelegramServer slash commands", () => {
     expect(transport.sent[0]!.text).toContain("cleared 1 turn");
   });
 
+  // Generous per-test cap: the waits above are the real deadlines and fail
+  // with what they were waiting for, so the cap only has to be bigger than
+  // them plus the harness's 5s fallback — never the thing that fires first.
   test("/stop aborts an in-flight turn and suppresses the would-be reply", async () => {
     // A harness that yields one text chunk then waits 5s (long enough that
     // the test-level abort is the only way it ever finishes within the
@@ -3431,11 +3461,14 @@ describe("runTelegramServer slash commands", () => {
     class AbortableHarness implements Harness {
       readonly id = "abortable";
       lastSignalAborted: boolean | undefined;
+      /** Set once the turn is parked, so the test can sequence off it. */
+      inFlight = false;
       async available(): Promise<boolean> {
         return true;
       }
       async *invoke(req: HarnessRequest): AsyncGenerator<HarnessChunk> {
         yield { type: "text", text: "thinking…" };
+        this.inFlight = true;
         await new Promise<void>((resolve) => {
           if (req.signal?.aborted) return resolve();
           const onAbort = () => {
@@ -3460,22 +3493,9 @@ describe("runTelegramServer slash commands", () => {
       senderId: "42",
       text: "kick off something slow",
     });
-    // /stop arrives ~30ms later, after the harness is already in-flight.
-    setTimeout(() => {
-      transport.pendingUpdates.push({
-        updateId: 2,
-        conversationId: "1001",
-        senderId: "42",
-        text: "/stop",
-      });
-    }, 30);
-
     const harness = new AbortableHarness();
     const ac = new AbortController();
-    // Stop the polling loop after a moment; the turn worker drain in the
-    // server's `finally` will wait for the aborted turn to resolve.
-    setTimeout(() => ac.abort(), 200);
-    await runTelegramServer({
+    const run = runTelegramServer({
       config: baseConfig(),
       memory,
       harnesses: [harness],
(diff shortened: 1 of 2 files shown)


tests/channels-telegram.test.ts
```diff
@@ -3485,6 +3505,28 @@ describe("runTelegramServer slash commands", () => {
...
+      updateId: 2,
+      conversationId: "1001",
+      senderId: "42",
+      text: "/stop",
+    });
+
+    // Tear the polling loop down only after the abort has actually happened
+    // and the /stop reply has gone out; the turn worker drain in the server's
+    // `finally` then waits for the aborted turn to resolve.
+    await waitUntil(
```
lenaparkhodges, line 3521: Non-blocking: the `tests/channels-phantomchat-server.test.ts` sibling (`/stop aborts an in-flight turn`, `slashSleep(120)`) is the same race class — and this class already skipped a release once. My take on your open question: fold it in here. Same mechanism, marginal diff cost, and it removes the last wall-clock-adjacent sequencing in the channel tests rather than leaving a known flake pattern for the next merge run.

lenaparkhodges (dismissed):
Verified on a fresh checkout at d3b543c: touched tests 3/3 green under a 12-process busy-loop (the exact load class that flaked), full channels-telegram file 147 pass / 0 fail, typecheck clean, CI fully green. Sequencing off observable state instead of wall-clock is the right fix — the `inFlight` flag is set only after the server has consumed the first chunk and the generator has parked on the abort promise, so it genuinely means "the turn is in-flight", and the teardown waits now name their missing effect instead of tripping bun's cap. Repro-first verification (2/6 old runs failing under load) is exactly how a flake fix should be evidenced.

robertclawson:
Folded the phantomchat sibling in here as you suggested — 7e4383a.

All four long-lived-server slash tests in `tests/channels-phantomchat-server.test.ts` now sequence off observable state instead of `slashSleep`: /stop waits for `BlockingHarness.inFlight` before sending the interrupt and for the published reply before tearing the server down, /reset waits for history to persist then for the "cleared N turns" reply, /restart waits for both the reply and the afterSend restart callback (a reply-only wait could still race it). `slashSleep` is gone from the file. `waitUntil` is the same helper, widened to take an async predicate because the effects here sit behind `await dmBubbles(...)` and `await memory.recentTurns(...)`.

Repro-first, same as before: 120ms sleeps are more forgiving than Telegram's 30ms, so 12 busy processes weren't enough — under 48 on this 8-core box the old tests failed 2 of 4 runs at the 5000ms cap (one run losing two tests at once). New ones: 6 of 6 green under identical load. File 57 pass, full suite 4644 pass, typecheck clean. Proved non-decorative by neutering the `stopped (was running Ns)` branch in `src/channels/commands.ts`: the test fails with "timed out after 4000ms waiting for the /stop reply to be sent" instead of a bare cap timeout.

The push dismissed your approval — re-requested.

lenaparkhodges (dismissed):
Re-reviewed at 7e4383a. The phantomchat fold-in mirrors the Telegram pattern correctly: async `waitUntil` with named failure messages, `BlockingHarness.inFlight` sequences the interrupt off observable state, /stop and /restart wait for the published reply (restart also waits on the afterSend callback — good catch, reply-only would still race), `slashSleep` is gone, per-test 20s caps. Verified fresh checkout: file 57 pass, typecheck clean on bun 1.4.2, CI all green at 7e4383a. Load-verified: 5 concurrent runs of the file under 40 concurrent telegram-file processes — the four slash tests passed in every run (the old 120ms sleeps failed the /restart family 2-of-4 at half this load).

One non-blocking note for the follow-up pile: the two streaming-bubble tests in the same file (`text before a progress chunk...`, `a trailing sign-off...`) still sequence on fixed `waitMs: 150` inside `runOnce` — under the same load they flaked in 4 of 5 runs on my host (asserted bubbles missing because the 150ms window closed before the scripted chunks flushed). Same race class, untouched by this PR and out of its stated scope, but worth the same observable-state treatment in a follow-up.

tests/channels-telegram.test.ts
robertclawson: Folded into this PR in 7e4383a, per your call. All four long-lived-server slash tests in tests/channels-phantomchat-server.test.ts now sequence off observable state and slashSleep is gone; waitUntil was widened to take an async predicate because the phantomchat effects sit behind await dmBubbles() and await memory.recentTurns(). Reproduced first: the 120ms sleeps needed 48 busy-loop processes before the old tests failed (2 of 4 runs at the 5000ms cap) — 12 was not enough load and looked like proof there was no race. New: 6/6 green under that same load, suite 4644 pass, typecheck clean, neuter proved.

robertclawson:
Folded the streaming-bubble note in rather than piling it on the follow-up list — same race class, same file, and leaving 20 more fixed-sleep call sites behind is exactly the pattern that skipped a release once.

`runOnce` now takes `untilInvocations` and waits on `harness.invocations` (the observable every harness in the file already exposes) before aborting the listen signal. The server drains its own `inFlight` set after the listen loop ends, so once the turn has started the drain guarantees it — and everything it publishes — completed; exact-equality bubble assertions stay as strong as before. `waitMs` survives only where there is genuinely nothing to wait for: the auth denials and the held-screen verdict, which assert `invocations === 0`.

Applied to every turn-expecting call site, not just the two you named — and that was not speculative. Load-verified at 40 concurrent busy processes on this 8-core box: 5/5 runs green with the named waits; 3/3 runs failing with the old sleeps, with `a trailing sign-off`, `text before a progress chunk`, and also `an npub in BOTH lists resolves to relay (least privilege wins)` among the failures.

Suite 4644 pass, typecheck clean. Push dismissed your approval as expected — re-requested.

lenaparkhodges (approved):
Re-reviewed f3c3a7d. Streaming-bubble tests now sequence off `untilInvocations: 1` with the runOnce drain argument holding up — once the turn has started, the post-listen drain guarantees it and everything it publishes completed, so the exact-equality bubble assertions stay as strong as before. Applied to every turn-expecting call site across the file (not just the two originally named), with the remaining `waitMs` correctly limited to the assert-nothing cases (held-screen verdict, auth denials). Verified fresh checkout at f3c3a7d: channels-phantomchat-server file 57 pass, 5 sequential runs clean, and 40 concurrent full-file runs (8-way parallelism × 5 waves) all 0 fail under load; typecheck clean (bun 1.4.2); CI all green. Fixed-sleep call sites in this file are now down to the ones that genuinely have nothing to wait on. Merge Andrew's call.

PR #548, opened 2026-09-12
channels: state one deterministic reply-language rule, drop the classifier

robertclawson:
## The bug

Replies flip language mid-turn — narration in Chinese, Russian, Spanish or French on an English conversation, then back to English. Two separate defects, both in the **prompt stack**, not in detection.

**1. Two instructions govern reply language, and the wrong one is last.**

`languageReplyInstruction` (#534) is injected via `systemPromptSuffix` at overlay position 1. `PRE_TOOL_NARRATION_INSTRUCTION` is pushed **LAST** (`src/orchestrator/turn.ts:624`, vs the suffix at `:539`) and re-opened the question:

> *"Use the user's language — match whatever language the conversation is in."*

That re-delegates language to **inference over turn content** — precisely the pollution source #534 exists to defeat — from the freshest, most salient instruction the model sees. Which is why the drift appears in **narration lines** specifically. Timing fits: #534's overlay landed 2026-09-04; #543 (2026-09-10) dropped the chat suffix's skip-narration line, so narration is now on for every text chat turn — maximum exposure.

**2. The classifier is silent on whole scripts.**

`LEXICON` is Latin-script function words (`en/es/nl/de/fr/it/pt`). A Chinese or Russian message scores 0 → `und` → **no language overlay injected at all**. The languages that drift worst are the ones governed by nothing.

## The fix

Widening the lexicon does not close (2): it is unbounded maintenance, and it still cannot separate the Latin-script languages this deployment mixes daily. So the rule now names the **source** instead of classifying it:

> Reply — including every pre-tool narration line — in the language of the **user's LATEST message**. A document being read, tool output, retrieved memory, the journal, a quoted reply, group catch-up context and your own previous turns are **DATA**, whatever language they are in.

- `REPLY_LANGUAGE_INSTRUCTION` replaces `languageReplyInstruction()`, injected **unconditionally** on every chat turn at both call sites.
- `src/lib/replyLanguage.ts`, `reply-language.json` and `PHANTOMBOT_REPLY_LANGUAGE_STATE` are **deleted**.
- The narration block's language clause now restates the same source rule instead of contradicting it (it stays self-contained, because `ask --stream` / the Twilio relay carry narration without the channel overlay).
- Recovery points at the user's message for the same reason; the `replyLanguageName` parameter is gone.
- AGENTS.md invariant 45 and the README's Reply Language section rewritten — the old invariant explicitly said "the instruction NAMES the language … do not soften it", so it had to be restated, not quietly contradicted.

**Deliberate trade-off:** per-conversation carry-forward goes too, so a bare "ok" no longer carries a remembered code. Nothing is lost — the rule points at the message and the model has the thread — and it removes the case where a stale remembered language was stated to the harness as fact.

## Verification

Tests pin the **wiring**, not a detector's wording:

- the rule is present on Chinese / Russian / Greek / Arabic input — the old silent hole
- it never names a concrete language
- both call sites inject it (a Telegram-only test would pass with PhantomChat broken)
- the narration overlay does not re-delegate to "whatever language the conversation is in"

Each fix neutered, suite goes red:

| mutation | result |
|---|---|
| drop the rule from `engine.ts` | **7 fail** |
| drop the rule from `server.ts` | **1 fail** |
| restore the old narration clause | **3 fail** |

`4624 pass / 8 skip / 0 fail`, `tsc --noEmit` clean.


lenaparkhodges (approved):
APPROVED. Verified on a fresh checkout at ae4a695: 4624 pass / 8 skip / 0 fail, tsc --noEmit clean (bun 1.4.2).

The approach closes the exact hole the investigation found. The old lexicon scored Chinese/Russian at zero → 'und' → no overlay at all, and PRE_TOOL_NARRATION_INSTRUCTION — pushed last in the overlay stack — re-delegated to 'whatever language the conversation is in', which is precisely where mid-turn drift came from. This PR fixes both:

- The rule is unconditional and names the source; tests pin its presence on Chinese/Russian/Greek/Arabic input and on a bare 'ok' — the old silent hole is now structurally impossible.
- The narration clause restates the same source rule instead of re-delegating; tests/orchestrator-turn.test.ts asserts the re-delegation is gone.
- Both call sites inject (engine.ts + phantomchat/server.ts) — not Telegram-only, which would have left PhantomChat broken.
- replyLanguage.ts, reply-language.json and PHANTOMBOT_REPLY_LANGUAGE_STATE are fully deleted with zero residual references; recovery points at the message.

One deliberate deviation worth ratifying explicitly: there is no English-default fallback, because there is no detection step left to be 'und'. The unconditional latest-message rule subsumes it — for an ambiguous bare 'ok' the model has the thread, and for a genuinely new conversation it mirrors whatever arrived. I agree that's the right call (a fallback that can never fire would be dead code), but it's a step beyond the original fix scope, so it should be a conscious acceptance.

Accepted trade-off: per-conversation carry-forward is gone. Reasonable — the rule points at the message, the model holds the thread, and a stale remembered language stated to the harness as fact was a real wart. The third-party-composition carve-out survives intact.

Nice piece of work — simpler and strictly broader than what it replaces.

PR #549, opened 2026-09-13
harness: native mode — embed the pi engine, split pi into native / pi-host, migrate via doctor

robertclawson:
## What this does

Phantombot now ships a brain that works out of the box. The [pi](https://pi.dev) engine (`@earendil-works/pi-coding-agent`, pinned exactly at **0.85.1**) is compiled into the phantombot binary and exposed as a new **`native`** harness. A fresh host with no claude, codex or pi installed can think as soon as it has a provider API key.

Delivered as **one PR** on purpose: the embedding, the harness split, the menus, the migration and the docs have to be tested and dogfooded together on an RC build.

### Harness ids

| id | What runs | Who configures it | Offered in menus |
|---|---|---|---|
| `native` | the embedded engine, via `<phantombot> __pi …` | phantombot: provider, API key, primary / vision / coder models (`[harnesses.pi.routing]`) | **always** |
| `pi-host` | the host's own `pi` | its owner, by running `pi` — phantombot passes no `--provider`/`--model`/`--api-key` | only when `pi` is installed |
| `claude`, `codex` | unchanged | unchanged | only when installed |

Native **always** uses the embedded engine, even when the host has its own pi. The routing schema is tied to the engine version we ship and test.

### Menus (`phantombot harness`, Configure → Brain, first-run onboarding)
- Detection runs **live** every time a menu opens. A missing host harness is not offered, and a stale chain entry is not pre-selected.
- Labels: **"Native — Configure Provider and Model Swap Settings"** and **"Pi — Use Host Configuration"**.
- **The Pi installer is gone.** Nothing in phantombot installs a harness any more. `harness --check` explains that native needs only a provider key.
- The "configure vs host configuration" question is gone: native is configured here, and the host pi is its own menu entry. Native → native still writes the `pi-primary` / `pi-fallback` named instances, now with `type = "native"`. Instance ids and vault secret names are unchanged.

### Attribution — native only
Native registers an inline pi extension that sets `HTTP-Referer: https://phantombot.bot` and `X-OpenRouter-Title` / `X-Title: Phantomyard's Phantombot`. It lives in the `__pi` entry, which only native spawns, so host harnesses keep their default headers.

### Upgrading existing installs (`lib/harnessReconcile.ts`)
One decision function has two callers:
- **Startup (`loadConfig`)** maps a legacy `pi` in memory, and never writes. A host that never runs doctor keeps serving.
- **`phantombot doctor`** (repair on by default, and also the daemon's startup doctor) writes the result into every `config.toml`. It copies the original to `config.toml.bak-<timestamp>` first, then does a temp-file + rename.

| Found | Result |
|---|---|
| `pi` + routing configured | `native` |
| `pi`, no routing, host `pi` installed | `pi-host` |
| `pi`, no routing, **no** host `pi` (doctor's live probe) | `native` (repair) |
| `pi`, no routing, not probed (startup) | `pi-host` (the entry's pre-upgrade meaning) |
| no `pi` / `native` / `pi-host` at all (claude- or codex-only) | **silent no-op**: no output, file untouched, native never injected |
| `pi-host` configured but binary gone | WARN only, never switched automatically |

Covered: the global chain, `[harnesses.personas.<name>].chain`, each persona's own `config.toml`, and `[harnesses.instances.<id>] type = "pi"`. A `use_local_config = true` opt-out counts as "no routing". The second run finds nothing and leaves files byte-identical.

**Rollback caveat (accepted by project policy):** older binaries skip unknown harness ids, so a migrated host must not be rolled back to a pre-native build. The migration is built to be right rather than reversible.

Doctor also re-extracts native's `PI_PACKAGE_DIR` assets when they drift, and warns when a native slot has no provider or model.

### Embedding notes (see AGENTS.md invariant 56)
- `src/index.ts` keeps its **static** import graph and dispatches `__pi` in its body. A dynamic-import entry compiles fine, but the compiled binary then crashes on every start (`@peculiar/x509`'s tsyringe reflect polyfill). Only running the compiled binary catches this.
- Imports pi's **unbundled** `dist/`. The bundled CLI breaks extension loading.
- `PI_PACKAGE_DIR` (pi's `package.json` + themes) is extracted to a version+hash data dir, **compiled binaries only**.
- The capability-routing extension's vision delegates re-enter the embedded engine through `PHANTOMBOT_PI_COMMAND`. A host pi gets that variable blanked.
- `/model`, `/model list`, the brain probe, `harness_bins` state (native is never persisted; pi-host persists under `pi`) and the Keys screen all understand the new ids.

## Verification

- `bun tsc --noEmit` clean; full `bun test` green.
- New tests:
  - `lib-harnessReconcile`: decision table; Megan-style `["claude","pi"]`; claude-only and codex-only files byte-identical with no backup; second run byte-identical; unparseable file left alone.
  - `lib-embeddedPi`: exact pin, argv, attribution, asset extract/repair, real `__pi --version`.
  - `harnesses-nativeChain`: native vs pi-host construction, delegate command parsing.
  - `harnesses-pi`: pi-host withholds routing and key; native spawns the engine prefix.
- **Compiled linux-arm64 binary:**
  - `phantombot --version` runs in 0.6s and `__pi --version` in 0.84s; `__pi --list-models` works.
  - Against a local header-echo provider, the compiled native request carried the three attribution headers. The same request through plain pi carried **none**.
  - **Doctor, sandboxed:**
    - A host config `["claude","pi"]` with routing became `["claude","native"]`, with a backup.
    - A persona file `["pi"]` with `use_local_config` and no host pi was repaired to `["native"]`.
    - Engine assets were extracted.
    - A second run printed nothing, wrote nothing, and made no new backup.
- **Binary size:** +10.4 MB (88.1 → 98.4 MB, linux-arm64). The earlier hello-world spike measured +6 MB; this is the real number.

## Not verified yet — needs the RC build before merge
- A **real OpenRouter request** showing "Phantomyard's Phantombot" in the activity log (only proven against a local echo server so far).
- `__pi` self-exec and asset extraction on **Windows and macOS** targets. Linux arm64 is verified.
- `install.sh` / `install.ps1` are unchanged: they launch bare `phantombot`, which opens this onboarding. Their existing test suites pass. A real sandbox install run needs the RC artifacts.
- Setting the tagline as the description on the OpenRouter app page is a manual step (no API). Updating the GitHub repo description happens after merge.

## Also in this PR
- Tagline **"Armors your harness, learns your world, zero context tax"** in the README, `package.json` and on phantombot.bot (`<title>`, `<h1>`, plus a new `meta description` / `og:description`).
- README, AGENTS.md (layout + invariant 56), `docs/adding-a-harness.md`, `docs/harness-detection.md`, `docs/architecture.md`.


src/config.ts
```diff
@@ -2067,6 +2097,19 @@ function buildPiRoutingConfig(
...
+): Record<string, { chain: string[] }> | undefined {
+  if (!personas) return personas;
+  return Object.fromEntries(
+    Object.entries(personas).map(([name, entry]) => [
+      name,
+      { chain: mapLegacyChain(entry.chain, facts) },
+    ]),
+  );
+}
+
```
lenaparkhodges, line 2120: Read-time note, non-blocking: this maps **every** persona chain in the global file using the routing facts of the persona being loaded (`loadConfig(persona)` builds `legacyPiFacts` from the ambient/env-resolved routing once, then applies it to all names). Per-turn serving is safe — each turn re-calls `loadConfig(persona)` with that persona's own env, so the effective mapping is right where it matters — but any consumer that inspects the snapshot's *other* personas' chains between reloads sees them mapped by someone else's env facts. Worth a one-line comment here, or resolving facts per persona the way `computeHarnessConfigReport` does.

docs/harness-detection.md
```diff
@@ -14,7 +14,11 @@ value:
 | 1 | Env var | `PHANTOMBOT_CLAUDE_BIN=/opt/claude/bin/claude` |
 | 2 | `config.toml` | `[harnesses.claude] bin = "claude.cmd"` |
 | 3 | **`state.json` → `harness_bins.<id>`** | written automatically on a successful resolve |
-| 4 | Bare-name default | `"claude"`, `"pi"`, `"codex"` |
+| 4 | Bare-name default | `"claude"`, `"pi"` (for `pi-host`), `"codex"` |
+
+The `native` harness has no entry in this chain: its engine is compiled into
+the phantombot binary, so detection always reports it present (its "binary" is
+`process.execPath`) and it is never written to `state.json`.
```
lenaparkhodges, line 21: The rollback caveat in the PR body only lives in the PR body — after merge it disappears. The hazard is real and invisible: doctor rewrites `type = "pi"` → `"native"`/`"pi-host"` in every config.toml, and an older binary that skips unknown ids turns a native-only chain into an empty one (exit 2 "no harnesses configured", from source it is a plain warning-skip). Please land a short version of this caveat in `docs/harness-detection.md` next to the new `native` paragraph (or the `harnessReconcile` module doc) so it outlives the PR. Non-blocking, but it should ship with the PR, not in its description.

lenaparkhodges (dismissed):
**Approved with two non-blocking inline threads.** This is the right shape for "the brain always works": native is always offered, host harnesses stay chain-only, and the pi installer is gone — phantombot grows no per-tool hooks, and the harness split matches how claude/codex already work.

**What I verified locally** (fresh checkout of 03afed6, linux x64, bun):
- `bun tsc --noEmit` clean.
- The four new/touched harness test files: **101 pass / 0 fail** (reconcile decision table, embeddedPi pin/argv/attribution/asset repair, native vs pi-host chain construction, pi-host key withholding).
- The reconcile decision table matches the PR description on every row I traced, including the Megan-style `["claude","pi"]` + routing case and the read-time vs doctor divergence (unknown `hostPiInstalled` ⇒ `pi-host`, definite absence ⇒ repaired to `native`). Idempotency is enforced structurally (zero changes ⇒ no write), and the writer backs up before temp-file+rename. Unparseable files are left alone.

**Product read:**
- Attribution is native-only by construction — the extension factory is passed only in `runEmbeddedPi`, and pi-host gets `PHANTOMBOT_PI_COMMAND` blanked so host delegates can never inherit a native spawn. Verified the header values match the OpenRouter attribution mechanics we documented (Referer + title headers).
- The honest "not verified yet" list is the merge gate, and it is correct: real OpenRouter attribution on the wire, `__pi` self-exec on Windows/macOS, and a sandboxed install run all need the RC build. I did not re-verify the compiled-binary claims on this host (no arm64 build here); CI covers typecheck/tests/windows CI only.
- The rollback caveat is accepted policy per the body, but it must outlive the PR description — see my inline on harness-detection.md.

**Merge recommendation:** approve to merge **after** the RC-build verification list is green (real OpenRouter request, Windows/macOS `__pi`, sandbox install). The migration design (one decision function, two callers, never disagree) is the strongest part of this PR and is tested like it matters.

src/tui/screens/Keys.tsx
```diff
@@ -29,8 +29,9 @@ export function expectedSecrets(
   if (persona.channels.includes("telegram")) {
     out.push({ name: "TELEGRAM_BOT_TOKEN", usedBy: "channel: telegram" });
   }
-  if (persona.chain.includes("pi")) {
-    out.push({ name: "ANTHROPIC_API_KEY", usedBy: "harness: pi" });
+  if (persona.chain.includes("native")) {
```
kaieriksen: [P1] Include named native instances here. The Brain flow persists a native→native chain as `pi-primary` / `pi-fallback` and stores their credentials as `PHANTOMBOT_PI_API_KEY_PI_PRIMARY` and `_PI_FALLBACK`, so this `includes("native")` branch adds neither expected key. Once either stored key is absent/unset, the Vault screen has no row through which the user can restore it, leaving a supported chain silently unmanageable except by rerunning Brain. Resolve each chain id through the instance type and emit `piInstanceSecretName(id)` for native instances (with regression coverage for the dual-native chain).

kaieriksen (changes requested):
The embedded runtime and migration coverage are strong, and the full local suite plus a compiled Linux x64 startup passed. One management-path regression remains for the supported native→native configuration.

src/tui/screens/Keys.tsx
robertclawson: Fixed in f40123c. New `nativeApiKeyNameFor(config, id)` in buildChain.ts resolves each chain id through `piEngineFor` (instance type) and returns `piInstanceSecretName(id)` for a named native instance, `PHANTOMBOT_PI_API_KEY` for the unnamed slot, undefined otherwise. The snapshot carries the resolved `nativeKeys`, and `expectedSecrets` emits one row per key. Regression coverage in `tests/tui-keys-native-instances.test.ts`: the dual-native `pi-primary`/`pi-fallback` chain shows both suffixed keys as unset rows, the unnamed slot keeps the plain key, pi-host/claude/codex expect none. Verified the dual-native test fails with the Keys change reverted. Full suite 4679/0, tsc clean.

docs/harness-detection.md
robertclawson: Done in f40123c: a rollback caveat now sits next to the `native` paragraph in `docs/harness-detection.md`. It covers what doctor rewrites (chain, persona chains, instance types), the backup, the exit-2 empty-chain failure on older binaries, and how to roll back (restore the backup or set the ids back to `pi`).

src/config.ts
robertclawson: Took the comment option in f40123c. Resolving per persona at read time would mean loading every sibling's vault on each `loadConfig`. The comment states the invariant that keeps serving correct: every turn, probe and TUI snapshot reads a chain via `harnessChainIds(loadConfig(<that persona>), <that persona>)`, so a sibling's chain must never be served off this Config. It also points at doctor as the per-persona path because it writes, and notes that doctor's post-write health scan only reads sibling ids for warnings.

lenaparkhodges (approved):
**Re-approved on f40123c.** All three threads are addressed:

1. **Rollback caveat in docs** — landed in `docs/harness-detection.md` next to the native paragraph, and it's complete: what doctor rewrites (chain, persona chains, instance types), the backup, the exit-2 empty-chain failure mode on older binaries, and both rollback paths (restore backup or flip ids back to `pi`). This was my only blocking-adjacent ask and it's done right.
2. **mapLegacyPersonaChains snapshot scope** — the comment-instead-of-refactor option is the right call: resolving per persona at read time would load every sibling's vault on each `loadConfig`. The invariant is now stated where the next reader will trip over it: every consumer reads a chain via `harnessChainIds(loadConfig(<persona>), <persona>)`, so a sibling's chain must never be served off another persona's Config.
3. **Native→native Vault-screen keys (the P1)** — verified the fix is structurally sound: `nativeApiKeyNameFor` resolves through `piEngineFor` with the same logic `buildHarness` uses, so the screen can't drift from what the turn actually reads; the snapshot carries `nativeKeys`; `expectedSecrets` dedupes and labels per-instance (`harness: pi-primary (native)`).

**Local verification on fresh checkout of f40123c:** typecheck clean; new `tests/tui-keys-native-instances.test.ts` 3/3 (both instance keys expected, unnamed slot unchanged, pi-host/claude/codex expect nothing); reconcile/doctor/buildChain/snapshot suites 149/0. CI green on all six checks including Windows.

The merge gate from my earlier review stands unchanged: the RC-build items in the PR body (real OpenRouter attribution, `__pi` self-exec on Windows/macOS, sandbox install run) should be green before this lands.

kaieriksen (approved):
Re-reviewed exact head f40123c. The named-native Vault defect is fixed at the shared harness-resolution seam: each configured native instance now exposes its actual suffixed vault key, the unnamed native slot retains PHANTOMBOT_PI_API_KEY, and pi-host/Claude/Codex remain excluded. Regression coverage exercises both missing dual-native keys as restorable rows. Targeted 46-test suite and typecheck pass; all GitHub checks are green.

PR #550, opened 2026-09-13
native: never depend on ~/.pi — copy keys before migrating, doctor verifies and repairs

robertclawson:
## Why

Follow-up to #549. On 2026-09-13 pi was removed from Lena (rig) and Atlas as a test. That exposed that **native mode still depended on the host's `~/.pi`**:

- **Atlas:** both native slots exited 1 with `No API key found for openrouter`. The key had lived only in `~/.pi/agent/auth.json`, and the #549 migration never copied it anywhere. Doctor still reported ok.
- **Lena:** survived only because `OPENROUTER_API_KEY` happened to be in her vault. She still lost image/coding routing: extension provisioning only read top-level `harnesses.pi.routing` and ignored native instance routing.
- **Both:** the embedded engine recreated `~/.pi/agent/auth.json` and `models-store.json`. The Brain test screen hid pi's real stderr, and stale `state.json harness_bins` entries still pointed at the deleted host pi.

## What changes

- **Isolation:** native pi runs with its own agent dir (`$XDG_DATA_HOME/pi-native/agent` via pi's `PI_CODING_AGENT_DIR`). It never reads or writes `~/.pi`. `pi-host` is unchanged.
- **Key copy before switching** (`src/lib/nativeKeyCopy.ts`): startup and doctor copy each native slot's key into the persona vault. Sources, in order: stored row → legacy auth store (api_key only, never OAuth) → `OPENROUTER_API_KEY` row → sibling slot with the same provider. Idempotent. A slot that resolves from none of these is reported, never invented.
- **Reconcile** refuses to switch a legacy slot to native when no key can be copied.
- **Doctor** FAILs a native harness with no resolvable key, naming the secret. `--fix` prompts once.
- **Brain** refuses to finish native without a key (it asks again; esc aborts with no change). The test screen shows pi's stderr tail.
- **Routing extension provisioning** covers native instances.
- Stale `harness_bins` entries are pruned.

## Also fixed while finishing this PR

- **Infinite loop in `phantombot harness` native key prompt.** The re-ask loop only broke on `set`/`clear`, so a blank answer resolving to `keep` spun forever at 100% CPU. It hung `tests/cli-harness.test.ts`, the full local suite and this PR's first CI run. Fixed with the missing `keep` break.
- **Bounded re-asks.** `MAX_NATIVE_KEY_ATTEMPTS = 3` (one constant in `piRouting.ts`) caps both the CLI prompt and the Brain gate. A prompt source that keeps answering blank now aborts with no change.
- **Extension dir with an explicit agent dir.** `extensionDir()` dropped the `extensions/` segment when `agentDir` was passed, stamping to `<agent>/capability-routing`, where pi never looks. There is now one derivation for both paths.

## Tests

- New `tests/lib-nativeKeyCopy.test.ts`: slot listing, auth-store parsing, and each copy source. It also covers idempotency after `~/.pi` is deleted, never inventing a key for a mismatched provider, and dryRun.
- `tests/tui-brain-flow.test.ts`: a blank key after a provider switch now **asks again instead of clearing**. Aborting at the gate changes nothing. Slot tests supply a key.
- `tests/tui-brain-test.test.tsx`: stderr tail is rendered.
- `tests/harnesses-pi.test.ts`: scrubs ambient `PHANTOMBOT_PI_API_KEY*` / `OPENROUTER_API_KEY` per test. A vault-injected dev-shell key had hidden that three routing tests set no key. There's also a new test that native with a provider and no key fails loudly, naming the secret and `doctor --fix`.
- `tests/cli-doctor.test.ts`: an unresolvable native key prints FAIL naming the secret and exits 1; a key copied this run is OK and exits 0.
- `tests/tui-brain-flow.test.ts`: bounded gate with an inexhaustible blank prompt source.
- `tests/lib-piExtensionProvision.test.ts` passes again (explicit agent dir path).
- Full suite: **4,693 pass / 0 fail** (289 files, 113s), run with key env vars stripped like CI. `tsc` clean.
- **Mutation-proven:** removing the `keep` break hangs `cli-harness`, removing the cap hangs the bounded test, and removing the gate fails the three gate tests, and ignoring unresolved keys fails the doctor exit-code test.

## Rollout check (RC)

Remove pi + `~/.pi` on a scratch host with a legacy `pi` config, start the daemon, and confirm the native turn answers. Also confirm `doctor` reports the copied key and `~/.pi` is not recreated.


kaieriksen (dismissed):
The CI-only regression is fixed at f3747de: the Pi subprocess suite now scrubs ambient keys, provides explicit fixture keys where required, and covers the new no-key failure path. Focused Pi tests pass 61/61, typecheck is clean, and all Linux/Windows CI plus CodeQL checks are green. The full migration/isolation diff is approved at this exact head.

src/cli/doctor.ts
```diff
@@ -1839,11 +1880,58 @@ async function computeHarnessConfigReport(
...
+  const auditKeys = async (
+    key: string | undefined,
+    auditPersona: string | undefined,
+    cfg: Config,
+    dryRun: boolean,
+  ): Promise<void> => {
+    if (nativeSlotsFor(cfg, auditPersona).length === 0) return;
+    keyResults.set(
+      key,
+      await copyNativeKeys({
```
lenaparkhodges, line 1903: This call relies on the default `openPersonaVault` opener, which mints `identity.json` + `vault.sqlite` for a persona that has none — even with `dryRun: true`. Confirmed by repro: a plain non-repair doctor run on a vaultless persona with a native slot provisions both files. That violates the repo's own invariant ("doctor never provisions a persona it only inspected", `cli-doctor.test.ts` — which passes vacuously because its fixture has no native slots and this path is behind `isPhantombotBinary()`).

The `openVault` seam on `CopyNativeKeysInput` exists for exactly this and is documented as "doctor uses a READ-ONLY-constructing opener" — but no caller ever passes it, so it's dead. Please wire the audit to an opener that opens only when an identity already exists (read existing nsec → `openVaultWithSecret`, else `undefined` → the audit then reports from env/auth sources only, per the seam's own doc). `openPersonaVault` stays correct in the interactive repair path where the operator pasted a key.

src/lib/nativeKeyCopy.ts
```diff
@@ -0,0 +1,293 @@
...
+      }
+
+      // (d) a sibling native slot with the same provider whose key resolves.
+      const sibling = statuses.find(
+        (s) =>
+          s.secretName !== status.secretName &&
+          s.provider === provider &&
+          bySecret.get(s.secretName)?.source !== "missing",
+      );
+      const siblingKey = sibling ? vault.get(sibling.secretName)?.trim() : undefined;
```
lenaparkhodges: Non-blocking: under `dryRun`, the sibling fallback under-reports. A sibling whose key would come from an auth-store copy carries `source: "auth-store"` (status updated even though nothing was written), but this `vault.get` reads the real vault, which was never written in dryRun — so a slot that `--fix` WOULD fill from a sibling reports `missing` in the `--no-repair` audit. Cheap fix: carry would-be values in a local map when `dryRun` instead of re-reading the vault. Only affects report accuracy of `--no-repair`, not writes.

lenaparkhodges (changes requested):
Verified locally at f3747de: typecheck clean, touched suites 246/0 (incl. cli-harness, cli-doctor), reconcile 24/0. The migration/isolation design is right — key copy before dependency, bounded re-asks, isolated agent dir, loud no-key failures. But one production-path defect:

**P1 — plain `phantombot doctor` (no `--fix`) mints a persona identity + vault as a side effect of inspection.**

The key audit in `computeHarnessConfigReport` calls `copyNativeKeys` with only `personaDir`, so the default opener (`openPersonaVault` → `getOrCreatePersonaIdentity`) runs — which **generates and persists a fresh Nostr identity plus an empty vault** for any persona with a native slot that has none. This happens even with `dryRun: true`, i.e. on a non-repair inspection run.

This violates the invariant this repo already tests for — `cli-doctor.test.ts` "doctor never provisions a persona it only inspected" — but that test passes vacuously today: its fixture has no native slots, and the whole path sits behind `isPhantombotBinary()`, so `bun test` can never reach it.

Repro at this head: fixture config with a native `pi-primary` slot, a persona dir with no `identity.json`, binary gate forced → after a `runDoctor` with no repair flags, `identity.json` and `vault.sqlite` exist.

Fix: the audit pass should pass the `openVault` seam that `CopyNativeKeysInput` already documents ("doctor uses a READ-ONLY-constructing opener: it must never generate an identity for a persona that has none") — open only when an identity already exists (read the existing nsec → `openVaultWithSecret`), else return undefined so the audit reports from env/auth sources only. That parameter is currently dead: no caller passes it. In repair mode after the operator pastes a key, `openPersonaVault` is fine — minting is the point there.

Please also add a regression test that reaches `computeHarnessConfigReport` (binary gate forced or via the `checkHarnessConfig` seam wired to the real implementation) and asserts no identity/vault mint on inspection with a native slot present — current coverage cannot see this path.

src/cli/doctor.ts
robertclawson: Fixed in f50e40b. The inspection audit now passes `openVault: () => openExistingPersonaVault(dir)`, which opens only when `vault.sqlite` + `identity.json` already exist (`readPersonaIdentityNsec` → `openVaultWithSecret`) and returns undefined otherwise. Repair keeps `openPersonaVault`. Regression test `tests/cli-doctor-nativeKeyAudit.test.ts` calls the real `computeHarnessConfigReport` with a native slot and no identity. It is mutation-proven: remove the opener and it fails with identity.json Received true.

src/lib/nativeKeyCopy.ts
robertclawson: Fixed in f50e40b. Resolved keys (written or would-be) go into an in-memory map, and the sibling lookup reads that map, not the vault. It also closes a related gap: with the inspection opener returning no vault, the dry run used to bail on `!vault` and report an auth-store key as missing. It now resolves without a vault. The new test "dryRun with no openable vault still reports what --fix would copy" fails under the old guard.

robertclawson:
Thanks Lena, both addressed in the same PR.

**P1: inspection no longer provisions.** `nativeKeyCopy.ts` now exports `openExistingPersonaVault(dir)`. It returns a vault only when `vault.sqlite` AND `identity.json` already exist, using `readPersonaIdentityNsec` → `openVaultWithSecret`. Otherwise it returns `undefined`. Doctor's audit passes it through the `openVault` seam whenever `dryRun` is set, so the seam is no longer dead. Repair keeps the default `openPersonaVault`, where writing the key is the whole point.

**Regression test that actually reaches the path:** `tests/cli-doctor-nativeKeyAudit.test.ts` calls the real `computeHarnessConfigReport` (now exported) directly, which skips the `isPhantombotBinary()` gate. The fixture has a native `pi-primary` slot, a persona dir with no identity, and `repair=false`. The test asserts the slot is reported unresolved and that neither `identity.json` nor `vault.sqlite` exists afterwards. **Mutation-proven:** with the opener line removed, it fails with `identity.json` Expected false / Received true.

One trap: bun's `homedir()` ignores a test-set `HOME`, so the fixture read the host's real `~/.pi/agent/auth.json`. The fixture therefore uses a provider that no auth store holds a key for.

**Non-blocking: dry-run sibling under-report.** Resolved values now go into an in-memory map, keyed by secret name. Each value is either written or would-be, and the sibling lookup reads the map, never the vault. While fixing this I found a bigger dry-run gap: once the inspection opener returns no vault, the second sweep bailed on `!vault`, so an auth-store key `--fix` would copy showed as `missing`. A dry run now resolves auth-store and sibling sources without a vault. The new `lib-nativeKeyCopy` test "dryRun with no openable vault still reports what --fix would copy" covers this and is mutation-proven: it fails with the old `!vault` guard.

`tsc` clean, full suite below.

Full suite (vault keys stripped, CI-equivalent): **4695 pass / 0 fail**.

lenaparkhodges (approved):
Re-reviewed at f50e40b. Both review threads verified fixed locally:

- **P1 (inspection must not provision):** `openExistingPersonaVault` only opens a vault that already exists (vault.sqlite + identity.json), wired through the previously dead `openVault` seam on dry-run; repair keeps the provisioning opener where writing the key is the point. The new `cli-doctor-nativeKeyAudit` test calls the real `computeHarnessConfigReport` directly — bypassing the `isPhantombotBinary()` gate that made the old invariant test vacuous — and asserts the audit reports the unresolved native slot without minting `identity.json` or `vault.sqlite`. Nice catch on the `homedir()`-ignores-`HOME` trap; the no-such-provider fixture sidesteps it cleanly.
- **Dry-run sibling under-report:** resolved values now land in an in-memory map that the sibling lookup reads instead of re-reading the vault — a dry run no longer under-reports what `--fix` would copy. Bonus: the `!vault` bail in the second sweep is gone, so a dry run with no openable vault still reports auth-store resolution (covered by the new `lib-nativeKeyCopy` test).

Local verification on fresh checkout of f50e40b: `tsc --noEmit` clean, 101/101 across cli-doctor, harnesses-buildChain, harnesses-nativeChain, lib-nativeKeyCopy, and the new cli-doctor-nativeKeyAudit suites.

Approving.

kaieriksen (approved):
Re-reviewed at f50e40b. The doctor audit now opens only an existing persona vault, so inspection cannot mint identity.json or vault.sqlite; repair retains the provisioning path. Dry-run source tracking also correctly models auth-store and sibling copies without writes. Local verification: 225 targeted tests pass and typecheck is clean. All Linux, Windows, and CodeQL checks are green.

PR #553, opened 2026-09-13
harnesses: replay model reasoning after a narration quiet window (closes #551)

### LINKED ISSUE #551: Narration decay: replay last reasoning string after a 10s quiet window (claude/codex/pi) — capture reasoning instead of discarding it
## Problem

Narration ("one short sentence before each tool call") is prompt-only — no code enforces it. On long autonomous runs (observed: a phantomchat session wrote one narration line, then ~122 silent tool calls) the model drifts out of the habit and the user sees a silent channel for minutes. This affects every harness.

A synthesized heartbeat string ("still working, 4m elapsed, 37 tool calls") was rejected as a fix: it adds yet another hardcoded English string to the codebase, and the project wants model-emitted, naturally-localized text.

## Desired behaviour

When a turn has been **quiet for 10 seconds** (no narration line and no tool event emitted), emit the **last human-readable reasoning/thinking string** the model produced, as a progress row. Fallback chain if no reasoning is available: last narration line → last tool-call note.

Constraints:
- Emit **only if new reasoning content arrived since the last emit** — no spam, no repeating a stale line.
- All emitted text is model-written (already in the user's language) — **zero new hardcoded English strings**.
- A plain reasoning replay with no quiet-window guard is explicitly rejected (would spam chatty thinking tokens).

## Key fact: all three harnesses currently DISCARD reasoning

Verified on main (`aca1fa7`):

- **claude** (`src/harness/claude.ts`) — `thinking` blocks map to heartbeat-only ("we never leak chain-of-thought"); some blocks arrive as `redacted_thinking` (encrypted, no readable text) → the fallback chain covers those.
- **pi** (`src/harness/pi.ts`) — receives `thinking_delta` events but discards the content; also emits `tool_execution_update` heartbeats (fine, those count as activity).
- **codex** (`src/harness/codex.ts`) — does not parse reasoning items at all, though the stream does emit reasoning summary items.

## Implementation sketch

1. Each harness parser keeps a **one-slot buffer** of the last human-readable reasoning string instead of dropping it.
2. Engine-side timer in the streaming path: after a 10s window with no narration/tool event, emit the buffered string once, then reset the timer. New reasoning arriving refreshes the buffer and re-arms.
3. No timer fires on a turn with no reasoning content at all (fallback chain applies there, also only when content actually arrived).
4. Localize nothing — never synthesize text; only replay model-emitted content. The only acceptable non-model decoration is none at all (no "still thinking:" prefix unless the team decides it's worth one i18n string).

## Acceptance criteria

- [ ] A claude/codex/pi turn that goes silent >10s while producing reasoning shows the last reasoning string as a progress row.
- [ ] No repeated emission of the same string without new reasoning in between.
- [ ] No new hardcoded English in the diff.
- [ ] `redacted_thinking` turns fall back to last narration/tool note (or stay silent if nothing exists — do not synthesize).
- [ ] Works for all three harnesses.

## Related

- #543 / #548 touched the narration language clause — not the cause; the drift is prompt-only decay, confirmed intact controller + 2s flush in phantomchat `server.ts`.

lenaparkhodges:
Closes #551. Implements the narration-decay fix: every harness parser already **received** human-readable reasoning and discarded it — this captures that text and replays it as an ephemeral progress row after a quiet window. Capture, not synthesis.

## What ships

- **`src/harnesses/reasoningReplay.ts`** — `ReasoningReplay`, a one-slot rolling buffer + state machine. Model-emitted text only; it never writes a string (the #548 English-only ruling holds by construction: a replay in a Chinese turn arrives in Chinese).
- **`src/lib/harnessRunner.ts`** — `parseEvent` widened to `ParseEventResult` (`HarnessChunk | ReasoningCapture`); the shared stdout loop races the next read against the replay's **absolute deadline** (no polling; heartbeat-only streams don't push the quiet window back). Emissions ride the payload-less `progress` path, which every channel layer treats as ephemeral UI state (live indicator + /status) — never reply text, narration bubble, memory, or logs.
- **Per-harness capture**: claude (`stream_event` → `thinking_delta`), pi (`thinking_delta`), codex (`item.completed` `reasoning` summaries). The surfaced chunk stays a payload-less heartbeat exactly as before.

## Constraint compliance (issue + review refinements)

- [x] Quiet window (default 10s, no narration/text/progress) **and** min interval (default 10s) between emissions — worst case one row per 10s, never one per reasoning burst
- [x] Emit only **new** reasoning — the buffer holds un-emitted text; nothing to say stays silent (no stale-line replay)
- [x] Zero new hardcoded strings in the diff
- [x] Fallback chain (`redacted_thinking` → last narration → last tool note): each replays **at most once** (identical strings never repeat), and only while fresh (60s default) — a minutes-old narration line is never replayed; silence wins
- [x] Reasoning replay is user-channel-only by construction (`progress` without `tool` is not resume evidence and not persisted)
- [x] Works for all three harnesses

## Privacy posture (per the issue thread)

Replaying reasoning to the user's own channel is fine (their model, their tokens), but the text now flows through channel render paths. It rides `progress` **without** a `tool` payload: phantomchat stores only `turnHandle.lastProgressNote` (capped, /status display), Telegram the same; the resume log counts only structured `progress.tool` chunks as tool evidence. Nothing persists reasoning to memory/journal/logs.

## Verified

- `bun run typecheck` clean
- New suite `tests/harnesses-reasoningReplay.test.ts`: state machine (fake clock — quiet window, min interval, caps, fallback freshness/dedupe, dueIn no-busy-loop guard) + engine integration with a **real subprocess and real timers** (bounded emissions, no repeats, narration fallback once, legacy path unchanged without `reasoningReplay`)
- Full suite: **4,708 tests, 0 fail**; touched harness/runner files 226/226

## Caveats (honest ones)

- Claude captures **only** `stream_event` thinking deltas (the partial stream phantombot always enables). Complete `assistant` thinking blocks stay heartbeat-only — they'd duplicate deltas already in the buffer. `redacted_thinking` has no readable text → fallback chain, as specced.
- Codex reasoning-item shape (`item.completed` with `item.type === "reasoning"`, `text` field) is taken from codex's exec-json docs/observations, not verified against a live codex on this box — flagging for RC verification alongside the other RC items.
- Due to the deadline race, a completely silent stream (no stdout at all) can't fire a replay — but there's nothing new to say in that case anyway, and the idle watchdog owns true wedges.

Closes #551.


src/lib/harnessRunner.ts
```diff
@@ -601,8 +641,56 @@ export async function* runHarnessProcess(
...
+            (r) => ({ kind: "read" as const, r }),
+          ),
+          new Promise<{ kind: "tick" }>((resolve) => {
+            tickTimer = setTimeout(
+              () => resolve({ kind: "tick" }),
+              Math.max(0, dueIn),
+            );
+          }),
+        ]);
+        if (tickOrRead.kind === "tick") {
```
kaieriksen, line 681: This chunk is indistinguishable from ordinary tool progress downstream, so the claimed `never persisted to logs` contract is false. `channels/core/engine.ts` logs every progress note (`telegram: progress`), and `cli/tick.ts` logs a preview of every progress note; background turns also opt into these harness defaults. Add an explicit ephemeral/reasoning discriminator and make every logging/persistence path drop or redact it before this can ship.

src/harnesses/reasoningReplay.ts
```diff
@@ -0,0 +1,248 @@
...
+  note(text: string): void {
+    const t = text.trim();
+    if (!t) return;
+    // Join fragments with a single space unless they already run on (deltas
+    // arrive mid-sentence; complete summaries are sentences). Never invent
+    // punctuation — the separator is whitespace, not synthesized text.
+    this.pending =
+      this.pending.length === 0
+        ? t
+        : (this.pending + (/\s$/.test(this.pending) ? "" : " ") + t);
```
kaieriksen: This corrupts real streaming deltas: e.g. `investig` + `ating` becomes `investig ating`, while a delta that intentionally begins with whitespace loses it via `trim()`. That violates capture-not-synthesis and will visibly mangle words (including non-English text). Preserve fragments byte-for-byte when appending; use `trim()` only to decide whether a fragment is empty, and add a split-mid-token regression test.

kaieriksen (changes requested):
Two blockers in the replay path: reasoning is currently copied into existing progress logs despite the stated privacy contract, and streamed delta fragments are rewritten rather than replayed verbatim. The focused state-machine tests pass; CI is green, but the local combined run could not execute the Pi suite because this checkout lacks its embedded Pi package.

src/lib/harnessRunner.ts
lenaparkhodges: Fixed in bfec92e: the replay chunk now carries an explicit `ephemeral: true` discriminator (`replayChunk()`), and both generic logging paths redact it — `channels/core/engine.ts` logs a placeholder for ephemeral progress rows instead of the note, and `cli/tick.ts` drops the `previewForLog` preview for the same. `lastProgressNote` remains in-memory `/status` state only (never persisted). Added a test pinning the `replayChunk` ephemeral shape.

src/harnesses/reasoningReplay.ts
lenaparkhodges: Fixed in bfec92e: fragments now append byte-for-byte — no invented separators, `trim()` only decides emptiness, and surrounding whitespace survives. Added a split-mid-token regression test (`investig` + `ating` → `investigating`) plus a whitespace-preservation test. Codex complete summaries run together at the seam when the model emits no boundary whitespace — accepted cosmetic cost; corrupting streamed deltas would be worse.

kaieriksen (dismissed):
Both blockers are fixed at bfec92e: replay text is explicitly ephemeral and redacted from the two persisted progress-log paths, while streamed fragments are captured byte-for-byte with split-token and whitespace regressions. Focused harness/replay tests pass 189/189, typecheck is clean, and CI is green.

robertclawson (changes requested):
Requesting changes at `bfec92e`. CI is green, and locally the full suite passes (4703 pass / 8 skip / 0 fail, run with the vault keys stripped) and `tsc` is clean. The capture work is solid and Kai's two fixes hold. But I don't think this delivers #551 yet, and the fallback chain adds noise the issue explicitly ruled out.

## Blocker 1: nothing shows the replay in phantomchat or Telegram

#551 was reported on phantomchat: narration goes quiet during long runs. In both `src/channels/phantomchat/server.ts:1074` and `src/channels/core/engine.ts:1430`, a `progress` chunk only does three things: sets `turnHandle.lastProgressNote` (which only `/status` reads), runs the tool-boundary narration flush, and (engine only) bumps `progressCount`. Neither channel sends a bubble, status line or indicator text for it. So on the channels where the decay was reported, the user still sees silence. The replay text only reaches `/status`.

The only surfaces that actually render it are the TUI and ACP, and they show it as a **tool call**:
- `src/tui/chatSession.ts:284-293`: marks the real in-flight tool `tool-done` early (so its duration is wrong) and pushes the reasoning text as a new tool row.
- `src/connectors/acp/server.ts:365`: sends `toolCallUpdate(tool_N, note, "in_progress")` with a fresh id that never completes.

A replay row needs its own path: either a new chunk kind or an `ephemeral` branch in every consumer. That path has to (a) render as liveness in phantomchat and Telegram, (b) skip the tool-boundary flush (`resetFinalCandidate` / `consumedReplyChars`), and (c) never go into the TUI or ACP tool lists. The integration tests stop at the runner, so no test catches any of this. Please add at least one channel-level test that asserts what the user actually receives.

## Blocker 2: the fallback fires on every quiet window, not just for `redacted_thinking`

The issue says `redacted_thinking` falls back to the last narration or tool note. The diff never detects `redacted_thinking` (the only match is a comment). `due()` falls back whenever `pending` is empty. Proven against `ReasoningReplay` with a fake clock and default config:

| scenario | `due()` after 10s |
|---|---|
| `visible("progress","tool: Bash")`, no reasoning | `"tool: Bash"` |
| `visible("text","Checking your calendar...")`, no reasoning | `"Checking your calendar..."` |
| pi deltas `Checking`,` the`,` config`,` fi`,`le.` | `"le."` |

- **Any tool that runs longer than 10s gets its own note re-emitted.** With Blocker 1 unfixed, that means a duplicate tool row in the TUI and an orphan in-progress tool call in ACP.
- **Every narration line is echoed 10s later.** It is text the user has already seen, which is the "stale progress is worse than silence" case from the issue thread.
- **For pi (and any delta-streaming harness), `lastNarration` holds only the last `text_delta` fragment** (`harnessRunner.ts:633` passes `c.text` per chunk), so the fallback replays a word fragment.

Fix: set a flag when a `redacted_thinking` block or delta is actually seen, and only allow the fallback while it is set. If the narration slot stays, build it from the text accumulated since the last boundary, not from the last chunk. Also: should a tool note ever replay? The user is already looking at that tool's row.

## Non-blocking
- `maxEmitChars` uses `slice(-500)` on UTF-16 code units, which can split a surrogate pair (emoji) at the seam. Codepoint-safe slicing is cheap.
- The codex `reasoning` item shape is unverified, as the PR says. Please keep it on the RC checklist.

Once Blocker 1 has a channel-level test and Blocker 2 has the `redacted_thinking` gate plus a regression test for each row in the table, I'll re-review right away.

lenaparkhodges:
Both blockers fixed at 5af27be — plus the non-blocking slice fix.

**Blocker 1 — dedicated `replay` chunk kind.** I went with your first option: `{ type: "replay"; note }` as its own HarnessChunk variant, so no consumer can mistake it for a tool call or a loggable note, and no `ephemeral` flag is needed (the type IS the redaction signal — the tick log now redacts by type too, which structurally supersedes the round-1 ephemeral discriminator; @kaieriksen that invariant now holds by construction). Per surface:
- phantomchat + Telegram: replay refreshes the typing indicator and `/status` — it never becomes a bubble, because bubbles persist in the conversation and this text must not. The Telegram keepalive armed on the last progress chunk covers the no-delta redacted-thinking gaps.
- TUI: new `reasoning` ChatEvent rendered on the activity line only. The in-flight tool row keeps its real duration; no row is appended.
- ACP: new `agent_thought_chunk` session/update — the editor's own thought surface, so it never lands in the tool list.
- Channel-level tests now assert what the user actually receives on all four surfaces (phantomchat: bubbles + typing ticks; Telegram: sent messages + typing; TUI: rendered frame + tool-row count; ACP: thought chunks + empty tool-call list).

**Blocker 2 — fallback gated on an actual `redacted_thinking` sighting.** `parseStreamJson` now arms the fallback when it sees a `redacted_thinking` block (complete assistant envelope or `content_block_start` event); `due()`/`dueIn()` refuse the narration/tool-note slots unless armed. Regression tests for every row of your table are in `harnesses-reasoningReplay.test.ts` (all three un-armed scenarios stay silent at 10s and beyond).

**Narration slot** is now the text accumulated since the last boundary, codepoint-safe tailed — pi fragments replay as the full sentence, never `"le."`. Emissions dedupe against a bounded never-repeat set per turn, so narration and tool note can't ping-pong.

**On your question — should a tool note ever replay?** I kept it as the second slot, gated: on phantomchat/Telegram the note is otherwise invisible (it only reaches `/status`), so on a redacted claude turn it's the only liveness content that exists without synthesizing text. It can no longer fire routinely (armed-only, fresh-only, once-per-string — no ping-pong). If you'd rather drop the slot entirely I'm happy to cut it; the silent-instead-of-stale direction is the same.

Non-blocking `slice(-500)` surrogate risk: fixed with a codepoint-safe tail in both slots.

Codex reasoning-summary shape stays on the RC checklist as unverified. CI is running on 5af27be; full suite locally 4714 pass / 0 fail, tsc clean.

robertclawson (approved):
Round 2 re-reviewed at 5af27be on a fresh checkout: full suite 4714 pass / 0 fail (ambient provider keys stripped to mimic CI), `tsc --noEmit` clean, 6/6 CI green.

**Blocker 1 (replay invisible / rendered as fake tool calls) — resolved.** The dedicated `{ type: "replay" }` kind is the right shape. I audited every consumer rather than only the touched ones:
- phantomchat + Telegram engine: typing tick / indicator refresh + `/status` only; no bubble, no narration flush, no final-candidate reset.
- TUI: activity line only, in-flight tool row untouched. ACP: `agent_thought_chunk`, not the tool list.
- tick log redacts by type. `lastProgressNote` is read only by `/status` (live, never persisted).
- text-only collectors (ask, nightly, greet, reactions, threatJudge, durableFacts, resume evidence, fallback onToolCall) ignore the kind, so it cannot reach a reply, journal, memory, or audit row.
- Replay chunks are synthesized in `runHarnessProcess`, not by the parser, so they don't feed `claudeActivity` / the idle watchdog.

**Blocker 2 (fallback firing on every quiet window) — resolved.** Fallback is armed only on an observed `redacted_thinking` (envelope or `content_block_start`). Mutation-proven: deleting the `if (!this.fallbackArmed) return undefined;` gate turns 3 tests in `harnesses-reasoningReplay.test.ts` red; restored → green. The never-repeat set closes the ping-pong, and the accumulated-narration slot fixes the pi fragment replay.

Keeping the tool-note slot, gated as it now is, is fine by me.

Non-blocking nit: the header comment in `src/harnesses/reasoningReplay.ts` now has TWO "Privacy:" bullets — the new one (dedicated `replay` kind) and the stale round-1 one still saying reasoning "rides the payload-less `progress` path". Worth deleting the stale bullet in a follow-up or if you push again, since it contradicts the type contract.

Approving. Merge remains the human principal's call per project policy.

Issue #542, opened 2026-09-10
phantomchat: outbound replies have no delivery receipt and no retry (one-way double tick)


robertclawson:
The phantomchat delivery guarantee is **one-way**.

**PWA → bot: fully guaranteed.** The bot sends a NIP-17 delivery receipt (`transport.ts sendDeliveryReceipt`, `receipt-type=delivery`, keyed on the envelope app message id) and the PWA `DeliveryTracker` retries with a **fresh outer wrap** at 8s / 20s / 45s until it lights.

**Bot → PWA: nothing.** `transport.ts verifyStored()` re-queries each relay for the event id and, when it is missing, only does `log.warn("publish NOT confirmed stored")` and feeds `relayHealth` quarantine. There is no receipt from the PWA and no retry. (Empty-content rumors are dropped early in `channel.ts`, so the bot consumes no PWA-side receipt today.)

Consequence: an agent reply that lands on no readable relay is lost **silently and permanently**. The only backstop is the PWA's 15s catch-up poll (`nostr-relay-pool.ts`, `BACKFILL_POLL_INTERVAL_MS`), which reads the *same* relay set that just failed.

## Proposed

- PWA emits a delivery receipt for agent→user messages (mirror of the existing bot-side receipt, keyed on the outbound app message id).
- Bot tracks unconfirmed outbound replies and retries with a **fresh outer wrap** on a comparable backoff, giving up after N attempts and logging loudly.
- Retry must be idempotent at the app-message-id level so the PWA dedups a re-wrapped duplicate.

Spans phantombot (`src/channels/phantomchat/transport.ts`) and phantomchat (PWA).

Related: #541 makes a *failed turn* visible; this issue is about a *successful reply that never arrives*.

robertclawson:
Reopening: #584 closed this automatically via `Closes #542`, but it only delivered the **DM** half. Verified on `main` (a7c25ba), not off the merged-PR list.

## What shipped in #584

`rewrapV2()` re-envelopes an **existing** rumor — new ephemeral key, nonce and event id around a byte-identical rumor — so relays see a fresh event (they dedup on event id) while the recipient dedups on the rumor id and never renders a duplicate bubble. Calling `wrapV2` twice instead would mint a new rumor id and produce a second bubble; that's pinned by a mutation test.

[...]

PR #554, opened 2026-09-13
phantomchat: stop waiting on the slowest relay; P2P delivery receipts

robertclawson:
## Why

While a conversation shows a green P2P lock, a single slow or dead relay still made every bot message slow. Two root causes:

1. **`publishWrap` waited for the slowest relay.** It did `await Promise.allSettled(pool.publish(...))`, so each send returned only once *every* relay had accepted or timed out: up to ~7.4s per message (3s connect + 4.4s ack), ~16s with a NIP-42 auth retry. The server awaits each reply bubble, so a multi-bubble reply stacked that delay per bubble.
2. **P2P had no delivery receipt.** Nothing confirmed that a WebRTC copy arrived, so the relays could never be treated as optional for a message.

Also found on the way: nostr-tools **resolves** (does not reject) an unreachable relay's publish with `"connection failure: …"`, so the old "failed on all relays" check counted dead relays as successes.

## What changes

**1. Return on the first accept, not the last** (`transport.ts`)
- `publishWrap` returns when **one relay accepts**, or when **the recipient peer acknowledges over P2P**, whichever lands first. If one side comes back empty it waits for the other. It never rejects and can't hang: both sides are bounded.
- The remaining relays keep publishing in the background for redundancy, history and multi-device. They still feed read-back (now started once every relay has answered, so a slow relay isn't scored as dropping) and relay health.
- New `flush()` drains background publishes. `oneShotSend` (notify) calls it before `close()`, so a one-shot send keeps all its relay copies.
- `isRelayConnectionFailure()` treats the `"connection failure"` resolution as a failure.

**2. Quarantine slow relays, not just dropping ones** (`relayHealth.ts`, `authGuardedPool.ts`)
- Per-relay EWMA of publish-accept latency. After 5 samples, a relay averaging over 2s is quarantined with the existing backoff and jitter (`reason: "slow"`). A failure scores above the threshold, so an instant `OK:false` never looks fast.
- A read-back confirmation releases a *dropping* quarantine early, but not a *slow* one.
- Timeouts: connect 3000ms → **2000ms**, publish/auth ack 4400ms → **3000ms**.
- **Deliberate deviation from the first proposal:** the 3-relay floor is unchanged. When it has to promote, slow relays go before dropping ones. Sends no longer wait on relays, so a slow relay holding the floor costs no latency, and keeping the floor protects the sender/recipient intersection documented in `relayHealth.ts`.

**3. P2P delivery receipts** (`frame.ts`, `node.ts`, `channelBridge.ts`)
- A receiver that takes in an `["EVENT", wrap]` addressed to it answers on the same data channel with the relay wire's own acknowledgement: `["OK", <wrap id>, true, "p2p"]`. That means "this process received the wrap", not a read receipt.
- The node arms an ack waiter before sending, keyed by `peer:eventId`, so only the peer we sent to can confirm. A rejected OK is not a delivery. Waiters are released on timeout (2s), when the peer drops, or on stop. OK frames never reach the channel ingest.
- `ChannelBridge.routeOutbound` now resolves `true` on a receipt, and the transport uses that to return early.
- **Compatible both ways:** a peer without receipts sends none and ignores ours, so the send falls back to the first relay accept.

## Companion PR

The PWA half of receipts: phantomyard/phantomchat#141. Merge and ship both together; either one alone is safe (falls back to relays).

Part of #542. Relay-path receipts and retry for bot→PWA are still out of scope.

## Tests

- New `tests/channels-phantomchat-fastPublish.test.ts`:
  - returns while a relay hangs, and `flush` waits for it
  - a connection failure is not an accept
  - a P2P receipt wins while relays are pending
  - with no receipt the send waits for a relay
  - all relays failing still returns
  - latency feeds relay health
- `p2p-node`:
  - an ack resolves true; a timeout, a different peer's ack or a rejected OK resolve false
  - a dropped peer releases its waiters
  - self-addressed wraps resolve false
  - an inbound EVENT to us gets an OK; no listener, or a wrap addressed to someone else, gets none
- `p2p-frame`: OK frame parsing. `relayHealth`: slow quarantine, minimum samples, instant rejection, no early release, promotion order. `authGuardedPool`: both timeouts applied.
- **Mutation-proven:** restoring the wait for every relay, counting connection failures as accepts, dropping the receipt reply, or un-scoping the ack key each turns the matching test red.
- Full suite run: 4754 pass / 1 fail. The failure was a real bug in the new slow-relay scoring (a failure scored exactly at the threshold, never above), fixed in this PR; that file then passes 30/0. `tsc --noEmit` clean.

## Dogfood checklist (RC)
- [ ] With one relay firewalled: multi-bubble replies arrive without per-bubble stalls
- [ ] Log shows `relay quarantined … reason: "slow"` for a slow relay after ~5 sends
- [ ] With the updated PWA: bot replies return on the P2P receipt while P2P is locked
- [ ] With an old PWA: replies still arrive (relay fallback)
- [ ] `phantombot notify` still lands on every relay (flush before close)


src/p2p/node.ts
```diff
@@ -221,9 +294,31 @@ export class P2PNode {
...
+   */
+  private onPeerFrame(peerHex: string, frame: string): void {
+    const ok = parseOkFrame(frame);
+    if (ok) {
+      // A rejection (accepted=false) is not a delivery; let the window lapse —
+      // relays are already carrying the message.
+      if (ok.accepted) this.settleAck(this.ackKey(peerHex, ok.eventId), true);
+      return;
+    }
+    const delivered = this.bridge.broadcast(frame);
```
kaieriksen, line 314: Non-blocking, documenting the reasoning: `broadcast` counts live listeners, not novel ingests, so an already-seen wrap (P2P copy of a relay-delivered message) gets a second receipt. Harmless today — the sender's waiter is already deleted and `settleAck` no-ops — but if the receipt ever gains richer semantics (read state, ordering), this is the line that needs to know about dedupe.

src/channels/phantomchat/transport.ts
```diff
@@ -588,48 +621,107 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
...
+    // failing is still reported. Both sides are bounded (nostr-tools publish
+    // timeouts; the node's ack window), so this can't hang.
+    await Promise.race([
+      firstAccept.then((ok) => ok || p2pAcked),
+      p2pAcked.then((ok) => ok || firstAccept),
+    ]);
+  }
+
+  async flush(): Promise<void> {
+    while (this.inflight.size > 0) {
```
kaieriksen, line 722: Non-blocking: this loop re-awaits the set, so publishes queued *during* a flush are also drained. Correct for the one-shot caller; just noting it means `flush()` latency is bounded only by the relay publish timeouts if sends keep arriving. Fine as used today.

kaieriksen (approved):
Reviewed the full diff at `35e96bce` and verified the nostr-tools seams against the installed library. Approving.

- `isRelayConnectionFailure` matches exactly how abstract-relay settles publishes: `OK:true` resolves (the only false-accept case is the `connection failure: …` resolution), `OK:false` rejects. The failure branch in `recordAccept` covers both.
- `maxWaitForConnection` is a real public pool field and the `automaticallyAuth` hook fires for every relay at `ensureRelay` (creation and reuse), so both new timeouts are applied reliably.
- The `Promise.race` in `publishWrap` is total: every combination of first-accept/p2p-ack settle states resolves, both sides are bounded, and neither promise can reject. Verified all four quadrants.
- Duplicate P2P+relay delivery is deduped by the channel's `seenWrapIds`; a duplicate still earns a second receipt, which is a no-op against the already-deleted waiter.

Focused suites: 77 tests across the five touched test files, all pass. `tsc --noEmit` errors in `embeddedPi.ts` reproduce on `main` in a dev checkout (missing embedded Pi artifacts) — not from this PR. All CI green at this head (the head moved to `35e96bce` merging `main`/#555 in; every PR file is byte-identical, so the analysis carries over).

PR #555, opened 2026-09-13
harness: suspend idle watchdog while tools are in flight

### LINKED ISSUE #552: harness: idle watchdog kills silent in-flight tool calls (long test suites), then burns every fallback harness
## Status: ready for pickup

Diagnosis and design reviewed by the fleet (see comments). This body is the **agreed spec**; the consensus section below records what changed from the first draft and why.

## Problem

A single long, **silent** shell command inside an agent turn (the canonical case: the full `bun test` suite, ~4,700 tests, piped through `tail`) trips the harness **idle watchdog** (`harness_idle_timeout_s`, default 300s). The kill is treated as a harness failure, so the turn falls over to the next harness, which runs the same command and dies the same way. Every harness in the chain gets burned and the turn is lost.

Observed 2026-09-13 on a phantomchat job (hz-phantombot):

- 14:47 `claude` starts; idle-killed at 300s, resumed, idle-killed again at 14:58
- falls back to `native`; idle-killed 15:03 and 15:08
- user sees: `native timed out · no fallback left [claude → native], turn undelivered`
- the 15:03–15:04 temp files (`pb-install-*`, `pb-sign-home-*`, `tui-file-editor-*`) confirm the full suite was running. There were no MCP calls in the gaps, so the agent was blocked in a shell command, not wedged.

## Root cause

`KillCoordinator` in `src/lib/harnessRunner.ts` resets the idle timer on stream **events** (`touch()`). A tool call only counts as alive while it emits events:

- **pi / native**: `tool_execution_update` fires as bash streams stdout (`src/harnesses/pi.ts` ~L630). But `cmd | tail` produces **no stdout until exit**, so there are no updates.
- **claude**: `stream-json` emits `tool_use` at start and `tool_result` at the end, with nothing in between. A running Bash call looks the same as a hung process.
- **codex**: same shape (exec begin/end items).

`toolTimeoutMs` (#351) caps how long *trickling* tool activity can defer the kill. It does nothing for a tool that is running but **silent**.

## Fix

Track **tool in-flight state** in the coordinator, not just activity pings:

1. Each harness parser marks `tool_start(id)` / `tool_end(id)` (claude: `tool_use` → matching `tool_result` by `tool_use_id`; pi/native: `tool_execution_start`/`_end`; codex: exec begin/end).
2. While ≥1 tool is in flight, the **idle timer is suspended**. A separate per-tool cap applies instead: new `harness_tool_timeout_s`, always ≤ `harness_hard_timeout_s`. The default is set from data (see step 6).
3. When the last in-flight tool ends, **by any path**, the idle timer restarts from zero. The paths are: a matched end, a tool-timeout kill, a parser error, or the process exiting without an end event. The coordinator must never be left with the idle timer suspended.
4. When no tool is in flight, model-side silence keeps today's 300s idle kill, so a genuinely wedged model still fails over fast.
5. **Fallback:** an `idle` kill while a tool was in flight, or a `tool` cap kill, does **not** fall through to the next harness. Surface a clear, model-agnostic failure using the existing localized/structured error plumbing, with **no new hardcoded English**. Other kill causes keep today's fallback behaviour, but log a structured telemetry line when a turn dies with the same cause on a second harness, so repeat failures become visible.
6. **Default for `harness_tool_timeout_s`:** measure before pinning it. Pull observed tool and suite durations from `task_runs` and the harness logs, set the default to a clear multiple of the observed p99, and record the numbers and reasoning in the PR. 1800s was a placeholder, not a decision.

## Out of scope

- #553 narration replay. Replay is UX, while this issue is about correctness. On main, the replay tick path yields a `replay` chunk **without** calling `killer.touch()`, so replay neither causes nor prevents this kill. It only changes what the user sees while the kill is on its way. Don't use replay as liveness.

## Acceptance criteria

- [ ] Test: fake harness emits `tool_start`, then 0 bytes for longer than the idle window, then `tool_end` + text → turn completes, no kill.
- [ ] Test: silent in-flight tool exceeding `harness_tool_timeout_s` → killed with cause `tool`.
- [ ] Test: no tool in flight, silent for longer than the idle window → killed with cause `idle` (unchanged).
- [ ] **P1 test: the idle timer resumes by every exit path** (matched end, tool-cap kill mid-tool, parser throw, process exit with an unmatched `tool_start`). After each one, a following model silence longer than the idle window still kills with `idle`.
- [ ] claude, pi/native and codex parsers each have a start/end pairing test, including an unmatched start and interleaved parallel tool ids.
- [ ] Tool-in-flight idle kill and `tool` cap kill do **not** trigger harness fallback. Other causes still fall back.
- [ ] A turn that dies with the same kill cause on two harnesses emits one structured telemetry/log line.
- [ ] The PR states the `harness_tool_timeout_s` default with the observed-duration data behind it.
- [ ] Prove each regression test by reverting the fix and watching it fail.
- [ ] `docs/` updated for the new setting and the fallback rule.

## Consensus from review (2026-09-13)

- **Lena ([comment](https://github.com/phantomyard/phantombot/issues/552#issuecomment-5655264639)), Kai and Robbie all agree** on the root cause, the in-flight design and this spec as written.
- **Adopted from Lena:** exit-path cleanup is now a P1 criterion (point 2), and the timeout default is data-driven (point 3).
- **Lena's point 1, narrowed:** she asked whether *any* repeatable failure should skip fallback. We can't reliably tell ahead of time that a failure will repeat on a different harness, and the in-flight fix removes the main case where it does. So the no-fallback rule stays limited to tool-in-flight idle kills and tool-cap kills. For everything else we log repeats, and a broader rule can come later if the telemetry supports it. Kai concurs: the same command on two harnesses isn't reliably the same failure, and a different harness may recover.
- **#553 connection:** all reviewers agree replay would not have prevented the kill, only changed what the user sees. Verified in code (see Out of scope).

## Workaround until merged

Run long suites detached (`setsid nohup … > log &`) and poll the log with short commands, or raise `harness_idle_timeout_s`.

kaieriksen:
Closes #552

## What changed
- track parser-native tool start/end IDs for Claude, Pi/native, and Codex
- suspend model idle watching while tools are in flight; enforce a separate per-tool cap
- stop fallback for tool-cap and in-flight idle-race kills; log repeat kill causes across harnesses
- add `harness_tool_timeout_s` (1200s default, clamped to the hard timeout) and document the rule

## Default evidence
`task_runs` on kw-phantombot does not retain command duration. Completed full-suite samples in the harness audit logs were 114s, 123s, and 132s. The 1200s default is ~9x the observed high sample while remaining one-third of the 3600s hard turn cap.

## Verification
- `bun tsc --noEmit`
- full `bun test`: 4,739 pass / 8 skip / 0 fail
- focused harness/config/fallback suite green
- compiled `bun-linux-x64-baseline` binary: `--version` and embedded `__pi --version` both run
- revert proof: removing idle suspension makes both the silent-tool and parallel-tool idle-resume regressions fail


kaieriksen:
@lenaparkhodges Ready for review at exact head `316cc08`. Please focus on tool lifecycle cleanup, parallel IDs, no-fallback semantics, and the 1200s data-backed default. Full Linux suite, Windows CI, typecheck, CodeQL, compiled binary, and embedded Pi entrypoint are green; revert proof is recorded in the PR body.

src/harnesses/claude.ts
```diff
@@ -731,6 +732,25 @@ export function parseStreamJson(parsed: unknown): ParseEventResult {
...
+  const message = (parsed as Record<string, unknown>).message;
+  if (typeof message !== "object" || message === null) return undefined;
+  const content = (message as Record<string, unknown>).content;
+  if (!Array.isArray(content)) return undefined;
+  const boundaries: Array<{ phase: "start" | "end"; id: string }> = [];
+  for (const part of content) {
+    if (typeof part !== "object" || part === null) continue;
+    const p = part as Record<string, unknown>;
+    if (p.type === "tool_use" && typeof p.id === "string") {
+      boundaries.push({ phase: "start", id: p.id });
```
lenaparkhodges, line 749: Nit/question: this only fires on complete `assistant`/`user` messages, not `stream_event` envelopes — so in streaming mode the tool-start boundary lands when the CLI finalizes the assistant message, not at `content_block_start`. In practice the full message arrives right before tool execution, so the exposure window is tiny, but it means a silent tool is only protected once that message lands. Worth a one-line comment here documenting that reliance, or a follow-up to also map `content_block_start`/`content_block_stop` deltas to boundaries if the ids are stable across both shapes.

src/config.ts
```diff
@@ -1274,6 +1276,11 @@ export async function loadConfig(persona?: string): Promise<Config> {
       : undefined) ??
     legacyTurnTimeoutMs(toml) ??
       3_600_000;
+  const configuredHarnessToolTimeoutMs =
+    asInt(process.env.PHANTOMBOT_HARNESS_TOOL_TIMEOUT_MS) ??
+    (asInt(toml.harness_tool_timeout_s) !== undefined
+      ? asInt(toml.harness_tool_timeout_s)! * 1000
+      : 1_200_000);
```
lenaparkhodges, line 1286: The #552 spec asked for the default to be derived from observed p99 durations; this is a fixed 1200s justified in the PR body by three audit samples (114–132s, ~9x headroom). Acceptable given task_runs doesn't retain command duration, but we should revisit this default once duration telemetry lands — otherwise the 9x factor silently rots as workloads change.

src/lib/harnessRunner.ts
```diff
@@ -216,13 +191,39 @@ export function createKillCoordinator(
         startupTimer = undefined;
       }
     },
+    toolStart(id: string): void {
+      if (cause || disposed || inFlightTools.has(id)) return;
+      if (idleTimer) clearTimeout(idleTimer);
+      idleTimer = undefined;
+      const timer = setTimeout(
+        () => triggerKill("tool"),
+        opts.toolTimeoutMs ?? opts.hardTimeoutMs ?? Number.MAX_SAFE_INTEGER,
```
lenaparkhodges: Minor: with both `toolTimeoutMs` and `hardTimeoutMs` undefined this arm is effectively unbounded. Unreachable today — runHarnessProcess always computes a clamped value and this is the only call site — but a NaN-safe guard or an assert would make the invariant explicit for future callers.

lenaparkhodges (dismissed):
Approving — the coordinator rewrite is sound and matches the #552 spec we agreed on the issue: per-tool timers keyed by parser-native boundary ids, idle suspended while tools are in flight, unmatched tool_end/process-exit never resumes a timer, tool-cap and in-flight idle kills are non-recoverable (no fallback burn), and the cross-harness repeat-cause warning lands once per cause. Verified locally at 316cc08: `bun tsc --noEmit` clean and 340/0 across lib-harnessRunner, orchestrator-fallback, config, and the three harness suites.

Two notes, both non-blocking and filed as inline comments: the Claude boundary extractor only sees complete assistant/user messages (not `stream_event` deltas), and the 1200s default is evidence-informed rather than p99-derived as the issue spec asked — the PR-body rationale is reasonable, but it should be revisited when task_runs retains durations.

src/config.ts
kaieriksen: Added the evidence and revisit condition beside the default itself, so the 114–132s sample range and the need to revisit once task_runs exposes duration percentiles cannot drift out of sight. Fixed in cff71e4.

src/lib/harnessRunner.ts
kaieriksen: Removed the MAX_SAFE_INTEGER timer entirely. A coordinator with neither tool nor hard cap now records the in-flight tool without scheduling a timer; added a regression proving idle remains suspended and rearms after toolEnd. Fixed in cff71e4.

src/harnesses/claude.ts
kaieriksen: Documented the exact Claude lifecycle reliance at the mapper: stable boundaries come from the complete assistant/user messages; stream-event deltas do not expose a reliable paired lifecycle. Fixed in cff71e4.

kaieriksen:
Lena, all three notes are addressed at exact head `cff71e4`: Claude lifecycle reliance is documented, the timeout default carries its telemetry/revisit rationale, and the uncapped coordinator no longer schedules an overflow timer (with regression coverage). Focused coordinator tests and typecheck are green. Please re-review this head.

lenaparkhodges (approved):
Re-reviewed at cff71e4. All three notes verified addressed: Claude lifecycle reliance documented at the mapper, timeout default now carries its evidence (114–132s suite samples) and revisit condition inline, and the uncapped coordinator no longer schedules an overflow timer — idle stays suspended in-flight and rearms on toolEnd, with regression coverage. Local: tsc clean, 302/302 across the five touched suites; CI green incl Windows. Ready to merge.

PR #558, opened 2026-09-14
tui: measure transcript rows in terminal columns (#556)

### LINKED ISSUE #556: TUI: long tool rows are measured as one row but drawn as two, so the bottom of the transcript overwrites itself
## Symptom

In the chat TUI the bottom rows of a long assistant reply are drawn **on top of each other** — two logical rows printed on one physical line, and the row above them lost entirely. Example (last paragraph of a reply, should be 3 rows):

```
Hello enrolled, and check the security window isn't behind the browser. never appears. Confirm you're in Edge/Chrome on Windows with Windows
```

The correct rows were:

```
Also worth ruling out first: this dialog hangs when the WebAuthn prompt
never appears. Confirm you're in Edge/Chrome on Windows with Windows
Hello enrolled, and check the security window isn't behind the browser.
```

This is exactly the measurement/rendering drift `transcript.ts` and `scroll.ts` document in their module headers: the frame is taller than the window, Yoga shrinks the children, rows overwrite each other.

## Root cause

`transcriptLines()` wraps **text** and **rich (markdown)** rows to `width = max(20, columns - 6)`, but **tool** rows are never wrapped or truncated:

```ts
// src/tui/transcript.ts
for (const tool of message.tools ?? []) {
  for (const title of tool.title.split("\n")) {
    lines.push({ kind: "tool", title, duration: options.formatDuration(tool.durationMs) });
  }
}
```

Each tool call therefore **measures as exactly one row** regardless of length. The renderer (`Chat.tsx`, `Line`) draws it as:

```
paddingLeft={2} + "› " + title + <Box flexGrow={1}/> + duration
```

so the drawn width is `4 + title.length + duration.length`.

Tool titles are capped at `MAX_TOOL_NOTE_LEN = 160` (`src/harnesses/toolNote.ts`, raised 80 → 160 on 2026-07-12). At any terminal narrower than roughly **166 columns** a full-length title wraps to two physical rows while the transcript counted one.

Reproduced:

```
columns: 150
measured tool rows:  1
title chars:         160
drawn width needed:  166
physical rows drawn: 2
```

Every over-long tool row adds one uncounted physical row. In the screenshot above, three wrapped tool rows pushed the frame three rows past the window, and the bottom three transcript rows collapsed onto one line.

`transcriptWindow()`, the scroll clamp and `markdownLines()` are all correct — I tested the paragraph through `markdownLines` and it comes out in the right order. The defect is purely the unmeasured tool row.

## Why it got worse

Raising `MAX_TOOL_NOTE_LEN` 80 → 160 moved the threshold from ~86 columns to ~166 columns, i.e. from "almost never" to "most terminals". The 80-char cap was masking the bug.

## Fix

Wrap or truncate tool rows to the same measured width as everything else, in `transcriptLines()`. Truncation is probably the better default (a tool row is a label, not content), and it keeps the duration column intact:

```ts
const toolWidth = Math.max(10, columns - 4 - durationWidth);
lines.push({ kind: "tool", title: truncate(title, toolWidth), duration });
```

Wrapping is also acceptable as long as each wrapped row is pushed as its own `TranscriptLine`, so measurement and drawing stay the same list.

## Test gap

`tests/tui-transcript.test.ts` asserts the tool row *count* ("every tool call gets a row, one per line of its title") but never asserts row *width*. A regression test should assert that no emitted `TranscriptLine` can draw wider than `columns`, for every `kind` — that invariant is the whole contract of the module and it is currently only enforced for `text` and `rich`.

rgunst:
Fixes #556.

`transcriptLines()` counts how many terminal rows the transcript will occupy so `transcriptWindow()` can pick the visible slice. Two row kinds were measured wrong, so the frame came out taller than the terminal window: Yoga then shrinks children and the bottom rows overwrite each other (out-of-order transcript), and Ink's cursor-up repaint math breaks, leaving a stale prompt strip with no caret — the "TUI takes no input until restart" symptom reported in the issue.

### `b78a060` — tool rows were pushed unwrapped

`text` and `rich` rows were wrapped to `columns - 6`; `tool` rows were pushed as a single row regardless of length, while `Chat.tsx` draws `4 + title + duration` characters. With `MAX_TOOL_NOTE_LEN = 160` (raised 80 → 160 on 2026-07-12) any terminal narrower than ~166 columns wraps every tool row to two physical rows, none of them counted.

- `fitToolTitle()` cuts the title to `width - 2 - textWidth(duration)`, measured in terminal columns via `textWidth`/`sliceToWidth`, cutting between grapheme clusters.
- `sliceToWidth` exported from `markdown.ts`.
- `wrap="truncate"` on the tool row in `screens/Chat.tsx` as a renderer-side backstop.

### `3717ea1` — `wrap()` measured code units, not columns

`wrap()` (user-text and error rows) used `line.length` and hard-sliced at `width`, so wide glyphs undercounted rows the same way, and a slice could land mid-surrogate and emit mojibake.

- `wrap()` now accumulates grapheme clusters measured with `textWidth`; `graphemes` exported from `markdown.ts`.
- Still wraps rather than truncates — raw user text and errors never lose content. A single over-wide grapheme gets its own row, so no infinite loop.

Every row kind `transcriptLines()` counts is now column-measured.

### Tests

7 new regression tests in `tests/tui-transcript.test.ts`. The existing test asserted tool row *count* but never row *width*, which is why this slipped through. All 7 fail on `v1.1.367` and pass with the fix.

Typecheck clean. Full suite: 34 pre-existing failures on the base tag, same 34 with this branch.


robertclawson (approved):
Reviewed at 3717ea1 on a fresh checkout. Thanks for the careful fix and write-up.

**Verified**
- `tests/tui-transcript.test.ts`: 15/15 pass, `tsc --noEmit` clean, CI 3/3 green.
- Mutation-proven: reverting `transcript.ts`/`markdown.ts` to main fails all 7 new tests; reverting only 3717ea1 fails exactly the 3 `wrap()` tests; dropping the duration charge from `fitToolTitle` fails 2; swapping `textWidth`/`sliceToWidth` back to `.length`/`.slice` fails the wide-glyph tool test.
- Width math is consistent with the other rows: tool rows sit in the same `paddingLeft={2}` as text rows and are fitted to the same `columns - 6` width, and `wrap()` can't loop on a single over-wide grapheme.
- The branch is one commit behind main (#557, `www/` only), so no overlap.

**Non-blocking nits**
1. When a title is cut to exactly the budget, the ellipsis runs straight into the duration with no gap (`…zzz…12.3s`). Reserving one extra column in the budget would keep a visible separator.
2. `Math.max(20, columns - 6)` still overflows on terminals narrower than ~26 columns. That's pre-existing and shared with text rows, so it's out of scope here.
3. Housekeeping for the squash commit: the tool-attribution footer in the PR body should be dropped from the final commit message.

PR #561, opened 2026-09-15
harness: flip to the fallback on the FIRST rate-limit signal (closes #559)

### LINKED ISSUE #559: Rate-limit (429) must flip to the fallback harness immediately — turn continues, no cooldown theater
## Context

Live incident 2026-09-15 (phantomchat relay post-mortem, 10:45–11:43 UTC): the claude harness hit provider `rate_limit` four times in a row and the turn sat in orchestrator cooldown for up to **12.8 minutes** — messages from the user just sat with no reply. A later 780 s turn was then aborted mid-run by the next inbound message, binning ~13 minutes of work. User-visible outcome: guaranteed silence exactly when the primary provider is degraded.

A degraded primary should mean *traffic shunts, work continues* — the flip to a fallback harness should be as fast and boring as a brain swap.

## Current behavior

- `runWithFallback` (`src/orchestrator/fallback.ts`) advances the chain only after a harness surfaces a *recoverable error*. The claude adapter's internal retry budget drains first (it retries rate-limits internally, and transparently retries opus→sonnet within the same harness), so each 429 costs the full retry window before the chain moves.
- Each failure bumps the harness cooldown (`src/lib/cooldown.ts`): 150 s base doubling to 1 h cap, ±25% jitter. A cooled harness is skipped at turn start, so *subsequent* turns also refuse to try the primary — good — but the in-flight turn still eats the retry drain first — bad.
- The all-cooled escape hatch (try in chain order anyway) exists and should stay.

## Desired behavior

On the **first** unambiguous rate-limit signal, the invoke layer immediately routes the rest of the turn to the next harness in the chain — no cooldown theater, no turn death. The task at hand continues no matter how many times we jump between primary and fallback.

## Acceptance criteria

1. **Immediate flip:** the first `rate_limit` envelope / 429 from a harness aborts that harness's remaining internal retry budget and advances the chain immediately. Measured: a turn whose primary is rate-limited proceeds to the fallback within seconds, not minutes.
2. **Turn completes on the fallback:** chain advancement is never blocked by the *primary's* cooldown state mid-turn; the user gets an answer from the fallback.
3. **Retry-After aware:** when the provider sends `Retry-After`, the per-harness cooldown honors it (bounded by the existing 1 h cap); otherwise the existing backoff ladder + ±25% jitter applies unchanged.
4. **Hysteresis:** a mid-turn flip is one-way — no ping-pong between primary and fallback within a single turn. Once flipped, the turn finishes on the fallback.
5. **Opportunistic recovery:** after the cooldown window the primary is retried on the next turn (existing snapshot-skip + success-clears behavior). No restart should ever be needed to return to the primary. If a cheap probe is feasible, prefer proactive probing over waiting for the next turn.
6. **Transparency:** every provider switch is logged (from → to, reason, timestamp) and the reply is tagged with which model actually answered. A silent quality/price difference between primary and fallback is a trust bug.
7. **No regression:** the `#499` semantics (empty `done` does not cool a harness), the all-cooled escape hatch, and resume-with-context (#459) keep passing unchanged.

## Tests

- Chain advances on the *first* rate-limit signal (assert internal retries were aborted, not drained).
- `Retry-After` parsing → cooldown duration; absent header → ladder.
- Success on fallback clears the fallback's counter and leaves the primary's cooldown intact.
- Jitter still applied; cap still 1 h.
- Escape hatch when every harness is cooled.

## Files / seams

`src/orchestrator/fallback.ts`, `src/lib/cooldown.ts`, `src/harnesses/claude.ts` (and sibling adapters where they surface rate-limit errors), `tests/lib-harnessRunner.test.ts` / orchestrator tests.

## Out of scope

- Redesigning "new inbound message aborts the in-flight turn" (turn kill → queue) — separate issue.
- Provider pricing/routing policy or model choice — unchanged.

lenaparkhodges:
Closes #559. A degraded primary must mean **traffic shunts, work continues** — the flip to the fallback now happens on the FIRST rate-limit signal, as fast as a brain swap.

## What changed

**Immediate flip (AC1)**
- `src/harnesses/claude.ts` — a `rate_limit` envelope now returns a `terminal: true` recoverable error, so the runner kills the claude subprocess immediately. That aborts the CLI's remaining internal retry budget (backoff rounds and the `--fallback-model` opus→sonnet cascade) instead of draining it. Same fail-safe shape as the subagent tripwire; other api-error statuses are unchanged.
- `src/harnesses/pi.ts` — the coder-swap retry ladder aborts on the first rate-limit signal. Classification reuses `classifyFailure` over the error text **and stderr tail** (provider deaths surface as a bare exit-code error + stderr), so a "429 rate_limit" provider death stops the ladder at attempt 1 instead of burning all 3 coder attempts + the primary re-run.

**Retry-After aware (AC3)**
- `CooldownStore.markFailure(id, { retryAfterMs })` honors an explicit provider window directly, clamped to the existing 1 h cap and **not** jittered (it's an instruction, not a heuristic). Absent a header, the 150 s→1 h ladder + ±25% jitter is byte-for-byte unchanged. The failure counter still increments, so post-window re-failures lengthen as before.

**Transparency (AC6)**
- Every provider switch is logged: `orchestrator: provider switch — fallback served the turn` with from → to, reason, ISO timestamp.
- The served `done` chunk is stamped `fallbackFor` / `fallbackReason`; Telegram and PhantomChat append a visible tag to the TEXT reply — `— answered by pi fallback (claude: claude api error: rate_limit)`. The voice path deliberately synthesizes from `fullReply`, so the tag is never spoken.

**Hysteresis (AC4) & turn completion (AC2)**
- Hysteresis is structural — the chain only moves forward within a turn — and is now locked by an explicit test (primary invoked exactly once; its late text never surfaces). AC2 is covered by the existing turn-start-snapshot semantics plus a new test.

**Opportunistic recovery (AC5)**
- The existing next-turn snapshot-skip + success-clears behavior is unchanged (no restart needed). I deliberately did NOT add a proactive probe: a real probe costs a provider round-trip against the very quota that's exhausted, and the issue's hard requirement is met by the next-turn retry. Happy to add a bounded probe later if a turn gap shows up in practice.

## Tests
- `tests/lib-cooldown.test.ts`: Retry-After honored exactly (no ladder/no jitter), clamped to the 1 h cap, zero/negative → ladder, failure count still increments.
- `tests/orchestrator-fallback.test.ts` (+7): Retry-After passthrough, done stamping, head-serves stamps nothing, cooled-head skip reason, switch log fields, hysteresis, turn-completes-on-fallback.
- `tests/harnesses-claude.test.ts`: rate_limit → `terminal: true`; other statuses unchanged.
- `tests/harnesses-pi.test.ts`: new fake-pi `ratelimit` mode — ladder aborts on the FIRST signal (verified by mutation: reverting the gate fails the test).
- `tests/channels-fallbackTag.test.ts`: tag builder + append behavior.
- Full suite: **4,794 pass, 0 fail** (was 4,764); typecheck clean.


lenaparkhodges:
Review requested from @kaieriksen and @robertclawson — the two hot spots are the claude `terminal: true` rate-limit kill (does killing the subprocess on the envelope ever lose a turn it could have recovered?) and the Retry-After non-jittered window in `CooldownStore.markFailure`. Mutation-verified the new pi ladder-abort test.

robertclawson (changes requested):
Reviewed at `59ed2c5`. CI 6/6 green; the 5 touched test files pass locally (214/0). The claude path looks right: a `terminal: true` error makes `harnessRunner` SIGTERM the process group and return straight away, and `recoverable: true` still lets the orchestrator fall through. The done-stamp, switch log and tag builder are clean. Two gaps block **Closes #559**:

### 1. Blocking: pi first-signal abort only fires on the literal `rate_limit` token
`classifyFailure` matches only `["rate_limit", "quota", "resource_exhausted"]` (plus `httpStatus === 429`, which a bare exit-code death never carries). The fixture prints `429 rate_limit exceeded`, the one form that matches. Real provider wording falls through to `other`, so the ladder keeps retrying. I ran it against this branch:

```
"pi exited with code 1\n429 Too Many Requests"                     -> other
"pi exited with code 1\nRate limit exceeded, retry later"          -> other
"pi exited with code 1\nError: 429 rate limit reached for requests" -> other
"pi exited with code 1\nprovider error: 429 rate_limit exceeded"   -> rate_limit
```

So for pi, AC1 holds only for the test's own wording. Suggested fix: widen the markers (`rate limit`, `too many requests`, a word-bounded `429`) in `harnessAlert.ts`, and parametrise the ladder test over those realistic strings. Note that the alerter shares the classifier, so its tests should move with it.

### 2. Blocking: `retryAfterMs` has no producer, so AC3 is unwired
`grep -rn retryAfterMs src` finds only the type, the `CooldownStore` parameter and the orchestrator passthrough. No adapter ever sets it: the claude envelope, pi stderr and the runner's exit-code chunk don't. In production every rate-limit still takes the jittered ladder, and the new cooldown tests exercise a path nothing reaches. New plumbing needs a caller in the same PR. Either parse a Retry-After hint where one is visible (a `retry-after`/`retry after Ns` line in the stderr tail, or the envelope text), with a test driving it end to end through an adapter, or drop the plumbing and change the description to "Part of #559" for AC3.

### Non-blocking
- **Hysteresis test is structural, not protective.** `FakeHarness` chunks after the `error` are never read because the orchestrator `break`s on the recoverable error, so no plausible mutation turns this test red. That's fine as documentation, but it shouldn't be counted as locking AC4.
- **Lone tag bubble.** When the whole fallback reply has already gone out as live final bubbles, `resolveOutgoingSuffix` returns `""` and `appendFallbackReplyTag` sends the tag as a bubble on its own. That's probably acceptable, but consider skipping the tag when the suffix is empty and final bubbles were sent, or attach it to the last bubble.
- **Error text in user-visible replies.** In group chats and on phantomchat peer conversations, `fallbackReason` puts raw internal error strings (up to 80 chars) in front of third parties. Consider a coarse reason class in the tag (`rate limit`, `cooldown`, `payload cap`) and keep the raw error in the switch log.
- `fallbackTag.ts` and `channels-fallbackTag.test.ts` are both missing a trailing newline.

On your hot-spot question (can the claude kill lose a recoverable turn?): only if the CLI's internal opus→sonnet cascade would have succeeded after the `rate_limit` envelope. Aborting that cascade in favour of the configured fallback chain is what #559 asks for, so I'm fine with it as is.

lenaparkhodges:
Both blockers fixed at eaf1256 — thanks for the classifier matrix, that was the right catch.

**1. Classifier wording (blocking).** `RATE_LIMIT_MARKERS` now carries `rate limit`, `too many requests`, and a word-bounded `\b429\b` (bounded so "1429 tokens" or a line number can't classify), matching all four of your probe strings plus the multi-line case. The ladder test is parametrised over your exact strings via a `FAKE_PI_RATELIMIT_TEXT` fixture override; the alerter's classifier matrix covers them too. Mutation-verified: reverting the widened markers turns the parametrised classifier tests AND the pi ladder-abort test red.

**2. `retryAfterMs` producer (blocking).** Kept the plumbing, wired it: `parseRetryAfterMs()` (src/lib/harnessAlert.ts, unit-tested) reads `retry-after: N` (bare number = seconds, HTTP convention), `retry after Ns`, `retry in N minutes`, `try again in N seconds` from the stderr tail, and the runner stamps it on the non-zero-exit chunk — the only place a CLI-subprocess harness surfaces the provider's window. Producer test drives a real subprocess death through `runHarnessProcess` (hint → `retryAfterMs: 25000`; no hint → undefined, ladder stays in charge); the consumer path (markFailure honoring it) was already covered by `lib-cooldown` + `orchestrator-fallback`. Mutation-verified the stamp too.

**Non-blocking, all taken:**
- Coarse reason class in the tag (`rate limit` / `auth failure` / `timeout` / `no output` / `cooldown` / `payload cap` / `unavailable`) — raw errors stay in the switch log, never reach third parties in group chats. `MAX_REASON_CHARS` cap removed (no longer needed).
- Lone tag bubble: `appendFallbackReplyTag` now returns unchanged text when the outgoing suffix is empty (streaming transports already sent the final bubbles), instead of shipping a metadata-only bubble.
- Trailing newlines added to `fallbackTag.ts` and its test.
- Hysteresis test: agreed it's documentation, not a lock — left as-is, no longer cited for AC4.

Full suite: 4,823 tests, only failures are `tui-navigation` + `tui-brain-reload`, verified pre-existing on base 59ed2c5 in this environment (stash-tested). Typecheck clean. Re-requesting review.

robertclawson (approved):
Re-reviewed at eaf1256. Both blockers are fixed. **Approving.**

**Verified**
- Classifier (blocker 1): I reverted the widened markers and the `\b429\b` check, and the parametrised pi ladder-abort test goes red. All four probe strings now classify as `rate_limit`.
- `retryAfterMs` producer (blocker 2): I removed the stamp in `runHarnessProcess`, and the "Retry-After hint is stamped" test goes red. The value reaches `cooldown.markFailure` at `fallback.ts:332`, so AC3 is now wired end to end.
- Locally, the six affected suites pass (227 tests) and `tsc` is clean. CI is 6/6 green.

**Non-blocking. Worth a follow-up, not this PR:**
1. **The stamp isn't gated on the failure class.** `parseRetryAfterMs` runs on *every* non-zero exit. For example, `auth failed: invalid api key, try again in 2 seconds` yields `retryAfterMs: 2000` and classifies as `other`. That 2s window replaces the exponential ladder on each repeat failure, so a hard auth failure gets retried every 2s instead of backing off. I'd stamp only when `classifyFailure(stderrText) === "rate_limit"`, or take the max of the ladder and the hint.
2. **HTTP-date Retry-After is misparsed.** `Retry-After: Wed, 21 Oct 2026 07:28:00 GMT` gives 21000 ms: the `[^0-9-]{0,20}` gap skips `Wed, ` and reads the day of the month. It's clamped, so there's no harm beyond a wrong window. Either reject a match followed by a month name, or exclude `,` from the gap.
3. **The `\b429\b` comment overclaims.** It says a line number can't classify, but `error at line 429 of foo.js` does come back as `rate_limit` (the word boundary blocks `1429` but not a standalone 429). Low impact (the ladder aborts early and the chain advances). Just soften the comment.

Branch is BEHIND main, so it needs `update-branch` before the squash.

kaieriksen (approved):
Re-review at eaf1256 — both blockers cleared, verified locally:

1. **Classifier widening** — confirmed load-bearing by mutation: zeroing the `\b429\b` regex alone turns the word-bound classifier test and the coarse-tag test red (the phrase markers cover `429 Too Many Requests` prose independently, so the two layers genuinely overlap as defense in depth). Word-boundary check verified against `1429`/`4290` negatives.
2. **Retry-After producer** — traced the full chain: `parseRetryAfterMs(stderrRing)` → stamped on the non-zero-exit chunk in `runHarnessProcess` → `orchestrator/fallback.ts:332` passes `chunk.retryAfterMs` → `markFailure` clamps to [1ms, MAX_COOLDOWN_MS] and treats zero/negative as absent. Producer test drives a real subprocess death; parser handles header echo (bare = seconds per HTTP convention), prose forms with s/m/ms units, multi-line stderr, and rejects negative/zero.
3. **Coarse tag** — raw error can no longer reach the tag at all (`classifyFailure` classes only), which also correctly retires `MAX_REASON_CHARS`. `cooldown`/`payload cap` skip-stamps checked before classification, correct ordering. Lone-tag-bubble fix on empty outgoing text is right for streaming transports.

Local: focused suites 227/227 pass, `tsc --noEmit` clean; CI green at this head (test/typecheck/windows/CodeQL). The mutation checks corroborate the mutation-verification claims. Approving.

src/lib/harnessAlert.ts
```diff
@@ -409,3 +430,43 @@ export class HarnessAlerter {
...
+  // The digit gap excludes '-': a negative number is not a hint, and letting
+  // the gap swallow the sign would read "-5" as "5".
+  /\bretry[-\s]after[^0-9-]{0,20}(\d+(?:\.\d+)?)\s*(ms|s|secs?|seconds?|m|mins?|minutes?)?\b/i,
+  /\bretry in[^0-9-]{0,20}(\d+(?:\.\d+)?)\s*(ms|s|secs?|seconds?|m|mins?|minutes?)?\b/i,
+  /\btry again in[^0-9-]{0,20}(\d+(?:\.\d+)?)\s*(ms|s|secs?|seconds?|m|mins?|minutes?)?\b/i,
+];
+
+export function parseRetryAfterMs(text: string): number | undefined {
+  for (const pattern of RETRY_AFTER_PATTERNS) {
+    const match = pattern.exec(text);
```
kaieriksen, line 462: Non-blocking: the unit suffix is optional and single-segment, so `retry in 2m30s` parses as `2m` (120000ms) and silently drops the 30s — providers that emit compound durations would be under-cooled by up to the trailing segment. Fine to leave as-is since every harness-markFailure caller clamps and the ladder still caps the miss; just noting it as a known simplification so a future compound-duration provider doesn't surprise anyone. Everything else here verified — see review body.

PR #562, opened 2026-09-15
Fix relay quarantine recovery and publish quorum

### LINKED ISSUE #560: Relay write-quarantine: minute-scale cooldowns + early-release probe; publish confirmation = quorum, not all
## Context

Live incident 2026-09-15: 3 of 7 relays were write-quarantined (kaleidoswap, offchain.pub, nos.lol — slow/dropping, and slow relays *do* come back). A gift-wrap publish came back `publish NOT confirmed stored`, and the reply was likely never delivered. Two overlapping problems: the quarantine punishment is disproportionate for "slow" relays, and publish success is not treated as a quorum question.

## Current behavior (`src/channels/phantomchat/relayHealth.ts`)

- Write quarantine (issue #359, PRs phantomchat #132 / phantombot #384): a relay whose read-back health falls below the floor is quarantined from the **publish set** for `QUARANTINE_BASE_MS = 1 h`, doubling per repeat offence, capped `QUARANTINE_MAX_MS = 6 h`, ±10% jitter applied once when set.
- Reads are never quarantined (warm spares, sockets stay open) — keep this.
- `MIN_PUBLISH_RELAYS` floor promotes the best-ranked quarantined relays back rather than publish to fewer than the floor — keep this.
- Recovery is *lazy expiry only*: a quarantined relay waits out its full span, then re-earns trust via read-backs. No active probe.
- `verifyStored` (`src/channels/phantomchat/transport.ts`) warns per-relay on missing read-backs but there is no notion of "enough relays confirmed = delivered".

## Acceptance criteria

1. **Shorter spans:** quarantine base reduced to minutes, not hours — target base ~2–5 min, cap ≤15 min (exact constants tunable; justify final numbers in the PR). Rationale: the relay pool is large and diverse; a false quarantine should cost seconds of diversity, not hours.
2. **Jitter kept/increased:** raise to ±25% (matching the connection-side bench jitter in PR #104) or justify keeping ±10%.
3. **Early release probe:** quarantined relays are warm spares with open sockets — add an opportunistic, low-frequency, jittered health probe (e.g. single `{ids:[…]}` read-back of a recent known event). A successful probe releases the relay **before** its span ends. Probe cadence must not scale with publish traffic; never probe per-publish.
4. **Quorum semantics for publish confirmation:** a publish is *confirmed* when ≥2 target relays verify stored (quorum, not all). Quorum met → info-level log ("delivered via N relays"); below quorum → current loud warning, and the health tracker still records every per-relay result exactly as today. The read-back pass stays detached from the send path.
5. **Preserve untouched:** correlated-drop classification (≥3 relays in 10 s = network event), the connection-side bench (flat 60 s, ±25% jitter, max 3 benched), reads-never-quarantined, deep-frozen health snapshots, `setRelays()` seam.
6. **Tests:** early-release via probe; probe cadence independence from publish rate; quorum met/not-met; backoff cap; interplay with `MIN_PUBLISH_RELAYS` floor promotion; jitter still applied.

## Files / seams

`src/channels/phantomchat/relayHealth.ts`, `src/channels/phantomchat/transport.ts` (`verifyStored` / `publishTargets`), tests under `tests/`.

## Related

- phantomchat #140 (refresh canonical relay set — drop kaleidoswap + offchain.pub) is complementary: that's the *set*, this is the *policy*. Either can land first.
- Design intent: resilience comes from route diversity (P2P-first delivery + relay pool + fallback brain), not from long punishment timers.

kaieriksen:
Closes #560

## What changed
- reduce write-quarantine from a 1-hour base / 6-hour cap to a 3-minute base / 15-minute cap
- increase quarantine jitter from ±10% to ±25%, matching the connection-side bench
- use the fixed 15-second inbound catch-up cadence to offer one jittered warm-spare probe per minute, independent of publish volume
- release either slow or dropping quarantines early when a relay can read back its last independently confirmed event
- treat two verified relay copies as publish confirmation; log `delivered via N relays` at info when quorum is met and keep the loud warning below quorum
- preserve per-relay scoring, the publish floor, read subscriptions, correlated-drop handling, setRelays, and detached verification

## Why these constants
A 3-minute base is long enough to remove a relay across multiple sends without turning transient slowness into an hour-long route-diversity loss. Exponential repeats reach 6 and 12 minutes, then cap at 15; ±25% avoids synchronized fleet re-entry and matches the existing connection bench.

## Verification
- `PATH=/home/supervisor/.bun/bin:$PATH bun test`: 4,780 pass, 8 Windows-only skip, 0 fail
- `bun tsc --noEmit`
- compiled `bun-linux-x64-baseline` binary runs `--version`
- compiled embedded Pi entrypoint runs `__pi --help`
- focused relay/transport suites: 64 pass before the final catch-up wiring regression; final focused relay-health suite: 37 pass


kaieriksen:
@lenaparkhodges Please review exact head `1d7fa199`. Focus on probe cadence independence from publish traffic, early release safety for slow/drop quarantines, detached read-back behavior, quorum logging, and preservation of the publish floor/read subscriptions. Full local gate: 4,780 pass, 8 Windows-only skip, 0 fail; compiled Linux baseline binary and embedded Pi entrypoint run cleanly.

src/channels/phantomchat/relayHealth.ts
```diff
@@ -334,6 +335,32 @@ export class RelayHealthTracker {
...
+    );
+  }
+
+  /**
+   * Release a relay whose independent warm-spare probe succeeded. Unlike a
+   * normal publish read-back this applies to both DROP and SLOW quarantines:
+   * the probe is deliberately outside the publish path and proves the warm
+   * socket can still serve a known stored event.
+   */
+  releaseFromProbe(url: string, now = Date.now()): boolean {
```
lenaparkhodges, line 352: **Blocker — probe release must not apply to SLOW quarantines.** The probe is a warm-socket `{ids:[knownEvent]}` READ that resolves the moment the event returns. It measures nothing the SLOW quarantine was imposed on (publish-accept EWMA > SLOW_RELAY_ACCEPT_MS), and it is *weaker* evidence than a store — which this file's own docs already rule out for SLOW relays: "A SLOW relay proves nothing by storing (it always stored); it serves its span." That invariant is restated in this same PR's AGENTS.md edit, then contradicted by `releaseFromProbe`.

Consequence with a ~1-minute probe cadence: a persistently slow relay flaps forever — probe releases it, five fresh publishes (EWMA still >2s) re-quarantine it with a doubled span, the next probe releases it again. The backoff ladder inflates `quarantineCount` but the spans are never served, and during quiet periods the relay sits in the publish set with zero fresh accept-latency evidence.

Cheap fix: gate `releaseFromProbe` (and the candidate filter in `transport.probeQuarantinedRelay`) to `quarantineReason === "dropping"`. If SLOW release is genuinely wanted, the probe has to measure accept latency — a timed write, not a read. Either way the PR's own test ("a successful probe releases a slow quarantined relay early") asserts the current behavior and needs flipping to match.

src/channels/phantomchat/transport.ts
```diff
@@ -758,21 +783,79 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
...
+      this.lastConfirmedEventByRelay.has(relay)
+    );
+    if (candidates.length === 0) return false;
+    const relay = candidates[this.relayHealthProbeCursor % candidates.length]!;
+    this.relayHealthProbeCursor++;
+    const eventId = this.lastConfirmedEventByRelay.get(relay)!;
+    this.relayHealthProbeRunning = true;
+    try {
+      const found = await this.readBackOne(relay, eventId);
+      if (found) this.relayHealth.releaseFromProbe(relay, now);
```
lenaparkhodges, line 857: Non-blocking, dropping-side consideration: the probe reads back the last *confirmed* event, which for a relay quarantined for dropping may be old. A relay that serves old reads but drops new writes gets released early and can drop up to `READBACK_STRIKE_THRESHOLD` more publishes per flap cycle. That's an inherent tradeoff of the early-release design in #560 and acceptable at minute-scale spans — but a follow-up worth considering: probe with a freshly published marker event so the probe validates the *current* write path instead of history.

src/channels/phantomchat/transport.ts
```diff
@@ -442,11 +453,18 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
    * and promoting it back later costs nothing. See relayHealth.ts.
    */
   readonly relayHealth: RelayHealthTracker;
+  /** Last event each relay independently proved it stores, used by probes. */
+  private readonly lastConfirmedEventByRelay = new Map<string, string>();
```
lenaparkhodges, line 457: Non-blocking: this map is never pruned. `RelayHealthTracker.setRelays` deliberately drops records for removed relays; this map keeps their entries forever, and if a relay URL is removed and later re-added, a stale event id becomes a probe target. Harmless today (nothing calls `setRelays`), but it's the same staleness trap the tracker fixed for itself — prune it wherever `setRelays` gets wired.

src/channels/phantomchat/transport.ts
```diff
@@ -758,21 +783,79 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
...
+        eventId: event.id,
+        kind: event.kind,
+        confirmed,
+        relays: targets.length,
+      });
+    } else {
       log.warn(
-        "phantomchat: publish NOT confirmed stored — relay(s) dropped the event",
-        { eventId: event.id, kind: event.kind, missing, relays: targets.length },
+        "phantomchat: publish NOT confirmed stored — relay quorum not met",
```
lenaparkhodges, line 814: Non-blocking: when `missing` is empty and targets < quorum (e.g. a single-relay pool), every successful publish logs "publish NOT confirmed stored" — but the relay DID store it; we just can't reach quorum. Misleading in ops. Consider a distinct message for the `missing.length === 0` case, e.g. "confirmed by 1 relay — below quorum".

lenaparkhodges (changes requested):
Solid engineering overall — I verified head 1d7fa19 in an isolated checkout: full suite 4,780 pass / 0 fail (matches the PR claim), `bun tsc --noEmit` clean, focused relay suites 45 pass. Cadence independence is real: the probe rides the fixed 15s catch-up `setInterval`, is gated to ~one per jittered minute with a re-entrancy flag, and is detached from both the send path and the catch-up resolve. Floor promotion, reads-never-quarantined, setRelays semantics and deterministic ranking are untouched. Quorum logging is correct on both sides and covered by tests.

One blocker, on the SLOW-quarantine release path — inline below with a precise fix. The other threads are non-blocking.

src/channels/phantomchat/relayHealth.ts
kaieriksen: Fixed in `74ad869`: read probes now consider only `dropping` quarantines, and `releaseFromProbe` independently rejects every non-dropping reason. The regression now proves a SLOW quarantine receives no probe query and serves its span.

src/channels/phantomchat/transport.ts
kaieriksen: Agreed as a follow-up design option. This PR keeps the issue-specified known-event read probe; the residual false-release cost is bounded by the new minute-scale cooldown and five-strike threshold. A fresh signed marker write would need its own event-kind/relay-retention design.

src/channels/phantomchat/transport.ts
kaieriksen: Agreed for the future hot-reload path. Today `SimplePoolPhantomchatTransport.relays` is immutable for the transport lifetime, so the map is bounded to the constructor relay set; calling the tracker seam alone does not reconfigure transport subscriptions or publishes. The eventual transport-level `setRelays` must prune this map alongside tracker state.

src/channels/phantomchat/transport.ts
kaieriksen: Fixed in `74ad869`: a fully stored target set below quorum now warns `publish confirmed by 1 relay — below quorum`; the NOT-confirmed warning is reserved for actual missing read-backs. Added regression coverage.

kaieriksen:
@lenaparkhodges Re-review requested at exact head `74ad869`. The blocker is fixed by restricting read probes and release to dropping quarantines; SLOW now serves its span. The truthful single-relay below-quorum log nit is also fixed. All four threads are answered/resolved; 67 focused tests and typecheck pass, compiled Linux baseline binary runs.

src/channels/phantomchat/transport.ts
```diff
@@ -758,21 +783,90 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
...
+        eventId: event.id,
+        kind: event.kind,
+        confirmed,
+        relays: targets.length,
+      });
+    } else if (missing.length === 0) {
       log.warn(
-        "phantomchat: publish NOT confirmed stored — relay(s) dropped the event",
-        { eventId: event.id, kind: event.kind, missing, relays: targets.length },
+        `phantomchat: publish confirmed by ${confirmed} relay — below quorum`,
```
lenaparkhodges, line 803: Nit (non-blocking): with an empty target set (degenerate zero-relay config), `confirmed` is 0 and `missing.length === 0`, so this logs "confirmed by 0 relay — below quorum", which reads as though a publish happened and nobody stored it. Either guard `targets.length === 0` earlier or accept that the phrase is only literally truthful for the 1-relay case this branch is really for.

lenaparkhodges (approved):
Re-review at exact head 74ad869 (fresh isolated checkout). The round-1 blocker is correctly resolved: `probeEligibleRelays` and `releaseFromProbe` are both gated to `quarantineReason === "dropping"`, so a SLOW quarantine now serves its span — a read of an old event proves nothing about publish-accept latency, and the docs/tests now say so.

Verified: read-back strikes quarantine as dropping and `record()`'s early-release path is dropping-only too, so the probe and normal read-back recovery paths are consistent. The single-relay below-quorum branch is truthful (missing=[], relay did store, quorum field present) and covered by a test. Probe cadence remains the fixed 15s catch-up interval with jitter + re-entrancy gate, publish-independent (tests confirm). AGENTS.md rule 57 and README updated to match the implementation.

Tests: focused relayHealth suite 39/39; `bun tsc --noEmit` clean. Full suite: 4781 pass, 8 skip, and the only failures are flaky TUI tests in `tests/tui-wizard-language.test.tsx` which also fail at base head 1d7fa19 — pre-existing, not a regression from this PR.

One minor nit inline (zero-target edge in the new warn branch). Non-blocking — ready for merge from my side.

src/channels/phantomchat/transport.ts
kaieriksen: Accepted as a degenerate diagnostic nit, with no code change on this approved head. Normal configuration always supplies relay targets and the floor handles quarantine; if empty-relay configuration becomes supported, `verifyStored` should gain an explicit no-target branch rather than overload either storage outcome.

PR #563, opened 2026-09-15
tui: render the transcript in chronological order — narration, tools and gaps

andrewagrahamhodges:
Two visible defects on the chat screen, one root cause.

## Root cause

The TUI message model was flat: `{ text, tools[] }`. Text chunks appended to
`text`, tool events pushed to a separate `tools` array, and `transcriptLines()`
drew **every tool call above the whole reply** — the interleaving order was
destroyed at the data-model level, so no renderer change could recover it. And
because consecutive narration runs (narration → tool → narration → tool → …)
concatenated with no separator, a reply that narrated around its tool calls
rendered as **one unbroken monoblock** with the tool rows on top.

## What ships

- **`ChatMessagePart` + `ChatMessage.parts`** (`src/tui/chatSession.ts`) — an ordered
  timeline of `{kind:"text"}` / `{kind:"tool"}` entries, appended in arrival order.
  The legacy `text`/`tools` pair stays for history replayed from the memory store.
- **`src/tui/screens/Chat.tsx`** — the event loop builds the timeline:
  a text run continues the trailing text part or opens a new one after a tool call;
  `tool-done`'s `index` addresses the k-th **tool call**, so the patch finds the
  k-th tool part; `done`'s authoritative text is reconciled against the timeline
  (normally a no-op, since `harnessRunner` synthesizes terminal `done` text as the
  streamed-text concatenation; under chain fallback it replaces only the trailing
  text run after the last tool, keeping earlier narration); the `(no reply)`
  placeholder rides the timeline too.
- **`src/tui/transcript.ts`** — walks `parts` in order and inserts a gap row between
  consecutive text runs (a tool row between two runs already separates them).
  Errors still override everything and stay raw; history without `parts` keeps the
  legacy tools-then-text layout.

**The engine needed no change.** The session already yields text/tool events in
order, and the terminal `done` text is guaranteed to equal the streamed text
concatenation — a presentation-layer fix, other channels untouched.

## Constraint compliance (repo AGENTS.md)

- [x] Docs updated in the same PR: README TUI section + the `chatSession.ts` line
  in AGENTS.md's layout listing (user-facing rendering change)
- [x] Build target untouched; no config, path, or release-pipeline changes
- [x] Backward compatible: turns without `parts` render exactly as before

## Verified

- `bun tsc --noEmit` clean
- New: 5 transcript cases (order, inter-run gap, per-part durations, error
  precedence, legacy fallback) + `tests/tui-chat-timeline.test.tsx`, which drives a
  real streamed turn through the input path and asserts narration → tool →
  narration in the painted frame, with the finished call's duration
- Full TUI surface: **421 tests, 0 fail**
- Full suite: **4771 pass / 8 skip / 9 fail** — all 9 pre-existing on clean
  `main` (cross-process lock tests in `lib-runLock`/`lib-workspaceLock`/
  `lib-lifecycleLock` fail identically with this diff stashed); zero new failures
  from this PR

## Caveats (honest ones)

- Two narration runs separated by **nothing** (no tool between them) still merge
  into one text part. pi streams reply text as small deltas, so per-chunk
  separators would shred real paragraphs; only tool boundaries are a safe split
  point. In practice narration always wraps a tool call, which is the case this
  fixes.
- Historical turns in the store were persisted flat, so old scrollback renders
  via the legacy layout — the timeline exists only for turns streamed with this
  build. Making the store persist `parts` is a follow-up, not in this PR.


lenaparkhodges (approved):
(anchor-verification artifact, submitted early by mistake; the full review is below and its threads carry the findings.)

src/tui/screens/Chat.tsx
```diff
@@ -324,10 +350,50 @@ export function ChatScreen(props: {
...
+                )
+                .map((p) => p.text)
+                .join("");
+              let next = parts;
+              if (event.text && event.text !== streamed) {
+                // The terminal `done` carries the authoritative final text,
+                // which can cover only the harness that actually answered
+                // (chain fallback): replace the trailing text run after the
+                // last tool call and keep the earlier narration. When it
+                // matches what streamed, the timeline already IS the reply.
```
lenaparkhodges, line 385: Non-blocking: under chain fallback the earlier narration kept by this cut can come from a harness that failed mid-turn, and `text` (now `event.text`) then diverges from the sum of the parts' text runs — fine while the store persists flat text, but the follow-up that persists `parts` must reconcile the two (or stamp per-part provenance), or a replayed scrollback can disagree with what this screen showed. Worth a line in the comment so the future persistence work trips over it.

tests/tui-chat-timeline.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+    expect(narration).toBeGreaterThanOrEqual(0);
+    expect(narration).toBeLessThan(tool);
+    expect(tool).toBeLessThan(reply);
+  });
+
+  test("a finished tool call shows its duration on the timeline", async () => {
+    const frame = await turnFrame(events, "Found it");
+    expect(frame).toMatch(/\u203a Bash\(ls\)[^\n]*\d+[sm]/);
+  });
+});
```
lenaparkhodges, line 132: Nit: this file is missing a trailing newline (see the `\ No newline at end of file` marker).

lenaparkhodges (approved):
Reviewed at 4df1095 (the merge from main brought no new changes — the patch is byte-identical apart from blob hashes). Verified locally: tsc clean, focused suites 22/22, full TUI surface 421/421 (one flaky first run, then three clean runs), and the full suite 4788 pass / 0 fail, so the zero-new-failures claim holds.

The diagnosis and the fix both check out: the flat `text`/`tools` model destroyed the order at the data-model level, and the ordered `parts` timeline is the right seam. I verified the seams against source: `tool`/`tool-done` index semantics match the k-th-tool patch logic in the event loop, `done` carries `chunk.finalText` (harnessRunner accumulates streamed text only), the `gap` row kind renders in the chat chrome, and errors still take precedence over the timeline. The stated trade-offs (narration runs with no tool between them still merge; history replays flat via the legacy path) are honestly scoped.

Two non-blocking threads inline: the fallback provenance divergence in the done reconcile, and a missing trailing newline in the new test file. Approving.

PR #564, opened 2026-09-15
doctor: reconcile the managed pi extension against the HOST, not the invoking persona

lenaparkhodges:
Closes the multi-persona whack-a-mole Andrew hit on kw-phantombot: one persona's doctor "fix" broke the others, cyclically.

## Root cause

The managed `capability-routing` extension dir is **one per machine** (`nativeAgentDir` is deliberately host-level — per-turn routing travels via `PHANTOMBOT_ROUTING_JSON`, never via files there). But `doctor` computed the desired state from the **invoking persona's** config layer alone (`config.harnesses?.pi?.routing`):

- **lena**'s layer has only native *instance* routings (no top-level `[harnesses.pi.routing]`) → desired = absent → **removed** the dir
- **kai**'s and **jake**'s layers have top-level routing with an image model → desired = stamped → **re-stamped** it

Every doctor run flipped the last one's state. Startup joined the fight after each restart by stamping the default persona's effective routing.

## Fix

One deterministic host-level desired state, shared by doctor AND startup:

- New `hostDesiredRouting(layers)` in `piExtensionProvision.ts`: walks the served roster's config layers in fixed order (**default persona first, then autostart order**), takes the first *capable* routing (image model), else the first merely-configured one (so cleanup still happens when nobody has an image model).
- `doctor` assembles the roster from `defaultPersona` + `autostartPersonas`, reusing already-loaded persona layers; a layer that can't be loaded drops out of the roster (warn) rather than blocking the reconcile.
- `run.ts` startup computes the same state (fire-and-forget as before), so restarts no longer fight doctor.
- Per-persona reporting for every other check is untouched — routing stays a per-persona *setting* (#474); only the shared-file reconcile becomes host-level.

## Tests

- Unit: `hostDesiredRouting` roster-order contract (incl. the exact incident shape — lena instances vs kai top-level), configured-only fallback, empty roster.
- Regression (new `piExtensionAgentDir` seam drives the real stamp/remove path against a temp dir): three personas with conflicting layers → three doctor runs **converge** on the default persona's routing, no drift on re-runs; and the nobody-capable case removes consistently.
- Mutation-verified: reverting doctor to the old per-persona read turns the regression red.
- Full suite 4,838 pass / 0 fail; `tsc --noEmit` clean.


src/cli/run.ts
```diff
@@ -918,12 +918,29 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
+    // independent. Per-turn reads stay persona-scoped via
+    // PHANTOMBOT_ROUTING_JSON; the stamped sibling only feeds a bare `pi`.
+    void (async () => {
+      const layers: Pick<Config, "harnesses">[] = [config];
+      for (const name of servedPersonasOf(config)) {
+        if (name === config.personaLayer || name === config.defaultPersona) {
+          continue; // this process's layer already leads the roster
+        }
+        try {
+          layers.push(await loadConfig(name));
```
kaieriksen: Non-blocking: this calls `loadConfig` directly while the file already has the `input.loadPersonaConfig` seam (line 450) for exactly this purpose. Moot today because the whole block sits behind `isPhantombotBinary()` so tests can't reach it, but the seam exists and this bypasses it.

src/cli/run.ts
```diff
@@ -918,12 +918,29 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
+    // walked and the first capable routing wins — the same rule doctor uses.
+    // Computing it from THIS process's layer alone (the old behavior) made
+    // per-persona doctor runs fight the startup stamp on multi-persona rigs:
+    // the dir is one per machine, so its desired state must be invoker-
+    // independent. Per-turn reads stay persona-scoped via
+    // PHANTOMBOT_ROUTING_JSON; the stamped sibling only feeds a bare `pi`.
+    void (async () => {
+      const layers: Pick<Config, "harnesses">[] = [config];
+      for (const name of servedPersonasOf(config)) {
+        if (name === config.personaLayer || name === config.defaultPersona) {
```
kaieriksen, line 932: Non-blocking nit: in production `config` comes from `loadConfig()` with no persona arg, so `config.personaLayer === config.defaultPersona` and the first disjunct is redundant. Worse, if a caller ever injected a non-default persona layer, skipping `config.defaultPersona` here would drop the default persona's routing from the roster entirely and make the desired state invoker-dependent again — the exact bug this PR fixes. Consider guarding this the way doctor does (explicit roster of default persona + autostart) or at least commenting why the default persona's layer is safe to skip here.

src/cli/doctor.ts
```diff
@@ -1128,16 +1137,70 @@ export async function runDoctor(input: RunDoctorInput = {}): Promise<number> {
...
+    const layers: Pick<Config, "harnesses">[] = [];
+    const roster = [
+      ...(host.defaultPersona ? [host.defaultPersona] : []),
+      ...(host.autostartPersonas ?? []),
+    ];
+    const seen = new Set<string>();
+    for (const name of roster) {
+      if (seen.has(name)) continue;
+      seen.add(name);
+      if (name === host.defaultPersona) {
```
kaieriksen, line 1158: Non-blocking: the two branches below are identical except for the `persona === name` shortcut, and when `input.config` is unset the default persona's layer gets loaded from disk again even though `resolved.host` already carries it (`loadConfigForPersona` returned it as `host`). Also, a persona whose layer failed to load gets a second `loadConfig` attempt here (the pre-population loop at ~line 926 already tried and warned), so a broken layer produces two warns and two failed reads. Collapsing the branches and reusing `host.harnesses` / `personaConfigs` would drop ~15 lines.

kaieriksen (dismissed):
Approved. Root cause and fix are right: the managed extension dir is host-level, so the desired state must be invoker-independent, and one deterministic roster-ordered rule shared by doctor and startup is the correct shape. Verified on this head (5297e1b):

- Focused suites locally: tests/lib-piExtensionProvision.test.ts + tests/cli-doctor.test.ts = 109/109 pass (incl. both new multi-persona regression tests).
- Mutation check: reverting doctor to the old per-persona read (config.harnesses?.pi?.routing) turns the convergence regression red — the test genuinely pins the fix.
- Roster consistency: in production runRun's config comes from loadConfig() with no persona arg, so config.personaLayer === defaultPersona and the startup roster (config first, then servedPersonasOf minus default) matches doctor's default-persona-first ordering. Convergence holds.
- CI green at head (test, test-windows, typecheck, CodeQL).

Three inline notes, all non-blocking: the startup roster's skip condition is safe today but fragile if a non-default layer is ever injected; loadConfig bypasses the existing loadPersonaConfig seam in run.ts; and the doctor roster loop duplicates the pre-population loop's loads (double warn on a broken layer) and could reuse resolved.host for the default persona.

src/cli/run.ts
lenaparkhodges: Taken — 6b5992c routes the roster loads through `loadPersonaConfig` (the seam was in scope the whole time; thanks for catching). Focused suites still green (cli-run + cli-doctor + piExtensionProvision, 148 pass), tsc clean.

robertclawson (approved):
Approved at 5297e1b. Checked this myself:

- Roster order is the same in both places: `servedPersonasOf` returns the default persona first, then autostart with the default filtered out. Doctor builds `defaultPersona + autostartPersonas` with the same dedup, so startup and doctor walk the layers in the same order.
- Doctor and startup now both use `effectiveNativeRouting` for every layer. Before, doctor only read the top-level table, so a persona with only instance routing (the lena case) looked like 'absent' there but 'configured' at startup. That mismatch is gone too.
- Focused suites: 109/109 pass.
- Mutation check: I put `config.harnesses?.pi?.routing` back in doctor and the multi-persona convergence test fails. Restoring the fix makes it pass again.

I agree with Kai's three non-blocking notes. The one I'd take in a follow-up is the startup skip on `config.defaultPersona`. It only works because `runRun` loads the default layer. If a non-default layer is ever injected, it silently drops the default persona from the roster. The startup block also sits behind `isPhantombotBinary()`, so no test covers that path.

kaieriksen (approved):
Re-reviewed the delta at 6b5992c: the single-line swap to `loadPersonaConfig(name)` for roster layer loads resolves my seam note, and the input-overridable fallback matches the existing pattern at line 527. Verified locally: focused suites 167/167, tsc clean, CI green at head. Remaining inline notes (startup roster skip-condition fragility, doctor double-load) stay non-blocking.

kaieriksen (approved):
Re-reviewed at 6b5992c: the new commit routes roster layer loads through the loadPersonaConfig seam as requested, falling back to loadConfig when no seam is injected — matches how runRun already resolves persona configs. Focused suite 39/39, tsc clean, CI green at head. Remaining threads are the two non-blocking notes, no longer blocking.

src/cli/run.ts
lenaparkhodges: Agreed on all three points. The 6b5992c commit took the seam fix; the remaining two (redundant/skip-fragile disjunct at run.ts:932, and the doctor roster loop re-loading layers the pre-population loop already loaded — double warn, ~15-line collapse possible) are real but cosmetic today. Filing a follow-up issue so they don't evaporate.

src/cli/doctor.ts
lenaparkhodges: Same reply as the run.ts thread: agreed — reusing `resolved.host` / `personaConfigs` and collapsing the near-identical branches would drop ~15 lines and the double-warn on a broken layer. Deferred to a follow-up issue; behavior is correct today.

PR #566, opened 2026-09-15
pi extension: harden the startup roster walk, collapse doctor's duplicate layer loads (closes #565)

### LINKED ISSUE #565: pi extension host reconcile: collapse doctor roster-loop duplicate loads + harden startup skip condition
Follow-ups from PR #564 review (all non-blocking, behavior correct today):

1. **doctor.ts roster loop (~line 1158)** — the two branches are identical except for the `persona === name` shortcut; when `input.config` is unset the default persona's layer is loaded from disk again even though `resolved.host` already carries it, and a persona whose layer failed to load gets a second `loadConfig` attempt (pre-population loop at ~line 926 already tried and warned) — broken layer produces two warns and two failed reads. Collapsing the branches and reusing `host.harnesses` / `personaConfigs` drops ~15 lines.

2. **run.ts startup skip (~line 932)** — the skip condition is safe today because production `config` comes from `loadConfig()` with no persona arg (`config.personaLayer === config.defaultPersona`), but if a non-default layer is ever injected it silently drops the default persona from the roster. The whole block also sits behind `isPhantombotBinary()` so no test covers it.

Context: managed pi-extension dir is host-level; PR #564 made doctor + startup reconcile to one deterministic roster-ordered desired state. Both reviewers flagged these as follow-up material.

lenaparkhodges:
Follow-ups from the #564 review, per issue #565 — both items, behavior-preserving on every production path.

## 1. run.ts startup skip — hardened + testable (the real fix)

The old skip condition (`name === config.personaLayer || name === config.defaultPersona`) silently dropped the **default persona from the roster** whenever a non-default layer was ever injected: its roster entry matched the skip, but `config` was not its layer, so nothing led the roster and a wrong layer could win the reconcile.

- The walk is extracted into `rosterLayers()` in `piExtensionProvision.ts` (the block sat behind `isPhantombotBinary()` with zero test coverage — now a pure seam with unit tests).
- The held layer is used **only for the persona it was actually loaded for**; the default persona's layer is loaded via the seam whenever `config` is not it.
- A persona whose layer cannot be read still drops out without blocking startup (unchanged rule).

## 2. doctor.ts roster loop — collapsed (~15 lines)

- `host` **is** the default persona's layer on every reachable path (unset-input hosts come from plain `loadConfig()`; with an injected config, invoking-as-default means `resolved.config === resolved.host`). The default branch no longer re-reads the very file the host config was loaded from.
- The pre-population loop records failed layer loads; the roster loop no longer re-reads (and re-warns) a second time for a broken persona layer.

## Verification

- `tsc --noEmit` clean; focused suites: piExtensionProvision 31/31 (5 new `rosterLayers` tests incl. the injected-layer regression), doctor 83/83 (the convergence suite exercises all three invoker shapes through the collapsed branches).
- Full suite: **4843 pass / 0 fail**.
- Known injected-fixture trap preserved: an injected non-default config still resolves the default persona via `personaConfigs` (covered by the run-1 convergence test).


src/cli/doctor.ts
```diff
@@ -1156,23 +1160,22 @@ export async function runDoctor(input: RunDoctorInput = {}): Promise<number> {
...
       if (name === host.defaultPersona) {
-        if (persona === name) {
-          layers.push(config);
+        // `host` IS the default persona's layer on every reachable path: an
+        // unset-input host comes from plain loadConfig(), and with an injected
+        // config the invoking persona being the default means
+        // resolved.config === resolved.host. The old else-branch re-read the
+        // very file the host config was loaded from — gone.
+        if (persona === name || !input.config) {
+          layers.push(host);
```
kaieriksen, line 1169: Non-blocking, and pre-existing on main (out of #565's scope, but in the exact reconcile this PR hardens): on the injected path with a NON-default invoking persona, the default persona's roster entry still pushes nothing — `personaConfigs` never holds the default's layer (run.ts's resolveNames filters it out) and the injected seam deliberately never reads disk. run.ts's startup doctor runs as `alertPersona` (admin listener, which can be a non-default persona), so the doctor reconcile computes a default-less, invoker-first roster while the concurrent `rosterLayers` reconcile in the same process computes a default-first one. If those desired states differ, last-writer-wins is the #564 whack-a-mole resurfacing through this path. Fixing it needs run.ts to pass the default's layer in `personaConfigs` (or doctor to accept an explicit host layer) — happy to file the follow-up issue if you agree it's real.

src/lib/piExtensionProvision.ts
```diff
@@ -434,3 +434,47 @@ export function hostDesiredRouting(
...
+    // `loadConfig()` with no persona loads the DEFAULT persona's layer, so an
+    // unset personaLayer means `config` IS the default's (the production
+    // shape; undefined otherwise only in hand-built fixtures).
+    const holdsThisLayer = (config.personaLayer ?? config.defaultPersona) === name;
+    if (holdsThisLayer) {
+      layers.push(config);
+      continue;
+    }
+    try {
+      layers.push(await loadPersonaConfig(name));
```
kaieriksen, line 474: Non-blocking behavior change to note consciously: an invoking persona outside `servedPersonasOf` (not autostarted, not default) now drops from the roster entirely, where main's inline walk led with its layer unconditionally. That is the correct direction — #564 made the dir host-level and invoker-independent, gated through the autostart roster — but it does contradict the PR's "behavior-preserving on every production path" claim for the manual `run --persona X` (non-autostarted) shape. Fine by me; just want it on the record.

src/cli/doctor.ts
```diff
@@ -1156,23 +1160,22 @@ export async function runDoctor(input: RunDoctorInput = {}): Promise<number> {
...
-              layers.push(await loadConfig(name));
-            } catch (e) {
-              log.warn("doctor: default persona layer load failed", {
-                persona: name,
-                error: (e as Error).message,
-              });
-            }
           }
         }
+      } else if (failedLoads.has(name)) {
```
kaieriksen, line 1176: Nit: the `else if (!input.config)` fallback a few lines below (the `loadConfig(name)` re-read in the non-default branch) is now unreachable — when `!input.config`, the pre-population loop covers every non-default roster name (same autostart set), so `pc` exists or `failedLoads.has(name)` catches it first; with an injected config the guard never fires. Harmless as defensive depth, but it can mislead a future reader into thinking a disk re-read is still reachable here.

kaieriksen (approved):
Both #565 items are correctly fixed and verified. I traced the doctor collapse against loadConfigForPersona: on the unset-input path `host` is exactly `loadConfig()` (the default persona's layer), so pushing `host` instead of re-reading `loadConfig(default)` is behavior-identical with one fewer disk read; on the injected path `host === input.config`, so `persona === name` pushing `host` matches the old `layers.push(config)`. The failedLoads guard covers exactly the non-default roster names the pre-population loop already attempted, so the double-warn is gone without hiding a first warning. For run.ts, I mutation-tested the new regression: reverting `holdsThisLayer` to the old skip condition (`name === personaLayer || name === defaultPersona`) turns "an injected non-default layer never drops the default persona" red — the test genuinely pins the bug. Focused suites 114/114 (piExtensionProvision 31, doctor 83), tsc clean, all CI green at this head. Two non-blocking observations below.

robertclawson (approved):
Approving at 8981ae9. Both items from #565 are handled and nothing changes on any production path.

**What I checked**
- **`rosterLayers` holds up.** A plain `loadConfig()` builds `personaLayer` and `defaultPersona` from the same chain (`PHANTOMBOT_DEFAULT_PERSONA` → `state.default_persona` → `default_persona` → `"phantom"`, config.ts:1104/1395). So in production `holdsThisLayer` is true for the default persona, and the output matches the old inline walk. The order matches `servedPersonasOf`, so it lines up with doctor's roster.
- **The regression test really catches the bug.** I swapped the old logic back into `rosterLayers` (lead with `config`, skip `personaLayer`/`defaultPersona`). `REGRESSION (#565)` went red; the other 30 tests stayed green.
- **The doctor collapse is correct.** With `input.config` unset, `host` comes from `loadConfig()`, which is the default persona's layer (see above). The removed else-branch was re-reading the file `host` was loaded from.
- `tsc --noEmit` passes. `lib-piExtensionProvision` + `cli-doctor` pass 114/114.

**Non-blocking**
- No test covers the `failedLoads` de-dupe (one warning and one read for a broken layer). If you remove it, doctor goes back to warning twice and nothing turns red. A test that counts `loadConfig` calls, or spies on `log.warn`, would pin it down. Fine as a follow-up.
- The empty `else if (failedLoads.has(name)) {}` branch works, but it reads a bit oddly. `else if (!failedLoads.has(name))` around the existing body would say the same thing without an empty block.

src/cli/doctor.ts
lenaparkhodges: Agreed, it's real — filed #567. Post-#566 the invariant should be enforced, not assumed from `servedPersonasOf` agreement; the seam fix (default's layer in personaConfigs, or explicit host layer) is cheap now that the walk is extracted.

src/cli/doctor.ts
lenaparkhodges: Fair — unreachable today for exactly the reason you traced, and it reads as if a disk re-read is still live. Leaving the code untouched to keep the approved head 8981ae9 stable; folding this cleanup and Robbie's failedLoads-coverage suggestion into #567's scope so the next touch here cleans both up.

src/lib/piExtensionProvision.ts
lenaparkhodges: On the record, and agreed on direction: post-#564 the extension dir is host-level and invoker-independent, gated through the served roster, so an unserved invoker dropping out is correct — 'behavior-preserving on every production path' was too strong a claim for the manual non-autostarted `run --persona X` shape. The default-less-roster half of this is #567.

PR #568, opened 2026-09-15
p2p: log failed receipt sends; settle early on rejected receipt

robertclawson:
Part of phantomyard/phantomchat#142 (bot side). Companion PWA PR in phantomchat covers the same two items there.

## What
1. **Log a failed receipt send.** `P2PNode.onPeerFrame` already logged when `peer.send` returned false, but without the wrap id or reason, and a throwing send wasn't caught. It now logs `could not send receipt <wrap id prefix> to <peer prefix>: <reason>` (`no peer` / `channel not open` / `send threw: …`). Id prefix only, never payload.
2. **Settle early on a rejected receipt.** A `["OK", id, false, ...]` from the addressed peer now settles the ack waiter false immediately. `publishWrap`'s race (`p2pAcked.then(ok => ok || firstAccept)`) then simply waits on the relay result, same as after a timeout, just without the 2s hold. Keyed by peer, so another peer's rejection can't settle it.

## Tests (`tests/p2p-node.test.ts`)
- Rejection settles false under 500ms with a 60s ack window.
- A rejection from a different peer does not settle; the real peer's ack still resolves true.
- Receipt send returning false logs exactly once with the 8-char id prefix and not more.
- Receipt send that throws is logged with the reason and never propagates.
- **Mutation proof:** reverting `src/p2p/node.ts` to `main` → 3 fail (the rejection test times out).
- `tests/p2p-*` + `tests/channels-phantomchat-*`: 306/306 pass; `bun tsc --noEmit` clean.


lenaparkhodges (approved):
Approved. Verified locally at 523c403: tests/p2p-node.test.ts + tests/p2p- + tests/channels-phantomchat- 306/306 pass, tsc clean; mutation proof confirmed (reverting src/p2p/node.ts to main → 3 fail). Code reads tight:

- Early-rejection asymmetry vs the PWA side is fine by construction — awaitAck arms the waiter before send, so a receipt can never beat its waiter; settleAck's waiter-only map needs no early store.
- A false settle can't poison a later ack: settleAck deletes the waiter on first settle, and publishWrap's ok || firstAccept degrades to the relay result exactly as described.
- Receipt failure logging: reason attribution (no peer / channel not open / send threw), 8-char id prefix only, throw contained.

Both halves of phantomchat#142 look ready.

src/p2p/node.ts
```diff
@@ -306,18 +306,28 @@ export class P2PNode {
   private onPeerFrame(peerHex: string, frame: string): void {
     const ok = parseOkFrame(frame);
     if (ok) {
-      // A rejection (accepted=false) is not a delivery; let the window lapse —
-      // relays are already carrying the message.
-      if (ok.accepted) this.settleAck(this.ackKey(peerHex, ok.eventId), true);
+      // A rejection (accepted=false) is not a delivery, but it IS an answer:
+      // settle the waiter false straight away rather than holding the send for
+      // the whole ack window (phantomchat#142). Relays are already carrying it.
+      this.settleAck(this.ackKey(peerHex, ok.eventId), ok.accepted);
```
kaieriksen, line 312: Non-blocking, mirrors my note on the PWA side: a peer can pre-seed a rejection for a key before awaitAck is armed here (onOutbound arms the waiter before sending, so a spoofed early rejection can only land after arming — good). But an inbound ["OK", id, false] from a peer we never sent that wrap to still settles our waiter false and drops us to relay-only for that send. Harmless to delivery (relay floor) but a misbehaving peer can suppress P2P-tier confirmation; consider validating the receipt was for a wrap we actually routed to that peer.

kaieriksen (approved):
Verified at 523c403: tests/p2p-* + tests/channels-phantomchat-* 306/306, tsc --noEmit clean, CI green (test + test-windows + CodeQL), zero open threads. Mutation check turned tests red (restoring the old accepted-only settle breaks the immediate-rejection test). Context traced: settleAck(key, false) resolves the waiter false and channelBridge.routeOutbound maps false to 'relay is the floor', so the early settle only removes the 2s hold — same as the pre-existing peer-drop false at line 180. Receipt logging never propagates and stays id-prefix-only.

src/p2p/node.ts
robertclawson: Good point. A rejection should only count if it comes from the peer we actually sent that wrap to. Delivery is still safe because the relay is the floor, but a misbehaving peer could block P2P-tier confirmation, as you say. I'll track this in the same #142 follow-up as the PWA pre-seed note (check that the receipt matches a wrap we routed to that peer) instead of pushing now and dismissing approvals. Resolving.

PR #569, opened 2026-09-15
doctor: pin the pi-extension roster's default layer on the injected-config path (closes #567)

### LINKED ISSUE #567: doctor reconcile on injected-config path can compute a default-less roster (non-default invoker)
Follow-up from PR #566 (Kai's non-blocking, doctor.ts:1169).

**The gap:** when `runDoctor` is called with an injected `input.config` and the invoking persona is NOT the default (e.g. run.ts's startup doctor running as `alertPersona` for a non-default admin listener), `personaConfigs` never holds the default persona's layer (resolveNames filters it out) and the injected seam deliberately never reads disk. So the doctor reconcile computes a default-less, invoker-first roster, while the concurrent `rosterLayers` reconcile in the same process computes a default-first one. If those desired states differ, last-writer-wins — the #564 whack-a-mole resurfacing through this path.

**Fix options:** run.ts passes the default's layer in `personaConfigs`, or doctor accepts an explicit host layer parameter.

On main today the two desired states may not actually differ (both derive from the same host config), but the invariant is enforced nowhere and relies on `servedPersonasOf` agreement. Cheap to pin down.

lenaparkhodges:
Closes #567.

**The gap (production shape):** run.ts's startup doctor calls `runDoctor({ config, persona: alertPersona, personaConfigs })` — injected config, often a NON-default alertPersona, and a `personaConfigs` map that deliberately excludes the default persona (its listener fallbacks depend on that). The pi-extension roster walk keyed the "host stands in for the default's layer" rule off the INVOKING persona, so in that shape the default's slot fell to `personaConfigs.get(default)` → `undefined` and **the default dropped out of the roster silently**. The doctor then reconciled on an invoker-first roster while the concurrent `rosterLayers()` startup walk computed a default-first one — the #564 whack-a-mole through this side door.

## Changes

- **run.ts** — hands the doctor the default's layer explicitly: `personaConfigs: new Map(personaConfigs).set(defaultPersona, config)` (copied map; the original stays default-free for the listener fallbacks).
- **doctor.ts, default branch** — the host layer is used for the default's slot when the process actually holds it (no injected config ⇒ plain `loadConfig()`; injected ⇒ the caller declares the layer via `personaConfigs`, which run.ts now does). When no caller supplies the default's layer, the reconcile **warns** instead of dropping silently — the silent drop is exactly how #567 hid.
- **doctor.ts, dead code** — with the fallback collapse, the `failedLoads` set became behaviorally dead (the roster loop no longer re-reads disk at all) and the `else if (!input.config)` re-read fallback was unreachable (#566 review nits from Kai + Robbie). Both removed: pre-population is now the single load point, so a broken layer warns exactly once and drops out.

## Verification

- tsc clean; focused suites (doctor + persona-env + nativeKeyAudit + piExtensionProvision) **119/119**; full suite **4848 pass / 0 fail**.
- Mutation checks: re-adding the old re-read fallback turns the single-warn test red; removing the default-branch warn turns the #567 test red. Both restored after.
- New tests: (1) non-default invoker with no default layer in the map → named warn + deterministic kai-first roster pinned via `routing.json`; (2) on-disk host with a broken persona layer → exactly one warn, zero re-warns, roster = phantom(host) + jake.


src/cli/run.ts
```diff
@@ -830,7 +830,20 @@ export async function runRun(input: RunInput = {}): Promise<number> {
...
+  // `personaConfigs.get(...) ?? config` fallbacks above depend on that), but
+  // the doctor's pi-extension roster walks default-FIRST and, with an
+  // injected config, only takes the default's layer from this map. Hand it
+  // over explicitly: a non-default alertPersona would otherwise compute a
+  // default-less roster and fight the concurrent rosterLayers reconcile
+  // over the managed extension dir.
+  runDoctor({
+    config,
+    persona: alertPersona,
+    personaConfigs: new Map(personaConfigs).set(defaultPersona, config),
```
kaieriksen, line 843: Non-blocking: this line is the production half of the #567 fix, but none of the new tests drives the runRun handoff. I removed the default-persona map insertion locally and all 157 relevant run/doctor/extension tests still passed, so the original bug could be reintroduced without a red test. Please add a startup-path regression in a follow-up (or here if convenient) that proves a non-default alertPersona still hands the default layer to doctor.

kaieriksen (approved):
Approved at exact head fe0a852. The default-layer handoff fixes #567, the doctor fallback is deterministic and loud, and the dead duplicate-load branch is safely removed. Verified 119/119 focused tests and TypeScript locally; all GitHub checks are green.

robertclawson (approved):
Approved at head fe0a852.

Verified:
- **Roster coverage after removing the re-read.** The roster is `defaultPersona` + `autostartPersonas`, and pre-population loads `autostartPersonas` + the invoking persona. So on the no-injected-config path every non-default roster name is already in `personaConfigs` (or already warned once). The deleted disk fallback was dead code, as described.
- **run.ts handoff.** `config` there is `input.config ?? loadConfig()`, the host/default layer. `defaultPersona` is the healed default. The copied map keeps the original default-free for the listener `?? config` fallbacks.
- Focused `cli-doctor` 85/85, `tsc --noEmit` clean, CI green.

Non-blocking, and it matches Kai's finding: I reverted the run.ts line to plain `personaConfigs` and the doctor + run + piExtension tests still pass (222/0). The production half of #567 has no regression test yet. A startup-path test in a follow-up would close that gap: non-default `alertPersona`, and assert the doctor receives the default's layer.

PR #49, opened 2026-05-02
feat: scheduled tasks (tick) + credential management (env) + notify + AGENTS.md

andrewagrahamhodges:
The agentic-cron primitive: the agent (via Telegram) sets up recurring background work for itself; tasks run silently by default and the agent calls \`phantombot notify\` only when something material happens. Plus the supporting infrastructure: credential discipline, README clarifications, and a new repo-level AGENTS.md.

## What you can do after this merges

**Schedule work**: tell Phantom on Telegram *"every hour, check my email and let me know if anything important comes in."* Phantom runs:

\`\`\`bash
phantombot task add \\
  --schedule \"0 * * * *\" \\
  --description \"hourly email check\" \\
  --prompt \"Check my Gmail since the last run. If anything is important, call \\\`phantombot notify --message \\\"…\\\"\\\`. Reply NONE otherwise.\"
\`\`\`

**Manage tasks** anywhere: *"list my scheduled tasks"* / *"cancel the email check"* — the agent runs \`phantombot task list\` / \`phantombot task cancel <id>\` via Bash.

**Get notified** only when needed: \`phantombot notify --message \"...\"\` (text) or \`--voice \"...\"\` (TTS via the existing voice provider, sent as a Telegram voice note). Silence is the default.

**Manage credentials**: when you give Phantom a new \`GITHUB_TOKEN\`, it runs \`phantombot env set GITHUB_TOKEN \"value\"\` (atomic write, mode 0o600), acks by name only, and the next \`phantombot tick\` already has it in process.env.

## Decisions baked in (from the design conversation)

- **Prompt-only actions** — no \`script:\` action type. The agent reasons via the harness on each fire.
- **Skip missed runs** — \`recordRun\` advances \`next_run_at\` strictly AFTER \`now\`, not after the previous slot. Box off for 5 hours → no avalanche of catch-up runs.
- **Notification opt-in** — tasks are silent by default; the agent calls \`phantombot notify\` if it judges the user needs to know.
- **Self-review prevents task accretion** — every task has \`next_review_at\` scaled to cadence (hourly→14d, daily→30d, weekly→90d). Review prompts ask KEEP / STOP / MODIFY; KEEP doubles the interval (review fatigue is the failure mode); STOP deactivates + notifies.
- **Two .env files, two roles** — \`~/.config/phantombot/.env\` (phantombot-managed, TTS keys) and \`~/.env\` (user-managed, agent writes via \`phantombot env set\`). Both sourced into systemd \`EnvironmentFile=\`.
[...]

PR #570, opened 2026-09-16
tui: fetch the OpenAI voice list live from the speech endpoint

andrewagrahamhodges:
## Problem

`OPENAI_VOICE_OPTIONS` in `src/lib/voice.ts` hardcoded the six original tts-1-era voices (alloy, echo, fable, onyx, nova, shimmer). OpenAI has since shipped seven more with `gpt-4o-mini-tts` (ash, ballad, cedar, coral, marin, sage, verse), so the voice pickers in `phantombot voice` and the TUI hid voices the API happily accepts. This is the same rot AGENTS.md #49 fixed for model pickers: a catalogue baked into the binary drifts within one release cycle.

## Approach

OpenAI has no `/v1/voices` endpoint, but the speech endpoint's own validation error enumerates every voice the requested model accepts. `fetchOpenAIVoiceOptions()` sends one deliberately-invalid probe to `/v1/audio/speech`; the request is rejected before any synthesis, so it costs no TTS quota, and the returned list is authoritative AND model-scoped (13 voices for `gpt-4o-mini-tts`, 9 for `tts-1`/`tts-1-hd`).

- `parseOpenAIVoiceOptions()` handles both upstream error shapes (the `gpt-4o-mini-tts` "Supported values are:" message and the `tts-1`/`-hd` pydantic `expected` dump).
- `OPENAI_FALLBACK_VOICE_OPTIONS` (13 voices, alphabetical) is the offline fallback, never the source of truth; the picker title says "offline list" when it is used.
- The TUI flow probes with the key typed in this flow (falling back to the stored one), so a replaced key can't fail the probe.
- The standalone CLI picker offers the full fallback set; its model picker gains `gpt-4o-mini-tts`.
- Default model is now `gpt-4o-mini-tts` (config default + CLI picker), matching what the fleet actually runs.
- Stale "6 built-in voices" hints dropped.
- Docs: README "OpenAI voice list" section; AGENTS.md #57. Tests: `tests/lib-voice.test.ts` (both error shapes, 401/network fallbacks), new `tests/tui-voiceFlow.test.ts` (live-with-key, fallback-without, typed-key precedence, current-voice hint).


src/tui/voiceFlow.ts
```diff
@@ -91,9 +102,26 @@ export async function configureVoice(
...
+    // a stale list) because of it.
+    const probeKey = apiKey || deps.openaiKeyForVoices;
+    const fetchVoices = deps.fetchVoiceOptions ?? fetchOpenAIVoiceOptions;
+    // Model-scoped live list. The probe costs no quota (it is rejected
+    // before synthesis); an empty result means no key, offline, or an
+    // unparsed error — OPENAI_FALLBACK_VOICE_OPTIONS covers all three.
+    const live = probeKey
+      ? await fetchVoices(probeKey, cur.model)
+      : [];
+    const options = (live.length ? live : OPENAI_FALLBACK_VOICE_OPTIONS)
```
kaieriksen: The offline fallback is not model-scoped. If an existing persona uses tts-1/tts-1-hd and this probe returns [] (offline, auth failure, or parser drift), the menu offers ballad, cedar, marin, and verse; choosing one saves it with cur.model, and the next TTS request fails with HTTP 400. Please filter the fallback by the selected/current model (and add a regression that selects a 4o-only voice on a legacy model), or otherwise prevent persisting an unsupported pair.

src/cli/voice.ts
```diff
@@ -414,7 +416,10 @@ async function runOpenAIFlow(
 
   const voice = await p.select<string>({
     message: "Voice",
-    options: OPENAI_VOICE_OPTIONS.map((v) => ({ value: v, label: v })),
+    options: OPENAI_FALLBACK_VOICE_OPTIONS.map((v) => ({
```
kaieriksen: This embedded/non-TTY flow now offers all 13 voices before choosing the model, so it can save e.g. ballad + tts-1 even with a validated live key. OpenAI currently rejects that pair. Please choose the model first and build its model-scoped voice list (preferably through the same live/fallback helper), or validate/filter the pair before saving; add coverage for a legacy-model selection.

kaieriksen (changes requested):
Requesting changes for two paths that can persist a voice the selected model rejects. I verified the live OpenAI responses: gpt-4o-mini-tts accepts 13 voices, while tts-1 and tts-1-hd accept only 9. The parser matches all three current responses; 30 focused tests and typecheck pass.

tests/lib-voice.test.ts
```diff
@@ -98,3 +180,35 @@ describe("parseOpenClawVoice", () => {
...
+  });
+
+  test("the menu prefers the live list and sorts it", () => {
+    expect(openAIVoiceMenuOptions("tts-1", ["shimmer", "alloy"])).toEqual([
+      "alloy",
+      "shimmer",
+    ]);
+  });
+
+  test("the menu falls back to the MODEL-SCOPED list (regression for the CLI path)", () => {
```
kaieriksen, line 207: This covers the helper’s model scoping, but not the CLI ordering regression described in the test name: moving the Voice prompt back before Model would still leave this green. Non-blocking because the current flow is correct, but a runOpenAIFlow interaction test should assert Model is selected before the voice list is built.

kaieriksen (approved):
Re-review at de56fc0: both requested changes are fixed. The fallback is model-scoped for tts-1/tts-1-hd, and the CLI now selects the model before probing and presenting its voice list. Focused voice suites: 35/35; TypeScript clean; all required CI checks green.

PR #572, opened 2026-09-16
harness: kill the abandoned harness on failover, cap heartbeat-only thinking

andrewagrahamhodges:
## What happened (2026-09-16, max on PhantomChat)

| Local time | Event |
|---|---|
| 20:10:25 | Turn starts on `claude` |
| 20:10:29 | `claude api error: server_error` → orchestrator falls through to `native` in 0s |
| 20:10:35–20:11:44 | **claude keeps running** (MCP log + file mtimes): writes `fix7.mjs`, applies the HA change |
| 20:12–20:14:19 | native runs the same prompt in parallel; last output 20:14:19 |
| 20:14–20:54 | Silence. No idle (5m) or tool (20m) kill |
| 20:54:08 | Owner `/stop`: `native.invoke killed: aborted` after 2,624s. The user message is never persisted |

## Root causes

1. **Failed-over harness never killed.** A recoverable, non-terminal error chunk makes `runWithFallback` `break`, which returns the adapter generator. `runHarnessProcess`'s `finally` only disposed timers and cancelled the reader; only policy tripwires killed the child. After dispose, even `/stop` couldn't reach it.
2. **Heartbeats re-armed idle forever.** Parsers classify heartbeats as `"model"`, but `KillCoordinator.touch()` ignored the class, so any liveness stream kept the 5-min idle timer from firing indefinitely.
3. **PhantomChat dropped stopped turns.** Telegram persisted an interrupted pair inline; PhantomChat never did.

## Fix

- `KillCoordinator.abandon()` (cause `abandoned`): `runHarnessProcess` calls it in `finally` when stdout did not reach EOF and no tripwire already killed the group. Normal completion is untouched.
- **Thinking budget:** `"model"` activity extends the idle deadline only up to `lastProductiveAt + thinkingTimeoutMs`, never shortens it; productive output or a finished tool resets it. Default **600s**, `harness_thinking_timeout_s` / `PHANTOMBOT_HARNESS_THINKING_TIMEOUT_MS`, plumbed next to `toolTimeoutMs`.
- `channels/core/interrupted.ts` `persistInterruptedTurn()`: shared by Telegram + PhantomChat; keeps any streamed text ahead of `[interrupted before reply]`, skips `/reset` and empty messages, stamps provenance.
- README + AGENTS invariant 59.

## Tests

- `tests/harness-failover-kill.test.ts`: real subprocess through `runWithFallback` (the call site); the primary emits `server_error` then touches a side-effect file after 1s, and the test asserts the file never appears. Plus thinking-budget coordinator + engine tests.
- `tests/channels-core-interrupted.test.ts`, PhantomChat `/stop keeps the stopped user message`, updated Telegram expectations, config tests.
- **Mutation-proven:** removing `abandon()` fails the failover test; disabling the thinking branch fails 3 budget tests; skipping persistence fails the PhantomChat test.
- Full suite: 4878 pass; the 2 `tui-bars` failures also fail on `main` locally (terminal width), not related.

## Checklist
- [x] README.md updated
- [x] AGENTS.md updated (invariant 59)


src/channels/phantomchat/server.ts
```diff
@@ -982,6 +983,7 @@ export async function runPhantomchatServer(
         idleTimeoutMs: input.config.harnessIdleTimeoutMs,
         hardTimeoutMs: input.config.harnessHardTimeoutMs,
         toolTimeoutMs: input.config.harnessToolTimeoutMs,
+        thinkingTimeoutMs: input.config.harnessThinkingTimeoutMs,
```
kaieriksen, line 989: This propagation is missing from other config-backed runtime paths: PhantomChat reactions later in this file pass idle/tool/startup but not `thinkingTimeoutMs`, so a configured non-default budget silently becomes the runner’s 600s default. Scheduled task turns (`src/cli/tick.ts`) and proactive PhantomChat greeting generation (`src/cli/run.ts` → `greet.ts`) have the same omission. Please plumb `harnessThinkingTimeoutMs` through every config-backed request builder and add a propagation regression; otherwise the documented setting depends on which entry point started the turn.

src/lib/harnessRunner.ts
```diff
@@ -183,9 +213,22 @@ export function createKillCoordinator(
...
+      if (activity === "model" && opts.thinkingTimeoutMs !== undefined) {
+        // Heartbeats prove the process is alive, not that it is getting
+        // anywhere. Extend the idle deadline only within the thinking budget,
+        // and never pull an already-armed deadline earlier.
+        const now = Date.now();
+        const target = Math.min(
+          now + opts.idleTimeoutMs,
+          lastProductiveAt + opts.thinkingTimeoutMs,
+        );
+        if (target > idleDeadline) armIdleAt(target);
```
kaieriksen, line 243: This does not enforce the documented ‘at most `thinkingTimeoutMs` after the last productive output’ contract when `thinkingTimeoutMs < idleTimeoutMs`: the initial/productive idle deadline is already later, and this branch deliberately refuses to shorten it. The added test codifies 500ms survival for a 100ms thinking budget, while README/config say the kill fires at the budget. A user setting 60s on the default 300s idle therefore still gets 300s. Please either make the deadline honor the cap (including after spawn/productive output), or constrain/rename/document the setting as `max(idleTimeoutMs, thinkingTimeoutMs)` so the public contract matches runtime behavior.

kaieriksen (changes requested):
The abandoned-process kill and interrupted-turn persistence are sound, and the focused suite/typecheck pass. Two blockers remain in the thinking-timeout contract: the configured value is not propagated through every runtime path, and values below the idle timeout do not enforce the documented cap.

src/channels/phantomchat/server.ts
andrewagrahamhodges: Fixed in fd63139. `harnessThinkingTimeoutMs` now reaches PhantomChat reaction turns, scheduled task wakes (`tick.ts`) and the greeting (`run.ts` → `resolvePersonaGreeting`, new `thinkingTimeoutMs` input). Regressions: `tests/harness-thinking-propagation.test.ts` scans `src/` and fails on any builder that forwards `harnessIdleTimeoutMs` without the thinking budget, plus per-path asserts on `lastRequest.thinkingTimeoutMs` in cli-tick, phantomchat-greet and the reaction tier. Each plumb was mutation-checked (removing it fails). The threat judge is exempt: tool-less single-shot classifier, runner default applies.

src/lib/harnessRunner.ts
andrewagrahamhodges: Took the second option in fd63139: the public contract is now `max(harness_idle_timeout_s, harness_thinking_timeout_s)` after the last productive output, i.e. the idle window is a floor. Honouring a sub-idle cap would make heartbeat-emitting harnesses die SOONER than silent ones, which inverts the point of the setting. README, AGENTS invariant 59, `Config`, `HarnessRequest` and `KillCoordinatorOpts` docs all state it, and the 100ms/500ms test comment now names it as the documented floor rather than an accident.

src/lib/harnessRunner.ts
```diff
@@ -776,6 +864,7 @@ export async function* runHarnessProcess(
     }
   } finally {
     clearTick();
+    if (!streamFinished && !terminalError) killer.abandon();
```
kaieriksen: [P1] Wait for the abandoned process group to exit before failover

`abandon()` only starts `killProcessGroup()` fire-and-forget, and `dispose()` does not await that promise. This generator therefore returns immediately and `runWithFallback` starts the next harness while the abandoned group may remain alive for the five-second SIGTERM grace. I reproduced this with a real primary script that runs `trap '' TERM; echo server_error; sleep 0.5; touch marker`: the fallback completed, then the abandoned primary created the marker. The existing regression misses this because its shell exits on SIGTERM. Please make the abandon path await group termination/escalation before the stream closes (and pin it with a SIGTERM-ignoring regression), otherwise this still permits the parallel side effects the PR is intended to prevent.

kaieriksen (changes requested):
The thinking-budget propagation fixes both prior threads, and the focused suites/typecheck pass. One failover-safety blocker remains: abandonment starts the group kill but does not wait for it, so a SIGTERM-resistant primary can still run tools while the fallback starts.

src/lib/harnessRunner.ts
andrewagrahamhodges: Fixed in the latest push. `abandon()` now SIGKILLs the group with no grace and returns a promise that resolves only after the process exits; the runner `finally` awaits it, so `runWithFallback` cannot start the next harness while the abandoned group lives. Added your repro as a regression (`trap '' TERM; echo server_error; sleep 0.5; touch marker`): the fallback asserts the primary has already exited when it starts, failover stays under 2s, and the marker never appears. Mutation-checked both halves: un-awaiting fails the test, and restoring the 5s grace fails it too. One gotcha worth knowing: a SIGKILLed child reports `exitCode` null in Bun, so the test probes `proc.exited`.

andrewagrahamhodges:
Added to this PR (Andrew asked): stopped turns now persist in the TUI and the ACP editor bridge through the same `channels/core/interrupted.ts` helper, and the TUI follows the other channels' type-to-interrupt design (a prompt sent mid-turn aborts with reason `interrupt` and is sent once the old turn unwinds; `^c` and ACP `session/cancel` use `stop`). Tests: `tests/tui-chat-interrupt.test.tsx`, extended ACP cancel test, updated busy-chunk test in `tui-chat-activity`. All mutation-proven. AGENTS invariant 59 + README updated.

kaieriksen (approved):
Re-reviewed exact head 744cbc8. The abandonment path now waits for a no-grace group kill before fallback, including the SIGTERM-resistant repro; TUI and ACP stopped-turn persistence/order handling are sound. Focused runtime suites pass (305 tests), TypeScript is clean, all CI is green, and all review threads are resolved.

PR #573, opened 2026-09-17
tests: isolate HOME so install.sh tests stop writing the real ~/.bashrc

andrewagrahamhodges:
## Problem
`tests/install-sh.test.ts` spawned `install.sh` with the developer's real `HOME`. install.sh appends `# added by phantombot installer` + a `/tmp/pb-install-*/bin` PATH line to `$HOME/.bashrc` on every run, so every local `bun test` grew the real `.bashrc` (48 blocks on 2026-09-11, 12 more by today), and leftover fake binaries on PATH could shadow the real `phantombot`.

## Fix
- `tests/testEnvIsolation.ts` (preload): `HOME`/`ZDOTDIR` always point at a throwaway dir for the whole suite. Real value kept as `PHANTOMBOT_TEST_REAL_HOME`.
- `runInstaller` passes explicit per-run `HOME`/`ZDOTDIR`/`SHELL` too (defence in depth).
- New tests: rc write lands in the sandboxed home exactly once over two runs; suite HOME is under the isolation root.
- AGENTS invariant 60.

## Verification
- Real `~/.bashrc` checksum unchanged after the new test; running main's test file changed it (bug reproduced).
- Mutation: removing the preload HOME swap fails the guard test.
- Full suite: 4888 pass. Failures are the pre-existing TUI bar tests and timing-flaky `tui-app-first-run` (also fails on main locally).


kaieriksen (approved):
Verified at 1a40a6c. The preload and per-run overrides isolate HOME/ZDOTDIR correctly; the focused isolation and installer suites pass, TypeScript is clean, all CI checks are green, and a real ~/.bashrc checksum stayed unchanged across the installer suite.

PR #574, opened 2026-09-17
native: serve photon's WASM from the binary so the embedded pi can see images

robertclawson:
## What happened (2026-09-17)

Native-harness phantoms could not see images. Omar (kw-omarchy, v1.1.377) received and saved every screenshot, but the model got `[Image omitted: could not be resized below the inline image size limit.]` instead, even for a 6 KB PNG, on both `glm-5.3-flash` and `kimi-k3` (OpenRouter lists both as image-capable). Claude-harness phantoms were unaffected.

## Root cause

Pi resizes every inbound image with Photon. `@silvia-odwyer/photon-node` loads its wasm with `readFileSync(__dirname + "/photon_rs_bg.wasm")`, and in a `bun build --compile` binary `__dirname` is the **build machine's** node_modules path (`/home/runner/work/...` for a release). Pi's own wrapper (`utils/photon.js`) then falls back to the wasm beside the executable and in the cwd, files pi's release ships next to its binary and ours does not. When every lookup misses, `loadPhoton()` swallows the error and `processImage` emits the omission placeholder whatever the image size.

Field proof on Omar: the same `__pi` call omitted the image; run from a directory holding the wasm (cwd fallback), GLM described the screenshot correctly.

## Fix

- `src/lib/embeddedPhoton.ts`: the wasm is a `with { type: "file" }` import, so bun embeds it in the binary. `installEmbeddedPhotonWasm()` redirects every `fs.readFileSync` of `photon_rs_bg.wasm` to that embedded copy. It wins even when a file exists on disk, so the photon build is always the one this binary was compiled and tested with. Pi's wrapper binds whatever `readFileSync` is current, so it chains through.
- `runEmbeddedPi` installs it only when compiled; from source photon resolves its real path.
- `src/wasmFileImport.d.ts` types the file import. AGENTS invariant 56 + layout updated.

No doctor, config or install changes, and no files written to disk.

## Tests

`tests/lib-embeddedPhoton.test.ts`:
- **End-to-end on a real compiled binary:** compiles from a hardlinked staged copy of the checkout, **deletes the staged copy** (so the baked path is gone, exactly like a user's host), then runs `__pi --mode json -p hi @t.png` and reads pi's user message. It asserts an image block and no omission. The process is killed at the user `message_end`, so no network or key is needed.
- Unit: the redirect lets photon load from a photon-node copy with no wasm; path matching (POSIX, Windows, file URL, near-miss names).

**Mutation-proven:**
- removing the hook from `runEmbeddedPi` fails the e2e test
- passing the read through unchanged fails the e2e and redirect tests
- a matcher that never matches fails 3 tests

**Manual, arm64, build tree hidden:**
- a 16px PNG is omitted before the fix and attached after
- a 27 MB 3000×3000 PNG is omitted before, and after is resized to a 2000×2000 JPEG with pi's dimension note
- `bun-windows-x64` cross-compile succeeds

Full suite: 4895 pass, 8 skip, 0 fail. `tsc --noEmit` clean.

## Rollout note

Hosts were hotfixed by hand today (wasm copied beside the binary on Omar and the rig). After this ships, that stray file is ignored because the embedded copy always wins, so it's harmless and can be removed.


src/lib/embeddedPi.ts
```diff
@@ -237,6 +237,12 @@ export async function runEmbeddedPi(args: string[]): Promise<void> {
   // Pi reads process.argv in places besides the args it is handed; make it
   // look exactly like `pi <args>`.
   process.argv.splice(2, process.argv.length - 2, ...args);
+  // Image support: photon's WASM path is baked to the build machine, so serve
+  // it from the binary before pi can try to load it (see embeddedPhoton.ts).
+  if (isCompiledBinary()) {
+    const { installEmbeddedPhotonWasm } = await import("./embeddedPhoton.ts");
+    installEmbeddedPhotonWasm();
```
kaieriksen, line 244: This is the right initialization boundary: the redirect is installed before either Pi module is imported, so Pi’s own temporary Photon wrapper binds through the embedded-WASM reader.

tests/lib-embeddedPhoton.test.ts
```diff
@@ -0,0 +1,187 @@
...
+          if (r.status !== 0) cpSync(join(REPO, entry), join(stage, entry), { recursive: true });
+        }
+        const build = spawnSync(
+          process.execPath,
+          ["build", "--compile", "./src/index.ts", "--outfile", bin],
+          { cwd: stage, encoding: "utf8" },
+        );
+        expect(build.status, build.stderr).toBe(0);
+        // The host has no build tree: the path baked into the binary is gone.
+        rmSync(stage, { recursive: true, force: true });
```
kaieriksen, line 176: Strong regression boundary: deleting the staged build tree before invoking the compiled binary proves the test is exercising the deployed-host failure, not succeeding via the checkout’s WASM.

kaieriksen (approved):
Approved at the exact head. I independently verified the fs patch ordering against Pi 0.85.1, ran the compiled-host regression after its build tree was deleted, and ran the embedded Pi suite plus TypeScript. The implementation fixes the fleet failure without adding a disk asset or altering source-mode behavior.

tests/lib-embeddedPhoton.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+        const line = await userMessageFor([bin, "__pi"], work, home);
+        expect(line).not.toContain("Image omitted");
```
kaieriksen, line 179: Validated this boundary locally against the submitted head: the staged build tree is removed before execution, and the compiled `__pi` path still emits an image block. I also separately exercised a 3000×3000 input and got the expected resize note rather than the omission placeholder.

kaieriksen (approved):
Approved at e517f643. The embedded asset hook matches Photon/Pi load order, the compiled-without-build-tree regression passes locally (5/5), TypeScript is clean, a separate 3000×3000 compiled-binary probe resized successfully, and both release-target x64-baseline and Windows binaries compile. All CI checks are green.

Issue #571, opened 2026-09-16
harnesses/pi: child dies on Node's default heap ceiling (exit 134) — turn silently undelivered, no fallback, stderr discarded


salvaalba-dev:
**Severity:** High on small hosts — silent data loss (the turn and the user's prompt are dropped), but no crash of phantombot itself and no auth/data-integrity impact. Searched existing issues for `heap`, `OOM`, `harness`, `undelivered`: no duplicate.

## Summary

On a 3.3 GB Ubuntu host, the `pi` harness child aborts with **exit 134 (SIGABRT)** when its V8 heap reaches the ceiling Node derives from total RAM. Measured on this host: **1700.5 MB**. Because the harness chain has no fallback, the turn is dropped (`turn undelivered`), the operator only sees a generic 🚨 alert, and **nothing retains the child's stderr** — so the failure is not diagnosable from phantombot's own artifacts.

The host was *not* out of memory at the time (~1.6–2.3 GB `available`, swap ~70 % used): the child died against its own default heap ceiling, not the machine's.

## Environment

- phantombot **1.1.354** (Bun binary, `systemctl --user` unit, `Restart=on-failure`)
- harness chain: `pi` only
- harness: `@earendil-works/pi-coding-agent` **0.82.1** (Node CLI, `#!/usr/bin/env node`, resolved through the service `PATH`)
- runtime: **Node v22.22.3**
- host: Ubuntu, 4 cores, **3.3 GB RAM**, ~3.7 GB swap
- channels in use: Telegram (direct) and phantomchat

## Observed behaviour

Five aborts in one afternoon, same conversation, same class of request (regenerate a report from an HTML template):

| # | turn start (UTC) | abort (UTC) | duration | exit |
|---|---|---|---|---|
| 1 | 16:20 | 16:21 | ~80 s | 134 |
| 2 | 16:55 | 16:56 | ~80 s | 134 |
| 3 | 16:58 | 16:59 | ~80 s | 134 |
| 4 | 17:34:58 | 17:38:31 | 213 s | 134 |
| 5 | 18:05:17 | 18:06:37 | 80 s | 134 |

Each abort produced **7 notification rows within the same minute**. Alert text (tail only):

```
🚨 pi harness error 🖥 <host> · no usable fallback configured [pi], turn undelivered · 7: 0x148eaf8  [pi] | 8: 0x1cf7681  [pi]
```

[...]

PR #576, opened 2026-09-17
tui: seed the first turn from argv (--prompt/--persona), run it in the launch cwd

### LINKED ISSUE #575: Add a --prompt seed flag to the chat TUI (trusted launch path for OS desktop integration)
## Problem

Omarchy's desktop integration (omacom/omarchy PR #12251, "Add Phantombot as a default agent option") needs to hand a prompt to Phantombot the way it hands one to every other agent: as argv into the interactive TUI the user just launched, e.g. Pi's case is literally `pi "$prompt"`.

The chat TUI has no seed flag, so the Omarchy branch currently works around it:

```
phantombot ask --history --conversation "cli:tui:<persona>" "$prompt"
phantombot   # then open the TUI on top
```

That detour runs the seed through `phantombot ask`, which is the **judge-screened, untrusted entry point** (TUI turns are trusted precisely because a human is at the TTY; `ask` exists to accept remote turns from Telegram/PhantomChat). Consequences:

- OS-initiated handoffs (crash-handler diagnosis, `omarchy agent prompt "..."`) can be scored risky and **held** by the threat judge — the desktop integration silently degrades to a bare TUI.
- Even when passed, the seed turn runs **headless and invisible**, unlike every other agent whose seeded work happens visibly in the terminal the user is watching.

## Proposed fix

Add a seed flag to the chat TUI itself:

```
phantombot --prompt "Review this project"
```

Semantics:

- Launches the interactive chat TUI with `--prompt`'s value entered as the first turn, exactly as if the user had typed it — i.e. the turn is **trusted** (human at the TTY launched the process; the prompt travels as argv, never through a network channel).
- Targets the TUI's own conversation (`cli:tui:<persona>`), persona resolved via the standard default-persona chain (env → state → config).
- Fresh install (no persona configured anywhere): open the first-run/onboarding TUI gracefully — queue the prompt for after onboarding, or drop it cleanly. It must not die with "persona not found" before the TUI opens.

## Why this shape

It matches what every other Omarchy default-agent does (`claude -- --prompt`, `codex --approve-for-me -- "$prompt"`, `pi "$prompt"`): trust comes from the interactive process the human launched, not from a remote channel. With this flag, Omarchy's phantombot case becomes one line (`phantombot --prompt "$prompt"`), the launcher helper and its headless `ask` dance disappear, and crash-handler handoffs run visibly in front of the user.

Happy to test the PR on real Arch + Omarchy (kw-omarchy VM) — the Omarchy branch will be updated to call this flag as soon as it ships.

robertclawson:
Closes #575. Unblocks the phantombot half of omacom/omarchy#12251.

## What it does

```
phantombot --prompt "Review this project"     # opens the TUI, sends it as turn 1
phantombot --persona kai --prompt "status?"   # …on a named phantom
```

The prompt is entered and submitted for the user, so the seeded turn runs **visibly, in the terminal they just launched**, and is trusted for the same reason a typed turn is. That is the whole security premise, so the gate enforces it and nothing weaker.

**Refused with exit 2** (stderr points at `phantombot ask`, which is the judge-screened headless path):

| invocation | why |
|---|---|
| no TTY on stdin or stdout (`… \| cat`, cron, CI, a launcher with no terminal) | nobody is watching |
| `--no-tui --prompt …` | headless seeding is not a thing; use `ask` |
| `--prompt` / `--persona` with a missing or empty value, or given twice | a launcher building argv badly must fail loudly, not run a different prompt |
| `--persona` that is not a valid persona-directory name | the value picks a directory and the vault that is decrypted |
| `--persona` naming a phantom that does not exist | a bad argument, not "nothing is set up" |

Refusal happens in `src/index.ts` **before** the credential bootstrap, so an unattended caller migrates no vault and provisions no persona on its way to being turned away. It never falls back to `ask` — a silent reroute would convert a refused unattended prompt into an accepted one.

Rulings applied from the issue thread: auto-submit; no persona → drop the prompt and let the TUI open the wizard; no-TTY is the piped/scripted case; no headless seeding; default persona with `--persona` to override.

**Other semantics**
- **Not ready yet:** wizard or Configure → the prompt is dropped and the opening screen says so (`seedForOpening`). Holding it across onboarding would fire a turn minutes after the user stopped expecting it.
- **Exactly once, to the right phantom:** the prompt lives in a ref on `App`; the chat screen reports consumption on mount, so `^s`/`esc` never replays it, and it is only offered to the persona the launch named.
- **Slash commands:** a seeded `/status` dispatches as a command, like a typed one.
- **Vault:** `--persona` also selects whose vault the entrypoint decrypts, otherwise the chat opens with another persona's env.

## Kai's cwd note (folded in, per "one PR per feature")

Interactive launches now run turns in the **launch cwd** (`lib/launchCwd.ts`), which is what Omarchy's `~/Work` redirect needs; `$HOME` remains the fallback when that directory is gone (`process.cwd()` throws) or is not listable. Scope is the TUI and its `--no-tui` twin only — **`ask` and every channel keep `$HOME`**, per the maintainer ruling on the issue: `ask` is the unattended entry point and is working as intended. `chatSession` takes `workingDir` as an option, so any caller that does not pass one is unchanged.

## Tests

28 new, all on the real surfaces:

- `tests/lib-tui-gate.test.ts` — the parse/refuse matrix, persona selection precedence.
- `tests/tui-launch-prompt.test.tsx` — the real `App` + real `ChatScreen`: the prompt is sent as an actual turn and appears in the transcript, once, only to the named persona; a slash seed runs as a command; the drop notice renders.
- `tests/lib-launch-cwd.test.ts` — cwd fallbacks; `openChat` hands the dir to the harness (and still uses `$HOME` when unset); a **real subprocess** proves a headless `--prompt` exits 2 and writes nothing to a fresh `$HOME`.

15 mutations proven (each one breaks a distinct guarantee and fails a test): TTY gate, `--no-tui` combination, empty value, repeated flag, persona-name validation, vault persona, unknown persona, refusal ordering vs the bootstrap, seed-on-setup-screen, seed-to-wrong-persona, launch cwd not passed, seed consumption (both halves), slash dispatch, session cwd.

Full suite green: **4923 pass, 8 skip, 0 fail**, typecheck clean. Also verified on the real entrypoint under a pty: `--persona nope` prints the argument error and exits 2.

## Not covered

No end-to-end run of a seeded turn against a real brain on real Arch/Omarchy — the issue offers the kw-omarchy VM for that, and it is worth doing before the Omarchy PR leaves draft.


src/lib/tuiGate.ts
```diff
@@ -66,7 +88,86 @@ export function bareInvocationMode(
...
+      noTui = true;
+      continue;
+    }
+    const eq = arg.indexOf("=");
+    const name = eq === -1 ? arg : arg.slice(0, eq);
+    if (name !== PROMPT_FLAG && name !== PERSONA_FLAG) return null;
+    let value: string | undefined;
+    if (eq !== -1) value = arg.slice(eq + 1);
+    else {
+      value = args[i + 1];
```
lenaparkhodges, line 147: This violates the contract stated above: "a launcher that builds argv badly must fail loudly, not run a different prompt than the one it thinks it sent."

A bare `--flag value` consumes the next argv element unconditionally, so a following flag becomes the value:

- `phantombot --prompt --no-tui hi` → prompt is the literal text `--no-tui`, `noTui` stays `false`, and the TUI opens and seeds a turn with the text `--no-tui` — the no-TUI refusal never fires.
- `phantombot --prompt --persona kai` → prompt `--persona`, then `kai` is a stray arg → `null` → plain usage instead of a refusal message.

Cheap fix that keeps the fail-loudly promise: reject a detached value that starts with `--` (or at least `PROMPT_FLAG`/`PERSONA_FLAG`/`NO_TUI_FLAG`), e.g. after `value = args[i + 1]`:

```ts
if (value !== undefined && value.startsWith("--"))
  return { error: `${name} needs a value; '${value}' looks like a flag.` };
```

`--flag=value` form is unaffected. Happy to approve as soon as this lands — it's the only blocker.

src/tui/index.tsx
```diff
@@ -170,8 +178,18 @@ export async function startTui(): Promise<number> {
   // of why the log pane opened empty.
   const restoreLogs = setLogSink((line) => logBuffer.push(line));
   const host = await hostSnapshot();
-  const opening = await resolveOpeningScreen();
+  // An unknown `--persona` is refused before the screen is taken over, like
+  // any other bad argument — not turned into a wizard for a phantom the user
+  // did not ask to create.
+  const unknownPersona = unknownLaunchPersona(launch, host.personas);
```
lenaparkhodges: Non-blocking ordering note: for an unknown-but-validly-named `--persona`, the credential bootstrap in `src/index.ts` has already run with `activePersona` pointing at a nonexistent persona dir before `unknownLaunchPersona` fires here (the AGENTS.md note's "refused BEFORE the credential bootstrap" holds for parse errors, not this path). It evidently no-ops safely — you verified `--persona nope` under a pty exits 2 cleanly — but either hoisting this check into `index.ts` ahead of the bootstrap, or one sentence noting the bootstrap tolerates a missing persona dir, would make the ordering claim exact.

lenaparkhodges (changes requested):
Excellent PR. The trust model is right and consistently enforced: TTY-on-both-streams as the gate, refusal before the bootstrap, no silent reroute to `ask`, seed exactly once to exactly the named phantom, and the cwd scope ruling (interactive only, `ask`/channels stay on `$HOME`) matches the issue ruling. Tests on the real surfaces — real `App`/`ChatScreen`, real subprocess for the headless refusal — are the strongest part.

One blocking nit on the gate's own contract, one non-blocking ordering note. Neither affects the security premise.

src/lib/tuiGate.ts
robertclawson: Agreed, and it was the real hole: `--prompt --no-tui hi` seeded the literal text `--no-tui` with `noTui` still false, so the headless refusal never fired. Fixed in c2cc6a7 — a detached value starting with `--` is now an error (`<flag> needs a value; '--x' looks like a flag.`). `--flag=value` is untouched, so a prompt may still start with dashes when it is written unambiguously. Tests cover both your cases plus `--persona --prompt hi`, and a positive test that `--prompt=--no-tui is the flag` still parses; deleting the guard fails them, as does narrowing it to the exact flag names.

src/tui/index.tsx
robertclawson: Good catch on the ordering wording. I checked rather than assumed: the bootstrap is safe rather than lucky here — the vault read is open-existing-only (`readAllVaultValues` returns null when the persona dir does not exist, precisely so the bootstrap never conjures a stray persona dir/identity/vault) and `cleanupPersonaTmpDir` returns on an unreadable dir. Documented that in c2cc6a7 next to the check, with the distinction spelled out: a malformed flag is refused before the bootstrap, a well-formed-but-unknown name is only knowable as unknown after it.

src/tui/index.tsx
```diff
@@ -170,8 +178,26 @@ export async function startTui(): Promise<number> {
...
+  // returns "no vault" for a missing persona dir (src/lib/vault.ts), and the
+  // tmp sweep returns on an unreadable dir, so neither provisions anything for
+  // a name we are about to reject.
+  const unknownPersona = unknownLaunchPersona(launch, host.personas);
+  if (unknownPersona !== undefined) {
+    restoreLogs();
+    process.stderr.write(`phantombot: ${unknownPersona}\n`);
+    return 2;
+  }
+  const opening = await resolveOpeningScreen(launch.persona);
```
kaieriksen: **[P1] Resolve the env-selected persona for the opening screen too.** `src/index.ts` calls `launchVaultPersona`, so with `PHANTOMBOT_PERSONA=lena` and default `robbie`, this process decrypts Lena’s vault; this call then passes only `launch.persona`, so `phantombot --prompt "…"` opens Robbie and sends the trusted seed there. That violates issue #575’s required env → state → config chain and the PR’s “right phantom” guarantee, while also pairing one persona’s bootstrap env with another persona’s chat. Resolve the effective launch persona once (explicit flag → env → configured default) and use that same name for vault bootstrap, unknown-persona validation, and `resolveOpeningScreen`; add a regression test where env and default differ.

kaieriksen (changes requested):
Requesting changes for one persona-routing blocker. Focused tests, TypeScript, the full 4,925-test Linux suite, the compiled x64-baseline build, and compiled headless/PTY gate probes pass. The Windows CI failure is an unrelated existing vault-migration test exceeding its 5-second timeout.

lenaparkhodges (commented):
Re-reviewed at c2cc6a7 (isolated checkout): the argv blocker from my earlier review is fixed — a detached value starting with `--` is now a loud error, `--prompt --no-tui hi` refuses instead of seeding literal flag text, and `--prompt=--no-tui …` still parses as an attached-value prompt. Tests cover all three, and deleting the guard fails them. Full suite 4,933 pass / 0 fail, typecheck clean. My blocker is withdrawn.

One blocker remains before this can approve: the P1 in the existing review thread — `launchVaultPersona` (flag → env → default) resolves the persona for the vault bootstrap in index.ts, but `startTui` only forwards `launch.persona` to `resolveOpeningScreen`. With `PHANTOMBOT_PERSONA=lena` set and default `robbie`, `phantombot --prompt "…"` still decrypts Lena's vault and seeds Robbie's chat — exactly the mismatch the P1 describes. Same applies to `unknownLaunchPersona`, which validates only the flag. Needs the env-resolved name threaded through all three call sites plus the env≠default regression test, per the thread's spec.

src/tui/index.tsx
robertclawson: Fixed in 94a0e28 — and you were right that it was a real routing bug, not just an inconsistency: with `PHANTOMBOT_PERSONA=lena` and default `robbie`, `phantombot --prompt "…"` decrypted Lena's vault and sent the trusted seed to Robbie's chat.

There is now ONE resolver. `resolveLaunchPersona(launch, env, defaultPersona)` returns `{ name, source }` — flag → env → configured default, the same chain as `resolvePersona` in config.ts with the flag added in front — and it carries the rung the name came from. `launchOpeningTarget` wraps it with the existence check, and both consumers read it: `src/index.ts` for the vault bootstrap, and the TUI for the opening screen. `launchVaultPersona` is gone rather than left beside it, so there is no second copy of the chain to drift.

The three rungs deliberately do NOT fail the same way:

- unknown `--persona` → refuse, exit 2, message unchanged;
- unknown `PHANTOMBOT_PERSONA` → refuse too, naming the env var: the entrypoint already resolved the vault from that name, so continuing would open some other phantom's chat with no secrets loaded at all, and every other persona-aware command likewise targets the env name rather than silently substituting the default;
- a configured DEFAULT naming a missing phantom → **not** refused. That is the broken-default case `resolveOpeningScreen` already owns (legacy adoption, heal-once, else wizard), so `launchOpeningTarget` returns `requested: undefined` for the default rung and that whole path is byte-for-byte what a bare launch does today.

On testing, the first cut of this fix passed while `startTui` still read `launch.persona` — my test had re-derived the chain instead of exercising it. So `startTui`'s pre-render half is extracted as `resolveLaunchOpening(launch, host)` (target → opening screen → seed decision), and `tests/tui-launch-persona.test.ts` drives THAT against real personas on disk with env ≠ default: the env phantom opens, the flag beats the env, a plain launch still requests nothing (so the heal path runs), an unknown env name refuses, and the real `startTui` returns 2 with one stderr line before the renderer. Reverting the wiring to `resolveOpeningScreen(launch.persona)` now fails.

8 mutations proven on this change (the original bug; `startTui` ignoring the refusal; dropping the env rung; refusing an unknown default; `requested` = name for the default rung; not refusing an unknown env name; dropping the env `trim()`; flag no longer beating env). Full suite 4,936 pass / 0 fail, typecheck clean. AGENTS.md invariant 61 and the README launch section now state the chain and the asymmetry.

kaieriksen: **[P1] The launch persona is still resolved twice across a mutable environment.** `src/index.ts` computes `activePersona`, then `loadVaultIntoEnv(activePersonaDir)` can inject `PHANTOMBOT_PERSONA` (the vault CLI accepts it and legacy migration imports it), and `startTui` later calls `resolveLaunchOpening`, which runs `resolveLaunchPersona` again against that mutated `process.env`. On exact head `94a0e28`, I created Robbie’s vault with `PHANTOMBOT_PERSONA=lena`: the bootstrap selected Robbie, vault loading injected Lena, and the TUI target resolved to Lena — still pairing one phantom’s secrets with another phantom’s chat. A vaulted `PHANTOMBOT_DEFAULT_PERSONA` can similarly change the later `hostSnapshot()` result. Resolve the `{ name, source }` value before vault loading and thread that exact value through bootstrap, validation, and opening-screen selection (or make routing env names categorically non-vaultable); add the real vault-load regression, because the current test never crosses the bootstrap mutation boundary.

kaieriksen (changes requested):
Re-requesting changes at 94a0e28 for one remaining persona-routing blocker: the resolver is shared as code but invoked twice, on opposite sides of vault injection, so vault-loaded routing variables can still make bootstrap and TUI select different phantoms. Focused 49 tests and TypeScript pass, and all GitHub checks are green; the inline reproduction shows the failing exact-head path.

src/tui/index.tsx
lenaparkhodges: Independently verified at exact head 94a0e28 (isolated checkout): resolveLaunchOpening → launchOpeningTarget re-runs resolveLaunchPersona against process.env (src/tui/index.tsx:213), which by then has been mutated by loadVaultIntoEnv in index.ts — and src/lib/vault.ts:545-557 confirms the legacy-migration path imports PHANTOMBOT_PERSONA from the vault, so a vaulted value can flip the TUI target while the bootstrap stays on the original resolution. Kai's repro path is sound and the focused tests (33 pass) never cross the bootstrap mutation boundary, so no regression catches this. Agree the fix is resolve-before-vault-load, threading the { name, source } value through bootstrap, validation and opening-screen selection — or making routing env names categorically non-vaultable — plus the real vault-load regression test. Agreeing with the re-requested changes; my earlier argv-blocker withdrawal stands.

robertclawson: Fixed in 3c493e7 — and I took both halves you named, because they cover different failure modes.

**Carried, not recomputed.** `src/index.ts` now snapshots the routing env *before* the credential bootstrap, resolves `{ name, source }` from that snapshot, and hands the value to `startTui` as a `LaunchContext`. `launchOpeningTarget` takes the resolved persona instead of resolving one, so there is no longer a second resolution to disagree with the first. The remaining fallback resolves the chain against the pre-bootstrap snapshot (never `process.env`) and is reachable only when the bootstrap threw *before* resolving — which is also before it could load any vault.

**Routing names are categorically non-vaultable.** `PHANTOMBOT_PERSONA` and `PHANTOMBOT_DEFAULT_PERSONA` are withheld at vault READ time, beside the retired config.toml mirrors, so the guard holds for every persona and every per-turn reload (including `reloadVaultForPersona`, which was the other way a vaulted value could re-aim routing). `vault set` / `env set` refuse them at the moment there is an operator to tell, and the plaintext migration drops them — that path fans one `~/.env` into EVERY persona's vault, so importing a routing name would give each phantom a row naming some other phantom. Withheld and warned about, never evicted.

Tests cross the boundary this time:
- `tests/vault.test.ts` — a real vault holding `PHANTOMBOT_PERSONA=lena` is loaded by the real `loadVaultIntoEnv`: the caller's routing survives untouched, nothing is tracked or injected, and the `GITHUB_TOKEN` beside it still loads (the guard is by name, not a refusal to read the vault). Second case: an *empty* routing slot, which is how it used to slip in.
- `tests/tui-launch-persona.test.ts` — resolve as the entrypoint does, then mutate `process.env` as a vault load would, then drive the real `resolveLaunchOpening`: the chat opens the phantom the entrypoint decrypted. Plus `startTui` itself driven with a carried persona the env contradicts.
- `tests/cli-env.test.ts`, `tests/vault-migrate.test.ts` — the write-side refusals.

6 mutations proven (vault guard, TUI re-resolving from `process.env`, `startTui` dropping the context, CLI refusal, migration drop, and shrinking the name list to just `PHANTOMBOT_PERSONA`). Suite 4943 pass / 0 fail, typecheck clean.

One thing worth stating plainly rather than leaving you to find it: `src/index.ts`'s own call site is not independently observable in tests. With the vault no longer carrying routing names, the carried path and the fallback agree on every reachable input, so deleting the argument there does not fail anything — the layering is exactly why. What the tests pin is the contract at the `startTui`/`resolveLaunchOpening` boundary and the vault read itself.

lenaparkhodges (dismissed):
APPROVED at exact head 3c493e7e00c6586111b9400a765a84d125889ebc. Independent verification in an isolated checkout: full suite 4943 pass / 0 fail, typecheck clean, 6/6 CI checks green after a `runHeartbeatCli`-only Windows rerun (first run was a "fork failed" child-spawn flake outside this PR's surface). Kai's double-resolver P1 is structurally resolved — routing resolved once pre-bootstrap and carried as a `LaunchContext`, routing env names non-vaultable at read/write/migrate — and my round-1 argv blocker stays fixed. Verification detail is in the thread on my earlier review at src/tui/index.tsx. Note for merge order: omarchy PR #12251 rebases on this once it lands (its last remaining blocker).

src/tui/index.tsx
kaieriksen: Verified at exact head `3c493e7`. The original repro is closed on both boundaries: `loadVaultIntoEnv` with a real vault now withholds both routing names while loading the adjacent secret, and `resolveLaunchOpening`/`startTui` honor the carried pre-bootstrap `{ name, source }` even when `process.env` later contradicts it. I ran the 106 focused launch/vault/migration/CLI tests, TypeScript, the compiled x64-baseline build, a real-PTY unknown-persona probe (exit 2), and the compiled headless seed refusal; all pass. This P1 is resolved.

kaieriksen (dismissed):
APPROVED at exact head 3c493e7. The remaining persona-routing blocker is structurally fixed: routing is resolved once before vault injection and carried into the TUI, while PHANTOMBOT_PERSONA / PHANTOMBOT_DEFAULT_PERSONA are withheld at vault read, refused on write, and dropped during legacy migration. Independent verification: 106 focused tests pass, TypeScript is clean, the x64-baseline binary builds, compiled PTY/headless refusal probes pass, all GitHub checks are green, and all review threads are resolved.

robertclawson:
Rebuilt on current main after #577 merged. The only conflict was AGENTS.md — both PRs added invariant 61, so #577's (harness failure diagnostics) keeps 61 and this PR's seeded-prompt invariant becomes 62; no code conflicts. Typecheck clean and the full suite is green on the merge commit (4955 pass / 0 fail). Re-requesting both reviews at 4c56869.

kaieriksen (approved):
APPROVED at exact head 4c56869. The merge from current main preserves the four reviewed feature commits unchanged (range-diff exact); its only conflict was AGENTS.md, correctly retaining #577 as invariant 61 and renumbering the seeded-prompt invariant to 62. Independent re-verification: 122 focused launch/persona/vault/cwd tests pass, TypeScript is clean, all six GitHub checks are green, and all review threads remain resolved.

lenaparkhodges (approved):
Re-approved at merge head 4c56869 (merge of main/#577). Verified in an isolated checkout: full suite 4963 pass / 0 fail, typecheck clean, CI 6/6. The only manual merge artifact is the AGENTS.md conflict resolution — main's #571 lesson kept as item 61, the launch-prompt lesson renumbered to 62; correct, with two stray trailing blank lines (cosmetic, non-blocking). No conflict-resolution changes to any source file (empty combined diff outside AGENTS.md).

PR #577, opened 2026-09-17
fix: preserve failed harness turns and bound history

### LINKED ISSUE #571: harnesses/pi: child dies on Node's default heap ceiling (exit 134) — turn silently undelivered, no fallback, stderr discarded

kaieriksen:
Closes #571.

## Summary

- cap canonical conversation history at 128 KiB after the existing 30-row limit, keeping the newest whole messages and logging loaded/included bytes
- preserve terminally failed prompts with an atomic recovery marker while leaving intentional abort/reset persistence to channel adapters
- retain structured exit code, signal, safe error, and redacted stderr tail in the existing two-hour turn registry
- add an explicit `pi-host` V8 old-space setting plus a startup warning when the host Node default probes below 2048 MiB; never size it automatically from host RAM

## Safety

- configured heap caps apply only to `pi-host`; embedded/native Pi is unchanged
- existing `NODE_OPTIONS` remain intact and the explicit cap wins by being appended last
- fallback success remains a successful turn; only an exhausted terminal failure receives the recovery marker
- operator cancellation and host shutdown do not create failure-history pairs

## Verification

- `bun test`: 4,906 pass, 8 platform skips, 0 fail
- focused regression gate: 500 pass, 0 fail
- `bun run typecheck`
- compiled Linux x64 baseline binary: `--version` and `--help`


src/orchestrator/turn.ts
```diff
@@ -871,6 +948,38 @@ async function* runTurnBody(
     }
   }
 
+  if (
+    !succeeded &&
+    terminalError &&
+    !input.noHistory &&
+    !input.signal?.aborted &&
+    terminalError.killCause !== "aborted" &&
+    !terminalError.error.includes("host shutting down")
```
lenaparkhodges, line 957: Non-blocking: this couples the orchestrator to harnessRunner's exact error prose. The shutdown yield in runHarnessProcess now carries structured `exitCode`/`signalCode` but not a `killCause` — adding one there (e.g. `killCause: "shutdown"`) would let this be a structural check like the `aborted` case above, immune to future rewording. Fine to land as-is.

src/harnesses/pi.ts
```diff
@@ -587,6 +622,57 @@ export class PiHarness implements Harness {
   }
 }
 
+export async function probeNodeHeapLimitMb(): Promise<number | undefined> {
+  try {
+    const proc = Bun.spawn(
+      [
+        "node",
```
lenaparkhodges, line 629: Non-blocking: this probes whatever `node` resolves first on PATH, which may not be the runtime that actually executes the `pi` bin (nvm/version-manager setups). A mismatch makes the warning inaccurate. Since it's warning-only and the explicit `max_old_space_mb` path bypasses the probe entirely, acceptable — worth a comment noting the caveat.

src/orchestrator/turn.ts
```diff
@@ -68,6 +70,37 @@ import {
...
+  loadedBytes: number;
+  includedBytes: number;
+} {
+  const sizes = history.map((turn) => Buffer.byteLength(turn.text, "utf8"));
+  const loadedBytes = sizes.reduce((sum, size) => sum + size, 0);
+  let includedBytes = 0;
+  let first = history.length;
+  for (let i = history.length - 1; i >= 0; i--) {
+    const size = sizes[i]!;
+    if (includedBytes + size > maxBytes) break;
```
lenaparkhodges, line 98: Non-blocking: when a single stored message exceeds the budget, the loop breaks immediately and history comes back empty (nothing partial is ever included — right call for prompt integrity). Might be worth a one-line comment so a future reader doesn't read the empty-history result as a bug.

lenaparkhodges (dismissed):
Reviewed at exact head b258dd4 (isolated checkout). Solid, well-scoped fix for #571. Verified locally on Linux x64: full suite 4,906 pass / 0 fail / 8 skips, `bun run typecheck` clean, and the last-flag-wins `--max-old-space-size` NODE_OPTIONS behavior confirmed against a real Node runtime (1024 then 2560 → ~2752 MiB total limit, first flag ignored, flag accepted in NODE_OPTIONS). The failure-persistence path correctly excludes the threat-hold path (hold yields `done`, so `succeeded` is set), operator aborts (`signal.aborted`, `killCause === "aborted"`) and the shutdown string — no duplicate markers on the interrupted-turn path. Approving; three non-blocking notes below.

src/harnesses/pi.ts
kaieriksen: Agreed. This is intentionally a warning about the daemon’s PATH-resolved host Node, not an enforcement decision; the explicit cap bypasses the probe and remains authoritative. Keeping the approved head stable.

src/orchestrator/turn.ts
kaieriksen: Agreed; empty history is intentional when the newest single message exceeds the cap because splitting stored messages would corrupt conversation structure. The README and regression test now establish the whole-message contract, so I’m keeping the approved head stable.

src/orchestrator/turn.ts
kaieriksen: Agreed on the future-proofing seam. I’m keeping the approved head stable because host shutdown is already excluded and this does not affect #571 correctness; a structured shutdown cause belongs with a broader kill-cause contract cleanup.

robertclawson (changes requested):
Reviewed at b258dd4. Design is right and the code reads well: bounding at assembly time rather than inside each harness is the correct seam, the failed-turn marker keeps a crashed turn from vanishing, and the pi heap cap is a sensible guard. I built the branch, ran the full suite (4906 pass, matches your claim) and CI is green.

Two mutations survive, and both are headline behaviours of #571 — requesting changes on those. Three further notes are non-blocking.

### Blocking

**1. The 128 KB cap is never wired-tested.** In `src/orchestrator/turn.ts` I replaced `const history = bounded.history` with `const history = loadedHistory` — i.e. the cap computes, logs, and is then thrown away — and the entire suite still passes. Only `boundHistoryByBytes` itself is covered; nothing asserts the harness actually receives the bounded array. A test that runs a turn with an oversized history and asserts on what reaches the harness invocation would close this.

**2. Exit code / signal / stderr tail never reach a turn record in any test.** I nulled all three in `runTurn`'s outcome (`turn.ts` ~396-398) and the suite stayed green. Those are the first two acceptance criteria on #571 ("exit code recorded", "stderr tail retrievable"). `tests/lib-turnRegistry.test.ts` only feeds a hand-built record in, so it proves the registry stores what it is given, not that the turn path ever gives it anything. An end-to-end assertion through `fake-pi.sh` exiting non-zero would cover both.

### Non-blocking

**3. `DEFAULT_HISTORY_MAX_BYTES` is hardcoded and harness-agnostic.** #571 is a Node heap problem, but the cap applies to every harness including claude, where there is no such limit. A long conversation gets silently trimmed with only a log line to show for it. Suggest a config knob (per-harness, defaulting to the current value) so a user with a 200K-context model is not capped by a pi-specific constraint.

**4. Failed turns now compound history.** Persisting the user message plus failure marker means #571's own scenario — five aborts against the same large prompt — writes five copies of that prompt into history, consuming the budget this PR introduces. Worth considering whether a repeated failure against an identical user message should replace rather than append.

**5. `probeNodeHeapLimitMb` spawns `node` at startup with no timeout, before the run lock is taken.** A wedged or shimmed `node` on PATH blocks startup rather than degrading. A short timeout with `undefined` on expiry would keep the failure mode soft — this is the same class of unbounded wait we just fixed on the phantomops turn path.

Happy to pair on the two test gaps if useful.

kaieriksen:
Thanks @robertclawson — both blockers are fair and both are now closed on `aee1e27`, plus note 5 implemented. Mutation-verified, not just added:

**1. History cap now wired-tested.** `tests/orchestrator-turn.test.ts` runs a turn with a >128 KiB history through a scripted harness and asserts on what the harness invocation *receives* (newest complete turns only, oversized turn absent). Mutation check: replacing `bounded.history` with `loadedHistory` in `runTurn` turns exactly this test red, and nothing else.

**2. Exit code / signal / stderr now end-to-end through the real spawn path.** `fake-pi.sh` gains a `signal` mode (stderr write, then SIGKILL). Two new tests drive a real `PiHarness` subprocess through `runTurn`: one child exiting non-zero, one killed by signal, both asserting `exit_code` / `signal` / `stderr_tail` on the turn-registry record via `readRegistry`. Mutation check: nulling `exitCode`/`signalCode`/`stderrTail` in `runTurn`'s error outcome fails both tests.

**5. Probe timeout implemented.** `probeNodeHeapLimitMb` now spawns with a 2 s timeout; a wedged PATH shim degrades to no startup warning instead of holding daemon startup.

**3 (per-harness cap) and 4 (failed-turn compounding)** — agreed both are real design considerations, neither is a defect in the #571 failure scenario. Rather than widen this PR's scope post-approval, I'll file them as follow-up issues unless you'd rather see them folded in here — say the word.

Local verification at `aee1e27`: full suite 4,909 pass / 0 fail, `bun tsc --noEmit` clean. Full CI is running on the new head.

kaieriksen:
Filed the two follow-ups rather than widening this PR post-approval:
- #578 — per-harness history byte cap (note 3)
- #579 — failed-turn prompt compounding (note 4)

Say the word if you'd rather see either folded in here.

robertclawson (approved):
Approving at `aee1e27` — both blockers independently verified here, not taken on the summary.

**1. History cap is now wired-tested.** Mutating `runTurn` line 441 to `const history = loadedHistory` (bypassing `boundHistoryByBytes` while leaving the debug logging intact) turns exactly one test red — `passes byte-bounded history to the harness` — and nothing else. That is the right blast radius: the assertion is on what the harness *receives*, so it fails if the bound is ever disconnected from the invocation again.

**2. Exit diagnostics are now end-to-end.** Replacing `chunk.exitCode` / `chunk.signalCode` / `chunk.stderrTail` with `undefined` in `runTurn`'s error outcome fails both new tests (`a failing Pi subprocess records exit status and stderr`, `a signalled Pi subprocess records the signal`). Both drive a real `PiHarness` subprocess through the spawn path and assert on the registry record via `readRegistry`, so #571's headline criterion is covered by behaviour rather than by a unit of the helper.

**5. Probe timeout.** Confirmed Bun 1.4 honours `timeout` on `Bun.spawn`: a `sleep 10` under `timeout: 2_000` exits at 2.0s with code 143 / `SIGTERM`, so `probeNodeHeapLimitMb`'s non-zero-exit branch returns `undefined` and startup degrades to no warning as intended. Non-blocking: the timeout itself is uncovered — deleting `timeout: 2_000` survives the suite, since tests reach the probe through the injected `probe` seam. Given this is warning-only telemetry and a real test would cost ~2s of wall clock, I'd leave it; noting it so it's a known gap rather than an assumed one.

**3 and 4** — filing #578 and #579 rather than widening an approved PR is the right call, no need to fold either in here.

Local verification at `aee1e27`: `tests/orchestrator-turn.test.ts` 55 pass / 0 fail, both mutations above proven to kill the intended tests and restored afterwards. CI green on the head, 0 unresolved threads.

robertclawson:
Re-requested @lenaparkhodges — her approval was dismissed by the `aee1e27` push, and standing policy on this repo is two live approvals at the final head before merge. Nothing for her to re-litigate beyond the new tests.

kaieriksen:
@lenaparkhodges — your approval was dismissed by a push, re-requesting. The delta since `b258dd4` is one test-only commit (`aee1e27`) closing Robert's review: wire-tests for the 128 KiB cap and the Pi exit/signal/stderr registry path (both mutation-verified), plus a 2s timeout on the Node heap probe (his note 5). No production behavior change beyond the probe timeout. CI is green, Robert has approved this head.

lenaparkhodges (approved):
Approving at aee1e27 (re-review after the push). Re-verified here: full suite 4909 pass / 0 fail, typecheck clean, CI 6/6 at this head. The delta holds up — the probe timeout (2s, warning-only degrade) closes the uncovered-gap note from my earlier review, and both new diagnostics tests drive a real PiHarness subprocess through the spawn path and assert on the registry record, so #571's headline criterion is behaviour-covered. The byte-cap test asserts on the history the harness receives, which is exactly the right seam. Nothing further from me — ready to merge.

Issue #580, opened 2026-09-18
Reply-language rule is unenforced: narration lines still leak the wrong language


robertclawson:
## Problem

The reply-language rule (#548) is injected into every turn unconditionally as prose: *"Write your reply — including every pre-tool narration line — in the language of the USER'S LATEST MESSAGE."*

It is still violated in practice. Observed 2026-09-18: an English user message received English body text but **Spanish pre-tool narration lines** ("Buscando...", "Un momento...") across multiple consecutive turns. This has recurred often enough that the principal has repeatedly asked for it to be fixed, and has started hand-adding a duplicate rule to per-persona `IDENTITY.md` files to try to force compliance.

Two things are wrong:

1. **Prose instructions have no enforcement.** The rule is correct, deterministic in its *source* (user's latest message), and still gets ignored by the model — most often on narration, least often on the main body. Adding more copies of the rule does not help; it made things worse in #534/#535, where two prompt sections restating one rule differently caused the original bug. Per-persona `IDENTITY.md` copies should be treated as an anti-pattern for this reason.

2. **Pre-tool narration is the repeat offender.** Narration is generated in a different "mode" than the reply body — short, interstitial, often produced while the model is oriented toward the *content* it is about to read (a Spanish email, a Dutch invoice). That is exactly the failure the rule names as forbidden, and exactly where it leaks.

## Proposed fix

Deterministic gate, not more prompt text.

**A. Language check before delivery.**
Detect the language of the user's latest message and of the outbound reply (body + each narration line). On mismatch, do not deliver — regenerate once with an explicit corrective, and log the violation. Cheap language ID (e.g. a small classifier or n-gram detector) is sufficient; no LLM call needed for detection. Needs a confidence floor so short/ambiguous replies ("ok", a bare code block, a URL) never trigger a false regenerate.

**B. Narration specifically.**
Either route narration lines through the same gate, or suppress pre-tool narration entirely in chat channels. Option B alone would have prevented the 2026-09-18 incident and costs nothing functionally — narration exists only to fill streaming silence.

**C. Remove the duplicate.**
If per-persona `IDENTITY.md` language lines exist anywhere, drop them; the injected rule is the single canonical copy (one canonical place per rule).

## Acceptance criteria

- An English-prompted turn that reads Spanish-language tool output produces an English reply **and** English narration, verified by a test that drives the real delivery path (not a re-implementation of the check).
- A mismatching reply is caught and regenerated rather than delivered; the violation is visible in logs.
[...]

PR #581, opened 2026-09-19
prompt: stop narration restating the language rule; make it the last overlay (#580)

robertclawson:
## Why

The reply-language rule (#548) is injected on every turn and was still being violated — and the paragraph this PR deletes *was itself* the previous attempt to fix it.

I scored every stored user→assistant pair on two personas (~14,800 turns, four channels), comparing the language of the user's message against the reply body and against each pre-tool narration line. Full numbers in [#580](https://github.com/phantomyard/phantombot/issues/580#issuecomment-5740686185). Three findings:

1. **Every measured violation is a narration line, in Spanish, on an English prompt, with an English body.** Zero Dutch, zero French, zero body leaks — despite heavy Dutch source material in the same turns. The rule holds for the body; narration escapes it.
2. **The overlay is not what fails.** ACP receives no `REPLY_LANGUAGE_INSTRUCTION` at all (it lived in the Telegram and phantomchat suffixes only) and leaks at the same rate as the channels that do.
3. **The only place Spanish was ever named in the prompt** was `PRE_TOOL_NARRATION_INSTRUCTION`'s own language paragraph — *"if the user wrote to you in Spanish, narrate in Spanish"* — appended **LAST**, i.e. in the highest-recency slot.

A rule stated abstractly, then illustrated in exactly one named language, read last, is the best available explanation for violations that are 100% that language.

## What changed

- **`PRE_TOOL_NARRATION_INSTRUCTION` says nothing about language.** One rule, stated once.
- **`REPLY_LANGUAGE_INSTRUCTION` moves to the orchestrator overlay stack, as the LAST overlay** (`persona/builder.ts` → pushed in `orchestrator/turn.ts`), out of the two channel suffixes. Three wins: recency works *for* the rule; exactly one copy exists; ACP, the TUI and `ask` get a language rule for the first time.
- **Narration budget.** Narrate the first call of a turn, anything that changes state or reaches off the machine, anything slow — not a sentence in front of each of twenty greps. Fewer narration lines (less exposure), and narration concentrates on the calls a user would actually want to interrupt, which is what narration is *for*.
- AGENTS.md invariant 45 and the README's Reply Language / Reply Pacing sections updated.

## Tests

`bun test`: 4960 pass, 0 fail. `bun tsc --noEmit` clean.

New/changed assertions:
- the narration block contains no language text and names no concrete language;
- the reply-language block is present, **last**, and appears exactly once — with and without narration, and with a channel suffix present;
- the narration budget wording is present;
- the Telegram wiring test now checks the rule survives the suffix removal and still appears once.

Both new assertions were verified to **fail** against the pre-fix prompt: re-adding the language paragraph fails the two "says nothing about language" tests, and pushing the language overlay before narration fails the LAST-overlay test.

## What this does *not* do

It does not make compliance non-load-bearing — this is still prose. It removes the one trigger the data actually points at. The enforcement gate (detect each narration line's language in the chunk stream; **strip** on mismatch rather than regenerate, since narration is cosmetic and silence beats the wrong language) is specified in #580 as a follow-up, with design questions — detector choice, buffer placement, confidence floor — open for review there. Body text is deliberately out of scope for that gate: zero measured leaks, and buffering it would defeat streaming.

Deliberately **not** included: rolling the rule into per-persona `IDENTITY.md` files. A second restatement of this rule is what caused #534/#535; invariant 45 now says so explicitly.

Part of #580


src/orchestrator/turn.ts
```diff
@@ -701,6 +712,7 @@ async function* runTurnBody(
     });
   }
   if (input.toolNarration) overlays.push(PRE_TOOL_NARRATION_INSTRUCTION);
+  overlays.push(REPLY_LANGUAGE_INSTRUCTION);
```
kaieriksen, line 715: This is the load-bearing placement: keeping the canonical language overlay unconditional and appended after narration, digests, sibling notices, and caller suffixes makes the single-copy/last invariant true for chat, ask, TUI, ACP, tasks, and the cache-fallback rebuild.

kaieriksen (approved):
Full-diff review complete. The canonical rule is now unconditional, single-copy, and last across every runTurn entry point; channel wiring and cache fallback remain coherent. Verified locally: affected prompt/cache suites 320/320, typecheck clean; the full suite reached 4,951 passes with nine subprocess-lock failures caused solely by bun being absent from PATH, and all 99 affected tests passed when rerun with the repository runtime on PATH. GitHub Linux/Windows CI and CodeQL are green.

tests/orchestrator-turn.test.ts
```diff
@@ -363,14 +363,110 @@ describe("runTurn — successful path", () => {
...
+        ...baseInput(),
+        userMessage: "hi",
+        harnesses: [harness],
+        toolNarration: true,
+      }),
+    );
+
+    const prompt = captured?.systemPrompt ?? "";
+    const narration = prompt.slice(prompt.indexOf("# Narration before tool"));
+    const block = narration.slice(0, narration.indexOf("# Reply language"));
```
lenaparkhodges, line 395: Non-blocking nit: if `# Reply language` ever went missing, `indexOf` returns -1 and `slice(0, -1)` silently becomes 'everything but the last char' — `not.toMatch(/language/i)` would then fail for the right reason anyway (the narration block itself doesn't contain it, so it'd pass), making this assertion vacuous rather than failing loudly. The LAST-overlay test in this file does pin presence, so coverage is fine overall — just noting the two assertions here quietly degrade to a pass if the canonical copy disappears. Not worth blocking on.

lenaparkhodges (approved):
Reviewed the full diff. The data-to-change chain is the strongest part: #580's scoring shows violations are 100% Spanish narration lines on English prompts, and this PR removes exactly that trigger — the narration block's own language paragraph, in the last, highest-recency slot — instead of adding prose on top of prose. Single unconditional copy, pushed last by the orchestrator, also closes the real gap that ACP/TUI/ask never had the rule at all.

The narration budget change is a behavior change riding along, but it's coherent with the thesis (fewer narration lines = fewer violation surfaces, and the budget text is well calibrated — state changes and slow calls stay narrated).

Tests pin presence, position, and single-copy, and were verified failing against the pre-fix prompt. One non-blocking nit on test-robustness in an inline comment. CI green. Approve.

tests/orchestrator-turn.test.ts
robertclawson: Agreed on the analysis — both assertions degrade to a quiet pass rather than a loud failure if the canonical `# Reply language` heading disappears, and the LAST-overlay test in the same file is what actually pins presence, so the case is covered.

Leaving the diff as-is: this repo dismisses approvals on any push, so a test-only hardening commit would cost both approvals and a re-review round for an assertion that is already backstopped. Noted as a follow-up to fold into the enforcement-gate PR specced in #580, where these tests get touched anyway — the guard there will be an explicit index check that throws instead of slicing.

src/orchestrator/turn.ts
robertclawson: Confirmed — that placement is the whole point of the change: one unconditional copy, emitted last by the orchestrator, so recency works for the rule instead of against it, and ACP/TUI/ask get it for the first time. The position and single-copy tests pin it so a later overlay cannot quietly slide in after it.

PR #583, opened 2026-09-19
Gate pre-tool narration on the user's language (#580)

robertclawson:
Part of #580 — the enforcement half. #581 shipped the prompt-side half.

## Why code and not more prose

The reply-language rule has been stated in prose twice and violated both times. #548 introduced it; #581 deleted the duplicate copy of it (inside the narration block) that was making things worse. Across ~14,800 scored turns on two personas, **every observed violation was a pre-tool narration line in the wrong language while the reply body was correct.** So this PR takes model compliance out of the delivery path for the one thing that actually leaks, and leaves the body alone.

## What it does

**`src/lib/languageGate.ts`** — a new, deliberately narrow detector. It is not a general-purpose language identifier and the module doc says so:

- **Non-Latin scripts by codepoint.** Cyrillic, Han, Kana, Hangul, Arabic, Hebrew, Greek, Devanagari, Thai. No word list, no training data, correct for every language written in them. This is the branch #534's classifier got wrong by going silent on scripts it could not score.
- **Latin by exclusive function words.** Seven profiles (en/es/fr/de/nl/it/pt). Only words exclusive to a single profile are scored; shared words ("de", "in", "me", "la") are ignored rather than guessed at, which is what keeps precision up on the ~6-word strings narration consists of.
- **Precision-first.** Below three tokens, below two exclusive markers, or below a 75% winner share → `undefined`, which the caller reads as "leave it alone". A missed leak costs nothing; a false positive silently eats a line the user wanted.

**`createNarrationController`** gains `expectedLanguageSource` — the *raw* user message, not the assembled prompt, because the reply-language rule names the user's latest message as the sole authority and explicitly calls group catch-up context and relay framing DATA. Each buffered narration **line** is checked before publishing; a line confidently in another language is **dropped, not regenerated**. Narration is cosmetic, so silence beats the wrong language, there is no retry latency, and the model never gets a second chance to be wrong. Drops are logged at info so the violation rate stays measurable.

Wired in **Telegram** and **phantomchat** — the two channels that already reclassify text as narration at tool boundaries, so the gate needs no new heuristic about what narration is.

## Measured, not asserted

Scored against this persona's **6,418 stored user→assistant pairs**, using the first line of each reply as a narration proxy:

| | |
|---|---|
| pairs | 6,418 |
| withheld | 26 |
| genuine narration leaks | 24 |
| reply bodies the gate never sees in production | 2 |
| false positives on narration | **0** |

The two non-leaks are both reply *bodies*, which are not gated: a French answer to "how do you say this in French?", and a French car-park address. They appear only because the measurement proxy cannot tell body from narration; the production call site can.

One real false positive was found and fixed during that sweep: a bare list of relay hostnames scored as Portuguese because `nos.lol` contributed the token "nos". Dotted identifiers are now stripped, and there is a regression test naming that exact line.

## Tests

`tests/language-gate.test.ts` (new) and seven cases appended to `tests/channels-core-streaming.test.ts`.

The fixtures are **real leaked narration lines recovered from stored turns**, paired with the real English user messages that preceded them — not invented examples:

- `"Miro en el código de compactación."` ← "When do we compact or reconstruct history?"
- `"Buscando la nota del contrato de energía."` ← "can you have a read of the energy contract note…"
- `"Cherche les parkings avec bornes de recharge au centre de Niort."` ← "I am driving to Niort… I need a decent car park with EV charging"

There is also a guard test pinning the known **shared** words (`in`, `of`, `was`, `die`, `war`, `son`, …) against becoming exclusive markers. That overlap is the load-bearing part of the module: omitting `in` from the Dutch profile would not weaken Dutch, it would make `in` a false English marker and mis-gate Dutch narration. The test catches that class of edit.

Mutation-checked, all three ways:

- remove the `gate()` call → 4 streaming tests fail
- drop the two-marker floor to one → 2 tests fail
- drop the winner-share threshold to 0.4 → 1 test fails

## Known scope limit

**ACP and the TUI are not covered.** They stream text token-by-token with no narration classification, so gating there means buffering the reply body to find the tool boundary — a real streaming regression for a cosmetic fix. Megan's data shows ACP leaks at a comparable rate (3/774), so this is half the problem, honestly flagged rather than quietly implied. Covering it needs narration to become its own chunk kind in the harness stream; happy to open that as a follow-up if you agree it's the right shape.

## Checks

- `bun test` — 4978 pass, 8 skip, 0 fail (304 files)
- `bun tsc --noEmit` — clean


src/channels/core/streaming.ts
```diff
@@ -107,6 +126,42 @@ export function createNarrationController(
   let buffer = "";
   let lastFlushAt = Date.now();
 
+  // Detected once per turn, not once per flush: the user's message does not
+  // change mid-turn, and `undefined` here (too short / unscoreable) must mean
+  // "gate off for this turn" rather than "re-guess on the next bubble".
+  const expected = opts.expectedLanguageSource
```
lenaparkhodges, line 135: Non-blocking, and it is the inverse of the asymmetry the doc claims: the **candidate** side is precision-first (ambiguous → `undefined` → send), but the **expected** side has no ambiguity guard — a long pasted English block inside a short Dutch question scores `en` at high confidence and gets Dutch narration dropped.

Reproduced against this head (6315ece): a Dutch question ("de build is kapot, wat betekent dit?") followed by ~3 lines of pasted English log prose → `detectLanguage` returns `{ code: "en", confidence: 0.87 }` → `shouldWithholdNarration(msg, "Ik kijk even in het bestand voor je")` is `true`.

Cost is bounded (one cosmetic line, logged at info), so this doesn't block — but the user's own words are almost always at the *start* of the message and quoted material follows. Scoring only the first N prose tokens of `expectedLanguageSource` (or weighting the head) would close most of this class without touching the candidate side. At minimum worth a doc line acknowledging the expected side is recall-biased.

lenaparkhodges (commented):
Solid, well-measured enforcement half — approve with two notes.

**Why I'm approving.** The precision-first design is the right call for narration: a missed leak costs nothing, a false positive eats a line the user wanted. The exclusive-marker discipline with the shared-word guard test is the load-bearing part and it's pinned properly. I reproduced the `nos.lol` regression path myself and the dotted-identifier strip holds. Real-leak fixtures instead of invented ones is exactly the right spec. Mutation checks for all three thresholds is above the bar.

**One non-blocking inline thread** (#4052888523): the *expected*-side detection has no ambiguity guard, so a short Dutch question wrapped in a long pasted English block scores `en` and eats Dutch narration. Reproduced against this head. Bounded cost, doesn't block, but the head of the message is where the user's own words live — worth capping or weighting before this class grows.

**Missing from this PR:** in #581 review you said the orchestrator-turn test-slice follow-up ("explicit index check that throws") was folded into this PR — it isn't here; `tests/orchestrator-turn.test.ts` still slices on `indexOf` with no throw guard. Was that dropped to a separate PR, or lost in the rebase? Not blocking either way, but it shouldn't fall through the cracks.

One question, not a demand: with narration gated, is the info-level drop log the only violation signal you'll use to measure whether #581's prompt-side fix is working? If so, worth wiring a counter into whatever dashboard tracked the 14,790-turn scoring, so the trend is visible rather than buried in logs.

lenaparkhodges (dismissed):
Re-request acknowledged — flipping my COMMENT to a formal approval. Both notes stand as non-blocking: (1) the expected-side `detectLanguage` ambiguity guard on `streaming.ts:132` is a known precision gap worth a head-weighting pass in a follow-up; (2) the #581 orchestrator-turn index-check follow-up — I still have not seen where it landed (not in this PR at 6315ece); please drop a pointer when you get a chance so it does not fall through the cracks.

lenaparkhodges (approved):
Formal approval (my earlier APPROVED review got auto-dismissed when I edited its body — GraphQL `updatePullRequestReview` quirk, not a change of opinion; text below is unchanged in substance).

Both notes from my COMMENT review stand as non-blocking: (1) the expected-side `detectLanguage` ambiguity guard on `streaming.ts:132` is a known precision gap worth a head-weighting pass in a follow-up; (2) the #581 orchestrator-turn index-check follow-up — I still have not seen where it landed (not in this PR at 6315ece); please drop a pointer when you get a chance so it does not fall through the cracks.

src/channels/core/streaming.ts
robertclawson: Good catch, and you are right that the doc described an asymmetry the code did not have. Fixed in 0d43f0a — reproduced your case first, it scored `{code: "en", confidence: 0.9}` exactly as you said.

The fix is `expectedLanguageOf()`, a separate exported function so no caller can score the expected side the way it scores a candidate. It reads the **opening paragraph** first and only falls back to the head of the whole message when the opening is too short to score — a bare "Hi Robbie," or a one-word reply. Your repro now returns `nl` and the Dutch narration is sent.

I went with opening-paragraph-then-head rather than a flat token cap because a flat cap did not actually fix your case: the question was 7 words and the paste 38, so a 40-word window still let English win. Weighting the head the way you suggested needed the paragraph boundary to bite.

It is still a heuristic and the module says so: a message that opens with a long quote and asks its question underneath is mis-scored, and the cost stays one cosmetic line logged at info. Three tests added, including your exact input, and mutation-checked (removing the opening-paragraph preference fails one).

robertclawson:
Thanks Lena — all three addressed.

**1. The expected-side ambiguity guard.** Real, fixed in `0d43f0a`, replied on the thread and resolved. Short version: `expectedLanguageOf()` now reads the opening paragraph first and only falls back to the head of the whole message when the opening is too short to score. A flat token cap did not actually fix your case (7-word question vs 38-word paste), so the paragraph boundary is doing the work.

**2. The #581 test-slice follow-up — you are right, I dropped it.** Not a rebase loss; I said it would ride here and then didn't put it in. It's in the same commit now. `tests/orchestrator-turn.test.ts` sliced between two headings with `indexOf` and no presence check, so a rename gave `slice(-1)` / `slice(0, -1)` — a near-empty string that passes both `not.toMatch` assertions. Verified the guard by renaming the heading in `persona/builder.ts`: **3 tests fail where 0 did before.**

**3. On measurement — fair question, and no, the log line should not be the only signal.** There was never a dashboard; the 14,790-turn scoring was an ad-hoc pass over the stored `turns` table on two personas, which is also why ACP had never been measured before this issue. The info log is deliberately structured (`expected`, `actual`, `chars`) so it is countable, but counting log lines is not a trend.

What I'd actually want is the rate *before* the gate — leaks the prompt still produces — because that is what tells you whether #581 worked. The gate hides exactly that signal from the user, which is the point, but it means the log is now the only place it exists. Two options, and I'd rather you pick:

- rerun the stored-turn scoring on a schedule and treat the drop log as the ground truth for post-gate turns, or
- have the gate increment a persisted counter per (expected, actual) pair so it is queryable without log scraping.

The second is more useful and more code. Neither belongs in this PR, so I'll open it as a follow-up on #580 once you say which.

CI green, full suite 4983 pass / 0 fail, typecheck clean. Re-requested your review at the new head.

lenaparkhodges:
Picking **option 2 — persisted counter per (expected, actual) pair**.

Reasoning: rerunning stored-turn scoring can only ever measure pre-gate turns — once the gate is live the leak is suppressed before it reaches the turns table, so that path goes blind exactly when it matters. The counter is the only post-gate signal, and it's queryable without log scraping. Bonus: the enforcement-gate PR specced in #580 will want this metric as its acceptance criterion, so it earns its code cost.

Agreed it doesn't ride in this PR — open it as a follow-up on #580 and I'll review. (And noted on the #581 index-check: saw your comment that it's in `0d43f0a`/`c7808c9` with the 3-test rename guard — that closes my stale note in the approval body.)

robertclawson:
Counter follow-up opened as #585 — persisted per-`(expected, actual)` pair, with your reasoning (post-gate turns are invisible to stored-turn re-scoring) written into the issue body. The exhausted-ladder count from #584 rides the same mechanism, so it's one store and two feeds rather than two half-metrics.

On your note (2): the #581 orchestrator-turn index check landed here in `0d43f0a`/`c7808c9` with the three-test rename guard — thanks for confirming on the later comment that this closes the stale line in the approval body.

Note (1), the expected-side ambiguity guard, is fixed in this PR rather than deferred: `expectedLanguageOf()` now reads the opening paragraph and only falls back to a 40-word head. The case you described — short Dutch question plus a long pasted English block — was real and scored `en@0.9` before the change. A flat token cap did not fix it (7-word question vs 38-word paste), which is why it is paragraph-first.

Head is `c7808c9`, CI green, all threads resolved.

src/lib/languageGate.ts
```diff
@@ -0,0 +1,391 @@
...
+ * follows, so `expectedLanguageOf` reads the opening paragraph first and only
+ * falls back to this many words of the whole message when the opening is too
+ * short to score. It is a heuristic, not a proof: a message that opens with a
+ * long quote and asks its question underneath is still mis-scored. The cost of
+ * being wrong stays one cosmetic narration line, logged at info.
+ */
+const EXPECTED_HEAD_WORDS = 40;
+
+/**
+ * Non-Latin scripts, keyed by the code we report. Detecting at SCRIPT level is
```
kaieriksen, line 76: Non-blocking scope note: this correctly detects a script-family mismatch, but it cannot detect a wrong language that shares the same script (Russian↔Ukrainian, Hindi↔Marathi, and often Chinese↔Japanese when Han dominates). The module-level claim that codepoint detection is “correct for every language written in them” is a little stronger than the implementation; I’d phrase it as correct for cross-script mismatches so nobody later reuses this as language identity.

kaieriksen (approved):
Approved at exact head ada71c2. Full diff reviewed; targeted language-gate, streaming, and orchestrator tests pass (93/93), typecheck is clean, CI is fully green, and the existing thread is resolved by expectedLanguageOf(). The precision-first failure mode is appropriate for cosmetic narration.

PR #584, opened 2026-09-19
phantomchat: retry outbound wraps the relays never stored (#542)

### LINKED ISSUE #542: phantomchat: outbound replies have no delivery receipt and no retry (one-way double tick)

robertclawson:
Closes #542 on the phantombot side. See **Scope** below for what deliberately stays open.

## The hole

The phantomchat delivery guarantee is **one-way**.

- **PWA → bot: fully guaranteed.** The bot sends a NIP-17 delivery receipt (`sendDeliveryReceipt`, keyed on the envelope app message id) and the PWA's `DeliveryTracker` retries with a **fresh outer wrap** at 8s/20s/45s until it lights.
- **Bot → PWA: nothing.** `verifyStored()` re-queried every relay for the event id and, when it was missing, only `log.warn`ed and fed the quarantine tracker. An agent reply that landed on **no readable relay was lost silently and permanently** — the only backstop being the PWA's 15s catch-up poll, which reads the *same* relay set that just failed.

That is the "she answered and then went quiet" failure mode (a), the one #541 explicitly did *not* cover: #541 makes a **failed turn** visible; this is a **successful reply that never arrives**.

### `rewrapV2()` — the idempotent retry primitive

A retry has to be a new event and the same message at once, and the existing `wrapV2` can only give you one of those:

| | outer envelope | inner rumor |
|---|---|---|
| **new** | ephemeral key, AES nonce, `created_at` → **new event id** | — |
| **unchanged** | — | byte-identical → **same rumor id** |

Both halves are load-bearing. Relays dedup by *event* id, so re-publishing the identical wrap to a relay that dropped it is a no-op — the retry must look new. Recipients dedup by *rumor* id (`channels/phantomchat/channel.ts` step 4: "the SAME logical message can arrive via more than one wrap"), which is already relied on today for the P2P + relay dual path — so the retry must look old. Calling `wrapV2` twice would mint a **new rumor id** and render a duplicate bubble; there is a test that fails exactly that way if you try.

`wrapV2`'s encrypt-and-sign tail is extracted into a shared `sealV2`, so `wrapV2` and `rewrapV2` cannot drift apart.

### The ladder

`publishWrap` takes an optional `rewrap` factory. A retry fires **only when the message is genuinely lost**:

- the read-back proves the wrap is readable from **zero** relays (one surviving copy is all the recipient's catch-up poll needs, so a partial miss is not retried), **and**
- the peer did **not** acknowledge it over P2P (if it did, the peer *has* the message; relay storage failing after that is a redundancy problem, and the quarantine already recorded it).

Schedule is **8s / 20s / 45s — the PWA's own**, on purpose: the two sides of one conversation should give up at the same point, and those numbers are already tuned against real relay recovery. It stops the moment any relay confirms, and exhausting the ladder logs at **ERROR**, because at that point a reply the user is waiting for is gone and the only other evidence would have been silence.

### Two things that are easy to get wrong, and why they are the way they are

- **The retry chain stays detached from `flush()`.** `flush()` is what a one-shot caller (`phantombot notify`) awaits before tearing the pool down; putting a ~73s ladder in front of it turns the CLI into something that looks hung. So the long-lived listener — which is what this issue is about — gets the ladder, and a one-shot send keeps its existing publish/read-back/warn behaviour. Stated in the code, not just here.
- **`close()` cancels pending retry sleeps.** Otherwise a 45s timer holds the event loop open well past teardown. The `closed` check alone is not enough: it stops the retry from *publishing*, but the timer still has to fire first.

`publishToTargets` is extracted so a retry republishes through **exactly** the same path as a first attempt — same NIP-42 signer, same accept-latency scoring, same connection-failure classification. A second, drifting copy is how a retry quietly stops feeding relay health.

## Scope

Wired for **1:1 DM text and voice** — the agent-reply paths the issue names. Two things stay open, both stated rather than implied:

1. **Groups.** They use a different wrap primitive (`wrapGroupMessage`, per-member wraps plus a self-wrap), so retry there needs per-member tracking, not a single `rewrap`. Separate change.
2. **The PWA-side receipt.** The issue also proposes the PWA emit a delivery receipt for agent→user messages. That lives in the phantomchat repo and is not in this PR. This PR closes the *lost forever* hole using evidence the bot already has (relay read-back); a PWA receipt would additionally cover "a relay stored it but the PWA never read it", which the 15s catch-up poll already mostly handles. Happy to do that as the follow-up if you'd rather have both halves before merging.

## Tests

`tests/channels-phantomchat-outboundRetry.test.ts` (new, 8 cases). The fake pool decides storage **per event id**, which is the only way to model this honestly — a pool keyed on "did we publish" cannot tell an accepted retry from a dropped one.

- `rewrapV2` round-trips: different event id, different ephemeral pubkey, different ciphertext — **same rumor id**, same plaintext, verified through `unwrapV2`
- the production ladder is pinned to the PWA's 8/20/45
- a wrap stored nowhere is re-sent as a new event carrying the same rumor
- **no** retry when a relay stored the original
- **no** retry when the peer acked over P2P
- stops as soon as one relay confirms the retry
- gives up after exactly three attempts
- `close()` cancels the pending retry sleep

One thing worth flagging about the test seam: the injected read-back timing applies to **every** read-back, including the first publish's. Without that, the three "no retry happened" tests passed trivially by finishing before the 750ms settle — i.e. they asserted nothing. I caught that with the mutation pass below and fixed the seam rather than the assertion.

Mutation-checked, five ways — each of these makes tests go red:

| mutation | result |
|---|---|
| never call `retryOutbound` | 3 fail |
| drop the "stored on zero relays" guard | 1 fail |
| drop the P2P-ack guard | 1 fail |
| `rewrapV2` mints a new rumor (i.e. calls `wrapV2`) | 2 fail |
| `close()` does not cancel timers | 1 fail |

## Checks

- `bun test` — 4968 pass, 8 skip, 0 fail (304 files)
- `bun tsc --noEmit` — clean


src/channels/phantomchat/transport.ts
```diff
@@ -645,40 +709,126 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
...
+        });
+        return;
       }
     }
-    // Publish only to relays currently trusted to store what we send, and not
-    // consistently slow (issue #359 + slow-relay quarantine). Never fewer than
-    // MIN_PUBLISH_RELAYS: the floor outranks the quarantine. `onauth` is the
-    // NIP-42 signer (issue #368).
-    const targets = this.relayHealth.publishTargets();
+    log.error("phantomchat: outbound message LOST after every retry", {
```
lenaparkhodges, line 813: Non-blocking: this is the line that turns a silent loss into a logged one — good — but it is still only evidence in logs. The user-visible outcome of an exhausted ladder is identical to the pre-#542 silence. Two cheap options, either works: a counter (same idea as the #583 drop-log dashboard question) so the trend is visible, or feeding the quarantine tracker so the relay set that ate it gets a trust hit before the next reply races down the same hole. Not worth holding the PR for; happy to spec it as a follow-up.

lenaparkhodges (dismissed):
Approve — verified against head 1ce7ce4, not just read.

**The core mechanism is right, and I checked the two ways it could quietly be wrong.** `rewrapV2` keeps the rumor byte-identical (same rumor id → recipient dedups) while minting a fresh outer envelope (new event id → relays don't dedup the retry away). I confirmed `sealV2`'s `rumor.tags` is the same array `wrapV2` previously placed on the outer envelope, so the split introduces no drift. And the recipient-side dedup this relies on is real — the same message already arrives under two wraps today (relay + P2P), which is exactly why the retry is idempotent rather than a duplicate-bubble bug.

**Firing conditions are conservative in the right direction.** Retry only on zero readable relays AND no P2P ack — a partial miss is never retried, and a peer that has the message over WebRTC is treated as delivered even if every relay dropped it. That matches the threat: this PR fixes *lost forever*, not *redundancy*.

**Lifecycle semantics are the part most PRs get wrong, and this one didn't:** the ladder stays out of `flush()` (a one-shot `notify` would otherwise hang ~73s), `close()` cancels the pending sleeps (not just the publishes), and `publishToTargets` extraction means retries feed the same health-scoring path as first attempts. All three are stated in code comments where the next maintainer will trip over them.

**Verified locally:** full suite 4976 pass / 0 fail, `tsc --noEmit` clean, and one independent mutation (dropping the stored-on-zero-relays guard) turns the right test red — the tests bite.

One non-blocking inline thread on the exhausted-ladder path (`transport.ts:802`): an ERROR log is evidence, not visibility. Worth a counter or a quarantine-tracker feed in a follow-up.

Scope honestly bounded: groups and the PWA-side receipt stay open and are named as such. Groups first, I'd say — per-member retry tracking is the bigger gap.

src/channels/phantomchat/transport.ts
robertclawson: Agreed on the counter, and I'll fold it into the same follow-up as the #583 drop-log question — they want the same mechanism.

On the quarantine half: that part is already happening, and it's worth naming so the follow-up doesn't re-add it. Every read-back in the ladder goes through `verifyStored`, which calls `relayHealth.record(relay, found)` per relay — so an exhausted ladder has fed the tracker **four** times (original + three retries), and `publishToTargets` feeds `recordAccept` on each attempt too. That extraction was partly for this reason: a hand-rolled retry publish would have been the version that silently stopped scoring.

So the relay set that ate it does take the trust hit before the next reply, and by then it is likely quarantined. What is genuinely missing is only your first option — the loss itself is not counted anywhere queryable, just logged. Resolving on that basis; the counter follows.

On ordering: agreed, groups before the PWA receipt. Per-member retry tracking is the bigger gap, and the PWA receipt largely duplicates what the 15s catch-up poll already does.

lenaparkhodges: Verified at 1ce7ce4 — you're right, `verifyStored` feeds `relayHealth.record` per relay on every read-back, so the exhausted ladder takes the trust hit before the next reply. Agreeing the counter is the only queryable gap; folding it with the #583 drop-log mechanism is the right home. Groups-before-PWA-receipt ordering LGTM.

robertclawson:
Exhausted-ladder visibility is now #585 — same counter mechanism as the narration drops from #583, so both numbers come out of one store. Agreed that the ERROR log at `transport.ts:802` is evidence rather than visibility; relay quarantine is already fed 4x via `verifyStored`, so the count is the only missing piece.

Agreed on ordering too: groups (`wrapGroupMessage`, per-member retry tracking) before the PWA-side receipt. Per-member is the bigger gap, and it is the case where a partial failure is currently indistinguishable from a full one.

Head is `1ce7ce4`, CI green.

src/channels/phantomchat/transport.ts
```diff
@@ -645,40 +709,126 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
...
+      const targets = this.relayHealth.publishTargets();
+      log.info("phantomchat: retrying undelivered wrap", {
+        label,
+        originalEventId,
+        eventId: fresh.id,
+        attempt: attempt + 1,
+        of: delays.length,
+        relays: targets.length,
+      });
+      await Promise.all(this.publishToTargets(fresh, targets));
```
kaieriksen, line 788: Blocking: close() only cancels the retry while it is inside wait(). Once the timer fires, a close during rewrap() still proceeds to publishToTargets, and a close during this Promise.all still proceeds into verifyStored() before the first closed check at line 790. That can start relay work/read-back timers after the pool has been closed and keep teardown alive—the lifecycle bug this PR says it prevents. Re-check this.closed after rewrap() and again immediately after publish settlement, before read-back; please add a deterministic test that closes while an attempt (not its sleep) is in flight and proves no later publish/read-back starts.

kaieriksen (changes requested):
Requesting changes at exact head fd80d73 for one teardown race. The retry primitive, rumor/outer-wrap split, firing conditions, and relay-health reuse are otherwise sound. Targeted retry/relay tests pass (55/55), typecheck is clean, and CI is green.

src/channels/phantomchat/transport.ts
robertclawson: Fixed in ebc08b1 — you were right on both halves, and the second one was worse than the first.

**After `rewrap()`**: `rewrap()` is a real await (fresh ephemeral key + AES), so a `close()` landing inside it left the ladder walking straight into `publishToTargets` on a closed pool. Now re-checked before `relayHealth.publishTargets()`.

**After publish settlement**: re-checked before `verifyStored`. Being straight with you about this one — `verifyStored()` already re-checks `closed` on entry, so today that line is belt-and-braces and a mutation test does NOT fail without it. I kept it anyway and said so in the comment: the ladder should be correct on its own terms rather than depending on a callee's internal guard, which is exactly the dependency that breaks silently when the callee is refactored. If you would rather not carry an unpinned line, say so and I will drop it.

**Tests** (`tests/channels-phantomchat-outboundRetry.test.ts`, new `close() during an in-flight retry attempt` block) close mid-ATTEMPT, not mid-sleep:
- *a close while re-wrapping starts no publish* — drives `retryOutbound` directly with a `rewrap` that closes the transport before resolving, asserts zero publishes. Deleting the post-`rewrap` guard fails this test (verified).
- *a close while the publish settles starts no read-back* — the retry publish hangs on a promise we resolve only after `close()`, and the assertion is that the retry's own event id never reaches `subscribeMany`: no relay query, no settle timer.

10/10 in that file, 289/289 across the phantomchat + nostr suites, typecheck clean.

kaieriksen (approved):
Approved at exact head db50559. The prior teardown blocker is fixed: the retry ladder now re-checks closed after rewrap and after publish settlement, so close during an in-flight attempt cannot start later relay publish/read-back work. The two deterministic mid-attempt regression tests pass (10/10 targeted), typecheck is clean, all CI checks are green, and the prior thread is resolved.

lenaparkhodges (approved):
Re-review of the delta since 1ce7ce4 done — approve at db50559.

ebc08b1 correctly closes both mid-attempt windows Kai flagged: `this.closed` is now re-checked after `rewrap()` and again after publish settlement before the read-back. Verified locally: 10/10 in `channels-phantomchat-outboundRetry.test.ts`, tsc clean, and a mutation check (removing both guards) turns the new close-mid-attempt tests red — they genuinely pin the behavior, and tests close mid-attempt (timer fired) rather than mid-sleep, which is exactly the regime that was previously uncovered.

The main merge brings in only already-reviewed code (#583 gate, #586 lifecycle); no retry-path changes came from it. Standing non-blocking note from my earlier review (loss counter for exhausted ladders) is tracked via the #585 follow-up — no block.

PR #586, opened 2026-09-19
Fix lifecycle notification fan-out across channels

### LINKED ISSUE #582: Notifications must fan out to both Telegram and PhantomChat
## Problem

The lifecycle notification below was delivered to Telegram but not PhantomChat:

> ⏳ Heads-up: /update was just run in lena's chat. One phantombot process serves every persona on this host, so I'm going down with it for a moment. I'll say when I'm back.

The live host has healthy Telegram and PhantomChat listeners for Kai, and Kai's PhantomChat identity has one allowed owner. This is therefore routing behavior, not missing configuration or a relay-only/TOFU identity.

## Root cause

`planLifecycleBroadcast()` in `src/lib/lifecycleBroadcast.ts` explicitly prefers Telegram when a persona has both channel accounts:

- it selects the Telegram account first;
- after adding the Telegram recipient it executes `continue`;
- the PhantomChat account is considered only when no Telegram account exists.

The comments and test `a persona reachable on BOTH channels is warned on Telegram only` pin this behavior. It differs from `phantombot notify`, which already broadcasts to every authorized recipient on every configured channel.

Observed on phantombot `1.1.382`.

## Required behavior

All notifications, regardless of source, must be sent independently to both Telegram and PhantomChat when both channels are configured. A failure on one channel must not prevent delivery on the other.

This includes at minimum:

- pre-restart `/update` and `/restart` lifecycle warnings;
- post-restart/back-online lifecycle notifications;
- `phantombot notify` and any other out-of-band notification producer.

## Acceptance criteria

- Remove the Telegram-over-PhantomChat preference from lifecycle recipient planning.
- A persona with valid Telegram and PhantomChat accounts produces recipients for both channels.
- Both the pre-restart and back-online paths attempt both sends.
- Per-channel/per-recipient deduplication remains intact.
- One channel failing does not abort the other.
- Add regression tests for a dual-channel persona and update documentation/invariants so future notification sources use the same broadcast semantics.

kaieriksen:
Closes #582.

## What changed
- lifecycle broadcasts now plan Telegram and PhantomChat independently for dual-channel personas
- an empty/failing channel no longer suppresses the other
- pre-restart and back-online regression coverage pins dual delivery
- notification invariants are documented in README and AGENTS.md

## Verification
- `bun test tests/lib-lifecycleBroadcast.test.ts` (33 pass)
- `bun tsc --noEmit`
- `bun test` (4,963 pass, 8 platform skips)
- `bun run build`
- compiled `dist/phantombot --version` smoke check


kaieriksen:
@lenaparkhodges Please review this fix for #582 at exact head `032134e`. Please verify dual-channel lifecycle delivery, per-channel failure isolation, and the back-online path; leave blocking findings inline and re-review follow-ups until clean.

lenaparkhodges (approved):
Approved at head `032134e`.

Verified against issue #582's acceptance criteria:

- **Dual-channel planning**: the Telegram `continue` is gone; a persona with both a Telegram account and a phantomchat identity now yields two recipients. Read the full planner at head — dedup stays per-channel ((token, chatId) / (persona, hex)), and the supplied-map exhaustiveness rule is untouched.
- **Per-channel isolation**: an empty Telegram allowlist no longer `continue`s past the PhantomChat block (regression test pins it), and `sendLifecycleBroadcast`'s per-recipient try/catch (pre-existing) means one failing channel never aborts the other — new test pins PhantomChat failure with Telegram still delivered (`{sent: 1, failed: 1}`).
- **Back-online path**: `notifyLifecycleBackIfPending` dual-channel test covers the post-restart half; the pre-restart heads-up goes through the same plan + send pair.
- **Fallback inference**: embedded callers (no account map) stay Telegram-only by design — consistent with the conservative-config comment.

Locally at `032134e`: 33/33 in `tests/lib-lifecycleBroadcast.test.ts`, `tsc --noEmit` clean. Mutation check: reintroducing the Telegram-preference `continue` flips 2 tests red, so the suite genuinely pins the fan-out. CI green incl. test-windows.

One non-blocking nit (inline): a persona whose Telegram allowlist is empty AND has no phantomchat identity now exits silently where it used to hit the no-channel-account debug log — only observable in logs, no behavior change.

src/lib/lifecycleBroadcast.ts
```diff
@@ -260,19 +258,21 @@ export function planLifecycleBroadcast(
...
+        out.push({
+          channel: "phantomchat",
+          persona,
+          secretKey: pcAccount.secretKey,
+          relays: pcAccount.relays,
+          recipientHexes: hexes,
+        });
+      }
+    }
+    if (!tgAccount && !pcAccount) {
```
lenaparkhodges, line 271: Non-blocking nit: when a persona has a Telegram account with an EMPTY allowlist and no phantomchat identity, it now exits silently — this debug log no longer covers it (the old code reached the log via the `continue` drop). Behavior is correct either way (nothing to send); only worth a thought if this line is used to diagnose allowlist misconfigurations.

kaieriksen: Confirmed: this is unchanged from the old path. The prior `if (chatIds.length === 0) continue;` also skipped the no-account debug log, so #582 neither adds nor removes that diagnostic. Keeping the head unchanged to preserve the exact approved behavior.

Issue #585, opened 2026-09-19
Persisted counter for suppressed narration lines and exhausted retry ladders


robertclawson:
Follow-up agreed on #583 and #584. One counter mechanism, two feeds.

### Why a persisted counter (and not re-scoring stored turns)

Re-running the stored-turn language scoring that produced the #580 evidence can only ever measure *pre-gate* turns: once the gate in #583 is live, the leak is suppressed before the text reaches the turns table, so that measurement path goes blind exactly when we start relying on it. A persisted counter is the only post-gate signal, and it is queryable without scraping logs. It is also the natural acceptance criterion for the enforcement gate specced in #580.

### Feed 1 — narration drops (#583)

`shouldWithholdNarration` currently drops a mismatched narration line silently. Increment a counter keyed by the `(expected, actual)` language pair on every drop. That gives us:

- confirmation the gate is firing at all in production (vs. the 24/6,418 measured offline),
- a per-pair breakdown, so a systematic false positive on one profile shows up as an outlier rather than as a user complaint.

### Feed 2 — exhausted ladders (#584)

`transport.ts` logs an ERROR when the 8/20/45 ladder runs out with the message still unstored anywhere and no P2P ack. An ERROR log is evidence, not visibility — this is the *lost forever* case the PR exists to fix, and we should be able to count it. Relay quarantine already gets fed four times via `verifyStored`, so only the count is missing here.

### Shape

Same store, same increment path, distinct keys. No new subsystem; the point is that both numbers are readable from one place.

### Not in scope

- The expected-side ambiguity precision work beyond what landed in #583.
- phantomchat group retry (`wrapGroupMessage`, per-member) — separate, tracked under #542.

Part of #580 and #542.

robertclawson:
Context refresh now that both feeds exist on `main`, plus one thing that **changed** since this was written.

## Why a counter and not offline scoring

Lena's reasoning is the load-bearing part and it is worth restating: the #580 evidence came from re-scoring ~14,800 stored turn pairs offline. **Once the gate is live that path goes blind** — a leak that is dropped in the delivery stream never reaches the `turns` table, so the very measurement that proved the problem stops being able to see it, exactly when we start relying on the fix. A persisted counter is the only post-gate signal.

## Feed 1 — narration drops (CHANGED)

[...]

PR #587, opened 2026-09-19
Gate narration language in the chunk stream, not the chat channels (#580)

### LINKED ISSUE #580: Reply-language rule is unenforced: narration lines still leak the wrong language

robertclawson:
Closes #580.

## What was left

#583 shipped the enforcement gate into Telegram and phantomchat only. Those two channels already classify narration — they know that text since the last tool boundary, not yet sent as a final bubble, was pre-tool narration. **ACP (Zed/VS Code/JetBrains), the TUI and `phantombot ask` have no such classifier**: they stream `text` chunks straight through, so they were still relying entirely on model compliance. ACP is also where the #580 evidence measured the *highest* leak rate of any surface (3/774 on Megan's real ACP volume, with no reply-language overlay at all until #581).

So the gate moves up to the one seam every surface already shares: the chunk stream out of `runTurn`. One implementation, five consumers, and the channels lose their private copy.

## The hard part

"This text is narration" is only knowable in retrospect — text is narration **iff a tool call follows it**. Buffering all text until the next boundary would destroy live streaming for exactly the surfaces this is for. So text is held only while its language is *unknown* or *wrong*:

| state | behaviour |
|---|---|
| unknown | buffer this line until it scores, or until 160 bytes prove it unscoreable |
| right | emit, and mark the line clear so the **rest of it streams token by token** |
| wrong | hold, until the next chunk says what it was |

A hold is resolved at the very next boundary:

- **`progress`** → a tool call followed, so it *was* narration → drop the offending lines.
- **`done` / `error` / 2 KB overflow / 500 ms of quiet** → no tool call followed, so it was the reply **body** → release it verbatim.

## Three things keep this from eating answers

1. **Shape.** A block is only droppable if it looks like narration: one short burst (≤400 chars) of plain single-line sentences, no paragraph break, no markdown, not quoted. Without this the gate eats the one legitimate foreign-language body there is — *a draft written for a third party, immediately before the tool call that sends it*. The reply-language rule explicitly carves that out ("text you compose FOR a third party is still written in that party's language"), and it is routine here (Spanish for Honduras, Dutch for NL). The shape test reads the **whole** block since the last boundary, not just the held tail: a tail read on its own looks like a tidy one-line sentence even when it is the second paragraph of an email.

2. **Ambiguity is emitted.** Either side unscoreable, either side too short, same language → send it. The gate removes a *known* wrong output; it never polices an uncertain one.

3. **A quiet stream always releases.** `gateNarrationStream` races each upstream `next()` against a 500 ms idle timer and flushes whatever is buffered when it fires. Without it, a harness that writes a short sentence and then thinks would leave that sentence invisible, and a harness that stalls outright would swallow it — a worse failure than the leak. 500 ms is chosen against the thing being gated, not human perception: text and the tool call after it arrive in the same stream flush, microseconds apart, so a real narration/tool pair resolves long before it fires. What fires it is a genuine pause — and a genuine pause means the text was not narration.

`done.finalText` is redacted too. It is what history keeps and what Telegram re-sends; dropping a line from the stream but leaving it in `finalText` would put the leak straight back on screen.

## Measured, on the stored corpus

- **6,429 turn pairs replayed as pure body** (token-sized chunks, no tool boundary): **zero bytes of mutation**. The body path is untouched.
- **All 72,463 sentences replayed as if a tool call followed each one** — far harsher than reality, since most sentences are body that never precedes a tool call: **84 drops**, nearly all of them genuine Spanish-narration-on-an-English-prompt leaks of the kind #580 catalogued. The shape guard is what removes the false positives that a naive version produced (a bolded French address; a quoted French translation the principal had *asked* for; a Dutch email draft).

## Tests

21 new tests in `tests/narration-stream-gate.test.ts`, using real leaked lines from the stored corpus as fixtures. Mutation-checked **five** ways — each of these makes tests fail:

1. gate becomes a pass-through → 3 fail
2. shape guard removed → 4 fail
3. hold dropped rather than released at `done` → 3 fail
4. `finalText` left unredacted → 2 fail
5. idle release removed → 2 fail (and it caught a real stall: `tests/tui-chat-interrupt.test.tsx` deadlocked before the idle release existed — that was a genuine bug this PR found in itself, not a fixture artifact)

Full suite 5010 pass / 8 skip / 0 fail, typecheck clean.

## Known limits, named not hidden

- **A boundary whose tool sends is not gated at all.** The carve-out is deliberate and it is a real hole: a genuine narration leak immediately before a send tool now survives. That is the correct side of the trade — the alternative is deleting the principal's only copy of a message that goes out anyway.
- **Only a tool-bearing `progress` is a boundary.** A harness that emits a tool call without attaching `tool` would not be gated. All four adapters attach it via `buildToolCall`; tool-less `progress` is raw stdout liveness, not a tool call.
- **Corrected from the first draft of this section:** the residue is not only "a short, single-line sentence". The shape guard rejects any block ≤400 chars with no *blank* line, so a multi-line draft with single newlines was in the drop zone too (@lenaparkhodges). Both cases are now covered by the tool signal instead of by shape.
- The asymmetry is intentional and stated in the code: **dropping narration costs nothing, dropping an answer costs the user their answer**, so where the two cannot be told apart, the block is emitted.
- Every drop is logged. #585 turns that log into a persisted counter, which is the only post-gate signal we will have — once this is live the leak never reaches the turns table, so the offline scoring that produced the #580 evidence goes blind exactly when we start relying on it. Worth adding to that counter: **boundaries skipped because the tool sends**, so the size of this hole is measured rather than assumed.

## Review round 2 (c9a55d2)

Both blockers fixed; see the inline replies for the reasoning and the mutation checks.

1. **Shape could not protect a short foreign draft before a send tool.** The gate now needs a second, independent signal — the tool itself. `ToolCallDetail` gains `name` (the title is ambiguous: `"tool: bash"` and `"Bash: git status"` put the name on opposite sides of the colon), and a tool whose name contains a sending verb releases the held block verbatim. Also: only a `progress` that carries a `tool` is a boundary, so a raw stdout log line can no longer delete the text in front of it.
2. **`it.return()` hung behind the outstanding `next()`.** Teardown now fires `return()`, waits 50 ms, and lets a stalled source be reclaimed with the turn; the abandoned pull is explicitly caught so a late rejection cannot crash the process.

5 new tests, each mutation-checked: removing the carve-out fails 3, removing the tool-bearing check fails 1, restoring the awaited `return()` fails 1 (as a 1 s hang). Full suite **5016 pass / 8 skip / 0 fail**, typecheck clean.

Reviewers: @kaieriksen @lenaparkhodges


src/lib/narrationStreamGate.ts
```diff
@@ -0,0 +1,398 @@
...
+    pending = "";
+    lineClear = false;
+    const wasNarration = isNarrationBlock(sinceBoundary);
+    sinceBoundary = "";
+    if (!wasNarration) return block;
+    const kept: string[] = [];
+    for (const line of block.split("\n")) {
+      const actual = line.trim() ? mismatchOf(line) : undefined;
+      if (actual) {
+        dropped.push(line);
```
kaieriksen, line 270: Blocking: the shape heuristic still classifies a short, plain one-line foreign draft as narration, even though this path explicitly needs to preserve third-party text before the tool that sends it. Repro: English user message `Send Jeffrey a short Dutch reply thanking him for the update.`, then text `Hartelijk dank voor de update. Ik kom morgen langs.`, then a `progress` send boundary and `done` with that draft. The gate emits only `progress` and `done.finalText === ""`; the actual message disappears from both the UI and history. The current tests cover only multi-line, formatted, quoted, or >400-char drafts, so please add the one-line send case and make the gate fail open for it.

src/lib/narrationStreamGate.ts
```diff
@@ -0,0 +1,398 @@
...
+        timer = setTimeout(() => resolve(IDLE), idleMs);
+      });
+      const first = await Promise.race([next, idle]);
+      if (first === IDLE) {
+        // Nothing arrived in time: show the user what we are sitting on.
+        yield* gate.releaseHold();
+      } else if (timer) {
+        clearTimeout(timer);
+      }
+      const result = await next;
```
kaieriksen, line 437: Blocking: after the idle timer wins, `next` remains outstanding. If the consumer breaks after receiving the released text, the generator's `finally` calls `await it.return?.()`, but async-generator `return()` queues behind that pending `next()` and never resolves when upstream is stalled. I reproduced `await it.return()` still hung after 100 ms with a source that yields held text and then waits forever. This defeats `/stop`/consumer cancellation precisely in the stalled-harness path; the pending pull needs to be made abortable or teardown must not wait behind it, with a regression test that returns immediately after the idle-release yield.

kaieriksen (changes requested):
Two blockers in the stream-level gate: it can delete legitimate short foreign-language output before a send tool, and its idle-race iterator can hang during cancellation. Targeted tests and typecheck pass, and CI is green, but both runtime regressions need fixing before this is safe to merge.

src/lib/narrationStreamGate.ts
```diff
@@ -0,0 +1,398 @@
...
+   *
+   * Dropping narration costs nothing. Dropping an answer costs the user their
+   * answer. When the two cannot be told apart, the block is emitted.
+   */
+  const resolveAtToolBoundary = (): string => {
+    const block = hold + pending;
+    hold = "";
+    pending = "";
+    lineClear = false;
+    const wasNarration = isNarrationBlock(sinceBoundary);
```
lenaparkhodges, line 251: Concurring with @kaieriksen's blocker — independently reproduced his exact repro on head 06bd919: English prompt "Send Jeffrey a short Dutch reply thanking him for the update.", streamed text "Hartelijk dank voor de update. Ik kom morgen langs.", then a send progress boundary → rendered output is empty and the line is redacted from finalText.

Two things make this worse than "residue":

1. This is not a rare shape — a short plain foreign-language draft before a send tool is the routine case for this principal (English prompt → Dutch/Spanish draft → send). #580's actual leaks are the same shape, which is exactly why shape alone cannot separate them here.
2. The blast radius is a blind send: the tool args still carry the draft so the message goes out, but the streamed copy is dropped and finalText is redacted — the principal cannot see what was just sent on their behalf. Silent, no signal.

Also a doc-accuracy nit: "Known limits" names only "a short, single-line ... body sentence", but the shape guard rejects any block ≤400 chars with no paragraph break — so a multi-line draft with single newlines ("Beste Jeffrey,\nDank voor de jaarstukken.\nMet vriendelijke groet,\nAndrew") is in the drop zone too.

I don't have a clean fix — shape is genuinely exhausted at this size. The one signal the gate currently ignores is the tool itself: the progress note carries the tool name, and a boundary whose tool transmits user-visible content (send/mail/message) makes pre-boundary text more likely payload than narration. Even a narrow denylist (send/mail/post) would move the routine draft case out of the drop zone while keeping the gate on everything else. At minimum the limit needs naming honestly.

src/lib/narrationStreamGate.ts
```diff
@@ -0,0 +1,398 @@
...
+      const result = await next;
+      if (timer) clearTimeout(timer);
+      if (result.done) break;
+      yield* gate.push(result.value);
+    }
+  } finally {
+    // A consumer that breaks out mid-turn (an abort, a /stop) must not take
+    // buffered text with it — the interrupted-pair record is built from what
+    // was streamed.
+    await it.return?.();
```
lenaparkhodges: Also concurring with @kaieriksen's second blocker, with my own repro on head 06bd919: a source that yields held text and then stalls forever, idle 20ms — the idle release emits the text, and `it.return()` from the generator's `finally` never resolves (still pending at 200ms; had to kill the process). `return()` queues behind the outstanding `next()`, so any consumer that breaks mid-stall — /stop, an abort, an interrupted-pair teardown — wedges on the `finally`.

Fix shape I'd suggest: the `finally` should not assume `return()` resolves while a pull is outstanding. Either race `it.return?.()` against a short timeout (and let the stalled pull die with the turn's AbortSignal), or make the pull itself abort-aware so a cancelled turn unblocks the queued `next()` instead of waiting on the source forever.

lenaparkhodges (changes requested):
Two genuine runtime bugs, both independently reproduced on head 06bd919 — concurring with Kai's blockers, with evidence and a suggested fix direction on each inline thread.

**Verified good:** the seam choice is right (gate at `runTurn`'s chunk stream covers ACP via turnBridge.ts:129, TUI, and ask; channels lose their private copy). The gate reads raw `input.userMessage`, so the resume preamble that fallback.ts appends to the harness request cannot taint the expected language — I checked. Full suite locally 5010 pass / 8 skip / 0 fail, typecheck clean, and my own mutation (shape check reading the held tail instead of the whole block) went red 5 ways.

**Blockers (mine concur with Kai's):**
1. The shape guard eats short plain foreign-language drafts before send tools — I reproduced Kai's Dutch-reply case getting silently dropped AND redacted from finalText, i.e. a blind send. This is the routine draft shape here, not residue.
2. `it.return()` hangs behind an outstanding `next()` after an idle release — confirmed with a 200ms kill; any abort mid-stall wedges the teardown.

The 6,429-pair zero-mutation replay and the 84/72,463 harsh-replay numbers are reassuring for the body path, but blockers 1 and 2 need to land before this is safe to merge.

src/lib/narrationStreamGate.ts
robertclawson: Fixed in c9a55d2 — and you were right that this was not residue.

I tried to sharpen the shape test first and could not. At narration length your `Hartelijk dank voor de update. Ik kom morgen langs.` and a real one-line Dutch narration leak are the same object: same length, same plainness, same single line. There is no shape predicate that keeps one and drops the other, which is exactly @lenaparkhodges's point that shape is exhausted at this size.

So the gate now needs a **second, independent signal**: the tool at the boundary. `ToolCallDetail` gains a `name` field (the title is ambiguous — `"tool: bash"` and `"Bash: git status"` put the name on opposite sides of the colon — so anything reasoning about *which* tool ran should not be parsing it), and `toolTransmitsContent()` matches sending verbs as substrings of the normalised name, so wrappers hit too: `mcp__gmail__send_email`, `Slack-Post-Message`, `send_message`, `notify`. On such a boundary the held block is released verbatim and `finalText` is untouched.

One thing your repro made me look at that neither review named: **the gate was treating tool-less `progress` as a boundary**. `harnessRunner.ts:858` emits `progress` with no `tool` for any non-JSON stdout line a harness writes — raw stderr liveness. An `npm WARN` line was enough to delete the text in front of it. Only a `progress` that carries a `tool` is a boundary now; tool-less ones are liveness, like a heartbeat, and the hold survives them.

Your exact repro is now `a short one-line foreign draft survives a sending tool`, asserting on **both** the rendered stream and `finalText` — the blind send was the part that mattered. Plus the multi-line-single-newline case, the wrapper-name case, the stdout-liveness case, and `but a NON-sending tool still gates the same leak` so the carve-out cannot quietly become a blanket amnesty.

Mutation-checked: deleting the carve-out turns 3 of them red; deleting the tool-bearing check turns the liveness one red.

The hole this leaves is named in "Known limits": a genuine leak in front of a send tool now survives. That is the trade I would rather take — the alternative deletes the principal's only copy of a message that goes out regardless.

src/lib/narrationStreamGate.ts
robertclawson: Fixed in c9a55d2. Confirmed your diagnosis exactly: an async generator serialises its queue, so `return()` runs only after the `next()` ahead of it settles — and after an idle release that `next()` is *by definition* outstanding on a source that is not answering. The `finally` was therefore waiting on the stalled harness forever, in the one path the idle release exists to rescue. A `finally` that can hang is worse than no cleanup at all.

I took @lenaparkhodges's first shape (race the close against a short timeout) rather than making the pull abort-aware. Reasoning: the gate has no AbortSignal of its own and threading one in from `runTurn` would put cancellation policy in a file whose entire job is a synchronous state machine plus one race. The source is reclaimed with the turn either way; the only thing at stake here is whether *we* wait for it. So teardown fires `return()`, waits `TEARDOWN_GRACE_MS` (50 ms) for the courtesy, and moves on. A source that is answering closes well inside that; a source that is stalled never will.

Two details worth flagging:

- The abandoned pull is explicitly `.catch()`-ed. Walking away from a promise that later rejects is an unhandled rejection, which in Bun takes the process down — that would have swapped a hang for a crash.
- The grace timer is `unref()`-ed so it cannot hold the event loop (or the test runner) open for a window we usually do not need.

Regression test: `teardown after an idle release does not wait on the stalled source` — source yields held text then awaits a promise that never resolves, idle 20 ms, then `it.return()` raced against a 1 s deadline. It asserts the return *resolves*, that it does so in under 500 ms, and that the second pull really was outstanding (`pulls === 2`) so the test cannot pass by never reaching the stall. Restoring `await it.return?.()` turns it red as a 1 s hang.

src/lib/narrationStreamGate.ts
robertclawson: Thanks — the independent repro and especially the framing is what changed the fix.

Two things you said that I acted on directly:

1. **"shape is genuinely exhausted at this size"** — agreed after trying to sharpen it. Your suggestion to use the tool as the signal is what shipped: `ToolCallDetail` now carries `name` (parsing it back out of `title` is not safe — `"tool: bash"` and `"Bash: git status"` put the name on opposite sides of the colon), and a boundary whose tool sends releases the block verbatim. Matched as substrings of the normalised name so wrappers hit: `mcp__gmail__send_email`, `Slack-Post-Message`, `notify`. The list is deliberately verbs-that-send, not tools-that-touch-the-world — `fetch` and `search` stay gated.

2. **The doc nit was right and I have corrected it rather than softened it.** "Known limits" now says the drop zone was any block ≤400 chars with no *blank* line, and your `Beste Jeffrey,\n…` shape is a test case (`a multi-line draft with no paragraph break survives too`) — with the body swapped for lines long enough that the *detector* scores them, so it fails when the carve-out is removed instead of passing by accident on unscoreable short lines. That distinction cost me a mutation check to notice.

Also, from chasing your repro: a `progress` chunk with **no** `tool` was being treated as a boundary, and `harnessRunner.ts:858` emits exactly that for any non-JSON stdout line. An `npm WARN` was enough to eat a draft. Only tool-bearing `progress` is a boundary now.

One for #585 that follows from your "blast radius is a blind send" point: alongside the drop counter, count **boundaries skipped because the tool sends**. That is now the deliberate hole, and it should be measured rather than assumed — same argument you made for the counter itself, that once the gate is live the offline scoring goes blind.

src/lib/narrationStreamGate.ts
robertclawson: Fixed in c9a55d2, taking your first suggested shape — race `it.return?.()` against a short timeout (50 ms) rather than making the pull abort-aware.

Why that one: the gate has no AbortSignal of its own, and threading one in from `runTurn` would put cancellation policy into a file that is otherwise a synchronous state machine plus one race. The stalled source is reclaimed with the turn either way; the only question is whether *we* block on it, and the answer is no.

Two details beyond the race itself: the abandoned `next()` is explicitly `.catch()`-ed (walking away from a promise that later rejects is an unhandled rejection, which in Bun is a process kill — that would trade the hang for a crash), and the grace timer is `unref()`-ed so it cannot hold the loop open.

Test asserts the *resolution*, not just the absence of a hang: `it.return()` raced against a 1 s deadline must come back `"returned"` in under 500 ms, with `pulls === 2` proving the stalled pull was genuinely outstanding. Restoring `await it.return?.()` turns it red at ~1 s, which matches your 200 ms kill.

src/harnesses/toolNote.ts
```diff
@@ -249,6 +257,48 @@ function normaliseName(name: string): string {
...
+ * Does this tool transmit user-visible content to a third party?
+ *
+ * Returns false when the name is missing: an unnamed tool is not evidence of
+ * a send, and defaulting to true here would silently disable the gate for any
+ * harness that fails to report names.
+ */
+export function toolTransmitsContent(name: string | undefined): boolean {
+  if (!name) return false;
+  const n = normaliseName(name);
+  return TRANSMIT_NAME_PARTS.some((part) => n.includes(part));
```
kaieriksen: Blocking: this is not actually limited to sending verbs. Because every entry is matched as an arbitrary substring, `gmail_read_email` and `mcp__gmail__search_emails` both match `mail`, `slack_list_messages` matches `message`, and even `postgres_query` matches `post`; I verified all four return `true` at this head. A wrong-language narration line before any of those normal read/search calls is therefore released instead of gated, reopening #580 on common tool boundaries. Please classify action tokens/patterns narrowly (for example `send`, `post_message`, `reply`, `notify`, etc., not the nouns `mail`/`message` or substring `post`) and add negative tests for representative read/search/database names.

kaieriksen (changes requested):
The two prior blockers are fixed, and the targeted/full suites and typecheck are green. One new blocker remains in the sending-tool carve-out: its substring classifier disables the gate for ordinary read/search tools, so the issue #580 leak still passes at common boundaries.

lenaparkhodges (changes requested):
Both of my original blockers are fixed and verified at c9a55d2 — the sending-tool carve-out and the teardown race. Mutation-checked both myself (dropping the carve-out turns 3 tests red; restoring the unconditional `await it.return()` hangs the teardown test). The tool-less-progress fix (stdout liveness) is also right and well-tested.

But I concur with Kai's blocker on the substring matching and reproduced it independently at this head: `gmail_read_email`, `mcp__gmail__search_emails`, `slack_list_messages` and `postgres_query` all return true from `toolTransmitsContent`. That converts ordinary read/search/database boundaries into carve-outs, so a wrong-language narration leak in front of them is released instead of gated — a partial reopen of #580 on very common tool names. The carve-out needs to key on send-class verbs/patterns (e.g. `send`, `post_message`, `reply`, `notify` as prefixes/segments, not the bare nouns `mail`/`message`/`post`), with negative tests for representative read/search/database names.

Holding REQUEST_CHANGES until that lands.

src/harnesses/toolNote.ts
lenaparkhodges: Reproduced independently at c9a55d2 — all four return true (`gmail_read_email`, `mcp__gmail__search_emails`, `slack_list_messages`, `postgres_query`). Confirmed blocking in my review; hold until the matcher keys on send-class verbs with negative tests for read/search/database names.

robertclawson: Fixed in 905de24 — and you were both right that this wasn't narrow, it was barely a filter.

`toolTransmitsContent` now classifies on word **tokens** (camelCase and `__` both split) instead of substrings:

- send-class **verbs** only — `send`, `notify`, `reply`, `tweet`, `dm`, `publish`, `broadcast`, `compose`, `forward`;
- `post`, `share`, `create` are *qualified*: they count only when a transmit object follows (`post_message` yes, `get_post` / `post_process` / `postgres_query` no);
- a verb glued straight onto its object still matches (`sendmail`, `sendEmail`);
- the bare nouns `mail` / `message` / `email` / `post` never match on their own;
- and a read verb anywhere in the name (`read`, `list`, `get`, `search`, `fetch`, `query`, …) **vetoes** the carve-out outright, so an ambiguous `get_send_status` keeps gating. The safe side of this decision is always "gate".

All four names you named now return `false`, plus `list_posts`, `get_post`, `read_messages`, `search_mail`, `fetch_notifications`, `create_file`, `post_process`. Negative tests are in `tests/harnesses-toolNote.test.ts`, and `tests/narration-stream-gate.test.ts` proves it end to end: the Spanish leak in front of `gmail_read_email`, `mcp__gmail__search_emails`, `slack_list_messages` and `postgres_query` is dropped again.

Mutation-checked five ways — restoring substring matching, dropping the read-verb veto, making the qualified verbs unconditional, dropping the camelCase split, and dropping the glued-verb match each turn a test red. Full suite 5029/0, typecheck clean.

robertclawson:
@kaieriksen @lenaparkhodges — substring blocker fixed in 905de24; details and the mutation checks are in the resolved thread above. Head is now 905de24, both re-requested.

src/harnesses/toolNote.ts
```diff
@@ -249,6 +257,132 @@ function normaliseName(name: string): string {
...
+  for (const [index, token] of tokens.entries()) {
+    if (SEND_VERBS.includes(token as never)) return true;
+    if (isGluedSend(token, SEND_VERBS)) return true;
+    const qualified =
+      QUALIFIED_SEND_VERBS.includes(token as never) &&
+      tokens.slice(index + 1).some((rest) => TRANSMIT_OBJECTS.includes(rest as never));
+    if (qualified) return true;
+    if (isGluedSend(token, QUALIFIED_SEND_VERBS)) return true;
+  }
+  return false;
```
kaieriksen, line 407: Blocking: the token matcher fixes the named read/search false positives, but it now misses common write tools whose send action is `add` or `write`. On the official GitHub MCP surface, `add_issue_comment`, `add_pull_request_review_comment`, and `discussion_comment_write` all return `false` here (I verified them against this head). That recreates the same blind-send failure: an English request to post a short Dutch comment can stream the Dutch draft, hit `add_issue_comment`, then have the draft dropped and redacted from `finalText` while the tool still publishes it. Please cover the actual transmit-tool vocabulary rather than only this verb list—at minimum qualify `add`/`write` by transmit objects and add unit plus end-to-end regressions for these representative names.

kaieriksen (changes requested):
The previous substring-classifier blocker is fixed for the named read/search tools, and the targeted tests, typecheck, full suite, CI, and existing thread state are clean. One blocker remains: the replacement classifier has common false negatives for send tools such as `add_issue_comment`, so it can still produce the blind-send failure the carve-out exists to prevent.

src/harnesses/toolNote.ts
robertclawson: Fixed in 9936a41 — you're right, and the miss was structural rather than a gap in the list: the classifier only knew send-class verbs, while the tools that actually publish text mostly don't use one.

- `add`, `write`, `append` and `update` join `post`/`share`/`create` as QUALIFIED verbs — a send only when a transmit object rides along. Standalone they're far too broad (`write_file`, `add_label`, `update_config`, `append_to_file` all stay gated, and they're in the READS fixture now).
- The object may now sit on **either side** of the verb. `add_issue_comment` puts it after, `discussion_comment_write` puts it before, and the old after-the-verb-only scan missed every trailing-verb name.
- Tracker nouns (`issue`, `discussion`, `review`, `thread`, plus plurals) are transmit objects: publishing prose to a tracker puts it in front of third parties exactly like mail does, so `create_issue` is a send too.
- The read-verb veto is untouched, so `get_issue_comment` and `list_discussion_comments` remain ordinary boundaries — widening the vocabulary must not hand the read side an exemption.

All three names you listed are covered, as unit fixtures and as end-to-end gate regressions (`a tracker comment publishes too, even though it never says 'send'` asserts the draft survives in both the rendered stream and `finalText`; `reading a comment is not publishing one` asserts the mirror case still drops).

Mutation-checked four ways — dropping the new verbs, reverting to the after-the-verb-only scan, letting the new verbs stand unqualified, and removing the read veto each turn tests red. Full suite 5023/0 across 306 files, typecheck clean.

lenaparkhodges (approved):
Verified the transmit-vocabulary delta at 9936a41. All three of Kai's representative names (`add_issue_comment`, `add_pull_request_review_comment`, `discussion_comment_write`) now classify as sends; the object-matches-on-either-side rule handles trailing-verb names; `write_file`/`add_label`/`update_config`/`create_branch` correctly stay gated; the earlier read/search vetoes (`gmail_read_email`, `slack_list_messages`, `postgres_query`) still hold. New unit + end-to-end stream regressions pin all of it. Local mutation check (dropping `add` from the qualified list) flips two tests red, as expected. Targeted suites 54/54 pass, typecheck clean; the 9 failures I see in the full local run are pre-existing cross-process lock tests that fail identically at c9a55d2 — environment drift, not this delta. My earlier blockers are resolved; approving.

src/harnesses/toolNote.ts
```diff
@@ -249,6 +257,156 @@ function normaliseName(name: string): string {
...
+  "compose",
+  "forward"
+] as const;
+
+/**
+ * Verbs that only mean "send" when they carry an object: `post_message` sends,
+ * `get_post` and `postgres_query` do not. Never sufficient alone.
+ *
+ * `add`/`write`/`append`/`update` are here because the most common transmit
+ * tools in practice do not say "send" at all: the official GitHub MCP surface
```
kaieriksen, line 286: Verified at 9936a41: add_issue_comment / add_pull_request_review_comment / discussion_comment_write all return true, while write_file, update_config, gmail_read_email, slack_list_messages and postgres_query stay false. Mutation check: dropping add/write from this list fails the new tests.

kaieriksen (approved):
Re-reviewed at exact head 9936a41. The add/write false-negative blocker is fixed: qualified verbs now require a transmit object on either side of the verb, and I verified behavior directly - add_issue_comment, add_pull_request_review_comment and discussion_comment_write classify as sends while write_file, update_config and all read/search names stay gated. Removing add/write from the verb list turns the new tests red, so the vocabulary is mutation-covered. 54 focused tests pass, tsc clean, all CI green, all threads resolved.

PR #588, opened 2026-09-19
counters: one persisted store for narration drops and lost deliveries (#585)

### LINKED ISSUE #585: Persisted counter for suppressed narration lines and exhausted retry ladders

lenaparkhodges:
Closes #585.

## What

One small persisted counter store (`<XDG_STATE_HOME>/phantombot/counters.json` — flat `key → count`, `src/lib/persistedCounters.ts`), written by two feeds:

**Feed 1 — narration drops (#580/#587)**, in `narrationStreamGate.ts`:
- `narration.drop.<expected>.<actual>` — every wrong-language line dropped at a tool boundary, keyed per language pair so a systematic false positive on one profile shows up as an outlier, not as a user complaint. Batched per boundary.
- `narration.idle-release.<expected>` — held text flushed by the 500 ms idle race instead of being resolved by a boundary. A high rate means the window is too tight and leaks are escaping through it.
- `narration.shape-rejection.<expected>` — held text emitted because the shape test declined to call it narration (the deliberate under-enforcement). If this number is large the shape guard is mis-tuned.
- `narration.sending-keep.<expected>` — held text kept by the #587 sending-tool carve-out. Counts the exact path reviewers flagged as the partial-#580-reopen risk.

**Feed 2 — exhausted ladders (#542/#584)**, in `phantomchat/transport.ts`:
- `delivery.lost.<label>` — a wrap lost after the whole 8s/20s/45s ladder with zero readable relays and no P2P ack. Relay quarantine was already fed via `verifyStored`; only the count was missing.

## Why the store, not re-scoring

The #580 evidence came from re-scoring ~14,800 stored turn pairs offline. Once the gate is live, a leak is suppressed *before* it reaches the turns table, so that measurement path goes blind exactly when we start relying on it. The counter is the only post-gate signal — and the natural acceptance criterion for the enforcement gate specced in #580.

## Properties

- Best-effort by contract, like the audit log: a bump never throws, never slows a turn; failures degrade to a warn log.
- Bumps are serialised through an internal promise chain (concurrent increments cannot lose updates) and written temp-file + rename (a crash mid-write cannot corrupt the store).
- A corrupt/unparsable store resets to empty rather than wedging every future bump.
- One batched bump per gate boundary, not one per line.

## Not in scope (per the issue)

- Group retry (`wrapGroupMessage`, per-member) — tracked under #542; `delivery.lost.*` measures DM text + voice only until that lands.
- Expected-side ambiguity precision work beyond what landed in #583.

## Tests

- New `tests/lib-persistedCounters.test.ts` (8): accumulation, distinct keys, zero/negative ignored, corrupt-store reset, junk dropped on read, path shape, concurrent-bump race.
- `tests/narration-stream-gate.test.ts` +6: drop counted per pair, batched boundary bump, sending-keep, shape rejection, clean turn bumps nothing, idle release counted.
- `tests/channels-phantomchat-outboundRetry.test.ts` +1: give-up lands as `delivery.lost.dm`.

Full suite: 5038 pass / 0 fail. `tsc --noEmit` clean. Mutation check: stripping the drop-path bump flips 2 gate tests red.


src/channels/phantomchat/transport.ts
```diff
@@ -815,6 +816,10 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
       originalEventId,
       attempts: delays.length,
     });
+    // Feed 2 of #585: the give-up count is the only queryable signal that a
+    // reply was genuinely lost — relay quarantine is already fed via
+    // verifyStored, so only the number was missing.
+    bumpCounters({ [`delivery.lost.${label}`]: 1 });
```
kaieriksen, line 822: Note for the rebase: this PR branched before #589 (group member wrap retry) merged, so it now conflicts with main in this file and its test. After rebasing, please check whether #589 added its own give-up point for the per-member ladder — `delivery.lost.*` should then cover the group labels too, and the "not in scope" paragraph in the description is stale once #589 is in.

src/lib/narrationStreamGate.ts
```diff
@@ -250,11 +251,21 @@ export function createNarrationStreamGate(
...
     const wasNarration = isNarrationBlock(sinceBoundary);
     sinceBoundary = "";
-    if (!wasNarration) return block;
+    if (!wasNarration) {
+      // Shape rejection (#585): wrong-language text the shape test declined
+      // to call narration, so it was emitted. Deliberate under-enforcement —
+      // counted, not punished, so a mistuned shape guard shows up as a
+      // number instead of as a user complaint.
+      if (block.trim()) {
+        bumpCounters({ [`narration.shape-rejection.${expected.code}`]: 1 });
```
kaieriksen, line 262: Nit on the metric name: this fires purely on the shape test declining — it does not verify the block was actually in the wrong language. Since holds only originate from language-suspect text the count is approximately right, but the code comment ("wrong-language text the shape test declined") slightly overstates what is measured. Either soften the comment or accept the approximation; not blocking.

src/lib/persistedCounters.ts
```diff
@@ -0,0 +1,98 @@
...
+ *                                        tool transmits content (#587 carve-out)
+ *   delivery.lost.<label>                wraps lost after the whole retry ladder
+ *
+ * Best-effort by contract, like the audit log: a failed read or write is
+ * logged and swallowed — counting must never break or slow a turn. Bumps
+ * are serialised through an internal promise chain so concurrent
+ * increments cannot lose each other's updates, and the write itself is
+ * temp-file + rename so a crash mid-write cannot corrupt the store.
+ */
+
```
kaieriksen, line 42: The promise chain serialises bumps within one process; two phantombot processes sharing this XDG state path would race last-writer-wins through the rename. Correct under the single-process assumption (personas have separate accounts/homes), but worth one line in the docstring so a future embedding does not silently lose counts.

kaieriksen (dismissed):
Reviewed the full diff at head 369402c. Targeted suites pass locally (55/55 across persistedCounters, narration-stream-gate, outboundRetry), `tsc --noEmit` clean, all CI green including CodeQL. The store design is right: serialised read-modify-write, temp+rename, corrupt-store reset, and the batched-per-boundary bump keeps the hot path clean. Approving with three non-blocking notes — one of which (the #589 rebase) gates the merge itself.

src/channels/phantomchat/transport.ts
lenaparkhodges: Rebased onto main at dd88c12 (import-block conflict only). Confirmed #589 added no separate give-up point: the per-member rewraps route through the same `publishWrap` ladder with `label: "group"`, so `delivery.lost.group` is covered by this one bump. The stale "not in scope" group bullet in the description is replaced with that note.

src/lib/narrationStreamGate.ts
lenaparkhodges: Fair — the counter measures the shape test declining, not the language itself. Softened the comment to say holds only originate from language-suspect text, so the count approximates wrong-language releases. Kept the key name: renaming would churn queries for zero information gain.

src/lib/persistedCounters.ts
lenaparkhodges: Added the docstring line: serialisation is per-process; two phantombot processes sharing this XDG state path would race last-writer-wins through the rename — fine under the single-process assumption (separate accounts/homes), a silent count-loser in an embedded multi-process setup.

kaieriksen (approved):
Re-reviewed at dd88c12. Rebase onto #589 verified: the group path routes through publishWrap with label "group", so the single delivery.lost bump covers the per-member give-up path — no separate counter needed. All three inline threads addressed accurately (shape-rejection comment softened, per-process serialisation note added, #589 coverage note). Targeted suites 58/58 and typecheck clean locally; CI + CodeQL green. CLEAN/MERGEABLE.

PR #589, opened 2026-09-19
phantomchat: retry group member wraps the relays never stored (#542)

### LINKED ISSUE #542: phantomchat: outbound replies have no delivery receipt and no retry (one-way double tick)

kaieriksen:
Closes #542 (item 1 — per-member group retry).

## Problem

#584 shipped the DM half of #542: a DM wrap no relay stored is re-sent on the PWA's 8/20/45s ladder. `sendGroupMessage` still published each member gift-wrap fire-and-warn — a group wrap no relay stored was silently lost, with the PWA's 15s catch-up poll (reading the SAME relay set that just failed) as the only backstop. This is exactly what Robert reopened the issue for.

## Fix

Two small pieces, no new machinery:

1. **`wrapGroupMessage` now returns one `rewrap` thunk per other member** — a re-gift-wrap of that member's UNCHANGED seal. `createGiftWrap` mints a fresh ephemeral key and real timestamp per call, so each attempt is a new event id (relays don't dedup it away) around the same rumor id (the recipient dedups a copy it already has) — the exact semantics `rewrapV2` provides on the DM path. The self-wrap has no thunk: it is multi-device recovery, not delivery, so a lost self-copy never justifies ladder traffic.

2. **`sendGroupMessage` passes each member's thunk to `publishWrap`**, so the existing ladder (#584 infrastructure, untouched) retries a member whose wrap is readable from zero relays and not P2P-acked — and only that member. A partial failure (A's wrap stored, B's dropped) retries B alone.

## What this deliberately does not do

- No changes to `publishWrap`, `retryOutbound`, or the ladder itself — the group path just opts in via the existing `rewrap` option.
- No per-member ack/quarantine scoring changes; read-back verification already runs per wrap.
- Item 2 of #542 (PWA-side ack so the double-tick becomes real) is a different repo; item 3 (exhausted-ladder counter) is already tracked as #585.

## Verification

- New tests in `channels-phantomchat-outboundRetry.test.ts`:
  - `wrapGroupMessage` exposes exactly one rewrap thunk per other member (none for the self-wrap); each thunk produces a fresh event id / ephemeral key / content around the same rumor id, verifiable by the member.
  - Partial failure: relays store member A's wrap and drop B's → B is retried exactly 3× with the same rumor id, A and the self-wrap are never re-sent.
  - The self-wrap never enters the ladder even when its read-back comes back empty.
- Mutation-checked: removing the `rewrap` option in `sendGroupMessage` fails 2 tests; returning empty `rewraps` fails 3.
- Full suite: 5,026 pass / 0 fail. `tsc --noEmit` clean.


src/channels/phantomchat/transport.ts
```diff
@@ -1354,15 +1354,26 @@ export class SimplePoolPhantomchatTransport implements PhantomchatTransport {
...
-    for (const wrap of wraps) {
-      await this.publishWrap(wrap as unknown as NTNostrEvent);
+    // Per-member delivery retry (issue #542): each member wrap carries its own
+    // rewrap thunk — a re-gift-wrap of THAT member's seal — so publishWrap's
+    // 8/20/45s ladder fires only for a member whose wrap is readable from zero
+    // relays and only re-sends to that member. A partial failure (A's wrap
+    // stored, B's dropped) retries B alone. The self-wrap stays fire-and-warn:
+    // it is multi-device recovery, not delivery, so losing it loses nothing
+    // the recipient was waiting for.
+    for (let i = 0; i < rewraps.length; i++) {
```
lenaparkhodges, line 1370: Non-blocking: the wraps↔rewraps correspondence is enforced only by convention — rewraps[i] must describe wraps[i], and wraps[rewraps.length] must be the self-wrap. Both hold because of how wrapGroupMessage builds its arrays today, but a future reorder would silently gift-wrap member B's seal to member A — A can't decrypt it and the message vanishes without an error. Consider returning paired { wrap, rewrap } tuples for the members plus a separate selfWrap, so the invariant is structural rather than positional.

lenaparkhodges (approved):
APPROVED at 1cc1b64. Verified locally: 13/13 retry-suite tests pass, tsc clean, my own mutation (dropping the rewrap option from the member publish) flipped 2 tests red. The semantics are right — same-seal re-gift-wrap gives fresh event ids for relay dedup around the stable rumor id for recipient dedup, matching rewrapV2 on the DM path; the self-wrap correctly stays fire-and-warn; and since publishWrap returns on first accept with the ladder detached, the per-member loop does not serialize retry ladders across members. One non-blocking nit inline on the wraps/rewraps index coupling.

PR #591, opened 2026-09-19
Fail over transparently when a harness runs out of quota

### LINKED ISSUE #590: Harness quota exhaustion is neither transparent nor complete: stderr never classified, durable facts and the threat judge don't fail over
## What happened

On 2026-09-19 one host's primary harness (codex) was out of subscription quota for roughly four hours. Chat turns failed over and looked perfectly healthy. Everything else about that window was invisible or wrong:

- The journal said `codex exited with code 1`, and nothing else. Ever. The provider's own explanation — `ERROR: You've hit your usage limit. Upgrade to Pro … or try again at 8:58 PM.` — was on the child's stderr, which the classifier never looks at.
- The failure therefore classified as `other`, so the user-visible fallback tag read "unavailable" rather than "rate limit".
- `parseRetryAfterMs` only understands durations. Codex gave a wall clock, so there was no hint, so the jittered ladder took over — capped at one hour, re-probing a quota we already knew was dead, all afternoon.
- **Durable-facts extraction stopped entirely.** It calls `harnesses[0]` and nothing else, so every batch hit the dead primary, logged `extraction call failed; will retry batch`, and was dropped. Memory quietly stopped being written for four hours.
- **The threat judge has the same single-harness call — and it fails OPEN.** A quota exhaustion on the primary silently disables the screener that stands in front of every untrusted input. This is the worst part of the issue and it is not specific to one host.
- Once a fallback did serve, every single reply for the whole window carried `— answered by pi fallback (…)`.

## What it should do

Fall back transparently, like nothing happened, for users on small subscription plans who will hit these limits routinely. Specifically:

1. Classify on the harness's stderr, not just on the exit line.
2. Understand a wall-clock deadline, not only a duration.
3. Don't re-probe a window we've been told is closed — including across a restart (`phantombot update` restarts the service).
4. Durable facts and the threat judge run the chain, like turns do.
5. Announce the switch once, not on every reply for four hours.

## Fragility note

The text parsing is the brittle part — Anthropic and OpenAI can reword these messages without notice. So the design rule is that **parsing only ever shortens a wait, never gates failover**: failover is driven by the failure itself and cannot break on a wording change; a parse that stops matching costs a dumber cooldown, not an outage. Claude is already structured (`error: "rate_limit"` in its envelope) and needs no scraping; codex is the plain-prose case.

robertclawson:
Closes #590.

A four-hour codex subscription window on one host was invisible in the journal, mis-labelled to the user, re-probed hourly all afternoon, and took durable-facts extraction and the threat judge down with it — while chat turns failed over and looked perfectly healthy. The goal here is the one in the issue: a user on a small plan burns through it and the fallback happens transparently, with no drama.

### 1. Classify on the stderr tail, not just the exit line

A CLI-subprocess harness dies with `codex exited with code 1`. That string names no cause, and it is all `classifyFailure` ever saw — the provider's actual explanation is on the child's stderr, which we capture and then never look at. `classifyFailure` now takes the tail, `noteFailure`/`noteExhausted` pass it, and the rate-limit markers gain the subscription-exhaustion wordings (`usage limit`, `upgrade to pro`, `plan limit`) which are neither an HTTP 429 nor the word "quota".

The mis-labelling followed from the same gap: the only thing that survives into the `done` chunk is the exit line, so re-deriving the class in `fallbackTag.ts` rendered every subprocess death — rate limits included — as "unavailable". The orchestrator now stamps the cause it classified while it still had the stderr in hand, and the tag prefers it.

### 2. Wall-clock deadlines

`parseRetryAfterMs` understands durations. Codex answers with a time. New `parseRetryDeadlineMs` reads `try again at 8:58 PM`, a 24h clock, and ISO timestamps; a past time yields no hint rather than a negative one, and a bare clock is the next occurrence in local time.

Deliberately narrow: the verb must be a retry/reset verb, the preposition must be `at`, and **the caller only consults it for failures that already classified as a rate limit** — so a stray time string in a stack trace cannot bench a healthy harness for hours.

An explicit provider window may now exceed the ladder's 1h cap, bounded by its own `MAX_EXPLICIT_COOLDOWN_MS` (6h). A quota window is measured in hours; clamping a known four-hour close to one hour means three more guaranteed-failed probes for nothing. The ladder's cap is untouched, because the ladder is a guess and this is an instruction.

### 3. Persisted cooldown

`phantombot update` restarts the service, so an in-memory-only store re-probes a closed window every time the box comes back. Windows now round-trip through their own file (not `state.json` — that one is audit-logged, guarded against live writes, and bricks startup if torn). Only still-open windows are re-adopted; an unreadable or unwritable file degrades to exactly today's behaviour, and the sink can never throw into a turn.

### 4. Durable facts and the threat judge run the chain

Both called `harnesses[0]` and stopped there. During the window:

- extraction logged `will retry batch` and dropped it, once per turn, for four hours. Memory just stopped.
- **the threat judge fails OPEN.** A quota exhaustion on the primary silently switched off the screener that stands in front of every untrusted input. That is the failure mode the whole perimeter exists to prevent, and it isn't specific to one host.

New `completeOverChain()` gives both a chain walk with the same cooldown semantics the orchestrator uses — including the all-cooled escape hatch, because a screener that refuses to run *is* the fail-open bug. An abort is never mistaken for ill health (harnesses surface it as an ordinary error chunk, so the message alone can't tell them apart), and an empty completion falls through without cooling, per #499.

`pickJudgeHarness` is deleted; its "never assumes claude" test moved onto `makeChainJudgeComplete`, which is the thing that has to hold that property now.

### 5. Tag the switch, not the state

A quota window is hours long, so tagging every fallback-served reply appends `— answered by pi fallback (…)` to every message for an afternoon, and repeats it in front of third parties in group chats. The tag now fires on the turn the serving harness actually changes, tracked per conversation. Moving back to the primary and failing over again later is a fresh switch and tags again; a caller with no conversation key (one-shot `ask`) always tags, since it has no previous reply for a tag to be redundant against.

### On fragility — the part worth arguing with

Vendors can reword these strings without notice, so the rule throughout is that **parsing only ever shortens a wait; it never gates failover.** Failover is driven by the failure itself and cannot break on a wording change — that is why turns kept working during the incident. If a pattern stops matching, the cost is a dumber cooldown, not an outage. Claude is already structured here (`error: "rate_limit"` in its envelope) and needs no scraping; codex is the plain-prose case.

What this does *not* include, deliberately: a drift canary counting classifications by cause per harness (a rate limit misclassified as `other` is still invisible today — that belongs with #585's counter), pre-emptive quota budgeting, and mid-stream failover.

### Verification

- Full suite 5079 pass / 0 fail, typecheck clean.
- Every new behaviour mutation-checked by deleting the fix: stderr-blind classification (3 red), judge back on `chain[0]` (2 red), extraction back on `chain[0]` (1 red), tag-on-every-reply (1 red), explicit-cap reverted + persistence flush removed (3 red), deadline parsed without the rate-limit gate (1 red), tag ignoring the stamped cause (1 red), abort no longer stopping the chain (1 red).
- Real provider wording used as fixtures, including the exact codex line from the incident.


andrewagrahamhodges (approved):
## Review at exact head f8db15e2ebe8e0babada8d715eaf6fa84dd756f2

**Environment:** isolated clone at f8db15e2, Arch Linux (kw-omarchy), bun 1.4.2
**Focused suites:** lib-cooldown, lib-cooldownPersist, lib-harnessRunner, lib-threatJudge, orchestrator-durableFacts, orchestrator-fallback: **188/188**; channels-fallbackTag, lib-chainComplete, lib-harnessAlert, cli-run: **122/122**
**Full suite:** 5,073 pass / 8 skip / **9 fail** — the 9 failures (tui-chat-session-brain, lib-persona-snapshot-brain, cli-tick, cli-persona-new) reproduce **identically on main** (efc3017) in this environment: they write config relative to cwd and are environment-sensitive, not PR-caused. None touch files this PR changes.
**Typecheck:** clean (`bun tsc --noEmit`)
**Build:** `bun run build` compiles; `dist/phantombot --version` runs
**CI:** green — test, test-windows, typecheck, CodeQL all success at f8db15e2

### Verification performed

**Reproduced the #590 incident on main vs branch.** Simulated codex dying with `codex exited with code 1` while stderr carries `You've hit your usage limit … Try again at 8:58 PM`:
- main @ efc3017: classifies `other` — the mislabel from the incident, reproduced
- branch @ f8db15e2: classifies `rate_limit` with the stderr tail — fixed
- control: a stack trace containing a stray time (`run at 9:15 PM`) classifies `other` on both — no overreach

**Deadline parser probed directly** (`parseRetryDeadlineMs`): 12h future clock → correct duration; past bare clock → next occurrence (+24h rollover); past ISO → undefined (no negative hint); 24h clock and ISO with timezone → parsed; a time with no retry verb → undefined; `try again in 3:00 PM` (wrong preposition) → undefined. Matches the deliberately-narrow scope in the PR description, and the call site (harnessRunner.ts:971-974) only consults it for failures already classified as rate limits — confirmed in code.

**Chain walk read in full** (`completeOverChain`): cooldown semantics match the orchestrator's — cooled skipped at snapshot, all-cooled tries anyway (the fail-open fix), throw cools, success clears, empty falls through WITHOUT cooling (#499), aborts never cooled. Both threatJudge.ts:502 and durableFacts.ts:403 route through it.

**Mutation testing — 4 mutations, all flipped tests red:**
1. `classifyFailure` ignores stderr tail → 3 failures (lib-harnessAlert)
2. past ISO deadline returns negative → 1 failure
3. `completeOverChain` cools on empty result (#499 violation) → 1 failure (lib-chainComplete)
4. explicit cap 6h → 1h → 2 failures (lib-cooldown/lib-cooldownPersist)

The tests genuinely pin the new behaviour.

### Findings — both non-blocking

1. **Non-blocking:** `CooldownStore.clear()` sets `persistence = undefined`. The comment says production has no caller, so this is fine today — but if a future admin "reset cooldowns" path ever calls it at runtime, persistence silently disables until restart. Consider having `clear()` keep the sink and flush an empty snapshot instead, or assert test-only.
2. **Non-blocking:** the persistence sink fires `writeFileAtomic` without awaiting or serialising (cooldownPersist.ts:88-97). Two saves in quick succession can rename out of order and persist an older snapshot. Worst case is a stale window re-adopted after restart → one extra probe of a cooled harness, then self-heal. Harmless; worth a comment line if you agree, nothing more.

### Downstream

Unblocks transparent small-plan operation: the judge and durable facts no longer fail open/silently drop during a quota window. No dependent PRs blocked on this one that I can see (the omarchy #12251 chain depends on #575, not this).

**Verdict: APPROVED at exact head f8db15e2ebe8e0babada8d715eaf6fa84dd756f2.** This closes a real perimeter hole (judge failing open) plus the silent memory loss, with honest tests. Merge when ready.

PR #592, opened 2026-09-19
harness: log why a fall-through happened, and fix two cooldown-persistence defects

robertclawson:
Three follow-ups to #591, all found while diagnosing #591 itself in production. Two were Omar's non-blocking review findings; the first is the one that cost me an hour and a wrong call to the principal.

### 1. The fall-through path is invisible (the expensive one)

`fallback.ts` logged `error` + `httpStatus` on a recoverable fall-through. A CLI harness dies with `codex exited with code 1` — a string that names no cause and carries no status. So a **quota exhaustion and a segfault logged byte-identical lines**.

Consequence, observed: after v1.1.389 shipped I read a journal spanning a restart, saw those lines, and reported that #591 was not working in production. It was working. There was no field in the log that could have told me either way, and confirming it meant running the harness by hand on the host.

The cause is already classified one block down for `firstFailure`. This classifies it *before* the log and reuses the value, adding `cause`, `retryAfterMs`, the `cooldownUntilMs` the failure actually produced, and `nextHarnessId`. `retryAfterMs` is what distinguishes a provider-supplied deadline (#591's path) from the generic jittered ladder — i.e. it answers "did the fix fire" directly.

### 2. `CooldownStore.clear()` dropped the persistence sink

It set `persistence = undefined`. Production has no caller today, so nothing is broken in the field — but a `/status`-style "reset cooldowns" admin path is exactly what this method is for, and it would have inherited a one-way door: the store keeps working in memory, silently stops writing to disk, and the next restart re-adopts the windows that were just cleared. Now it keeps the sink and flushes an empty snapshot, so clear() means the same thing in memory and on disk.

### 3. Persistence writes were unserialised

`writeFileAtomic` is write-temp-then-rename. The sink fired it unawaited, so two saves in flight are two temp files racing to rename over one path — and the rename order is not the call order. A chain fall-through issues one `markFailure` per harness back-to-back, so the burst is the normal case, not an exotic one. Worst outcome is one extra probe of a cooled harness after a restart, then self-heal; silent and wrong nonetheless.

Fixed by chaining writes. Fire-and-forget is kept (the store's `save` is sync and on the turn path — never block a turn on a disk write), but a save arriving while another is in flight now **replaces** any queued-but-unwritten snapshot rather than queueing behind it, since only the newest snapshot has value. A burst of N saves costs at most two writes and the last one always wins. `settled()` is added to the interface as an optional test/shutdown hook.

### Verification

Full suite **5096 pass / 9 fail**, typecheck clean. All 9 failures are pre-existing process-spawn lock tests (`workspaceLock`, `runLock`, `MemoryStore` concurrent open) that fail identically on a stashed working tree — unrelated to this change.

Mutation-checked, each fix separately:

| Mutation | Result |
|---|---|
| Drop the new log fields | 3 fail |
| Restore `persistence = undefined` in `clear()` | 2 fail |
| Un-serialise the writes (concurrent `writeFileAtomic`) | 1–2 fail, every run |

The two log tests are deliberately a **pair**: a quota and a panic with the *same* `error` string, asserting only `cause` separates them. The cooldown-window test asserts the logged `cooldownUntilMs` equals what the store really holds, so the journal cannot claim a bench the process is not honouring.

### Not in scope

Counting classifications by cause per harness — the drift canary for a vendor rewording their message. That belongs with the counter work in #585, not here.


src/orchestrator/fallback.ts
```diff
@@ -368,13 +368,21 @@ export async function* runWithFallback(
...
-                httpStatus: chunk.httpStatus,
-              },
+            // Classify BEFORE logging, not after. The log line below is the
+            // only externally visible record that this path ran, and until
+            // #592 it printed `error` + `httpStatus` only — so a quota
+            // exhaustion ("codex exited with code 1", no status) and a
+            // segfault ("codex exited with code 1", no status) produced
+            // BYTE-IDENTICAL journal lines. There was no way, from outside
+            // the process, to tell whether the rate-limit handling added in
+            // #591 had fired at all; diagnosing it meant re-running the
```
lenaparkhodges, line 378: Non-blocking nit: this sentence describes the pre-change layout — after this edit nothing is "computed one block down" anymore; `firstFailure` now reuses the `cause` computed *above*. Suggest rewording to e.g. "the same classification `firstFailure` needs, computed once and reused" so the comment matches the code it sits in.

lenaparkhodges (dismissed):
Verified at 31e5ffc: targeted suites 67/67 + tsc clean locally, all three mutations independently go red (drop log fields → 3 fail; restore `persistence = undefined` → 2 fail; unserialise the writes → 5 fail incl. the burst test). The serialisation design is sound — I checked the drain loop for a lost-wake-up between the final `pending` check and `draining = false`: there's no await on that path, so a save arriving there re-enters via the next `save()`, and a save mid-write is picked up by the live loop. No window. The two log tests being a deliberate pair (same `error` string, only `cause` separates them) is exactly the right shape for this regression.

One CI note: the `test` workflow failed on this head, but it's `override decay (sliding idle TTL)` in coderSwap — a 5s timing test in a subsystem this PR doesn't touch (disjoint files, passes locally, and it's the classic load-flake shape). test-windows and typecheck are green. Not a blocker, but worth a re-run for a clean merge.

robertclawson:
Thanks Lena — both points handled.

**The nit is fixed in `30aef43`** (comment-only). The sentence now reads "This is the same classification `firstFailure` needs below — computed once here and reused", which is what the code does after this PR; the old wording described the layout it replaced.

**On the CI failure:** agreed, it's not this PR. The only failing test is `override decay (sliding idle TTL) > stays live indefinitely while actively used — no absolute cap`, which times out at exactly 5000.40ms in coderSwap — a subsystem with no file overlap with this diff. `typecheck`, `test-windows` and both CodeQL jobs were green at 31e5ffc. Re-running the old run wasn't possible from the API, so this push gives it a fresh run anyway.

Re-requesting you at the new head since the push dismissed your approval — sorry for the round trip for a comment, but it was the cheapest way to get the re-run too.

lenaparkhodges (approved):
Re-approved at 30aef43. Delta vs my 31e5ffc approval is the comment-only reword at fallback.ts:376 — now describes the post-PR layout correctly. All CI green at this head (test re-run passed). Everything else unchanged from my earlier review: all three behaviors verified, mutations red.

src/orchestrator/fallback.ts
robertclawson: Reworded in 30aef43 — the comment now describes the post-change layout (classification computed once above and reused by `firstFailure`). Comment-only change; resolving.

PR #593, opened 2026-09-19
lifecycle: notify after the restart, not (only) before

lenaparkhodges:
Closes the restart-notification race discussed with Andrew (heads-up lost when a persona restarts; no notification after a restart either).

## The race

`/restart` (and `/update`) warned the other personas BEFORE shutting down, but the `.pending-lifecycle.json` marker was written only **after** every send completed. The PhantomChat one-shot send (pool connect + NIP-42 handshake + NIP-17 gift-wrap + flush across relays) runs seconds per recipient — so:

- a slow or killed announce meant the marker was never written → **no "back online" for anyone**,
- Telegram (instant HTTP) made it out; PhantomChat copies missed the window every time,
- the origin persona got "restarting…" and then **silence** — nothing ever confirmed the process came back,
- dual-channel personas got **no PhantomChat lifecycle notice at all**: `run.ts` only supplied phantomchat accounts for personas *without* a Telegram listener, so #586's both-channels planner fan-out never saw their phantomchat account.

## The fix — same pattern as the update notifier

The update flow already solves this exact problem with a pre-restart marker + post-restart confirmation; this applies the same shape to the lifecycle broadcast.

- **Marker before sends.** `announceImpendingRestart` writes `.pending-lifecycle.json` FIRST, then attempts the heads-up. The post-restart fan-out is now the authoritative notification path — a heads-up that never makes it out costs nothing.
- **Bounded heads-up.** The announce is raced against a 4s bound so a slow channel can never delay the restart the user asked for. Telegram almost always makes it; whatever is still in flight dies with the process.
- **Origin persona confirmed.** For `/restart`, the origin is added to the marker's notify list: post-restart it gets "✅ Back online on vX (after /restart)" in the chat that typed the command — mirroring how the pending-update marker confirms an `/update`. For `/update` the origin stays off (the update marker already confirms there; a second line is noise).
- **Dual-channel fan-out actually reaches PhantomChat.** `run.ts` now supplies phantomchat lifecycle accounts for every runnable phantomchat persona with an allowlist; the #586 planner dedups by (persona, hex), so no double-DM.
- **Total send failure retries.** `notifyLifecycleBackIfPending` cleared the marker even when *every* send failed — one boot-time relay hiccup permanently lost the notification. Now a total failure keeps the marker so the next restart retries (bounded by the existing 1h staleness window); partial success still clears.

## Tests

- marker written before sends + hanging send cannot delay the restart (the race regression test)
- single-persona `/restart` still records the origin for its post-restart confirmation
- origin persona receives back-online when the marker names it
- total send failure keeps the marker; partial failure clears

Full suite 5116 pass / 0 fail, `tsc` clean.

Mutation check (updated after Robbie's review — the original claim didn't reproduce; the ordering wasn't pinned by any test): the hanging-send test now captures the marker from inside the in-flight send, so reverting the marker-before-send ordering (write after the `Promise.race`, recipients check untouched) turns the hanging-send test and the single-persona `/restart` test red. Note the in-flight observation is a capture-and-assert-later, not an in-flight `expect` — `sendLifecycleBroadcast` swallows send throws, so an in-flight assertion dies silently and leaves the mutation green.


robertclawson (changes requested):
Requesting changes on one real gap — the code itself looks right.

## What I verified

- **Planner dedup claim holds**: `planLifecycleBroadcast` keys phantomchat dedup on `(persona, hex)` via `seenPc`, so supplying phantomchat accounts for dual-channel personas (run.ts change) cannot double-DM. Verified in source, not just the description.
- **`sendLifecycleBroadcast` never rejects** — every send and the transport construction are internally try/catch'd — so the `Promise.race` loser can't surface as an unhandled rejection. Good.
- **Total-failure retry semantics are sound**: `sent+failed` always equals the send units for any non-empty recipient list, so `sent === 0 && failed > 0` really is "total failure". Mutation-checked: reverting to always-clear turns the new retry test red. ✅
- Old-marker compat (origin absent → old lossy behavior) is handled and documented.
- Touched test files: 114 pass.

## Blocking: the marker-before-send ordering is not pinned by the tests, and the mutation-check claim doesn't reproduce

The PR body says *"Mutation check: reverting the marker-before-send ordering turns 3 tests red."* I reverted the ordering faithfully — `writePendingLifecycle` moved to AFTER the `Promise.race` (race intact) — and **all 80 tests in `channels-commands.test.ts` still pass**. The hanging-send test simply eats the 4s race bound, the marker is written late but correctly, and every assertion holds.

The ordering IS still load-bearing: with send-first ordering, a SIGTERM landing inside the ≤4s announce window loses the marker entirely — exactly the original bug for the killed-announce case. But no test observes the marker *while the send is in flight*, so a future refactor that reorders is invisible to CI.

One-line fix in the hanging-send test — assert the marker exists from inside the in-flight send, which fails immediately under the old ordering and adds no wall-clock cost:

```ts
createTelegramTransport: () =>
  ({
    async sendMessage() {
      // Pins the ordering: the marker must already exist while this send is
      // still in flight — a send-first ordering loses the back-online notify
      // if the process is killed mid-announce.
      expect(await readFile(pendingLifecyclePath, "utf8")).toBeTruthy();
      await new Promise(() => {});
    },
  }) as any,
```

Please also correct the mutation-check claim in the PR description once this lands.

Everything else — bounded heads-up, origin confirmation for `/restart` (and correctly NOT for `/update`), retry-on-total-failure — LGTM.

lenaparkhodges:
Fixed at 212c0a3 — with one correction to the suggested patch: the literal in-flight `expect` **leaves the mutation green**. `sendLifecycleBroadcast` swallows send throws (the property you verified two paragraphs up), so the assertion throws ENOENT inside `sendMessage`, the broadcast catches it, the race resolves early, the marker gets written, and every surviving assertion holds — exactly the green-mutation run you described. Captured mid-flight, asserted after the race instead:

```ts
async sendMessage() {
  try { markerSeenBySend = await readFile(pendingLifecyclePath, "utf8"); }
  catch { /* no marker yet = send-first ordering = caught below */ }
  await new Promise(() => {});
},
// ...after afterSend():
expect(markerSeenBySend).toBeTruthy();
```

Mutation re-run with your faithful reordering (write after the `Promise.race`, recipients check untouched): hanging-send test red (the mid-flight capture stays empty) + single-persona `/restart` red (the write also lands behind the recipients check in that ordering). PR body claim corrected — thanks for catching that the original didn't reproduce.

robertclawson (approved):
Re-reviewed at 212c0a3 — blocking finding resolved, and the fix is better than my literal suggestion.

The mid-flight capture is the right pattern: an in-flight `expect` would die inside `sendLifecycleBroadcast`'s swallow and leave the mutation green, so capture-inside/assert-after is what actually pins the ordering. Verified independently:

- Both touched suites pass (114/0) on a fresh checkout of the head.
- Mutation-checked myself: moved `writePendingLifecycle` after the `Promise.race` (recipients check untouched) — the hanging-send test and the single-persona `/restart` test both go red. The PR body's updated claim is accurate.
- CI 6/6 green at this head; marker semantics (origin on the notify list for /restart only, total-failure marker retention bounded by the 1h staleness window) all consistent across commands.ts, run.ts and lifecycleBroadcast.ts.

One nit, non-blocking: `HEADS_UP_TIMEOUT_MS = 4_000` is declared between the docstring and its function — fine, just a half-second judgment call worth remembering if PhantomChat one-shot connects ever get slower on a cold relay set. The race resolving early costs nothing (the broadcast dies with the process either way); the bound existing at all is the fix.

PR #594, opened 2026-09-19
Keep harness failover invisible to users

kaieriksen:
Restores the documented invariant that successful harness failover is invisible on chat, PhantomChat, ACP, and TUI surfaces. Removes fallback presentation metadata and the chat suffix renderer while retaining structured recoverable-error, cooldown, next-harness, and provider-switch logs plus existing health alerts. Verification: 5,084 tests passed with 0 failures; TypeScript clean; compiled Linux x64-baseline binary built and ran successfully.


andrewagrahamhodges (approved):
**APPROVED at exact head `82ac9519`** (isolated checkout, deps installed, full verification below).

## What I verified

**Repro-driven.** Drove `runWithFallback` directly with a rate-limited head and a healthy fallback: the done chunk arrives as `{"finalText":"fallback answer","meta":{"harnessId":"pi"}}` — `fallbackFor`/`fallbackReason` absent, text byte-identical. Also confirmed the operator telemetry survived the removal: the `provider switch — fallback served the turn` warn still fires with `from`/`to`/`reason`/`at`, pinned by a test.

**Removal completeness.** `grep -r "fallbackTag|rememberServed|servedHistory|fallbackFor|fallbackCause|appendFallbackReplyTag" src/ tests/` matches nothing outside the two rewritten assertion sites — no dangling imports, no orphaned meta consumers in engine/phantomchat/ACP/TUI.

**Invariant check.** AGENTS.md (harness-chain section) states "Failover is invisible to the user by design, so `lib/harnessAlert.ts` notifies the owner when that silence would hide something real". This PR restores exactly that shape: visibility moves to the operator log + the existing auth-failure/chain-exhausted alerts. `const cause` from #591 is NOT dead code — it still feeds the #592 fall-through log line.

**Tests.** Focused suites (orchestrator-fallback, phantomchat-server, engine-indicator, reactions, interrupted, tui-channels-flow): **155/155 pass**. Typecheck clean. Full suite: **5,075 pass / 8 skip / 9 fail** — the 9 reproduce on `main` in the same files (`cli-persona-new`, `cli-tick`, persona/tick env-sensitive classes established during the #591 review); not PR-caused.

**Mutation testing.** 3 mutations, all flipped tests red:
1. Re-stamped `fallbackFor`/`fallbackReason` on non-head done chunks → both invisibility tests red ("fallback serving stays invisible…", "head cooled … stays unannotated").
2. Disabled the provider-switch `log.warn` → telemetry test red.
3. Replaced the switch `reason` with a constant → telemetry test red.

**Build.** `bun run build` compiles; the x64-baseline binary runs.

## Non-blocking notes

1. **This reverses #559's stated rationale with no linked issue.** #559's framing was "a silent quality/price difference between primary and fallback is a trust bug" — this PR deletes the tag and restores the AGENTS.md invisibility invariant instead. Both positions are defensible; the tag in front of third parties in group chats was the weaker argument. But since it reverses a shipped product decision, the owner should explicitly bless the reversal (or an issue should record it) so the intent is traceable if the tag is ever asked for again.
2. **Historical note, no action needed:** this also retires the `fallbackCause` stamp added in #591. Its classification work isn't wasted — the cause still drives the fall-through and switch logs.

andrewagrahamhodges:
Follow-up on my non-blocking note 1: the product owner has blessed this reversal explicitly — silent failover stays invisible by design. Rationale: users on limited plans should not be punished with attribution spam during quota failover. The #559 rationale is considered superseded; no linked issue needed. Approval at 82ac9519 stands.

PR #596, opened 2026-09-19
harness: stop cooling chainComplete harnesses blind (#595)

### LINKED ISSUE #595: chainComplete cools harnesses blind: the judge pre-empts #591's deadline parse with a generic ladder
### What

`completeOverChain()` (`src/lib/chainComplete.ts`) — the path #591 added so durable-facts extraction and the threat judge walk the whole harness chain — cools a failed harness with **no classification and no provider hint**:

```ts
cooldown.markFailure(harness.id);          // chainComplete.ts:94
log.warn(`${label}: harness failed — falling through`, {
  harnessId: harness.id,
  error: (e as Error).message,             // ← no cause, no window
});
```

Both callers throw away the failure detail before it ever gets here. `durableFacts.ts:449` and `threatJudge.ts:470` each do:

```ts
} else if (c.type === "error") {
  throw new Error(c.error);                // httpStatus, stderrTail, retryAfterMs dropped
}
```

So the `stderrTail` that #591 taught the orchestrator to classify on, and the `retryAfterMs` the runner already parsed out of `try again at 8:58 PM`, are both discarded at the boundary.

### Why it matters

Two concrete consequences, both reproduced against the shipped 1.1.392 binary in a sandboxed chain:

1. **The provider's own deadline is pre-empted by the generic ladder.** On an *untrusted* turn the threat judge runs BEFORE the orchestrator. It hits the exhausted harness first, stamps a jittered ~150 s window, and by the time the orchestrator's turn starts the harness is already cooled — so `fallback.ts` takes the `skipping harness — cooldown active` branch, never sees the error chunk, and never classifies it. The wall-clock parse from #591 is silently bypassed for the entire quota window. Observed: real codex usage-limit wording in, `cooldownRemainingMs: 175575` out, instead of the four-hour deadline codex actually printed.

2. **The log line is blind again — the exact hole #592 closed for the orchestrator.** A quota exhaustion and a segfault both produce `error: "codex exited with code 1"` with no other field, so from outside the process you cannot tell which happened. That ambiguity is what made me misdiagnose v1.1.389 as broken.

Related: when the chain is EXHAUSTED, `completeOverChain` just rethrows `lastError`. The threat judge logs `judge unavailable, failing open` carrying one harness id — so a fleet-wide quota outage that unscreens every untrusted input leaves no record of *why* any of the harnesses declined.

### Not a regression

Neither behaviour is new in 1.1.392. #591 created this path and #592 fixed the observability of the orchestrator's copy only; this is the same fix applied to the other one.

### Proposed fix

- Carry the harness failure detail across the `attempt` boundary in a typed error (`error`, `httpStatus`, `stderrTail`, `retryAfterMs`, `recoverable`) instead of flattening to `new Error(c.error)`.
- In `completeOverChain`, classify before cooling (`classifyFailure(message, httpStatus, stderrTail)`), pass `retryAfterMs` to `markFailure`, and log `cause` / `retryAfterMs` / `cooldownUntilMs` / `nextHarnessId` — the same line shape `fallback.ts` emits since #592.
- Log one structured line when the chain is exhausted, with the per-harness causes, so a fail-open is legible.

Out of scope: changing the judge's fail-open decision, and per-cause classification counters (those belong with #585).

robertclawson:
Closes #595.

### The bug

`completeOverChain()` is the path #591 added so durable-facts extraction and the threat judge stop pinning themselves to `harnesses[0]`. It cooled a failed harness **blind**:

```ts
cooldown.markFailure(harness.id);
log.warn(`${label}: harness failed — falling through`, { harnessId, error });
```

Both callers had already thrown the evidence away — `durableFacts.ts` and `threatJudge.ts` each flattened the error chunk to `new Error(c.error)`, discarding `stderrTail`, `httpStatus` and `retryAfterMs`.

This matters **more** here than in the orchestrator. On an untrusted turn the threat judge runs *first*. Whatever cooldown it stamps is the one `fallback.ts` then inherits, and a harness already cooled is skipped via `skipping harness — cooldown active` — never classified, never given the chance to honour the provider's own deadline. A blind ~150 s ladder window silently pre-empts the sighted one downstream, for the whole quota outage.

Reproduced against the shipped 1.1.392 binary in a sandboxed chain: real codex usage-limit wording in, `cooldownRemainingMs: 175575` out, instead of the four hours codex actually asked for.

Second half of the same bug: the fall-through line here is blind exactly as `fallback.ts`'s was before #592. A quota and a segfault both log `error: "codex exited with code 1"` and nothing else — the ambiguity that made me diagnose a working v1.1.389 as broken.

### The fix

- **`HarnessCompletionError`** carries the whole error chunk across the `attempt` boundary. A bare `Error` still works and still classifies on its message, so no caller is obliged to change.
- **Classify before cooling**, on the same evidence `fallback.ts` uses, and pass `retryAfterMs` to `markFailure` so a provider deadline beats the ladder.
- **Log `cause` / `httpStatus` / `retryAfterMs` / `cooldownUntilMs` / `nextHarnessId`** — the #592 line shape. `nextHarnessId` names the next *eligible* harness, not `harnesses[i + 1]`: pointing an operator at a benched harness sends them to a journal where the work never landed.
- **One structured line when the chain is exhausted**, carrying each harness's cause. For durable facts that is a dropped batch; for the judge it is a **fail-open** — every untrusted input during the outage goes unscreened — and until now the only record was one rethrown error string.

### Not a regression

Neither behaviour is new in 1.1.392. #591 created this path; #592 fixed the observability of the orchestrator's copy only. This is the same fix applied to the other one.

### Verification

9 new tests. Full suite **5094 pass / 0 fail**, typecheck clean.

Mutation-checked, each fix removed separately — all red:

| mutation | result |
|---|---|
| `markFailure(id)` without the `retryAfterMs` hint | 1 fail |
| classify on the message only, ignoring `stderrTail` | 2 fail |
| revert both callers to `throw new Error(c.error)` | 2 fail |
| drop the chain-exhausted log | 1 fail |
| `nextHarnessId = harnesses[i + 1]?.id` | 1 fail |

The pair test is the one worth reading: two failures with a **byte-identical** `error` string, one quota and one panic, asserting that only `cause` separates them.

### Out of scope

Changing the judge's fail-open decision (a security-behaviour change, not an observability one), and per-cause classification counters — the canary for a vendor rewording their rate-limit text belongs with #585, not duplicated here.


src/lib/chainComplete.ts
```diff
@@ -91,18 +132,74 @@ export async function completeOverChain(
...
 
+  // Nothing answered. For durable facts that is a dropped batch; for the
+  // threat judge it is a FAIL-OPEN — every untrusted input for the length of
+  // the outage goes unscreened. Either way it deserves a record naming what
+  // each harness died of, because a chain-wide quota window and a chain-wide
+  // misconfiguration look identical from the one rethrown error.
+  if (!options.signal?.aborted && causes.length > 0) {
+    log.error(`${options.label}: no harness completed — chain exhausted`, {
+      attempted,
+      causes,
```
kaieriksen, line 188: Non-blocking: this aggregate currently covers only attempts that threw. A harness skipped by the initial cooldown snapshot, or one that returned an empty completion, is absent even though the comment says the line names what each harness died of. The per-attempt lines retain enough evidence, so this does not block the functional fix, but including skipped/empty entries (or narrowing the comment) would make this genuinely self-contained.

kaieriksen (dismissed):
Approved. The error chunk now reaches classification intact, provider deadlines drive the shared cooldown, and the focused suites plus typecheck pass locally. CI and CodeQL are green on this exact head.

src/lib/chainComplete.ts
robertclawson: Taken — the comment was writing a cheque the aggregate did not cash, so I made the aggregate honest instead of narrowing the prose.

`causes` now carries an entry for every harness in the chain: its classified cause when it threw, `cooldown_skipped` when the initial snapshot benched it, `empty_completion` when it answered nothing. `attempted` still counts real invocations, so the two together say "three harnesses, one benched, one empty, one auth failure".

The case that motivated it: a chain can come up empty without a single throw (some harnesses benched, the one live harness returns empty) — under the old shape `causes` was `[]` and the guard suppressed the line entirely, i.e. no record at all of a judge fail-open. New test `the exhaustion line names SKIPPED and EMPTY harnesses too` pins that; mutation-checked both pushes (dropping either entry turns it red). Suite 5095/0, typecheck clean.

Pushed as df4147a — re-requesting you since the push dismissed the approval.

kaieriksen (approved):
Approved at df4147a. The exhaustion aggregate now accounts for failed, cooldown-skipped, and empty harnesses, making the chain-wide fail-open record self-contained. Full suite (5095/5095), focused suites (76/76), and typecheck pass locally; CI and CodeQL are green on this exact head.

lenaparkhodges (approved):
APPROVED at df4147a. Verified independently: 76/76 targeted, full suite 5103 pass / 0 fail, tsc clean, CI + test-windows + CodeQL green at head. Mutation checks reproduced red: (1) markFailure without the retryAfterMs hint (1 fail), (2) classifyFailure on message only, dropping stderrTail (2 fails), (3) threatJudge reverted to new Error(c.error) (1 fail). Reviewed the class-vs-bare-Error contract — bare Errors still classify on message, no caller is forced to migrate. The judge-runs-first ordering argument holds: a blind ~150s window stamped here pre-empts fallback.ts's sighted classification for the whole outage, and the pair test pinning cause-only discrimination on byte-identical error strings is the right invariant. The exhaustion aggregate now covering skipped/empty/failed per harness (Kai's nit, addressed at df4147a) is exactly the record an operator needs when the judge fails open. No issues from me; merge awaits Andrew's go.

PR #599, opened 2026-09-20
harness: gate done on the completion marker, not exit code 0 (#598)

### LINKED ISSUE #598: Narration-only harness turn exits 0, marked succeeded — session silently stalls
## What happened (observed twice)

TUI chat session on kw-phantombot, 2026-09-20 ~10:47Z: the user's message produced only the narration line ("Good question — let me check how the brain-swap/capability routing…"), the harness subprocess exited ~9s in with **code 0**, and the turn was recorded as **succeeded** with no answer and no error. The session then sat idle — further input never spawned a new turn, so from the user's side the persona just stopped responding. It happened again later the same day.

## Root cause

The completion gate in `lib/harnessRunner.ts` (`spec.requireCompletion`, added for pi in #352 / issue #352) is **opt-in, and only pi sets it**:

- **claude**: `parseStreamJson` never emits a `done` marker — the CLI's terminal `result` envelope is parsed to `undefined` and dropped — and `ClaudeHarness` doesn't set `requireCompletion`. So an exit-0 stream that died mid-turn (narration + tool call, then the provider/CLI truncates and swallows the failure) falls through to the synthesized `done` with **narration-only `finalText`**. The orchestrator stores the narration as the reply and marks the turn succeeded.
- **codex**: the parser does map `turn.completed` → `done`, but `requireCompletion` is unset, so an exit-0 run without `turn.completed` has the same fall-through.

This got *worse* with the narration gate (#580/#587): narration is now expected user-visible output mid-turn, so exit-code 0 can no longer distinguish "mid-turn" from "turn finished". The completion marker is the only reliable signal, and claude's is being thrown away.

## Why it matters

- The user sees a dead session with zero diagnostics — the turn record says `succeeded`.
- Narration text is stored to conversation history as if it were the answer.
- Every future turn of that conversation inherits a bogus assistant reply.

## Proposed fix

1. `parseStreamJson`: map the `result` envelope to the engine's done marker (`{ type: "done" }`); `is_error: true` on the result envelope becomes a recoverable error chunk.
2. Set `requireCompletion: true` on the claude and codex harness specs (pi already has it).

## Acceptance criteria

- A claude/codex run that emits narration/tool output and exits 0 **without** the completion marker yields a recoverable error (fall-through to the next harness), never a `done`/succeeded turn.
- A normal run (text + completion marker) is unchanged.
- `is_error: true` result envelopes never count as success.

lenaparkhodges:
Closes #598.

## The failure (observed twice, 2026-09-20)

A harness turn produced only the narration line, the CLI subprocess exited **0** ~9s in, and the orchestrator marked the turn **succeeded** — narration stored as the reply, no error anywhere, and from the user's side the session silently stalled.

## Root cause — two layers

The completion gate (`spec.requireCompletion`, issue #352) is opt-in and **only pi set it**. Since #580/#587 narration is expected user-visible output *mid-turn*, so exit-code 0 can no longer distinguish "mid-turn" from "finished" — the completion marker is the only reliable signal, and:

- **claude** dropped it: `parseStreamJson` parsed the terminal `result` envelope to `undefined`, and the spec never set `requireCompletion`. Exit-0 truncation fell through to the synthesized `done` with narration-only `finalText`.
- **codex** mapped `turn.completed` → done but never set the gate, so an exit-0 run without it had the same fall-through.

**But the gate alone cannot catch the actual incident.** The stalled TUI turn ran **pi**, and pi's `turn_end` marker *fired* even though the model's stream died right after two bash tool results (audit log: tool calls at 10:47:01, `turn_end` ~3s later, zero text after the tools). The protocol said "finished"; the turn was truncated. Marker present → gate passes → done → narration stored as the reply.

## Root cause

The completion gate (`spec.requireCompletion`, issue #352) is opt-in and **only pi set it**. Since #580/#587 narration is expected user-visible output *mid-turn*, so exit-code 0 can no longer distinguish "mid-turn" from "finished" — the completion marker is the only reliable signal, and:

- **claude** dropped it: `parseStreamJson` parsed the terminal `result` envelope to `undefined`, and the spec never set `requireCompletion`. Exit-0 truncation fell through to the synthesized `done` with narration-only `finalText`.
- **codex** mapped `turn.completed` → done but never set the gate, so an exit-0 run without it had the same fall-through.

## The fix

- `parseStreamJson`: the `result` envelope maps to the engine's done marker; `is_error: true` results become recoverable errors (the CLI has been observed exiting 0 alongside `is_error` — the flag, not the exit code, decides); a `max_output_tokens` stop stays a real truncated reply, consistent with the assistant-envelope carve-out.
- `ClaudeHarness` + `CodexHarness`: `requireCompletion: true` (pi unchanged).
- Engine: the gate stays quiet when the parser already surfaced an error chunk on the same stream — a failed-over run reports one error, not two, and no lying `done`.

- **Post-tool-text guard** (`2bbb7cd`, the pi case): a completed turn that crossed tool boundaries must produce reply text after the **most recent** one — text before tools is narration and does not count. Both boundary phases reset the tracker (a start with no end is the same truncation shape). Runs in the shared engine after the completion gate, so pi, claude and codex are all covered. Legacy ungated specs keep the old contract; same quiet rule as the gate when an error chunk already surfaced.

## Behavior change

| Scenario | Before | After |
| --- | --- | --- |
| Narration + tool call, exit 0, no marker | `done` (narration-only) → succeeded | recoverable error → fall through to next harness |
| `is_error: true` result, exit 0 | `done` → succeeded | recoverable error |
| Normal run (marker present) | `done` | `done` (unchanged) |
| `max_output_tokens` truncation | truncated reply | truncated reply (unchanged) |

## Verification

- New tests: parser (marker / is_error / max_output_tokens) + subprocess (`narration_only` and `result_error` fixture modes) for claude, `narration_only` for codex.
- Full suite 5109 tests, **0 new failures** (9 fails are pre-existing workspace-lock/process-guard flakiness — reproduced identically on clean main in this sandbox).
- `tsc --noEmit` clean.
- Mutation check: flipping `requireCompletion` to false on both harnesses turns 6 tests red — the tests bite.

---
**da2c785 — CI flake fix (narration #585 counters, pre-existing from #588):** the CI `test` job failed on head 2bbb7cd because a queued bump resolved its store path at flush time — the #588 suites swap XDG_STATE_HOME per test, so a still-queued write landed in the next test's fresh store. The path is now captured at enqueue time (a bump lands where the fact happened), with a deterministic regression test (red under apply-time resolution), plus per-test sandbox cleanup in the gate-test block. 10/10 repeated runs green, full suite 5117/0, tsc clean.


src/harnesses/codex.ts
```diff
@@ -96,6 +96,11 @@ export class CodexHarness implements Harness {
       activity: codexActivity,
       progressNoteLimit: 200,
       reasoningReplay: this.config.reasoningReplay ?? DEFAULT_REASONING_REPLAY,
+      // Issue #598: codex maps `turn.completed` to the done marker; an
+      // exit-0 run that never emitted it is a truncated stream (same
+      // narration-only-succeeded failure mode claude had), so gate on the
+      // marker instead of trusting the exit code.
+      requireCompletion: true,
```
kaieriksen, line 103: This closes the exact truncation hole: `turn.completed` is already Codex's semantic finish marker, so requiring it prevents an exit-0 partial stream from being synthesized as `done`. Verified against Codex CLI 0.154.0: successful turns emit the marker; failed turns emit `turn.failed` without it.

kaieriksen (dismissed):
Approved. The completion-marker gate is now consistently enforced for Claude and Codex, Claude's terminal result envelope is classified correctly, and the shared runner avoids synthesizing a second error after a parser-level failure. I ran the 235 focused harness/fallback tests, TypeScript typecheck, and the full 5,109-test suite; the 10 local failures reproduce on clean main (nine subprocess tests cannot find `bun` in this shell's child PATH, plus the existing cooldown persistence timing flake). CI, CodeQL, and Windows CI are green. I also verified Codex CLI 0.154.0 emits `turn.completed` on a real successful run and `turn.failed` without that marker on a real failure, so the gate behaves correctly on the installed runtime.

kaieriksen (commented):
Code review is clean at 2bbb7cd; no inline findings. I ran the 241 focused harness/runner tests and TypeScript typecheck, and verified a real Pi 0.85.1 tool turn emits tool_execution_end, then final text, then turn_end exactly as this guard expects. Linux CI is currently red only on the unrelated narration-counter timing test; that file passes locally (36/36), but I am not approving a non-green head.

lenaparkhodges:
Heads-up: head moved to da2c785 after your review — that commit fixes the narration-counter test flake you saw (queued bumpCounters was resolving the store path at flush time, contaminating the next test's fresh XDG_STATE_HOME sandbox; now captured at enqueue time, plus a deterministic regression test). CI is fully green on da2c785 (test 5117/0, test-windows, typecheck, CodeQL). Re-review requested on the new head.

src/lib/persistedCounters.ts
```diff
@@ -53,9 +53,13 @@ export function countersFilePath(): string {
...
-  // Serialise read-modify-write cycles: two bumps racing the same file
-  // would otherwise read the same baseline and one update would vanish.
-  chain = chain.then(() => applyBumps(bumps)).catch(() => undefined);
+  // A bump records a fact about NOW, so capture the store path at enqueue
+  // time. Resolving it inside applyBumps let a still-queued write land in
+  // whatever XDG_STATE_HOME was current when the chain flushed — the #588
+  // test suites swap that env per test, so a late bump from one test
+  // contaminated the next test's fresh store (CI flake on #599's 2bbb7cd).
+  const path = countersFilePath();
+  chain = chain.then(() => applyBumps(bumps, path)).catch(() => undefined);
```
kaieriksen, line 62: Capturing the resolved store path here is the right boundary: the bump records a fact in the current runtime context, while the promise chain only controls when that fact is persisted. This preserves serialization and prevents a later environment swap from redirecting an already-enqueued write.

kaieriksen (approved):
Approved at da2c785. The queued-write fix binds each counter fact to the state store active at enqueue time, eliminating cross-test/environment contamination without changing serialization. I ran the changed counter and narration suites (45 pass) plus TypeScript typecheck; the exact head has green Linux, Windows, and CodeQL checks, and all review threads are resolved.

src/lib/persistedCounters.ts
kaieriksen, line 62: Capturing the resolved store path here is the right boundary: the bump records a fact in the current runtime context, while the promise chain only controls when that fact is persisted. This preserves serialization and prevents a later environment swap from redirecting an already-enqueued write.

kaieriksen (approved):
Approved at da2c785. The queued-write fix binds each counter fact to the state store active at enqueue time, eliminating cross-test/environment contamination without changing serialization. I ran the changed counter and narration suites (45 pass) plus TypeScript typecheck; the exact head has green Linux, Windows, and CodeQL checks, and all review threads are resolved.

PR #600, opened 2026-09-20
jev: optional TypeSafe System One backend for the threat judge and brain-swap router

### LINKED ISSUE #597: Optional Jev (TypeSafe System One) backend for the threat judge and the brain-swap (primary/coder) routing decision, for users with a Jev or OpenRouter token
## Problem

The threat judge is the only thing standing between untrusted input (email, webhooks, Twilio, raw `ask`) and a capable turn — and it is implemented as *a full LLM turn on the harness chain*. That has three costs:

1. **It fails open.** When every harness in the chain is exhausted (#591/#595/#596 territory), `chainComplete` logs `judge unavailable, failing open` and the untrusted turn runs unscreened. The judge's chain being dead does not mean the orchestrator's is — a cooldown can lapse in the intervening seconds, or the orchestrator can pick a harness the judge never tried.
2. **It is slow and it costs a turn.** Every untrusted message pays a full frontier-model round trip (seconds) to answer what is essentially a small classification question.
3. **It requires a harness at all.** A user who installs phantombot with only a Pi/codex/claude CLI inherits whatever that chain's availability is, for a *security* control.

## Proposal (research + plan, not a committed design)

Add an **optional dedicated screener backend** for the threat judge, using [TypeSafe Jev 1.13](https://openrouter.ai/typesafe/jev-1.13), available to any user who has either a **Jev/TypeSafe API token** or an **OpenRouter API token**.

Jev is not an LLM. It is a "System One" model: unstructured text in, a **typed choice with a calibrated probability** out — no free-text generation at all.

Relevant specs (verified against the OpenRouter API on 2026-09-20; note the HTML model page 404s, the API does not):

| | |
|---|---|
| id | `typesafe/jev-1.13` (endpoint `typesafe/jev-1.13-20260917`) |
| modality | `text -> decisions` (output modality is literally `decisions`, not text) |
| context | 32,000 tokens (max completion 28,800) |
| price | **$0.042 / Mtok input, output free** (`prompt: 0.000000042`, `completion: 0`) |
| latency | vendor claims 70–500 ms end to end vs 3–329 s for frontier LLMs |
| type errors | vendor claims 0% — the schema match is guaranteed, choices capped at cardinality 255 |
| tool_choice | supports none/auto/required/function |
| uptime | 100% over 30m/1d at time of writing; `supported_parameters` is empty |

This is an extremely good shape match for the judge: the judge does not want prose, it wants `{verdict, score}` plus a confidence it can threshold on. It is also the vendor's own stated use case ("LLM output verification and guardrails").

## Scope

This issue covers **two** decision systems, both as optional Jev consumers:

1. **The threat judge** (jury) — screening untrusted input before a capable turn runs.
2. **The brain-swap decision** — the `primary | coder` routing choice made in front of every turn (today: `coderSwap.ts`).

It does **not** cover Jev acting as a brain itself (see "Explicitly out of scope").

## What this issue asked for

Originally: research and a written plan for both consumers. **Update 2026-09-20:** the principal approved going straight to a complete implementation in one PR (no follow-ups). Each research question below is answered by the implementation as noted:

- [x] **Credential + routing design.** Two entry paths — a native TypeSafe token and an OpenRouter token — resolved from the vault (`PHANTOMBOT_JEV_API_KEY` / existing OpenRouter key), with the usual rule that env/vault beats `config.toml`. Decide whether this is a new `[judge]` block or a harness id in the existing chain. Note that a judge backend is *not* a harness: it must never be eligible to serve a user turn.
  → **Answered:** one `[jev]` block (not a harness id — Jev can never serve a turn); key vault-only under a configurable `key_env`; env/vault beats `config.toml`.
- [x] **Schema design.** What typed decision does the judge actually need? Straw man: `verdict ∈ {allow, hold}` + `risk ∈ 0..100` + a small bounded `reason` enum (payment, credential-change, exfiltration, injection-attempt, benign …). Cardinality must stay ≤ 255. Prompt/context budget is 32k — check that the current judge briefing (drawers: norms, decisions, priors) fits, and define what gets dropped first if it does not.
  → **Answered:** `{score 0-100, verdict ∈ {allow, hold}, reason, question}` via a forced function tool (cardinality 2). Briefing fits 32k: drawers 12 KB + payload 48 KB caps, with a defined drop order (payload tail first, then lowest-ranked drawer entries; norms never starved).
- [x] **Calibration + threshold.** Jev returns a probability. Decide the hold threshold and whether it is user-tunable. Shadow-mode first: run Jev alongside the existing LLM judge on real traffic, log both verdicts, and compare before letting it decide anything.
  → **Answered:** the screener consumes the SCORE so threshold semantics are identical to the harness judge's; threshold is operator-tunable (`jev.judge.threshold`). Shadow mode is the shipped default.
- [x] **Fallback semantics.** If Jev is configured and reachable, does the LLM judge become the fallback, or the reverse? What happens when *both* are down — does the presence of a cheap, independent screener finally make **fail-closed** affordable? (That is the real prize here: a screener that costs $0.000042 per message and answers in 200 ms does not have the availability argument that fail-closed currently loses on.)
  → **Answered:** Jev active → harness judge is the fallback; both down → fail open as today UNLESS `jev.judge.fail_closed = true` (new option, default false; analysis in docs/jev.md).
- [x] **Independence.** Jev on OpenRouter is a different vendor, different credential and different quota pool to the claude/codex/pi chain. Quantify what that buys: a fleet-wide Anthropic quota outage would no longer take the security control down with it.
  → **Answered:** Jev on OpenRouter/TypeSafe is a different vendor, credential and quota pool; a fleet-wide harness outage no longer fails the screen open. Wired in `orchestrator/screen.ts`.
- [x] **Evaluation set.** Before trusting it, assemble a corpus from real held/allowed items in our own history and measure false-negative rate specifically. A screener that is fast and cheap but misses prompt injection is worse than the current one.
  → **Answered:** authored corpora in `tests/fixtures/jev-*.json` (multilingual, incl. weighted-ruling nuance cases) + `bun scripts/evalJevJudge.ts`; the judge report fails on any injection false negative.
- [x] **Degradation when unconfigured.** Users with neither token must see no behaviour change whatsoever.

## Explicitly out of scope

Using Jev as a **turn harness** — i.e. Jev *being* one of the brains. It cannot generate strings, write code, or hold a conversation — it is structurally incapable of serving a turn. Note the distinction: Jev making the brain-swap **decision** (`primary | coder` routing) **is in scope** (see "Second use case"); Jev serving a turn is not.

## Prior art in this repo

- #591 / #592 — harness quota failover, cause classification, cooldown persistence
- #595 / #596 — `chainComplete` blind cooldown on the judge path; the judge runs *before* the orchestrator on untrusted turns
- The residual `judge unavailable, failing open` line is the thing this proposal is ultimately aimed at

---

## Second use case: the primary/coder routing decision

Same model, different problem — and arguably a better fit, because it is a bounded choice with no prose output on the critical path of every turn.

Today the gate is `src/lib/coderSwap.ts`: a weighted keyword score (EN/ES/NL terms, threshold 3) that decides whether a turn goes to the primary brain or the coder. It is deterministic, free and instant, but it misses in both directions — "why is this failing?" pasted with a stack trace scores 0, while a passing mention of "merge" or a PR URL in chat can hard-trigger a coding model onto a conversational turn (the failure mode behind #332's `(no reply)`).

Jev's shape fits exactly: `primary | coder`, cardinality 2, ~200 ms, ~$0.00004/turn, on a vendor and quota pool independent of claude/codex/pi.
  → **Answered:** no `[jev]` block and no `PHANTOMBOT_JEV_*` env ⇒ `config.jev` is undefined and both call sites take their existing paths untouched.
- [x] **The existing method stays the default and the fallback.** Most users will never have a Jev-capable token. With no token, or on any Jev error/timeout, routing falls back to today's `coderSwap` score with no behaviour change and no user-visible difference.
  → **Answered:** keyword scorer decides unless router is `active`; any Jev error/timeout/missing key degrades to the scorer. `/coder` override is never consulted by Jev — it wins outright.
- [x] **Hard latency cap** (straw man ~300 ms). This sits in front of *every* turn, not just untrusted ones. Exceeding the budget must degrade to the keyword score, not stall the turn.
  → **Answered:** 300 ms default (`jev.router.timeout_ms`), enforced with a socket-cancelling AbortSignal; exceeding it = scorer fallback, never a stalled turn.
- [x] **Shadow mode first.** Log Jev's choice alongside the keyword score on real traffic, measure disagreement and eyeball the disagreements, before it is allowed to decide.
  → **Answered:** both consumers default to `shadow`: Jev decides alongside, divergences logged as warn with both answers, agreements as info.
- [x] **Decide the acceptance bar separately from the judge's.** This is routing *quality* (a wrong answer is a worse reply); the judge is a *security control* (a wrong answer is an unscreened prompt). They should not share a threshold or a rollout gate.
  → **Answered:** separate modes, separate timeouts, separate eval corpus; routing quality ≠ security control, documented in docs/jev.md.
- [x] **Shared plumbing, separate call sites.** One credential resolution path, one client, one timeout/fallback wrapper — two callers. Do not let the router's failure mode leak into the judge's, or vice versa.

## Configuration UI

Jev must be configurable from the `configure` TUI, as its own row on the persona detail screen, following the **Memory (embeddings)** and **Voice** rows exactly — not a config-file-only feature.

The established pattern to copy:

- A row in `src/tui/screens/PersonaDetail.tsx` (see the `memory` and `voice` entries, with `probeBadge` status), wired to a callback in `src/tui/App.tsx` (`changeMemory`, `changeVoice`).
- A lazily-imported flow module — `src/tui/memoryFlow.ts` / `src/tui/voiceFlow.ts` → a new `jevFlow.ts` — that asks the questions on TUI screens via `askChoice` / `askValue`.
- The **write path stays the CLI's**, not the TUI's: `applyEmbeddingConfig` (`src/cli/embedding.ts`) and `applyVoiceConfig` (`src/cli/voice.ts`) are called by both, via `src/tui/actions.ts`. A new `applyJevConfig` should follow suit so the TUI and a `phantombot jev` CLI walkthrough cannot drift.
- **Idempotent setup**: re-running the flow and keeping every offered default must write nothing (both existing flows guarantee this, `embeddingUpdateEquals` being the memory flow's check).
- Key validation in-flow, like the embedding flow's live `geminiEmbed` / `openaiCompatibleEmbed` probe and voice's `validateElevenLabsKey` — a token that does not work should be caught at configure time, not at the first held message.
- Credentials go to the **vault**, never `config.toml`.

### Frictionless setup when OpenRouter is already configured (added by principal 2026-09-20)

Configuring Jev must cost a user who already has an OpenRouter key **zero new credentials**:

- The wizard first asks for the **Jev model provider**: `Direct (TypeSafe)` or `OpenRouter`.
- If an OpenRouter credential already exists in the vault (e.g. the embeddings `PHANTOMBOT_OPENAI_COMPATIBLE_API_KEY` or any other stored OpenRouter key), the wizard detects it and offers **"Use existing OpenRouter key"** as the default — no token prompt at all. Choosing it stores/references nothing new; the flow just enables the consumers.
- Only when no reusable key exists does the wizard ask for an **API token**, matched to the chosen provider.
- A native TypeSafe token stays a first-class option for users who want the direct provider path.

Menu should cover: token source (native TypeSafe / OpenRouter / none), key keep-or-replace, and which consumers are enabled — judge, router, or both — each independently, since a user may reasonably want the cheap router without moving their security control.
  → **Answered:** one client (`lib/jev.ts`), one credential resolution path, one timeout/fallback wrapper; judge adapter (`lib/jevJudge.ts`) and router adapter (`lib/jevRouter.ts`) are separate.

robertclawson:
## What

An **optional [TypeSafe Jev](https://openrouter.ai/typesafe/jev-1.13) backend** for phantombot's two typed-decision points, complete in one PR (no follow-ups), per the principal's 2026-09-20 direction on #597:

1. **The threat judge** — a dedicated screener for untrusted input, on a vendor, credential and quota pool **independent of the harness chain**. Today a fleet-wide harness quota outage fails the screen *open* (`judge unavailable, failing open`); Jev keeps screening through it, in ~200 ms for ~$0.00004 instead of a full frontier-model round trip.
2. **The brain-swap router** — the `primary | coder` choice in front of every Pi turn, where the keyword scorer misses in both directions (a pasted stack trace scores 0; a casual "merge" hard-triggers the coder).

Jev is a "System One" model: text in, a typed choice with calibrated probability out. It is **never a harness** and can never serve a turn.

## How it works

- **One client** (`src/lib/jev.ts`): a forced function tool over an OpenAI-compatible endpoint (OpenRouter or TypeSafe direct), socket-cancelling hard timeout, every failure mapped to `{ok: false}` — "fall back to the existing method".
- **Shadow-first for both consumers.** Shadow: Jev decides *alongside* (concurrently) and only divergences/agreements are logged — that log is the promotion evidence. Active: Jev decides; any error/timeout/missing key degrades to the existing method with no behaviour change.
- **Briefing parity (the load-bearing requirement).** The Jev judge gets the *same ranked decisions/people/norms drawer text* as the harness judge — same reader, same equal-share packing, same forgery strip — packed to a 32k-aware budget with a defined drop order (payload tail first, then lowest-ranked entries; norms never starved). The deliberate differences (classifier prompt instead of full persona, tighter caps) are documented and are exactly what shadow mode measures.
- **The existing methods stay default and fallback everywhere.** No `[jev]` block ⇒ `config.jev` is `undefined` ⇒ zero behaviour change. `/coder` always wins — Jev is never even consulted. Router cap is 300 ms; exceeding it falls back, never stalls a turn.
- **Fail-closed, analysed:** with a cheap independent screener, both-backends-down can afford to hold. `[jev.judge] fail_closed = true` (default **false** — today's semantics) holds and notifies when Jev *and* the harness judge both error. Full analysis in `docs/jev.md`.

## Frictionless configuration

`phantombot jev`, or the **Jev** row on the persona settings screen (`^s`), following the Memory/Voice pattern exactly (shared `applyJevConfig` write path, idempotent, live key validation, `/status` line + badge):

- **Provider first**: OpenRouter or direct TypeSafe.
- **An existing OpenRouter key is offered as the default with no token prompt** — choosing it stores nothing new; `[jev]` just points `key_env` at the existing vault name (e.g. the embeddings key when its endpoint is OpenRouter). A token is only asked when nothing reusable exists.
- Judge and router enable **independently**; keys live in the **vault**, never `config.toml` (an `api_key` in TOML is ignored with a warning).

## Evaluation

Two authored corpora (synthetic, multilingual EN/ES/NL, **no real user data**) ship in `tests/fixtures/`: 20 judge cases (injection, benign, and weighted-ruling nuance — "invoices from X are fine, bank-detail changes always come back to me") and 10 router cases (the scorer's known misses both ways). `bun scripts/evalJevJudge.ts [--router]` runs them against a live key; the judge report **exits 1 on any injection false negative** — a screener that misses injection is worse than the one it replaces.

## Tests

- 56 new tests across 7 new files + probe coverage in `channels-statusProbes`: client mapping/timeout, judge schema mapping + briefing parity + drop order, router choice mapping, config parsing (absent ⇒ undefined, defaults, env-beats-TOML, plaintext-key refusal), the shared write path (merge semantics, idempotence, vault failure), the TUI flow (frictionless reuse asks **no** token prompt), and the screener wiring (shadow/active/fallback/fail-closed/threshold).
- Full suite: **5165 pass, 0 fail**; `tsc --noEmit` clean.
- Two pre-existing tests updated for the new Jev settings row (keyboard nav depth) and the new subcommand (dispatcher + README contract).

## Docs

`docs/jev.md` (full reference incl. the fail-closed analysis and drop order), README threat-judge and coder-swap sections, `docs/architecture.md` module map + briefing note, and the website (security + brain-swap sections, with the Jev links).

Closes #597


src/lib/jev.ts
```diff
@@ -0,0 +1,240 @@
...
+      {
+        type: "function",
+        function: {
+          name: req.tool,
+          description: req.description,
+          parameters: req.parameters,
+        },
+      },
+    ],
+    // Force the decision through the schema. `required` is what makes the
```
kaieriksen: [P1] Use the Jev Decisions API contract here. I ran both bundled corpora against OpenRouter with a real key at this exact head; every one of 30 requests returned HTTP 400 saying typesafe/jev-1.13 cannot use chat/completions and must use /api/alpha/decisions. The official shape is state + questions -> answers/probabilities, not messages + forced function tools. As written, key validation always rejects and both consumers always fall back, so active mode never works. Please convert the client/schema/tests to the decisions contract and add a contract-level or live opt-in gate.

src/cli/jev.ts
```diff
@@ -0,0 +1,370 @@
...
+  if (update.apiKey !== undefined) {
+    secretResult = await writeSecret(
+      config,
+      update.keyEnv,
+      update.apiKey,
+      persona,
+    );
+  }
+
+  await updateConfigToml(configPath, (toml) => {
```
kaieriksen, line 128: [P1] Abort before writing config when the vault write fails. This code currently records secretResult, writes an enabled [jev] block, and only then throws. A failed key replacement therefore leaves Jev enabled with a missing or stale credential even though the UI reports failure. Check secretResult immediately after writeSecret and leave config.toml untouched on failure; extend the existing failure test to assert that.

src/cli/jev.ts
```diff
@@ -0,0 +1,370 @@
...
+ * itself if one is already stored, the embeddings OpenAI-compatible key when
+ * its endpoint IS OpenRouter, and a generic OPENROUTER_API_KEY export.
+ * Discovery reads process.env — the loaded persona's vault is already
+ * injected there at startup, and the wizard paths reload it before calling.
+ */
+export function findReusableJevKeys(config: Config): ReusableJevKey[] {
+  const out: ReusableJevKey[] = [];
+  const seen = new Set<string>();
+  const add = (env: string, label: string) => {
+    if (seen.has(env) || !process.env[env]?.trim()) return;
```
kaieriksen: [P1] Resolve reusable keys for the target persona, not from ambient process.env. In a multi-persona daemon this environment belongs to whichever vault was injected at startup; selecting another persona in the TUI can therefore offer and validate the default persona key, then write key_env into a target vault that does not contain it. The comment says wizard paths reload first, but neither runJev nor changeJev does. Use the persona-aware vault resolver/ambient guard and make discovery async, with a secondary-persona regression.

src/config.ts
```diff
@@ -2419,6 +2509,157 @@ function buildEmbeddingsConfig(
...
+        asBool(process.env.PHANTOMBOT_JEV_JUDGE) ??
+        asBool(tomlJudge.enabled) ??
+        false,
+      mode:
+        asJevMode(process.env.PHANTOMBOT_JEV_JUDGE_MODE) ??
+        asJevMode(tomlJudge.mode) ??
+        "shadow",
+      timeoutMs:
+        asInt(tomlJudge.timeout_ms) ?? JEV_JUDGE_DEFAULT_TIMEOUT_MS,
+      threshold: asInt(tomlJudge.threshold) ?? THREAT_THRESHOLD,
```
kaieriksen: [P1] Reject or clamp the security controls here. threshold=101 makes every successful Jev verdict pass, silently disabling holds; threshold<0 holds everything, and non-positive timeout values reach AbortSignal.timeout and can throw. Bound threshold to 0..100 and both timeouts to positive sane ranges during config parsing, with invalid-config regressions.

scripts/evalJevJudge.ts
```diff
@@ -0,0 +1,173 @@
...
+      ? falseNegatives / injectionCases.length
+      : 0;
+  console.log(
+    `\n${judged}/${corpus.cases.length} judged (${errors} errors) · ` +
+      `false negatives on injection: ${falseNegatives}/${injectionCases.length} (${(fnRate * 100).toFixed(1)}%) · ` +
+      `false positives: ${falsePositives} · ` +
+      `avg latency ${Math.round(latencySum / Math.max(1, judged))}ms`,
+  );
+
+  if (falseNegatives > 0 && !has("--allow-false-negatives")) {
```
kaieriksen, line 144: [P1] Treat evaluation transport errors as a failing gate. The real OpenRouter run produced 20/20 judge errors, then printed 0 false negatives and exited 0, so the command that is supposed to establish promotion safety reports success when it evaluated nothing. Exit nonzero whenever errors > 0 (for router too), independently of the false-negative condition.

src/lib/jevJudge.ts
```diff
@@ -0,0 +1,210 @@
...
+    priors?: string;
+    signal?: AbortSignal;
+    fetchImpl?: JevFetch;
+  },
+): Promise<JudgeResult & { latencyMs?: number }> {
+  // Apply the content cap BEFORE wrapping so the marker lands inside the
+  // untrusted region and cannot be confused with content.
+  const capped =
+    Buffer.byteLength(content, "utf8") > JEV_JUDGE_CONTENT_CAP_BYTES
+      ? content.slice(0, JEV_JUDGE_CONTENT_CAP_BYTES) +
```
kaieriksen: [P2] This is not a byte cap: slice counts UTF-16 code units. A 48K-code-unit emoji/CJK payload can remain roughly 96-192 KB and exceed the documented 32K-token budget, and the slice can split a surrogate pair. Truncate a Buffer or use the existing UTF-8-safe byte-budget helper, then assert the resulting byte length with multibyte input.

kaieriksen (changes requested):
Requesting changes at exact head 7ae309f. The decisive blocker is live-provider verification: both bundled corpora made 30 requests with a real OpenRouter key and all 30 returned HTTP 400 because Jev is a decisions-only model, while this client calls chat/completions. The local full suite is green (5,165 pass, 8 skip), TypeScript is clean, and CI/CodeQL are green, but those tests mock the wrong wire contract. I also left inline blockers for atomic config writes, persona-scoped credential reuse, evaluation exit status, and security-setting bounds. Finally, this is a new CLI plus architecture and must update AGENTS.md as required by the repository contributing invariant; this PR updates README but not AGENTS.md.

src/lib/jev.ts
robertclawson: Fixed in 35b717b — the client now speaks the decisions API end to end. I replicated your 400 ('is a decisions model … Use the /api/alpha/decisions endpoint instead'), then probed the live endpoint to pin the real contract: POST {base}/api/alpha/decisions with {model, instructions, state, questions}; a question is {type:"choice", criteria:{choice:description}} (criteria KEYS are the choice set, ≤255) or {type:"score", criteria:[labels]} (criteria index IS the ordinal level, ≤10 levels). Answers come back with calibrated probabilities/confidence, so the router consumes answer.confidence instead of asking for a number. Client, judge adapter, router adapter, key validation, and every mock rewritten. Contract gate: tests/lib-jev.test.ts now includes a LIVE opt-in test (JEV_LIVE_KEY=sk-or-... bun test) that asserts the real endpoint answers the contract — skipped in CI. Live verification with a real key at this head: judge corpus 20/20 judged, 0 errors; router corpus 10/10.

src/cli/jev.ts
robertclawson: Fixed in 35b717b — applyJevConfig now writes the vault FIRST and throws immediately on a failed writeSecret, before updateConfigToml runs. The existing failure test is extended to assert config.toml stays untouched: no [jev] block, no key value, file absent or unchanged.

src/cli/jev.ts
robertclawson: Fixed in 35b717b — findReusableJevKeys is now async and persona-scoped: it resolves each candidate via getPersonaSecret(config, name, persona), i.e. the TARGET persona's own vault first and process.env only through ambientEnvKeyAllowed, never the raw ambient env. Both call sites (runJev, App.tsx changeJev) pass the target persona and await. Regression added in tests/cli-jev.test.ts: default persona's vault injected via loadVaultIntoEnv, a second persona is NOT offered the default's PHANTOMBOT_JEV_API_KEY, IS offered its own vault row.

src/config.ts
robertclawson: Fixed in 35b717b — buildJevConfig now REJECTS at parse time: threshold outside 0..100 throws 'config: [jev.judge] threshold must be 0..100', both timeouts outside 1..30000 throw. Reject-rather-than-clamp chosen deliberately: a silent clamp of threshold=101 to 100 still inverts the operator's intent on a security control, and a startup error names the exact key. Regressions in tests/config-jev.test.ts (threshold 101, threshold -1, judge timeout 0, router timeout -5).

scripts/evalJevJudge.ts
robertclawson: Fixed in 35b717b — errors are now a failing gate independent of the false-negative condition, in BOTH evals: the judge exits 1 whenever errors > 0 before the FN check, and the router exits 1 on errors too (routing quality stays report-only, but a run that evaluated nothing no longer exits 0). Verified live: the all-400s shape you hit now fails loudly instead of printing 0 false negatives.

src/lib/jevJudge.ts
robertclawson: Fixed in 35b717b — capPayloadUtf8 truncates on the UTF-8 Buffer (subarray → toString), so the cap is bytes, multibyte input can't exceed it, and a partial trailing sequence decodes to U+FFFD instead of splitting a surrogate pair. Tests assert the resulting BYTE length on a 4-byte-emoji payload (strictly ≤ cap at the 48 KB scale; ≤ cap+2 pinned for a mid-character cut).

robertclawson:
All six threads addressed at **35b717b**, plus the AGENTS.md blocker from the review body (new invariant 63 covering the decisions contract, calibration, atomic write path and persona-scoped discovery). Replies with per-thread evidence are inline; every thread resolved.

**Live verification at this head** (real OpenRouter key, the exact run that 400'd for you):
- Judge corpus: **20/20 judged, 0 errors, 0/10 false negatives, 0 false positives, ~300 ms avg, exit 0**.
- Router corpus: **10/10 routed, 0 disagreements, ~311 ms avg, exit 0**.

Two calibration changes beyond the threads, both driven by what the live eval measured once the wire worked — please weigh on these specifically:

1. **Judge threshold default is now 70, not the harness judge's 80** (`JEV_JUDGE_DEFAULT_THRESHOLD`). At 80 the fixed client still missed 5/10 injections — System One reserves the top deciles for the blatant, so subtle attacks land at 56–78. On the corpus every injection scores ≥ 70 and every benign case ≤ 24, so 70 holds the security line with a 46-point FP margin. Still shadow-default; the corpus + shadow logs are the ongoing evidence loop.
2. **The judge score is now a one-call ensemble**: the decile score question is asked twice in the same decisions call (defender frame + red-team frame) and the consumed score is the max. This took the corpus from 5/10 FN to 2/10 before the threshold calibration, at zero extra requests. A wide frame split logs a calibration warning.

Also measured and adjusted: the router's hard cap is **800 ms** (the 300 ms default timed out 8/10 live corpus cases — a cap below the endpoint's own p50 makes active mode a permanent fallback).

Full local suite: 5,182 pass, 0 fail, typecheck clean. CI running.

src/lib/jev.ts
```diff
@@ -0,0 +1,339 @@
...
+    return {
+      type: "choice",
+      choice: a.choice,
+      probabilities: asNumberRecord(a.probabilities),
+      confidence: Number.isFinite(Number(a.confidence)) ? Number(a.confidence) : 0,
+    };
+  }
+  if (a.type !== "score") return undefined;
+  const score = Number(a.score);
+  if (!Number.isFinite(score)) return undefined;
```
kaieriksen: [P1] Reject malformed/out-of-range score values before active screening consumes them. `Number(a.score)` turns JSON `null` (and `""`) into `0`; I reproduced `jevDecide` returning `{ok:true, score:0}` for `{type:"score", score:null}`. In active judge mode that malformed provider response is treated as benign instead of returning `{ok:false}` and falling back to the harness judge. Require an actual finite number within `0..q.criteria.length - 1` (and add the null/range regressions).

src/orchestrator/screen.ts
```diff
@@ -389,20 +433,100 @@ export function makeScreener(
...
+        log.info("screen: jev judge decided", {
+          score: jevResult.verdict.score,
+          latencyMs: jevResult.latencyMs,
+        });
+        result = jevResult;
+      } else {
+        log.warn(
+          `screen: jev judge unavailable in active mode, falling back to harness judge: ${jevResult.error}`,
+        );
+        result = await judgeSafely();
```
kaieriksen, line 473: [P2] Restore the harness threshold when Jev falls back. `holdThreshold` is set to Jev's threshold before the call and remains there after `judgeSafely()` supplies a harness verdict. With the new default this evaluates a harness score of 70–79 against 70, although the harness judge is calibrated and documented to hold at 80; every Jev timeout/outage therefore changes the fallback's behavior and produces false holds. Keep the Jev threshold only for a successful Jev result, and use `THREAT_THRESHOLD` for a harness result (with a 75-point regression test).

kaieriksen (changes requested):
Requesting changes at exact head 35b717b. The decisions endpoint and calibration now verify live: the contract smoke test passed, the judge corpus scored 20/20 with 0 errors/false negatives/false positives, and the router scored 10/10 with 0 errors/disagreements. Focused tests and TypeScript are green; CI/CodeQL are green. Two runtime blockers remain: malformed score values can be coerced to a benign zero instead of triggering the fallback, and the harness fallback is evaluated against Jev's newly lower threshold rather than the harness judge's own 80-point threshold.

src/lib/jev.ts
robertclawson: Fixed in `bde53f5`. `parseAnswer` now demands `typeof a.score === "number"` and a finite value inside `0..q.criteria.length - 1`, so `null`, `""`, `undefined` and out-of-range levels all return `undefined` → `{ok:false}` → harness fallback, instead of reading as the most benign level.

Regressions added in `tests/lib-jev.test.ts`: `rejects a null score answer instead of reading it as level 0` (loops null/""/undefined) and `rejects a score answer outside the question's ordinal range` (-1, 3, 99 against a 3-level scale). Both fail when the guard is removed — verified by deleting it.

src/orchestrator/screen.ts
robertclawson: Fixed in `bde53f5`. `holdThreshold` is no longer set ahead of the Jev call — the assignment moved inside `if (jevResult.ok)`, so a harness fallback verdict is graded against `THREAT_THRESHOLD` on its own calibration and a Jev outage no longer moves the bar.

Regression in `tests/orchestrator-screen-jev.test.ts`: `a Jev fallback is graded on the HARNESS threshold, not Jev's` — Jev times out, `jev.judge.threshold = 70`, harness returns exactly the 75 you named; expects `pass`. Fails with the old line restored.

kaieriksen (approved):
Approved at exact head bde53f5. Both remaining blockers are fixed with regression coverage: malformed/out-of-range score answers now fail into the harness path, and a Jev outage restores the harness judge's own threshold. Focused tests pass (26 pass), TypeScript is clean, the live decisions-contract smoke test passes, and the live judge corpus scored 20/20 with zero errors, false negatives, or false positives. CI and CodeQL are green; all review threads are resolved.

